PR #20, opened 2015-01-28
Pressing Alt + ←/→ does not interfere now with browser

### LINKED ISSUE #19: Jumping one word forward/backward using Alt + →/← goes forward/backward to browser history
The event should prevent default [here](https://github.com/sourcelair/xterm.js/blob/master/src/xterm.js#L2203)

This issue is present in Windows/Chrome.

akalipetis:
Previously, when on Chrome for Windows, if default was not prevented the browser history backward/forward was triggered.

Fixes #19

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -2201,7 +2201,7 @@
         // left-arrow
         case 37:
           if (ev.altKey) {
-            ev.preventDefault();
+            this.cancel(ev, true);
             key = '\x1bb' // Jump a word back
             break;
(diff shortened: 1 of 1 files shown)


parisk:
Could you please change `ev.preventDefault()` with `cancel(ev)` please? This complies with the overall code styling of the file.

The `cancel` function is being defined at https://github.com/sourcelair/xterm.js/blob/master/src/xterm.js#L4653

After this, I am okay to merge.

parisk:
:cake:

Issue #28, opened 2015-07-02
Fire double backspace event in Safari


alexeyfrank:
Hi, I found a bug in Safari. When I press backspace, "keypress" event fired in Safari. But in Chrome this event is not fired. 

We can fix it in keyPress function:

```

     if (ev.charCode) {
       key = ev.charCode;
     } else if (ev.which == null) {
       key = ev.keyCode;
     } else if (ev.which !== 0 && ev.charCode !== 0) {
       key = ev.which;
     } else {
       return false;
     }

    // fix backspace in Safari
    if (key === 8) {
      return false;
    }

```

TDaglis:
Is this still an issue?

TDaglis:
Closing due to inactivity. If the problem still exists, reopen

PR #29, opened 2015-07-02
Fix double backspace in Safari

alexeyfrank:


src/xterm.js
```diff
@@ -4782,4 +4786,4 @@
     Terminal.cancel = cancel;
 
     return Terminal;
-});
\ No newline at end of file
+});
```
parisk: Could you please enter a new line here to comply with the code styling of the repository?

parisk:
Thanks for your PR @alexeyfrank :smile:.

I assume that this PR closes #28.

xterm.js seems to work fine in Safari when using backspace though. Can you help me reproduce this issue?

alexeyfrank:
@parisk sorry, i found a bug in my code. If our css accidentally make textarea in terminal hidden - width: 0, height: 0  - then we can't focus on this textarea in Safari.

parisk:
@alexeyfrank sounds like this is a bug of your application and not of xterm.js. Are you okay with closing this PR if this is the case?

PR #31, opened 2015-08-25
Fixed IE issues with classList.add

akalipetis:
`classList.add` in IE does not take more than one parameter and this would not work. Split in 3 statements to fix it.


parisk:
LGTM :cake:

PR #32, opened 2015-09-08
Update main.js

chani:
just a tiny change, if you are using the demo file as example to start working with, prompt as variable might cause trouble if you do use prompt() rather use shellprompt as variable name instead of prompt. I'd suggest to rename term's prompt() as well to not cause confusion.


parisk:
You 're totally right @chani. Thanks for the patch :smile:.

PR #36, opened 2015-10-22
Added lenient option to linkify

TDaglis:
This PR introduces the ability to chose between lenient and strict url recognision.
By strict I mean that the url **must** have match a protocol clause, while with lenient the functionality remains unchanged.
If not provided, `lenient` defaults to true.

### DIFF
--- a/addons/linkify/linkify.js
+++ b/addons/linkify/linkify.js
@@ -7,14 +7,14 @@
     } else {
         /*
          * Plain browser environment
-         */ 
+         */
         linkify(this.Xterm);
     }
 })(function (Xterm) {
 	'use strict';
 
     /**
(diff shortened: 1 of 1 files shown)


parisk:
What about keeping only the strict URL recognition to make it simpler?

Some tests that we did together yesterday showed that the lenient parsing is wrong in many cases.

Anyway this gets definitely a LGTM :cake:, so if you prefer keeping both options let's keep it that way.

TDaglis:
I prefer keeping them both and fixing the lenient parsing. Since fixing the lenient parsing won't be that easy, I'm merging for now and I'll open a new PR with a fix for lenient :smiley:

PR #38, opened 2015-10-26
Fix xterm focus events

TDaglis:
blur and focus events are now emitted by the blur and focus events of the textarea

EDIT: also added .editorconfig to preserve uniform code styling


akalipetis:
LGTM

PR #40, opened 2015-10-29
Better filtering of input

TDaglis:
In default case, do not treat `Ctrl` + `Shift` + `<character/symbol>` the same way as `Ctrl` + `<character/symbol>`.
This lead to cases like `Ctrl` + `Shift` + `C` being handled like `Ctrl` + `C`

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -2346,7 +2346,7 @@
           break;
         default:
           // a-z and space
-          if (ev.ctrlKey && !ev.shiftKey) {
+          if (ev.ctrlKey && !ev.shiftKey && !ev.altKey && !ev.metaKey) {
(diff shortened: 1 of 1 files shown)


src/xterm.js
```diff
@@ -2346,7 +2346,7 @@
           break;
         default:
           // a-z and space
-          if (ev.ctrlKey) {
+          if (ev.ctrlKey && !ev.shiftKey) {
```
parisk: Should you handle `ev.altKey` and `ev.metaKey` (Cmd/Start button) not being pressed as well?

TDaglis: Sure. I'm updating it

TDaglis:
@parisk done

parisk:
LGTM :cake:

PR #42, opened 2015-11-03
Fixed width of fit

### LINKED ISSUE #35: Fit addon does not work properly when terminal container has heavily customized text-related CSS

TDaglis:
Fit now utilizes the full width provided.
Also removed firefox hack since it is no longer needed

fixes #35


akalipetis:
LGTM

PR #43, opened 2015-11-10
Fixed marginal width geometry

TDaglis:
#42 was an attempt to better utilize the full width provided and not cut the lines short.
The new calculations were too marginal and there were cases were a `-` character would introduce an unnecessary line break.


parisk:
:+1:

PR #44, opened 2015-11-11
Cursor is now invisible in IE

### LINKED ISSUE #18: Terminal cursor is visible in Internet Explorer
screenshot 2015-01-21 22 46 35

TDaglis:
The change can be easily read in the second commit
fixes #18

also added *.css rules in .editorconfig for code-style consistency


akalipetis:
Could you please provide some reasoning/documentation on how/why this works please? A comment in the code would be event better.

TDaglis:
Well the issue here is that on the top left of the terminal we have the small helper-textarea and right on top of it the xterm.
Now even though helper-textarea is supposed to be hidden, IE for some reason displays the cursor of the focused helper-textarea.
What happens here is that the text area is moved so far to the left that it is no longer displayed on the screen, so even IE can no longer display the cursor

TDaglis:
Added a comment to explain that in the code as well

akalipetis:
LGTM :+1:

PR #49, opened 2016-04-11
nodejs module definition

noseglid:
Allow importing xterm into  a nodejs application (for instance an electron
application, or perhaps a test environment using phantomjs which
also has a DOM),


parisk:
Awesome. Thanks a lot!

PR #51, opened 2016-04-20
Fix cursor blink

blink1073:
Otherwise this line always bails: https://github.com/sourcelair/xterm.js/blob/master/src/xterm.js#L1123

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -379,6 +379,8 @@
 
     Terminal.options = {};
 
+    Terminal.focus = null;
+
     each(keys(Terminal.defaults), function(key) {
       Terminal[key] = Terminal.defaults[key];
       Terminal.options[key] = Terminal.defaults[key];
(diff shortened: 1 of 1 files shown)


akalipetis:
Shouldn't you also include `Terminal.focus = null;` somewhere here?
https://github.com/sourcelair/xterm.js/blob/master/src/xterm.js#L415

This seems like an old term.js logic that was partially removed.

blink1073:
@akalipetis, thanks, updated.

akalipetis:
LGTM 🐻

blink1073:
Thanks @parisk, are you at a point where you can make a release?

parisk:
Yes, I just published the 0.33 release: https://github.com/sourcelair/xterm.js/releases/tag/0.33.

blink1073:
Great, thanks!

PR #53, opened 2016-05-03
Make all add-ons CommonJS importable

### LINKED ISSUE #52: Make all files importable with CommonJS
All we have to do actually is apply #49 to all add-ons.

parisk:
Make all add-ons importable with CommonJS. Both of the following coding snippets should work:

## Example for the attach add-on
```javascript
var attach = require('./addons/attach');
```

or

```javascript
var attach = require('./addons/attach/attach');
```

This PR closes #52.

### DIFF
--- a/addons/attach/attach.js
+++ b/addons/attach/attach.js
@@ -7,7 +7,12 @@
  */
 
 (function (attach) {
-    if (typeof define == 'function') {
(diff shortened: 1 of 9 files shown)


akalipetis:
LGTM 🐻

PR #56, opened 2016-05-26
Explicitly adjust height of terminal rows and child elements

### LINKED ISSUE #54: Style issue with ansi characters
Greetings.

I'm sure someone who enjoys css more than me has a better fix for the following, but when I have a terminal app (mcedit is a good example) that brings up a dialog with ansi characters in it the padding seems to blow out.

image

I can fix it with 
```css
.xterm-rows div {
    height: 16px;
}
```
but I imagine it's not the best fix (breaks scaling)

parisk:
Closes #54

### DIFF
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -2134,3 +2134,19 @@
 .terminal .xterm-bg-color-255 {
     background-color: #eeeeee;
 }
+
+/**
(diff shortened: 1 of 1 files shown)


akalipetis:
LGTM 🐻

PR #58, opened 2016-05-30
Proper rendering of inverse data in default colors

### LINKED ISSUE #57: Inverse text styles are not applied correctly
I'm using xterm.js in conjunction with [ssh2](https://github.com/mscdex/ssh2) to build a SSH client, which has been working quite well so far (thanks for the awesome library!). One issue that I have encountered though is that inverse text styles appear as normal text. For example, on nearly any terminal, running `nano` will show the nano editor with a white status bar across the top implemented using inverse styles. In xterm.js, these appear as normal text instead. This issue also appears when using the [colors.js](https://github.com/Marak/colors.js) library.

Peeking into the code, the problem seems to lie in the styles being applied by [`Terminal.prototype.refresh()`](https://github.com/sourcelair/xterm.js/blob/master/src/xterm.js#L1058). For inverse text, the classes applied to the surrounding `<span>` in each `div.xterm-rows` is `xterm-bg-color-257  xterm-color-256` neither of which are defined in `xterm.css` and aren't valid xterm-256 colors. The correct classes should be `xterm-bg-color-15  xterm-color-0` for a white background with black text.

For nano, the `data` variable that sets the inverse text in `Terminal.prototype.refresh()` is 2228992 (I'm not sure if that helps but that seems to be the value that causes `bg` and `fg` to evaluate to incorrect values).

parisk:
Fix #57

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -968,6 +968,17 @@
     };
 
 
+    /**
+     * Flags used to render terminal text properly
+     */
+    Terminal.flags = {
+      BOLD: 0b00001,
+      UNDERLINE: 0b00010,
+      BLINK: 0b00100,
+      INVERSE: 0b01000,
+      INVISIBLE: 0b10000
+    }
+
     /*
      * Rendering Engine
      *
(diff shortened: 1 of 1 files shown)


TDaglis:
LGTM :panda_face:

PR #60, opened 2016-06-01
Fix cross platform input problems

runarberg:
- Fifth level keys were not working. That prevented some inputting
  important characters (like pipe `|` and caret `^`) on some keyboards.

- Control characters didn't prevent default browser behavior on some
  platforms. `ctrl + D` for example sent the EOT *and* opened the
  bookmark section on MS Edge.

- Meta key behavior simply wasn't working on Mac OS. There is some logic
  that maps `metaKey` on Mac OS to the `altKey` on other
  systems. However that logic didn't work.

Instead of parsing the user-agent string to find the users os, we now
look into the `platform` attribute of the `navigator` object.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -659,10 +659,25 @@
       * Parse User-Agent
       */
       if (this.context.navigator && this.context.navigator.userAgent) {
-        this.isMac = !!~this.context.navigator.userAgent.indexOf('Mac');
-        this.isIpad = !!~this.context.navigator.userAgent.indexOf('iPad');
-        this.isIphone = !!~this.context.navigator.userAgent.indexOf('iPhone');
         this.isMSIE = !!~this.context.navigator.userAgent.indexOf('MSIE');
+      }
+
+      /*
+      * Find the users platform. We use this to interpret the meta key
+      * and ISO third level shifts.
+      * http://stackoverflow.com/questions/19877924/what-is-the-list-of-possible-values-for-navigator-platform-as-of-today
+      */
+      if (this.context.navigator && this.context.navigator.platform) {
+        this.isMac = contains(
+          this.context.navigator.platform,
+          ['Macintosh', 'MacIntel', 'MacPPC', 'Mac68K']
+        );
+        this.isIpad = this.context.navigator.platform === 'iPad';
+        this.isIphone = this.context.navigator.platform === 'iPhone';
+        this.isMSWindows = contains(
+          this.context.navigator.platform,
+          ['Windows', 'Win16', 'Win32', 'WinCE']
+        );
       }
 
       /*
@@ -2337,12 +2352,16 @@
         return this.cancel(ev);
       }
 
-      if (result.cancel) {
+      if (isThirdLevelShift(this, ev)) {
+        return true;
+      }
+
+      if (result.cancel ) {
         // The event is canceled at the end already, is this necessary?
         this.cancel(ev, true);
       }
 
-      if (!result.key || (this.isMac && ev.metaKey)) {
+      if (!result.key) {
         return true;
       }
 
@@ -2532,7 +2551,8 @@
               // ^] - group sep
               result.key = String.fromCharCode(29);
             }
-          } else if ((!this.isMac && ev.altKey) || (this.isMac && ev.metaKey)) {
+          } else if (!this.isMac && ev.altKey && !ev.ctrlKey && !ev.metaKey) {
+            // On Mac this is a third level shift. Use <Esc> instead.
             if (ev.keyCode >= 65 && ev.keyCode <= 90) {
               result.key = '\x1b' + String.fromCharCode(ev.keyCode + 32);
             } else if (ev.keyCode === 192) {
@@ -2571,7 +2591,9 @@
         return false;
       }
 
-      if (!key || ev.ctrlKey || ev.altKey || ev.metaKey) {
+      if (!key || (
+        (ev.altKey || ev.ctrlKey || ev.metaKey) && !isThirdLevelShift(this, ev)
+      )) {
         return false;
       }
 
@@ -4398,6 +4420,15 @@
      * Helpers
      */
 
+    function contains(el, arr) {
+      for (var i = 0; i < arr.length; i += 1) {
+        if (el === arr[i]) {
+          return true;
+        }
+      }
+      return false;
+    }
+
     function on(el, type, handler, capture) {
       if (!Array.isArray(el)) {
         el = [el];
@@ -4452,6 +4483,15 @@
         if (obj[i] === el) return i;
       }
       return -1;
+    }
+
+  function isThirdLevelShift(term, ev) {
+      var thirdLevelKey =
+          (term.isMac && ev.altKey && !ev.ctrlKey && !ev.metaKey) ||
+          (term.isMSWindows && ev.altKey && ev.ctrlKey && !ev.metaKey);
+
+      // Don't invoke for arrows, pageDown, home, backspace, etc.
+      return thirdLevelKey && (!ev.keyCode || ev.keyCode > 47);
     }
 
     function isWide(ch) {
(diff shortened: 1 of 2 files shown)


runarberg:
I've been having some serious cross platform inconsistencies using this package. Most serious is not being able to input fifth-level keys with my Icelandic keyboard. Meaning I'm unable to enter keys like the pipe `|`, tilde `~` or backtick `` ` ``. Also on some my mac the meta key isn't working (I have to press `esc` + `d` for "forward delete word" for example, while `alt + d` works on my linux). And on top of that some control sequences trigger browser side-effects (like `ctrl + k`, "kill to end of line", opens a new tab with current URL on Edge on my Windows machine).

I've tested this with Icelandic keyboard on firefox 46.0.1 and chrome 50.0.2661.102 (64-bit) on Mac OS X 10.11.5 as well as on MS Edge on Windows 10.

Note that this brakes copy-pasting using `ctrl + C` and `ctrl + V` on windows and `cmd + C` and `cmd + V` on Mac OS X. You can still copy by right click → `copy` on all browsers, and paste by `edit` → `paste` in chrome and (sometimes) in Edge. This can (maybe) be fixed by creating a custom context menu on terminal that includes a paste button, that will paste from the clipboard into the textbox.

src/xterm.js
```diff
@@ -2411,9 +2430,19 @@
...
+      if (!key || fifthLevel) {
         return true;
+      } else {
+        // Terminal takes presidence of control key (i.e. ``ctrl + D''
+        // won't trigger bookmark on windows).
+        //
+        // TODO: This also brakes copy-pasting into the terminal using
+        // ``ctrl + C'' and ``ctrl + V''. Find a different way of
+        // allowing users to copy-paste intuitively.
+        this.cancel(ev, true);
```
runarberg: @parisk If I remove this part, would you consider merging the rest?

runarberg:
@parisk I don't know if the `language` and `languages` attributes will help for a few reason, at least not for detecting the input.

1. There is not a clear mapping between language and keyboards. Some languages have several popular layouts, for example, some Esperanto layouts use dead keys to enter esperanto specific characters, while others override the non-esperanto characters (<kbd>y</kbd> → <kbd>ŭ</kbd>) and then use the fifth level to reach the english equivalent.
2. A user might have a preferred language, but still prefer a keyboard language in another language. Because the task at hand is non-optimal using their preferred language’s keyboard layout.
3. A user might have a custom layout that deviates from their preferred language, for example a default keyboard layout for a particular language might use dead keys to enter symbol, but a user might have remapped them to use a fifth level key.
4. A user might simply fail to set a preferred language in their system, and therefor the attributes will give a false image of the user.

In this PR I opted for detecting the system, rather than the preferred language, because generally the popular OS platforms are consistent in how a fifth-level key is reached:

* Windows machines use <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>key</kbd>, also mapped to <kbd>AltGr</kbd>.
* Mac OS machines use <kbd>⌥ Option</kbd> + <kbd>key</kbd>.
* Linux machines (X windowing system) have a native notion of `ISO_Level5_Shift` that is usually mapped to the <kbd>AltGr</kbd> on modern keyboards (although for us it doesn't matter since the DOM event handles it correctly).

runarberg:
I amended the commit. I stopped preventing the browsers default behavior, stopped referring to the *“fifth-level”* (after some investigation, it is actually called a *“third level shift”*), and simplified the commit message.

parisk:
Thanks for the info and commits @runarberg. What you are writing sound quite rational. We just have to make sure that they keep the desired user experience for most of potential users.

I will give a deeper look into this today and also make a local rebase with #78 to see if everything works 🆗 .

parisk:
I pulled this branch, checked it out, rebased with master and still I cannot copy and paste using my keyboard on OS X.

The existing experience should not break for the majority of users. @runarberg can you please list all behavior that is expected to break after your latest patches and possible workarounds for each of them?

runarberg:
@parisk This is because we are using <kbd>⌘ Cmd</kbd> on Mac OS as <kbd>Alt</kbd> on other platforms. If you prevent default behavior on the terminal key event, this key won't emit default behavior. For example <kbd>⌘ Cmd</kbd> + <kbd>C</kbd> will act as <kbd>Alt</kbd> + <kbd>C</kbd> on other systems, and emit “Capitalize word from cursor” to the terminal instead of “Copy”  to the browser window.

This is the only expected behavior that brakes that I can think of, unless some windows users have browser default behavior to <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>key</kbd>.

runarberg:
I amended the commend, stopped using Mac OS <kbd>⌘cmd</kbd> as <kbd>Alt</kbd> on other systems (now Mac OS users have to use <kbd>Esc</kbd> like before), and rebased onto upstream/master. Keyboard copy-pasting works again on Macs.

Ps. I discovered an unintended side-effect where chrome users on ubuntu can use <kbd>⇧shift</kbd> + <kbd>Ctrl</kbd> + <kbd>V</kbd> to paste (<kbd>⇧shift</kbd> + <kbd>Ctrl</kbd> + <kbd>C</kbd> wont work for copying, nor does this work on firefox, nor edge on windows).

parisk:
Thanks a lot for making progress on this. I move forward with pulling your branch and testing right now.

parisk:
Okay the only issue that I found out is that <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>V</kbd>. Won't work for pasting in Firefox. Was that just me or happens to you as well?

Also I think that we can manage with making <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>C</kbd> work, as it invokes the developer tools by default.

runarberg:
@parisk I don't think <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>key</kbd> should work as a replacement for <kbd>Ctrl</kbd> + <kbd>key</kbd> in firefox on ubuntu, it didn't doesn't do so upstream either. It works as a replacement in chrome (and *Terminator*) for some reason.

We could bind <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + {<kbd>V</kbd>, <kbd>C</kbd>} to `document.execCommand("paste")` and `document.execCommand("copy")` (again paste won't work in firefox unless the user changes their settings), but I don't believe that belongs in this PR. Ubuntu and Windows users could also use <kbd>Shift</kbd> + <kbd>Insert</kbd> to paste, but xterm.js will send a `\x1b[2~` to the terminal (I can send another PR that fixes that, but it doesn't really belong here since it is also the behavior upstream).

runarberg:
I submitted #107 to fix keyboard copy-pasting on non-macs.

runarberg:
Ammended: Fixed a bug in the last patch, and rebased onto upstream.

parisk:
Hi @runarberg, thanks again for your patches. It seems that <kbd>Alt</kbd> + <kbd>←</kbd>/<kbd>→</kbd> does not work any more to move back and forward with words (tested on Chrome on Mac).

Is there any workaround for this?

parisk:
Also could you please provide us some documentation/reference in order to see how third level keys should work?

This way we will be able to test the new functionality on top of the existing one and make sure things won't break for mainstream users.

runarberg:
Amended: Added unit tests, and rebased on top of upstream.

runarberg:
@parisk I don't really know much about keyboard input methods, except from experience of using ISO keyboard, and frustration when ANSI keyboard is forced upon me. But [here is a wikipedia article](https://en.wikipedia.org/wiki/ISO/IEC_9995#Levels_.28.E2.80.9Cunshifted.E2.80.9D_.E2.80.9Cshifted.E2.80.9D.2C_.E2.80.9CAltGr.E2.80.9D.29) about the ISO keyboard standard.

parisk:
👍  thanks a lot for one more time. This seems to be working great for now. Merging.

PR #78, opened 2016-06-02
Drop hidden textarea for input handling

### LINKED ISSUE #34: Selection disappears after copying

### LINKED ISSUE #66: Copying terminal output copies all trailing whitespace
Copying this:

image

Pasted result:

image

### LINKED ISSUE #67: The .xterm-helpers element containing textarea prevents selection in top left corner of terminal
Cursor is over the second character `y` in this image:

image

parisk:
This Pull Request drops completely the hidden textarea for input handling. It relies completely on keyboard event handling on the `element` attribute of the terminal. This simplifies the structure of the code and removes a few weird bugs of the previous implementation.

The only hack that had to be done here was to allow pasting with right click. Using `contentEditable` was not a choice as we would not be able to get rid of the default cursor applied on the terminal by the browser. To handle right-click paste I set the `contentEditable` attribute of the terminal to `true` only on right click and then unset it on `mouseup` this way the default cursor is only seen for an instance, while pasting on the terminal with right click is made possible. 

This PR deals with the following issues:

- Closes #34 
- Closes #67
- Closes #66


TDaglis:
I encountered various issues with this PR. I'll try to pinpoint their cause and document them here tomorrow.

parisk:
Just pushed an update that fixes several pasting issues across different browsers. @TDaglis could you please take a look at Windows?

TDaglis:
Sure. I'm on it

TDaglis:
Issues I faced after the update:
- the blinking caret is visible for a while after you click the terminal
- the focus event is not triggered anymore (it should be moved from the textarea to element instead of being removed completely)
- cut works :sweat_smile:

parisk:
Just pushed a commit that should fix all cut and copy issues (it also trims white spaces when copying; fixes #66) and emits the focus and blur events again 😊 .

The blinking caret is a result of leasing contentEditable for 5 seconds when the terminals that the user is about to paste. Maybe we could move the caret somewhere relatively invisible, but not sure if it is worth the effort at this time.

akalipetis:
LGTM 🐻 , apart from some strange Edge copy/paste issues.

Sometimes I'm able to copy on Edge, but I cannot paste at any time. I don't know if this is something we can take care of, just noticed it.

Code looks 👌 apart from minor comments.

demo/main.js
```diff
@@ -30,3 +32,7 @@ term.on('key', function (key, ev) {
     term.write(key);
   }
 });
+
+term.on('paste', function (data, ev) {
+	term.write(data);
```
akalipetis: I think the indentation is off here.

Tyriar:
See https://github.com/sourcelair/xterm.js/issues/124, no native text control means this makes the operating system's IME not activate. As far as I can tell this cannot be worked around as there are no hooks into the IME from JavaScript and the `KeyboardEvent` received contains the regular character.

PR #82, opened 2016-06-07
Make cursor hollow on blur

### LINKED ISSUE #79: Show hollow cursor when not focused
Right now the solid cursor is shown even when the terminal is not focused, making it more difficult to tell where the application's focus is. Ideally this should be represented by a character something like this:

image

parisk:
Fix #79

### DIFF
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -51,6 +51,12 @@
     color: #000;
 }
 
+.terminal:not(:focus) .terminal-cursor {
+    outline: 1px solid #fff;
+    outline-offset: -1px;
+    background-color: transparent;
+}
+
 /*
  *  Determine default colors for xterm.js
  */
--- a/demo/style.css
+++ b/demo/style.css
@@ -21,6 +21,6 @@ h1 {
     padding: 2px;
 }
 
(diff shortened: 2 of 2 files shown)


src/xterm.css
```diff
@@ -47,8 +47,15 @@
 }
 
 .terminal .terminal-cursor {
+    border: 1px solid #fff;
```
Tyriar: It would look a lot better if the hollow cursor didn't stick out an extra pixel. Something like this would be ideal:

```css
.terminal:not(:focus) .terminal-cursor {
  outline: 1px solid #fafafa;
  outline-offset: -1px;
}
```

It may need a fallback though, not sure how good `outline`'s support is.

TDaglis:
LGTM :panda_face:

src/xterm.css
Tyriar: Pretty good apparently http://caniuse.com/#feat=outline

parisk:
Thanks for the comment @Tyriar (https://github.com/sourcelair/xterm.js/pull/82#discussion-diff-66041689).

Seems like it works great now.

PR #83, opened 2016-06-07
Set line height of terminal rows to normal

### LINKED ISSUE #76: Multi-line selections do not touch each other
At least on Chrome. This is because of:

```css
body {
    line-height: 1.4;
}
```

Before:

image

After:

image

Same on demo:

image

parisk:
Fix #76


Tyriar:
Looks good :+1:

PR #84, opened 2016-06-07
Fix emission of terminal focus&blur events

TDaglis:
Since the removal of the hidden textarea (https://github.com/sourcelair/xterm.js/pull/78) the blur and focus events are not triggered correctly. Specifically, the blur event is not triggered at all.

Moved the focus/blur event emission to onfocus/onblur events instead of focus and blur prototype functions to fix this.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -401,7 +401,6 @@
 
       this.showCursor();
       this.element.focus();
-      this.emit('focus', {terminal: this});
     };
 
     Terminal.prototype.blur = function() {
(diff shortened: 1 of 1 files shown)


parisk:
LGTM 🍰

akalipetis:
I believe a better way of doing this is calling `terminal.focus()` and `terminal.blur()` on the respective DOM events.

This will keep 100% backwards compatibility.

TDaglis:
What this PR does is fix the backwards compatibility. By merging this, the events will be triggered just like they were triggered before.
We can create a separate issue and rework the focus/blur functions if it is necessary.

Issue #27, opened 2015-04-29
Linkify addon breaks the terminal when typing anything that gets linkified


TDaglis:
Before:
![image](https://cloud.githubusercontent.com/assets/3427236/7387177/c5df7886-ee63-11e4-8fb3-a6cc7a676c64.png)

After:
![image](https://cloud.githubusercontent.com/assets/3427236/7387179/cda737e8-ee63-11e4-9d41-eb092d9ee021.png)

After ++:
![image](https://cloud.githubusercontent.com/assets/3427236/7387186/e25dcd0a-ee63-11e4-8998-c0acfd3beee6.png)
The situation gets worse with every keystroke.

This problem occurs only when you type the url. If it appears in another way it is properly likified
This is especially annoying since typing for any reason a file name breaks the terminal (#26)

Tyriar:
Closing in favor of #455

Issue #24, opened 2015-04-28
Link matchers do not work with multi-line links


TDaglis:

martellaj:
Is there anyone looking into this issue? It's a killer for VS Code's integrated terminal.

webdeb:
+1

parisk:
Hi @martellaj, right now this is not assigned to anyone, so no one is looking into this.

We will get on top of this, as soon as we find the required amount of time.

And of course PRs are always welcome 😁.

Tyriar:
I believe all the pre-requisites for this are now done:

- The buffer knows when lines are wrapped (at least on Linux/macOS)
- There is a relatively fast method to grab the entire line as a string
- Links no longer need to juggle the DOM so it will be possible to get multi-line lines with both parts reacting when hovering

I won't be able to pick this up any time soon. Here's what needs to be done for those interested:

- Query [`isWrapped`](https://github.com/sourcelair/xterm.js/blob/313d1e6fecd98a4a5e8dddebb832a6c4c16a828d/src/InputHandler.ts#L68) to identify lines that are wrapped when [linkifying rows](https://github.com/sourcelair/xterm.js/blob/313d1e6fecd98a4a5e8dddebb832a6c4c16a828d/src/Linkifier.ts#L202).
- Modify [`MouseZone`](https://github.com/sourcelair/xterm.js/blob/313d1e6fecd98a4a5e8dddebb832a6c4c16a828d/src/input/MouseZoneManager.ts#L158) and [`MouseZoneManager`](https://github.com/sourcelair/xterm.js/blob/313d1e6fecd98a4a5e8dddebb832a6c4c16a828d/src/input/MouseZoneManager.ts#L20) to support wrapping. This is probably best done by replacing `MouseZone.y` with `y1` and `y2`.

Tyriar:
We also want to make sure links work when the data goes off the screen to the right. This should be tackled at the same time as multi-line links

chirgjn:
I would like to take a stab at this. I have done a fair bit of js for past 2 years and am looking forward to get my first pull request. I can pick this up on the coming weekend if that is fine with everyone 😊

Tyriar:
[...]

PR #90, opened 2016-06-08
Add linkify addon to demo

Tyriar:


parisk:
The linkify addon has some major issues right now (#27, #24) which should be fixed before adding it to the demo.

Will try to deal with linkify, after dealing with CJK characters.

Tyriar:
I wouldn't consider #24 a blocker though, seems like a harmless limitation. It would be a great enhancement to support only single-line URLs imo. #27 and #96 block adoption in vscode from what I can see so far.

parisk:
#27 is the big blocker actually. I hope that I will be able to deal with this, along with https://github.com/sourcelair/xterm.js/issues/96#issuecomment-225559809 within tomorrow.

Tyriar:
Closing this off, it's trivial to do if we want.

PR #93, opened 2016-06-08
Add IP addresses to linkify

### LINKED ISSUE #91: The linkify addon doesn't linkify IP addresses
image

Tyriar:
Fixes #91 

---

image

The fact that the invalid port IP address is still matched is to do with the current `negatedPathCharacterSet`, it will linkify `:http://www.google.com:` too.


akalipetis:
Shall we put the port clause after either host, or IP? This would make more sense IMO.

Thanks for the contribution 🎆

Tyriar:
@akalipetis done

Tyriar:
+tests

akalipetis:
LGTM, thanks a lot!

PR #97, opened 2016-06-09
Don't resize when unnecessary

Tyriar:
This saves some boilerplate for consumers

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -2585,6 +2585,10 @@
         , j
         , ch;
 
+      if (x === this.cols && y === this.rows) {
+        return;
+      }
+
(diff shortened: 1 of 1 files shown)


parisk:
Could you please provide some testing instructions? When does an unnecessary resize happen?

Tyriar:
`term.resize(term.cols, term.rows)` should exit early as it's not needed. You could confirm this with devtools.

A unit test that should be made eventually is `term.refresh` should not be triggered when calling resize with the same cols and rows.

parisk:
Great. Thanks for this.

Now that we set up the environment for testing (#105), this will be more straightforward to implement I guess 😄 .

PR #101, opened 2016-06-09
Fix buffer corruption after resizing rows

### LINKED ISSUE #98: Reducing rows in a resize messes up the terminal
This is likely a recent regression as it does not appear to reproduce on sourcelair.

Here is the server demo:

image

Now run this through devtools: `term.resize(term.cols, term.rows - 1)`

image

The cursor goes to the correct position but the content of the terminal appears to be corrupted permanently. Note that this does not happen when *increasing* the number of rows, only decreasing.

### LINKED ISSUE #99: Increasing rows in a resize should keep the bottom line at bottom
It would be more natural if when increasing the number of rows, the previously final visible row was kept as the last row after the resize (provided the buffer contains sufficient lines above the viewport to fill it in). `gnome-terminal` behaves as I'm describing.

**Steps to reproduce**

1. Load the demo and add some content so the viewport is filled

  image
2. Run: `term.resize(term.cols, term.rows + 1)`
3. Notice the line at the bottom is empty

  image

Tyriar:
Lines from the buffer were incorrectly being removed when the viewport was
resized. This change removes the lines when the cursor is above them, if not it
shifts the viewport down. Some basic jsdoc comments were added to some Terminal
properties for future reference.

Fixes #98
Fixes #99

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -202,10 +202,26 @@
         this.on('data', options.handler);
       }
 
+      /**
+       * The scroll position of the y cursor, ie. ybase + y = the y position within the entire
+       */
       this.ybase = 0;
+
+      /**
+       * The scroll position of the viewport
+       */
       this.ydisp = 0;
+
+      /**
+       * The cursor's x position after ybase
+       */
       this.x = 0;
+
+      /**
+       * The cursor's y position after ybase
+       */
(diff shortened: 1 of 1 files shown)


parisk:
Thanks a lot for this! The attached patch seems to be working great.

PR #107, opened 2016-06-09
Allow Ctrl/Shift + insert to copy/paste on windows and linux

runarberg:
Many systems (including MS Windows and many linuxes) map <kbd>Ctrl</kbd> + <kbd>Insert</kbd> to copy and <kbd>Shift</kbd> + <kbd>Insert</kbd> to paste. That serves as a handy fallback when the more common <kbd>Ctrl</kbd> + <kbd>C</kbd> and <kbd>Ctrl</kbd> + <kbd>V</kbd> keybindings have their default prevented to send signals to the terminal.

Currently all keydown-events with the insert key send `\x1b[2~` to the terminal. Instead this PR won’t send anything if either the `shiftKey` or the `ctrlKey` are present in the event. Also it will enable `contentEditable` when `shiftKey` is present to allow for pasting in Firefox and Edge.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -519,6 +519,11 @@
             term.leaseContentEditable();
           }
         }
+
+        if (!term.isMac && ev.keyCode == 45 && ev.shiftKey && !ev.ctrlKey) {
+          // Shift + Insert pastes on windows and many linuxes
+          term.leaseContentEditable();
+        }
       });
 
       /**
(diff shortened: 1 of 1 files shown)


Tyriar:
Ideally this should be configurable, I'd like VS Code users to be able to use their preferred method (the default of which will probably be <kbd>ctrl</kbd>|<kbd>cmd</kbd>+<kbd>shift</kbd>+<kbd>c</kbd>/<kbd>v</kbd>). `gnome-terminal` has a settings dialog which allows customizing of a bunch of keybindings that aren't native

image

Perhaps allowing consumers to do more generic mappings would be a good idea?

Also <kbd>shift</kbd>+<kbd>insert</kbd> at least on `gnome-terminal` appears to be a shortcut for middle-click to paste selection, not from clipboard.

parisk:
While I agree with @Tyriar that some kind of customization/extensibility considering key bindings, I do not believe that this is blocking at the moment.

This PR works 👍 . LGTM.

PR #112, opened 2016-06-10
Improve scroll to work with blank lines after the cursor

### LINKED ISSUE #111: Resizing down and then up messed up cursor position
Looks like I missed a case in #101

Steps to reproduce:

1. Load demo
2. Fill with output
3. Run:

  ```js
  term.resize(term.cols, term.rows + 2)
  term.resize(term.cols, term.rows - 2)
  ```

image

Tyriar:
This commit works fixes scrolling when there were blank lines after the cursor.
Here is what it does (blank rows are those added by running `clear`):

```
when increasing rows:
  if there are blank rows below the cursor:
    add a blank row to the bottom
  else if there is room in the buffer above the viewport
    scroll up
  else
    add a blank row to the bottom

when decreasing rows:
  if there are blank rows below the cursor:
    remove a blank row from the bottom
  else
    scroll down
```

Fixes #111

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -2607,7 +2607,8 @@
         , el
         , i
         , j
-        , ch;
+        , ch
+        , addToY;
 
       if (x === this.cols && y === this.rows) {
         return;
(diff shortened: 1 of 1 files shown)


parisk:
Works great, thanks!

joerick:
Good stuff @Tyriar! Really pleased the VSCode guys are hopping on this project - a rising tide lifts all boats!

PR #117, opened 2016-06-10
Bound refresh max rows to this.rows not this.lines

### LINKED ISSUE #86: Potential cannot set property of undefined exception in refresh
I've managed to reproduce an issue in jeremyramin/term.js where calling refresh while explicitly indicating an end row can cause a undefined exception, see https://github.com/Microsoft/vscode/issues/7357

Looking at the code for xterm.js it looks like this particular case will still occur so here's what I'm doing:

Calling `this.terminal.refresh(0, this.terminal.rows);` causes an exception within the `refresh` function on the line `this.children[y].innerHTML = out;`. In other words, this.terminal.rows is not sufficiently validated or there are missing children that are expected.

Tyriar:
this.row is the size of rows in the viewport, this.lines is the buffer. It's
only possible to refresh `0` to `this.rows - 1`.

Fixes #86 

---

This prevents an exception when calling refresh incorrectly.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -1135,9 +1135,9 @@
       width = this.cols;
       y = start;
 
-      if (end >= this.lines.length) {
+      if (end >= this.rows.length) {
         this.log('`end` is too large. Most likely a bad CSR.');
-        end = this.lines.length - 1;
+        end = this.rows.length - 1;
       }
 
       for (; y <= end; y++) {


parisk:
Thanks a lot for this!

parisk:
@Tyriar I think that there might be a chance that this PR made the focus/blur events go a little bit crazy. Could you please try setting a listener to each event and compare the event emission rate before and after the attached commit?

PR #121, opened 2016-06-11
Fix cursor blinking when enabled

### LINKED ISSUE #119: Cursor blinking seems to be broken
Add option to demo:

```js
    term = new Terminal({
      cursorBlink: true
    })
```

Doesn't work

Tyriar:
Fixes #119


parisk:
What about implementing this with CSS animations? Seems that browser support is pretty good: http://caniuse.com/#feat=css-animation.

Tyriar:
Could be a nice improvement. Would be good to get this in to fix it though.

parisk:
Are there any particular issues that prevent us from going with CSS animations within the scope of this PR?

Tyriar:
Time was the only reason, I went ahead and did it though :stuck_out_tongue: 

It works fine when https://github.com/sourcelair/xterm.js/pull/133 is applied. It's actually quite convenient as refresh recreates the cursor which resets the cursor, so no need to manually reset it on keydown.

parisk:
Thanks a lot @Tyriar

PR #122, opened 2016-06-11
Add basic options selection to demo

Tyriar:
This will make it easier to test options without needing to modify the demo.


parisk:
Great. Hope that this will help out.

PR #123, opened 2016-06-11
Add some jsdoc for important functions

Tyriar:
Part of #13


parisk:
That's great! Thanks a lot.

PR #125, opened 2016-06-13
Implement addon loader (CommonJS + RequireJS)

### LINKED ISSUE #96: Cannot import addons in Electron
`require('xterm')` points at `main` in package.json (`src/xterm.js`). As far as I'm aware, there is no way to grab the addon JS modules from a node API. Something like this would probably work:

&lt;reporoot&gt;/index.js

```js
module.exports = {
  Terminal: require('./src/xterm'),
  addons: {
    fit: require('./addons/fit'),
    ...
  }
}
```

parisk:
Closes #96

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -759,6 +759,26 @@
       this.emit('open');
     };
 
+
+    /**
+     * Attempts to load an add-on using CommonJS or RequireJS (whichever is available).
+     * @param {string} addon The name of the addon to load
+     * @static
+     */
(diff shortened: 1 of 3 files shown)


Tyriar:
This may not work for some build systems for the web, specifically those that derive which modules to pull in by following `require`s and then concat sources.

parisk:
Since RequireJS is not messing with the node path and can work seamlessly with relative paths, requirejs users can stick with the native way of loading these modules.

PR #128, opened 2016-06-13
Allow refresh to execute 30 times a second maximum

### LINKED ISSUE #126: Long running processes cannot be stopped with SIGINT
Steps to reproduce:

1. `npm start`, load in browser
2. Run `ls -lR /usr/share` on Linux, `ls -lR /Applications` on Mac (or some other big dir)
3. Try interrupt using <kbd>ctrl</kbd>+<kbd>c</kbd>

### LINKED ISSUE #127: A busy stream maxes out CPU due to constant refreshes
Steps to reproduce:

1. `npm start`, load in browser
2. Run `ls -lR /usr/share` on Linux, `ls -lR /Applications` on Mac (or some other big dir)
3. View process CPU%

I believe this is happening because `refresh` is called at a faster rate than it takes to run.

See https://github.com/Microsoft/vscode/issues/7484#issuecomment-225662697

Tyriar:
For commands that pass a significant amount of output to the write function,
this prevents the terminal maxing out the CPU and making the UI unresponsive.
While commands can still run beyond what they do on the terminal, it is far
better with a debounce in place as every single terminal manipulation does not
need to be constructed in the DOM.

A side-effect of this is that it makes ^C to interrupt a process seem more
responsive.

Fixes #127
Fixes #126

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -230,6 +230,16 @@
        */
       this.y = 0;
 
+      /**
+       * Used to debounce the refresh function
+       */
+      this.isRefreshing = false;
+
+      /**
+       * Whether there is a full terminal refresh queued
+       */
+      this.queuedRefresh = false;
+
       this.cursorState = 0;
       this.cursorHidden = false;
       this.convertEol;
@@ -297,6 +307,7 @@
 
       this.tabs;
       this.setupStops();
+      this.debounceRefresh();
     }
 
     inherits(Terminal, EventEmitter);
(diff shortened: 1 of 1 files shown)


parisk:
LGTM. Thanks a lot!

parisk:
@Tyriar I think that there might be a chance that this PR made the focus/blur events go a little bit crazy. Could you please try setting a listener to each event and compare the event emission rate before and after the attached commit?

It seems that `debounceRefresh` runs perpetually and not only when requested. Ref:  https://github.com/sourcelair/xterm.js/pull/128/files#diff-d4f745d6f29faa0617f2e02bb95f0c44R310

PR #131, opened 2016-06-14
Fix focus and blur events emission

parisk:


Tyriar:
It seems to perform around the same, my main test for this was that selection of the `h1` was still smooth while `ls -lR /usr/share` ran. It's still pretty slow in vscode, refresh seems to still 100% CPU, this happened before and after this change though.

Before:

image

After:

image

PR #136, opened 2016-06-17
Demo improvements

dbkaplun:
This PR makes the demo more functional and app-like. The goal is to make the demo an unobstrusive portal into xterm.js as it should be used.


parisk:
Hi @dbkaplun, thanks for your contribution!

Could you please elaborate a little bit on what would you like to achieve with this PR? What is the functionality you would like to add to xterm.js' demo?

Also, it seems that the `login` command with which you replaced `bash` does not work without root

Last, I do not feel that well with renaming "demo" to "app". The role of the demo is to literally demonstrate the capabilities of this library, which is not a standalone app.

dbkaplun:
My goal in this PR was to lay the groundwork to simplify the demo page into a barebones wrapper around xterm.js. Hopefully other developers will see the page and will have an easier time implementing based on example. I can remove the renaming from the PR but I do think it should be called 'app' rather than 'demo' since it's what's launched by `npm start`.

re `login` command: I will update the PR to use `$SHELL` instead. (It worked without root on OSX.)

I have some other branches I've been working on around the demo page, relating to resize handling and other stuff which I will file when I have the time. Please let me know how I can improve this PR!

app/main.js
```diff
@@ -4,23 +4,22 @@ var term,
...
+}
+createTerminal(opts);
 
-function createTerminal() {
+function createTerminal(opts) {
   while (terminalContainer.children.length) {
     terminalContainer.removeChild(terminalContainer.children[0]);
   }
-  term = new Terminal({
-    cursorBlink: optionElements.cursorBlink.checked
```
Tyriar: How do you turn `cursorBlink` on in this PR?

dbkaplun: You can turn it on by opening your browser's JS console on the demo page and typing `localStorage.opts = JSON.stringify({cursorBlink: true})`. The goal here is to make it possible to change all config params without changing the code, not just `cursorBlink`.

parisk:
Hi @dbkaplun thanks a lot for replying! I am still missing the main point of this PR though.

Could you please [open an issue](https://github.com/sourcelair/xterm.js/issues/new) in this repository, where you will pin-point the particular things that "annoy" you with the current demo, and propose your solutions there?

This way we will be able to discuss it and end up in the solution that we believe serves best the purpose of this project.

app/main.js
Tyriar: This is just the beginning for the options in the demo though, my plan was to eventually have all of the properly supported options there so that they can easily be demoed. Forcing the user to run that in the console is far from intuitive and would hide the feature from potential users and annoy developers who want to test the feature.

dbkaplun:
I have decided to move these changes into their own repository so it can be developed more rapidly. Closing this PR.

PR #137, opened 2016-06-19
Fix cumbersome pasting with Cmd + V

parisk:
For some reason, setting the element's contentEditable value to true within a keyboard event, while the element has focus, did not allow pasting with clipboard, unless the element gets clicked explicitly.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -456,7 +456,7 @@
      * @public
      */
     Terminal.prototype.blur = function() {
-      return terminal.element.blur();
+      return this.element.blur();
     };
 
     /**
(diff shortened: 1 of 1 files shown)


Tyriar:
This might be a browser security feature, ensuring that a user action is performed before accessing the clipboard.

parisk:
I do not think so. The thing is that when you press <kbd>Cmd</kbd> + <kbd>V</kbd> on OS X Chrome (which is a user action), it does not fire the `paste` event. This happens only when the textarea's `contenteditable` value is being set to `true` from `false` and no click has happened before the keyboard event.

PR #138, opened 2016-06-23
Disable ligatures which mess up rendering for certain fonts

dbkaplun:
A ligature is when two characters render as one glyph. This makes text look better but is problematic for a fixed-width environment such as a terminal. For instance [this monospace font](https://github.com/powerline/fonts/tree/master/Meslo) uses turns "fi" into a ligature which should be disabled for proper rendering.

https://en.wikipedia.org/wiki/Typographic_ligature
https://developer.mozilla.org/en-US/docs/Web/CSS/font-feature-settings

### DIFF
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -39,6 +39,7 @@
     background-color: #000;
     color: #fff;
     font-family: courier-new, courier, monospace;
+    font-feature-settings: "liga" 0;
     position: relative;
 }
 


parisk:
Seems to be working great. Thanks!

parisk:
@dbkaplun could you please take a look at @Tyriar's comment: https://github.com/sourcelair/xterm.js/commit/4c6b99345875171c476a8e4d2f880e7372094060#commitcomment-17998829?

Would such a change provide a better result?

Tyriar:
Chrome doesn't seem to support `font-variant-ligatures: none;`

PR #142, opened 2016-06-24
Only blink cursor if the terminal has focus

### LINKED ISSUE #141: Cursor should only blink when terminal has focus

Tyriar:
Fixes #141


parisk:
Great, thanks!

Issue #70, opened 2016-06-02
Moving over CJK text jumps 2 characters at a time


Tyriar:
Related: https://github.com/sourcelair/xterm.js/issues/61

Initial:

![image](https://cloud.githubusercontent.com/assets/2193314/15739200/d8f27dca-2863-11e6-9620-66b58c7f7ede.png)

<kbd>&larr;</kbd>

![image](https://cloud.githubusercontent.com/assets/2193314/15739222/e42be014-2863-11e6-999c-715acf2df966.png)

Text: 안녕하십니까?

parisk:
Thanks for reporting. Will take a look at it, along with the rest of international character issues.

parisk:
Closed in #144.

Issue #62, opened 2016-06-02
CJK characters are not treated as double width chars


Tyriar:
See https://github.com/chjj/term.js/issues/96

![image](https://cloud.githubusercontent.com/assets/2193314/15738068/539d8c82-285e-11e6-91ce-b839db904383.png)

vs gnome-terminal:

![image](https://cloud.githubusercontent.com/assets/2193314/15738083/67542448-285e-11e6-83be-6db1b670e607.png)

Notice the characters are exactly 2 ascii characters in width

parisk:
Thanks for reporting. Will take a look at this along with the rest international character issues.

parisk:
Seems like the weird line break is actually the div overflowing. This indeed happens because these are double byte characters, the terminal ignores this fact and counts them as single-byte ones.

Working on a fix.

## Screenshot
![image](https://cloud.githubusercontent.com/assets/1188592/15852933/afdcc7a0-2cab-11e6-961a-0ae79cfe399e.png)

jerch:
@parisk To support this you gonna have to include the wcwidth calculation for any Unicode codepoint (see `man wcwidth`) and adjust the taken space for fullwidth characters to 2 terminal cells (terminals are based on the idea that a single cell can hold one halfwidth character since all western characters are halfwidth).
Problem with Javascript and Unicode is - it is a total mess with the UTF-16 encoding (surrogates) and very expensive to calculate. This gets even more complicated if you plan to support stackable combining characters (wcwidth will report a width of zero for those). This might be easier to implement once the Unicode functions with real codepoints are widely adapted by the JS engines.
I tried to implement this according to the Unicode spec but well - it ended up as a total code mess in my own terminal emulator.

jerch:
I had a look at your line at cell abstraction - this is pretty much the same I did in my emulator. If you like I can extend your isWide and write functions with the wcwidth stuff in a PR.
Another problem is the rendering with Unicode. For higher BMP or non BMP codepoints the fonts might differ in the glyph width from wcwidth (if they have that glyph at all). You are likely to need a CSS class fixing misaligned glyphs into place to keep the output a well formed cell grid.

parisk:
Interesting. Do you believe that this could be handy? http://code.woong.org/wcwidth.js

jerch:
[...]

Issue #72, opened 2016-06-02
Strange characters show up when moving a cursor over hindi


Tyriar:
Text: अर्धतत्सम

![image](https://cloud.githubusercontent.com/assets/2193314/15739616/bd055860-2865-11e6-9a15-98df3a82db58.png)

![image](https://cloud.githubusercontent.com/assets/2193314/15739630/ce01b5b4-2865-11e6-8d53-b98a21245d6c.png)

Characters change:

![image](https://cloud.githubusercontent.com/assets/2193314/15739673/f71f079e-2865-11e6-9371-d3cdab539fee.png)

Can't move further back that this:

![image](https://cloud.githubusercontent.com/assets/2193314/15739653/e2cda5de-2865-11e6-9890-6ea87f594706.png)

jerch:
This is directly related to the missing handling of combining characters mentioned in #62

jerch:
This is what I get under Linux and FF with default monospace font (Liberation Sans Mono) at 30px:

![1](https://cloud.githubusercontent.com/assets/6193135/16388275/b1c9daec-3c98-11e6-88fc-9ae0298e1b78.png)
![2](https://cloud.githubusercontent.com/assets/6193135/16388276/b1e6dade-3c98-11e6-8ead-97b0e4a8083e.png)
![3](https://cloud.githubusercontent.com/assets/6193135/16388281/b22482bc-3c98-11e6-9f09-7aadefc85a0a.png)
![4](https://cloud.githubusercontent.com/assets/6193135/16388277/b1ff812e-3c98-11e6-90f6-46b353c6ed37.png)
![5](https://cloud.githubusercontent.com/assets/6193135/16388280/b21cba8c-3c98-11e6-840d-b4af76dba339.png)
![6](https://cloud.githubusercontent.com/assets/6193135/16388278/b215680e-3c98-11e6-81f1-850d6f592e76.png)
![7](https://cloud.githubusercontent.com/assets/6193135/16388279/b219022a-3c98-11e6-91eb-369c9cdf3880.png)

[...]

PR #144, opened 2016-06-24
wcwidth calculation

jerch:
Quick replacement of `isWide()` with `wcwidth`.  Since xterm.js has no frontend packaging atm I just copied my version into the source file.

wcwidth fixes several issues:
- full unicode (all planes supported at data stage, output depends on font capability)
- combining characters
- half/fullwidth distinction for all unicode codepoints (though lookup table needs fixes to get xterm or font compatibility)

Misaligned glyphs are not fixed by this pull request. This would need a way to introspect the current font and glyph widths. This could be done by prerendering faulty glyphs onto a canvas, getting their width with  `measureText` and aligning them with a CSS rule. I played with this approach here https://github.com/netzkolchose/jquery.browserterminal/blob/master/dist/js/jquery.browserterminal.js#L289, the glyph width calculation is pretty slow but is only needed once.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -70,6 +70,122 @@
     var window = this, document = this.document;
 
     /**
+     * wcwidth
+     */
+    var wcswidth = (function() {
+        var _WIDTH_COMBINING = [
+            [0x0300, 0x036F], [0x0483, 0x0486], [0x0488, 0x0489],
+            [0x0591, 0x05BD], [0x05BF, 0x05BF], [0x05C1, 0x05C2],
+            [0x05C4, 0x05C5], [0x05C7, 0x05C7], [0x0600, 0x0603],
+            [0x0610, 0x0615], [0x064B, 0x065E], [0x0670, 0x0670],
+            [0x06D6, 0x06E4], [0x06E7, 0x06E8], [0x06EA, 0x06ED],
+            [0x070F, 0x070F], [0x0711, 0x0711], [0x0730, 0x074A],
+            [0x07A6, 0x07B0], [0x07EB, 0x07F3], [0x0901, 0x0902],
+            [0x093C, 0x093C], [0x0941, 0x0948], [0x094D, 0x094D],
+            [0x0951, 0x0954], [0x0962, 0x0963], [0x0981, 0x0981],
+            [0x09BC, 0x09BC], [0x09C1, 0x09C4], [0x09CD, 0x09CD],
+            [0x09E2, 0x09E3], [0x0A01, 0x0A02], [0x0A3C, 0x0A3C],
+            [0x0A41, 0x0A42], [0x0A47, 0x0A48], [0x0A4B, 0x0A4D],
+            [0x0A70, 0x0A71], [0x0A81, 0x0A82], [0x0ABC, 0x0ABC],
+            [0x0AC1, 0x0AC5], [0x0AC7, 0x0AC8], [0x0ACD, 0x0ACD],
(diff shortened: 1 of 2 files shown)


parisk:
Hi @jerch, thanks for your contribution! Before moving on with testing this out, could you please mention the particular GitHub issues that this PR is supposed to fix?

jerch:
Issues affected by this: #72 , maybe #70 , #62

src/xterm.js
```diff
@@ -70,6 +70,122 @@
     var window = this, document = this.document;
 
     /**
+     * wcwidth
+     */
+    var wcswidth = (function() {
+        var _WIDTH_COMBINING = [
+            [0x0300, 0x036F], [0x0483, 0x0486], [0x0488, 0x0489],
```
Tyriar: Where was this list sourced from?

src/xterm.js
```diff
@@ -1468,8 +1589,30 @@
         this.maxRange();
       }
 
+      // apply leftover surrogate left from last write
+      if (this.surrogate_left) {
+            data = this.surrogate_left + data;
```
Tyriar: Use 2 space indents

src/xterm.js
```diff
@@ -1468,8 +1589,30 @@
...
+        }
+
       for (; i < l; i++) {
         ch = data[i];
+
+        code = data.charCodeAt(i);
+        // surrogate high
+        if (0xD800 <= code && code <= 0xDBFF) {
+            low = data.charCodeAt(i+1);
+            if (low !== low) {  // end of data stream, we need to save to first part of surrogate
```
Tyriar: `low` will always equal `low`?

src/xterm.js
```diff
@@ -70,6 +70,122 @@
     var window = this, document = this.document;
 
     /**
+     * wcwidth
+     */
+    var wcswidth = (function() {
```
Tyriar: General helper functions should go at the end of the file.

src/xterm.js
jerch: https://www.cl.cam.ac.uk/%7Emgk25/ucs/wcwidth.c

src/xterm.js
jerch: `charCodeAt` with an illegal index returns NaN, (NaN !== NaN) evaluates always to true.

Tyriar: Can you use `isNaN()` for clarity?

jerch:
I have cleaned up the code and fixed the indentation.

Tyriar:
Thanks @jerch! I'll have to take a chunk of time to look at this in detail, can you think of some test cases that could be added for `wcwidth`?

jerch:
yes I have some test cases, gonna try to apply them to xterm.js.

There are some edge cases that need proper testing:
- fullwidth chars just go to next line if entered at `cols-2`, what to do with the last cell itself? It will need at least the attributes set.
- combining chars should go to last active cell - needs tests for all cells, no clue yet how solve this after cursor jumps (xterm just insert them at the cursor after jumps)

jerch:
One test case fails - I commented it out since the it refers to a missing feature in default state (wrap true/false is not handled there at all).

parisk:
Okay, just found the chance to take a deeper look into this. #72 does not seem to be closed by this (still weird characters appear, but different than the ones mentioned in this issue).

This PR closes #70 and closes #62 successfully though.

I will move on with merging this PR, since it great steps forward have been done here. I also believe that #72 can be tackled autonomously, since it appears only when moving the cursor.

Thanks a lot!

Issue #134, opened 2016-06-16
Performance of refresh is still subpar


Tyriar:
Follow up from https://github.com/sourcelair/xterm.js/pull/131#issuecomment-226566256

It looks like refresh and the layout is still taking far too long in a real application scenario. Consider the following portion of a Chromium timeline when running `ls -lR /usr/share` in VS Code, which takes around 3.5s for the process to run but ~25s until it's finished in the UI.

![image](https://cloud.githubusercontent.com/assets/2193314/16127983/a93b7856-33b3-11e6-8995-7b4e0bf7a16d.png)

My initial thoughts on this were that it was taking too long to parse the incoming data in `Terminal.prototype.write` but this timeline seems to refute that. All time is spent in refresh and functions/events it triggers.

Here is the timeline for a single refresh:

![image](https://cloud.githubusercontent.com/assets/2193314/16128017/cdc9db18-33b3-11e6-93be-4261ba40dbb7.png)

And zoomed in even more to the portion of refresh that gets repeated:

![image](https://cloud.githubusercontent.com/assets/2193314/16128068/02d4bddc-33b4-11e6-96c2-1eb64d0a33d7.png)

The basic flow of the section that repeats is:
1. Blur/focusout events
2. Recalc style
3. Lots of separate parse HTML sections
4. Recalc style
5. Layout
6. Focus/focusin events
7. Recalc style
8. Layout

Looking at this, refresh's performance could still certainly be improved. In the meantime another thing that would be good to have to protect against this is to skip refreshes if a refresh is already in progress.

Tyriar:
@parisk I can have a go at fixing this considering I have the most context.

parisk:
@Tyriar I think that we are already skipping such refreshes in https://github.com/sourcelair/xterm.js/commit/a6e85ad5f5aa224cd5770a57806f1219779719d4#diff-d4f745d6f29faa0617f2e02bb95f0c44R1206. Do you mean something else?

Feel free to give it a try if you would like. What is the approach you are thinking to take?

Tyriar:
@parisk using a DocumentFragment to prevent layout thrashing. I believe right now if a refresh takes longer than 34ms, it may not wait before starting a new one, ie. `this._refreshIsQueued = false;` should be at the end of refresh.

jerch:
I had the same problem with https://github.com/netzkolchose/jquery.browserterminal and played alot with performance tips floating around. What helped to increase the output performance and interactivity was:

[...]

PR #146, opened 2016-06-25
Keep slave program in sync with terminal

jerch:
Since I got no anwser to https://github.com/sourcelair/xterm.js/issues/134#issuecomment-227474863 here is a PR addressing the out of sync problem of the slave program and the terminal emulator.

The problem boils down to the nonblocking websocket handling, which cant propagate back the "uhmm, I'm still busy, please wait" of the terminal emulator and buffers more and more bytes. In the result the emulator lacks far behind in data processing. Thus a long running output heavy task seems not to be interruptable (which is not the case, it is just the terminal emulator, which has to get through the big amount of data to the point where the interrupt happened).
This PR uses XON/XOFF flow control to keep the backend and frontend in sync.

You can test this with any long running output intensive task, e.g. `ls -lR /`

Related issues are #134 and https://github.com/Microsoft/vscode/issues/7801

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -32,6 +32,9 @@ app.ws('/bash', function(ws, req) {
   });
   term.on('data', function(data) {
     try {
+      // XOFF - stop pty pipe
+      // XON will be triggered by emulator before processing data chunk
+      term.write('\x13');
       ws.send(data);
     } catch (ex) {
       // The WebSocket is not open, ignore
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -294,6 +294,9 @@
       this.prefix = '';
       this.postfix = '';
 
+      // user input states
+      this.user_xoff = false;  // user pressed XOFF
+
       /**
        * An array of all lines in the entire buffer, including the prompt. The lines are array of
        * characters which are 2-length arrays where [0] is an attribute and [1] is the character.
@@ -1458,6 +1461,11 @@
      * @public
      */
     Terminal.prototype.write = function(data) {
+      // XON - about to process data, thus we can get more
+      // dont send XON if user pressed XOFF
+      if (!this.user_xoff)
+        this.send('\x11');
+
       var l = data.length, i = 0, j, cs, ch;
 
       this.refreshStart = this.y;
@@ -2655,6 +2663,10 @@
           }
           break;
       }
+      if (result.key === '\x13')
+        this.user_xoff = true;
+      else if (result.key === '\x11')
+        this.user_xoff = false;
       return result;
     };
 
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -1462,7 +1462,7 @@
      */
     Terminal.prototype.write = function(data) {
       // XON - about to process data, thus we can get more
-      // dont send XON if user pressed XOFF
+      // dont lift XOFF if user pressed it
       if (!this.user_xoff)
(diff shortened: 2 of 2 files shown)


Tyriar:
While this does seem to work, and is quite snappy when sending `SIGINT`, I don't think we necessarily want to ensure that absolutely everything has been synced to the UI. While the frontend can be optimized further (as detailed in #134), this change negates the work that was done in https://github.com/sourcelair/xterm.js/pull/128 which allows as much data as possible through, we just only print the last page of it.

Consider this command with this change:

```
$ time ls -lR ~
...
real    3m21.946s                                                                                 
user    0m2.796s                                                                                  
sys     0m4.752s
```

And without the change:

```
$ time ls -lR ~
...
real    0m3.289s                                                                                  
user    0m1.124s                                                                                  
sys     0m2.056s
```

`real` is not actually accurate on the second output which is what is so awesome about this change, but the command only takes about 10 seconds to run.

Perhaps this could be hooked up somehow to `refresh` instead of `write` so that we can get the benefits of both the `refresh` throttling and `XON`/`XOFF`?

jerch:
Ah yes your 30fps refresh patch will have no effect with all those xon/xoffs, since it will stop the backend with every websocket transfer. Although the stops bind any keyboard action to the next websocket action (therfore feeling snappy and in sync) it is not suitable - the average websocket transfer buffer is about 1k to 3k even for output heavy tasks, that means the backend gets stopped every few thousand bytes. Bummer.

I created a new branch with a different approach: https://github.com/jerch/xterm.js/tree/flowcontrol_buffer
Here I use a high/low watermark buffer to trigger the flow control. The high mark is used to halt the backend (out of sync gap to big) and the low mark is used to avoid waiting time on the frontend for new data.
All settings are adjustable, the commit contains the settings where even my FF can handle it smoothly (which is not the case with just your fps patch). SIGINT gets through at the average max-min mark (the buffer state shows a sawtooth behavior).

Regarding the xon/xoff at refresh time:
Imho there is much room for rendering optimizations which should be applied first. (The switch to `pre` with `span` nearly doubled the throughput, same with fragments instead of HTML strings).

parisk:
Sorry about the delay to get back on this, but finally I got my input. I am not quite fond of the idea of synchronizing the back-end with the front-end. IMO the best option is optimizing rendering as much as possible and maybe offering this feature as an add-on or a flag (`sync`?) to the [attach add-on](https://github.com/sourcelair/xterm.js/blob/master/addons/attach/attach.js).

Also it would be great if you could provide us with a PR which swaps `div`s with `pre`s, in order to try it out and see how does it perform.

jerch:
> Also it would be great if you could provide us with a PR which swaps divs with pres, in order to try it out and see how does it perform.

This would need an almost full rewrite of the `refresh` method. Gonna try it if I find some time for that.

parisk:
Another thing we could do would be opening a GitHub issue, proposing a refactoring of the `refresh` method based on this principles, in order to create a place for discussion and make it available for anyone to take it over.

Do you think you could do that (creating a GitHub issue)?

jerch:
Sure - feel free to add more ideas and discuss it here #150

parisk:
Awesome, thanks!

jerch:
Gonna close the PR since it is obsolete atm...

PR #148, opened 2016-06-27
Fix MouseEvent cloning

bfis:
Cloneing MouseEvent instances via __proto__ (line 1099) caused the cloned object to be broken and throw "TypeError: Illegal invocation" during property access ("ev.shiftKey" line 1020).

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -838,7 +838,7 @@
 
         sendEvent(button, pos);
 
-        switch (ev.type) {
+        switch (ev.overrideType || ev.type) {
           case 'mousedown':
             pressed = button;
             break;
@@ -988,7 +988,7 @@
         // 3 = release
         // wheel up/down:
         // 1, and 2 - with 64 added
-        switch (ev.type) {
+        switch (ev.overrideType || ev.type) {
           case 'mousedown':
             button = ev.button != null
               ? +ev.button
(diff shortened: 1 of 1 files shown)


parisk:
Hi @bfis! Thanks a lot for your contribution. Before merging this I would like to ask for a couple of things:

1. instructions on how to reproduce the initial error, in order to check out what this PR fixes
2. could you add any tests that cover this case?

bfis:
# How to reproduce:

1. Use the terminal (in Browser) with some program that sets the vt100mouse mode (in my case htop does it)
2. Have your developers console open
3. Click somewhere

Now, you should get an uncaught exception, i.e:

- "Uncaught TypeError: Illegal invocation" in Chromium 51.0.2704.79
- "TypeError: 'shiftKey' getter called on an object that does not implement interface MouseEvent." in Firefox 45.0

Since this is a problem specific to the special handling of MouseEvents in Browsers, I have no clue how to implement a test (nor if one could do this with reasonable effort).

bfis:
This "browser feature" can be easily checked with:

```javascript
e = new MouseEvent("mousedown") // => MouseEvent {isTrusted: false, screenX: 0, screenY: 0, clientX: 0, clientY: 0…}
e.screenX // => 0
c = {__proto__: e, foo: 123} // => MouseEvent {foo: 123}
c.foo // => 123
c.screenX // throws: Uncaught TypeError: Illegal invocation(…)(anonymous function)
```

parisk:
👍  thanks a lot!

Issue #118, opened 2016-06-10
Support custom keybindings


Tyriar:
Raised from a discussion in https://github.com/sourcelair/xterm.js/pull/107, I would like to eventually let vscode users to be able to configure keybindings for:

- Copy and paste (https://github.com/Microsoft/vscode/issues/6451, this one's particularly urgent)
- Clear terminal (⌘K on OS X https://github.com/Microsoft/vscode/issues/7240, general command request https://github.com/Microsoft/vscode/issues/6603)

What I'm thinking is allowing consumers of xterm.js to be able to plug key -> escape sequence mappings into `evaluateKeyEscapeSequence`. This could return the same result interface if there is a match otherwise `undefined`.

parisk:
I was thinking that an interesting solution would be offloading the copy/paste functionality to an addon, so as to be able to apply any functionality programmatically on top of the core terminal.

Tyriar:
@parisk exposing some API like `Terminal.prototype.registerKeyMappings(event)` would be good, then that can be used by both addons for common customizations like copy and paste and also lets consumers do whatever they want with it.

PR #153, opened 2016-06-30
Fix Ctrl+Shift+V pasting

TDaglis:
Lease contentEditable for non Mac devices on Ctrl+Shift+V.
Even though it was mentioned in the listener documentation it had not been implemented and there were cases where pasting in Windows (and other non mac devices) would not work

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -551,8 +551,8 @@
 
       /**
        * Hack pasting with keyboard, in order to make it work without contentEditable.
-       * When a user types Ctrl + Shift + V or Cmd + V on a Mac, lease the contentEditable value
-       * as true.
+       * When a user types Ctrl + Shift + V or Shift + Insert on a non Mac or Cmd + V on a Mac,
+       * lease the contentEditable value as true.
        */
       on(term.element, 'keydown', function (ev) {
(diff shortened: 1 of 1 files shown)


Tyriar:
Is <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>c</kbd> for copy going to be hardcoded along the same lines until #118 is tackled?

TDaglis:
I think copy does not require special handling to work since it does not need the contentEditable hack. Are there any issues with copying?

TDaglis:
My bad. Copy requires special care as well indeed.

akalipetis:
@Tyriar this is not really hard-coded, since it's still using the browser's behavior for pasting - just adds the `contenteditable` attribute.

In the case of copying and, we'll need to use something like `document.execCommand('Copy');`. Let's keep the discussion about copying any other keybindings at #118.

parisk:
LGTM. Thank!

serjN:
On version 1.1.3 of xterm.js this not works.

parisk:
@serjN could you please open up an issue for this in order to investigate it and see if and how we can fix it?

Tyriar:
Works fine for me on Ubuntu.

PR #160, opened 2016-07-04
Have ' ' chars get inserted correctly after a resize

### LINKED ISSUE #159: &nbsp; characters are not reinstated when horizontally shrinking and increasing viewport

Tyriar:
Since the character spec didn't contain a character width it was being skipped
during refresh.

Fixes #159

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -2828,7 +2828,7 @@
       // resize cols
       j = this.cols;
       if (j < x) {
-        ch = [this.defAttr, ' ']; // does xterm use the default attr?
+        ch = [this.defAttr, ' ', 1]; // does xterm use the default attr?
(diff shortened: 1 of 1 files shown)


parisk:
@Tyriar can you provide us with some testing instructions on this? How can we verify that this PR works?

jerch:
@parisk This was introduced by my wcwidth patch. I tried to fix all new cell creations with the character width as third array element. This one slipped through ;)
You can test it with a `resize` or `resize+refresh` test case - the sum of the third cell element of a line should always be equal to `term.cols` after resizing. Same for the length of a not styled line string containing only halfwidth characters in `refresh`.

parisk:
👍  Thanks a lot!

Issue #156, opened 2016-07-03
Proposal: use more tests


jerch:
The emulator is a fairly complex beast already and has like no test cases. Do you have any plans for testing xterm.js?
I feel a bit uneasy about playing with the code and making PRs with literally no chance to spot regressions at least. For example those #145 and #147 issues are candidates for fixing a problem while introducing a bigger one at some random place.

Tyriar:
The current plan is to add tests as we go, mocha was only just added so the more tests you can contribute in PRs, the better.

jerch:
I have some xterm output comparison tests, can try to apply them. They are high level therefore it will be tricky to spot the faulty piece of code with just those.
At low level stage I have complete tests for this [parser](http://www.vt100.net/emu/dec_ansi_parser), but xterm.js does the parsing differently. Maybe I can adopt some of them idk.

jerch:
Added the xterm tests --> https://github.com/jerch/xterm.js/commit/36930e9661123a86e44a8f2471e6dfed2f3b6824
10 of 44 basic tests are passing, this is a beginning :) The tests are just a high level indicator of xterm conformance (comparable to vttest) and not usable for tracking down a problem to a single line of code. That would need tests with a higher granularity (like testing every single function and escape sequence separately).
**Note:** I took many of the xterm test cases from https://github.com/MarkLodato/vt100-parser/tree/master/test and adjusted them slightly.

parisk:
That's great @jerch and it's definitely a very good start. Even getting 10 out of 44 vttest tests passing at first we will at least now if a PR breaks any of these (and there it will be much easier to debug the code).

I am really looking forward to seeing these tests incorporated in xterm.js. Please let me know how can I help you out with this.

jerch:
Had a closer look at the current escape sequence parsing. My test cases cant be applied to it, they are not compatible state and transition-wise. The parsing doesn't fit into Paul's reference parser model, therefore I'm not sure if the parser always does the right thing (Paul claims to have the most complete model). Up for discussion, since it would need a really big re-factoring to get it in line....

jerch:
@parisk Well the xterm tests would have to be tested and fixed one by one. You can select a specific test case by altering the for loop in the test file and switch the noisy output on to compare the outputs.
[...]

PR #161, opened 2016-07-05
xterm test cases

jerch:
Adding xterm conformance tests. See #156.

I fixed most of the test case input files (up to test 44), any higher test needs to be checked before using. I also switched off all failing test cases for now.

If you are going to apply the test cases please also give the original author of the input files the appropriate credits (https://github.com/MarkLodato).


test/escape_sequence_files/run_tests.py
```diff
@@ -0,0 +1,73 @@
+from glob import glob
```
parisk: Is this file used at all?

test/escape_sequences.js
```diff
@@ -0,0 +1,101 @@
...
+    // omit stack trace for escape sequence files
+    Error.stackTraceLimit = 0;
+    var files = glob.sync('test/escape_sequence_files/*.in');
+    // comment/uncomment one of the following
+    // only sucessful tests
+    var successful = [0, 2, 6, 12, 13, 18, 20, 22, 27, 28];
+    for (var a in successful) {
+        var i = successful[a];
+    // first 44 are the most basic escape sequences
+    //for (var i=0; i<44; ++i) {
```
parisk: Do we need this commented piece of code?

test/escape_sequences.js
```diff
@@ -0,0 +1,101 @@
...
+    // all tests
+    //for (var i=0; i<files.length; ++i) {
```
parisk: Same here as well (https://github.com/sourcelair/xterm.js/pull/161/files#r69700664).

parisk:
Hi @jerch thanks a lot for this contribution! I made some comments at first, after resolving I will dig deeper into this.

test/escape_sequence_files/run_tests.py
jerch: ah well with this file I created the expected xterm outputs. It sits there just in case some test case is faulty or needs to be adjusted. For testing itself it is not used.

test/escape_sequences.js
jerch: This will run all tests, not only the basic ones. The higher tests are more complex testing different escape code at once. They wont run unless the basic ones are fixed. Also some of the higher tests need to be fixed before usage, some of them even need a real pty with a shell between. I can go through the higher tests as we get there.

test/escape_sequences.js
jerch: Nope. It just sits there as I used it to track down the initial passing tests and needed it to switch back and forth.

parisk: Could you please remove this line then, since the repository doesn't get any value at this point from it? We 'd like to keep the codebase as lean as possible. We can bring this line back, after we make sure most (if not all) tests pass.

test/escape_sequences.js
parisk: Can we remove this commented line of code as well, since the repository does not get any value currently from it then? We can bring it back as soon as we fix the majority of tests.

test/escape_sequence_files/run_tests.py
parisk: Let's remove this file then, since it is not being used directly by this repository, in order to keep the codebase lean 😊 .

parisk:
@jerch thanks a lot for the last patch! Could you please also rebase with master and resolve conflicts? Then we can merge this.

PR #168, opened 2016-07-07
:arrow_up: pty.js@0.3.1

freebroccolo:
This PR updates `pty.js` to `0.3.1`. On my system (OS X `10.11.5`, node  `v6.3.0`) the package fails during the install stage with the previous version of `pty` but everything seems to work with `0.3.1` so far. I had the same problem on another system running an earlier version of node 6.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "devDependencies": {
     "express": "4.13.4",
     "express-ws": "2.0.0-rc.1",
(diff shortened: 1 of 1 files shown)


Tyriar:
:+1: Might be able to get rid of my fork with this

parisk:
Works great on OS X and Linux. Thanks!

Issue #68, opened 2016-06-02
Selection goes crazy when you start dragging and then use mouse wheel


Tyriar:
Start:

![image](https://cloud.githubusercontent.com/assets/2193314/15738918/78877e50-2862-11e6-91bc-17fbcb4ab59b.png)

Mouse wheel up:

![image](https://cloud.githubusercontent.com/assets/2193314/15738936/8ae2c244-2862-11e6-99d0-5e9f72df7c61.png)

parisk:
Wild guess is that this is related to #67.

Tyriar:
I think it's to do with the DOM being changed radically after a scroll, so the cursor ends up at the top-left (start) of the terminal.

parisk:
This could be the case as well. Implementing natural scrolling, instead of virtual one, seems like could fix most of these issues.

Tyriar:
That would probably lead to perf issues if the terminal buffer is too big, plus scrolling would not jump in character length blocks which would feel strange.

parisk:
We could apply a limit to the terminal buffer size, which is something that even native apps do (e.g. [Terminal.app](https://discussions.apple.com/thread/2123671?tstart=0)).

Tyriar:
I think it would be worth trying to get around the scrolling issues doing something like https://github.com/sourcelair/xterm.js/issues/69 before considering this route. That's a lot of DOM elements and imo the way it currently works is one of the great things about term.js; that it's footprint on the DOM is relatively small while maintaining a large buffer.

Plus it would probably need to recreate most lines on resize which could cause performance issues with a reasonably sized buffer limit.

Tyriar:
Fixed in https://github.com/sourcelair/xterm.js/pull/670

Issue #149, opened 2016-06-27
Lines containing unicode characters differ in height


Tyriar:
Using the demo, lines with CJK characters at 24px, regular lines are 18px, and my prompt character is 20px.

![image](https://cloud.githubusercontent.com/assets/2193314/16393767/e56a18d0-3c66-11e6-8fd0-d9f994726392.png)

This is the reason that on my computer the size of the terminal increases every time I press enter; because the `❯` character is being added, pushing out the height of the line.

parisk:
@Tyriar are you aware of any way we can fix this? Could we somehow (based on font-size I guess) know what is the appropriate line height to set, in order not to have lines with variant line heights?

Tyriar:
The way I'm probably going to approach a fit for vscode at least which allows a configurable `font-size` and `line-height`, is to make the default `line-height` a fixed pixel size and set it as the `height` as well. Setting `height` and `line-height` forces all chars to be aligned the same and all lines to be the same height.

Issue #69, opened 2016-06-02
Selection is only possible on the text in the current viewport


Tyriar:
While potentially difficult, it's probably possible to do a smooth selection over multiple pages by:
- Recording start selection position
- Scroll the terminal up when dragging from terminal and cursor is close to/beyond top
- Scroll the terminal down when dragging from terminal and cursor is close to/beyond bottom
- Whenever the terminal is scrolled, reset the selection position
- When copy is triggered, intercept and select the text between the start and end (probably won't work in older browsers)

parisk:
I think that this will be handled across with all other scrolling issues. The best thing to do IMO is bring scrolling as close as possible to native terminals found in modern OSs (e.g. Ubuntu Terminal and Terminal.app)

Tyriar:
[hterm](https://chromium.googlesource.com/apps/libapps/+/HEAD/hterm) handles this really well but only has the currently displayed viewport in the DOM. We should see how they implement it.

parisk:
We could mimic the way https://github.com/codemirror/codemirror does it. This could also help with #77.

Tyriar:
Fixed in https://github.com/sourcelair/xterm.js/pull/670

PR #175, opened 2016-07-12
Add textarea back and support IMEs

### LINKED ISSUE #124: Terminal doesn't accept input via IME
Using Ubuntu 16.04 native Korean IME the terminal just enters the English characters.

Tyriar:
Korean input using Korean (Hangul) (IBus) on Ubuntu 16.04

image

Japanese input using Japanese (Anthy) (IBus) on Ubuntu 16.04

image

image

Covered in this PR:

- [x] Manually revert: Drop hidden textarea for input handling #78
- [x] Support CSS focus state using the `.focus` class instead of a `:focus`
- [x] Support IMEs by hooking into composition events #124
- [x] Add a bunch of tests for `CompositionHelper`

Some of the testing that needs to be done on final version:

- [x] Copy using <kbd>ctrl</kbd>+<kbd>ins</kbd>
- [x] Paste using <kbd>shift</kbd>+<kbd>ins</kbd> and <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>v</kbd>
- [x] Inputting without IME continues to work as expected
- [x] Inputting via Korean IME
  - [x] Test: ㅇㅇ (automated)
  - [x] Test: 아아 (automated)
  - [x] Test: 앙앙 (automated)
  - [x] Pressing arrow keys to finalize composition
  - [x] Pressing enter to finalize and execute composition
  - [x] Numbers and punctuation can be input while IME is active (see remaining issues)
- [x] Inputting via Japanese IME
  - [x] Test multiple character compositions (automated)
  - [x] Test hiragana -> katakana conversion (automated)
  - [x] Test hiragana -> kanji conversion (automated)
- [x] Inputting non-composition character (eg. 1) after composition character (eg. ㅇ) (automated)
- [x] Check support on Windows (tested using Microsoft Korean IME and Google Japanese IME, works fine except for the fact that the terminal cursor disappears which makes the composition appear on top of text)
- [x] Check support on OS X (tested using Apple's 2-Set Korean IME)
- [x] Check browser support (seems to behave consistently across browsers)
  - [x] Chrome 51 (Linux, OS X, Windows)
  - [x] Firefox 48 (Linux)
  - [x] Safari 9.1.1 (OS X)

Fixes #124 (Terminal doesn't accept input via IME)

Remaining issues:

- Currently the composition view is placed on top of the cursor, it's not aligned currently due to #149 (Lines containing unicode characters differ in height)
- The composition view can appear on top of text on Windows only because the cursor seems to disappear. Input works fine, it's just that it can overlap on text.


Tyriar:
@parisk please review when you get a chance, so far I've only tested this on Korean but it's probably the most complex. I'll be able to test multi-character compositions with Japanese but it will likely be small tweaks inside `CompositionHelper` to get this working if it's not already.

The code itself is very well sectioned off from the rest so you don't have to worry so much about it introducing instability when not using an IME. A section is added to the top of `Terminal.prototype.keyDown` which prevents the key event going through if the `isComposing` flag is active. There is also a call at the end of refresh to position the composition view element on top of the cursor. All other changes not in `CompositionHelper` are just reverting the textarea change.

parisk:
Thanks a lot for this @Tyriar! Right now I will focus on what is being needed to release **1.0.0** by the end of the week (e.g. adding a CONTRIBUTING.md file, updating the README, etc.) and I will get on top of this starting next week at most.

Tyriar:
@parisk thanks, probably a good idea to release v1 before pushing a change this big. If we can land this within a week it will get a bunch of testing in our Insider's build before the stable release.

Tyriar:
Ok done and tested, at least under Chrome 51 and Ubuntu 16.04. I moved even more into the `CompositionHelper` so now the only changes made outside are:

- Creating the `CompositionHelper`
- Attaching events to it
- Calling into `CompositionHelper.prototype.keydown` manually at the beginning of `Terminal.prototype.keyDown`

yuvipanda:
(I can also help test this at paws.wmflabs.org once it lands!)

Tyriar:
@parisk good to go :+1: the only remaining issue is the IME helper can popup in the wrong spot on Windows due to the cursor disappearing. The fix for that likely lies in pty.js or winpty though.

parisk:
Thanks a lot @Tyriar! Moving on with testing on OS X.

src/xterm.js
```diff
@@ -757,11 +867,39 @@
...
+      this.helperContainer.classList.add('xterm-helpers');
+      // TODO: This should probably be inserted once it's filled to prevent an additional layout
+      this.element.appendChild(this.helperContainer);
+      this.textarea = document.createElement('textarea');
+      this.textarea.classList.add('xterm-helper-textarea');
+      this.textarea.setAttribute('autocorrect', 'off');
+      this.textarea.setAttribute('autocapitalize', 'off');
+      this.textarea.setAttribute('spellcheck', 'false');
+      this.textarea.tabIndex = 0;
+      this.textarea.onfocus = function() {
```
parisk: Could you please use `this.textarea.addEventListener('focus, ...)` instead here please?

src/xterm.js
```diff
@@ -757,11 +867,39 @@
...
+        self.emit('focus', {terminal: self});
+      }
+      this.textarea.onblur = function() {
```
parisk: Could you please use `this.textarea.addEventListener('blur', ...)` instead here please?

src/xterm.js
```diff
@@ -436,7 +625,18 @@
      * Focus the terminal. Delegates focus handling to the terminal's DOM element.
      */
     Terminal.prototype.focus = function() {
-      return this.element.focus();
+      if (document.activeElement === this.textarea) {
+        return;
+      }
+
+      if (this.sendFocus) {
+        this.send('\x1b[I');
```
parisk: This is already done inside the `focus` callback that is set up in the `Terminal.bindFocus` static method. I would prefer to keep any "business logic" inside high level callbacks. Defaults can be set up in `bindFocus` which is run on every new instance.

src/xterm.js
```diff
@@ -460,7 +660,18 @@
...
+      if (Terminal.focus !== this) {
+        return;
+      }
+
+      this.element.classList.remove('focus');
+      this.cursorState = 0;
+      this.refresh(this.y, this.y);
+      this.textarea.blur();
+      if (this.sendFocus) {
+        this.send('\x1b[0]');
```
parisk: This is already done inside the callback that is set up in the `Terminal.bindBlur` static method. I would prefer to keep any "business logic" inside high level callbacks. Defaults can be set up in `bindBlur` which is run on every new instance.

parisk:
🆗 , I made my first comments on code styling and I have a few more on functionality on OS X.

1. **Cannot paste using right click**. Quick hack proposal: Reposition the helper textarea below the mouse position on `mousedown` if the right click is being pressed. This should trigger the `contextmenu` event on the textarea and the user should be able to paste using right click.
2. **Cannot enter emojis**. Xterm ignores emojis completely when attempting to enter them with the native input method on OS X. Well this is **definitely non-blocking** for merging this PR (no matter how much I love emojis in the terminal 😍 ), but thought about mentioning.

 In general this PR behaves really well and is exceptionally written and documented! 👍  @Tyriar!

src/xterm.js
Tyriar: See https://github.com/sourcelair/xterm.js/pull/175/commits/4595a18197a9f7378077b078c06719ecb16adb38, let me know if that's what you meant.

Tyriar:
@parisk so I took a shot at fixing the paste on right click behavior and while I got it to work by positioning the textarea under the cursor and then redirecting the event to the textarea and cancelling the original event, it removes the selection which breaks means you cannot copy using right click. Here's the super messy/hacky code I used to accomplish this:

```
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -1220,7 +1220,41 @@
         };
       }
 
+function getOffset( el ) {
+    var _x = 0;
+    var _y = 0;
+    while( el && !isNaN( el.offsetLeft ) && !isNaN( el.offsetTop ) ) {
+        _x += el.offsetLeft - el.scrollLeft;
+        _y += el.offsetTop - el.scrollTop;
+        el = el.offsetParent;
+    }
+    return { top: _y, left: _x };
+}
+
+self.textarea.addEventListener('mousedown', function(ev) {
+  console.log('textarea mousedown', ev);
+ev.stopPropagation();
+});
+
       on(el, 'mousedown', function(ev) {
+
+        self.textarea.style.position = 'absolute';
+var offset = getOffset(self.element);
+        self.textarea.style.left = (ev.pageX - offset.left - 10) + 'px';
+        self.textarea.style.top = (ev.pageY - offset.top - 10) + 'px';
+self.textarea.style.width = '20px';
+self.textarea.style.height = '20px';
+self.textarea.style.zIndex = '0';
+        redirectMouseEvent(ev, self.textarea);
+
+function redirectMouseEvent(event, el) {
+  el.dispatchEvent(new MouseEvent(event.type, event));
+}
+ev.preventDefault();
+    ev.stopPropagation();
+    return;
+
+
         if (!self.mouseEvents) return;
 
         // send the button
```

Perhaps this should be deferred until selection is handled in a non-native way which would fix #68 and #69. I think that's how hterm behaves and it's how the VS Code editor maintains selection no matter where the focus is, seems like the right solution without compromising.

There is always <kbd>ctrl</kbd>+<kbd>shift</kbd>+<kbd>v</kbd> and <kbd>cmd</kbd>+<kbd>v</kbd> to paste now which are the defaults on Linux/Mac.

I'd prefer to get this in and accept the regression with the context menu for now as this change needs to get in for VS Code 1.4 (code freeze/testing starts on the 25th).

parisk:
🆗  let's move forward with merging this right now and I will create an issue in order to fix copy and paste with mouse until the next release (1.1.0) at 22 August 2016.

Tyriar:
https://github.com/sourcelair/xterm.js/pull/175/commits/4595a18197a9f7378077b078c06719ecb16adb38 actually broke the IMEs as it changed focus/blur behavior on the textarea, investigating...

PR #184, opened 2016-07-13
Add statement on browser support

### LINKED ISSUE #183: State official browser support in README
Recommended for v1.0 milestone, here's what the sourcelair documentation says:

- Chrome	48+
- Firefox	44+
- Safari	8+
- Internet Explorer	11+
- Opera	35+

Having this will also make it easier in the future to tell why a particular version is required as you can check the commit in which they were bumped.

Tyriar:
Fixes #183 

---

These can be tweaked as necessary later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,12 @@ Xterm.js supplies a modular, event-based interface that lets developers build ad
 
 ## Browser Support
 
-- Chrome	48+
-- Firefox	44+
-- Safari	8+
-- Internet Explorer	11+
-- Opera	35+
+- Chrome 48+
+- Edge 13+
+- Firefox 44+
+- Internet Explorer 11+
+- Opera 35+
+- Safari 8+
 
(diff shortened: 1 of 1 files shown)


parisk:
Great, thanks! Should we mention Edge as well, or is it considered part of Internet Explorer?

One non-blocking question also; are there any specific features that make xterm.js incompatible with IE <11, Chrome <48, etc.? If yes, does it make any sense to refer them?

Tyriar:
Yes I'll add Edge.

Potentially, it might be good to just state a baseline and adjust it accordingly though. Not sure I'll have time to investigate whether it works in the older browsers.

parisk:
> Potentially, it might be good to just state a baseline and adjust it accordingly though

👍  definitely. Let's add Edge and merge this!

parisk:
Thanks a lot @Tyriar!

PR #186, opened 2016-07-15
Allow custom keydown handler to be attached

### LINKED ISSUE #118: Support custom keybindings

Tyriar:
Fixes #118 

---

Example usage:

```js
// Don't handle any events
this.xterm.attachCustomKeydownHandler(function (ev) {
  return false;
});

// Handle everything except ctrl+m
this.xterm.attachCustomKeydownHandler(function (ev) {
  return !(ev.ctrlKey && ev.keyCode === 77);
});
```


src/xterm.js
```diff
@@ -2529,12 +2530,27 @@
...
+    }
+
+    /**
      * Handle a keydown event
      * Key Resources:
      *   - https://developer.mozilla.org/en-US/docs/DOM/KeyboardEvent
      * @param {KeyboardEvent} ev The keydown event to be handled.
      */
     Terminal.prototype.keyDown = function(ev) {
+      if (this.customKeydownHandler && !this.customKeydownHandler(ev)) {
```
parisk: Could you please change `!this.customKeydownHandler(ev)` with `(this.customKeydownHandler(ev) === false)` in order to require users to explicitly return false to stop further handling of keys?

src/xterm.js
```diff
@@ -2529,12 +2530,27 @@
     };
 
     /**
+     * Attaches a custom keydown handler which is run before keys are processed, giving consumers of
+     * xterm.js ultimate control as to what keys should be processed by the terminal and what keys
+     * should not.
+     * @param {function} customKeydownHandler The custom KeyboardEvent handler to attach. This is a
+     *   function that takes a KeyboardEvent, allowing consumers to stop propogation and/or prevent
+     *   the default action. The function returns whether the event should be processed by xterm.js.
+     */
```
parisk: Could you please also document the meaning and importance of return values here?

Tyriar: That's here:

>  The function returns whether the event should be processed by xterm.js.

Not sure if there is a way to type functions in jsdoc like in closure compiler.

parisk: It has the `@returns` directive; http://usejsdoc.org/tags-returns.html.

Tyriar: As in on where `this.customKeydownHandler` is defined? `@return` can't be used on `attachCustomKeydownHandler` since it's the return of the param function, not the function itself.

parisk: Yes, you 're right. Sorry, my bad!

parisk:
@Tyriar could you rebase with master please? I think we can move on with merging this afterwards.

Tyriar:
Looking into test failure

Issue #77, opened 2016-06-02
No scroll bar on terminal


Tyriar:
It would be great if a virtual scroll bar was provided.

parisk:
The terminal scrolling definitely needs an enhancement and physical scrolling with a scrollbar would be a great addition. The only tricky thing to do there is handling switching back and forth with curses applications like Vim.

Tyriar:
[hterm](https://chromium.googlesource.com/apps/libapps/+/HEAD/hterm) handles the scrollbar really well but only had the currently displayed viewport in the DOM. We should see how they implement it.

Tyriar:
https://chromium.googlesource.com/apps/libapps/+/HEAD/hterm/js/hterm_scrollport.js

Tyriar:
I'm going to work on this as part of vscode 1.5.0, the current rough plan/approach is:

- Detect row height and set `height` and `line-height` explicitly on `.xterm-rows` (will resolve #149)
- Wrap `.xterm-rows` in a `.xterm-viewport` element which has `overflow-y: scroll`
- Create an `.xterm-scroll-area` element which sizes based on the total number of rows in the buffer, this will be invisible but will position itself based on `y`, `ybase` and `ydisp`

parisk:
Sounds 👍 . What is the purpose of `.xterm-scroll-area` element though? Will it be used for the displaying of the scroll bar?

Tyriar:
@parisk yes, it's a large element that's programmatically sized and is used for the scrollbar. After playing with this implementation I ended up getting it working but needed to pull `.xterm-rows` out of `.xterm-viewport`, otherwise every refresh would recall the scroll method. That made it very difficult to handle the 3 scroll cases; wheel, drag and drop scroll bar and up/down arrows.

I'll probably put out a PR later today.

PR #188, opened 2016-07-16
Use constant line-height

### LINKED ISSUE #149: Lines containing unicode characters differ in height

Tyriar:
Unicode characters have a constant height when using a decimal (eg. 1.2) but
can overlap which looks unattractive on some browsers that use transparent
selections (Chrome). "normal" line-height differs for CJK and other unicode
chars, only exactly pixel sized seems to work all the time.

Fixes #149

### DIFF
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -2141,9 +2141,9 @@
 }
 
 /**
- * All terminal rows should have explicitly declared height,
- * in order to allow child elements to adjust.
+ * All terminal rows should have explicitly declared height, in order to allow child elements to
+ * adjust. line-height:normal does not provide a fixed height for all unicode characters.
  */
 .terminal .xterm-rows > div {
-    line-height: normal;
+    line-height: 18px;
 }


parisk:
Hmmm, not sure if hardcoding font-size to px is the right way to go here. What about setting both font size and line height to 1em?

Changing this also means that we should document how different font sizes should work on xterm.js (demanding updating of both font size and line height).

Tyriar:
I believe 1em translates directly to 16px, 18px just seemed to work for default font size. You can see how I handled this case in vscode here https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/parts/terminal/electron-browser/terminalPanel.ts#L292

Tyriar:
Taking more control over row height turns out to be a pre-requisite for a virtual scrollbar as the height of rows much be known. Going to close this and handle #149 in the PR for #77

PR #189, opened 2016-07-17
:memo: bring in the quickstart

basarat:
Assuming its a `fork` this should just work https://github.com/chjj/term.js#example :rose:

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,52 @@ npm start
 
 Then open http://0.0.0.0:3000 in a web browser (use http://127.0.0.1:3000 is running under Windows).
 
+## Example
+
+```
+npm install xterm
+```
+
+Server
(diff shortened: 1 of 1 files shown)


parisk:
Hi @basarat thanks for your contribution! This won't work actually in xterm.js as it does not use socket.io anywhere, while the example depends on it.

May I ask you why would you like to see this example incorporated in the README? Does the Demo section of the README fit your needs (https://github.com/sourcelair/xterm.js/#demo)? If not, what extra value would you like to get out of it?

basarat:
> Does the Demo section of the README fit your needs (https://github.com/sourcelair/xterm.js/#demo)

It does. The demo is good and easy to understand :rose: My bad ❤️

Issue #158, opened 2016-07-04
Introduce some build system to enable more than one js file


Tyriar:
It's difficult to simplify the code base if it all sits in a 5000 line file.

parisk:
I totally agree with that. Let' get on top of this after finishing documentation and releasing 1.0.0.

Tyriar:
I suggest webpack ([prototype here](https://github.com/sourcelair/xterm.js/pull/196)) as I've used it before and it does exactly what we want (bundles several files into 1 using `require()`). I've also barely touched the surface of its capabilities as well and it's very popular.

Tyriar:
Another option is [TypeScript](https://www.typescriptlang.org/) which I've become quite fond of after working on vscode. This would allow multiple files to be built into one and provide type safety which would promote code quality (especially good since unit tests are patchy).

parisk:
I think that plain ES6 or TypeScript are more suitable for this. Webpack seems to be more of an overkill at this moment.

The most important thing that we have to take into account is how are going to ship the code here. This most probably will introduce breaking changes (especially for Bower users), since we will need to find a way to ship the built artifacts to keep this library working in the browser out of the box.

I am adding this to the "open" milestone for version **2.0** and start some experimentation on this, preferably with plain ES6, which is close to TypeScript.

My main concerns with TypeScript are:

1. barrier to entry for new contributors
2. none of the rest maintainers has worked with TypeScript in the past (if I can recall right), me included

Tyriar:
When using bower most projects with a build system add the built files to a dist directory and those are the ones you link to. It would introduce a breakage though so v2 would be appropriate.

jerch:
+1 for using Typescript. No clue about the current state of ES6 support, I think generators are still at an early stage in some JS engines and therefore not usable yet.

Tyriar:
I'll assign myself to put together a TypeScript prototype.

131:
May i suggest you start with ES6 & Babel ? Leveraging typescript is a good thing, yet, it will come with a price.

Having an es6 project will allow your user to just "var Terminal = require('xterm')" and use the raw source code (browserify will work just fine)". When with typescript, you'll have a different file for the build and the source of your application.

May i help here ?

Tyriar:
Sourcemaps gets around that if needed though, the type safety is the main benefit imo.

131:
[...]

PR #196, opened 2016-07-19
Initial webpack implementation

Tyriar:
Part of #158 

---

@parisk this PR is a bit messy at the moment, let me know your thoughts on the general approach here. Basically the following changes would come with/after this change:

- Builds are done using webpack, the module is built at `npm install` time and the file that is relied upon is `./dist/xterm.js` (see `main` in `package.json`)
- Addons will be pulled out into their own repositories and published as independent modules, this means that not all addons need to be loaded at once and are more modular in nature
- We can split out all the different pieces of logic in xterm.js into a well structured tree
- Development is done by running:
  - First terminal: `npm run dev` to watch/compile sources
  - Second terminal: `npm start` to start server as before
- Eventually change to exposing the library in an xtermjs object that contains a Terminal object would be preferable particularly for TypeScript as a proper typings file could be written for the library (ie. `export { Terminal: Terminal };` vs `export Terminal`)
- We can start using babel or similar to enable more elegant/efficient syntax without hurting compatibility in the resulting distributable
- We can get rid of the AMD header as it's all handled by webpack
- Allow minification/removal of comments in prod version, pushing the size of the distributable down significantly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,6 @@ Makefile.gyp
 *.target.gyp.mk
 *.node
 example/*.log
+dist/
 docs/
 npm-debug.log
--- a/demo/app.js
+++ b/demo/app.js
@@ -4,8 +4,10 @@ var expressWs = require('express-ws')(app);
 var os = require('os');
 var pty = require('pty.js');
 
+app.use('/dist', express.static(__dirname + '/../dist'));
+// TODO: Expose CSS in some static dir, away from the JS
 app.use('/src', express.static(__dirname + '/../src'));
-app.use('/addons', express.static(__dirname + '/../addons'));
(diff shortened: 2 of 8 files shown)


Tyriar:
@parisk any thoughts on the approach? Didn't want to continue until choice of tech, method, etc. was cool.

parisk:
Sorry for being late here @Tyriar.

I have no significant webpack knowledge at the moment and TBH I am really reluctant to move forward with such a move without laying down first the pros and cons and examining alternatives.

I think that #158 is the best place to make some discussion about our options and choose the one that fits the project's needs better 😊 .

Tyriar:
Closing as it's stale and going to try a TypeScript implementation.

PR #203, opened 2016-07-22
Correct blur/focus delegation

Tyriar:
Focus and blur events on the Terminal object are delegated to the textarea
which handles everything.

---

@parisk please merge if it looks fine, what was happening before was that after the small refactor in https://github.com/sourcelair/xterm.js/pull/175/commits/4595a18197a9f7378077b078c06719ecb16adb38, the IME was toggling focus one too many times like so:

1. Focus Terminal
2. Blur textarea
3. Focus textarea

This is changed to:

1. Focus Terminal, which delegates to textarea

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -625,7 +625,7 @@
      * Focus the terminal. Delegates focus handling to the terminal's DOM element.
      */
     Terminal.prototype.focus = function() {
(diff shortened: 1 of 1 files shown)


parisk:
Looks great! Thanks @Tyriar! 👍

PR #210, opened 2016-07-28
Hide textarea off-screen when not composing

### LINKED ISSUE #209: Textarea is positioned near the cursor even when the composition is not active
See https://github.com/Microsoft/vscode/issues/9747 for context.

Tyriar:
Fixes #209

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -324,7 +324,6 @@
       this.textarea.style.top = '';
     }
 
-
     /**
      * States
      */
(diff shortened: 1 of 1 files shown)


parisk:
Works great for me 👍 .

parisk:
Seems like this test has passed but Travis did not update the status accordingly: https://travis-ci.org/sourcelair/xterm.js/builds/148110796.

parisk:
Merged as it needed admin access, because of the commit status not reported.

Issue #198, opened 2016-07-20
It would be cool if xterm.js described its mission or distinguishing features


vincentwoo:
There's beginning to be a bunch of JS terminal emulators in active development and use, in addition to this one. Notably, 

* [Hyperterm](https://hyperterm.org/) with [hterm](https://chromium.googlesource.com/apps/libapps/+/master/hterm) underneath. I don't know too much about hterm.
* [term.js](https://github.com/chjj/term.js/), which xterm.js is based on.
* [terminal.js](https://github.com/Gottox/terminal.js), which tries to support several renderers.

As a user, it can be hard to know which one to pick. I think it would be great if the Readme included a statement concerning:

* What use cases xterm.js focuses on
* What limitations xterm.js has at this time
* What important things are planned for the future
* How xterm.js differs from other libraries, either in implementation or philosophy

What do you think?

Tyriar:
I agree, it would be good to have some more detail in the README. Here are some of my thoughts on the topic, in no particular order:

- xterm.js succeeds term.js, pushing it forward as the original author no longer has the time to maintain term.js
- The current use case being focused on is editor integration as the two main products integrating it are VS Code and SourceLair
- To be clear: hterm is comparable to xterm.js, Hyperterm is not.
- IME support will be a big stand out feature once it lands #175, I know hterm doesn't handle this, not sure about terminal.js
- Currently while performance is not perfect by any means (#134, #150), xterm.js seems to perform significantly better than hterm when processing large amounts of data.
- Some areas for improvement are general compatibility with xterm and test coverage https://github.com/sourcelair/xterm.js/issues/194 #156

parisk:
This is a great idea @vincentwoo, thanks for proposing! I am putting this into the `1.1.0` milestone, which is set for release approximately one month after the `1.0.0` release.

All of @Tyriar's comments about xterm.js are true. In addition we should mention that xterm.js is a self contained library. This means that it has no external dependencies, thus it is lighter and easier to integrate. While this might change in the future if there are important reasons to do so, we have no such plans right now.

vincentwoo:
Thanks for the responses. Another thing I think could be interesting to say is that xterm might get a lot more eyeballs on it because of its exposure in VS Code. I'm not sure what the relative popularity is like, but it seems like a relevant bit.

parisk:
@vincentwoo could you please take a look at the attached patches at #219? It should clarify things a little bit hopefully 😊 .

PR #214, opened 2016-08-02
Make linkify links open in new tab/window

mentos1386:
I believe this behavior is more appropriate.

### DIFF
--- a/addons/linkify/linkify.js
+++ b/addons/linkify/linkify.js
@@ -54,7 +54,8 @@
    *								  		 "linkified".
    * @param {boolean} lenient - The regex type that will be used to identify links. If lenient is
    *                            false, the regex requires a protocol clause. Defaults to true.
-   * @param {string} target -  Sets target="" attribute with value provided to links. Defaults to '_self'
+   * @param {string} target -  Sets target="" attribute with value provided to links.
+   *                           Default doesn't set target attribute
    * @emits linkify
    * @emits linkify:line
    */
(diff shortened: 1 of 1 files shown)


vincentwoo:
I think the appropriateness of this flag can vary wildly by setting. Here's a random opinion piece on the topic: https://css-tricks.com/use-target_blank/.

I think setting `target=_blank` by default is fine, but it would be cool if you could toggle it off.

parisk:
Hi @mentos1386, thanks a lot for your contribution! I think it fits the needs of xterm.js a lot, based on its main use cases. @vincentwoo is right though that the appropriateness of this flag can vary and if #198 was closed it would definitely help.

What I would like to see here in order to move with merging is making this an option, by adding an additional argument to the `linkify` and `linkifyTerminalLine` called `target` that will default to `null` and if set it would set the link's `target` attribute to the given value.

At this moment it is important not to break the current behavior of xterm.js.

mentos1386:
I changed it so that you can provide argument and it will set target to what you provide. Default behavior is that it doesn't change anything.

addons/linkify/linkify.js
```diff
@@ -67,6 +69,8 @@
       throw new TypeError(message);
     }
 
+    typeof target === 'undefined' ? target = '' : target = 'target="' + target + '"';
```
parisk: Could you please turn this in the following form to make it more human understandable?

```
if (typeof target === 'undefined') {
  target = '';
} else {
  target = 'target="' + target + '"';
}
```

parisk:
Works great, thanks @mentos1386 🎉 !

PR #218, opened 2016-08-03
Implement scroll bar

### LINKED ISSUE #77: No scroll bar on terminal

### LINKED ISSUE #149: Lines containing unicode characters differ in height

Tyriar:
Fixes #77
Fixes #149

### Implementation

The scroll bar is implemented using a `.xterm-scroll-area` element which is resized based on what the entire buffer is, this is contained within `.xterm-viewport` which uses `overflow-y: scroll` to show a native scroll bar. The viewport is kept separate from the rows in order to prevent calls to `refresh` triggering `scroll` events on the viewport.

Scroll and wheel events are then delegated to the viewport which in turn passes it on to the terminal. This can happen in two ways, events originating from the viewport (mouse wheel, drag and drop scroll bar, up/down buttons on scroll bar) and events originating from the terminal (scrolling via keyboard, new output, etc.).

The implementation supports various scroll configurations, properly utilizing `WheelEvent.deltaY` based on the [`WheelEvent.deltaMode`](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode), this allows certain configurations to specify whether the user prefers to scroll by pixels, lines or pages.

### Remaining tasks

- [x] Write some tests
- [x] Manually test on Windows
  - [x] Edge
  - [x] Chrome
  - [x] Firefox
  - [x] IE11
- [ ] See if anything can be improved with OS X scroll bar visibility
- [ ] Allow custom `line-height`?
- [x] Disable/fill scroll bar when in an application such as `vim`

### Notes

- I dropped support for `mousewheel` and `DOMMouseScroll` events in favor of `WheelEvent` since it's just complicating the support and xterm.js' [supported browsers](https://github.com/sourcelair/xterm.js#browser-support) are well within [the browsers supporting `WheelEvent`](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent#Browser_compatibility)
- Eventually a bunch of `Terminal`'s variables (`ydisp`, `y`, `ybase`) could be pulled into `Viewport` to simplify code, the virtual selection implementation would probably best live here too.
- Safari's scroll bar seems a little glitchy around drag of the scroll bar. This is not a big issue though considering it's probably not a very common use case on Macs.

### What it looks like

Ubuntu 16.04/Chrome 52:

image

Ubuntu 16.04/Firefox 47

image

Windows 10/Edge 14

image

### DIFF
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -43,7 +43,8 @@
     position: relative;
 }
 
-.terminal.focus {
+.terminal.focus,
+.terminal:focus {
     outline: none;
 }
 
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -103,6 +103,13 @@
     display: block;
 }
 
+.terminal .xterm-char-measure-element {
+    display: inline-block;
+    visibility: hidden;
+    position: absolute;
+    left: -9999em;
+}
+
 /*
  *  Determine default colors for xterm.js
  */
@@ -2168,12 +2175,4 @@
 
 .terminal .xterm-bg-color-255 {
     background-color: #eeeeee;
-}
-
-/**
- * All terminal rows should have explicitly declared height,
- * in order to allow child elements to adjust.
- */
-.terminal .xterm-rows > div {
-    line-height: normal;
 }
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -103,6 +103,20 @@
     display: block;
 }
 
+.terminal .xterm-viewport {
+    overflow-y: scroll;
+}
+
+.terminal .xterm-rows {
+    position: absolute;
+    left: 0;
+    top: 0;
+}
+
+.terminal .xterm-scroll-area {
+    visibility: hidden;
+}
+
 .terminal .xterm-char-measure-element {
     display: inline-block;
     visibility: hidden;
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -104,6 +104,8 @@
 }
 
 .terminal .xterm-viewport {
+    /* On OS X this is required in order for the scroll bar to appear fully opaque */
+    background-color: #000;
     overflow-y: scroll;
 }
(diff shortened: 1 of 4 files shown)


parisk:
The scrollbar works great in OS X Chrome. A couple of things I noticed that it would be great if they were fixed:

- Clicking on the scrollbar focuses on the terminal's parent element, which gets an outline styling (see attached screenshot)
- The scrollbar appears only after scrolling first using the trackpad (like before), while on OS X if a scrollbar exists, it should appear even when hovering the corresponding element

## Screenshot
screenshot

parisk:
Thanks a lot for this @Tyriar 🎉 .

soichih:
How can I try this? Cloning the master doesn't seem to add the scrollbar. Is there any instruction?

parisk:
Hi @soichih! You have to checkout this branch at @Tyriar's fork. To do this, run the following commands in your terminal:

```
git checkout -b Tyriar-77_scroll_bar master
git pull https://github.com/Tyriar/xterm.js.git 77_scroll_bar
```

Let us know if you ran into any trouble.

Tyriar:
@parisk RE your comments:

- I added `.terminal:focus` rule to remove outline
- It seems to be the platform convention that hovering over the side does not trigger the scrollbar, tested on Chrome, Outlook and Safari.

I fixed the issue with the transparent scroll bar on OS X. This is ready to merge imo, let me know what you think.

parisk:
👍  awesome. LGTM.

Feel free to merge!

PR #219, opened 2016-08-04
Update README and remove screenshot

### LINKED ISSUE #198: It would be cool if xterm.js described its mission or distinguishing features

parisk:
Update README to add more information about what xterm.js is, what are its use cases and who maintains it.

Also removed the screenshot, as it might be more confusing than helpful.

Closes #198.


vincentwoo:
good start! long term I think there really should be docs on how xterm works in practice (how does it handle line wrapping? can you use css word-wrap or does it need to be literally the same number of rows/cols as the backend terminal?) but getting the ball rolling is great

couple typos:

It enables applications **to** provide fully featured terminals to their users and create great development experiences.

Modular, event-based API: Let's (no **apostrophe**) you build addons and themes with **ease**

parisk:
Thanks for the comments @vincentwoo! I fixed the typos.

You can find the API reference at http://docs.xtermjs.org.

Also we have started a repo for the website but no work has been done there yet: https://github.com/xtermjs/xtermjs.org. Hopefully I will get on top of this within the next couple of weeks.

vincentwoo:
I didn't know about the docs, but I don't think it constitutes a "real" API reference. It's mostly one-sentence descriptions of various functions supported by xterm. I believe most of the doc content is basically old WIP commentary left by @chjj when he wrote the original term.js. It's not clear what changes you've made since, and thus the docsite feels very untrustworthy.

A "real" reference would explain the internal model of xterm, its corner cases, and how you might achieve certain usecases, etc.

PR #221, opened 2016-08-05
Update pty.js

odino:
ptyjs fails to build on node 6 :) didnt test on old versions  of node -- can you give it a look?


parisk:
Thanks for that! We had bumped this to `0.3.1` but it got rolled back by mistake.

odino:
np!

PR #223, opened 2016-08-08
By more explicit in security warning

Tyriar:
Calling out that the insecurity is due to the barebones implementation
of the demo will prevent people thinking that the library is inherently
insecure.


parisk:
Thanks @Tyriar! Clarification is always welcome.

PR #224, opened 2016-08-08
Sync scroll bar height if the viewport height changes

Tyriar:
Missed this in original implementation. When the viewport height changes, the resize event will trigger a `syncScrollHeight`, this change is required to force a `refresh`.


parisk:
👍  works great for me and code LGTM.

PR #226, opened 2016-08-12
Reposition textview after composition view

### LINKED ISSUE #208: The terminal IME composition view appears on top of the text when at the bottom of the screen
Using ibus Japanese IME, notice the view appears on top of the second character that is being composed.

image

## Details
- OS version: Ubuntu 16.04
- xterm.js version: 65b2be7

Tyriar:
Call the function again via setTimeout to allow changes after composition
events. This prevents the IME windows jumping sometimes on thesecond key
press.

Fixes #208

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -302,7 +302,7 @@
      * Positions the composition view on top of the cursor and the textarea just below it (so the
      * IME helper dialog is positioned correctly).
      */
-    CompositionHelper.prototype.updateCompositionElements = function() {
+    CompositionHelper.prototype.updateCompositionElements = function(dontRecurse) {
       if (!this.isComposing) {
         return;
       }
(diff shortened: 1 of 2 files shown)


src/xterm.js
```diff
@@ -302,17 +302,21 @@
...
       if (cursor) {
         this.compositionView.style.left = cursor.offsetLeft + 'px';
         this.compositionView.style.top = cursor.offsetTop + 'px';
-        this.textarea.style.left = cursor.offsetLeft + 'px';
+        var compositionViewBounds = this.compositionView.getBoundingClientRect();
+        this.textarea.style.left = cursor.offsetLeft + compositionViewBounds.width + 'px';
         this.textarea.style.top = (cursor.offsetTop + cursor.offsetHeight) + 'px';
       }
+      if (!dontRecurse) {
+        setTimeout(this.updateCompositionElements.bind(this, true), 0);
```
parisk: I would propose changing `0` to `false` as `dontRecurse` seems to be closer to boolean semantics, rather than to number. Not breaking though.

parisk:
Just a tiny comment. Other than this LGTM.

src/xterm.js
Tyriar: `setTimeout` takes a number though, not sure it's defined what is meant to happen with a boolean?

parisk: Nah, never mind, I got confused with it. Disregard my comment. Sorry 😅.

parisk:
LGTM 👍 .

PR #228, opened 2016-08-13
restore customKeydownHandler on reset(RIS)

ayapi:
hello, thank you cool project
i noticed, my `customKeydownHandler` may be lost.
it happens after i execute `less` , `fzf`, etc.
these commands have something in common. they raise "ESC c Full Reset (RIS)".
please review this pr for fix it.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -3467,7 +3467,9 @@
     Terminal.prototype.reset = function() {
       this.options.rows = this.rows;
       this.options.cols = this.cols;
+      var customKeydownHandler = this.customKeydownHandler;
       Terminal.call(this, this.options);
+      this.customKeydownHandler = customKeydownHandler;
       this.refresh(0, this.rows - 1);
     };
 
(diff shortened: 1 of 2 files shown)


src/xterm.js
```diff
@@ -3467,7 +3467,9 @@
     Terminal.prototype.reset = function() {
       this.options.rows = this.rows;
       this.options.cols = this.cols;
+      var customKeydownHandler = this.customKeydownHandler;
       Terminal.call(this, this.options);
```
Tyriar: Maybe this should be `Terminal.call(this, this.options, this.customKeydownHandler)`? Thoughts @parisk?

ayapi: thank you,
how about we add `customKeydownHandler` to `options` 's member?

Tyriar: Not sure if we would want a function to be an option?

parisk: Hmmm... not sure about which is the best way to go and I do not want to bike shed TBH. I believe that currently proposed implementation does the work and it does not break any public APIs, so I am 👍  with it.

parisk:
Thanks a lot @ayapi for your contribution. LGTM 👍 .

ayapi:
thank you very much, i'll use this terminal every day!

PR #230, opened 2016-08-17
Refresh the height of the viewport when the line-height changed

ayapi:
hello,
we can't scroll buffer correctly after the line-height was changed, until `rows` is changed.
this pr will fix it, so please check.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -357,12 +357,14 @@
     Viewport.prototype.refresh = function(charSize) {
       var size = charSize || this.charMeasureElement.getBoundingClientRect();
       if (size.height > 0) {
-        if (size.height !== this.currentRowHeight) {
+        var rowHeightChanged = size.height !== this.currentRowHeight;
+        if (rowHeightChanged) {
           this.currentRowHeight = size.height;
           this.viewportElement.style.lineHeight = size.height + 'px';
           this.terminal.rowContainer.style.lineHeight = size.height + 'px';
         }
-        if (this.lastRecordedViewportHeight !== this.terminal.rows) {
+        var viewportHeightChanged = this.lastRecordedViewportHeight !== this.terminal.rows;
+        if (rowHeightChanged || viewportHeightChanged) {
           this.lastRecordedViewportHeight = this.terminal.rows;
           this.viewportElement.style.height = size.height * this.terminal.rows + 'px';
         }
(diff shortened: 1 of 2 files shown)


src/xterm.js
```diff
@@ -357,12 +357,15 @@
...
+        var rowHeightChanged = false;
         if (size.height !== this.currentRowHeight) {
           this.currentRowHeight = size.height;
           this.viewportElement.style.lineHeight = size.height + 'px';
           this.terminal.rowContainer.style.lineHeight = size.height + 'px';
+          rowHeightChanged = true;
         }
-        if (this.lastRecordedViewportHeight !== this.terminal.rows) {
+        if (rowHeightChanged
+          || this.lastRecordedViewportHeight !== this.terminal.rows) {
```
parisk: Could you please save `this.lastRecordedViewportHeight !== this.terminal.rows` in a variable with a semantic name (e.g. `viewportHeightChanged`) and then move it on the line above (e.g. `if (rowHeightChanged || viewportHeightChanged) {`), for better readability?

ayapi: @parisk thank you, i have modified that as you said just now.

parisk:
Great, @ayapi can you provide some instructions on how to test this along with:

1. what was happening before
2. what should we expect now

src/xterm.js
```diff
@@ -357,12 +357,15 @@
     Viewport.prototype.refresh = function(charSize) {
       var size = charSize || this.charMeasureElement.getBoundingClientRect();
       if (size.height > 0) {
+        var rowHeightChanged = false;
         if (size.height !== this.currentRowHeight) {
```
Tyriar: Related to the previous comment, could you put this into `rowHeightChanged` to remove the line at the end of this block:

```js
var rowHeightChanged = size.height !== this.currentRowHeight;
if (rowHeightChanged) {
```

Looks good to me other than that, thanks for finding this :smiley:

Tyriar:
@parisk I think you can repro this currently in VSCode Insiders by changing `terminal.integrated.lineHeight` and trying to scroll.

ayapi:
@parisk 
i've finished to write a minimum example code.
please check this gist
https://gist.github.com/ayapi/9ecee21a70c2d56516799fdcceaa26db
and run it as demo `http://0.0.0.0:3000/`

## before
xterm-refresh-viewport-height-before

we can't see last line, we also can't scroll down.

## after
xterm-refresh-viewport-height-after

src/xterm.js
ayapi: @Tyriar thank you, i've done!

parisk:
Got it! Works great. Thanks for your contribution!

ayapi:
thank you very much!

Tyriar:
Thanks @ayapi, works great in vscode :smiley:

PR #231, opened 2016-08-17
Fix incorrect mouse position

meefik:
Incorrect mouse position in Midnight Commander


parisk:
Thanks a lot @meefik! Works great!

PR #232, opened 2016-08-17
[WIP] Implement virtual selection

parisk:
This is still in very early PoC state and it is barely usable. I am opening this mainly for feedback on the implementation.

Still lots to be done here.

### DIFF
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -2193,3 +2193,20 @@
 .terminal .xterm-bg-color-255 {
     background-color: #eeeeee;
 }
+
+.terminal::-moz-selection,
+.xterm-row::-moz-selection,
+.xterm-row span::-moz-selection {
+    background-color: transparent;
+}
+
+.terminal::selection,
+.xterm-row::selection,
+.xterm-row span::selection {
+    background-color: transparent;
+}
+
+.xterm-selection{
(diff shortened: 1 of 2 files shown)


src/xterm.css
```diff
@@ -2193,3 +2193,20 @@
...
+}
+
+.terminal::selection,
+.xterm-row::selection,
+.xterm-row span::selection {
+    background-color: transparent;
+}
+
+.xterm-selection{
+    background-color: yellow;
```
parisk: This is here just to make sure that the virtual selection stands out.

Tyriar:
I was thinking to bind mouseup/mousedown, tracking the start and end position. Then draw 1-3+ rectangles `div`s under the text but over the background, depending on which text was selected. This would enable things like selecting more text than is in the viewport and scrolling the terminal when the mouse is down and the cursor approaches the top/bottom boundaries.

Some things to consider:

- In a browser, copying will probably require copying the text to a temporary textarea
- Middle click to paste may be messed up?
- I would like to keep the inverted selection color on vscode so maybe more than 3 rectangles would be required, what about adding `.selected` to `span` elements when the entire `span` is selected

parisk:
Closing because of #293.

Issue #151, opened 2016-06-28
zsh / oh-my-zsh - no scroll


wallverb:
If i start demo with `zsh` instead of `bash` and have `oh-my-zsh` installed then the scroll will not work in xterm.js

I am not sure if this is an issue with `xterm.js` or `oh-my-zsh`

The command that causes scroll to stop working is this command: `echoti smkx` that is located in `.oh-my-zsh/lib/key-bindings.zsh`

One can also start `xterm.js` without `oh-my-zsh` and type that command into terminal and it will stop scrolling.

parisk:
Thanks for reporting @wallverb. I do not have access to zsh right now, so I am not able to test this out.

Could you please take a look at your console in the developer tools of your browser and provide us with any errors/logs that might appear when attempting to scroll when in zsh with oh-my-zsh?

wallverb:
@parisk - when zsh/oh-my-zsh is executing that `echoti smkx` command then the following data is being sent to   `Terminal.prototype.write = function(data) `:

```
[?1l>
xterm.js:1467 


xterm.js:1467 ]2;echoti smkx
xterm.js:1467 ]1;echoti[?1h=[1m[7m%[m[1m[m    
```

which seems like that `=` char is causing this:

```
 // ESC = Application Keypad (DECPAM).
              case '=':
                this.log('Serial port requested application keypad.');
                this.applicationKeypad = true;
                this.state = normal;
                break;
```

and then ` this.applicationKeypad = true;` is causing return in scroll code:

```
      on(el, wheelEvent, function(ev) {
        if (self.mouseEvents) return;
        if (self.applicationKeypad) return;
```

richarddavenport:
@parisk, any movement on fixing this? Really appreciate the work!

parisk:
@richarddavenport unfortunately I didn't have time to deal with this and I do not project to be able to do so by the end of the month.

It would be extremely helpful if we could get an external contribution on this and I can definitely help anyone out with figuring out the internals of xterm.js.

richarddavenport:
@parisk I'd love to help! Is there any documentation anywhere, or could you help me get going with debugging? Is there an editor or IDE that works best? VS Code/Webstorm/Chrome Dev Tools? Thanks!

Tyriar:
[...]

PR #236, opened 2016-08-19
Disable onScroll event during application mode

### LINKED ISSUE #235: Entering application mode with a large buffer will scroll to the top of the buffer temporarily
This is happening due to the reduction of the scroll bar when entering application mode but the scrollTop event is still going off.

See https://github.com/Microsoft/vscode/issues/10685 for original report.

Tyriar:
Fixes #235

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -425,6 +425,10 @@
      * @param {Event} ev The scroll event.
      */
     Viewport.prototype.onScroll = function(ev) {
+      if (this.isApplicationMode) {
+        // Scrolling via the scroll bar is disabled during application mode
+        return;
+      }
       var newRow = Math.round(this.viewportElement.scrollTop / this.currentRowHeight);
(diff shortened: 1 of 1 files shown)


ayapi:
@Tyriar ,
related to that i mentioned on #151, i think, to begin with, we shouldn't freeze scrolling on the viewport.
because, on original XTerm, even if it entered to application mode, we can scroll by <kbd>shift+pageup</kbd> and <kbd>shift+pagedown</kbd>

Tyriar:
@ayapi scrolling using the scroll bar already is disabled, this just prevents a weird bug where it scrolls to top when entering an app. Mouse wheel events should probably be passed to the application rather than handled by the scroll bar during application mode (which they aren't doing currently), for example to scroll the cursor in vim.

ayapi:
@Tyriar thank you, i think maybe i get it.

parisk:
Works great! LGTM 👍 .

PR #243, opened 2016-08-23
Disable application mode in viewport on soft reset

### LINKED ISSUE #242: Scroll stops working after running vim
## Details
- Browser and browser version: Chrome 52.0.2743.116 (64-bit)
- OS version: OSX 10.11.6
- xterm.js version: master (01c20807c255911174289a015ed79a986c25dfae) I guess its 1.1.2.

### Steps to reproduce

1. Make sure there is scrollable content on the terminal (e.g. run `dmesg`).
2. Make sure you can scroll the text using normal mouse wheel scrolling.
3. Start and quit `vim`.
4. Realize you no longer can scroll the text (it can scroll again if you press the `ENTER` key).

Tyriar:
Fixes #242 

---

I could not reproduce but it makes sense that this would it be, it's the only place that `Viewport.setApplicationMode` is not being set but should be.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -4561,8 +4561,8 @@
             break;
           case 66:
             this.log('Switching back to normal keypad.');
-            this.viewport.setApplicationMode(false);
             this.applicationKeypad = false;
+            this.viewport.setApplicationMode(false);
             break;
           case 9: // X10 Mouse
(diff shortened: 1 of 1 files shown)


parisk:
This should be OK to merge no.

PR #245, opened 2016-08-23
ES6 modularisation

131:
Fix  https://github.com/sourcelair/xterm.js/issues/158

Slice code in ES6 module.

Compositionhelper, viewport & eventemitter are proper ES6 classes.
The Terminal classe is big, yet wcwidth and escapeSequence have been sliced too, this increase overall readability and testing scope. ( slicing can be done on other parts of Terminal class, if desired)

Tests are fine
Documentation generation is fine

new pm task
* npm run build:dist to build standalone version ( exactly same signature as current version) in /dist folder

Strongly simplify demo/app.js (server side)
  no need for external fetch, just use WS socket all along
Strongly simplify demo/index.html / demo/main.js

Strongly simplify addon design & loading

in package.json
  node_modules/.bin is in npm run path, so no need to set it

PR #251, opened 2016-08-26
Introduce Babel as the build system, convert src/xterm.js to ES2015 and export EventEmitter

### LINKED ISSUE #158: Introduce some build system to enable more than one js file

parisk:
This PR introduces the following major changes in the codebase:

- sets Babel (with the es-2015 preset) as the build system for xterm.js
- converts `src/xterm.js` to an ES2015 native module
- exports EventEmitter into its own module to demonstrate how this should work from now on

and eventually just closes #158.

## Goals
- lay the ground to break the core of xterm.js into multiple files to increase maintainability
- make only the needed changes to achieve the above point and consequently keep this as reviewable as possible

## Details
- Babel is used in conjuction with browserify, derequire and exorcist in order to build the core into a standalone UMD module that can run in CommonJS, RequireJS and plain browser environment
- `bin/build` is introduced to build xterm.js
    - regular builds go to `build/` directory, which is ignored by Git
    - release builds go to the `dist/` directory and are committed into the repository
- `bin/demo-server` is introduced to server the demo
    - builds all files before starting
    - is being controlled by nodemon to auto-restart when a file changes (in the core or an add-on)

## Known issues
- The indentation level of xterm.js is totally wrong at this moment, after I removed the wrapping function, but I kept it as-is to maintain reviewability. This will be fixed in a next PR.
- Addons are not pulled into their own directories, neither they got converted to ES 2015 in order to focus on as less things as possible in this PR. They still work with their old public API. This will be fixed in a next PR as well.

## ⚠️ Breaking changes
- ~~The constructor class of xterm.js is now accesses through `xterm.Terminal`. This is valid across all module systems (CommonJS, RequireJS and plain browser environment).~~ This changed and now the Public API of `Terminal` stays the same.
- Xterm is now imported through the `dist` directory instead of `src`

**What I would like to achieve, but I did not manage to do** is maintain the same public API for xterm, but there was no way I could get around this with browserify and UMD export.

Rollup worked great with this (before switching to babel), but it completely messed up the tests and I couldn't fix them.

This PR is greatly inspired by the great work done in #245 and #247 by @131 and @Tyriar.

While both of the above PRs are great, both of them included additional code and concepts that were not described in #158 (e.g. introducing TypeScript for type safety, or changing the communication protocol of the demo) and introduce complexity in reviewing of both PRs.

As a result I would like us to address all those topics in additional standalone issues, in order to be certain about our choices.


131:
@parisk , would you please reconsider #245  if i just remove the changes in the demo app ?
I'll give more details in a few seconds

131:
1) using shell scripts will break windows build (there is no need for that, a npm task works just fine)
(see package.json in my PR)

2) Changing the public API (expose xTerm.Terminal) is un-necessary, as xterm (repository) is composed of proper ES6 modules, developpers wanting to access ViewPort will just need to do a require('xterm/utils/viewport')

131:
In the last commit of my PR, i'v preserved the /src/ folder as it was in the master branch, so now #245  provides : 

* no api change
* no directory structure change
* es6 modularisation
* windows build compatibility
* (and the workflow in the demo app)

bin/demo-server
```diff
@@ -0,0 +1,5 @@
+#! /usr/bin/env sh
+
+# Watch for file changes inside the `src` and `addons` directories.
+# Every time a file changes, then build xterm.js and restart the demo server.
+node_modules/.bin/nodemon --watch src --watch addons --exec bash -c "./bin/build && node demo/app"
```
Tyriar: Any non-trivial scripts like this should be put directly in scripts for cross-platform compatibility.

bin/build
```diff
@@ -0,0 +1,19 @@
+#! /usr/bin/env sh
```
Tyriar: We should have a plan to bring Windows compat back for the build.

src/xterm.js
```diff
@@ -5444,6 +5358,4 @@
     Terminal.off = off;
     Terminal.cancel = cancel;
 
-
-    return Terminal;
-});
+export { Terminal };
```
Tyriar: Does this mean you're for the API change I proposed?

src/xterm.js
```diff
@@ -63,76 +46,11 @@
...
-
-    /**
-     * EventEmitter
-     */
-
-    function EventEmitter() {
-      this._events = this._events || {};
+    // Let it work inside Node.js for automated testing purposes.
+    if (typeof document == 'undefined') {
+      var document = null;
```
Tyriar: Would be clearer if var was defined outside of the if:

```js
var document;
if () { ... }
```

or

```js
var document = typeof document == 'undefined' ? null : document;
```

test/viewport-test.js
```diff
@@ -1,11 +1,12 @@
 var assert = require('chai').assert;
-var Terminal = require('../src/xterm');
+var Terminal = require('../src/xterm').Terminal;
 
 describe('Viewport', function () {
   var terminal;
   var viewportElement;
   var charMeasureElement;
   var viewport;
+  var scrollAreaElement;
```
Tyriar: Why is this needed?

bin/prepare-release
```diff
@@ -17,12 +17,16 @@ CURRENT_BOWER_JSON_VERSION=$(cat bower.json \
   | sed 's/[",]//g' \
   | tr -d '[[:space:]]')
 
-# Update version in package.json and bower.json
-sed -i "s/\"version\": \"$CURRENT_PACKAGE_JSON_VERSION\"/\"version\": \"$NEW_VERSION\"/g" package.json
-sed -i "s/\"version\": \"$CURRENT_BOWER_JSON_VERSION\"/\"version\": \"$NEW_VERSION\"/g" bower.json
+# Build xterm.js into `dist`
```
Tyriar: I like the build/dist distinction :+1: that solves the issue I was having.

test/viewport-test.js
parisk: Tests were breaking 😕 : https://travis-ci.org/sourcelair/xterm.js/builds/155375888.

Any clue?

src/xterm.js
parisk: I would prefer if I could avoid this, but the way that browserify is building UMD modules it will break the API anyway for plain browsers.

The thing is that I could use the default export in general, but browserify would just create a UMD module named `xterm` with the constructor placed in `xterm.default` :trollface:.

test/viewport-test.js
Tyriar: I guess Babel adds `'use strict'`

bin/build
parisk: I will rewrite this in Node.

parisk: Right now.

src/xterm.js
131: .. just juse the `--standalone Terminal` flag for browserify, it will expose term.js main export as it is now...

package.json
```diff
@@ -28,7 +28,7 @@
     "sleep": "^3.0.1"
   },
   "scripts": {
-    "start": "./bin/demo-server",
+    "start": "node_modules/.bin/nodemon --watch src --watch addons --exec bash -c './bin/build && node demo/app'",
```
Tyriar: This can be just nodemon, no need for the `node_modules/.bin` (same with jsdoc).

test/viewport-test.js
parisk: From what I read this seems to be job of the es2015 preset.

parisk:
🆗  folks, this should be considered ready. Feel free to take a look, in order to merge this and move forward with further fixes.

Tyriar:
Ship it :+1:, I'll make sure it works in vscode in a few days to a week. We're going through our testing/release phase now so this will probably hit the Insiders build in a week or so.

parisk:
Awesome. I will move on with creating some issues to further improve the codebase after this.

PR #252, opened 2016-08-26
add modifier+delete, modifier+F1-F12 keys

ayapi:
<kbd>ctrl+delete</kbd>, <kbd>shift+delete</kbd>, <kbd>shift+f1</kbd>... etc.
i thought it would be nice to use these keys. how about like this?

reference -> http://invisible-island.net/xterm/xterm-function-keys.html

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -3016,18 +3016,90 @@
           }
           break;
         // F1-F12
-        case 112: result.key = '\x1bOP'; break;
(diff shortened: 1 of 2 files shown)


parisk:
Will get back to this after we move on with a build system for the repository 😊 .

parisk:
Works great, thanks 😄 .

ayapi:
thank you!

PR #257, opened 2016-08-29
Improve copyright line

Tyriar:
This has bugged me for a while :stuck_out_tongue:


parisk:
Oh my god, that was a leftover from a long time ago when we used to have that banding.

Thanks! 😄

PR #269, opened 2016-09-15
Set package.json main to build/xterm.js

Tyriar:
build/xterm.js should be used when require is used when installed via npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "test",
     ".gitignore"
   ],
-  "main": "src/xterm.js",
+  "main": "build/xterm.js",
   "repository": "https://github.com/sourcelair/xterm.js",
   "license": "MIT",
   "devDependencies": {


parisk:
Could you please run `npm build` into the `postinstall` script? Otherwise it might load a wrong version of the library.

Tyriar:
@parisk does postinstall run when it's installed via npm? If so babel and the like would need to be `dependencies`, not `devDependencies`. It would also mean it rebuilds what we push as a release in ./build on install.

Tyriar:
We want to check in `./build/` when we publish right?

parisk:
Yes, I think eventually the `postinstall` thing is not a good idea.

Also, yes we will have to check in `./build/` when we publish, in order to make this work out of the box for everyone, plain browser to amd, commonjs and node.

parisk:
Please give me a little bit more time here to think about any other possible solutions.

Tyriar:
@parisk pretty sure we need this if we want people who use npm to use the version in `build/` (the most compatible version).

parisk:
This was reverted in https://github.com/sourcelair/xterm.js/commit/68906b5def33e138013354b0577b069bcbbe0018#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R9 as people need to fetch from `dist`. `build` is [ignored by Git](https://github.com/sourcelair/xterm.js/blob/master/.gitignore#L3) and is there for development purposes.

PR #272, opened 2016-09-16
Add setOption API

### LINKED ISSUE #271: Come up with API for changing settings after initialization
Currently there are issues (https://github.com/Microsoft/vscode/issues/11049) when changing settings during runtime since the terminal gets "reset" when application mode is entered and exited. Reset uses `term.options[settingKey]` whereas the actual setting used is `term[key]`. Changing a setting requires both of these to be set.

Current:

```js
xterm.options.cursorBlink = blink;
xterm.cursorBlink = blink;
```

Proposed:

```js
xterm.setOption('cursorBlink', true);
```

Requesting feedback from @parisk & @akalipetis 

Related: https://github.com/sourcelair/xterm.js/commit/c2aa0d75ec210d50348051492365241f39a884cd

## Details
- OS version: Ubuntu 16.04
- xterm.js version: ecfa44b0e8e5ebd410b330ce2b582a3adee67cc2

Tyriar:
Fixes #271

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -358,6 +358,17 @@ Terminal.prototype.focus = function() {
 };
 
 /**
+ * Sets an option on the terminal.
+ */
+Terminal.prototype.setOption = function(key, value) {
+  if (!(key in Terminal.defaults)) {
+    throw new Error('No option with key "' + key + '"');
+  }
+  this[key] = value;
+  this.options[key] = value;
+};
+
+/**
  * Binds the desired focus behavior on a given terminal object.
  *
  * @static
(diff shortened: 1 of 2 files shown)


src/xterm.js
```diff
@@ -358,6 +358,17 @@ Terminal.prototype.focus = function() {
 };
 
 /**
+ * Sets an option on the terminal.
+ */
```
parisk: Could you please use the [`@param`](http://usejsdoc.org/tags-param.html) JSDoc syntax here to document the arguments?

Tyriar: Opps. That's what CR is for!

parisk: Yep, I also wanted to try the new review features more, so this was a good chance 😅 .

PR #276, opened 2016-09-17
Expose API to clear the terminal

### LINKED ISSUE #173: Expose clear function
To allow external code to clear the terminal, for example in order to implement a button or ⌘K keybinding to clear the terminal (#118), a clear function should be exposed. This is not as simple as calling `eraseInDisplay([2])`

This clear function should:

- Clear the display above the line the cursor is on
- Scroll the viewport such that the cursor line is at the top

Since it's difficult/impossible to determine whether your prompt spans multiple lines (like my own), I suggest we do what iTerm does on OS X which is to clear all lines except the one the cursor is on.

Related: https://github.com/Microsoft/vscode/issues/6603

Tyriar:
Fixes #173

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -3058,6 +3058,10 @@ Terminal.prototype.eraseLeft = function(x, y) {
  * Clears the entire buffer, making the prompt line the new first line.
  */
 Terminal.prototype.clear = function() {
+  if (this.ybase === 0 && this.y === 0) {
+    // Don't clear if it's already clear
+    return;
+  }
   this.ydisp = 0;
   this.ybase = 0;
   this.y = 0;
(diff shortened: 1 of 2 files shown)


parisk:
Code LGTM 👍 .

While testing it I found out that if you run `term.clear()` when only the prompt is visible, the prompt disappears, while the cursor stays at the same place (I attached a screenshot to see how it looks like).

Is this the intended behavior here?

## Screenshot
image

Tyriar:
As in `term.clear(); term.clear();`? Looks like a bug

Tyriar:
Actually there is still an issue, your screenshot also occurs when the buffer is <= `term.rows`, looking into it now.

Tyriar:
Shuffled things around a bit to fix that problem and added a test to check 2x clears. Merging

PR #283, opened 2016-09-22
Workaround a known bug in Babel JS

iMoses:
Details can be found here: https://github.com/babel/babel/issues/3826

It causes tests to fail after the first run, due to babel cache which save unicode string wrong.
The fix was simply to replace the string representation of the character with a numeric one, then convert it into a string.

I also made a minor change to the .gitignore file to ignore .idea, the settings directory Injellij IDE.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ Makefile.gyp
 example/*.log
 docs/
 npm-debug.log
+/.idea/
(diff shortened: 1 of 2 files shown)


parisk:
Thanks! Will keep an eye on the original issue to revert this in the future, as the syntax is much better with literals.

PR #285, opened 2016-09-24
separate the interface logic from the core logic into different files

iMoses:
This was done as the initial step of an attempt to create a react component for the xterm.js core.
I believe it would make refactoring much easier, once the layers are completely separated, and this way the core can be required separately and new interfaces would be easier to implement. 
The between the two communication is done using events.

I hardly changed any code, I only separate the methods into different files. Core containing everything that doesn't have to do with the browser, and Interface takes care of the rest. Interface inherits from Core and it is what xterm.js exports, so it shouldn't affect anyone, unless they're listening to the 'refresh' event which was changed a bit.

Let me know what you think :)


Tyriar:
Things do listen to refresh, what changed exactly?

```js
term.on('refresh', term.compositionHelper.updateCompositionElements.bind(term.compositionHelper));
```

Also "Interface" is a bit ambiguous, Browser.js may be better? @parisk thoughts? @iMoses what were your plans in regards separating the other modules?

iMoses:
I haven't notice any use of the refresh event in the code, so I assume it is being used by an external source. I can change the name of the new event or revert the event arguments and adjust my code changes to work with it.

What has changed is that instead of calling 'this.refresh' directly, the Core now emits an event which can be caught by any interface to be rendered as it see fits. The event changed a bit, instead of returning a single object argument, it now returns two arguments, start and end. The main benefit is that Core can now be initialized as a stand alone without breaking anything and it still announces UI related actions using events.

I agree about Interface being too ambiguous, haven't thought of a better name yet.
Eventually I see the entire "Interface" becomes the View that is used to render the UI, and views can change according to implementation, such as a React Component.

In regard to other modules, I started playing with the concept of separating everything into smaller files/modules, but it was too big of a change and I ain't sure it actually made anything more readable. So what I did, locally, is separate "charset", "colors" and "events" into helper files. I'm also thinking about turning CSI and ESC into external commands which the Core uses.

I'm trying to focus first on the Core, which is a single module, separating everything else away from it. Currently I'm working on a React UI which uses this Core, that way I'm slowly learning what more can be done to support multiple interfaces. Once my experiment will be done, I'll go back and refactor the "Interface" (events, Viewport, CompositionHelper, addons) according to the changes. So for the moment I don't have a complete plan for all modules, I hope to get there slowly :)

parisk:
I think I have to take a deeper look in this PR to see if namings are 👍 .

@iMoses how does the current architecture affect you in creating a React component for xterm.js?

iMoses:
The current architecture is tightly coupled with rendering logic. 
There's no way to import or run the Terminal without it.
It highly complicates the creation of different interfaces, such as a React component that takes care of it's own rendering logic.

I think a better, more complete example would be a better explanation. I'll work on it but it would take me a couple of days.

What I've done, in my own project, is removed everything that is somehow connect to the `document` or `window` objects from Terminal, and replaced some of them with event, such as 'refresh' and 'resize' that passes the same parameters that the original function received. (I realized that some of it overrides current events, in the final code I could change them if necessary)

In my customized interface I listen to  and emit events to communicate with the library.

parisk:
@iMoses, I am moving forward with closing this, since it got quite obsolete and many changes have happened till this got opened.

We will open up a new issue for separating rendering and state logic, which will be helpful hopefully.

PR #293, opened 2016-09-28
Allow right-click paste (again)

### LINKED ISSUE #202: Right-click paste does not work
Right-click paste does not work. This regression was caused after #175, since we moved input handling to a textarea instead of xterm itself.

## Proposals
This could be solved in the following two ways (additional ideas are always welcome though)

1. Temporarily move the textarea element below the mouse cursor on right click (https://github.com/sourcelair/xterm.js/pull/175#issuecomment-233876524). This would need non-native selection handling though (https://github.com/sourcelair/xterm.js/pull/175#issuecomment-234344851).
2. Temporarily give focus and enable contenteditable on the terminal element to allow pasting without messing with text selection issues.

Out of these two proposals IMO **1.** is the best way to go (it will also fix other bugs like #116), but it will need definitely more work, while **2.** is a quick hack that could do the job for now.

## Details
- Browser and browser version: All browsers
- OS version: All OSes
- xterm.js version: >= 10fbf0dcccdb1f8a9f065029c788ca86fb99fce8

### Steps to reproduce

1. Right click on the terminal
2. Notice that there is no **Paste** option in the right click menu, while it should

parisk:
Allow right-click paste in the terminal, with the hidden textarea.

For the purposes of this PR, I branched out all clipboard logic into a new module and documented everything with JSDoc.

This PR closes #202 and renders #232 useless for the moment.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -34,6 +34,7 @@
 import { CompositionHelper } from './CompositionHelper.js';
 import { EventEmitter } from './EventEmitter.js';
 import { Viewport } from './Viewport.js';
+import { rightClickHandler, pasteHandler, copyHandler } from './handlers/Clipboard.js';
 
 /**
  * Terminal Emulation References:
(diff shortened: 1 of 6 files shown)


src/xterm.js
```diff
@@ -1691,15 +1650,15 @@ Terminal.prototype.write = function(data) {
           case '=':
             this.log('Serial port requested application keypad.');
             this.applicationKeypad = true;
-            this.viewport.syncScrollArea();
+            this.viewport.setApplicationMode(true);
```
Tyriar: Accidental diff?

src/xterm.js
```diff
@@ -431,52 +432,20 @@ Terminal.bindBlur = function (term) {
...
-      term.handler(text);
-      term.textarea.value = '';
-      return term.cancel(ev);
-    }
+  // Bind clipboard functionality
+  on(this.element, 'copy', copyHandler);
+  on(this.textarea, 'paste', function (ev) {
+    pasteHandler.call(this, ev, term);
+  });
+  on(this.element, 'contextmenu', function (ev) {
```
Tyriar: I think this might conflict with vscode which overrides the context menu to show options to create new terminal, paste and copy.

src/xterm.js
```diff
@@ -431,52 +432,20 @@ Terminal.bindBlur = function (term) {
...
-  on([term.textarea, term.element], 'paste', function(ev) {
-    ev.stopPropagation();
-    if (ev.clipboardData) {
-      var text = ev.clipboardData.getData('text/plain');
-      term.handler(text);
-      term.textarea.value = '';
-      return term.cancel(ev);
-    }
+  // Bind clipboard functionality
+  on(this.element, 'copy', copyHandler);
```
Tyriar: Should copy be exposed via the API so it can be done programmatically?

src/xterm.js
parisk: Wow. I have no idea how this appeared here.

src/xterm.js
parisk: Are you sure? I am not cancelling the event.

src/xterm.js
parisk: All methods are already exposed in Clipboard.js. This also would not work in browsers, so I am skeptical about exposing such a public API.

src/xterm.js
parisk: Rewrote history, so this is gone now. Thanks for pointing out!

src/xterm.js
Tyriar: @parisk wouldn't it cause 2 context menus to appear or something?

Tyriar:
Let me know when this is good to go and I'll do a quick test with vscode, I suspect there might be some breakages for Electron apps with custom right click menus.

src/xterm.js
parisk: No. All I am doing here is move the textarea **when** the context menu appears. Is there any easy way to try this out?

parisk:
@Tyriar feel free to test

PR #296, opened 2016-10-01
Move to TypeScript build

Tyriar:
A second much slimmer go at moving to TypeScript. Here's a summary:
- JavaScript files are compiled using the `--allowJs` flag, this enables us to convert incrementally to TS
- Typings are now tracked in typings.json, this is strong type information to recognize node for now (we won't need this when we convert fully though)
- Didn't change the API


parisk:
PR seems good in general, but I have a few comments and questions:

1. Tests are failing, but I guess this should only require to build with TypeScript before running the tests
2. Does this work also in browsers without a module system?

Last, I have to point out that while I am kind of skeptical of introducing TypeScript syntax into the core (possible barrier to entry for new contributors and found out tools like [Flow](https://github.com/facebook/flow) that allow static type checking in pure JavaScript) the build system seems **much better** than the previous one which needed lots of different npm packages that do the job.

Tyriar:
1. Tests are actually passing fine locally, need to look into why it's failing on Travis
2. TypeScript builds to es5/commonjs and then that goes through browserify similar to what we were doing before for a UMD wrapped file

Regarding Flow, that would be another way to get static typing but the power of it would only really unlock when you embrace explicit typing and other features of the language (which looks very similar to TypeScript). Regardless, they both look very similar to JavaScript and I'm not so sure worrying about *potentially* losing some contributions is worth avoiding type safe code.

bin/build
```diff
@@ -1,38 +1,7 @@
...
-
-// Copy CSS into $BUILD_DIR
-console.log('  - Building ' + buildDir + '/xterm.css...');
-
-fs.createReadStream('src/xterm.css').pipe(
-  fs.createWriteStream(buildDir + '/xterm.css')
-);
-console.log('    OK.');
+mkdir -p build
+browserify src/xterm.js --standalone Terminal -p [ tsify ] --outfile build/xterm.js
```
Tyriar: TS 2 may have made this step redundant? https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#support-for-umd-module-definitions

parisk: I think that this solves most of problems with TypeScript building.

parisk:
Do we have any idea why tests might be failing on Travis?

blink1073:
It looks like you're running your mocha tests against ES6 code.

Tyriar:
I'll sync with master and fix up the tests soon 😃

bin/build
Tyriar: I looked into this, while it does support outputting as `umd` module types only `amd` and `system` are supported with the `--outFile` argument. They recommend using browserify or webpack for bundling umds.

Tyriar:
@parisk since 2.1.0 is out, is it fine to merge https://github.com/Tyriar/xterm.js/pull/1 into this PR and then into master provided the tests pass?

parisk:
Sure @Tyriar, let's move on.

PR #303, opened 2016-10-05
Add `Terminado` example

staticfloat:
This adds a (not-quite barebones) example of how to use `Xterm.js` with `Terminado`.  Note that `terminado_attach.js` is extremely similar to `attach.js`, perhaps there's some way we can provide arguments to `attach.js` so that the file is completely unnecessary.  I'm very much not a Javascript coder, as you can probably tell by my hacked-up `index.html` that begs, borrows and steals from things like `fit.js` to hopefully cut down on the amount of code needed to get you up and running on Python.

If you think this can be made simpler, (I definitely think it can) then feel free to just close this PR and remake your own based upon this, I just thought this might be useful for those that want to use a Python backend instead of a node.js one.


parisk:
In general the content of this PR is quite nice and interesting, but I believe it is out of scope in specific parts.

While a terminado add-on would be useful, I do not think that a terminado demo should be inside this repository.

## My proposal

- [ ] Make `terminado_attach` an add-on. It could be called just `terminado` also, to make things easier.
- [ ] Remove the demo completely
- [ ] Add a **Connecting/Integrating with Terminado** guide at http://xtermjs.org/docs (repo: https://github.com/xtermjs/xtermjs.org)

I think this is the best idea, in order to preserve the scope and purpose of this repository, while helping people that want to use xterm.js with Terminado.

Tyriar:
:+1: to moving the demo to a guide in http://xtermjs.org/docs/
:-1: to adding this as an addon, that means we need to maintain it until the end of time

scopatz:
> :-1: to adding this as an addon, that means we need to maintain it until the end of time

I don't think that is what that means.  It just means that as long as it is useful, your users can actually find it. Feel free to deprecate it in the future if the state of the art moves on.

Tyriar:
@scopatz since addons hook into private APIs by design this limits the ability to evolve the code base, since addons are essentially part of the API (or at least will be when they're documented).

scopatz:
I think it is OK to break addons as needed, and then people can submit patches to fix them.  They are explicitly not part of the core. Folks can always roll back to last working version, if needed.  My point is that without an addon, the user is forced to do this work. I am not a hugely experienced web or JS dev, and it was complex enough that after about 8 hours trying to get xtermjs work over the past couple of days, I gave up.  I would really appreciate *anything* that makes the user experience less painful.

Tyriar:
I was looking at the code on my phone and I think I missed some, after another look I know what you mean. It doesn't actually touch any private APIs which is the main thing I was concerned about. So :+1: for the addon.

terminado_demo/terminado_attach.js
```diff
@@ -0,0 +1,142 @@
...
+   * @param {WebSocket} socket - The socket to attach the current terminal.
+   * @param {boolean} bidirectional - Whether the terminal should send data
+   *                                  to the socket as well.
+   * @param {boolean} buffered - Whether the rendering of incoming data
+   *                             should happen instantly or at a maximum
+   *                             frequency of 1 rendering per 10ms.
+   */
+  exports.attach = function (term, socket, bidirectional, buffered) {
+    bidirectional = (typeof bidirectional == 'undefined') ? true : bidirectional;
+    term.socket = socket;
```
Tyriar: @parisk related to my comments on the PR, we should come up with some namespacing convention to prevent variable collisions with other addons. Something like:

```js
// _ prefix for private, followed by addon name
term._terminado = {};
term._terminado.socket = socket;
term._terminado._flushBuffer = function () {
...

// addon name prefix
Xterm.prototype.terminadoAttach = ...
```

staticfloat:
Great, if you think it's better to have it as a separate addon, I can definitely do that.  In the same spirit as the `attach` plugin, should I include a demo `index.html` page?

addons/terminado/terminado.js
```diff
@@ -0,0 +1,142 @@
...
+   * Attaches the current terminal to the given socket
+   *
+   * @param {WebSocket} socket - The socket to attach the current terminal.
+   * @param {boolean} bidirectional - Whether the terminal should send data
+   *                                  to the socket as well.
+   * @param {boolean} buffered - Whether the rendering of incoming data
+   *                             should happen instantly or at a maximum
+   *                             frequency of 1 rendering per 10ms.
+   */
+  Xterm.prototype.attach = function (socket, bidirectional, buffered) {
```
parisk: Could you please rename this to `terminadoAttach` or something similar, in order to avoid clashes with the `attach` addon?

addons/terminado/terminado.js
```diff
@@ -0,0 +1,142 @@
...
+    return exports.attach(this, socket, bidirectional, buffered);
+  };
+
+  /**
+   * Detaches the current terminal from the given socket.
+   *
+   * @param {WebSocket} socket - The socket from which to detach the current
+   *                             terminal.
+   */
+  Xterm.prototype.detach = function (socket) {
```
parisk: Could you please rename this to `terminadoDetach` or something similar, in order to avoid clashes with the `attach` addon?

parisk:
Do you plan to open a PR at https://github.com/xtermjs/xtermjs.org with a more complete documentation about how to connect to a terminado back-end?

addons/terminado/terminado.js
staticfloat: If I rename `Xterm.prototype.attach` to `Xterm.prototype.terminadoAttach`, will I need to change the `attach` names on lines [9](https://github.com/sourcelair/xterm.js/pull/303/files/fe54da34f81c0fc15bb9986cb71bb8a82000f9b8#diff-36ae33bbbcb6d2f58585e7e15063d570R9), [14](https://github.com/sourcelair/xterm.js/pull/303/files/fe54da34f81c0fc15bb9986cb71bb8a82000f9b8#diff-36ae33bbbcb6d2f58585e7e15063d570R14), [19](https://github.com/sourcelair/xterm.js/pull/303/files/fe54da34f81c0fc15bb9986cb71bb8a82000f9b8#diff-36ae33bbbcb6d2f58585e7e15063d570R19) and [24](https://github.com/sourcelair/xterm.js/pull/303/files/fe54da34f81c0fc15bb9986cb71bb8a82000f9b8#diff-36ae33bbbcb6d2f58585e7e15063d570R24)?

staticfloat:
> Do you plan to open a PR at https://github.com/xtermjs/xtermjs.org with a more complete documentation about how to connect to a terminado back-end?

Yes

parisk:
Moving on with some functional tests.

parisk:
@staticfloat do you have any ready example that I can use to test this out?

staticfloat:
Documentation here: https://github.com/xtermjs/xtermjs.org/pull/4

addons/terminado/terminado.js
```diff
@@ -0,0 +1,142 @@
...
+    };
+
+    socket.addEventListener('message', term._getMessage);
+
+    if (bidirectional) {
+      term.on('data', term._sendData);
+    }
+    term.on('resize', term._setSize);
+
+    socket.addEventListener('close', term.detach.bind(term, socket));
```
parisk: Change this to `term.terminadoDetach` please.

addons/terminado/terminado.js
```diff
@@ -0,0 +1,142 @@
...
+    socket.addEventListener('error', term.detach.bind(term, socket));
```
parisk: Change this to `term.terminadoDetach` please.

addons/terminado/terminado.js
```diff
@@ -0,0 +1,142 @@
...
+   *
+   * @param {Xterm} term - The terminal to be attached to the given socket.
+   * @param {WebSocket} socket - The socket to attach the current terminal.
+   * @param {boolean} bidirectional - Whether the terminal should send data
+   *                                  to the socket as well.
+   * @param {boolean} buffered - Whether the rendering of incoming data
+   *                             should happen instantly or at a maximum
+   *                             frequency of 1 rendering per 10ms.
+   */
+  exports.attach = function (term, socket, bidirectional, buffered) {
```
parisk: Maybe renaming exported functions to `terminadoAttach` and `terminadoDetach` as well would be a better option, to reduce confusion.

staticfloat:
Done, and tested, this time with the console open to make sure there are no errors.

takluyver:
:+1: to having something like this to make things easy for people coming from the Python world. I'd like to add a link from the Terminado docs once this and the associated doc PR are done.

takluyver:
For that matter, if it makes life easier, we could add an option to terminado to send/receive data in the format supported by the existing `attach` addon. Looking over that code, though, I don't think it can tell the server when the terminal has been resized, so I still like the option to have a slightly richer protocol.

parisk:
Everything seems to be working great. Thanks for implementing this!

@takluyver let's keep the scope of this PR to where it is right now and add extra functionality down the road in an as-needed base.

takluyver:
Thanks!

PRs welcome to add a link to terminado docs, or I'll get round to it sooner or later.

parisk:
@staticfloat just letting you know that we mentioned you for this contribution in the [**2.1** announcement](http://xtermjs.org/blog/articles/2016-11-10-xtermjs-2.1). Thank you!

PR #304, opened 2016-10-05
Make add-ons load Terminal from dist

### LINKED ISSUE #305: Addon loader doesn't work in Electron
Since the 2.0 release the loadAddon method doesn't work anymore in Electron apps.

Code:
```
var Terminal = require('xterm');
Terminal.loadAddon('fit');
```
Error:
`Uncaught Error: Cannot find module '/src/../addons/fit'`

The same code does work with version 1.1.3.

## Details
- Browser and browser version: Electron 1.4.3 (Chrome 53.0.2785.113)
- OS version: MacOS Sierra
- xterm.js version: 2.0.0 / 2.0.1

parisk:
Make add-ons load Terminal from `dist`, or else they will break on environments with module systems that do not support ES2015 modules.

I also tweaked the tests to made sure that the `dist` constructor gets overloaded, even if  `loadAddon` is being called from `src/xterm.js`. I am not quite happy with this hack though 😕  .

Fix #305.

### DIFF
--- a/addons/attach/attach.js
+++ b/addons/attach/attach.js
@@ -11,12 +11,12 @@
     /*
      * CommonJS environment
      */
-    module.exports = attach(require('../../src/xterm'));
+    module.exports = attach(require('../../dist/xterm'));
   } else if (typeof define == 'function') {
     /*
      * Require.js is available
      */
-    define(['../../src/xterm'], attach);
+    define(['../../dist/xterm'], attach);
   } else {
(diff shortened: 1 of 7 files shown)


Tyriar:
Some addon tests are failing

parisk:
Working on this right now.

parisk:
Test are passing @Tyriar, can you please take another look at this?

Issue #216, opened 2016-08-03
Viewport should only scroll down on input, not when the pty writes


Tyriar:
Similar to `gnome-terminal`, if the user is scrolling up to look at past output and the currently running program adds output to the terminal, the viewport should not scroll and interrupt what they are looking at.
## Details
- Browser and browser version: Chrome 52
- OS version: Ubuntu 16.04
- xterm.js version: d2cef6f29db8e4e47b82000f8415b849a7756728
### Steps to reproduce
1. Run `while [ "true" ]; do echo "ping"; sleep 1; done`
2. Scroll up, notice the viewport is moved to the bottom every second

jerch:
This is maybe another candidate for a customisation setting as it is in xterm.

PR #326, opened 2016-10-26
prevent terminal scrolling when scrollback

jupe:
Quick fix to prevent terminal scrolling when user is looking into scrollback.

**Initial state:**
there is hundreds of line in xterm, and just some of lines are visible -> user can scroll view..
More lines is coming to xterm all the time..

**Behaviors before:**
If user scroll up it goes back to end of buffer when next line is appearing which are quite annoying.

**Behaviors After:**
User can scroll up without annoying "auto-scroll-down" effect.


src/xterm.js
```diff
@@ -1339,11 +1352,11 @@ Terminal.prototype.write = function(data) {
   this.refreshStart = this.y;
   this.refreshEnd = this.y;
 
-  if (this.ybase !== this.ydisp) {
-    this.ydisp = this.ybase;
-    this.emit('scroll', this.ydisp);
-    this.maxRange();
-  }
+  // if (this.ybase !== this.ydisp) {
```
parisk: Remove this peace of code completely, if it's not needed any more.

parisk:
Thanks for your contribution! This PR seems 🆒 .

I 'd like to ask a couple of things:

1. fix my code comment above
2. share some instructions on how to test this PR from the user's perspective to compare "before" and "after"

jupe:
added short introduction to PR description :).

parisk:
Thanks @jupe! Seems like this PR is fixing #216. Right?

Also please take a look at https://github.com/sourcelair/xterm.js/pull/326#discussion_r85089031 to remove redundant inline comments 😊 .

src/xterm.js
```diff
@@ -231,6 +231,9 @@ function Terminal(options) {
 
   this.tabs;
   this.setupStops();
+
+  // Store if user went browsing history in scrollback
+  this.userScrolling = false;
```
Tyriar: I'm not sure there needs to be additional object state for this. Basically we only want to scroll down if the terminal's viewport is already at the bottom of the buffer, so this could be pulled into a function and computed on the fly instead. That would remove the risk of the property becoming stale.

src/xterm.js
Tyriar: The main case this covered was to scroll down whenever there are new key strokes, how is this handled now? We want to not scroll down when the process writes something, but we do want to scroll down if the write origin comes from a `key*` event.

dist/xterm.css
```diff
@@ -61,6 +61,7 @@
     position: absolute;
     opacity: 0;
     left: -9999em;
+    top: -9999em;
```
Tyriar: Also no need to update the dist files, ideally this is only done on release to reduce the amount of noise in the commit log. Don't revert it as that would double the noise, just FYI for the future :smiley:

src/xterm.js
jerch: @Tyriar To handle those cases I suggest to distingish two separate boolean states `scrollOnInput` and `scrollOnOutput` as gnome-terminal does.

xterm itself has also 2 settings (`XTerm*ScrollTTYOutput` and `XTerm*ScrollTTYKeypress`) but behaves different - scrolling itself can't be stopped and the 2 settings just toggle the jump to bottom on data.

Since all other popular emulators (konsole, gnome-terminal, rxvt) do a full stop of scrolling (until the scroll buffer runs out) I suggest to go with that behaviour, too.

parisk:
@jupe did you have the chance to take a look at the above comments?

I think it would be great if we could merge this before [2.1](https://github.com/sourcelair/xterm.js/milestone/5)!

jupe:
Unfortunately I doesn't have time for this at the moment..

parisk:
That's 🆗  @jupe.

As I would like to include this in the next release which will be on Monday 7 November, are you OK with me taking over or do you believe that you can make it by the end of the week?

jupe:
it's ok if you take this over.

parisk:
Closing this in favor of #336.

PR #336, opened 2016-11-03
Stop auto scrolling to bottom on output

### LINKED ISSUE #216: Viewport should only scroll down on input, not when the pty writes

parisk:
This PR is a continuation of #326 by @jupe. I kept all original commits, to maintain the appropriate credits 😊 .

This PR introduces this new behavior:

- When the user has scrolled up the terminal and output is being printed (by the back-end), the terminal now is not scrolling down to bottom, as it used to
- The terminal scrolls down to bottom when the user presses a key

Closes #216.

/cc @Tyriar to take a look since you opened the initial issue and commented also in the original PR.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -231,6 +231,9 @@ function Terminal(options) {
 
   this.tabs;
   this.setupStops();
+
+  // Store if user went browsing history in scrollback
+  this.userScrolling = false;
 }
 
 inherits(Terminal, EventEmitter);
@@ -1250,7 +1253,9 @@ Terminal.prototype.scroll = function() {
     this.lines = this.lines.slice(-(this.ybase + this.rows) + 1);
   }
 
-  this.ydisp = this.ybase;
+  if (!this.userScrolling) {
+    this.ydisp = this.ybase;
+  }
 
   // last line
   row = this.ybase + this.rows - 1;
@@ -1272,7 +1277,9 @@ Terminal.prototype.scroll = function() {
   if (this.scrollTop !== 0) {
     if (this.ybase !== 0) {
       this.ybase--;
-      this.ydisp = this.ybase;
+      if (!this.userScrolling) {
+        this.ydisp = this.ybase;
+      }
     }
     this.lines.splice(this.ybase + this.scrollTop, 1);
   }
@@ -1292,6 +1299,12 @@ Terminal.prototype.scroll = function() {
  * viewport originally.
  */
 Terminal.prototype.scrollDisp = function(disp, suppressScrollEvent) {
+  if (disp < 0) {
+    this.userScrolling = true;
+  } else if (disp + this.ydisp >= this.ybase) {
+    this.userScrolling = false;
+  }
+
   this.ydisp += disp;
 
   if (this.ydisp > this.ybase) {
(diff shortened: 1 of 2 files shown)


src/xterm.js
```diff
@@ -2413,6 +2420,13 @@ Terminal.prototype.attachCustomKeydownHandler = function(customKeydownHandler) {
  * @param {KeyboardEvent} ev The keydown event to be handled.
  */
 Terminal.prototype.keyDown = function(ev) {
+  // Scroll down to prompt, whenever the user presses a key.
+  if (this.ybase !== this.ydisp) {
```
Tyriar: Should `this.userScrolling` be set here?

Also I think having this at the start of `keyDown` means that the DOM could be updated twice in a row, once before the key is evaluated and once after.

parisk: > Should `this.userScrolling` be set here?

I am not sure if it should. Strictly speaking, the user is not scrolling here. We are scrolling to bottom for them, in order to be able to see their input.

> Also I think having this at the start of `keyDown` means that the DOM could be updated twice in a row, once before the key is evaluated and once after.

That's true. I put it there because I thought that we should scroll down no matter how the key will be handled (sounds like a universal terminal behavior).

I think it's not bad that the DOM might get updated twice, since this will happen **only** if the  terminal needs to scroll down, meaning that in the next typing there will be no scrolling of the viewport, thus no second (or first 😄 ) DOM update.

Tyriar:
Would it be easy to write some unit tests for this? For example:

- Key down always scrolls to bottom of buffer
- Write scrolls down to bottom of buffer when it's already at the bottom
- Write does not scroll down to bottom of buffer when not at the bottom

parisk:
🆗  I moved on with the following:

1. Used [`scrollToBottom`](https://github.com/sourcelair/xterm.js/pull/308) instead of manipulating `ydisp` and `ybase` myself and emitting `scroll`
2. Implemented test to make sure that terminal scrolls to bottom, when scrolled up and key is pressed
3. Cleaned up and squashed commits in order to avoid touching the `dist` directory at all and keeping the initial commit of @jupe.

@Tyriar I don't think I could easily implement tests for [`write`](http://xtermjs.org/docs/api/Terminal/#writetext) as it interacts with the DOM heavily.

Let me know if this is good to go.

test/test.js
```diff
@@ -181,6 +181,25 @@ describe('xterm.js', function() {
         assert.equal(xterm.ydisp, startYDisp);
       });
     });
+
+    describe('keyDown', function () {
+      it('should scroll down, when a key is pressed and terminal is scrolled up', function () {
+        var terminal = new Terminal();
+
+        // Do not process the keyDown event, to avoid side-effects
+        terminal.attachCustomKeydownHandler(function () {
```
Tyriar: Clever trick :smile:

parisk: Thanks!

parisk:
@onttoni just letting you know that we mentioned you for your commit in this PR in the [**2.1** announcement](http://xtermjs.org/blog/articles/2016-11-10-xtermjs-2.1). Thank you!

PR #337, opened 2016-11-03
Add main property to bower.json

IanLewis:
This patch adds the main property to the project's bower.json for better
interoperability with tools like wiredep etc.


parisk:
@IanLewis just letting you know that we mentioned you for this contribution, in the [**2.1** announcement](http://xtermjs.org/blog/articles/2016-11-10-xtermjs-2.1). Thank you!

PR #339, opened 2016-11-04
Fixed copy event in IE11

hiro-su:
hello!,
an error occurs by the copy event in IE11.
problems similar to https://github.com/sourcelair/xterm.js/pull/334
this pr will fix it, so please check.

### DIFF
--- a/src/handlers/Clipboard.js
+++ b/src/handlers/Clipboard.js
@@ -35,11 +35,16 @@ function prepareTextForClipboard(text) {
  * Binds copy functionality to the given terminal.
  * @param {ClipboardEvent} ev The original copy event to be handled
  */
-function copyHandler (ev) {
+function copyHandler(ev, term) {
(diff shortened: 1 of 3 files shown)


src/handlers/Clipboard.js
```diff
@@ -134,6 +140,15 @@ function rightClickHandler(ev, term) {
   }
 }
 
+// Browser detection using the user agent
+function detectBrowser(userAgent) {
```
parisk: I am not sure if here is the right place for this as it is not a clipboard specific utility.

Let's leave this here though until we find a better place for it, but feel free to drop any other ideas.

parisk:
@hiro-su can you switch to `term.browser` for browser and platform detecting as introduced in #342?

hiro-su:
@parisk Thank you for your suggestions!
I corrected the parts you pointed out. Please check it again.

PR #341, opened 2016-11-04
Copy(Ctrl + c) failed in the Windows environment except IE11

hiro-su:
Hi,
Could not be copied(<kbd>ctrl</kbd> + <kbd>c</kbd>) due to the SIGINT is sent to the process.

It was modified so that it does not send the SIGINT In the range selection.
Could you please check this pr.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -2693,7 +2693,10 @@ Terminal.prototype.evaluateKeyEscapeSequence = function(ev) {
       break;
     default:
       // a-z and space
-      if (ev.ctrlKey && !ev.shiftKey && !ev.altKey && !ev.metaKey) {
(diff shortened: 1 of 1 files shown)


parisk:
Thanks for putting time on this, but this functionality is not part of xterm and therefore out of xterm.js' scope.

You can use the [`attachCustomKeydownHandler`](http://xtermjs.org/docs/api/Terminal/#attachcustomkeydownhandlercustomkeydownhandler) method though, in order to highjack default functionality on such cases.

hiro-su:
@parisk Thank you for advice!! I will use the [attachCustomKeydownHandler](http://xtermjs.org/docs/api/Terminal/#attachcustomkeydownhandlercustomkeydownhandler) method.

PR #342, opened 2016-11-04
Make right-click work on all browsers

### LINKED ISSUE #310: Right-click paste does not work in Firefox and IE
Since 2.0 version, Right-click paste work in Chrome, but does not work in Firefox and IE.

Details

Browser and browser version: Firefox 49.0.1.6109 / IE 11.0.9600
OS version: Windown 7
xterm.js version: 2.0.0 / 2.0.1

parisk:
Made right-click work on all browsers.

The issue with the previous implementation was that it was bound to the [`contextmenu`](https://developer.mozilla.org/en/docs/Web/Events/contextmenu) DOM event.

While this works for Chrome and Safari (and Edge?), Firefox and Internet Explorer fire the `contextmenu` event **after** the menu appears (thus highjacking the textarea position is pointless). So, I moved on and detected right-click using the [`mousedown`](https://developer.mozilla.org/en/docs/Web/Events/mousedown) DOM event on Firefox and MSIE.

In addition, I moved all browser detecting utilities into a new `utils/Browser.js` module.

I would like to add some tests for this module but I couldn't find a way to make it work with Mocha, since `navigator` is absent.

Closes #310

### DIFF
--- a/src/handlers/Clipboard.js
+++ b/src/handlers/Clipboard.js
@@ -86,7 +86,7 @@ function pasteHandler(ev, term) {
  */
 function rightClickHandler(ev, term) {
   var s = document.getSelection(),
-      sText = prepareTextForClipboard(s.toString()),
+      selectedText = prepareTextForClipboard(s.toString()),
       clickIsOnSelection = false;
 
   if (s.rangeCount) {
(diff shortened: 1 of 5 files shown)


src/utils/Generic.js
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2016, SourceLair Private Company <www.sourcelair.com> (MIT License)
+ */
+
+/**
+ * Generic utilities module. This module contains generic methods that can be helpful at
+ * different parts of the code base.
+ * @module xterm/utils/Generic
+ */
+
+export let contains = function(el, arr) {
```
Tyriar: `arr, el` would be more intuitive ordering imo "does *arr* contain *el*"

src/utils/Generic.js
```diff
@@ -0,0 +1,19 @@
...
+  for (var i = 0; i < arr.length; i += 1) {
```
Tyriar:
```js
return arr.indexOf(el) >= 0;
```

src/xterm.js
```diff
@@ -446,9 +447,21 @@ Terminal.prototype.initGlobal = function() {
...
   });
-  on(this.element, 'contextmenu', function (ev) {
+
+
+  function rightClickHandlerWrapper (ev) {
     rightClickHandler.call(this, ev, term);
-  });
+  }
+
+  if (isFirefox || isMSIE) {
```
Tyriar: Is it a concern you're forking the logic here? If both work equally well with `mousedown` maybe that would be better so all browsers use the same code path?

src/handlers/Clipboard.js
```diff
@@ -102,23 +102,27 @@ function rightClickHandler(ev, term) {
...
   }
 
   // Bring textarea at the cursor position
   if (!clickIsOnSelection) {
     term.textarea.style.position = 'fixed';
-    term.textarea.style.width = '10px';
-    term.textarea.style.height = '10px';
-    term.textarea.style.left = x + 'px';
-    term.textarea.style.top = y + 'px';
+    term.textarea.style.width = '20px';
```
Tyriar: Curious why this is bigger all of a sudden?

parisk: Just gave it a little bit of padding, because I think it could run into corner cases.

src/utils/Generic.js
parisk: 👍  changing

src/utils/Generic.js
parisk: 👍  changing

parisk:
@Tyriar do you have any clue why this particular test might be failing?

Tyriar:
@parisk probably related to `platform.isMac`. Try  adding a few `console.log`s to the test should find the problem fairly easily.

parisk:
Thanks @Tyriar, the error message wasn't that helpful 😅 . I was setting `xterm.isMSWindows` instead of the :new: `xterm.browser.isMSWindows`.

parisk:
@Tyriar tests are passing, feel free to take another look, when it's 🆗  for you.

PR #349, opened 2016-11-10
Fix selection lost on keyup

nicot:
Fixes https://github.com/sourcelair/xterm.js/issues/348.

Currently, releasing a key destroys the selection, even a modifier key like shift or command. This leads to bugs like https://github.com/Microsoft/vscode/issues/15200. This does not change the behavior that inserting a character will remove the selection.

This is the same behavior that other terminals have.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -487,8 +487,6 @@ Terminal.bindKeys = function(term) {
     term.keyPress(ev);
   }, true);
 
-  on(term.element, 'keyup', term.focus.bind(term));
-
   on(term.textarea, 'keydown', function(ev) {
     term.keyDown(ev);
   }, true);


Tyriar:
@parisk do you know why this was added originally? Perhaps about redirecting the focus from the element to the textarea or something?

parisk:
@Tyriar yes, mostly because of that.

parisk:
Thanks for your contribution @nicot! One thing that does not work in this PR, but used to work and works on my local terminal as well is:

1. Select some text
2. Copy
3. Press <kbd>Ctrl</kbd> + <kbd>V</kbd> (or <kbd>Cmd</kbd> + <kbd>V</kbd> on your Mac)

I expected it to paste, but it didn't. Can you please take care of that before moving on?

nicot:
@parisk: Whoah, that is definitely not good, I'm glad you caught that! Weirdly enough, I can't get that to happen on my computer running macOS in Safari, Firefox or Chrome.

What browser/OS are you running?

parisk:
NP @nicot! I am on macOS Sierra Chrome.

nicot:
I run the same system, and I can't reproduce the issue you're running into. Pasting is working fine for me.

I think pasting happens on keydown, not on keyup, so this shouldn't change anything there.

parisk:
Could you please try in another browser as well or try disabling your add-ons?

I am still running into this issue and after running the following code in the console the issue goes away (along with the functionality of this PR 😅 ):

```javascript
term.element.addEventListener('keyup', term.focus.bind(term));
```

nicot:
I can't reproduce this at all! I've tried three different browsers, in Incognito mode and with extensions disabled. I'll close this PR, as I can't debug this issue further without reproducing it.

PR #353, opened 2016-11-12
Point to correct assets after TS build

Tyriar:
The demo was still pointing at `./build`, not `./dist` so the assets were 404ing

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -2,14 +2,14 @@
 <html>
     <head>
         <title>xterm.js demo</title>
-        <link rel="stylesheet" href="../build/xterm.css" />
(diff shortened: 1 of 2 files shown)


demo/index.html
```diff
@@ -2,14 +2,14 @@
 <html>
     <head>
         <title>xterm.js demo</title>
-        <link rel="stylesheet" href="../build/xterm.css" />
-        <link rel="stylesheet" href="../build/addons/fullscreen/fullscreen.css" />
+        <link rel="stylesheet" href="../dist/xterm.css" />
```
parisk: Should we use absolute URLs here as well?

PR #367, opened 2016-11-21
Prevent keyboard modifier events from scrolling down

### LINKED ISSUE #363: Modifier keys by themselves should not be counted as key presses to scroll to bottom
## Details
- Browser and browser version: vscode 1.8.0-insiders
- OS version: Ubuntu 16.04
- xterm.js version: 2.1.0

### Steps to reproduce

1. Run a command to give you a scrollbar
2. Scroll up
3. Press shift, xterm.js scrolls down to the bottom

I don't believe the shift is routed to the terminal's shell process so it shouldn't scroll down as it isn't technically input.

Tyriar:
Fixes #363 

---

Notes:

- This is the first TypeScript file
- I'm suggesting a new format for .ts test files so they lay right next to the source file they're testing, this prevents duplicating the directory structure under the `test/` dir
- The new `@types` dependencies provide type information for `mocha` and `chai`, allowing TypeScript to check the types when compiling. From TS 2 this is as easy as running `npm i -D @types/mocha`

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,8 @@
   "repository": "https://github.com/sourcelair/xterm.js",
   "license": "MIT",
   "devDependencies": {
+    "@types/chai": "^3.4.34",
+    "@types/mocha": "^2.2.33",
     "@types/node": "^6.0.41",
     "browserify": "^13.1.0",
     "chai": "3.5.0",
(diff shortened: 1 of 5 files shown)


parisk:
I am totally 👍  for the new test file format. Sounds much more sane.

Issue #335, opened 2016-11-02
Migrate code base to TypeScript


parisk:
One of the things proposed in order to increase the quality of this code base as it grows (both in code and popularity) was migrate from JavaScript to TypeScript.

TypeScript sports several cool features like type safety, classes etc. and is a superset of JavaScript, so not much will change.

## Questions
There are some questions though that have to be answered before making such a "big" move:

- What problems that we have already ran into will be solved by TypeScript's features (e.g. type safety, classes etc.)?
- **Timing**: Does it need to be implemented now (why?) or can wait until we really need this?
- What are possible regressions that might be introduced by this?
- What should we change in our current toolset?

## Concerns
My main concerns about moving code base to TypeScript are:

1. Not sure if this will be a barrier to entry for new contributors
2. Not all maintainers are familiar with TypeScripts (so most of us will need to learn a new language and adapt to its ecosystem)
3. Might be an overkill for this project

/cc @Tyriar since IMO you can contribute better into this.

blink1073:
Hi @parisk, a few comments from the peanut gallery based on experience in https://github.com/jupyterlab/jupyterlab.

1) We have had contributors who were new to JavaScript, as well as ones who had used JavaScript heavily before.  The use of TypeScript made it much easier as maintainers to have confidence in PRs from new contributors.  Having types also forces a minimal set of documentation about arguments and return types, that technically could be in JSDoc but is prone to forgetfulness or getting out of sync.  Most of the reservations we heard from the wider community were difficulty in getting a new project in TypeScript set up, which is a lot easier with TypeScript 2.0.  Other complaints included how to get typings for dependencies, which is again mitigated in TypeScript 2.0.

2) You can use as much or as little of the language as you like, depending on your use of compiler settings, and can gradually opt in to things like null checking.

3) Once you have the build system set up, it overall is a decrease in cognitive burden for both maintainers and new contributors, and is a great aid when refactoring.

Tyriar:
**What problems that we have already ran into will be solved by TypeScript's features (e.g. type safety, classes etc.)?**

- The biggest problem for me that I think this solves is to help with refactoring by making the code less fragile. There's quite a large portion of xterm.js which is very difficult to touch and essentially all PRs need to be tested manually to ensure quality. I'm having difficulty starting with https://github.com/sourcelair/xterm.js/issues/266 for example which is essentially a bit refactor.
[...]

PR #371, opened 2016-11-22
Convert CompositionHelper to TypeScript

Tyriar:
Part of #335


src/CompositionHelper.ts
```diff
@@ -0,0 +1,233 @@
+/**
+ * xterm.js: xterm, in the browser
+ * Copyright (c) 2014-2016, SourceLair Private Company (www.sourcelair.com (MIT License)
+ */
+
+import { ITerminal } from './Interfaces';
+
+interface IPosition {
```
Tyriar: The interface is more light-weight than a class and since the interface gives us all we need (type checking) that's sufficient. I would make it a class if we has any operations that worked on `start` or `end`.

Issue #356, opened 2016-11-14
WARNING in ./~/xterm/dist/xterm.js Critical dependencies: 1:481-488 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.  @ ./~/xterm/dist/xterm.js 1:481-488


veeramarni:
Receiving warning after upgrade to latest xterm version. 

I used NPM to install xterm and webpack to build the client app.

Tyriar:
This is expected since we compile and commit the assets to `dist/`. We do this so we can distribute with `bower` and also with `npm` without needing to install babel. It's probably best to just ignore the warning as then you may not be able to upgrade if we change our build system (like we're doing in v2.2.0).

parisk:
@veeramarni as @Tyriar said the best thing you can do is ignore this warning.

Another alternative would be requiring xterm.js as a relative dependency in your webpack bundle and use the `node_modules/xterm/src/xterm.js` path.

I don't think that this will work out of the box though, since we migrated to building with TypeScript.

Closing this as there are not much that we can do to suppress this warning from our side.

blink1073:
In our projects, we have a `src/` folder that has ES6 TypeScript.  We output to a `lib/` directory that is ES5, and our `main` in `package.json` points to the index in that directory.  We ship the `lib/` directory on npm, but keep only `src/` in version control.  We also ship a separate `dist/` directory that could be used for bower.   The `package.json` should point to *node* code, and `bower.json` to the `dist` code.  

cf https://github.com/jupyterlab/services

blink1073:
In this way, consumers require `xterm` in a way that works best for their use case, without needing extra config or importing from `node_modules`.

Tyriar:
@blink1073 dist/ will contain un-minified ES5, from 2.2.0 we now build with TypeScript which you obviously can't use in your build process. Unless I'm misunderstanding, I believe pointing main at dist/xterm.js is ideal for your use case.

blink1073:
Great, all I was looking for was for `main` in `package.json` to point to ES5 commonjs code, and pointing out how we do it ourselves.

veeramarni:
As a workaround, I'm using noParse to prevent the warnings. 

  `"noParse": ["node_modules/localforage/dist/localforage.js", "node_modules/xterm/dist/xterm.js"]`

parisk:
This should not be the case starting with 2.3.0, due to #398.

PR #390, opened 2016-12-05
Bring back the `build` directory for temporary builds

### LINKED ISSUE #375: Proposals for fixing the 'dist' issue: publish on a release branch
We could only commit `dist/` on a release branch and publish that instead of master, this is probably the best way to avoid dist diffs showing up in PRs constantly.

For a release we could branch off to say `release/2.2` (or just `release`), when a release is done it could then be tagged (for bower).

parisk:
Fix #375

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -3,7 +3,7 @@
 set -e
 
 # $BUILD_DIR should default to "build"
-export BUILD_DIR=${BUILD_DIR:=build}
+BUILD_DIR=${BUILD_DIR:=build}
 
 # Create the build directory
 mkdir -p $BUILD_DIR
--- a/bin/build
+++ b/bin/build
@@ -1,26 +1,31 @@
 #! /usr/bin/env bash
 
+set -e
+
+# $BUILD_DIR should default to "build"
+export BUILD_DIR=${BUILD_DIR:=build}
+
 # Clean out/* to prevent confusion if files were deleted in src/
 rm -rf out/*
 
 # Build all TypeScript files (including tests) to out/
 tsc
 
-# Concat all xterm.js files into a single file and output as a UMD to dist/xterm.js
-browserify ./out/xterm.js --standalone Terminal --debug --outfile ./dist/xterm.js
-cat ./dist/xterm.js | exorcist ./dist/xterm.js.map -b ./dist > ./dist/xterm.temp.js
-rm ./dist/xterm.js
-mv ./dist/xterm.temp.js ./dist/xterm.js
+# Concat all xterm.js files into a single file and output as a UMD to $BUILD_DIR/xterm.js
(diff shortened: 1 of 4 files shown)


blink1073:
Can we also point `main` in `package.json` to this folder, to avoid #356?

parisk:
This is not something we are willing to do at the moment.

If we did so, requiring xterm with `require('xterm');` would not work out of the box and would require building on the library from the consumer. This is not an experience we would like to provide ATM.

If you are using TypeScript you can require each of xterm.js' components relatively from the `src` directory.

Tyriar:
@parisk to clarify, release commits will set `BUILD_DIR`, run the script and check in `dist/`?

parisk:
Yep, exactly. They will set `BUILD_DIR` to `dist` and then do the rest.

blink1073:
I disagree.  In our projects, we have a `src/` folder that has ES6 TypeScript.  We output to a `lib/` directory that is ES5, and our `main` in `package.json` points to the index in that directory.  We ship the `lib/` directory on npm, but keep only `src/` in version control.  We also ship a separate `dist/` directory that could be used for bower.   The `package.json` should point to *node* code, and `bower.json` to the `dist` code.  

cf https://github.com/jupyterlab/services

blink1073:
In this way, consumers require `xterm` in a way that works best for their use case, without needing extra config or importing from `node_modules`.

parisk:
@blink1073 thanks for your comments and proposals.

What you are proposing is change the way we distribute xterm.js. While this is something that can change and can be debated, it is outside the scope of this PR.

The issue that this PR is trying to solve is that when we make temporary builds of xterm.js when we develop xterm.js, this happens in a directory that is being checked into Git by default (`dist`), which is outside the scope of your proposal.

Also, while I understand that your proposal can solve this issue and that it works great for Jupyter, currently I'd like to keep the "importing API" as-is, with all imports/requires etc. requiring files from the `dist` directory **by default**.

blink1073:
Fair enough, I'll move the comment over to #356.

bin/build
```diff
@@ -1,26 +1,35 @@
 #! /usr/bin/env bash
 
+set -e
+
+# $BUILD_DIR should default to "build"
+export BUILD_DIR=${BUILD_DIR:=build}
```
Tyriar: I don't think you want to export here, that means it will be available outside the scope of the script.

bin/build
```diff
@@ -1,26 +1,35 @@
 #! /usr/bin/env bash
 
+set -e
+
+# $BUILD_DIR should default to "build"
```
Tyriar: Should we use args instead so we don't need to mess with the env?

PR #393, opened 2016-12-06
Migrate Clipboard handler to TypeScript

parisk:
Also apply JavaScript rules in [`.editorconfig`](https://github.com/sourcelair/xterm.js/blob/824a9c6dbf12dd29685b63bee8899601308fe89c/.editorconfig) in TypeScript files as well.

Part of #335

### DIFF
--- a/src/Interfaces.ts
+++ b/src/Interfaces.ts
@@ -2,12 +2,26 @@
  * @license MIT
  */
 
+export interface IBrowser {
+  isNode: boolean;
+  userAgent: string;
+  platform: string;
+  isFirefox: boolean;
+  isMSIE: boolean;
+  isMac: boolean;
+  isIpad: boolean;
+  isIphone: boolean;
+  isMSWindows: boolean;
+}
+
 export interface ITerminal {
   element: HTMLElement;
   rowContainer: HTMLElement;
+  textarea: HTMLTextAreaElement;
   ydisp: number;
   lines: string[];
   rows: number;
+  browser: IBrowser;
 
   /**
    * Emit the 'data' event and populate the given data.
@@ -16,4 +30,5 @@ export interface ITerminal {
   handler(data: string);
   on(event: string, callback: () => void);
   scrollDisp(disp: number, suppressScrollEvent: boolean);
+  cancel(ev: Event, force?: boolean);
 }
--- a/src/handlers/Clipboard.ts
+++ b/src/handlers/Clipboard.ts
@@ -5,6 +5,15 @@
  * @license MIT
  */
 
+import { ITerminal } from '../Interfaces';
+
+// We are extending the ClipboardEvent interface, since TypeScript has not declared the
+// clientX and clientY properties that we use.
+interface IClipboardEvent extends ClipboardEvent {
+  clientX: number;
+  clientY: number;
+}
+
 /**
  * Prepares text copied from terminal selection, to be saved in the clipboard by:
  *   1. stripping all trailing white spaces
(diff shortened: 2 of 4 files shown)


src/Interfaces.ts
```diff
@@ -2,12 +2,26 @@
  * @license MIT
  */
 
+export interface IBrowser {
```
Tyriar: `IEnvironment` might be a better name for this as it's not necessarily a browser?

src/handlers/Clipboard.ts
```diff
@@ -5,21 +5,30 @@
...
 /**
  * Prepares text copied from terminal selection, to be saved in the clipboard by:
  *   1. stripping all trailing white spaces
  *   2. converting all non-breaking spaces to regular spaces
  * @param {string} text The copied text that needs processing for storing in clipboard
  * @returns {string}
  */
-function prepareTextForClipboard(text) {
-  var space = String.fromCharCode(32),
+export function prepareTextForClipboard(text: string) {
```
Tyriar: Specify return type:

```ts
export function prepareTextForClipboard(text: string): string {
```

src/handlers/Clipboard.ts
```diff
@@ -5,21 +5,30 @@
...
+interface IClipboardEvent extends ClipboardEvent {
+  clientX: number;
+  clientY: number;
+}
+
 /**
  * Prepares text copied from terminal selection, to be saved in the clipboard by:
  *   1. stripping all trailing white spaces
  *   2. converting all non-breaking spaces to regular spaces
  * @param {string} text The copied text that needs processing for storing in clipboard
```
Tyriar: You don't need to include types in the jsdoc anymore as TypeScript knows about it, do you need these for a tool you're using?

src/handlers/Clipboard.ts
```diff
@@ -49,23 +61,26 @@ function copyHandler(ev, term) {
  * @param {ClipboardEvent} ev The original paste event to be handled
  * @param {Terminal} term The terminal on which to apply the handled paste event
  */
-function pasteHandler(ev, term) {
+export function pasteHandler(ev: IClipboardEvent, term: ITerminal) {
   ev.stopPropagation();
 
-  var dispatchPaste = function(text) {
+  let w = <any>window,
```
Tyriar: If you need additional interfaces on it you could try:

```ts
interface IWindow extends Window {
  clipboardData?: {
    getData(type: string): string;
  };
}

declare var window: IWindow;
```

I don't think it's standard on `Window` which is why it's optional via `?`. That way you still get type checking.

You could also cast it to an `IWindow` instead of `any` and not use the declare for the same effect.

src/Interfaces.ts
parisk: I would prefer leaving it as `IBrowser` for now since:

1. xterm.js in general runs in the browser (node is the exception for testing)
2. it is named after the `Browser` util`

src/handlers/Clipboard.ts
parisk: Yep, sometimes I build docstrings with JSDoc (`npm run build:docs`) and this is useful information in general.

parisk:
@Tyriar feel free to review the updated PR and my replies to your comments.

parisk:
@Tyriar I move on with merging this in order to move forward with issuing the 2.2.0 release, since I fixed most of your comments, while the dangling ones are more of naming nature and can be changed post-release without affecting the user experience.

Tyriar:
lgtm!

PR #396, opened 2016-12-07
fix: change build dir to dist

imsnif:
I'm not sure whether this is intentional or not, but I think this is just a small oversight?
The docs say the addons are in the dist directory, but they're not - and the build script builds them into the build directory...

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -3,7 +3,7 @@
 set -e
 
 # $BUILD_DIR should default to "build"
-BUILD_DIR=${BUILD_DIR:=build}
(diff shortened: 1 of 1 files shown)


parisk:
@imsnif this is because we have not issues the 2.2.0 release yet, which will happen within the next minutes.

Sorry about the trouble.

parisk:
@imsnif this should be fixed in the [2.2.1 tag](https://github.com/sourcelair/xterm.js/tree/2.2.1) that just got pushed.

imsnif:
Cool, thanks! Sorry - it just stopped working for me with the latest pull.

PR #398, opened 2016-12-09
Add prepublish script and change main

### LINKED ISSUE #397: Failure when building with browserify
When building my project with browserify and babelify, I get the following error:

```
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: Cannot find module './Generic.js' from '/opt/project/node_modules/xterm/dist'
    at /opt/project/node_modules/resolve/lib/async.js:55:21
    at load (/opt/project/node_modules/resolve/lib/async.js:69:43)
    at onex (/opt/project/node_modules/resolve/lib/async.js:92:31)
    at /opt/project/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:123:15)
```

This is new from version 2.2.0 and 2.2.1.

## Details
- OS version: ubuntu-12.04
- xterm.js version: 2.2.0 and 2.2.1

### Steps to reproduce

1. Create a project witch `require('xterm')`
2. Build this project with browserify and babelify

budde377:
Building TS before publish and pointing the main entry to the builded files.

Closes #397

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,12 @@
     "test",
     ".gitignore"
   ],
-  "main": "dist/xterm.js",
+  "main": "lib/xterm.js",
   "repository": "https://github.com/sourcelair/xterm.js",
   "license": "MIT",
+  "files": [
+    "lib"
+  ],
   "devDependencies": {
     "@types/chai": "^3.4.34",
     "@types/mocha": "^2.2.33",
@@ -30,8 +33,9 @@
   },
   "scripts": {
     "start": "nodemon --watch src --watch addons --watch demo --exec bash -c './bin/build && node demo/app'",
-    "test": "./bin/build && mocha --recursive ./out",
+    "test": "mocha --recursive ./lib",
     "build:docs": "jsdoc -c jsdoc.json",
-    "build": "./bin/build"
+    "build": "./bin/build",
+    "prepublish": "tsc --outDir lib"
   }
 }
--- a/.npmignore
+++ b/.npmignore
@@ -2,7 +2,6 @@ node_modules/
 *.swp
 .lock-wscript
 out/
-lib/
 Makefile.gyp
 *.Makefile
 *.target.gyp.mk


package.json
```diff
@@ -6,7 +6,7 @@
     "test",
     ".gitignore"
   ],
-  "main": "dist/xterm.js",
+  "main": "out/xterm.js",
```
parisk: Can you leave this as `dist/xterm.js` please, since the browserify case is not our main use case?

package.json
```diff
@@ -32,6 +32,7 @@
     "start": "nodemon --watch src --watch addons --watch demo --exec bash -c './bin/build && node demo/app'",
     "test": "./bin/build && mocha --recursive ./out",
     "build:docs": "jsdoc -c jsdoc.json",
-    "build": "./bin/build"
+    "build": "./bin/build",
+    "prepublish": "./bin/build"
```
parisk: I think that you can use just `tsc` here? /cc @Tyriar

parisk:
Thanks for your contribution @budde377! The proposal is pretty good. Being able to also require the built pre-bundled JavaScript files can definitely have it's use cases.

Now if we decide to ship these files in npm's xterm.js distribution, maybe it's a good idea to change the name of TypeScript's output directory to `lib` which is a little bit more semantic than `out` and implies that inside there there are some kind of "library files".

@Tyriar any thoughts on this?

package.json
budde377: Of course. 
But that would entail that you could never use `require('xterm')` in Node, as we have seen. You would always have to write something like `require('xterm/out/xterm.js')` which isn't really that pretty. Also, I see no downside with letting main to point to the `out/xterm.js`. It would still work in the browser using the ´<script src="node_modules/dist/xterm.js"></script>` tag and it would work natively in node..

package.json
budde377: I think that you are right

Tyriar:
That would entail committing `lib` then as well right? Maybe we should re-consider the release branch? Keeping 2 copies of built files around in the repo feels really messy.

blink1073:
You can release a sub-folder on npm without committing it, using `files`: https://docs.npmjs.com/files/package.json#files.

Tyriar:
@blink1073 didn't know about that

parisk:
@Tyriar what's your opinion on publishing `lib` on NPM only, while ignoring this directory in Git.

Tyriar:
@parisk whatever we can do to keep compiled files out of the master branch is good imo, so if it works 👍

budde377:
Are you guys sure that you want to keep the entrypoint for npm to be `dist/xterm.js`. See my comment above.

Tyriar:
If we use `files` in `package.json` then I think we should change it to point at `lib/xterm.js`

budde377:
I agree.

Tyriar:
It's worth bringing up (again) that if we don't need to check in `lib` for `npm` support, all that is keeping us checking in `dist` is Bower which is on the decline. Most devs are moving away from it as `npm` now downloads deps flatly and Bower is a pain for module owners to support.

parisk:
@budde377 can you also include the `lib` directory using the [`files`](https://docs.npmjs.com/files/package.json#files) field in the `package.json` file?

@Tyriar yep, I think it's time to drop Bower: #406.

budde377:
Done

parisk:
Works great, thanks!

Tyriar:
Woo! Great result :smile:

blink1073:
Yes!  Thank you so much for this.

logicminds:
Having the same issue.  Despite having this merged I don't understand what I need to do when using browserify to require xterm in my project.

Is this what I need to do in my index.js file?

```
var xterm = require('./node_modules/xterm/lib/xterm');
var fetch = require('fetch');
var attach = require('./addons/attach/attach.js');
var fit = require('./addons/fit/fit.js');
var fullscreen = require('./addons/fullscreen/fullscreen.js');
var linkify = require('./addons/linkify/linkify.js');

```

PR #399, opened 2016-12-09
Don't apply the 32 delta when in sgr mode - fix #394

### LINKED ISSUE #394: Mouse buttons are wrong with extended coordinates
The 1006 mode (SGR mouse, `\033[?1006h`) is the recommended way to deal with extended coordinates without risking conflicts between sequences. Unfortunately, it doesn't work with xterm.js:

 With mintty:

```
left mouse click: 0
left mouse drag: 32
mouse wheel down: 64
```

With xterm.js:

```
left mouse click: 32
left mouse drag: 64
mouse wheel down: 96
```

I think the reason for these discrepencies is that the `pressed` value ([here](https://github.com/sourcelair/xterm.js/blob/master/src/xterm.js#L652), [here](https://github.com/sourcelair/xterm.js/blob/master/src/xterm.js#L677) and [here](https://github.com/sourcelair/xterm.js/blob/master/src/xterm.js#L863)) always has a value of at least 32, which incorrectly shift all button numbers. As you can see in the [XTerm documentation](http://invisible-island.net/xterm/ctlseqs/ctlseqs.html), you should never add this extra value when operating under the SGR mode, because it encodes button numbers as numbers instead of ASCII characters:

> ### SGR (1006)
>
> The normal mouse response is altered to use CSI < followed by semicolon-separated encoded button value, the Cx and Cy ordinates and a final character which is M  for button press and m for button release.
>
> - **The encoded button value in this case does not add 32 since that was useful only in the X10 scheme for ensuring that the byte containing the button value is a printable code.**
>
> - The modifiers are encoded in the same way.
>
> - A different final character is used for button release to resolve the X10 ambiguity regarding which button was released.
>
> The highlight tracking responses are also modified to an SGR-like format, using the same SGR-style scheme and button-encodings.

This bug is quite damaging, since it's not possible to easily detect whether the input is a standard left mouse drag vs a buggy left mouse click, or a standard mouse wheel down vs a buggy left mouse drag.

bottleofwater:
Cf #394 - "Mouse buttons are wrong with extended coordinates"

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -785,7 +785,7 @@ Terminal.prototype.bindMouse = function() {
       pos.x -= 32;
       pos.y -= 32;
       self.send('\x1b[<'
-                + ((button & 3) === 3 ? button & ~3 : button)
+                + (((button & 3) === 3 ? button & ~3 : button) - 32)
                 + ';'
                 + pos.x
                 + ';'


parisk:
Seems to be working great, thanks!

PR #400, opened 2016-12-10
Use a fixed size to account for the scrollbar

arcanis:
The current fit algorithm removes one column, which I assume is to account for the scrollbar. It's unfortunately not enough, since the size of a character might be less than the size of the scrollbar, in which case a visual artifact occurs.

image

This commit solves this by removing a fixed amount of pixels from the available pixels. The amount itself (17) is the scrollbar size on various browsers. A better implementation would probably try to compute it dynamically, but I felt like using a static value could a good start (especially since apparently noone reported this issue until now).


parisk:
Thanks for your contribution @arcanis! Looks great 😄 .

PR #402, opened 2016-12-12
Improve the package.json scripts

akalipetis:
1. Include a `prestart` script, which runs the build
2. Make the `start` script just start the demo server
3. Add a new `dev` script, which contains the nodemon logic
4. nodemon now watches for changes in `*.ts` files and runs the start command

This allows for easier development, while also makes the different scripts we use simpler and self-contained.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ npm-debug.log
 /.idea/
 .env
 build/
+.vscode/
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,1 @@
+web: npm start
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,1 +1,1 @@
-web: npm start
+web: npm run dev
--- a/package.json
+++ b/package.json
@@ -52,7 +52,9 @@
     "typescript": "^2.0.3"
   },
   "scripts": {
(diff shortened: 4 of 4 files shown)


parisk:
Overall, this is a great addition 👍 , just two comments:

1. Can you please change the `web` script to `npm run dev` in `Procfile.dev`?
2. Should we add also a `Procfile` with `web: npm start`?

.gitignore
```diff
@@ -12,3 +12,4 @@ npm-debug.log
 /.idea/
 .env
 build/
+.vscode/
```
Tyriar: Curious what you're ignoring here, did an extension or something create a file?

akalipetis:
Thanks @parisk and @Tyriar for your comments - Paris I have updated the `Procfile*`s.

I have also rebased with master.

.gitignore
parisk: Let's leave this here, since vscode might be creating any files, I think it's a good idea to ignore them for sure.

Non-blocking question: Should we change this to `/.vscode/` like the `/.idea/` above or not?

Tyriar: @akalipetis thanks, I was mainly just wondering if an extension was being naughty 😛 

.vscode can only appear at the root of the repo so `.vscode/` should be fine.

PR #403, opened 2016-12-12
Improve the Docker image

akalipetis:
1. Bump Node version to 6.9 - the latest LTS
2. Include the `cpio` binary, used during building
3. Ignore node_modules and .git directories for faster builds
4. Run the tests and build during the image build, to make sure the image is not built if these break
5. Make npm run dev the default command
6. Add an entrypoint to automatically install Node modules if they do nott exist

Signed-off-by: Antonis Kalipetis <akalipetis@gmail.com>

Based on #402 - since it uses the `npm run dev` script.


.gitignore
```diff
@@ -12,3 +12,4 @@ npm-debug.log
 /.idea/
 .env
 build/
+.vscode/
```
Tyriar: Curious what you're ignoring here, did an extension or something create a file?

entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
```
Tyriar: Can this be inlined into Dockerfile or something? Seems a bit messy to have a file in root that just runs npm install.

parisk: I think that we can run `npm install` "on building" of image by using the `node:6.9.2-onbuild` image instead.

I agree that the `entrypoint.sh` in the root of the project kinda bloats it. I somehow missed it before.

akalipetis: This only helps during development - @parisk the `on-build` image flavor will still not work since we use volumes to mount the code inside the container.

We can just omit this and improve our docs - users should do `docker-compose run --rm web npm install && docker-compose up` the first time they want to spin up the project.

I'm 👍 with that, just let me know if you're Ok with this approach and I'll do a quick update.

parisk: Would it work if we inlined the ENTRYPOINT into the Dockerfile?

```dockerfile
ENTRYPOINT "[ ! -d node_modules ] && npm install; $@"
```

Or something like the above?

akalipetis:
Rebased with master and fixes an issue with EOL characters.

Tyriar:
LGTM, I'm no authority on docker though :smile:

PR #404, opened 2016-12-12
Line wrap

imsnif:
Continuing the conversation in this PR: https://github.com/sourcelair/xterm.js/pull/386
And addressing this issue: https://github.com/sourcelair/xterm.js/issues/325

This is a rough implementation, and is not yet 100% ready - but it is mostly working.
The reason I'm making this PR now is that I've made quite a few assumptions regarding the way you'd like to solve this issue.
Namely, I've added a fourth "cell" to the array representing each character. The cell can either be undefined, 'wrapped', or 'pad':
'wrapped' would mean that this character has been wrapped when the terminal window was shrunk - we have to mark this because we need to know to 'unwrap' it when the terminal window is expanded again.
'pad' would mean that this character is just a padding character (when we wrapped a line and had to fill the space up until the column width of the terminal) - we have to mark this so that we know these characters can be discarded when we unwrap the line (we can't just do this with normal whitespace, because otherwise we'd end up potentially removing actual 'wanted' whitespace).

If this solution is generally acceptable, I would be happy to expand the test coverage, tidy up the code and fix some small issues (namely the cursor x position when expanding, and integrating it with the current wrapping mechanism in the write method).

Would very much love to hear your thoughts.


src/utils/LineWrap.js
```diff
@@ -0,0 +1,47 @@
+export function padLine (line, x, defAttr) {
```
Tyriar: Missing a copyright statement (copy from some other file)

src/utils/LineWrap.js
```diff
@@ -0,0 +1,47 @@
...
+  var ch = [defAttr, ' ', 1, 'pad']; // does xterm use the default attr?
```
Tyriar: All new files are being being built in TypeScript, this will basically involve renaming the file to use extension `.ts` and fixing up the errors on `npm start` (if any).

src/xterm.js
```diff
@@ -2856,24 +2857,62 @@ Terminal.prototype.resize = function(x, y) {
...
-  j = this.cols;
-  if (j < x) {
-    ch = [this.defAttr, ' ', 1]; // does xterm use the default attr?
-    i = this.lines.length;
-    while (i--) {
-      while (this.lines[i].length < x) {
-        this.lines[i].push(ch);
+  var origLineCount = this.lines.length
+  this.lines = this.lines
+  .map(l => unpadLine(l))
```
Tyriar: I'm a little concerned about how many arrays are being constructed here on every resize as `this.lines` is capable of becoming very large (configuring `scrollback`). Have you considered an approach similar to this?

- On resize, calculate how many rows each line will take up so that we can fetch a 'virtual line' for any row `y`, Encapsulate this inside a TypeScript class (a row index would know which index in `this.lines` and also which index of the line to start/end at).
- Inside `refresh`, fetch each line using the lookup generated in on resize.
- Potentially remove the padding from the lines all together if it's reasonably fast to determine the width of each character in the line.

imsnif: Sounds good - that's a pretty neat solution that would do away with most of this. Thanks!
Just to make sure I understand correctly - the Typescript class would be inserted into this.lines instead of the respective line (if resize deems it should be wrapped or unwrapped), and be dealt with in the refresh method so it displays accordingly?
That would also mean moving the resize method's update of this.y, this.ybase and this.ydisp into the refresh method as well, I think?

In any case, I'll try to whip something up sometime this week.

Tyriar: So initially I'm thinking just keeping a small class or array alongside `this.lines` that mapped to the various indexes. Something like this:

```ts
export class IRowPosition {
  lineIndex: number; // the index of this.lines
  startIndex: number; // The start index in this.lines[lineIndex]
  endIndex: number; // The end index in this.lines[lineIndex]
}

// I'm sure there's a better name for this
export class WrappedLineLookup {
  // this.lines[y]
  // ->
  // var row = this.lineLookup.getRow(y);
  // var line = this.lines[row.lineIndex]
  // for (var x = row.startIndex; x < row.endIndex; x++) { ... }
  public getRow(index: number): any { ... }

  // this.line.length -> this.lineLookup.rowCount
  public get rowCount(): number { ... }
}
```

We should try getting the naming on this right though as it could quickly get confusing (line, row, wrappedLine, `IRowPosition`'s members, etc.)

Eventually we should completely wrap `this.lines` in a `TerminalBuffer` class (utilizing CircularList from my [WIP branch](https://github.com/Tyriar/xterm.js/blob/361_circular_list_scrollback/src/utils/CircularList.ts)) and moving `ybase`, `ydisp`, `y` into `Viewport`. That's quite a lot of refactoring though, we should slowly move to a better design so we don't break everything in the meantime 😉

imsnif: Wouldn't keeping an additional object to this.lines be problematic for the same reason my solution from the previous PR (with this.lineCache) was problematic? That if this.lines is changed outside the resize method everything would get messed up?

Tyriar: Good point 😦

Tyriar: This is quite a tricky problem to solve without totally rewriting the structure of `this.lines` (which could invalidate much of the work [here](https://github.com/Tyriar/xterm.js/blob/361_circular_list_scrollback/src/utils/CircularList.ts)). Keeping the indexes alongside the lines would get stale so they would need to be offsets, but that means we potentially need to traverse the entire list up to a point to find the actual index of an item.

imsnif: So how about trying for my first interpretation of your suggestion?
Any line in this.lines that needs to be wrapped is transformed into an instance of a Typescript class* that keeps the original this.lines contract (so that nothing else will be broken), but also allows to display the line as wrapped in the this.refresh method? (as well as change the wrapping or entirely unwrap it and turn it back to a 'normal' array as necessary)
It's a little hacky, but as something that transitions us towards TerminalBuffer - maybe it'll work out?

*I'm not sure a Typescript class would fit this case, I'd have to play around with it a little. Worse case I'll just compose an array with some appropriate methods.

Tyriar: The problem with that is that big chunks of the viewport are discarded when the buffer limit is reached, that would mean recalculating this a lot if there is a lot of output.

That led me to thinking about a relative index (ie. a line knows how many rows it consumes), but then if we wanted to view something in the middle of the terminal we would need to go over 50% of the lines to find the correct position.

I don't think it's been mentioned yet but all this is only necessary to support being able to scroll to every row within a line, otherwise scrolling would probably feel bad.

imsnif: I'm not sure I understand the problem with the relative index. Are there many cases in which we need to insert data in the middle of the terminal screen?

src/utils/LineWrap.js
```diff
@@ -0,0 +1,47 @@
...
+  while (line.length <= x) {
+    line.push(ch)
+  }
+  return line
+}
+
+export function unpadLine (line) {
```
parisk: This method needs documentation, especially since it is in single-line syntax, which can be hard to read and understand.

src/utils/LineWrap.js
```diff
@@ -0,0 +1,47 @@
...
+  return line.filter(c => !c[3] || c[3] !== 'pad')
+}
+
+export function wrapLines (unwrappedLine, x, defAttr) {
+  return unwrappedLine.reduce((memo, e, index) => {
```
parisk: Please use traditional function syntax, when the function has "actual body", which is closer to the current coding style.

src/utils/LineWrap.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+    if (memo[memo.length - 1].length >= x) {
+      memo.push([])
+    }
+    if (index !== unwrappedLine.length - 1 || index === unwrappedLine.length - 1 && e[1] !== ' ') {
+      memo[memo.length - 1].push(e)
+    }
+    return memo
+  }, [])
+  .map((wrappedLine, index) => {
```
parisk: Please use traditional function syntax, when the function has "actual body", which is closer to the current coding style.

imsnif: My apologies for the rough implementation. I plan on cleaning everything up once we decide on a proper solution to the problem discussed above.

parisk:
First of all thanks a lot for your contribution @imsnif! I 'd really like to see this moving forward and help as much as possible with this, since it's a great addition and we have a working PoC.

Let me share a few thoughts and concerns with the current implementation, which are definitely "tackle-able".

## 1 - Code formatting
Before merging this we should ensure that the code formatting and styling is as close as possible to the code base's formatting and styling. This requires the following:

- Code is in TypeScript (If you don't feel comfortable with this, I can help)
- All methods are documented using JSDoc
- Use traditional function syntax instead of arrow functions in **full-bodied**  functions, to keep consistency

## 2 - Code structure
Considering code structure, I am quite skeptical for introducing a fourth flag. Amending the same object in many different places can be a mess and the core parts of xterm.js that did not change after we forked term.js are still a little bit messy.

Let's not add any more complexity to this. It will only make things worse.

What I propose instead is handle this in a "self-contained environment" (e.g. a resize handler at `handlers/Resize.ts`, where we could migrate all resize logic also afterwards). There you can keep a "wrapping state" object, which will contain all information you need (which characters of which lines are wrapped or padded).

A proposed quick-draft format of such an object I thought might work is:

```javascript
// Object in the following format:
//
// {
//   lineNumber: lineWrapData,
//   anotherLineNumber: anotherLineWrapData,
//   andSoOnLineNumber: andSoOnLineWrapData
//  }

var wrapState = {
  1: {
    'state': 'wrap',
    'wrapStart': 75
  }
};
```

Does this make sense at all? (Ideally these structs should be TypeScript classes or interfaces.

/cc @Tyriar to take a look as well

src/utils/LineWrap.js
parisk: No problem. Thought about it, but I also thought to mention this nevertheless!

Please take a look at my comment below which might be helpful: https://github.com/sourcelair/xterm.js/pull/404#issuecomment-266826478

The work you have done is really great. Thank you so much for this.

Tyriar:
@parisk I think a large resize handler is a bit too heavy, I think it looks great right now that I don't have to debounce the `resize` event, we should try to keep that.

Also as mentioned in [this thread](https://github.com/sourcelair/xterm.js/pull/404#discussion_r92017594), there's an issue keeping this data away from the `lines` object as they will easily get out of sync when lines is trimmed (due to `scrollback`) or changed (lines in the middle of the buffer can change).

src/xterm.js
imsnif: I see.
So - as I understand it now (please do correct me if I'm wrong), we can either:

1. Use a relative index. This will be problematic with scrolling. Because when we scroll, we could either:
    1. Scroll by the non-relative this.lines index, and have scrolling be a little weird and inconsistent when it encounters 'unevenly wrapped' lines.
    2. Scroll by the relative index, and then have potential performance issues when scrolling to the middle of the buffer.
2. Use an external index (as both of you @Tyriar and @parisk suggested). But then we'd have to adjust all places in which this.lines is updated (outside the resize method as well) so that they also update the external index.

Is this about right? Are there more options?

If there aren't any other options, which option do you guys think is best? I'd be willing to implement or POC any of them - though with some of them (particularly the external index one) I could definitely use some help to avoid some gotchas stemming from my unfamiliarity with the codebase.

parisk: I think that the best option is going with an external index. What I can definitely do is refactor `term.lines` into a TypeScript class that won't be amended directly like a JavaScript array.

This way we will also be able to unit test the functionality of this class and ensure its integrity.

Now, this class can emit some events, where the wrap handler will be able to listen to in order to update it's state.

How does that sound? Thinking about preparing a PoC right now.

imsnif: Sounds great to me @parisk 
Maybe just make sure with @Tyriar , because I think he mentioned earlier that this sort of solution might break some of his ongoing work?
If we're going with this, I'd love to amend this PR to include the wrap handler.

parisk:
@Tyriar the resize logic is complex and actually is an important part of the rendering process. Moving it into it's own module eventually with < 300 lines of code in total will make things more manageable. (this does not have to happen right now)

src/xterm.js
parisk: Yep, let's wait for @Tyriar to get on top of the discussion before moving forward.

In the meantime I threw a quick PR (which doesn't have to be merged at all) on how the `Lines` class could work: #410. The demo works after implementing these changes btw.

I tried to stay as close as possible to the `Array` API, in order to avoid breaking code in `src/xterm.js`.

Since we have "managed" methods now, this means that we can emit events to let the rest of the code that a change happened in the lines.

I think that if we implement tests for this functionality (appropriate event emitting) we can rely on it for keeping everything in sync.

@imsnif in the meantime, could you share your opinion on if this PR, accompanied with event emitting, could help in implementing a proper external index?

imsnif: Looks great to me. I think it can cover it. I'd be happy to whip up a new PR that would implement it. (If all is okay with everyone, of course).

parisk: @imsnif let's wait for @Tyriar to push his branch including the CircularList, which implements the functionality I showcased in my own branch. Then you will be able to re-use the code included there as a foundation for the line wrapping.

Discussion thread: https://github.com/sourcelair/xterm.js/pull/410#issuecomment-267022749

parisk: Thanks again for your effort and for your co-operation so far!

imsnif: Sure thing. Let me know when I can start working on it.

imsnif: I see https://github.com/sourcelair/xterm.js/pull/422 was merged. Is that it then, shall I start working on line wrapping over the weekend?

Tyriar: That would be great :smiley: I suggest starting out by extending `CIrcularList` in a `Lines` class similar to what https://github.com/sourcelair/xterm.js/pull/410/files was trying to accomplish before `CircularList` existed. That way we can keep `CircularList` nice and generic and keep that complexity in that class.

jerch:
An additional note on this: the terminal can actually decide at the time the characters are newly inserted, at which position a line overflows and wraps. This position could be used to for an initial external index, which would enable enlarging a terminal window with correct wrapping/reflowing behavior, too.

Tyriar:
@imsnif some big changes landed in master last night so you'll want to sync your branch with upstream. Let me know if you're having issues with the solution 😃

imsnif:
Thanks @Tyriar, I'm currently working out some kinks in my prototype (it's basically working - had some trouble with identifying the correct x position after wrapping, but I'm on my way to resolving them). I'll pull before I clean up.

Thanks for the heads up, and I'll definitely take you up on your offer of advice if I get stuck.
Sorry it's taking more than I expected, I'm slowly familiarizing myself with the code. :)

parisk:
@imsnif thanks a lot for putting in the effort for this. It's really important for the project.

No surprise it took more than expected. This is a big update after all.

Reach out whenever you need some help. 👍

parisk:
@imsnif may I ask if there is any progress on this? It would be great if we could get it into the code base now that it got much cleaner!

Please feel free to ask for assistance, if you'd like. We'd love to help!

Again, thanks a lot for putting the effort on this.

imsnif:
Hey @parisk - I was actually about to post something today or tomorrow about my progress, so you beat me to it.
I've had less time to work on this than I'd like in the past couple of weeks - and again I'm sorry this is taking a while.
What I have right now in my prototype:
The basic resize functionality is working. I did this in an outside ts class (temporarily named 'LineWrap') and in the refresh and resize methods. This means that resizing the terminal wraps and unwraps lines successfully.
I'm currently working on what I think will be the last major part of this PR, which is the write method. This method has its own overflow mechanism (as mentioned here above by @jerch). I started playing around with having it talk to my LineWrap class this morning - but I haven't yet managed to get it to work. I hope to finish that part this weekend.
After that, I plan to write some tests so that I can easily merge with the new changes in upstream. That will also give me an opportunity to do some refactoring and make it readable.
Following this, I'll have to go over all of the code and incorporate such things as 'eraseRight' and such with the new logic. I will definitely like some input at that stage from people who are more familiar with the code base than I am.
Finally, I'm going to clean everything up - fix my horrible Typescript faux pas (this is my first experience with Typescript) and then submit everything to your review.

I hope it's okay that this is taking a short while. I'm very much invested in this fix and plan to see it through. I'll try to post some updates every now and then, and meanwhile (if you don't mind my mess), you can follow my progress here: https://github.com/imsnif/xterm.js/tree/line-wrap-prototype

parisk:
Awesome @imsnif!

It's totally OK that this is taking some. It's a big change and will be beneficial for all users.

xterm.js was also my first experience with TypeScript, so don't feel bad about that 😄 .

imsnif:
Hey, just a quick update. I have a very rough prototype working with the external index implementation we discussed above.
The code is still a mess, and there are still some minor issues (cursor x positioning doesn't work at times, vim/fish/top/etc. support, extra lines added to this.lines causing scroll issues, etc.)
I will now start cleaning stuff up, merging with the changes in upstream and fixing the various issues.
I made a small very basic demonstration if you want to see how it looks: (not embedding because then this thread will become a little cumbersome, I think :) ):

http://i.imgur.com/dkQsiYg.gif

Tyriar:
@imsnif the gif looks great :smiley:

amir-arad:
hey @imsnif any ETA?

imsnif:
@amir-arad hoping to be done by the end of this month.

parisk:
That's great @imsnif, thank you.

Everyone let's hold on asking ETAs though. We should put no pressure on contributors.

Tyriar:
@imsnif you might want to rebase against origin/master and recreate the PR to fix the non-you commits, the diff and get rid of the older conversations.

imsnif:
Oh dear, pushed to the wrong branch. :)
This is still very much a WIP.

LucianBuzzo:
@imsnif Gentlest of pings 🙏 
I've put together my own implementation of the line wrap feature -https://github.com/LucianBuzzo/xterm.js/tree/lb-line-wrap
Maybe you could integrate it into the work you've done so far or perhaps I could assist with the work you're doing?

imsnif:
Hey @LucianBuzzo,
First - thanks for this! I appreciate the offer of help. There's really no need to be gentle, not on my part. I think we all just really want this to be implemented. :) I do appreciate it though.

This looks great! A quick glance through the code shows me that this is implemented slightly differently than what we talked about above here, correct me if I'm wrong? Not with an external index but rather with caching the overflow?
I think this might make it somewhat difficult to reconcile the two branches. :/
If you think you can bring this to completion quickly, and @Tyriar and @parisk don't mind changing the implementation to work in this manner, I'd be more than happy to hand this hat over to you. :)

If you'd like to help out with my current wip branch (https://github.com/imsnif/xterm.js/tree/line-wrap-wip) - I'd also love the help. It really is almost finished... This mostly suffered from an unexpected lack of free time on my part.
There are just a few quirks to work out, and then some refactoring and more tests. 
Would love to discuss these all and get some more helping hands for them if that's the course you'd like to take.
@Tyriar also mentioned to me privately that he'd like to help. So how about the three of us get into this and see what can be done to get it finished?

LucianBuzzo:
@imsnif If you're close to completion on your branch then I'll dive in and help on that. I suggest opening up issues on your fork for outstanding tasks and I'll submit PR's directly to your WIP branch.
What do you think?

imsnif:
@LucianBuzzo - sounds great. I'll start opening issues and tag you. If anyone else wants in, just let me know.

imsnif:
So, I've opened some issues in my fork. Anyone who feels like helping out: please, by all means. If anything is unclear, please let me know and I'll do my best to get back to you ASAP.
If anything is left by this weekend, I'll also keep going. :)

LucianBuzzo:
After a review of the external index approach and some discussion between @imsnif and myself here https://github.com/imsnif/xterm.js/issues/1#issuecomment-285179369, we feel that the solution in this branch https://github.com/LucianBuzzo/xterm.js/tree/lb-line-wrap may be the best course of action.
@parisk @Tyriar If you are happy with this I think we should redirect our efforts into completing the POC branch I've linked above.

parisk:
Thanks for putting the joint effort on this @LucianBuzzo and @imsnif.

If the end experience is the same for the end user and the code quality is satisfying, then there is no objection to move on by me. So let's move on 😄 .

Is there any additional input needed on our behalf at this moment?

parisk:
Closing this in favor of #609.

PR #407, opened 2016-12-13
Always build TypeScript in `lib/`

parisk:
Now that we publish `lib` to npm after #398, I do not think that there is any reason to keep the `out` directory.

Also this will make it easier for people that clone the bare repo to work (or else they will have to run the "prepublish" script).

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,9 @@
 node_modules/
 *.swp
 .lock-wscript
-out/
+lib/*.test.js
+lib/*.test.js.map
+lib/test/
 Makefile.gyp
 *.Makefile
 *.target.gyp.mk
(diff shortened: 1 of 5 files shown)


Tyriar:
You will want to remove any generated test files in lib/, we don't want them in the npm package. Do glob patterns work in package.json `files`?

blink1073:
Yes, they do: https://github.com/jupyterlab/services/blob/eb5b3d78c0485e0b761116341e7a98729852f966/package.json#L57

.npmignore
```diff
@@ -1,7 +1,8 @@
 node_modules/
 *.swp
 .lock-wscript
-out/
+lib/*.test.js
+lib/*.test.js.map
```
parisk: Great catch. Rebased: 230d277

parisk: Re-rebased :trollface: : e23956c

PR #413, opened 2016-12-15
Addon attach-electron for attaching to ipcRenderer 

saswatds:
attach-electron addon for using the ipcRenderer as the attachment. Also contains the documentation of how to use it.


README.md
```diff
@@ -52,6 +52,28 @@ npm start
...
+term.open(document.getElementById('#terminal'));
+term.write('Hello from \033[1;3;31mxterm.js\033[0m $ ')
+```
+To get the data from xterm to send to a server or for proccessing you can
+
+```js
+function onDataReceived(data) {
+   // Your logic
+}
+term.on('data', onDataReceived);
```
Tyriar: This is very similar to our quick start on http://xtermjs.org/

Maybe it's worth calling out `term.on('data', _)` there and potentially pulling that into the readme as a complete example like on the site? @parisk?

saswatds:
Sorry i messaged up while creating another pull request for the addon I made. Have opened the getting started as a separate branch

saswatds:
@parisk is this an acceptable addon for the project? Or should i find out a way to integrate it into the attach addon?

parisk:
@saswatds sorry for being late to get on top of this.

Can you share some details on the use case of this PR? I am not sure I understand how and why should someone use the new ipc attach addon.

Thanks!

Tyriar:
This isn't how I embed using electron, I'm also curious what the use case for this is exactly.

saswatds:
@Tyriar @parisk The application that I am working on uses different transport methods socket.io and zeromq to communicate with around 100 of ssh clients all running simultaneously. So the task of handling these connections happens in the main process, but somehow I needed to listen to these channel and render  the output to the terminal based on which client the user selects on the front-end. In that case i would just require to bind the xterm to the required ipcRenderer channel. This kind of solved my required task.  

P.S - I would also like to know what your approach is to embed xterm using electron.

Tyriar:
This seems a little too specific to your use case to include as an addon. I use [pty.js](https://github.com/Tyriar/pty.js) to embed it.

saswatds:
@Tyriar I guess so its bit specific to my use case, but there might be occasions where someone might just want to decouple pty.js to the main process and might need the ipc to communicate with the render process. Like in the case where we just want to keep the  ssh session open and decouple it from the front-end life cycle. 

In my use case i spawn a lot of child processes with pty.js to connect to a lot of raspberrypi's ...

parisk:
@saswatds I agree with @Tyriar that this is quite specific right now to include it in the main tree of xterm.js, which implies support by us.

I propose closing this PR now, until we have at least a more generic use case for this (I am also missing a few parts here, but let's not stick on details).

If you are OK, I will close this PR and please feel free to open an issue to continue discussion there or send a message in our Gitter chatroom: https://gitter.im/sourcelair/xterm.js.

saswatds:
Yeah Sure.. No problem ..

PR #415, opened 2016-12-15
Update README.md

saswatds:
Adding a getting started instruction for adding xterm to a parent div and listening for data output from xterm

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,25 +54,25 @@ Then open http://0.0.0.0:3000 in a web browser (use http://127.0.0.1:3000 if run
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


parisk:
Thanks your contribution @saswatds! I think the best thing would be shaping this based on the **Quick start** section at http://xtermjs.org/, which provides also a more complete example.

Do you believe you can refactor this accordingly?

Tyriar:
Here's the repo: https://github.com/xtermjs/xtermjs.org/

saswatds:
Yeah sure.. I am on.

parisk:
Thanks a lot for this @saswatds!

PR #417, opened 2016-12-16
changed escape sequence for alt-arrow to work on bash on os x

BobReid:
These are the escape sequences that make alt-arrow work on os x bash. 

I have just started poking around the code base and am not super familiar with it. Any guidance on how to make this cross platform / cross shell would be appreciated.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -2501,7 +2501,7 @@ Terminal.prototype.evaluateKeyEscapeSequence = function(ev) {
         // HACK: Make Alt + left-arrow behave like Ctrl + left-arrow: move one word backwards
         // http://unix.stackexchange.com/a/108106
         if (result.key == '\x1b[1;3D') {
-          result.key = '\x1b[1;5D';
+          result.key = '\x1bb';
         }
       } else if (this.applicationCursor) {
         result.key = '\x1bOD';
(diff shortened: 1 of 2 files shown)


BobReid:
Fix for https://github.com/sourcelair/xterm.js/issues/225

src/xterm.js
```diff
@@ -2501,7 +2501,7 @@ Terminal.prototype.evaluateKeyEscapeSequence = function(ev) {
         // HACK: Make Alt + left-arrow behave like Ctrl + left-arrow: move one word backwards
         // http://unix.stackexchange.com/a/108106
         if (result.key == '\x1b[1;3D') {
-          result.key = '\x1b[1;5D';
+          result.key = '\x1bb';
```
Tyriar: You can wrap this in a platform check by using `if (browser.isMac)`. This change also fails a test which will need its own if statement.

@parisk do you expect this to break anything? I'm concerned about how https://github.com/sourcelair/xterm.js/issues/225#issuecomment-267358011 could not repro and other it may not behave in other shells like zsh.

BobReid: I just tested with zsh. 
I actually reprod the issue with zsh as well with the old escape codes. The new escape codes actually works with both bash and zsh on my mac which actually caught me by surprise. 

I will wrap the changes in the isMac check and update the tests

BobReid: I see that there are already some test that mock out the browser check so I will go with that for now.

BobReid:
Updated the PR. 
Added the browser check and fixed tests

parisk:
Thanks @BobReid!

aliatsis:
this still doesn't fix alt+Backspace to delete words on mac.
should that be a separate issue?

BobReid:
alt+Backspace does not have a special function on my mac (bash & zsh on stock terminal and iTerm)

parisk:
On my tests `Alt` + `Backspace` just deletes a character in bash.

@aliatsis what macOS version are you running currently? Also are you using the stock terminal with bash?

aliatsis:
@parisk I'm on Sierra (10.12).. it was the same in Yosemite before I upgraded.
Yes, I am using the stock terminal with bash

PR #419, opened 2016-12-16
Clean up npm publish

blink1073:
Make sure the full build script is run before publishing, ensuring that CSS files are copied, for example

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,26 +10,12 @@
   "repository": "https://github.com/sourcelair/xterm.js",
   "license": "MIT",
   "files": [
-    "dist/*.css",
-    "dist/**/*.css",
-    "dist/*.js",
-    "dist/*.js.map",
-    "dist/**/*.js",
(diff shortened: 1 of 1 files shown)


parisk:
@blink1073 thanks for putting the effort on this, but is not the intended behavior.

The issue is that this would break applications that depend on either the built monolith or the TypeScript sources and install xterm via npm.

To make it work in such a case it would require a build step before using it, which is not desired.

Let me know if I understood well and it is OK to close this. Otherwise, please share the reasons you would like to remove these files from the NPM package (e.g. are they introducing issues in your application that depends on xterm.js?).

blink1073:
The intent of this PR was to make sure the CSS file was getting published, removing the other paths was in response to https://github.com/sourcelair/xterm.js/pull/416#discussion_r92718913.  Updating now.

parisk:
Thanks a lot @blink1073!

PR #420, opened 2016-12-16
Remove dist/ from source tree

Tyriar:
I don't believe we need it anymore.


parisk:
We will have to migrate the Bower package to another repository that will have `dist/` checked out before removing it from the main repo.

parisk:
Should we close this, until we lay down a plan for xterm.js 3, where we can deprecate/drop bower support?

vincentwoo:
Hi, sorry to ping a closed issue, but I would consider doing this sooner rather than later. One of the annoyances in debugging xterm.js against a project that uses it is that you train your users to link against `dist/xterm.js` by default. Let's say, then, that you want to test against master!

1. You try to just npm install against `sourcelair/xterm.js`. At first this seems like it succeeds but there appear to be no changes, even though you confirm that the commit ref is HEAD for sure.
2. You figure you just need to build the dist script, and that it's probably just not updated on every commit (true). You muck around with tooling a bit and manage to get a build going but you discover that the npm package doesn't bundle `gulpfile.js`, which is needed to build.
3. You blow away your package, checkout from git directly instead, and manage to get a build going. There still appear to be no changes to your client code.
4. After 5 minutes of debugging you realize that the build process is copying results into `/build` and not `/dist`! You change your application to link against `/build` and start drinking.

I think xterm.js should follow codemirror's lead and not include /dist in the sourcetree and that FURTHERMORE on install from npm, it should build itself either *on the spot or not at all*. You can either install all build tooling on any npm install (fine by me) or just require the client application have some kind of build process capable of consuming xterm's entry point. Either way, I *definitely* think stale builds should not be packaged in `/dist` in the package source tree.

Tyriar:
Publishing lib with `files` in package.json works well, I don't think we should be installing all the build deps as some take a lot of time to download. 

Creating the bower distributable repo was a good idea.

PR #422, opened 2016-12-21
Use a circular list to manage buffer

### LINKED ISSUE #361: scrollback doesn't work as expected

Tyriar:
Fixes #361

This is a work in progress, I'm putting the PR out now so the changes are easier to view and so I can get early feedback. Initial testing shows this is a lot faster in processing a large `ls -lR` as well as being responsive and showing the output as it's going (as opposed to locking up).

Remaining tasks:

- [x] BUG: Applications that change middle of buffer screw up everything (eg. `git log`) 
- [x] Make sure vim works
- [x] Ensure that all changes to the list that would trigger a change in `ybase` and `ydisp` are handled
- [x] Do thorough manual testing and ensure automated tests pass
- [x] Benchmark against current
- [x] Replace usages of `CircularList.splice` with `CircularList.shiftElements` where possible


Tyriar:
@akalipetis @parisk feel free to give this some early feedback if you have the time.

parisk:
Awesome. This looks great (code and functionality-wise).

Giving it a quick try works like a charm 👍 . Also running `ls -lR` feels **by far** more responsive.

I witnessed only mess-ups with git log / top:

If I scrolled `x` pages down in `git log` and then scrolled `y` pages up (`y`<`x`), some weird flashing happened.

👍  let's move this forward. I'd be happy to help wherever needed.

Tyriar:
I still need to do more verification but I believe the bugs are worked out. There are a few lingering TODOs still to do though. After this change we should then get rid of the majority if not all of the `CircularList.splice` calls and replace them with a much faster function(s) that sets some element and shifts `n` elements by 1 index as that seems to be its main purpose (see `reverseIndex`).

parisk:
@Tyriar a few comments on the TODO list:

> Ensure that all changes to the list that would trigger a change in ybase and ydisp are handled

This will be handled in automated tests, right?

> Do thorough manual testing [...]

Can you update this with the test cases you have in mind, in order for the rest of us to contribute as well?

> Benchmark against current

Should we consider this as *✔︎ done*, since we have the [`ls -lR ~` metric you provided](https://github.com/sourcelair/xterm.js/pull/422#issuecomment-268766435).

Tyriar:
@parisk 

> This will be handled in automated tests, right?

If possible, having had jumped into this escape sequence code I have a far better understanding of how it all works. I might follow this one with some refactors and tests, rather than jamming it all into this PR.

> Can you update this with the test cases you have in mind, in order for the rest of us to contribute as well?

Ideas right now:

- See if we can expand the automated escape sequence tests that are already present
- Play with git, vim, less, etc. when the buffer is full and not full

> Should we consider this as *✔︎ done*, since we have the [`ls -lR ~` metric you provided](https://github.com/sourcelair/xterm.js/pull/422#issuecomment-268766435).

I might see if I can write an automated perf test first, it would be nice to have something a little more solid than a timer on my phone :smile: 

I was also thinking maybe we could look at adding the PR that syncs https://github.com/sourcelair/xterm.js/pull/146 to see if we've overcome the issues that made this slow. It would be awesome if `time` was accurate, both for the end user and for testing purposes.

Tyriar:
Automated testing is probably best left for better support within xterm.js. I did some more tests and compared it to `gnome-terminal` too:

| | master | CircularList | gnome-terminal
|---|---|---|---
| Print 1..500000 | 55s | 8s | 3.5s
| ls -lR ~ (524582 files) | 68s | 29s | 6s

Tyriar:
This is ready for a full review now.

src/xterm.js
```diff
@@ -1228,16 +1229,14 @@ Terminal.prototype.showCursor = function() {
...
 Terminal.prototype.scroll = function() {
   var row;
 
-  if (++this.ybase === this.scrollback) {
-    this.ybase = this.ybase / 2 | 0;
-    this.lines = this.lines.slice(-(this.ybase + this.rows) + 1);
-  }
+  this.ybase++;
 
+  // TODO: Why is this done twice?
```
Tyriar: It's done just below, it could probably be merged with the one below.

src/utils/CircularList.ts
```diff
@@ -0,0 +1,186 @@
+/**
```
Tyriar: https://github.com/sourcelair/xterm.js/pull/422/commits/50cfe2b21a172706ea37b55176f017fcf9c3f97b

PR #435, opened 2016-12-31
Introduce build system based on Gulp

### LINKED ISSUE #359: Introduce task runner for builds
grunt or gulp will allow us to get cross-platform builds (currently Windows won't build) and simplify the build logic.

parisk:
This PR closes #359.

## Details
- Add Gulp and new dependencies to `package.json`
- Add `gulpfile.js` with two tasks:
    - `tsc`: For building TypeScript sources
    - `build`: For building and bundling JavaScript modules in a monolith
- Clean up `Dockerfile`, since `cpio` is not needed any more
- Clean up not needed dependencies from `package.json`
- Remove `bin/build`
- Update `bin/release` to use `npm run build` instead of `./bin/build`


gulpfile.js
```diff
@@ -0,0 +1,56 @@
+const browserify = require('browserify');
+const buffer = require('vinyl-buffer');
+const fs = require('fs-extra');
+const gulp = require('gulp');
+const merge = require('merge-stream');
+const source = require('vinyl-source-stream');
+const sourcemaps = require('gulp-sourcemaps');
```
Tyriar: Sourcemaps don't appear to be working on the demo.

image

`sorcery` was being used because the browserify version needs to resolve dist/xterm.js -> lib/xterm.js > src/xterm.js.

package.json
```diff
@@ -58,7 +64,7 @@
     "lint": "tslint src/**/*.ts",
     "test": "mocha --recursive ./lib",
     "build:docs": "jsdoc -c jsdoc.json",
-    "build": "./bin/build",
+    "build": "gulp build",
```
Tyriar: Unrelated but can we merge `npm run dev` into `npm start`? I was confused as to why watching wasn't happening on `npm start`. I don't think there's any point to not watch on the dev version?

gulpfile.js
```diff
@@ -0,0 +1,56 @@
...
+    basedir: 'lib',
+    debug: true,
+    entries: ['xterm.js'],
+    standalone: 'Terminal'
+  };
+  let bundle = browserify(browserifyOptions).bundle()
+        .pipe(source('xterm.js'))
+        .pipe(buffer())
+        .pipe(sourcemaps.init({
+          loadMaps: true
```
Tyriar: Might be more readable on one line:

```js
.pipe(sourcemaps.init({ loadMaps: true }))
```

package.json
```diff
@@ -38,18 +38,24 @@
     "browserify": "^13.1.0",
     "chai": "3.5.0",
     "docdash": "0.4.0",
-    "exorcist": "^0.4.0",
     "express": "4.13.4",
     "express-ws": "2.0.0-rc.1",
+    "fs-extra": "^1.0.0",
     "glob": "^7.0.5",
+    "gulp": "^3.9.1",
+    "gulp-cli": "^1.2.2",
```
Tyriar: I don't think you install this in the project, only globally.

parisk: When used in an package.json script it works. Npm adds `./node_modules/.bin` in PATH when running via `npm run ...`.

gulpfile.js
parisk: Hmm, I'll take a look. gulp-sourcemaps was supposed to do this when using [`loadMaps`](https://github.com/floridoo/gulp-sourcemaps#load-existing-source-maps).

package.json
parisk: Let's leave this for another PR to keep this as much in context as possible.

gulpfile.js
Tyriar: If this could be fixed before this was merged that would be great as I use dev tools pretty heavily for xterm.js dev.

parisk:
Status update based on @Tyriar's comment https://github.com/sourcelair/xterm.js/pull/435#discussion_r94273433 on broken source maps.

It seems that [`gulp-typescript`](https://github.com/ivogabe/gulp-typescript) with [`gulp-sourcemaps`](https://github.com/floridoo/gulp-sourcemaps) produce different source maps, compared to the ones produced by `tsc`.

The most important change I found out is that `tsc` does not inline source content into the source maps, while `gulp-sourcemaps` does.

Other than that it seems that the structure is different as well and a wild guess is that the new source maps have messed up `sourceRoot`. I am taking a deeper look and will post an update.

parisk:
@Tyriar everything should be fine now (except for the tests on Travis, which I have to fix).

Can you please give it a try and also run the tests and let me know if `npm run test` crashes on your environment as well? Runs fine on SourceLair, but crashes on Travis.

Tyriar:
I see the same issue, must be a Linux problem with the paths (it should contain `xterm.js`):

```
Error: ENOENT: no such file or directory, open '/home/daniel/dev/source
lair/lib/CompositionHelper.js.map'
```

Tyriar:
It's not working because the initial sourcemaps are wrong:

```
{"version":3,"sources":["/src/xterm.js"],"
```

Should be:

```
{"version":3,"sources":["../src/xterm.js"],"
```

parisk:
@Tyriar this should be 👌  for review. The error was caused by https://github.com/floridoo/gulp-sourcemaps/commit/fb4027a496898b8b7992ea5af9b3d707f1d82d38.

PR #436, opened 2016-12-31
Fix scroll in tmux with max scrollback

### LINKED ISSUE #434: Big output in tmux (application) mode breaks rendering
Printing lots of output in tmux (application) mode results in broken rendering

## Details
- Browser and browser version: Chrome 55
- OS version: macOS
- xterm.js version: 23472e6fb4b1def129ed6b3f77a53a5e3a16afd3

### Steps to reproduce

1. Run the demo and open it in the browser
2. Run `tmux`
3. Run `ls -lR` into xterm.js' directory

#### `tmux`
image

#### `ls -lR` (no tmux status bar)
image

Tyriar:
I believe this is related to when scrollBottom is not the last row in the viewport

Fixes #434

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -1234,6 +1234,15 @@ Terminal.prototype.showCursor = function() {
 Terminal.prototype.scroll = function() {
   var row;
 
+  // Make room for the new row in lines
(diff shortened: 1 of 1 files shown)


Tyriar:
Tested after merge :+1:

PR #438, opened 2016-12-31
Improve refresh queue

### LINKED ISSUE #280: Refresh limit system makes performance bad on Windows
The integrated terminal in VS Code has a performance issue on Windows outlined here https://github.com/Microsoft/vscode/issues/10328#issuecomment-248116656, @codec-abc suggests it's due to the limit system in the refresh function:

``` js
setTimeout(function () {
  self.refresh(start, end, false);
}, 34)
```
## Details
- Browser and browser version: vscode v1.6/electron
- OS version: Windows 10
- xterm.js version: https://github.com/sourcelair/xterm.js/commit/3c9d8b0d84e9262c013a7698aad57f02cccad2f3

Tyriar:
Use requestAnimationFrame in addition to a queue to refresh every animation
frame but only when a refresh is needed.

Fixes #280 

---

This change results in a slight degredation in performance in terms of raw numbers but improves by being far more responsive by losing minimal frames and not locking up the web page while processing the output. `requestAnimationFrame` will only fire once the callback has completed so it will drop frames when necessary.

Below tests were running `ls -lR /usr/lib`, notice the increased rendering, paint and scripting time on the after timeline.

Before:

image

After:

image

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -581,11 +581,7 @@ Terminal.prototype.open = function(parent) {
 
   // Setup loop that draws to screen
   this.queueRefresh(0, this.rows - 1);
-  function refreshLoop() {
-    self.refresh();
-    window.requestAnimationFrame(refreshLoop);
-  }
-  window.requestAnimationFrame(refreshLoop);
+  this.refreshLoop();
 
   // Initialize global actions that
   // need to be taken on the document.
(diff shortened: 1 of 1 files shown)


jerch:
Yeah the frame updates happen more often, making it feeling snappier (for the price of some more paints). IMHO this can only be improved any further by separating the websocket communication and terminal handling into a separate thread. See https://github.com/sourcelair/xterm.js/issues/150#issuecomment-229825748 and https://github.com/sourcelair/xterm.js/issues/150#issuecomment-230169015 for a comparison on my local computer with webworker. The second one with the webworker basically removes the websocket time (grey in the pie) and the terminal status updates (most of the yellow scripting time) from the main thread and adds some smaller scripting time for syncing between the threads. (Well it was based on the old code from July, might not work anymore with the new changes).

PR #440, opened 2016-12-31
Style wide characters at charWidth*2 in DOM

### LINKED ISSUE #439: Wide characters should be sized to 2 character widths
Notice how gnome-terminal sizes CJK characters to 2 character widths:

image

We currently do this mostly for fitting characters in rows only, we should space the characters visually as well.

Tyriar:
Fixes #439

---

Adds a new util class that encapsulates measuring of characters and restricts it to only be performed during at start up and after a resize (which I think should also be called when char size is changed).

Before:

image

After:

image

It appears to work when resizing using crazy sizes too:

image


jerch:
Note that some half width chars are rendered to wide with some fonts. Also fonts may differ in the wcwidth calculation for some glyphs. Thats the reason why I ended up with a precalculated dictionary of glyph widths for a critial subset of characters.

Tyriar:
`Viewport`'s usage of `CharMeasure` in this PR should actually make a significant performance impact as long running commands seem to spend ~20% time in `getBoundingClientRect` :open_mouth:

src/xterm.js
```diff
@@ -573,17 +574,21 @@ Terminal.prototype.open = function(parent) {
   this.compositionHelper = new CompositionHelper(this.textarea, this.compositionView, this);
   this.helperContainer.appendChild(this.compositionView);
 
-  this.charMeasureElement = document.createElement('div');
-  this.charMeasureElement.classList.add('xterm-char-measure-element');
-  this.charMeasureElement.innerHTML = 'W';
-  this.helperContainer.appendChild(this.charMeasureElement);
+  this.charSizeStyleElement = document.createElement('style');
```
parisk: Nice idea 👍 .

Tyriar:
Tested after merge :+1:

PR #446, opened 2017-01-03
Remove var usage from Viewport

Tyriar:
Part of #335 

---

I'll merge this after #440 is pushed and conflicts are resolved.


Tyriar:
Ready for review

PR #447, opened 2017-01-03
Sync with pty process using XOFF and XON

### LINKED ISSUE #425: Reconsider XON/XOFF PR to keep slave in sync with terminal
See if https://github.com/sourcelair/xterm.js/pull/146 performs better after https://github.com/sourcelair/xterm.js/pull/422 lands. It could also be changed to let larger chunks of data (~1 viewport page?) as that should suffice for keeping ctrl+c more responsive as well as making commands such as `time` accurate.

Tyriar:
Fixes #425

This PR adds the following:

- Implements XOFF (ctrl+S) and XON (ctrl+Q)
- Separates writes from the processing of the writes by means of a write buffer (`Terminal.writeBuffer`) and `setTimeout(...,0)` to start the write async
- XOFF is sent to the pty process when a threshold is reached so that xterm.js can catch up
- Writes are batched in separate async calls in order to prevent the possibility of overloading the UI thread. Calling them all async is bad for performance as it means that the renderer would need to catch up before every write.
- Frames are now skipped in order to save CPU time for the processing of data, not useless rendering of output that will be replaces in ~1000/60ms

Notes:

- Since things are not synced with the backing process programs such as `time` are now accurate!
- `^C` is now responsive
- There is a large diff due to an indentation change in `write` (now `innerWrite`), you only need to look at the beginning and end of the function though


Tyriar:
Here are results from the perf test in https://github.com/sourcelair/xterm.js/pull/422#issuecomment-269080082 to show how this affects perf:

| | Before perf work | CircularList | **XOFF/XON** | gnome-terminal
|---|---|---|---|---
| Print 1..500000 | 55s | 8s | **12s** | 3.5s
| ls -lR ~ (524582 files) | 68s | 29s | **22s**  | 6s

Keep in mind while there is a regression, the UI is now completely responsive (at least on my laptop with these config values) and you can SIGINT the process at any time.

Tyriar:
Also this is what an `ls -lR /usr/lib` timeline looks like which I've been using in other issues:

image

Tyriar:
Ready for review

PR #450, opened 2017-01-04
Keep row spans in an object pool to reduce garbage collection by reusing DOM nodes

### LINKED ISSUE #449: Investigate reuse of DOM nodes to reduce browser garbage collection
In timeline of https://github.com/sourcelair/xterm.js/pull/447#issuecomment-270262849 I see a lot of minor GC's, I believe this is due to `.refresh` throwing out all DOM nodes. If we keep a pool of span elements  instead of reconstructing all of them then this may go down.

image

### LINKED ISSUE #467: Better support for unicode characters
Apparently the characters that `vtop` outputs are not supported currently by xterm.js (whereas they are in hterm. We should support them!

https://github.com/MrRio/vtop

Additional context: https://github.com/zeit/hyper/issues/1275#issuecomment-271679999

Tyriar:
Fixes #449 
Fixes #467

---

This PR adds and utilizes an object pool in `Terminal.refresh` so that all the spans and text nodes are not being thrown away sometimes milliseconds later.

This introduces one change in behavior in that it wraps plain text in a span, not a text node as there seemed to be issues with adding text nodes only containing `&nbsp;`. I think this should have minimal impact on perf though.

A good illustration of it is action is the below shot of the DOM after a `for x in {1..500000}; do echo $x; done`, this would previously have generated 100-1000 DOM elements (the number is low due to refresh rate limiting and frame skipping) but you can see the IDs which are assigned in sequence are all in the double digits as the nodes are being reused:

image


Tyriar:
Testing this is a little more involved, putting on hold until the other more important perf PRs land.

Tyriar:
Ready for review

Tyriar:
This is particularly tricky to perf test. What (I think) I'm seeing is a slight degradation in overall command speed (~5%?), more DOM nodes created (which needs to be investigated) but less JS memory usage. I do think it's a valuable change, it needs more work though.

Tyriar:
This is ready to review/merge.

Update:

- The performance impacts are negligible - so close they're difficult to test effectively.
- This refactor improves the quality of the code significantly.
- Single-width unicode characters not have their width explicitly set which fixes https://github.com/sourcelair/xterm.js/issues/467 (an adoption blocker for [Hyper](https://hyper.is/)).

Tyriar:
On performance of unicode heavy apps like `vtop`, rendering and painting time goes down (presumably due to reuse of spans) but scripting time goes up (wrapping almost all chars in spans).

**Before:**

screen shot 2017-05-14 at 2 51 30 pm

**After:**

screen shot 2017-05-14 at 2 51 41 pm

src/utils/DomElementObjectPool.ts
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * @param type The DOM element type (div, span, etc.).
+   */
+  constructor(private type: string) {
+    this._type = type;
+    this._pool = [];
+    this._inUse = {};
+  }
+
+  public acquire(): HTMLElement {
```
parisk: Document this method please, with a comment above this line.

src/utils/DomElementObjectPool.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (this._pool.length === 0) {
+      element = this._createNew();
+    } else {
+      element = this._pool.pop();
+    }
+    this._inUse[element.getAttribute(DomElementObjectPool.OBJECT_ID_ATTRIBUTE)] = element;
+    return element;
+  }
+
+  public release(element: HTMLElement) {
```
parisk: Document this method please, with a comment above this line.

src/utils/DomElementObjectPool.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (!this._inUse[element.getAttribute(DomElementObjectPool.OBJECT_ID_ATTRIBUTE)]) {
+      throw new Error('Could not release an element not yet acquired');
+    }
+    delete this._inUse[element.getAttribute(DomElementObjectPool.OBJECT_ID_ATTRIBUTE)];
+    this._cleanElement(element);
+    this._pool.push(element);
+  }
+
+  private _createNew(): HTMLElement {
```
parisk: Document this method please, with a comment above this line.

src/utils/DomElementObjectPool.ts
```diff
@@ -0,0 +1,59 @@
...
+    const element = document.createElement(this._type);
+    const id = DomElementObjectPool._objectCount++;
+    element.setAttribute(DomElementObjectPool.OBJECT_ID_ATTRIBUTE, id.toString(10));
+    return element;
+  }
+
+  private _cleanElement(element: HTMLElement): void {
```
parisk: Document this method please, with a comment above this line.

src/xterm.js
```diff
@@ -760,7 +760,9 @@ Terminal.loadAddon = function(addon, callback) {
  * character width has been changed.
  */
 Terminal.prototype.updateCharSizeCSS = function() {
-  this.charSizeStyleElement.textContent = '.xterm-wide-char{width:' + (this.charMeasure.width * 2) + 'px;}';
+  this.charSizeStyleElement.textContent =
+      '.xterm-wide-char{width:' + (this.charMeasure.width * 2) + 'px;}' +
```
parisk: Non-blocking: Should we use template literals here as well?

Tyriar: :+1:

PR #453, opened 2017-01-04
Add disableStdin option

### LINKED ISSUE #452: Provide option to disable stdin
This has come through as an API request for vscode (https://github.com/Microsoft/vscode/issues/17896) and is also necessary if I want to keep xterm.js open and function after the process has shutdown (eg. https://github.com/Microsoft/vscode/issues/15583).

Proposal:

```
Terminal.getOption('disableStdin') => boolean
Terminal.setOption('disableStdin', true | false)
```

Tyriar:
Fixes #452 

---

Testing done:

1. Launch demo
2. In dev tools `term.setOption('disableStdin', true);`
3. Check input doesn't work but scrolling does
2. In dev tools `term.setOption('disableStdin', false);`
3. Check everything works

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -326,7 +326,8 @@ Terminal.defaults = {
   scrollback: 1000,
   screenKeys: false,
   debug: false,
-  cancelEvents: false
+  cancelEvents: false,
+  disableStdin: false
   // programFeatures: false,
   // focusKeys: false,
 };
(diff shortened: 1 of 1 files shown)


parisk:
Isn't this something that can be easily implemented by the consumer app, by using the attachCustomKeydownHandler method?

Tyriar:
Mouse input is also stdin, check references to `Terminal.send`.

parisk:
That's right 👍

PR #458, opened 2017-01-08
Add additional products that use xTerm

TylerJewell:
Hi xTerm team. Over the past 6 months, we have been studying xTerm as a replacement for Eclipse Che's term.js. It's taken us a bit of an effort, but the work is now done!  You can expect that Eclipse Che engineers (abot 60 contributors) will make direct contributions back to xTerm ongoing now that we are fully integrated in. We integrate xTerm alongside GWT.  As we more natively integrate the solution, we'll make additional PRs directly back to the project.

As background on the research and integration work that we did, these are the postings from our engineering team's research from the past six months. Some of these issues have already been resolved, or in the process of being resolved.

We also plan a couple of blog posts to our forums (~250K followers) about xTerm integration later in Q1 after we make our 5.0.0 announcements. I could also add Codenvy to this PR as well since it has been released with the version based upon xTerm, but with Sourcelair being the initiators we didn't want to make it seem like we were trying to be competitive. This PR is about the promotion and success of xTerm, for which we are fully committed to.

## Xtermjs 
This document is result of investigation using xtemjs ui terminal instead of current term.js. It contains analyze pluses and minuses using xtermjs (release version 2.2.3) and technical problems.  Investigation issue: https://github.com/eclipse/che/issues/3210

## Description
For now we are using our own fork of the https://github.com/chjj/term.js for user interface websocket-terminal. But actually this project is no longer maintained. So we can not get new releases from this project and we need support this script on our own. Original project contains link for a maintained fork https://github.com/sourcelair/xterm.js. This fork uses MIT license and community intensive develops this project. It has such users like: Microsoft Visual Studio Code, SourceLair, ttyd. xterm.js had already done 46 releases https://github.com/sourcelair/xterm.js/releases.

## Technical Advantages of xTerm:
* xterm.js has js tests. We have not any test for old term.js.
* We can periodically update xtem.js by new release.
* Added parameter to cancel browser events.
* Default 256 colours moved from js to css.
* Added ability to set terminal theme.
* Improved resize mehanizm (added resize event).
* Added fit.ts script to fits terminal size to original height and width of parent div.
* Fixed lost text selection from current active line(when blink is enabled).
* Improved copy/paster mehanizm.
* Linkify URL feature.
* Implement moving back and forward across words with "Alt + ←" and "Alt + →" respectively.
* Improved special key handling.
* Added addons to exdens xterm.js
* Improved mehanizm copy/paster from clipboard.
* Fixed cross platform input problems(For IPad, Iphone, MacIntel, MacPPC and so on).
* Did some work to support UTF-8 symbols.
* Implemented scrollbar.
* Fixed incorrect mouse position for application with  pseudo-graphic user interface (for example Midnight Commander).
* Fix to prevent terminal scrolling when user is looking into scrollback(similar to gnome-terminal, if the user is scrolling up to look at past output and the currently running program adds output to the terminal, the viewport of xterm.js should not scroll and interrupt what they are looking at.)
* Make right-click work on all browsers.

##  Broken Changes: 
We can move on our changes to realize copy/paste by hotkeys Ctrl + C/Ctrl +V. But xterm.js support hotkeys more common for terminal: Ctrl + Insert and Shift + Insert and maybe that's enough.

Drop support for old mouse wheel APIs: all browsers have supported the WheelEvent (onwheel) for sometime now, since Firefox does not support onmousewheel which is also non-standard but works with the standard interface, it makes sense to drop support now.

## Browser Support
Since xterm.js is typically implemented as a developer tool, only modern browsers are supported officially. Here is a list of the versions to support:
Chrome 48+
Edge 13+
Firefox 44+
Internet Explorer 11+
Opera 35+
Safari 8+

Xterm.js works seamlessly in Electron apps and may even work on earlier versions of the browsers but these are the browsers xterm.js developers strive to keep working.

## Known major bugs:
https://github.com/sourcelair/xterm.js/issues/307
https://github.com/sourcelair/xterm.js/issues/362
https://github.com/sourcelair/xterm.js/issues/348
https://github.com/sourcelair/xterm.js/issues/325 Data loss when resizing terminal (but this bug is exist in the current terminal ui in the CHE too). Community has pull request to fix this issue for xtermjs https://github.com/sourcelair/xterm.js/pull/404 .
 
Additional information: 

Code base xterm.js consist of files written on native javascript and typescript. For manage js dependency and configuration used npm, typings, bower and node. Development tendency: rewrite xtemr.js completely on the Typescript(information about this included to the release notes 2.2.3).

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Xterm.js is used in several world-class applications to provide great terminal e
 - [**Microsoft Visual Studio Code**](http://code.visualstudio.com/): Modern, versatile and powerful open source code editor that provides an integrated terminal based on xterm.js
 - [**ttyd**](https://github.com/tsl0922/ttyd): A command-line tool for sharing terminal over the web, with fully-featured terminal emulation based on xterm.js
 - [**Katacoda**](https://www.katacoda.com/): Katacoda is an Interactive Learning Platform for software developers, covering the latest Cloud Native technologies.
+- [**Eclipse Che**](http://www.eclipse.org/che): Developer workspace server, cloud IDE, and Eclipse next-generation IDE
 
 
 Do you use xterm.js in your application as well? Please [open a Pull Request](https://github.com/sourcelair/xterm.js/pulls) to include it here. We would love to have it in our list.
@@ -116,6 +117,10 @@ Visit https://lair.io/sourcelair/xterm and follow the instructions. All developm
 #### [Visual Studio Code](http://code.visualstudio.com/)
 
 [Download Visual Studio Code](http://code.visualstudio.com/Download), clone xterm.js and you are all set.
+
+#### [Eclipse Che](http://www.eclipse.org/che)
+
+You can start Eclipse Che with `docker run eclipse/che start`
 
 ## License Agreement
 


Tyriar:
Sounds great @TylerJewell, more contributors are always welcome!

Here are some notes on your investigation:

- **Linkify URL feature.** This has been broken ever since I started contributing, I did however just write up a proposal to get it implemented in a good way https://github.com/sourcelair/xterm.js/issues/455
- **Copy and paste** In electron clients you can control this completely, it's only an issue in the browser. @parisk knows more on this
- We have a bunch of performance changes currently in review that fixes all UI blocking issues in addition to making processing of output and rendering *a lot* faster.

I'll leave this open for @parisk to view/merge.

parisk:
Thanks a lot for your contribution @TylerJewell, looks great. Also, please feel free to add Codenvy to the list as well.

As @Tyriar mentioned:

1. the linkify addon is broken since almost the start of time 😅 , but we will take care of this soon, as it's a beloved feature
2. the clipboard has indeed issues in the browser, but this is something I plan to improve in this release

Two more cents I'd like to add are that within the next few (1-2) releases you should also expect:

1. The code base converted completely to TypeScript
2. A public and stable add-on API

TylerJewell:
Thank you, @parisk and @Tyriar - I will link in the various engineers that are doing our xTerm work and keep them apprised.  I will issue another PR for the Codenvy addition.

Issue #460, opened 2017-01-08
Pull escape sequences into their own named file


Tyriar:
Organize escape sequences constants into a new EscapeSequences.ts file that gives them names and centralizes the values.

TODO:

- [x] C0 codes https://github.com/sourcelair/xterm.js/pull/461
- [x] Replace all references to C0 codes in xterm.js https://github.com/sourcelair/xterm.js/pull/461
- [ ] CSI codes
- [ ] Replace all references to CSI codes in xterm.js
- [ ] OSC codes
- [ ] Replace all references to OSC codes in xterm.js

Tyriar:
The current layout is good

PR #461, opened 2017-01-08
Pull C0 escape sequences into its own file

Tyriar:
Part of #460 

---

This is the plan for #460 but to apply it to CSI and OSC as well. Once we move the relevant code to TS we'll be able to use template literals to make the code very readable. Example:

```js
result.key = '\x1b[1;' + (modifiers + 1) + 'C';
```

Would become something like:

```ts
result.key = `${CSI.START}1;${modifiers + 1}${CSI.CUF}`;
```

Or potentially:

```ts
result.key = csiCode(CSI.CUF, '1', modifiers + 1);
```

We can also look into having TS inline the constants: https://github.com/Microsoft/TypeScript/issues/3922 https://github.com/Microsoft/TypeScript/issues/3192

@parisk let me know what you think of the structure.


Tyriar:
I'm beginning to reconsider this after spending some time redoing the parser in https://github.com/sourcelair/xterm.js/pull/462

Having C0 codes are definitely useful as they're nonsensical without it (eg. \x08 is backspace). CSI codes however are more difficult to name as they have multiple names, the information on them are probably best located within the InputHandler and the Parser itself. The CSI final letter is the best way to identify the behavior.

parisk:
What if we use methods that will get parameters instead of attributes for CSI?

Or what if each code had multiple "children" and acted like a map?

Would such a thing work?

Tyriar:
I think that's probably overkill. Check out the parser PR for what I'm thinking now, the descriptive comments and csi code names move with the implementations to InputHandler. 

Also remember if it's a map they're not inlined which could cause some significant bloat when dealing with so many codes.

PR #464, opened 2017-01-09
Add Codenvy to the list of xTerm supporters

TylerJewell:
This is the follow-on PR per @Tyriar and @parisk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,8 @@ Xterm.js is used in several world-class applications to provide great terminal e
 - [**Microsoft Visual Studio Code**](http://code.visualstudio.com/): Modern, versatile and powerful open source code editor that provides an integrated terminal based on xterm.js
 - [**ttyd**](https://github.com/tsl0922/ttyd): A command-line tool for sharing terminal over the web, with fully-featured terminal emulation based on xterm.js
 - [**Katacoda**](https://www.katacoda.com/): Katacoda is an Interactive Learning Platform for software developers, covering the latest Cloud Native technologies.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,7 +121,11 @@ Visit https://lair.io/sourcelair/xterm and follow the instructions. All developm
 
 #### [Eclipse Che](http://www.eclipse.org/che)
 
-You can start Eclipse Che with `docker run eclipse/che start`
+You can start Eclipse Che with `docker run eclipse/che start`.
```
Tyriar: @parisk should we remove the "Known tools that use xterm.js internally" list as it pretty much duplicates the real world uses?

TylerJewell:
It did strike me as odd that there were two sections - I couldn't really differentiate the meaning of the two.

Tyriar:
I initially thought it was:

- Here are products that use xterm.js
- Here are tools you can use to develop it

The second section isn't really necessary though imo.

TylerJewell:
Just to keep things a bit more organized, with Eclipse Che, we have had to move anything related to making a contribution into either a CONTRIBUTION.md file (the highlights) and if you are an active developer of the project, we move all of the development workflow into a series of documents contained within the project's wiki.  It's harder to maintain, but it keeps the initial README a bit simpler to get your arms around, and then have it more focused on the end user than a project contributor.

parisk:
Yep, now that it started getting bigger, I think it pretty much looks like a duplicate.

I will move forward with merging it and I will open up a new PR to rephrase things.

Thanks @TylerJewell for your contribution!

PR #466, opened 2017-01-10
Perform CharMeasure.measure async

### LINKED ISSUE #465: Sometimes initial CharMeasure.measure() does not give accurate measurements
See https://github.com/Microsoft/vscode/issues/18370

It appears that calling `getBoundingClientRect` directly after an `appendChild` does not work _sometimes_. It's probably better from a perf perspective to run this async anyway and never remove the element.

Tyriar:
This fixes the edge case where getBoundingClientRect was returning a width and
height of 0,0.

Fixes #465

### DIFF
--- a/src/utils/CharMeasure.ts
+++ b/src/utils/CharMeasure.ts
@@ -28,24 +28,29 @@ export class CharMeasure extends EventEmitter {
   }
 
   public measure(): void {
-    const oldWidth = this._width;
-    const oldHeight = this._height;
-
     if (!this._measureElement) {
       this._measureElement = document.createElement('span');
       this._measureElement.style.position = 'absolute';
       this._measureElement.style.top = '0';
       this._measureElement.style.left = '-9999em';
       this._measureElement.textContent = 'W';
+      this._parentElement.appendChild(this._measureElement);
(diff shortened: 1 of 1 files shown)


src/utils/CharMeasure.ts
```diff
@@ -28,24 +28,29 @@ export class CharMeasure extends EventEmitter {
...
     if (!this._measureElement) {
       this._measureElement = document.createElement('span');
       this._measureElement.style.position = 'absolute';
       this._measureElement.style.top = '0';
       this._measureElement.style.left = '-9999em';
       this._measureElement.textContent = 'W';
+      this._parentElement.appendChild(this._measureElement);
+      // Perform _doMeasure async if the element was just attached as sometimes
+      // getBoundingClientRect does not return accurate values without this.
+      setTimeout(() => this._doMeasure(), 0);
```
parisk: Should we use 4ms which [is the default for HTML5](https://developer.mozilla.org/en-US/docs/Web/API/WindowTimers/setTimeout#Reasons_for_delays_longer_than_specified)?

Tyriar: 0 should be fine as it's handled by the browser, it's also the most common way to do it. I'd personally be a bit confused if it was 4.

PR #479, opened 2017-01-14
Apply cursorBlink via setOption

### LINKED ISSUE #478: Setting cursorBlink needs a refresh before applying
1. Open demo
2. Toggle cursor blink option, notice nothing happens as there is no refresh

Tyriar:
Fixes #478


parisk:
What if we changed `cursorBlink` to work purely with CSS (and CSS animations)?

This way we will avoid redrawing.

Tyriar:
We do control it completely with CSS. Now that you say that we could grab the cursor element if it's available and apply the class to it. Not sure it's worth duplicating the cursor logic since we're only refreshing a single line which is a pretty fast operation anyway.

parisk:
We can set the CSS class on `term.element` and add the appropriate CSS attributes based on the following-rule (or sth like that): `.xterm.xterm-cursor-blink .terminal-cursor`.

Tyriar:
I saw there is an escape sequence which handles changing the cursor style. I wonder if it would be useful to support that? Maybe existing config files could set it then?

parisk:
I think we should support it via CSS as well.

parisk:
Works great. Also tests have passed, but have not been reported: https://travis-ci.org/sourcelair/xterm.js/builds/191852761.

Only thing needed is a rebase with master.

Issue #487, opened 2017-01-17
Come up with good story for custom (non-xterm) keybindings support


Tyriar:
We currently support things like alt+arrow to jump words which is done by overriding what escape sequence we send to the pty. We should look at making this configurable with safe defaults so that consumer(s) can turn it off/configure if they desire.

parisk:
This could be a `keyMap` like option, where we can map either key bindings or escape codes to terminal actions.

How does that sound?

Tyriar:
Something like that is what I was thinking, however it gets tricky in the case like `ctrl+v` where I think the escape sequence we generate is the same as `v`

Tyriar:
This is also the reason we can't use an `~/.inputrc` file to get option+backspace to delete a word on Mac.

mofux:
I've been thinking a lot about this recently, here is my idea (considering the many issues open on hyper related on this topic):

The main problem is, that accelerators of the app (e.g. hyper) collide with accelerators that the app running in the terminal wants to consume. To get around this problem, most apps introduce a meta key that helps to escape from the terminal. While this works in most cases, it is uncomfortable to use, because you need three fingers. So why don't we...

**Use a keyboard gesture to escape the terminal**
The idea is that if you double-tap a modifier key (ctrl, alt, shift, meta?), then the key will not be sent to the terminal, but will be forwarded to user, so his app can consume it. Without the double-tap, the key(s) will be consumed by the terminal.

Example: `ctrl...ctrl+x` or `alt...alt+left` will not be consumed by the terminal but forwarded as `ctrl+x` and `alt+left` to the user.

In addition, for the very commonly used `ctrl+c` we should let it copy if text is selected, otherwise we should write the control sequence to the terminal.

Pasting can be done with `ctrl...ctrl+v`, middle or right mouse button.

Tyriar:
@mofux for VS Code I introduced a `commandsToSkipShell` setting which lists off a bunch of commands whose keybindings will not be processed by xterm.js. So you should be able to intercept cases like this using `Terminal.attachCustomKeydownHandler` and handle it however you want at a higher level (ctrl+c + text selected -> copy, ctrl+c + no text selected -> send to xterm.js).

My personal thinking is this is more about having xterm.js handle the common keybindings that are not standard but generally desirable, but allowing consumers to opt out and/or handle it themselves.

mofux:
[...]

PR #491, opened 2017-01-18
Allow pasting on Windows

BenHall:
This appears to allow Ctrl+V pasting to work on Windows.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -2839,6 +2839,10 @@ Terminal.prototype.evaluateKeyEscapeSequence = function(ev) {
     default:
       // a-z and space
       if (ev.ctrlKey && !ev.shiftKey && !ev.altKey && !ev.metaKey) {
+        if (ev.keyCode == 86) {
+          // console.log("PASTING!!!!");
+          break;
+        }
         if (ev.keyCode >= 65 && ev.keyCode <= 90) {
           result.key = String.fromCharCode(ev.keyCode - 64);
         } else if (ev.keyCode === 32) {


src/xterm.js
```diff
@@ -2839,6 +2839,10 @@ Terminal.prototype.evaluateKeyEscapeSequence = function(ev) {
     default:
       // a-z and space
       if (ev.ctrlKey && !ev.shiftKey && !ev.altKey && !ev.metaKey) {
+        if (ev.keyCode == 86) {
```
Tyriar: I believe this would make ctrl+v never send the v key which is may not be desirable. VS Code for example allows users to customize their copy/paste keys which are intercepted before this is hit and the event is then canceled.

I just filed this which is closely related this this issue of keybinding configuration https://github.com/sourcelair/xterm.js/issues/487

@parisk would it make sense to always cancel the keydown event for ctrl+v when in a non-node/electron environment? Or should this be more configurable?

BenHall: My first attempt, when it was at the top of the CASE, did block v as you described. However, this approach appears to allow it. 

This way feels a little hacky, but not having Ctrl+V working on Windows caused us headaches. #487 might be a cleaner approach.

Tyriar: This would also override ctrl+v on Mac (cmd+v is typically the default) as well as on Linux (where ctrl+shift+v is typically the default). We could target just Windows but I think it should be somewhat configurable so that it's not forced upon consumers that don't want it (such as VS Code's case where paste's keybinding is configurable).

parisk:
@BenHall are you 🆗  with putting this on hold until ew conclude on #487? I think it is related and might be a neater solution.

BenHall:
Of course. This was something which worked for us and got us out of a problem. If there is a cleaner way coming, great!

Tyriar:
Going to close this off as the fix would need to come on a new branch anyway once #487 is done.

PR #492, opened 2017-01-18
Implement CSI Ps SP q Set cursor style

### LINKED ISSUE #480: Support set cursor style (DECSCUSR)
We should support the following CSI code to allow more fine-grained control of the cursor and also allow configuration via `.bashrc` (eg. adding `echo -e "\e[5q"` for blinking bar)

```
CSI Ps SP q
          Set cursor style (DECSCUSR, VT520).
            Ps = 0  -> blinking block.
            Ps = 1  -> blinking block (default).
            Ps = 2  -> steady block.
            Ps = 3  -> blinking underline.
            Ps = 4  -> steady underline.
            Ps = 5  -> blinking bar (xterm).
            Ps = 6  -> steady bar (xterm).
```

Both `xterm` and `gnome-terminal` support this out of the box. This can be done after the Parser refactor is complete https://github.com/sourcelair/xterm.js/pull/462.

Tyriar:
Fixes #480 

cursor-styles

### DIFF
--- a/demo/style.css
+++ b/demo/style.css
@@ -20,7 +20,3 @@ h1 {
     color: #fafafa;
     padding: 2px;
 }
-
-#terminal-container .terminal:focus .terminal-cursor {
-    background-color: #fafafa;
-}
--- a/src/InputHandler.test.ts
+++ b/src/InputHandler.test.ts
@@ -1,0 +1,49 @@
+import { assert } from 'chai';
+import { InputHandler } from './InputHandler';
+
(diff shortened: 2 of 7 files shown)


Tyriar:
Testing is so easy now https://github.com/sourcelair/xterm.js/pull/492/commits/9227aed90607820c5ca4ce92228a38ec0ad310ac

We just need to bring in phantomjs to make it even nicer!

PR #493, opened 2017-01-19
Convert xterm.js to TypeScript!

Tyriar:
Part of #335

This PR does the following:

- Moves contents of xterm.js to Terminal.ts
- Expose entrypoint at xterm.ts
- Convert Terminal to a class
- Moved old 'static' `Terminal.member` to functions/variables scoped to file
- Replaced the majority of usages of `Terminal[optionsKey]` with `Terminal.options[optionsKey]`
- The type `any` is used in many places to get the conversion over with quickly, these can be refined later

Verification done:

Testing done in Chrome 51 and Firefox 50
- [x] npm test
- [x] ls -lR
- [x] Scrolling via mouse/keyboard
- [x] tmux input/split/scroll
- [x] vim input/scroll

### DIFF
--- a/src/EventEmitter.ts
+++ b/src/EventEmitter.ts
@@ -53,12 +53,11 @@ export class EventEmitter {
     return this.on(type, on);
   }
 
-  public emit(type): void {
+  public emit(type, ...args: any[]): void {
     if (!this._events[type]) {
       return;
     }
 
-    let args = Array.prototype.slice.call(arguments, 1);
     let obj = this._events[type];
 
     for (let i = 0; i < obj.length; i++) {
--- a/src/Interfaces.ts
+++ b/src/Interfaces.ts
@@ -2,6 +2,20 @@
  * @license MIT
  */
 
+export interface ICircularList<T> {
+  length: number;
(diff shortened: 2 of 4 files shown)


parisk:
Great! I got two proposals before moving on a full review:

1. We wait to merge all hanging PRs first, before dealing with this, since it's the biggest one.
2. Keep `xterm.ts` as `xterm.js` just for reference to the project's name. The syntax looks like it can stay exactly the same.

Tyriar:
Sounds good, the line wrap PR is the main thing that needs to move before this to prevent some merge headaches.

hsk:
Nice too meet you.
I worked without noticing pullrequest.
Please refer if you like.

https://github.com/hsk/xterm.js/blob/master/src/xterm.ts

This work difference strict types than your Convert.

Tyriar:
@hsk the reason this wasn't merged is because it's a large diff and there were (and still are) community PRs out that would cause bit headaches for contributors. Pulling things out of xterm.js slowly is probably the better approach but probably not until https://github.com/sourcelair/xterm.js/pull/609 is merged

PR #495, opened 2017-01-20
Only keep > 0 geometry in CharMeasure

### LINKED ISSUE #494: Resizing a terminal that is not being displayed will break the viewport/scroll area
See https://github.com/Microsoft/vscode/issues/18332

This is happening because the terminal container is being hidden which causes `CharMeasure` to return 0 which puts the viewport into an invalid state. There should be a reasonable way of doing this to prevent obscure bugs like above.

Tyriar:
Fixes #494

### DIFF
--- a/src/utils/CharMeasure.ts
+++ b/src/utils/CharMeasure.ts
@@ -44,10 +44,12 @@ export class CharMeasure extends EventEmitter {
   }
 
   private _doMeasure(): void {
-    const oldWidth = this._width;
-    const oldHeight = this._height;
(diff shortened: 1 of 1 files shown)


Tyriar:
This fixes the major issue of the viewport size completely breaking, there is one remaining issue though. When you show the element after the resize, if the resize was smaller then the scroll bar will still be out of sync. This requires a manual `term.emit('scroll')` which is fine. I created https://github.com/xtermjs/xtermjs.org/pull/12 to document this.

PR #497, opened 2017-01-20
Implement configurable tab stop width

### LINKED ISSUE #488: Option for configurable tab width?
`xterm.js` inherits from `term.js` the [default tab width of 8 characters](https://github.com/sourcelair/xterm.js/blob/ed13218f9098a997283d996e7570d5cdb24ed87f/src/xterm.js#L3146):

```js
for (; i < this.cols; i += 8) {
  this.tabs[i] = true;
}
```

I think it would be very useful, especially for space-constrained web applications, to be able to set this width to a lower setting.

parisk:
This PR introduces configurable tab stop width.

## Current issues
- [x] No way to call `term.setupStops()` after setting the option, without a hack into the private API. A private `_optionChanged` event could definitely solve this. Ideas @Tyriar?
- [ ] There is no easy way to re-render already printed tab stops. In order to achieve this we will have to separate the data from the presentation layer and re-render the whole thing, which is a big issue.

Documentation PR: xtermjs/xtermjs.org#13

Closes #488.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -17,13 +17,13 @@
         <div>
           <h2>Options</h2>
           <p>
-            <label>cursorBlink <input type="checkbox" id="option-cursor-blink"></label>
+            <label><input type="checkbox" id="option-cursor-blink"> cursorBlink</label>
           </p>
           <p>
             <label>scrollback <input type="number" id="option-scrollback" value="1000" /></label>
           </p>
           <p>
-            <label>tabStopWidth <input type="number" id="option-tabstopwidth" value="4" /></label>
+            <label>tabStopWidth <input type="number" id="option-tabstopwidth" value="8" /></label>
           </p>
           <div>
           	<h3>Size</h3>
--- a/demo/index.html
+++ b/demo/index.html
@@ -17,10 +17,13 @@
         <div>
           <h2>Options</h2>
           <p>
-            <label><input type="checkbox" id="option-cursor-blink"> cursorBlink</label>
+            <label>cursorBlink <input type="checkbox" id="option-cursor-blink"></label>
           </p>
           <p>
-            <label>Scrollback <input type="number" id="option-scrollback" value="1000" /></label>
+            <label>scrollback <input type="number" id="option-scrollback" value="1000" /></label>
+          </p>
+          <p>
+            <label>tabStopWidth <input type="number" id="option-tabstopwidth" value="4" /></label>
           </p>
           <div>
           	<h3>Size</h3>
--- a/demo/main.js
+++ b/demo/main.js
@@ -16,8 +16,8 @@ var terminalContainer = document.getElementById('terminal-container'),
     rowsElement = document.getElementById('rows');
 
 function setTerminalSize () {
-  var cols = parseInt(colsElement.value),
-      rows = parseInt(rowsElement.value),
+  var cols = parseInt(colsElement.value, 10),
+      rows = parseInt(rowsElement.value, 10),
       width = (cols * charWidth).toString() + 'px',
       height = (rows * charHeight).toString() + 'px';
 
@@ -36,7 +36,7 @@ optionElements.scrollback.addEventListener('change', function () {
   term.setOption('scrollback', parseInt(optionElements.scrollback.value, 10));
 });
 optionElements.tabstopwidth.addEventListener('change', function () {
-  term.setOption('tabStopWidth', parseInt(optionElements.tabstopwidth.value));
+  term.setOption('tabStopWidth', parseInt(optionElements.tabstopwidth.value, 10));
 });
 
 createTerminal();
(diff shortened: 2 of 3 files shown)


Tyriar:
What does `setupStops` do exactly? We can call it from within `getOption` like is done to fix `scrollback`?

parisk:
`setupStops` defines the position in the terminal where tab stops are located.

I added it into `setOption` just like we do with `cursorBlink`, although I wasn't quite fond of the idea at first.

parisk:
@Tyriar I rebased with master and resolved conflicts. This is ready for review.

demo/index.html
```diff
@@ -17,10 +17,13 @@
         <div>
           <h2>Options</h2>
           <p>
-            <label><input type="checkbox" id="option-cursor-blink"> cursorBlink</label>
+            <label>cursorBlink <input type="checkbox" id="option-cursor-blink"></label>
```
Tyriar: Normally checkboxes are at the start of the line?

demo/index.html
```diff
@@ -17,10 +17,13 @@
...
           </p>
           <p>
-            <label>Scrollback <input type="number" id="option-scrollback" value="1000" /></label>
+            <label>scrollback <input type="number" id="option-scrollback" value="1000" /></label>
+          </p>
+          <p>
+            <label>tabStopWidth <input type="number" id="option-tabstopwidth" value="4" /></label>
```
Tyriar: value=8 as the default?

demo/main.js
```diff
@@ -32,7 +33,10 @@ optionElements.cursorBlink.addEventListener('change', function () {
   term.setOption('cursorBlink', optionElements.cursorBlink.checked);
 });
 optionElements.scrollback.addEventListener('change', function () {
-  terminal.setOption('scrollback', parseInt(optionElements.scrollback.value, 10));
+  term.setOption('scrollback', parseInt(optionElements.scrollback.value, 10));
+});
+optionElements.tabstopwidth.addEventListener('change', function () {
+  term.setOption('tabStopWidth', parseInt(optionElements.tabstopwidth.value));
```
Tyriar: Add radix of 10 as second arg as it's a mandatory arg https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

demo/main.js
```diff
@@ -44,7 +48,8 @@ function createTerminal() {
   }
   term = new Terminal({
     cursorBlink: optionElements.cursorBlink.checked,
-    scrollback: parseInt(optionElements.scrollback.value, 10)
+    scrollback: parseInt(optionElements.scrollback.value, 10),
+    tabStopWidth: parseInt(optionElements.tabstopwidth.value)
```
Tyriar: Radix arg

src/xterm.js
```diff
@@ -433,6 +434,7 @@ Terminal.prototype.setOption = function(key, value) {
       this.element.classList.toggle(`xterm-cursor-style-underline`, value === 'underline');
       this.element.classList.toggle(`xterm-cursor-style-bar`, value === 'bar');
       break;
+    case 'tabStopWidth': this.setupStops(); break;
```
Tyriar: Eventually we could move this to event listeners which would probably be a better structure.

Tyriar:
Changing tab will likely break `ls` output? How does `ls` fetch the terminal's tab size? Related: https://github.com/sourcelair/xterm.js/issues/259

demo/index.html
parisk: Did this to keep uniformity with the rest. Will revert.

demo/index.html
parisk: Nah, forgot it from testing.

src/xterm.js
parisk: Yep. It does the job for now, so let's keep this for another PR.

Tyriar:
Looking at the tab stuff for https://github.com/sourcelair/xterm.js/issues/259, I don't think re-rendering would actually be desirable. There is the issue of how programs get the tab size though.

parisk:
Everything seems to be working fine. I think that re-rendering is an open question and that we can revisit it in the future.

`ls` looks to be working just fine. I am pretty sure though that it is using spaces and not tabs.

## Screnshot
image

parisk:
I also fixed comments in demo, so @Tyriar feel free to take another look.

Tyriar:
It was certainly tabs on my Macbook, I'll grab your branch and verify.

Tyriar:
This is what I see on a dir with a bunch of files of varying lengths. The first is default (8), the second is 4 via `setOption`:

screen shot 2017-01-20 at 2 49 47 pm

Tyriar:
This also happens on Terminal.app, maybe this is just a price you pay?

Tyriar:
You can fix this on linux by doing something like:

```
alias ls=ls --tabsize=4
```

I guess that's just how it is 😄

parisk:
Well, while this view sucks, technically there is nothing wrong about it. Take a close look and you will see that each file name is being printed at the first available tab stop.

The tab stop width though is not available somehow to `ls` in order to take care of properly aligning everything. I guess this is where you have to use `ls --tabsize=4`.

My opinion is that this is the price you have to pay eventually for such a feature and that you should be careful with using it.

I guess it has it's use cases though.

Tyriar:
Yeah, this is the behavior of other terminal emulators that set via calling `tabs` so :+1:

vincentwoo:
Sorry was late to the party. I agree that tab stop width changes should *not* rerender previous content. Otherwise, I have no comments!

PR #499, opened 2017-01-21
Add support for rest of VT100 character sets

Tyriar:
This PR adds British, Dutch, Finnish, French, French Canadian, German, Italian, Norwegian/Danish, Spanish, Swedish and Swiss character sets. They can be turn on using `echo -e '\x1b(A'` where A is the language code, this example will turn on the UK character set, changing any `#` to `£`.

### DIFF
--- a/src/Charsets.ts
+++ b/src/Charsets.ts
@@ -14,14 +14,16 @@ export const CHARSETS: {[key: string]: {[key: string]: string}} = {};
  */
 export const DEFAULT_CHARSET = CHARSETS['B'];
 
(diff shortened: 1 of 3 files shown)


parisk:
Great, maybe we can create an example page in xtermjs/xtermjs-org to document how a consumer app can change the charset.

parisk:
Should we also expose a public API for setting charset?

Tyriar:
I don't think so, this seems like more of a legacy thing since changing keyboars layouts works just fine without this. This just allows programs to output certain Unicode characters without actually using Unicode if they desire.

PR #500, opened 2017-01-21
Fix many vt tests

Tyriar:
This PR fixes many of the tests in escape-sequence-tests.js, see the commits for details on the individual changes. Previously 10 tests were passing, now 30+ are.

The major fixes:

- `CAN` has been implemented which cancels any escape sequence, the corresponding test does not work still due to complex ignore logic in xterm that we don't have parity with
- Index, LF/VT/FF, CUB, CUD, VPR behavior was change to stay at the end of the column if it was otherwise going to go to wraparound
- When wraparound mode is disabled output will be cut off the end, no longer will the last character always be displayed in the last column
- There was an error which cause causing our tests to compare only the first page of the buffer, not the viewport.
- Expected output was trimmed as we don't care, at least not currently (we trim all output from xterm.js anyway).

There is a failing test:

```
1) xterm.js unicode - surrogates 2 characters per cell over line end without autowrap:

      AssertionError: expected 'a' to deeply equal '𐀀'
      + expected - actual

      -a
      +𐀀
```

This is likely because it relied on the previous wraparound disabled logic which I believe was wrong. @jerch thoughts since you added it?

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -123,6 +123,9 @@ export class InputHandler implements IInputHandler {
       this._terminal.y--;
       this._terminal.scroll();
     }
+    if (this._terminal.x >= this._terminal.cols) {
+      this._terminal.x--;
+    }
   }
 
   /**
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -220,6 +220,9 @@ export class InputHandler implements IInputHandler {
     if (this._terminal.y >= this._terminal.rows) {
       this._terminal.y = this._terminal.rows - 1;
     }
+    if (this._terminal.x >= this._terminal.cols) {
+      this._terminal.x--;
+    }
   }
 
   /**
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -246,6 +246,9 @@ export class InputHandler implements IInputHandler {
     if (param < 1) {
       param = 1;
     }
+    if (this._terminal.x >= this._terminal.cols) {
+      this._terminal.x--;
+    }
     this._terminal.x -= param;
     if (this._terminal.x < 0) {
       this._terminal.x = 0;
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -701,6 +701,9 @@ export class InputHandler implements IInputHandler {
     if (this._terminal.y >= this._terminal.rows) {
       this._terminal.y = this._terminal.rows - 1;
     }
+    if (this._terminal.x >= this._terminal.cols) {
+      this._terminal.x--;
+    }
   }
 
   /**
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -123,6 +123,7 @@ export class InputHandler implements IInputHandler {
       this._terminal.y--;
       this._terminal.scroll();
     }
+    // If the end of the line is hit, prevent this action from wrapping around to the next line.
     if (this._terminal.x >= this._terminal.cols) {
       this._terminal.x--;
     }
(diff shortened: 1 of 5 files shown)


src/InputHandler.ts
```diff
@@ -124,6 +123,9 @@ export class InputHandler implements IInputHandler {
       this._terminal.y--;
       this._terminal.scroll();
     }
+    if (this._terminal.x >= this._terminal.cols) {
```
parisk: What is the role of these lines here?

parisk: Can you add a line comment please?

parisk: Seems like you are shifting things up when you are inserting a line feed, when you are in the bottom line?

src/InputHandler.ts
```diff
@@ -218,6 +220,9 @@ export class InputHandler implements IInputHandler {
     if (this._terminal.y >= this._terminal.rows) {
       this._terminal.y = this._terminal.rows - 1;
     }
+    if (this._terminal.x >= this._terminal.cols) {
```
parisk: Can you add a line comment of what this if clause does please? Seems identical to https://github.com/sourcelair/xterm.js/pull/500/files#r97194377.

src/InputHandler.ts
```diff
@@ -244,6 +249,9 @@ export class InputHandler implements IInputHandler {
     if (param < 1) {
       param = 1;
     }
+    if (this._terminal.x >= this._terminal.cols) {
```
parisk: Can you add a line comment of what this if clause does please? Seems identical to https://github.com/sourcelair/xterm.js/pull/500/files#r97194377.

src/InputHandler.ts
```diff
@@ -693,6 +701,9 @@ export class InputHandler implements IInputHandler {
     if (this._terminal.y >= this._terminal.rows) {
       this._terminal.y = this._terminal.rows - 1;
     }
+    if (this._terminal.x >= this._terminal.cols) {
```
parisk: Can you add a line comment of what this if clause does please? Seems identical to https://github.com/sourcelair/xterm.js/pull/500/files#r97194377.

src/xterm.js
```diff
@@ -2303,6 +2303,9 @@ Terminal.prototype.index = function() {
     this.y--;
     this.scroll();
   }
+  if (this.x >= this.cols) {
```
parisk: Can you add a line comment of what this if clause does please? Seems identical to https://github.com/sourcelair/xterm.js/pull/500/files#r97194377.

parisk:
This is the test case failing: https://github.com/sourcelair/xterm.js/blob/4d5046b5e7047bfcf1f1dba088acb343df1969bb/src/test/test.js#L598

Not sure which of these three `expect`s fail. I think that you can just uncomment all your changes in inputhandler, parser and xterm and then by putting one piece at a time back in you will be able to find what broke this test.

parisk:
Any way I can help?

Tyriar:
It's the first expect and I think it's to do with https://github.com/Tyriar/xterm.js/commit/43c3349494c7b922577fa80d50a57b23896e0818

I believe it's verifying the wrong behavior.

jerch:
@Tyriar Yeah the test checks that with `wraparoundMode = false` surrogate chars correctly end up the last cell. Since you changed that behavior to truncating the line, the test is not applicable anymore.
(Btw truncating the line is not conform to the old VTs, they were not able to move the cursor beyond the display, thats why they simply inserted the char at the last cell)

Tyriar:
@jerch thanks for the clarification, I'll pull the test out :smile:

Tyriar:
@parisk good to look at again

PR #505, opened 2017-01-24
Add CoderPad to real world use cases

vincentwoo:
Hi everyone,

I'm very pleased to say I have cut CoderPad.io over to using xterm.js (replacing term.js). I'm very pleased with the work you've done on it. You've implemented the many terrible hacks on top of term.js I had to write, except you've actually implemented stuff the right way.

I look forward to your continued progress on the project!

Cheers,
Vincent


Tyriar:
Great to hear! It's also nice to see it being used for different use cases - and it's obvious now why tab width was needed 😄

parisk:
Thanks a lot @vincentwoo! We are glad to have you on the list 😄 .

PR #509, opened 2017-01-25
Support debugging mocha tests in vscode

Tyriar:
This file allows debugging of the mocha tests in vscode, this will be super helpful particularly when fixing the vt tests.

image

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,23 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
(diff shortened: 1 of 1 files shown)


Tyriar:
It will only work if it's located at `<opened folder>/.vscode/launch.json` and it's a dot file so it will be hidden by default on Unix-like OS'. It's quite an inconvenience to copy and paste it like that. Note that we're still ignoring the directory, just not this file (which shouldn't change automatically).

PR #513, opened 2017-01-27
Add useFlowControl option

### LINKED ISSUE #511: Paste doesn't work on oh-my-zsh under macOS
This appears to be a recent regression. Pasting behaves really strangely.

See https://github.com/Microsoft/vscode/issues/18833

Tyriar:
Fixes #511


mofux:
Out of curiosity, what was the reason to disable flow control by default?

mofux:
Found the answer in #511, seems reasonable.

Issue #164, opened 2016-07-06
Uncaught TypeError: Cannot set property '0' of undefined


tqlihuiqi:
execute vi / vim command throw exception

error:
        Uncaught TypeError: Cannot set property '0' of undefined

debug:
        Serial port requested application keypad.

code:
        <script src="/static/plugins/xterm/src/xterm.js"></script>
        <link rel="stylesheet" href="/static/plugins/xterm/src/xterm.css">

```
    var currentWs = new ReconnectingWebSocket('ws://{{ wsServer }}/ws/terminal/current');
    currentWs.onopen =  function(event) {
        var xterm = new Terminal({
            geometry: [98, 26],
            debug: true
        });
        xterm.open(document.getElementById('currentRecord'));
        currentWs.onmessage = function(event) {
            xterm.write(event.data);
        }; 
    }
```

akalipetis:
Hello @tqlihuiqi - thanks for reporting.

Are you testing on the latest master branch? Also, in which browser you're testing?
I just tried running `vi`/`vim` on the Xterm demo but could not reproduce this on Chrome/Firefox on OSX.

tqlihuiqi:
Hi, @akalipetis 

I use the latest master branch on Chrome/Firefox on OSX.

I've already resolve this issue.

code:
    var currentWs = new ReconnectingWebSocket('ws://{{ wsServer }}/ws/terminal/current');
    currentWs.onopen =  function(event) {
        var xterm = new Terminal({
            geometry: [98, 26],
            debug: true
        });
        xterm.open(document.getElementById('currentRecord'));
        currentWs.onmessage = function(event) {

            if(!xterm.write(event.data)){
                xterm.write('');
            }

        }; 
    }

tqlihuiqi:
Hi, @akalipetis

I use the latest master branch on Chrome/Firefox on OSX.

I've already resolve this issue.

code:

    var currentWs = new ReconnectingWebSocket('ws://{{ wsServer }}/ws/terminal/current');
    currentWs.onopen =  function(event) {
        var xterm = new Terminal({
            geometry: [98, 26],
            debug: true
        });
        xterm.open(document.getElementById('currentRecord'));
        currentWs.onmessage = function(event) {
            if(!xterm.write(event.data)){
                xterm.write('');
            }
        }; 
    }

jerch:
@tqlihuiqi Could you provide more info about what you did, the error (with stacktrace?) and what exactly fixed the problem? The
```
[...]

PR #514, opened 2017-01-27
[issue #164] Fix bug that can occur when setting a scroll region

LucianBuzzo:
This connects to #164 
Fixes an issue where a scroll region can be set out of bounds.


parisk:
Thanks for this PR @LucianBuzzo! Can you provide us with some steps to test this PR out?

Maybe some before vs after results?

LucianBuzzo:
@parisk Using the demo, create a file called `test.sh`  with the following code:
```sh
printf "\033[1;300r"
grep -Rl ' ' .
```
Then run it with `sh test.sh`, this will throw errors in the console and make the term unusable.
This PR fixes the issue.

parisk:
Yep, works great. Thanks!

PR #520, opened 2017-02-01
Add null checks in eraseRight and eraseLeft

### LINKED ISSUE #519: eraseRight can throw because line can be undefined
See https://github.com/Microsoft/vscode/issues/19651

Tyriar:
Fixes #519


Tyriar:
It was running some script that was piping data between xterm.js and a remote server, I think it was happening due to either interleaved CSI codes or the delay between server and host. I'll keep an eye out.

Tyriar:
Looks like the remote process was running with different dimensions to the xterm.js instance which likely caused this.

PR #521, opened 2017-02-01
Don't requeue animation frame if not needed

### LINKED ISSUE #517: Refresh animation frame should be turned off when idle
Downstream issue https://github.com/Microsoft/vscode/issues/19640

Since we know when the next update is going to happen we `queueRefresh` could just turn the frame on instead of what is currently does and then `refreshLoop` can turn it off when there are no more changes.

Tyriar:
Fixes #517 

---

This should improve idle CPU/battery life, particularly when multiple terminals are running.


parisk:
There is only one `window` reference issue in tests and then is good to go 🚀 .

PR #522, opened 2017-02-01
Pull renderer out of xterm.js

### LINKED ISSUE #266: Allow xterm.js to exist and receive/send data before it's attached to the DOM
In v1.6 of VS Code the newly introduced Terminal API will allow processes to exist before the actual terminal panel/element is created. While this is currently possible, xterm cannot be initialized until the element exists, which means the output before the element is attached is lost. It would be good if everything was functional only no rendering took place until `open` is invoked.

Tyriar:
Fixes #266
Merge https://github.com/sourcelair/xterm.js/pull/521 first

---

This makes it possible to hook up messages with the pty and they are retained before `Terminal.open` is called.

### DIFF
--- a/src/Renderer.ts
+++ b/src/Renderer.ts
@@ -1,0 +1,303 @@
+import { ITerminal } from './Interfaces';
+
+/**
+ * The maximum number of refresh frames to skip when the write buffer is non-
+ * empty. Note that these frames may be intermingled with frames that are
+ * skipped via requestAnimationFrame's mechanism.
+ */
+const MAX_REFRESH_FRAME_SKIP = 5;
+
+// TODO: Convert flags to number enum
+/**
+ * Flags used to render terminal text properly
+ */
+const FLAGS = {
+  BOLD: 1,
+  UNDERLINE: 2,
+  BLINK: 4,
+  INVERSE: 8,
+  INVISIBLE: 16
+};
+
+let brokenBold: boolean = null;
+
+export class Renderer {
+  /** A queue of the rows to be refreshed */
+  private _refreshRowsQueue: {start: number, end: number}[] = [];
+  private _refreshFramesSkipped = 0;
+  private _refreshAnimationFrame = null;
+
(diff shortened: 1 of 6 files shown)


src/Renderer.ts
```diff
@@ -0,0 +1,303 @@
+import { ITerminal } from './Interfaces';
```
parisk: License comment? 😊

src/xterm.js
```diff
@@ -1065,261 +1050,26 @@ Terminal.prototype.destroy = function() {
...
-        }
-      }
-      this.refreshRowsQueue = [];
-      this.refresh(start, end);
-    }
-  }
-  window.requestAnimationFrame(this.refreshLoop.bind(this));
-}
+Terminal.prototype.refresh = function(start, end) {
+  this.queueRefresh(start, end);
```
parisk: Should we call `this.renderer.queueRefresh` directly?

src/xterm.js
```diff
@@ -1065,261 +1050,26 @@ Terminal.prototype.destroy = function() {
...
-    if (attr !== this.defAttr) {
-      out += '</span>';
-    }
-
-    this.children[y].innerHTML = out;
-  }
-
-  if (parent) {
-    this.element.appendChild(this.rowContainer);
+Terminal.prototype.queueRefresh = function(start, end) {
```
parisk: Do we need this method in the Terminal prototype?

I think we can "hide" from Terminal that queueing even exists.

Tyriar: This was a facade to reduce the diff, I'll clean it up.

src/xterm.js
Tyriar: I like it, queueRefresh was a facade to simplify the change but there aren't as many calls as I expected so it's easy to change.

Tyriar:
@parisk made the changes, did some clean up and fixed a nasty bug (renderer disappears after a reset). I'll merge this now :+1:

PR #525, opened 2017-02-05
Fix issue when resizing hidden terminals

LucianBuzzo:
Whilst using the 'fit' addon, if you run `Terminal.fit()` whilst the terminal is hidden (with `display: none` for example), then the terminal will become unusable.
The solution I'm proposing is to exit the `Terminal.resize()` function if either the x or y parameter is a NaN value.


parisk:
Thanks @LucianBuzzo

vincentwoo:
This change broke IE, which does not support `Number.isNaN`. For alternatives, please see http://stackoverflow.com/questions/26962341/number-isnan-doesnt-exists-in-ie

PR #530, opened 2017-02-08
Disable escape sequence tests on windows

Tyriar:
Before the tests error out, now they all pass on Windows 10.


Tyriar:
Merging for 2.3.0 since it's test only.

PR #532, opened 2017-02-08
Remove excess library.

AndrienkoAleksandr:
Tsify lib used for browserify typescript sources. But folder /lib consist of js files has already compiled from TypeScript sources (by gulp task "tsc"). So Tsify library does nothing in this case.


Tyriar:
LGTM, I'll merge this soon after the 2.3.0 release has been verified.

PR #538, opened 2017-02-10
Implement web links and custom link matcher registration

### LINKED ISSUE #455: Support links in the terminal
## Context

The linkify addon has been broken for some time with little attention to it, yet it remains quite a popular request to have be have active links for VS Code (https://github.com/Microsoft/vscode/issues/7321). I'm proposing we pull the linkify logic into the core and expose it as an option.

## Why?

One of the reasons linkify is currently broken with no way forward as it is implemented as an addon outside of xterm.js, dealing with things that are very much internal. In addition, with all the performance changes on their way in (particularly this one https://github.com/sourcelair/xterm.js/pull/450), it is not safe to modify `.xterm-rows`. This is also very performance critical so it should not be messed with at all my addons.

### When linkify gets triggered

In order to prevent this from causing a big performance hit I propose we only check for links a certain period of time (100ms, 200ms?) after there has been no changes to the viewport's buffer. This is in contrast to the current implementation which I believe requires an external call.

### Web vs local

The current implementation only linkifies hyperlinks as it was designed just for the browser. One of the most compelling use cases however is to linkify local files so that they can be opened in the editor.

Obviously regular `<a>` tags won't cut it for non-hyperlinks as they won't know how to handle them, we will need something like an `attachLinkifyHandler` function (even better deprecate and merge `attachCustomKeydownHandler` into some generic `attachHandler`/`registerHandler` function).

It's also important to note that in Electron it's probably not desirable to jump use `<a>` tags but handle them custom as well. We do however want to keep `<a>` tags for web files to get the nice middle click/ctrl+click behavior. So the links will be:

- web: `<a>` with href set
- local: `<a>` with no href provided a linkify handler has been attached, otherwise don't check for local links

### Better viewport row reuse

How `Terminal.refresh` works currently, all links would disappear when a single line is added to bottom of the buffer. To better support linkify and also save some CPU cycles while we're at it, we should look into reusing rows by informing `Terminal.rowContainers` that we simply shifted some elements whenever possible and only run a targeted refresh on a specific row.

### Make the span object pool aware of the links

The DomElementObjectPool logic within `Terminal.refresh` which keeps `<span>`s around so not as many DOM nodes are constructed will need to be made aware of the potential for `<a>` tags wrapping `<span>` and even splitting `<span>`s in some cases.

Tyriar:
Fixes #455

### :new: Web links

This PR adds web links to the terminal which can be clicked to open the URL, a custom handler can be attached which will change the links to use an `onclick` handler instead.

```js
Terminal.prototype.attachCustomKeydownHandler = function(customKeydownHandler) { ... }
```

image

### :new: Custom link matchers

The ability for consumers to register custom link matchers has also been added, these are basically the same as web links but can match literally anything the consumer wants.

```js
Terminal.prototype.registerLinkMatcher = function(regex, handler, matchIndex) { ... };
Terminal.prototype.deregisterLinkMatcher = function(matcherId) { ... };
```

image

### Implementation details

- Web links use the old linkify addon's `strictUrlRegex`, meaning only http(s) protocol prefixed links will activate
- Linkification should happen 200ms after a row was last updated, provided no other refresh occurred at which point the timer would be reset. With the default styling of mouse over underline, users likely wouldn't even notice this delay.
- Link matchers are evaluated in the order in which they are added, starting with the builtin web link handler. This will likely need to change later and support priorities as generally new matchers should be higher priority but the http handler needs to beat out many of them of paths would be incorrectly matched.
- The web link handler cannot be deregistered.
- The link matcher API allows the regex match ID to be set, to allow greater flexibility.
- Testing should be easy after https://github.com/sourcelair/xterm.js/pull/535 lands.


Tyriar:
@parisk I'm having second thoughts on how deregister/register works. While currently it's very simple to use, it could be made far more powerful if the link matchers were assigned to actual rows. Consider the following which would be a really cool experience:

1. Scan for `ls` line
2. Register an `ls` link matcher which will linkify all items in the output
3. Scan for end of prompt
4. Deregister `ls` link matcher

Currently the above will do not linkify anything as the link matcher is immediately deregistered. Before merging this I'm going to look at tighter integration between the linkifier and the rows, so instead of just discarding the link matcher completely on deregister, the rows where it was active would keep a reference to it (which would now live outside Linkifier) and eventually be discarded when all references to it are gone (when all rows that had that link matcher go off the end of the buffer and outside references are dropped).

parisk:
I think I get it and indeed sounds powerful. Just a clarification question:

How will you register link matchers to lines (e.g. using line number, DOM element, something else)?

Also, will the linkifying happen automatically or the host application has to invoke it manually at will.

Tyriar:
There would need to be another step where when a row is written to it assigns references to the current link handlers on `lines`. As for where exactly, thinking about it.

parisk:
Do you plan to implement this in this PR or as an enhancement in a next one?

Tyriar:
In a new one would probably be better provided we don't advertise the API yet as it could change.

Tyriar:
Merging this, created the following follow ups: https://github.com/sourcelair/xterm.js/issues/549, https://github.com/sourcelair/xterm.js/issues/550, https://github.com/sourcelair/xterm.js/issues/551

puzrin:
@Tyriar I'd suggest to take a look at https://github.com/markdown-it/linkify-it for link pattern search algorythms.

Tyriar:
@puzrin thanks for the link!

PR #542, opened 2017-02-10
Perform initial viewport sync after CharMeasure is ready

### LINKED ISSUE #539: Terminal background initially has wrong size
We've started seeing this since we upgraded xterm from 2.2.3 to 2.3.1 in Jupyter:

terminal problem

We're passing in `rows` and `cols` options when we instantiate `Terminal`, and I've checked that `rows` is being passed a sensible value, but the background (`xterm-viewport` div) is not getting the correct height. As soon as resizing the window triggers the terminal to resize, it corrects itself and displays correctly.

Apologies if we're doing something stupid here, but we're currently stumped as to what it could be.

## Details
- Browser and browser version: Reproduced on Firefox 51 and Chromium.
- OS version: Ubuntu 16.04
- xterm.js version: 2.3.1

Cross-linking jupyter/notebook#2172

Tyriar:
Fixes #539 

---

@parisk I recommend putting this in for 2.3.2 as otherwise the background won't be the right size until after resize is called.

### DIFF
--- a/src/Viewport.test.ts
+++ b/src/Viewport.test.ts
@@ -41,13 +41,17 @@ describe('Viewport', () => {
   });
 
   describe('refresh', () => {
-    it('should set the line-height of the terminal', () => {
-      assert.equal(viewportElement.style.lineHeight, CHARACTER_HEIGHT + 'px');
-      assert.equal(terminal.rowContainer.style.lineHeight, CHARACTER_HEIGHT + 'px');
-      charMeasure.height = 1;
-      viewport.refresh();
-      assert.equal(viewportElement.style.lineHeight, '1px');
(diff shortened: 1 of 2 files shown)


blink1073:
Thanks for this!

Tyriar:
@blink1073 thanks for helping with finding and reporting :smiley:

takluyver:
Thanks! Note you've got a couple of test failures:

```
2 failing
  1) Viewport refresh should set the line-height of the terminal:
     AssertionError: expected 0 to equal '10px'
  
  2) Viewport syncScrollArea should sync the scroll area:
     AssertionError: expected 0 to equal '0px'
```

parisk:
Merging ASAP and publishing release.

Issue #346, opened 2016-11-08
Can not input on Safari


tsl0922:
Can not input on safari, but works fine on Chrome.

## Details

- Browser version : Safari 10.0.1 (12602.2.14.0.7)
- OS version: macOS Sierra 10.12.1 (16B2555)
- xterm.js version: master (2cba2ff3d4dc545d10efcc2cc040b150a9bb5685)

### Steps to reproduce

1. clone source code from github
2. `npm install` then `npm start`
3. open http://localhost:3000 on Safari
4. input some char, nothing displayed back

This may not always happen, It seems to be something with the input focus.

parisk:
This could be the WebSocket closing. Could you please try refreshing the window and us know if it works after refreshing?

tsl0922:
Refresh doesn't work, it seems not related to websocket, I can input a char by clicking on the cursor and type quickly.

I'm trying to port ttyd from hterm to xterm.js (https://github.com/tsl0922/ttyd/pull/22), can't input on safari, too.

parisk:
Just tried this out and works 👍  for me.

Can you try in a private window please? It might be an extension / addon issue.

tsl0922:
Tried in a private window, not work, then removed all the safari extension, not work too.

parisk:
🆒 . Do you get any console logs/errors?

Also can you please:

1. make a test by binding to the [`keydown`](http://xtermjs.org/docs/api/Terminal/#keydown) event and see if keystrokes are being captured properly
2. take a look at the websocket frames sent and see if the proper character sequences are being sent to the back-end?

tsl0922:
Hi, @parisk 

I can confirm the key event not captured properly, this is my test code (`demo/main.js`), nothing logged when I type.

```js
function runRealTerminal() {
  term.attach(socket);
  term._initialized = true;
  term.on("keydown", function(e){
    console.log(e);
  });
  term.on("key", function(key, e){
    console.log(key, e);
  });
}
```

If I click the terminal with mouse and input quickly, key event can be captured, but only one click one char.

tsl0922:
Hi, again.

I've resolved this issue by clearing the browser cache from safari's developer menu. Don't know why this helps, but thanks for your help!

parisk:
🆒  happy it worked out at last!

achimnol:
[...]

PR #543, opened 2017-02-13
Fix non-IME character input in Safari (both mobile and desktop)

achimnol:
Regarding my comment to #346, I propose a PR to fix it.

I have some updates to my comment as well:

 * I found that duplicate keystrokes in my Firefox was a bogus error and it never happened again.
 * Also, broken Korean IME input in Chrome was actually related to the shell's locale and inputrc settings<sup>[(1)](#fn1)</sup> and not a bug in xterm.js.

<sup><a name="fn1">(1)</a></sup> I had to uncomment `set convert-meta off` in `/etc/inputrc` inside my tested Ubuntu container connected via xterm.js, and run `locale-gen en_US.UTF-8` with setting `LANG` environment variable. In docker containers, base images often omit default locale settings to keep the image sizes small.

So, with my patch, I confirmed that both IME and non-IME input works well with latest versions of Chrome/Safari/Firefox on macOS, Safari on iOS, and Edge on Windows 10.

Known problem:

 * When typing fast (repeating to press another key before release the previous one), sometimes keystrokes are duplicate. But on normal typing, this almost doesn't happen. It seems to be a Safari bug in handling keyboard events.

Here follows the commit log:

 * Safari does not fire keypress events but textInput events when typing
   individual English characters.

 * Chrome and IE 9 also support textInput event but they also support
   keypress event -- let's favor the more widely adopted name, keypress,
   if they provide equivalent functionality.

 * Modern practices favor feature detection instead of browser
   detection, but I resorted to the latter via `Terminal.isSafari` flag.
   It's difficult to use feature detection for textInput event as there
   is no "ontextinput" attribute and Safari also provides "onkeypress"
   event handler though it is not fired as expected.

### DIFF
--- a/src/utils/Browser.ts
+++ b/src/utils/Browser.ts
@@ -7,11 +7,16 @@
 import { contains } from './Generic';
 
 const isNode = (typeof navigator === 'undefined') ? true : false;
+const g = (isNode) ? global : (<any>window);
 const userAgent = (isNode) ? 'node' : navigator.userAgent;
 const platform = (isNode) ? 'node' : navigator.platform;
 
 export const isFirefox = !!~userAgent.indexOf('Firefox');
 export const isMSIE = !!~userAgent.indexOf('MSIE') || !!~userAgent.indexOf('Trident');
+export const isSafari = (isNode) ? false : (/constructor/i.test(g.HTMLElement)
+                                            || ((p) => {
+                                                 return p.toString() === "[object SafariRemoteNotification]";
+                                               })(!g['safari'] || g['safari'].pushNotification));
 
 // Find the users platform. We use this to interpret the meta key
 // and ISO third level shifts.
--- a/src/utils/Browser.ts
+++ b/src/utils/Browser.ts
@@ -15,7 +15,7 @@ export const isFirefox = !!~userAgent.indexOf('Firefox');
 export const isMSIE = !!~userAgent.indexOf('MSIE') || !!~userAgent.indexOf('Trident');
 export const isSafari = (isNode) ? false : (/constructor/i.test(g.HTMLElement)
                                             || ((p) => {
-                                                 return p.toString() === "[object SafariRemoteNotification]";
+                                                 return p.toString() === '[object SafariRemoteNotification]';
                                                })(!g['safari'] || g['safari'].pushNotification));
 
 // Find the users platform. We use this to interpret the meta key
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -335,16 +335,6 @@ Terminal.vcolors = (function() {
 })();
 
 /**
- * Browser compatibility flags.
- */
-
-// Check if we are running on Safari
-// (ref: http://stackoverflow.com/questions/9847580)
-Terminal.isSafari =  /constructor/i.test(window.HTMLElement)
-                 || (function (p) { return p.toString() === "[object SafariRemoteNotification]"; })
-                    (!window['safari'] || safari.pushNotification);
-
-/**
  * Options
  */
 
@@ -546,11 +536,11 @@ Terminal.bindKeys = function(term) {
     term.keyDown(ev);
   }, true);
 
-  var keyPressEventName = (Terminal.isSafari) ? 'textInput' : 'keypress';
+  var keyPressEventName = (term.browser.isSafari) ? 'textInput' : 'keypress';
   on(term.textarea, keyPressEventName, function(ev) {
     term.keyPress(ev);
     // Truncate the textarea's value, since it is not needed
-    term.textarea.value = '';
+    this.value = '';
   }, true);
 
   on(term.textarea, 'compositionstart', term.compositionHelper.compositionstart.bind(term.compositionHelper));
@@ -1646,7 +1636,7 @@ Terminal.prototype.keyPress = function(ev) {
     return false;
   }
 
-  if (Terminal.isSafari) {
+  if (this.browser.isSafari) {
     key = ev.data;
   } else {
     if (ev.charCode) {
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -335,6 +335,16 @@ Terminal.vcolors = (function() {
 })();
 
 /**
+ * Browser compatibility flags.
+ */
+
+// Check if we are running on Safari
+// (ref: http://stackoverflow.com/questions/9847580)
+Terminal.isSafari =  /constructor/i.test(window.HTMLElement)
+                 || (function (p) { return p.toString() === "[object SafariRemoteNotification]"; })
+                    (!window['safari'] || safari.pushNotification);
+
+/**
  * Options
  */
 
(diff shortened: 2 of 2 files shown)


achimnol:
The test failure is related to use of `window` object which does not exist in Node. I'm not very familiar with Javascript unit tests, please suggest me a better way to fix this.

src/utils/Browser.ts
```diff
@@ -7,11 +7,16 @@
 import { contains } from './Generic';
 
 const isNode = (typeof navigator === 'undefined') ? true : false;
+const g = (isNode) ? global : (<any>window);
 const userAgent = (isNode) ? 'node' : navigator.userAgent;
 const platform = (isNode) ? 'node' : navigator.platform;
 
 export const isFirefox = !!~userAgent.indexOf('Firefox');
 export const isMSIE = !!~userAgent.indexOf('MSIE') || !!~userAgent.indexOf('Trident');
+export const isSafari = (isNode) ? false : (/constructor/i.test(g.HTMLElement)
```
Tyriar: This is very difficult to read, also why don't you just test the user agent http://stackoverflow.com/a/7944490/1156119?

src/xterm.js
```diff
@@ -536,7 +536,8 @@ Terminal.bindKeys = function(term) {
     term.keyDown(ev);
   }, true);
 
-  on(term.textarea, 'keypress', function(ev) {
+  var keyPressEventName = (term.browser.isSafari) ? 'textInput' : 'keypress';
```
Tyriar: I'll defer this to @parisk, I'm a little concerned about forking the logic like this depending on the browser, unless it's absolutely necessary.

src/utils/Browser.ts
achimnol: I have taken the code from http://stackoverflow.com/questions/9847580 which seems to be working more reliably than user-agent string checks.

parisk:
👍  while I am skeptical with branching input handling, I will take a look after we merge and release #542.

src/utils/Browser.ts
parisk: While this might be the most accurate way to detect safari, it is indeed quite complex.

I think that just testing `navigator.userAgent` over containing `Safari` and **not containing** `Chrome` is more than OK and keeps the code cleaner.

src/xterm.js
parisk: @achimnol can you provide some documentation on the `textInput` event?

Is this possibly the [`input` event](https://developer.mozilla.org/en-US/docs/Web/Events/input).

achimnol:
@parisk Hm... I didn't know it is actually working in other people's computer because I and my colleagues experienced the same problem. Not sure, but, as we are Korean, we set macOS's language priority to Korean as primary and English as secondary. I have heard lots of cases this priority settings often causes unexpected bugs in many applications. I will soon test this scenario.
ps. I suspected also [my 3rd party IME (Gureum)](https://github.com/gureum/gureum) but my colleague uses the vanilla one.

With a small sample HTML file, `input` events seems to work consistently across different browsers but it lacks the information about which key is pressed; it just notifies that the content of element has changed. I used `textinput` because it works similarly to `keypress` event and xterm.js' existing `keyPress` event handler expects the event argument to have the keycode data.

src/utils/Browser.ts
achimnol: My concern is that many modern browsers' user-agent string has "Safari" and "AppleWebkit". Even if use of "textinput" event for all such browsers works, the naming "isSafari" then gets confusing because it would be true for non-Safari browsers sharing the engine.
Maybe it's better to use feature-centric namings such as `requiresTextInputHack` if you feel uncomfortable with that this complex piece of code is here just to say "it's Safari".
Anyway, if you still want to keep this code in similar complexity with other lines, I will change it to use user-agent strings.

src/xterm.js
achimnol: The reference I consulted is http://help.dottoro.com/ljhiwalm.php . An issue with `input` event is that it lacks the information about which key is pressed, unlike `keypress` or `textinput` event.

parisk: Hmm... Can you confirm if `textInput` is a standard event? Can't find it anywhere in https://developer.mozilla.org/en-US/docs/Web/Events, https://en.wikipedia.org/wiki/DOM_events or https://www.w3.org/TR/input-events/.

src/utils/Browser.ts
Tyriar: The standard way seems to be to check for "Safari/..." and not "Chrome/" though, other browsers will likely have both to try to boost compatibility.

Tyriar:
If you switch to English does input work fine? It could be related to IMEs within Safari which is handled in the `CompositionHelper` class.

parisk:
@achimnol can you confirm if textInput is a standard event? Can't find it anywhere in https://developer.mozilla.org/en-US/docs/Web/Events, https://en.wikipedia.org/wiki/DOM_events or https://www.w3.org/TR/input-events/.

achimnol:
I had changed the default language of macOS to English and using the defualt English IME but there was no difference.

src/xterm.js
achimnol: Though [there seems no comprehensive documentation for Safari/Webkit](http://stackoverflow.com/questions/871460/webkit-javascript-reference), I found some references to `textInput` event in Webkit's source code, which is to abstract/simplify IME composition events:
 * https://github.com/WebKit/webkit/blob/d13624/Source/WebCore/dom/EventNames.h#L220
 * https://github.com/WebKit/webkit/blob/727c7a/LayoutTests/http/tests/navigation/resources/keyboard-events-test.js
 * https://github.com/WebKit/webkit/blob/19c8b9/LayoutTests/fast/events/key-events-in-input-button.html#L21-L22

There is also `oninput`-related events:
 * https://github.com/WebKit/webkit/blob/382a078/LayoutTests/fast/events/input-events-typing-data.html
 * https://github.com/WebKit/webkit/blob/382a078/LayoutTests/fast/events/input-events-typing-data-expected.txt

It seems that `oninput` event is the *intended* one to be used in our case, but for some reason, only `textInput` events (related to IME composition) are fired on our environments.

parisk:
As far as I understand based on https://github.com/sourcelair/xterm.js/pull/543#discussion_r101914586, `textInput` is not an officially supported and documented event, even for Safari/Webkit only.

To be honest, I am quite reluctant in utilizing it because of this, but we can hear @Tyriar's opinion too.

@achimnol do you believe you can find any standard event or way (or at least officially supported by Safari) that could do the job, instead of `textInput`?

Tyriar:
I never fully understood why we have both keydown and keypress so I'd like some clarity on that first.

Also based on http://stackoverflow.com/a/4890059/1156119 I would expect Safari to fire the keypress event for regular input?

parisk:
Ping @achimnol do you believe you can find any standard event or way (or at least officially supported by Safari) that could do the job, instead of textInput?

achimnol:
@parisk Sorry for late response, I have been busy for my business affairs for last few weeks.

I have tried to reproduce this keypress missing issue and found that some Korean Mac users does not experience it. Their Safari get keypress events like a charm! 😧  Still, I could not figure out what differences in the systems make it working or not working...

The only similar online reference to this problem (but without solution/answers) is: https://forums.developer.apple.com/thread/61607
NOTE: As I have tested, using vanilla IME did not solve my case.

parisk:
Thanks for the update @achimnol.

If that's OK with you I will move on to close this PR, since we have no standards-based way of reproducing and fixing this right now.

If you find anything that can fix this issue, using documented standards-based APIs, please open up a new PR. We would love to review it!

Thanks a lot for the effort you have put on this.

achimnol:
Thanks for reviews and discussions.
I think this may be a bug in macOS or Safari, or something else.
If I have any news, I'll revisit later.

PR #544, opened 2017-02-13
Make fit plugin guard against non-existent parent element.

LucianBuzzo:
If the terminal's parent element is removed from the DOM and the `.fit()` method is called, an error is thrown.
This PR introduces guard logic that checks if the parent exists.

### DIFF
--- a/src/addons/fit/fit.js
+++ b/src/addons/fit/fit.js
@@ -32,6 +32,9 @@
   var exports = {};
 
   exports.proposeGeometry = function (term) {
+    if (!term.element.parentElement) {
+      return null;
+    }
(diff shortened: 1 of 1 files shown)


src/addons/fit/fit.js
```diff
@@ -32,6 +32,9 @@
   var exports = {};
 
   exports.proposeGeometry = function (term) {
+    if (!term.element.parentElement) {
+      return false;
```
Tyriar: I think `null` would be more fitting here, other than that looks good

LucianBuzzo: Sure thing, I'll update the PR.

PR #571, opened 2017-02-27
Implement link validation

### LINKED ISSUE #570: Support link validation
Currently links in the terminal are always "active", I propose adding a way to set a callback when registering a pattern that the consumer would return `true` or `false` on depending on whether the link is valid.

Example:

```ts
term.registerLinkMatcher(/\w+\.json/, () => console.log(Date.now()), {
  validationCallback: function (url, cb) {
    // Use a cb so this can be async to check the disk
    cb(url === 'tsconfig.json');
  }
});
```

Tyriar:
Fixes #570 

---

This allows consumers to invalidate links:

```ts
term.registerLinkMatcher(/\w+\.json/, () => console.log(Date.now()), {
  validationCallback: function (uri, cb) {
    // Only enable URIs that exist on disk
    cb(fs.existsSync(uri));
  }
});
```

As part of this I also removed PhantomJS in favor of much more lightweight jsdom. This makes the tests far more readable, is way faster to download and run and everything runs through the mocha gulp task so they're included in the coverage report!

The one downside with jsdom is that it doesn't do layouts so we have to fake them in the tests, this is a small price to pay for the benefits though imo.


test-harness.html
```diff
@@ -14,6 +14,7 @@
       assert = chai.assert
     </script>
     <script src="build/xterm.js"></script>
+    <script src="lib/Linkifier.phantom.js"></script>
```
parisk: Should we change the naming from `*.phantom.js` to `*.jsdom.js`?

test-harness.html
```diff
@@ -14,6 +14,7 @@
...
     <script src="lib/utils/CharMeasure.phantom.js"></script>
```
parisk: Same here, should we change the naming from `*.phantom.js` to `*.jsdom.js` since we ditched PhantomJS in favor of jsdom?

test-harness.html
Tyriar: Opps, the test harness was meant to be deleted. Fixed in next commit. jsdom tests are now run with the regular tests so no need for a custom name.

PR #572, opened 2017-02-28
Use tsconfig.json data in gulpfile.js instead of hardcoded values.

AndrienkoAleksandr:
Use tsconfig.json data in gulpfile.js instead of hardcoded values. Little code clean up

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -13,27 +13,28 @@ const ts = require('gulp-typescript');
 
 
 let buildDir = process.env.BUILD_DIR || 'build';
-
+let tsProject = ts.createProject('tsconfig.json');
+let srcDir = tsProject.config.compilerOptions.rootDir;
+let outDir = tsProject.config.compilerOptions.outDir;
 
 /**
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -14,27 +14,28 @@ const ts = require('gulp-typescript');
 
 
 let buildDir = process.env.BUILD_DIR || 'build';
-
+let tsProject = ts.createProject('tsconfig.json');
```
Tyriar: Any reason you're not just requiring the json file here?

parisk: I think this is because `tsProject.src()` is being used below by Gulp.

This was the case also before, but @AndrienkoAleksandr seems to have brought `createProject` at the top level, to be able to access it's settings inside the tasks.

parisk:
@AndrienkoAleksandr can you please rebase with master, in order to merge this?

AndrienkoAleksandr:
@parisk branch rebased.

Tyriar:
Let's wait until 2.4 is released before we merge this (just in case).

parisk:
Looks great, thanks!

PR #579, opened 2017-03-03
Require ctrl for link clicks

### LINKED ISSUE #578: Change terminal links to require ctrl
I think we should change the link behavior to require ctrl+click for the following reasons:

- This is how most terminal emulators work
- Users can accidentally open links when they want to focus the terminal or select text
- It could mess with terminal applications

Tyriar:
Fixes #578

### DIFF
--- a/src/Interfaces.ts
+++ b/src/Interfaces.ts
@@ -86,12 +86,6 @@ export interface LinkMatcherOptions {
    */
   validationCallback?: LinkMatcherValidationCallback;
   /**
-   * A callback that fires when a link click fails due to ctrl or cmd (mac) not
-   * being pressed. This allows hinting to the user after a likely  failed
-   * click.
-   */
-  clickFailCallback?: (event: MouseEvent) => any;
-  /**
    * The priority of the link matcher, this defines the order in which the link
    * matcher is evaluated relative to others, from highest to lowest. The
    * default value is 0.
--- a/src/Interfaces.ts
+++ b/src/Interfaces.ts
@@ -86,6 +86,12 @@ export interface LinkMatcherOptions {
    */
   validationCallback?: LinkMatcherValidationCallback;
   /**
+   * A callback that fires when a link click fails due to ctrl or cmd (mac) not
+   * being pressed. This allows hinting to the user after a likely  failed
+   * click.
+   */
+  clickFailCallback?: (event: MouseEvent) => any;
+  /**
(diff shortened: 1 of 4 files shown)


mofux:
I actually quite like the way cloud9 does it [https://github.com/c9/core](https://github.com/c9/core):

If you hover a resolved link, it will get underlined and you get a cursor: pointer, and on left click they will show a context menu that allows you to interact with the file. Adding the context menu should be handled by the implementor, but it would be awesome if we could listen to link clicked events, e.g.:

```javascript

term.on('link-open', (evt, uri) => {

  // only continue if the ctrl key was held
  if (!evt.ctrlKey) return evt.preventDefault();

  // show context menu for this uri
  ...

});

```

screen shot 2017-03-03 at 12 38 29

mofux:
@Tyriar Just checked how c9 does their link matching, as it works pretty well. Here's how they do it, just if you're interested: [https://github.com/c9/core/blob/master/plugins/c9.ide.terminal/aceterm/hover_link.js](https://github.com/c9/core/blob/master/plugins/c9.ide.terminal/aceterm/hover_link.js)

parisk:
@Tyriar I have my concerns about making Ctrl + Click the default behavior, since it's a non-standard.

I think that we should just expose the linkified element in some sort of event and let the consumer app manipulate it. So IMO just exposing the element to the handler should do the job.

Tyriar:
@parisk you mean the event not the element?

mofux:
The event would contain a reference to the element (evt.target), wouldn't it? I think a good reason for exposing the event would be the fact that it could be prevented (to stop it following the link on click for example)

Tyriar:
@mofux yes it would, but also expose things like `MouseEvent.ctrlKey` so consumers can do ctrl key themselves.

mofux:
@Tyriar perfect, and the href attribute of the element would probably equal the uri from the matcher?

Tyriar:
The uri would need to be a param as well: `(event: MouseEvent, uri: string) => boolean` since only standard web links use `href`

mofux:
Perfect, looks good to me.

mofux:
What is the reason not to `href` for example `file://` uri's? I think opening the file locally when clicked and not prevent would be a good default.

Tyriar:
Yes it allows a handler to override the hypertext logic.

Issue #157, opened 2016-07-04
No bold text in terminal output


protasenko:
`isBoldBroken()` always returns true since it inserts the test span at `body` level where the font is most likely not mono-spaced. Instead, it should test at or below the node where terminal font is specified (`.terminal`). As an example, the below version will allow bold text to be rendered by terminal:

```
/**
 * @param container terminal's parent node
 */
function isBoldBroken(container) {
      container = container.getElementsByClassName("terminal")[0];
      var el = document.createElement('span');
      el.innerHTML = 'hello world';
      container.appendChild(el);
      var w1 = el.scrollWidth;
      el.style.fontWeight = 'bold';
      var w2 = el.scrollWidth;
      container.removeChild(el);
      return w1 !== w2;
}
```

jerch:
Imo this can't be tested programmatically with some random characters in a reliable manner unless you test a critical subset. There are fonts that handle bold for some glyphs and dont for others (esp. bold box drawing chars are kinda broken in all monospace fonts).
Imo this needs handpicking and documentation of useable and unuseable fonts for a specific task (one font might handle bold cyrillic good while rendering the terminal unusable for greek etc...)

parisk:
Hi @protasenko, thanks for reporting. Running `echo "this is $(tput bold)bold$(tput sgr0) but this isn't"` prints the bold text normally.

I think that adding a test which evaluates that printing a fixed sentence with a bold word that will place the `xterm-bold` class to the bold word would suffice.

If the host application of xterm does not display the bold characters after ensuring that this test succeeds most probably means that this is a font issue and not an xterm.js issue.

@protasenko are you OK with closing this issue?

protasenko:
Hi @parisk, just cloned latest source and started demo app.js. Your `echo` test didn't show bold word. Now if I modify index.html like so: `<body style="font-family: 'courier new'">...` then the bold text shows as expected (assuming 'courier new' is installed on the system of course).
This goes to show the fact that isBoldBroken() test is done at the wrong DOM node `document.body`, whereas it should be done at `#terminal` node or below. `document.body` in most cases is not styled to use monospace font, therefore the isBoldBroken() is true even though there's nothing wrong with the actual terminal font.
[...]

PR #582, opened 2017-03-03
Support checking height in checkBoldBroken

### LINKED ISSUE #581: Consider to set height to each rows
Some fonts might have different character height between bold and regular.
As a result, some rows have not expected height.

This is the result on my environment. The row has only regular weight chars, gets 19.43px height.

image

But row has some bold weight chars, get 20.57px height.
image

Please consider to set the height to each rows.

## Details
- Browser and browser version: Chrome Canary (v58.0.3027.0)
- OS version: Windows 7 (6.1.7601)
- xterm.js version: HEAD of master(8693d06)

### Steps to reproduce

1. Type 'echo "test"'
2. Type 'echo -e "\033[1mtest\033[0m"
3. Measure row height 1 and 2.

tmyt:
This PR resolve #581 

After apply this patch, resolve issue and gets these results.

- With no Bold chars

image

- With Bold chars

image

And each row DOM elements has styles.
image

Please consider to fix.

### DIFF
--- a/src/Renderer.ts
+++ b/src/Renderer.ts
@@ -296,9 +296,9 @@ function checkBoldBroken(terminal) {
   const el = document.createElement('span');
   el.innerHTML = 'hello world';
   terminal.appendChild(el);
-  const w1 = el.scrollWidth;
+  const w1 = el.offsetWidth;
   el.style.fontWeight = 'bold';
-  const w2 = el.scrollWidth;
+  const w2 = el.offsetWidth;
   terminal.removeChild(el);
   return w1 !== w2;
 }
--- a/src/Renderer.ts
+++ b/src/Renderer.ts
@@ -297,8 +297,10 @@ function checkBoldBroken(terminal) {
   el.innerHTML = 'hello world';
   terminal.appendChild(el);
   const w1 = el.offsetWidth;
+  const h1 = el.offsetHeight;
   el.style.fontWeight = 'bold';
   const w2 = el.offsetWidth;
+  const h2 = el.offsetHeight;
   terminal.removeChild(el);
-  return w1 !== w2;
+  return w1 !== w2 || h1 !== h2;
 }
--- a/src/Renderer.ts
+++ b/src/Renderer.ts
@@ -34,7 +34,7 @@ export class Renderer {
     // Figure out whether boldness affects
     // the character width of monospace fonts.
     if (brokenBold === null) {
-      brokenBold = checkBoldBroken((<any>this._terminal).document);
+      brokenBold = checkBoldBroken((<any>this._terminal).element);
     }
 
     // TODO: Pull more DOM interactions into Renderer.constructor, element for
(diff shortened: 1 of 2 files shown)


Tyriar:
@tmyt what font are you seeing the issue on?

tmyt:
@Tyriar my browser choose 'courier' from demo app default stylesheet.

Tyriar:
Instead of this approach I think we should roll this into `checkBoldBroken` in Renderer.ts. This function checks whether an element's width is the same with and without bold, if not it will disable bold in the terminal. Fundamentally this is an issue with the font and it shouldn't be used.

I also think we should add a `console.error` when bold is broken to make it clear in the console why bold doesn't work.

tmyt:
Thanks to your review, I agree with that.
Then I tried to detect broken bold with `checkBoldBroken`, but the function does not detect correctly.
Because, generally `body` does not have monospace font. and Chrome does not calculate scrollWidth or scrollHeight for `display: inline` element. The reason similarly to #157.

First, needs to fix `checkBoldBroken` to works correctly.

src/Renderer.ts
```diff
@@ -291,14 +291,16 @@ export class Renderer {
 
 // if bold is broken, we can't
 // use it in the terminal.
-function checkBoldBroken(document) {
-  const body = document.getElementsByTagName('body')[0];
+function checkBoldBroken(terminal) {
+  const document = terminal.ownerDocument;
```
Tyriar: Based on this https://github.com/sourcelair/xterm.js/blob/ec91a6c7f4b8032e3483a8730e6ed66a672415d0/src/xterm.js#L589

Shouldn't `terminal.document` be the same as `terminal.ownerDocument` since terminal.parent should be owned by the same document?

src/Renderer.ts
```diff
@@ -291,14 +291,16 @@ export class Renderer {
...
   const el = document.createElement('span');
   el.innerHTML = 'hello world';
-  body.appendChild(el);
-  const w1 = el.scrollWidth;
+  terminal.appendChild(el);
+  const w1 = el.offsetWidth;
```
Tyriar: I believe we want `scrollWidth`/`scrollHeight` here as `offset*` may not give the correct results if the element is off screen? https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/Determining_the_dimensions_of_elements

Did doing this fix the issue for you?

tmyt: I seems to fix the issue on my environment (Chrome 59 & Firefox 54).
This changes for `scroll*` does not work for `display: inline` element (e.g. `span`) on Chrome. Then I change `scroll*` to `offset*`.
Should I change `offset*` to `scroll*` and use `div` instead of `span`?

src/Renderer.ts
tmyt: This is bad naming `terminal`. The `terminal` passed with `terminal.element`. So should I change naming `terminal` to `terminalElement`.
`terminal.element` is created by `terminal.parent.ownerDocument.createElement`.
As a result `terminal.document` same as `terminalElement.ownerDocument`.

src/Renderer.ts
Tyriar: Main just concerned about if this would break when the terminal is either detached from the DOM or is invisible (`display:none`, `visibility:hidden`).

src/Renderer.ts
Tyriar: If terminal.document is the same as terminalElement.ownerDocument why not leave it as it was? `(<any>this._terminal).document`

tmyt: I needed (<any>this._terminal).element for apply monospace font.
I think, document could get from element, then I want to simplify function arguments.

src/Renderer.ts
tmyt: Both properties could not get the correct size from detached or invisible element. Anyway, currently or my PR both needs on visible DOM tree. Do you have you any great ideas?

src/Renderer.ts
Tyriar: Ah makes sense, you add it to terminal instead of body which is much nicer.

src/Renderer.ts
Tyriar: Thanks for checking if there's no regression this is all good :+1:

PR #586, opened 2017-03-06
Clean up tsconfig.json.

AndrienkoAleksandr:
'include' field in the tsconfig.json defines source scope (top level of this scope is "src" folder), so we do not need 'exclude' folders in the root of the project, because this folders had already excluded by 'include' scope.


Tyriar:
Thanks @AndrienkoAleksandr :smiley:

PR #598, opened 2017-03-14
Fixed characterHeight calculation, which could cause overflowing

mikesir87:
The previous calculation was simply using the height of the letter
W, but wasn't taking into account that each row has a lineHeight,
which might be greater. If the lineHeight is .4px off, after many
rows, it will cause the proposedGeometry to have an extra row,
potentially causing hidden rows (based on layout).

### DIFF
--- a/src/addons/fit/fit.js
+++ b/src/addons/fit/fit.js
@@ -56,7 +56,7 @@
     subjectRow.innerHTML = 'W'; // Common character for measuring width, although on monospace
     characterWidth = subjectRow.getBoundingClientRect().width;
     subjectRow.style.display = ''; // Revert style before calculating height, since they differ.
-    characterHeight = parseFloat(term.rowContainer.style.lineHeight);
+    characterHeight = subjectRow.getBoundingClientRect().height;
     subjectRow.innerHTML = contentBuffer;
 
     rows = parseInt(availableHeight / characterHeight);
--- a/src/addons/fit/fit.js
+++ b/src/addons/fit/fit.js
@@ -56,7 +56,7 @@
     subjectRow.innerHTML = 'W'; // Common character for measuring width, although on monospace
     characterWidth = subjectRow.getBoundingClientRect().width;
     subjectRow.style.display = ''; // Revert style before calculating height, since they differ.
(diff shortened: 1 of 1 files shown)


Tyriar:
Deferring to @parisk since I don't use the fit addon.

mikesir87:
> It seems that the offsetHeight attribute takes into account the actual height, padding and border of an element, so it does not ignore the parent element's lineHeight.

You're absolutely correct here.  The problem is that `offsetHeight` and `clientHeight` return the value as an integer.  But, what if the value was non-integer?  Chrome seems to truncate the decimal completely, while Safari and FF behave differently by using the decimal value. The Codepen below demonstrates this.  You'll see in FF/Safari that if you adjust the line-height to another decimal value, the overall height of the container changes, while in Chrome, it stays constant.

http://codepen.io/anon/pen/LWzEYj

> My objection with this implementation is that it uses the `style` attribute, while it would be better if we could somehow extract the calculated height of the line.

Haha... no objections about that here!  Does seem like a "hack" of a solution.  Looking at how the `CharMeasure` works and is being used, I think a better approach is to use get the `boundingClientRect`'s height.  I tested in FF/Safari/Chrome and they all return the expected value (supporting non-integer values).  New commit reflects this change.

src/addons/fit/fit.js
```diff
@@ -56,7 +56,7 @@
     subjectRow.innerHTML = 'W'; // Common character for measuring width, although on monospace
     characterWidth = subjectRow.getBoundingClientRect().width;
     subjectRow.style.display = ''; // Revert style before calculating height, since they differ.
-    characterHeight = parseFloat(term.rowContainer.style.lineHeight);
+    characterHeight = term.getBoundingClientRect().height;
```
mofux: `term` is the terminal instance, not the subject row that you will want to measure the height on. I'm afraid this line will crash... 🤔

mikesir87: Good catch and sorry about that... yes, it would crash. Just pushed the fix.

parisk:
Looks great now. Thanks a lot @mikesir87!

PR #607, opened 2017-03-15
node-pty@0.6.3

Tyriar:
Bunch of releases have happened in node-pty, should pull in all that good stuff :smile: https://github.com/Tyriar/node-pty/releases


parisk:
Great! Any idea why tests might be failing though?

https://travis-ci.org/sourcelair/xterm.js/jobs/211531244

Tyriar:
It's happening because it was relying on a private API, I'll fix this up when I get a chance.

Tyriar:
I'll close this off since I just released a new version. Didn't fix this particular thing though.

PR #609, opened 2017-03-16
Add logic for reflowing lines and preserving data when resizing

### LINKED ISSUE #622: Support reflowing lines on resize
On resize ideally we want the text to reflow and wrap to the next line where appropriate.

LucianBuzzo:
Fixes #622


mofux:
Very nice approach, love it. I found some issues though:

1. Create some output, then shrink the terminal by half of its previous cols, and scroll up. Even if the wrapped output is below 1000 lines of the scrollback buffer, I am unable so see the start of the scrollback.

2. Run `ls -alh` with 80 cols. Shrink the terminal to 20 cols. run `ls -alh` again. Resize back to 80 cols. The second `ls -alh` output is still wrapped. This does not happen in Terminal.app, not sure how they do it, because technically the wrapped lines are arriving as-is from the pty process. I'm wondering if there's any control sequence that marks the end of the wrapped line?

wrap-unwrap

blink1073:
Thank you so much for this, @LucianBuzzo!

LucianBuzzo:
@mofux A simpler approach to this issue might be to just implement the same behaviour as [hyperterm](https://hyper.is/) and ignore reflowing altogether.
TLDR hyperterm retains the line information but crops the line visually when resizing smaller.

jerch:
> Run ls -alh with 80 cols. Shrink the terminal to 20 cols. run ls -alh again. Resize back to 80 cols. The second ls -alh output is still wrapped. This does not happen in Terminal.app, not sure how they do it, because technically the wrapped lines are arriving as-is from the pty process. I'm wondering if there's any control sequence that marks the end of the wrapped line?

This is possible because the pty normally just inserts '\r' at EOL. The actual newline step is inserted by the emulator itself with the autowrap setting (without '\n' char). By ignoring the '\r' at EOL the lines still belong together. From there you can flow lines in both direction - shrinking and enlarging.

Tyriar:
@mofux on your first point, one of the more tricky issues that arises when implementing this is that when lines wrap and reflow and are no longer fixed, the number of rows expand and contract and the viewport/scrollbar needs to be made aware of it. Scrolling the mouse wheel one notch should jump a consistent number of actual lines as are currently wrapped, not the amount they originated with.

Consider the extreme case which is quite reasonable in a terminal environment of a huge line that consumes the entire viewport, if the terminal is then resized to 50% of the width and a reflow occurs (so the content consumes 2x viewports), you must be able to scroll a consistent number of rows.

Based on your first comment it would seem that only the first 1000 rows are shown and the scroll bar isn't touched at all. We should probably just trim rows that go beyond the scrollback when the terminal width is reduced.

On your second issue I would guess it's related to the pty as @jerch indicates, if that can be worked around that would be cool 😄

LucianBuzzo:
@Tyriar @mofux I think point 2 should be straightforward to fix, I'll take a look 👍

jerch:
@LucianBuzzo You can distinct the EOLs in the write method - any real EOL will be triggered by a '\n' char in the data, therefore any line jumped to by '\n' is a real new line. All other lines are autowrapped and can reflow to a single line by enlarging. Note - that will not work with programs that use the terminal as canvas with heavy cursor jump usage (like vim or any nurses stuff), but since most of those implement the WINCH signal, they will "autorepair" itself afterwards. If they dont, hmm the output will be broken anyways. It might be possible to circumvent reflow for "jumped" lines at all and go with the cropping thingy you mentioned above.

mofux:
I'm probably thinking to naive here, but given what @jerch is saying, if we were only creating a new line if we see '\n', and then let the browser do the wrapping for us, wouldn't that work?

Like 
```css
.xterm-rows > div {
  word-break: break-all;
}
```

would then automatically break the characters into a new line if they don't fit into one. Not sure what this does to the xterm viewport though.

jerch:
@mofux This would work, but introduce tons of problems with the [col, row] grid idea of a terminal and make the cursor jump escape sequences very hard to calculate.

Imho an easier approach would be to mark all terminal lines as "real EOL" on a line level (pseudocode):
```
for line in terminal:
    line.eol = true;
```
and set this to `false` once a wrap happens in write (pseudocode):
```
write(s):
    ...
    if (wrapped):
        line.eol = false;
    ...
```

During resizing the reflow logic would just have to search for a line with ``eol == true`` and fit all following lines into the new width up to the next real EOL line or sumthing like that.

The tricky part is the handling of curses like apps with WINCH. A heuristic approach would be to set eol to ``true`` again as soon as a cursor jump happens into a line by an escape sequence. Another heuristic hint could be the ICANON flag of the pty, if it is available for the emulator.

Tyriar:
> if we were only creating a new line if we see '\n', and then let the browser do the wrapping for us, wouldn't that work?

@mofux doing it that way would be ideal in many respects such getting links spanning multiple rows trivially. *However*, I imagine there's a bunch of problems with doing it as @jerch mentions so it's probably better to make the links more clever instead. Another potential problem area for this is when I get "virtual selection" working https://github.com/sourcelair/xterm.js/issues/207, if things are even a pixel off it would look wrong.

src/Renderer.ts
```diff
@@ -128,21 +129,46 @@ export class Renderer {
...
+
+    let renderCount = 0;
+    let rowsToRender = end - start;
+
+    // Lines are rendered in reverse, from bottom to top, to help line wrapping logic
+    while (renderCount <= rowsToRender) {
+
+      // If there are overflow lines use the last one
+      if (overflowBuffer.length) {
+        line = overflowBuffer.pop();
```
Tyriar: I believe this would suffer from the issue I mentioned in https://github.com/sourcelair/xterm.js/pull/609#issuecomment-287113016 where is a really long line in the terminal that is reflowed within a smaller width gets larger than the viewport, you won't be able to see the rows. And resulting from this scrolling will feel awkward because it will jump the amount of rows that the line is now wrapped, instead of the correct 1.

src/Renderer.ts
```diff
@@ -159,11 +185,15 @@ export class Renderer {
       for (; i < width; i++) {
         if (!line[i]) {
           // Continue if the character is not available, this means a resize is currently in progress
-          continue;
+          data = this._terminal.defAttr;
```
Tyriar: `data`, `ch` and `ch_width` can be shared among each character, should we expose a `Terminal.defaultCharacter` variable or something (perhaps function if defAttr changes). `Terminal.blankLine` reuses these empty characters for a single line.

src/utils/LineWrap.ts
```diff
@@ -0,0 +1,76 @@
...
+ * a coloured bg (ie vim).
+ * Returns a shallow copy of the original array.
+ *
+ * @param {array} line - A terminal line
+ * @param {number} min - The minimum length to trim the line to
+ * @param {number} blank - The code for a default blank character
+ *
+ * @return {array} - The trimmed terminal line
+ */
+export const trimBlank = (line, min, blank) => {
```
Tyriar: Not necessarily in this PR, but it might make sense to pull all this into a new class `BufferModel` which extends `CircularList`, I've felt a class like that has been needed to encapsulate operations on the buffer for a while. This way we can potentially solve the scrolling problem I've mentioned by making the buffer itself aware of the reflows.

LucianBuzzo:
I've updated this PR to include a more complete line reflow solution that should fix the issues raised in comments here.
I still need to fix sticky scrolling and add a caching/memoization layer in front of the buffer, but for the most part it seems to work quite well.

src/Viewport.ts
```diff
@@ -66,9 +67,10 @@ export class Viewport {
    * Updates dimensions and synchronizes the scroll area if necessary.
    */
   public syncScrollArea(): void {
-    if (this.lastRecordedBufferLength !== this.terminal.lines.length) {
+    let totalLines = this.terminal.lines.totalLinesAtWidth(this.terminal.cols);
```
Tyriar: `syncScrollArea` is too performance critical, we cannot call `totalLinesAtWidth` here. I think Buffer needs some `reflow` function to be called when `cols` change, this will do all the heavy lifting of mapping the underlying `CircularList` lines with the reflowed lines only when a resize occurs.

Tyriar:
@LucianBuzzo is this ready for a full review?

LucianBuzzo:
@Tyriar It's very close! I made some changes that dramatically improved performance and it seems to be working very well.
Could you please give it a shakedown and see if there's anything obviously broken?

mofux:
I found a bug with Mac's default `top` application, tested using the latest version of this pull request. Running `top`, then resizing the terminal to 40 cols starts to wrap the lines of `top`s output. In contrast, the default mac Terminal.app will truncate the lines. Anyway, when quitting `top` the terminal is dysfunctional, leaving you with characters all over the place, and I am unable to recover with `clear`.

top

LucianBuzzo:
@mofux The latest commit should fix the issue you're seeing.

mofux:
@LucianBuzzo `top` works perfect now, thanks! Testing some more, I have found another issue with `vim`. If I write some multiline text into vim and then resize from 81 cols to 40 cols for example, the text is wrapped incorrectly, and strange things start to happen when typing some more...

vim

Tyriar:
@LucianBuzzo a bit busy at the moment with VS Code release stuff, so I probably won't be able to have a look for a few days. Very keen to get this in though.

LucianBuzzo:
@mofux When resizing the terminal with vim running, vim appears to actually send the last character twice. Here is the log output straight from the socket `[m[H[2J[1;30Hqwertt[2;1Hy`.
As you can see the `t` character in `qwerty` appears twice at the end of the line.
On the master branch the additional character is never rendered so it's not an issue. 
I'm not sure why this happens and plan to do some more investigation.

LucianBuzzo:
Hmm so this problem doesn't happen with neovim.

mofux:
You are right, I can confirm this is only happening in `vim` but not `nvim`. The big question is, why does the extra character that is sent not rendered?

LucianBuzzo:
@mofux I think there's a typo in your "big question"

mofux:
I am not sure if it is related, but when I change the pty name in [demo/app.js](https://github.com/sourcelair/xterm.js/blob/master/demo/app.js#L28) from `xterm-color` to `vt100`, I get an exception as soon as I start typing in vim:

buffer-error

I actually wanted to test if a different term environment yields the same "duplicate character on line break" error that we are seeing in vim.

LucianBuzzo:
@mofux, this issue should now be fixed for `vt100`. I looked and we're getting the same vim output :(

Tyriar:
Found an issue:

1. Open demo
2. Run `ls`
3. Resize to 60 cols, the output should be wrapped such that there is no scroll bar. It should fill the space below the prompt when resized if there is "empty" space available

image

Tyriar:
Another issue:

1. Type in a lot of content into prompt, the output should wrap like normal and no wrap onto the same line (as in master)
2. Backspace to remove the content

**branch behavior**

image

image


**master behavior**

image

image

Tyriar:
Another issue:

1. Open vim
2. Press i
3. Paste in this: Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse euismod urna quis nisi dapibus sodales tristique a ante. Nunc quis lectus dui.

Notice the line is wrapped incorrectly and the scroll bar is active:

image

Master behavior:

image

LucianBuzzo:
@tyriar vim seems to be sending duplicate characters. Any ideas how to handle this or why it is happening?

jerch:
@LucianBuzzo I think this is related to [DECAWM](http://vt100.net/docs/vt510-rm/DECAWM.html), vim seems to write over EOL. No clue why it does...

LucianBuzzo:
I've been trying to find out what's causing this vim issue. I'm not sure that it's related to DECAWM and I've been unable to find out any info on the duplicate character behaviour.

LucianBuzzo:
Here's the relevant code that causes this https://github.com/vim/vim/blob/master/src/screen.c#L5707
I'm a bit puzzled as to how to get around this behaviour, or how other emulators deal with it.

mofux:
Hmmm.... my theory is that other terminals will also reflow the additional character into the next line, but it then gets overwritten because the cursor is manually positioned to where the extra character would be (see the tty raw data, there is a control sequence that manually sets the cursor to the start of this new line, in your example above `�[m�[H�[2J�[1;30Hqwertt�[2;1Hy` its the `�[2;1H` that sets the cursor to row 2 column 1, which is where the extra character would be). I believe that we have to account the wrapped line as a real line, so when the cursor is set to row 2, it has to go to the wrapped line.

LucianBuzzo:
@mofux, yep this is my conclusion as well. Currently, my implementation moves the cursor to the *source* row, rather than the rendered row, which is why it isn't overwriting the duped character. 
Oh well, at least I know how to progress with this feature now!

parisk:
This thread has been inactive for about 10 days, so I just want to remind @LucianBuzzo that if you need any input at any point here, just ask 😄 !

I think that this is a great enhancement to the xterm.js library, so I will put priority on this, in order to help out.

LucianBuzzo:
@parisk I pushed some commits a few days ago fixing the vim issue and the scroll to bottom issue reported by @Tyriar 
AFAICT We only need to sort the line wrap issue on the command prompt and we're there.
If you can give the branch a shakedown and see if anything is broken or showing weird behaviour that would be ace! 👍

src/Buffer.ts
```diff
@@ -0,0 +1,223 @@
...
+  constructor(scrollback: number, private blank: string) {
+    super(scrollback);
+  }
+
+  /**
+   * Faster version of Math.ceil()
+   * see https://jsperf.com/math-ceil-vs-bitwise
+   *
+   * @param {number} n - The value to round up.
+   *
```
Tyriar: Style nit: be more concise here:

```
@param ...
@return ..

@example
...
```

src/Buffer.ts
```diff
@@ -0,0 +1,223 @@
...
+   * @example
+   *
+   * let num = buffer.fastCeil(3.4); // -->  4
+   */
+  public fastCeil(n: number): number {
+    let f = (n << 0);
+    return f === n ? f : f + 1;
+  }
+  /**
```
Tyriar: Style nit: Empty line here

src/InputHandler.ts
```diff
@@ -18,80 +18,64 @@ export class InputHandler implements IInputHandler {
...
-          this._terminal.updateRange(this._terminal.y);
-        }
-        return;
-      }
 
-      // goto next line if ch would overflow
-      // TODO: needs a global min terminal width of 2
-      if (this._terminal.x + ch_width - 1 >= this._terminal.cols) {
-        // autowrap - DECAWM
-        if (this._terminal.wraparoundMode) {
```
Tyriar: I think we still want to respect this

src/Interfaces.ts
```diff
@@ -60,7 +61,7 @@ export interface ILinkifier {
   deregisterLinkMatcher(matcherId: number): boolean;
 }
 
-interface ICircularList<T> {
+interface IBuffer<T> {
```
Tyriar: Keeping the IBuffer and ICircularList interfaces separate would be cleaner, you should be able to do this:

```
interface IBuffer extends ICircularList<string> {
...
}
```

Also I don't think IBuffer should be generic.

src/Interfaces.ts
```diff
@@ -23,7 +23,8 @@ export interface ITerminal {
   textarea: HTMLTextAreaElement;
   ybase: number;
   ydisp: number;
-  lines: ICircularList<string>;
+  scrollBase: number;
```
Tyriar: It used to be an `ICircularList<string>` but I'm not sure this is true, I think lines is an `ICircularList<any[]>`? (Until we move xterm.js fully to TS)

src/xterm.js
```diff
@@ -1230,7 +1241,7 @@ Terminal.prototype.scrollToTop = function() {
  * Scrolls the display of the terminal to the bottom.
  */
 Terminal.prototype.scrollToBottom = function() {
-  this.scrollDisp(this.ybase - this.ydisp);
+  this.scrollDisp(this.scrollBase - this.ydisp);
```
Tyriar: What's the difference between ybase and scrollbase?

LucianBuzzo: `scrollBase` compensates for wrapped lines.

LucianBuzzo:
I'm closing this PR in favour of https://github.com/sourcelair/xterm.js/pull/644

LucianBuzzo:
This implement

Issue #599, opened 2017-03-14
Support chrome-devtools scheme in web link matcher


Tyriar:
VS Code issue: https://github.com/Microsoft/vscode/issues/21668

Tyriar:
Probably best deferred to https://github.com/sourcelair/xterm.js/issues/583 as the http link handler stuff will be replaced in that.

Tyriar:
Merging into https://github.com/xtermjs/xterm.js/issues/583

Issue #601, opened 2017-03-15
Could not linkify the URL with super-long TLDs.


tmyt:
Currently Linkifier implimentation could not linkify the URL with super-long TLDs.

Linkifier assume the TLD lenght in 2 - 6 chars in the here,
https://github.com/sourcelair/xterm.js/blob/master/src/Linkifier.ts#L14

But, some new gTLDs have over 6 chars. (ex: .enterprises, .foundation, and so on).
Needs to update `tldClause` to support these super-long TLD names.

## Details
- xterm.js version: master (ff29fd6)

### Steps to reproduce

1.  echo 'https://example.enterprises'
2.  URL does not linked.

Tyriar:
PRs welcome, should be an easy fix :+1:

jens1o:
@Tyriar Would happy to contribute, for me 256 ought be enough for everybody? ;)

Tyriar:
@jens1o the concern with working more on the web link matcher as is, is that it will eventually be replaced by something much smarter that can understand things like when brackets () are within or outside of a URL, see https://github.com/sourcelair/xterm.js/issues/583

And that one's a bit more work 😛 as it's replacing all the regex logic with a parser that will probably work by finding the start of a link and then move forward until the end is found.

jens1o:
Oh, sorry. I think I won't meet your expectations... Let's hope there will be some nice contributor with some good skills :D

Tyriar:
No worries 😄

Tyriar:
Merging into https://github.com/xtermjs/xterm.js/issues/583

Issue #583, opened 2017-03-03
Use a proper parser for link detection


Tyriar:
There will always be issues with URL matching using regex as some cases simply can't be caught with regex. An example where regex will fail is including brackets in URLs but only when the brackets are opened within the url:

- Include brackets: `http://<domain>.com/foo(bar)`
- Don't include wrapping brackets: `(http://<domain>.com/foobar)`
- Include commas: `http://<domain>.com/foo,bar`
- Don't include trailing commas: `http://<domain>.com/foo, other text`
- Detect spaces (paths are ambiguous?) `c:\Users\X\My Documents`

egmontkob:
FYI: These particular cases are handled correctly by `gnome-terminal` using regexes (pretty complex ones though).

See [big rewrite bug](https://bugzilla.gnome.org/show_bug.cgi?id=756038), [balanced parentheses bug](https://bugzilla.gnome.org/show_bug.cgi?id=763980), [current implementation](https://git.gnome.org/browse/gnome-terminal/tree/src/terminal-regex.h), [unittests](https://git.gnome.org/browse/gnome-terminal/tree/src/terminal-regex.c).

Tyriar:
When someone tackles this issue there are some more cases on the VS Code issue tracking this https://github.com/Microsoft/vscode/issues/21125

Tyriar:
Merged some issues into this one to support chrome-devtools scheme https://github.com/xtermjs/xterm.js/issues/599 and long TLDs https://github.com/xtermjs/xterm.js/issues/601.

Tyriar:
Merging UNC path support into this, eg. `\\localhost\c$\GitDevelopment\standup\server\db.js`

graph:
wowi, looking at this thread this problem is so much harder than I thought. So many different formats. 

Currently relative paths without a "./" don't work e.g.

```
echo "README.md"
```
That doesn't work. But if I put a "./README.md" it works. Python sometimes outputs paths without a "./" and those errors do not link.

Tyriar:
mocha 5 uses this format (no initial ./):

```
src/test/testUtils.ts:5:37 - error TS2322: Type 'Node<K, V> | null' is not assignable to type 'Node<K, V>'.
  Type 'null' is not assignable to type 'Node<K, V>'.

5     public get root(): Node<K, V> { return this._root; }
                                      ~~~~~~~~~~~~~~~~~~
```

ElleshaHackett:
[...]

PR #611, opened 2017-03-17
Replace Linkifier RegExp to RFC3986 based one

tmyt:
Current Linkifier has some problems.
The PR replace Linkifier RegExp to RFC3986 based one, and aim to resolve issues #599 and #601.

These URLs are now Linkify.

- http://example.enterprises/
- http://[::1]/
- ftp://example.com/
- chrome-devtools://devtools/
- mailto:xxx@example.com
- about:blank

Could you discuss "should" or "should not" linkify the URI without "//"?


tmyt:
The code wrote myself with refer the Appendix A of [IETF RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) documentation.

Strictly, Unicode characters is not allowed in URI and should be encode with Percent or Punycode. Then this implementation does not support Unicode chars.

Tyriar:
@tmyt strictly, same with brackets which get encoded. Browsers and such will try to help the user out by allowing unencoding characters in URLs, there are also legal unicode domain names now. I wonder if doing this is even worth it as the real fix imo is to implement a proper parser to get the ideal behavior as in https://github.com/sourcelair/xterm.js/issues/583

tmyt:
@Tyriar Ah, I could not check #583 that implement in mean time, the PR is not needed. If not needed, please reject this.

Tyriar:
@tmyt while this would be an improvement, it's probably a good idea to not switch to this and then a parser. Thanks for understanding :smiley: something like this is what I'm thinking for the parser https://github.com/Microsoft/vscode/blob/master/src/vs/editor/common/modes/linkComputer.ts

PR #613, opened 2017-03-17
Implement getState method

### LINKED ISSUE #595: Request: Headless Mode
Hi everyone,

With `xterm.js` maturing, I think a great feature to expand the range of applications `xterm.js` could be useful for would be the formalization of a headless mode. What do I mean? Well basically, for the ability to:

1. Run a client without a renderer.
2. Communicate to a client *with* a renderer the state of a headless client and vice versa.

Right now, the only way to reliably "set" the state of a client is to replay the set of all commands that produced that editor's state (at least via the public API). However, you could easily imagine wanting to do things like:

* scrubbing back and forth through a terminal's history
* running a terminal in the cloud that you intermittently connect to via the client
* (my use case) having multiple users have *literally* the same terminal with the ability to come and go

A lot of this is basically almost already possible if you just serialize `term.rows` and assorted state flags, but the adoption of a formal API would help a lot. It may also help improve the testability and robustness of the core library, too!

Thoughts?
Vincent

parisk:
This is a PoC for a `getState` terminal method that will return the serialized current state of the terminal.

## Things to include in state

- [x] Geometry
- [x] Mode
- [x] Buffer
- [x] Alt buffer
- [x] Cursor position
- [x] Cursor style
- [x] Options
- [ ] Version

## To do

- [x] Add all members in state
- [ ] Implement tests for `getState`

## State size
It seems that the serialized state of the terminal can get kinda big, relatively easily. This happens because of the (verbose) way we are storing the buffers.

Maybe we should entertain the idea of keeping less information in memory and computing some things on the fly instead, considering buffers.

### Example
1. Open the xterm.js demo
2. Run `less typings.json` in the terminal
3. Then use `term.getState()` in the browser's dev tools to retrieve the state
4. The resulting state is almost 250KBs and 25k+ lines of prettified JSON data

### Screenshot
image

This PR should eventually close #595.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -266,6 +266,74 @@ function Terminal(options) {
 inherits(Terminal, EventEmitter);
 
 /**
+ * Returns the current position and style of the terminal cursor.
+ * It returns an object in the following form: {position: [x, y], style: "cursorStyle"}
+ */
+Terminal.prototype._getCursor = function() {
+  return {
+    position: [this.x, this.y],
+    style: this.getOption('cursorStyle')
+  };
+};
+
+
+/**
+ * Returns the current mode of the terminal. Can be one of the following:
+ *   - application
+ *   - insert
+ *   - normal
+ */
+Terminal.prototype._getMode = function() {
+  if (this.applicationKeypad) {
+    return 'application';
+  } else if (this.insertMode) {
+    return 'insert';
+  }
+
+  return 'normal';
+};
+
+
+/**
+ * Serialize and return the terminal's state as a JSON object.
+ * Also iterate through all terminal properties that have a `getState` methods and embed their
+ * state as well.
+ */
+Terminal.prototype.getState = function() {
+  const properties = Object.keys(this);
+  const availableOptions = [
+    'cursorBlink', 'disableStdin', 'scrollback', 'tabStopWidth', 'useFlowControl'
+  ];
+
+  let state = {
+    cursor: this._getCursor(),
+    geometry: this.geometry,
+    mode: this._getMode(),
+    options: {}
+  };
+
+  availableOptions.forEach(function (option) {
+    state.options[option] = term.getOption(option);
+  });
+
+  // Iterate through all terminal properties to embed their state as well.
+  for (let i in properties) {
+    let key = properties[i];
+    let value = this[key];
+
+    if (!value) {
+      continue;
+    }
+
+    if (typeof value.getState == 'function') {
+      state[key] = value.getState();
+    }
+  }
+
+  return state;
+};
+
+/**
  * back_color_erase feature for xterm.
  */
 Terminal.prototype.eraseAttr = function() {
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -305,8 +305,20 @@ Terminal.prototype.getState = function() {
     'cursorBlink', 'disableStdin', 'scrollback', 'tabStopWidth', 'useFlowControl'
   ];
 
+  let normalBuffer = this.lines._array;
+  let altBuffer = null;
+
+  // This means that the alt buffer is active
+  if (this.normal) {
+    altBuffer = normalBuffer;
+    normalBuffer = this.normal.lines._array;
+  }
+
   let state = {
-    buffer: this.lines._array,
+    buffers: {
+      normal: normalBuffer,
+      alt: altBuffer
+    },
     cursor: this._getCursor(),
     geometry: this.geometry,
     mode: this._getMode(),
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -306,6 +306,7 @@ Terminal.prototype.getState = function() {
   ];
 
   let state = {
+    buffer: this.lines._array,
     cursor: this._getCursor(),
     geometry: this.geometry,
     mode: this._getMode(),
--- a/demo/main.js
+++ b/demo/main.js
@@ -68,7 +68,7 @@ function createTerminal() {
   protocol = (location.protocol === 'https:') ? 'wss://' : 'ws://';
   socketURL = protocol + location.hostname + ((location.port) ? (':' + location.port) : '') + '/terminals/';
 
-  term.open(terminalContainer);
+  term.open(terminalContainer, true);
   term.fit();
 
   var initialGeometry = term.proposeGeometry(),


src/xterm.js
```diff
@@ -251,6 +251,52 @@ function Terminal(options) {
 inherits(Terminal, EventEmitter);
 
 /**
+ * Returns the current mode of the terminal. Can be one of the following:
```
Tyriar: If only 1 of these modes can be active at once we should turn that into an enum.

src/xterm.js
```diff
@@ -251,6 +251,52 @@ function Terminal(options) {
...
+  // Iterate through all terminal properties to embed their state as well.
+  for (let i in properties) {
+    let key = properties[i];
+    let value = this[key];
+
+    if (!value) {
+      continue;
+    }
+
+    if (typeof value.getState == 'function') {
```
Tyriar: :+1: when `Terminal` is in TS I think we can do if `value instanceof ISerializable` which is even nicer.

src/xterm.js
parisk: Yep, I was thinking about that, but I am not exactly sure how this works right now.

I took a look at the following links and I was not able to find any reference to the `origin` mode, so I ditched it. Also I cannot understand if insert mode can co-exist with application mode:

- http://invisible-island.net/xterm/manpage/xterm.html
- http://invisible-island.net/xterm/ctlseqs/ctlseqs.html

parisk:
@Tyriar I am thinking about whether I should include the buffers as "plain text to be rendered" or as an array of lines (each line can either be a string, or an array of characters).

I have not decided yet, but my gut tells me that having an array of lines (each line should be a string) should do the work pretty well.

Any opinions on this?

mofux:
@parisk My use cases for this would be to:

1. store the state of the terminal when the application quits, and restore the state including the scrollback buffer etc. when the user reopens the app. Additionally, If I can't get hold of the old tty session, I would spawn a new tty in the background, and write a one-line notice to the terminal that indicates that the session was re-created. 

2. provide a viewfinder, similar to what code editors like Sublime Text have on the right hand side of the code pane, where you can see a zoomed out version of the code, and allow you to scroll through  it using a view finder. To get this working, It would massively help if I would get a line representation that can be easily transformed into html, e.g.

```javascript
lines: [
  // one line, saying "hello world", 
  // "hello" has green foreground and white background (ascii color)
  [
    { text: 'hello', fg: 'green', bg: 'white', bold: false, underline: false },
    { text: ' world': fg: 'fg', bg: 'bg', bold: false, underline: false }
  ]
]
```

In general, having some sort of parser that can be used to transform raw tty data into something that can be easily rendered by a browser would be very useful.

Tyriar:
@mofux 

1. Thanks for sharing your use case
2. xterm.js has enough performance challenges without considering a minimap imo, the fact that the output changes so frequently and so much means that the minimap would have to be re-rendered a lot. Sure it's possible but there are better areas of opportunity to focus on

vincentwoo:
I'm excited to see this PR progress along!

parisk:
OK, I just pushed an update here, that makes `getState()` fully-functional for a v1 IMO.

My only thought here is that the buffers can get really really big, really quickly. This implies that:

1. The serialized state of the terminal can get king-size 
2. This will make testing (and debugging) of serialized buffers cumbersome (thinking of using pre-baked `.json` files with already serialized states

Any ideas on how we could work around better with this issue?

If nothing comes, I will move on with implementing tests utilizing static files for the buffer data.

src/xterm.js
```diff
@@ -260,6 +260,87 @@ function Terminal(options) {
...
+
+  // This means that the alt buffer is active
+  if (this.normal) {
+    altBuffer = normalBuffer;
+    normalBuffer = this.normal.lines._array;
+  }
+
+  let state = {
+    buffers: {
+      normal: normalBuffer,
```
Tyriar: We should create a `CircularList.getState` function which returns data from indexes 0 to `CircularList.length`. That's all the data that's meaningful:

screen shot 2017-05-27 at 3 14 09 pm

With the 24 lines from the CircularList it should be easy to create a new CircularList constructor or `setState` method which sets the length to 24 and puts those 24 lines into the list.

src/xterm.js
```diff
@@ -260,6 +260,87 @@ function Terminal(options) {
...
+    mode: this._getMode(),
+    options: {}
+  };
+
+  availableOptions.forEach(function (option) {
+    state.options[option] = term.getOption(option);
+  });
+
+  // Iterate through all terminal properties to embed their state as well.
+  for (let i in properties) {
```
Tyriar: Does this retrieve any data yet? Maybe we should comment it out if it's not used yet?

vincentwoo:
@parisk I imagine there's pretty straightforward compression you can do if many entries in the buffer share the same mode?

Tyriar:
@vincentwoo if size matters I was thinking consumers could always gzip it or something. Seems like the better way to go to avoid additional complexity in the codebase imo.

vincentwoo:
Compression in-lib is usually better because a problem often ends up being de/serialization costs. If you like, I can probably write a compression utility for the data format we can include as an addon?

Tyriar:
@vincentwoo that might make better sense as an addon, sounds like a good idea to me 👍

vincentwoo:
Okay, is the format stable enough for me to program against yet or should I hold off?

Tyriar:
@vincentwoo I'd leave that up to @parisk, I figured you would be piping it through a common compression algorithm like gzip and back again in which case the format shouldn't matter?

vincentwoo:
No, I'm talking about format-aware stuff. For instance, let's say you have an array buffer like

`[ ['a,', 485499], ['b,', 485499], ['a,', 485499] ... ]`

where the mode number is always the same. the problems are:

1. gzip can tokenize the common number, but still has to express something like ['a', TOKEN], ['b', TOKEN], etc.
2. the server has to pay the penalty of stringifying a large datastructure
3. the client has to pay the penality of deserializing a large datastructure

instead, of we represent this as something like ['a', 'b', 'a'] with a side-structure of "mode 485499 for the whole line", the client and server can both go much faster when producing the blob to send over the wire. the smaller payload gzipped will also be smaller than the larger blob gzipped, because we have simplified entropy in a way that gzip is not capable of figuring out.

Tyriar:
Maybe this should just go in Buffer.getState/setState post https://github.com/sourcelair/xterm.js/pull/717

If the serialize functions are nicely separated it should be pretty simple.

vincentwoo:
Should we do it before then or before a release? Changes to the buffer
format are breaking between releases at that point.

On Jun 21, 2017 3:48 PM, "Daniel Imms" <notifications@github.com> wrote:

Maybe this should just go in Buffer.getState/setState post #717
<https://github.com/sourcelair/xterm.js/pull/717>

If the serialize functions are nicely separated it should be pretty simple.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/sourcelair/xterm.js/pull/613#issuecomment-310226679>,
or mute the thread

Tyriar:
@vincentwoo I'd wait until at least this and #717 are merged.

The buffer format may indeed change in the future as it's still using the original character format from term.js which is a little strange (`[number, string, number]`), we may also want to store more metadata some day as well. Certainly don't want to lock ourselves in to anything.

parisk:
@vincentwoo let's hold off until we merge the new buffer classes into `master` and the first implementation of `getState` and `setState`, both of which turned out to need significantly more work than I initially thought 😅 .

After that, I think it will be pretty much easier to iterate over performance improving, benchmarking etc.

ioquatix:
I would find this useful in my Atom package `script-runner` as it needs to save and restore window state.

parisk:
@ioquatix this will hopefully land in 3.0 🙂.

Tyriar:
I think I mentioned it before but I think we should only move forward with this if we allow breakages across versions. Otherwise this will severely impact our ability to improve the buffer and other components in the future.

We could do this in a way that minimizes breakages for users though, the state could have a `version` property which declares the breakage. So when consumers try to use the state, maybe it would return false if the version is not what is expected?

vincentwoo:
I am perfectly fine with that, I don't want to constraint xterm.js development speed at all. I think anyone who uses this feature is liable to be a savvier user and probably can deal with the consequences.

ioquatix:
In the case of script-runner, it's really not that important, it's more like, nice to have it working correctly - if the version is updated and some old panels fail to reload correctly, it's not a big deal.

ioquatix:
I can't tell you how much I'm looking forward to this PR :)

parisk:
@Tyriar I think it's OK to introduce breaking changes when we have a new major version out (e.g. 4.0).

Considering versioning the serialized state, I think it's a great idea, so I'll put it in a todo.

Tyriar:
@parisk I think we should allow "breaking changes" to the getState data even between minor versions, not doing so could be overly restricting. Basically defer the breakage to a separate schema version contained in the `getState` json which may change a between versions.

parisk:
@Tyriar do you have any components in might that might need to break their serialized form more than once in (let's say) a year?

Tyriar:
@parisk buffer improvements https://github.com/sourcelair/xterm.js/issues/791 and true color https://github.com/sourcelair/xterm.js/issues/484 are the main ones that come to mind immediately. 

Additional styles would change the format, perhaps in a backwards friendly way though https://github.com/sourcelair/xterm.js/issues/580

parisk:
Closing this, as I will open a fresh one for `3.0.0`.

ioquatix:
Was this ever implemented?

parisk:
@ioquatix, nope. This was not implemented at all. PRs are welcome though. I would love to review.

ioquatix:
Okay. I don't really know enough about it nor do I have the time :(

PR #615, opened 2017-03-18
Added Spyder IDE Terminal as a xterm case-use application

andfoy:
At Spyder IDE organization we've implemented a embed version of xterm.js inside a Qt Widget, this project, named [Spyder Terminal](https://github.com/spyder-ide/spyder-terminal) allows users to interact with a terminal inside the IDE

captura de pantalla de 2017-03-18 15-54-00


parisk:
Awesome, thanks a lot @andfoy!

PR #616, opened 2017-03-19
Prevent line data from being discarded on resize.

LucianBuzzo:
One of the core issues I'm trying to address in this PR https://github.com/sourcelair/xterm.js/pull/609 is that xterm.js will discard line data when resizing to smaller widths. As the discussion around reflow behaviour and architecture is still ongoing and may be for some time yet, I'm submitting this PR as an interim fix.
This PR prevents the character discard and uses behaviour similar to hyperterm.

linepreserve

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -1805,14 +1805,8 @@ Terminal.prototype.resize = function(x, y) {
         this.lines.get(i).push(ch);
       }
     }
-  } else { // (j > x)
-    i = this.lines.length;
-    while (i--) {
-      while (this.lines.get(i).length > x) {
-        this.lines.get(i).pop();
-      }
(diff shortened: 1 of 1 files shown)


parisk:
Thanks for this @LucianBuzzo!

blink1073:
@parisk, is it possible to cut a bug fix release with this change?

parisk:
@blink1073 this cannot happen, since we have pushed several commits with "new features" after we issued the release, so this wouldn't be a pure bugfix release.

You can either clone the repo and build it on your own, or just wait until the first week of April, for the 2.5 release.

blink1073:
No worries, had to ask :wink:.  Thanks!

PR #621, opened 2017-03-30
Added Cloud Commander to real world uses

coderaiser:
[Cloud Commander](https://cloudcmd.io) uses `xterm` as an embedded terminal which can be additionally installed and enabled (with help of [gritty](https://github.com/cloudcmd/gritty)). 
menu

Can be opened with `Shift + ~` and closed with `Shift + Esc`.

terminal


Tyriar:
Thanks @coderaiser, looks cool :smiley:

PR #624, opened 2017-03-31
Fix typos in comment

InDieTasten:


Tyriar:
Wow I was having a bad day then :sweat_smile: Thanks @InDieTasten!

InDieTasten:
@Tyriar No worries! As long as it's only in comments 😄

PR #626, opened 2017-04-03
Use setInterval over animation for cursor blink

### LINKED ISSUE #625: Animation performance on Chrome is terrible we should prefer setInterval for cursorBlink
See https://github.com/Microsoft/vscode/issues/22900

Verified it's a problem for xterm.js by looking at the Chrome task manager, idle CPU jumps from 1% with no blinking to 5-13% with blinking.

Tyriar:
Fixes #625

### DIFF
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -97,16 +97,9 @@
     background-color: transparent;
 }
 
-.terminal:not(.xterm-cursor-style-underline):not(.xterm-cursor-style-bar).focus.xterm-cursor-blink .terminal-cursor {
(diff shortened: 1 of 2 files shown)


rebornix:
LGTM. On my MacBook Pro (Retina, 15-inch, Mid 2015) with Processor 2.2 GHz Intel Core i7 , the blinking cursor in latest Chrome in `interval` eats around 1% CPU, not perfect but still better than animations.

Tyriar:
@parisk hopefully they will animations like this more efficient in the future

PR #628, opened 2017-04-06
Added Codevolve.com to README

saulcosta18:
We use xterm.js for our courses at Codevolve.


Tyriar:
Thanks @saulcosta18 :smile:

PR #632, opened 2017-04-12
Workaround IME composition bug in Chrome v55 and v56

rebornix:
Bug in Chrome v55 and v56: compositionupdate .data is broken so in these two versions of Chrome, in composition state, unconfirmed text of CJK IME collapse to one character:

terminalcjk

Electron shipped their latest version with Chrome v56 so anyone who is adopting the latest version of Chrome can run into this IME bug. We already fixed that for editor in VS Code https://github.com/Microsoft/vscode/issues/24557 but we may still want to fix this in xterm.js for Integrated Terminal.

### DIFF
--- a/src/CompositionHelper.ts
+++ b/src/CompositionHelper.ts
@@ -10,6 +10,7 @@ interface IPosition {
 }
 
 const isChromev55v56 = (
+  typeof navigator === 'object' &&
   (navigator.userAgent.indexOf('Chrome/55.') >= 0 || navigator.userAgent.indexOf('Chrome/56.') >= 0)
   /* Edge likes to impersonate Chrome sometimes */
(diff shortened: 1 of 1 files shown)


rebornix:
The last commit fixed https://github.com/sourcelair/xterm.js/issues/634. Whenever we don't trust the composition update from browser, we have to get the correct composition text/state from textarea. It's not perfect but it works.

Tyriar:
As discussed in person we'll close this off since it introduces other regressions. We'll just live with the bug in Chrome 55/56 https://bugs.chromium.org/p/chromium/issues/detail?id=677050&can=1&q=blink%20component%3ABlink%3EEditing%3EIME%20&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified

PR #635, opened 2017-04-14
Update README with RStudio under real-world uses.

gtritchie:
RStudio 1.1 is under development and uses xterm.js for an integrated terminal on Mac, Windows, Linux, and server.

Daily builds: https://www.rstudio.org/download/daily/


parisk:
Thanks!

PR #636, opened 2017-04-16
Implement < ESC [ 21 m > (un-bold)

dolmen:
*`ESC`*`[21m` is the reverse of *`ESC`*`[1m`.

Test case from a Unix shell: `printf "normal\x1B[1mbold\x1B[21mnormal\n"`

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -1148,6 +1148,7 @@ export class InputHandler implements IInputHandler {
    *     Ps = 5  -> Blink (appears as Bold).
    *     Ps = 7  -> Inverse.
    *     Ps = 8  -> Invisible, i.e., hidden (VT300).
+   *     Ps = 2 1  -> Not bold.
    *     Ps = 2 2  -> Normal (neither bold nor faint).
    *     Ps = 2 4  -> Not underlined.
    *     Ps = 2 5  -> Steady (not blinking).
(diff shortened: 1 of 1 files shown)


Tyriar:
@dolmen according to our principle source (authored by a long time xterm developer) `[21m` represents doubly underline?

> Ps = 2 1  -> Doubly-underlined (ISO 6429).

http://invisible-island.net/xterm/ctlseqs/ctlseqs.html

jerch:
For the old DEC VTs `CSI 22 m` is the code to reset bold. `CSI 21 m` is not defined there and got somewhat overloaded by later implementations. If you merge em all up for 21 it would read as
- reset bold or faint to normal
- set double underscore

Not sure if there is any terminal (emulator) out there supporting it this way.

Tyriar:
My feeling here is to go with the current popular terminal emulators. I only have a mac on me right now but Terminal.app doesn't seem to reset bold. Thoughts @jerch?

jerch:
`CSI 22 m` does not reset bold in xterm and Terminal.app. It seems to be supported by most others - libvte emulators support it (gnome-terminal, xfce-terminal etc.), linux kernel console does as well.
Note, this PR does not add a missing feature, `CSI 21 m` seems to reset bold on all terminals with support for bold.
Since xterm.js somewhat implies xterm compatibility I would not implement it.

Tyriar:
@jerch I agree, thanks for your input :smiley:

@dolmen I'll close this off as we don't want to deviate from xterm in this respect.

dolmen:
Agreed.

PR #643, opened 2017-04-23
Add "Terminal for Atom" to Real-World Uses

jsmecham:
I've just released the first version of an Xterm.js-based terminal for Atom that takes full advantage of Docks and Panes.


parisk:
Thanks!

PR #645, opened 2017-04-26
Add `focus` argument in `Terminal.prototype.open`

### LINKED ISSUE #640: Preventing page scroll on initialization
I'm trying to place terminal scrolled out of view on a page, however when it's initialized the page scrolls to the terminal element. I've been through the source code some and don't see a clean way of disabling this functionality ― any suggestions?

parisk:
Fix #640

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -607,8 +607,9 @@ Terminal.prototype.insertRow = function (row) {
  * Opens the terminal within an element.
  *
  * @param {HTMLElement} parent The element to create the terminal within.
+ * @param {boolean} focus Focus the terminal, after it gets instantiated in the DOM
  */
-Terminal.prototype.open = function(parent) {
+Terminal.prototype.open = function(parent, focus) {
   var self=this, i=0, div;
 
   this.parent = parent || this.parent;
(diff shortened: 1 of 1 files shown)


parisk:
Opened up a documentation PR as well: xtermjs/xtermjs.org#20.

Tyriar:
I don't think we should add an option to the API, isn't this the same as doing this?

```js
term.open(el);
term.focus();
```

We could call it out in the release notes as a minor change, I don't see it breaking many applications though.

parisk:
@Tyriar, this will break all applications that use xterm.js actually, since the focus will be elsewhere now.

Since this is changing the way it works and it is a user facing issue, I don't want to break the API in a minor release. It's a very simple piece of code, which can go away in the next major release.

coderaiser:
Would be better to use `options` object instead of [last boolean argument](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap) in v3. This way:

```js
term.open(container, {
    focus: true
});
```
It is much cleaner and obvious then last boolean parameter.
What if there should be some other options, not only `focus` in a future?

Tyriar:
A new object might cause confusion with `Terminal(options)`. I would prefer if we removed the parameter outright in v3 so it's ignored.

parisk:
I agree. Keeping this in `options` sounds quite better.

PR #652, opened 2017-05-03
Force failing mocha tests to return a exit code of 1

### LINKED ISSUE #651: npm test returns an exit code of 0 when it fails
See https://github.com/sourcelair/xterm.js/pull/644, https://travis-ci.org/sourcelair/xterm.js/jobs/228226085

Tyriar:
Fixes #651

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -83,6 +83,7 @@ gulp.task('instrument-test', function () {
 gulp.task('mocha', ['instrument-test'], function () {
   return gulp.src([`${outDir}/*test.js`, `${outDir}/**/*test.js`], {read: false})
       .pipe(mocha())
+      .once('error', () => process.exit(1))
       .pipe(istanbul.writeReports());
 });
 


gulpfile.js
```diff
@@ -83,6 +83,7 @@ gulp.task('instrument-test', function () {
 gulp.task('mocha', ['instrument-test'], function () {
   return gulp.src([`${outDir}/*test.js`, `${outDir}/**/*test.js`], {read: false})
       .pipe(mocha())
+      .once('error', () => process.exit(1))
```
parisk: I got only two questions and this is good to go:

1. Does this stop the test suit completely if an error occurs at some time? (I would consider this blocking)
2. Does this prevent instanbul from writing coverage reports when tests fail? (Non-blocking)

Tyriar: 1. I assume so yes
2. Yes

Don't we want it to fail completely? Partial coverage reports aren't very useful imo.

PR #659, opened 2017-05-09
Normalize line endings on paste

### LINKED ISSUE #660: Line endings cause issues when pasting on Windows
Same issue as https://github.com/Microsoft/vscode/issues/22887, we should handle it in the paste handler.

jdanyow:
Enables pasting multi-line text into the terminal.

cc @EisenbergEffect, @yangl900, @DuncanmaMSFT

Fixes #660

### DIFF
--- a/src/handlers/Clipboard.ts
+++ b/src/handlers/Clipboard.ts
@@ -39,6 +39,17 @@ export function prepareTextForClipboard(text: string): string {
 }
 
 /**
+ * Prepares text to be pasted into the terminal by normalizing the line endings
+ * @param text The pasted text that needs processing before inserting into the terminal
+ */
+export function prepareTextForTerminal(text: string, isMSWindows: boolean): string {
+  if (isMSWindows) {
+    return text.replace(/\r?\n/g, '\n');
+  }
+  return text;
+}
+
+/**
  * Binds copy functionality to the given terminal.
  * @param {ClipboardEvent} ev The original copy event to be handled
  */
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@Tyriar This the PR I mentioned in our email exchange earlier :)

src/handlers/Clipboard.ts
```diff
@@ -68,6 +76,7 @@ export function pasteHandler(ev: ClipboardEvent, term: ITerminal) {
   let text: string;
 
   let dispatchPaste = function(text) {
+    text = prepareTextForTerminal(text);
```
Tyriar: I think we should probably apply this only on Windows:

```ts
import { isMSWindows } from '../Browser';
...
if (isMSWindows {
  ...
}
```

Thoughts @jdanyow?

Tyriar:
I created an issue for tracking and amended the top comment so it will close off automatically :smiley:

src/handlers/Clipboard.ts
```diff
@@ -39,6 +39,17 @@ export function prepareTextForClipboard(text: string): string {
 }
 
 /**
+ * Prepares text to be pasted into the terminal by normalizing the line endings
+ * @param text The pasted text that needs processing before inserting into the terminal
+ */
+export function prepareTextForTerminal(text: string, isMSWindows: boolean): string {
```
jdanyow: this signature made the tests easy to implement- let me know if you'd rather see this rolled into dispatchPaste or some other form

Tyriar: Easy tests are good :ok_hand:

Tyriar:
@parisk coveralls doesn't seem to be completing, the build seems to have completed successfully though https://travis-ci.org/sourcelair/xterm.js/jobs/230548323

parisk:
Seems to be OK now 😄 .

Tyriar:
I'm seeing some weird behavior after this. For the following string:

```
a
b
c
```

Cmd pastes on same line:

image

Powershell pastes in reverse and the cursor is on the first line:

image

PR #674, opened 2017-05-27
Restore the "paste" terminal event

### LINKED ISSUE #671: paste still doesn't work
## Details
- OS version: 
- xterm.js version: 

Browser  chrome
OS  mac
xterm.js latest version

### Steps to reproduce

1.  run demo project use fake terminal without websocket

parisk:
Restores the `paste` terminal event (which is also used in the demo).

Fixes #671.

### DIFF
--- a/src/handlers/Clipboard.ts
+++ b/src/handlers/Clipboard.ts
@@ -82,6 +82,8 @@ export function pasteHandler(ev: ClipboardEvent, term: ITerminal) {
     text = prepareTextForTerminal(text, term.browser.isMSWindows);
     term.handler(text);
     term.textarea.value = '';
+    term.emit('paste', text);
+
     return term.cancel(ev);
   };
 


src/handlers/Clipboard.ts
```diff
@@ -82,6 +82,8 @@ export function pasteHandler(ev: ClipboardEvent, term: ITerminal) {
     text = prepareTextForTerminal(text, term.browser.isMSWindows);
     term.handler(text);
     term.textarea.value = '';
+    term.emit('paste', text);
```
Tyriar: What's listening for this event exactly?

parisk: The demo and any app that wants to consume the paste event in a single place, as it can occur both in the hidden textarea and the terminal element, depending on the browser.

PR #676, opened 2017-05-30
Use the Menlo font when possible

arcanis:
This font seems to work better than the alternatives on OSX.

Before:

After:

### DIFF
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -38,7 +38,7 @@
 .terminal {
     background-color: #000;
     color: #fff;
-    font-family: courier-new, courier, monospace;
+    font-family: Menlo, courier-new, courier, monospace;
     font-feature-settings: "liga" 0;
     position: relative;
 }


Tyriar:
@parisk maybe we should just use monospace here by default?

arcanis:
Monospace seems broken as well with unicode characters :(

You can check here if you want to make some tests: https://manaflair.github.io/mylittledom/demo/#complex-form

jerch:
my 2 cents on this:
Dont try to fix the monospace font nightmare in browsers by tagging it to this or that for some platform. It will work for your tested chars of interest and fail for others. Source of the problems are:
- fonts always miss some glyphs --> system font renderer will step in to get some "sane" output
- fonts dont agree on the right way to layout ancient glyphs --> while one font handles glyph XY nice other fonts dont (font creators cant be blamed since "nice" is not specified)
- system font renderer have different heuristics to solve the "uhmm, there is no glyph for this in the actual font", they are likely to try to get a glyph from some other font and can completely mess up monospace output

There is only one way to get around this font rendering mess - if you get some ancient char, render it offscreen, measure and try to fit it into the monospace grid.

parisk:
Hi @arcanis thanks for submitting this PR, but I do not think that there is any benefit to the project from using Menlo by default.

Plus, this would "break" existing installations that rely on our default type face.

I suggest you use Menlo in your own installation, by utilizing the `.xterm-...` classes, as we use Source Code Pro in SourceLair. It should be fairly easy.

Thanks again for submitting this!

PR #685, opened 2017-06-09
Inherit CSS properties to simplify styling of terminal

aradzie:


Tyriar:
This will cause issues on some bg/fg colors with the terminal colors (white, bright white, etc.) not contrasting well. I think it's best to leave this as is since consumers can always override the colors by creating more specific CSS rules.

PR #690, opened 2017-06-09
Skip failing tests of macOS and add to CI

### LINKED ISSUE #687: Fix tests on macOS and setup CI
Right now a few xterm conformance tests fail on macOS:

```
xterm output comparison t0003-line_wrap.in
xterm output comparison t0005-CR.in
xterm output comparison t0009-NEL.in
xterm output comparison t0503-zsh_ls_color.in
```

We should fix these and then add macOS to TravisCI.

Tyriar:
Fixes #687


parisk:
Seems to be crashing at "install" time because of node-gyp.

Tyriar:
Yeah something related to compilation of the `sleep` module. Not sure how to fix it right now.

Might be easier to look for alternatives here https://github.com/sourcelair/xterm.js/blob/795cdb52dc585ebdf5b47a93f50582e7ce954ad4/src/test/escape-sequences-test.js#L30

PR #704, opened 2017-06-14
Implement very basic find API

### LINKED ISSUE #553: Implement search / find API
I would like to request a buffer search API. This would highlight all instances inside the buffer which match the search with the current search position highlighted uniquely.

Alternatively if you would like to make this more generic and only provide the means by which one could implement search, I would like the ability to highlight parts of the buffer via the API as well as the ability to flatten the buffer ignoring any special colouring and character tags etc (right now when i use innerHTML, colouring breaks words and sentences searching, innerText for some reason is return character code 160 instead of 32 for spaces).

Tyriar:
This PR exposes 2 new functions on `Terminal`:

```ts
findNext(term: string): boolean
findPrevious(term: string): boolean
```

Currently they scroll to and select the matching search term (using the selection model introduced in https://github.com/sourcelair/xterm.js/pull/670).

Fixes #553 

image


src/SelectionManager.ts
```diff
@@ -156,6 +157,9 @@ export class SelectionManager extends EventEmitter {
     this._buffer = buffer;
   }
 
+  public get selectionStart(): [number, number] { return this._model.finalSelectionStart; }
```
parisk: Can you document these with JSDoc please?

src/SelectionManager.ts
```diff
@@ -566,14 +521,22 @@ export class SelectionManager extends EventEmitter {
     return charIndex;
   }
 
+  public setSelection(col: number, row: number, length: number): void {
```
parisk: Can you document this with JSDoc please?

Tyriar:
👍 I can look into pulling this out into an addon. It will have to use private API. as long as we have good testing to prevent regressions that shouldn't matter though.

parisk:
Since we are the ones implementing and maintaining it, I think it's totally OK to use private APIs in add-ons.

Tyriar:
@parisk any comments on the current approach? I managed to get search working as both the first multi-file and first TypeScript addon. Right now we can't touch anything outside of `addons/search`, including interfaces, there's probably a way to share interfaces nicely but not sure atm.

TODO:

- [ ] Test non-browser environments
- [ ] Clean up the gulpfile.js
- [ ] jsdoc

parisk:
@Tyriar the implementation looks good to me 👍 . I also liked that this introduces the first TypeScript add-on 😄 .

A couple of to-dos **after merging this PR**:

- Pulling BufferLine into the Buffer class
- Finding a better way to write add-ons, than encapsulating everything in a closure

Tyriar:
@parisk ready for review, after wrestling with Travis for 30 minutes 😛

parisk:
Wooo 🎉 !

PR #708, opened 2017-06-14
add powershell instructions

jdanyow:


Tyriar:
Not sure these instructions are necessary, the better way to do it is to simply run powershell, bash.exe, etc. from the cmd.ex session. Maybe more generic instructions would be better, indicating you can launch zsh, powershell, etc. by running it in the demo?

jdanyow:
doh- make sense! closing this.

PR #717, opened 2017-06-19
Introduce Buffer and BufferSet classes

parisk:
This PR introduces the `Buffer` and `BufferSet` classes to help:

1. Organize all buffer information into a single place (`Buffer` class)
2. Manage a set of `normal` and `alt` buffer (`BufferSet` class)
3. Know when a buffer gets activated (`BufferSet` events)

## To-do
- [x] Replace usage of `lines` with `Buffer` everywhere
- [x] Get existing tests work
- [x] Write tests for `Buffer` and `BufferSet` classes
- [x] Document `Buffer` and `BufferSet` classes

This will eventually help ship #613.

### DIFF
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -13,7 +13,7 @@ import { CircularList } from './utils/CircularList';
  *   - scroll position
  */
 export class Buffer {
-  public lines: CircularList<[string, number, string]>;
+  public lines: CircularList<[number, string, number][]>;
 
   /**
    * Create a new Buffer.
@@ -33,7 +33,7 @@ export class Buffer {
     public scrollTop: number = 0,
     public tabs: any = {},
   ) {
-    this.lines = new CircularList<[string, number, string]>(this.terminal.scrollback);
+    this.lines = new CircularList<[number, string, number][]>(this.terminal.scrollback);
     this.scrollBottom = this.terminal.rows - 1;
   }
 }
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -29,7 +29,7 @@ export class Buffer {
     public scrollTop: number = 0,
     public tabs: any = {},
   ) {
-    this.lines = new CircularList<string>(this.terminal.scrollback);
+    this.lines = new CircularList<[string, number, string]>(this.terminal.scrollback);
     this.scrollBottom = this.terminal.rows - 1;
   }
 }
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -10,7 +10,6 @@ import { CircularList } from './utils/CircularList';
  */
 export class Buffer {
   public lines: CircularList<any>;
-  private _tabs: any;
 
   /**
    * Create a new Buffer.
@@ -26,7 +25,11 @@ export class Buffer {
     public ybase: number = 0,
     public y: number = 0,
     public x: number = 0,
+    public scrollBottom: number = 0,
+    public scrollTop: number = 0,
+    public tabs: any = {},
   ) {
     this.lines = new CircularList(this.terminal.scrollback);
+    this.scrollBottom = this.terminal.rows - 1;
   }
 }
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -9,7 +9,7 @@ import { CircularList } from './utils/CircularList';
  * This class represents a terminal buffer (an internal state of the terminal)/
  */
 export class Buffer {
-  public lines: CircularList<any>;
+  public lines: CircularList<string>;
 
   /**
    * Create a new Buffer.
@@ -29,7 +29,7 @@ export class Buffer {
     public scrollTop: number = 0,
     public tabs: any = {},
   ) {
-    this.lines = new CircularList(this.terminal.scrollback);
+    this.lines = new CircularList<string>(this.terminal.scrollback);
     this.scrollBottom = this.terminal.rows - 1;
   }
 }
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -9,7 +9,7 @@ import { CircularList } from './utils/CircularList';
  * This class represents a terminal buffer (an internal state of the terminal)/
  */
 export class Buffer {
-  private _lines: CircularList<any>;
+  public lines: CircularList<any>;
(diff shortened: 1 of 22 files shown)


src/InputHandler.ts
```diff
@@ -519,8 +519,8 @@ export class InputHandler implements IInputHandler {
   public scrollUp(params: number[]): void {
     let param = params[0] || 1;
     while (param--) {
-      this._terminal.lines.splice(this._terminal.ybase + this._terminal.scrollTop, 1);
-      this._terminal.lines.splice(this._terminal.ybase + this._terminal.scrollBottom, 0, this._terminal.blankLine());
+      this._terminal.buffer.lines.splice(this._terminal.buffer.ybase + this._terminal.scrollTop, 1);
+      this._terminal.buffer.lines.splice(this._terminal.buffer.ybase + this._terminal.scrollBottom, 0, this._terminal.blankLine());
     }
     // this.maxRange();
     this._terminal.updateRange(this._terminal.scrollTop);
```
AndrienkoAleksandr: this._terminal.ScrollTop and this._terminal.ScrollBottom seems should be in the buffer too.

src/Buffer.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+import { ITerminal } from './Interfaces';
+import { CircularList } from './utils/CircularList';
+
+/**
+ * This class represents a terminal buffer (an internal state of the terminal)/
+ */
+export class Buffer {
+  public lines: CircularList<string>;
```
Tyriar: FYI this is a `CircularList<[string, number, string]>`. Also it will change soon, see https://github.com/sourcelair/xterm.js/pull/756

src/InputHandler.ts
```diff
@@ -900,7 +900,7 @@ export class InputHandler implements IInputHandler {
 
           // TODO: Why are params[0] compares nested within a switch for params[0]?
 
-          this._terminal.x10Mouse = params[0] === 9;
+          this._terminal.buffer.x10Mouse = params[0] === 9;
```
AndrienkoAleksandr: @parisk Are you sure this line should be changed?

src/InputHandler.ts
```diff
@@ -1094,7 +1078,7 @@ export class InputHandler implements IInputHandler {
         case 1000: // vt200 mouse
         case 1002: // button event mouse
         case 1003: // any event mouse
-          this._terminal.x10Mouse = false;
+          this._terminal.buffer.x10Mouse = false;
```
AndrienkoAleksandr: @parisk Are you sure this line should be changed?

src/Interfaces.ts
```diff
@@ -23,19 +23,18 @@ export interface ITerminal {
...
   browser: IBrowser;
   writeBuffer: string[];
   children: HTMLElement[];
   cursorHidden: boolean;
   cursorState: number;
-  x: number;
-  y: number;
   defAttr: number;
+  scrollback: number;
+  buffers: any;  // This should be a `BufferSet` class, but it would result in circular dependency
```
Tyriar: Let's use interfaces here so they're typed.

src/xterm.js
```diff
@@ -242,21 +219,27 @@ function Terminal(options) {
   // leftover surrogate high from previous write invocation
   this.surrogate_high = '';
 
+  // Create the terminal's buffers and set the current buffer
+  this.buffers = new BufferSet(this);
+  this.buffer = this.buffers.active;  // Convenience shortcut;
```
Tyriar: I think this should note:

// TODO: Change to a getter when moved to TypeScript

src/xterm.js
```diff
@@ -2301,9 +2284,13 @@ Terminal.prototype.reset = function() {
   this.options.cols = this.cols;
   var customKeyEventHandler = this.customKeyEventHandler;
   var cursorBlinkInterval = this.cursorBlinkInterval;
+  var inputHandler = this.inputHandler;
+  var buf = this.buffers;
```
Tyriar: Nit: this is the only one here that doesn't use the full name.

src/xterm.js
parisk: Are you aware if there any chances of this affecting performance if it becomes a getter since `term.buffer` is being accessed quite frequently?

src/Interfaces.ts
parisk: I tried this and it displays a weird error that `ICircularList` does not have a `_length` attribute 😕.

I added the `_length` attribute to `ICircularList` and it displays an error that this is a private attribute :trollface:.

I am not expert with these, so if you have any tip here it could help.

parisk:
This should be ready for a final review 😄. /cc @Tyriar

src/xterm.js
Tyriar: It looks like getters should be pretty fast, this TS:

```ts
export class CircularList<T> extends EventEmitter {
  ...
  public get maxLength(): number {
    return this._array.length;
  }
```

Compiles to:

```js
var CircularList = (function (_super) {
    ...
    Object.defineProperty(CircularList.prototype, "maxLength", {
        get: function () {
            return this._array.length;
        },
```

src/Interfaces.ts
Tyriar: Hmm, I would only expect that error if you added `_length` to the interface.

Tyriar: I fixed this up and pushed to this branch. The types were a bit messy, the list is actually a `CircularList<[number, string, number][]>` 😅

src/xterm.js
```diff
@@ -243,21 +220,27 @@ function Terminal(options) {
...
   this.surrogate_high = '';
 
+  // Create the terminal's buffers and set the current buffer
+  this.buffers = new BufferSet(this);
+  this.buffer = this.buffers.active;  // Convenience shortcut;
+  this.buffers.on('activate', function (buffer) {
+    this._terminal.buffer = buffer;
+  });
+
   /**
```
Tyriar: This comment is no longer relevant, should be safe to remove as I'm going to heavily document the buffer format in https://github.com/sourcelair/xterm.js/pull/756

Tyriar:
Also > 70% test coverage :tada:

PR #728, opened 2017-06-25
Fix turning from alt screen to normal screen and vice versa.

AndrienkoAleksandr:
Fix turning from alt screen to normal screen and vice versa. 
Fix https://github.com/sourcelair/xterm.js/issues/229.
Target branch: buffer-set

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -936,16 +936,14 @@ export class InputHandler implements IInputHandler {
           this._terminal.cursorHidden = false;
           break;
         case 1049: // alt screen buffer cursor
-          // this._terminal.saveCursor();
-          ; // FALL-THROUGH
+          this.saveCursor(params);
+          // FALL-THROUGH
         case 47: // alt screen buffer
         case 1047: // alt screen buffer
-          if (!this._terminal.normal) {
-            let normal = {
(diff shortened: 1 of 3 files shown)


AndrienkoAleksandr:
@parisk can you look?

src/InputHandler.ts
```diff
@@ -1440,8 +1441,8 @@ export class InputHandler implements IInputHandler {
    *   Save cursor (ANSI.SYS).
    */
   public saveCursor(params: number[]): void {
-    this._terminal.savedX = this._terminal.buffer.x;
-    this._terminal.savedY = this._terminal.buffer.y;
+    this._terminal.buffers.active.x = this._terminal.buffer.x;
```
parisk: `this._terminal.buffers.active` is the same with `this._terminal.buffer`.

I am not sure yet though if the `savedX` and `savedY` cursor positions have to be stored in the buffer or universally in the terminal.

Any thoughts?

parisk:
Thanks @AndrienkoAleksandr! This seems to be working.

Can you take a look at https://github.com/sourcelair/xterm.js/pull/728/files#r123939766 though, as I will need to change this piece of code?

src/InputHandler.ts
AndrienkoAleksandr: Ok I will look.

PR #736, opened 2017-06-27
Fix 'CR' issue on Linux based OS.

AndrienkoAleksandr:
Fix 'CR' issue on Linux based OS. Fix https://github.com/sourcelair/xterm.js/issues/735

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,18 +1,3 @@
 * text=auto eol=lf
 logo.png binary
-src/test/escape_sequence_files/t0005-CR.in eol=cr
-src/test/escape_sequence_files/t0011-RI_scroll.in eol=cr
-src/test/escape_sequence_files/t0014-CAN.in eol=cr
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1 +1,18 @@
 * text=auto eol=lf
+logo.png binary
+src/test/escape_sequence_files/t0005-CR.in eol=cr
```
Tyriar: Can we use patterns here? `src/test/escape_sequence_files/*` probably makes the most sense as these are test files that we can't touch.

.gitattributes
```diff
@@ -1 +1,3 @@
 * text=auto eol=lf
+logo.png binary
+src/test/escape_sequence_files/* binary
```
Tyriar: Also `text=auto` might be more appropriate as they're not binary files, I think that just allows whatever.

AndrienkoAleksandr: Ok, I think it's a make sense.

Issue #594, opened 2017-03-13
Support ballistic scrolling via touch


Tyriar:
This isn't supported currently because of how the viewport is actually underneath the row divs.

Tyriar:
Related to https://github.com/sourcelair/xterm.js/issues/512

Tyriar:
Scrolling with touch currently works, but it doesn't scrolling when you do a "flick" gesture.

Tyriar:
This is still needed, see https://github.com/xtermjs/xterm.js/pull/748 for a reference on where to get started.

Tyriar:
I think this will work now, just it's broken due to another problem: https://github.com/xtermjs/xterm.js/issues/5489

PR #747, opened 2017-07-01
Add scroll functionality for mobile

anishathalye:
This patch adds functionality to be able to scroll on mobile (tested with an iOS device).

### DIFF
--- a/src/Viewport.ts
+++ b/src/Viewport.ts
@@ -123,10 +123,18 @@ export class Viewport {
     ev.preventDefault();
   };
 
+  /**
+   * Handles the touchstart event, recording the touch occurred.
+   * @param ev The touch event.
+   */
   public onTouchStart(ev: TouchEvent) {
     this.lastTouchY = ev.touches[0].pageY;
   };
 
+  /**
+   * Handles the touchmove event, scrolling the viewport if the position shifted.
+   * @param ev The touch event.
+   */
   public onTouchMove(ev: TouchEvent) {
     let deltaY = this.lastTouchY - ev.touches[0].pageY;
     this.lastTouchY = ev.touches[0].pageY;
--- a/src/Viewport.ts
+++ b/src/Viewport.ts
@@ -13,6 +13,7 @@ export class Viewport {
   private currentRowHeight: number;
   private lastRecordedBufferLength: number;
   private lastRecordedViewportHeight: number;
+  private lastTouchY: number;
 
   /**
    * Creates a new Viewport.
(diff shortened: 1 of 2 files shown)


src/Viewport.ts
```diff
@@ -34,6 +34,7 @@ export class Viewport {
     this.terminal.on('scroll', this.syncScrollArea.bind(this));
     this.terminal.on('resize', this.syncScrollArea.bind(this));
     this.viewportElement.addEventListener('scroll', this.onScroll.bind(this));
+    this.touchY = null;
```
Tyriar: You'll need to declare this as a private member above to remove the compile warning. A better name might also be `lastTouchY`:

```ts
private lastTouchY: number;
```

src/xterm.js
```diff
@@ -1086,6 +1086,18 @@ Terminal.prototype.bindMouse = function() {
     self.viewport.onWheel(ev);
     return self.cancel(ev);
   });
+
+  on(el, 'touchstart', function(ev) {
+    if (self.mouseEvents) return;
+    self.viewport.onTouchStart(ev);
+    return self.cancel(ev);
+  }
```
Tyriar: Missing a `);` at the end

src/xterm.js
```diff
@@ -1086,6 +1086,18 @@ Terminal.prototype.bindMouse = function() {
...
+
+  on(el, 'touchmove', function(ev) {
+    if (self.mouseEvents) return;
+    self.viewport.onTouchMove(ev);
+    return self.cancel(ev);
+  }
```
Tyriar: Missing a `);` at the end

anishathalye:
Thanks for the feedback! I rebased onto the current master branch and also made the changes you asked for.

(for some reason, GitHub doesn't seem to recognize my `commit --amend` as a change...)

Let me know if there's anything else I should change.

Tyriar:
@anishathalye yeah with GitHub you typically add individual "fix" commits after a PR is made as amending a commit that has already been commented on can turn out a bit weird in GH. The repo owners can squash the commits if they want when it's merged.

tbodt:
The problem with this is it doesn't support the iOS scroll momentum thing. I've gotten that to work by removing these listeners, using z-index to put the viewport div above the rows div, and putting `-webkit-overflow-scrolling: touch` on the viewport div. Unfortunately, putting the viewport div above the rows div breaks selection.

tbodt:
Actually, it turns out that setting the z-index is unnecessary, and not setting the z-index makes selection work. So this is the perfect solution.

Tyriar:
@tbodt https://github.com/xtermjs/xterm.js/issues/594 was kept open since ballistic scroll doesn't work yet.

PR #748, opened 2017-07-02
Prototype for ballistic touch scroll

### LINKED ISSUE #594: Support ballistic scrolling via touch

Tyriar:
Fixes #594 

Before pushing, verify that:

- [ ] Cursor style is a text cursor when not in mouse mode
- [ ] Cursor style is a pointer when in mouse mode
- [ ] Cursor style is a pointer when mousing over the scroll bar
- [ ] Selection/scrolling works
- [ ] Mouse events work in mouse mode
- [ ] Verified in each supported browser https://github.com/sourcelair/xterm.js#browser-support


anishathalye:
What needs to be done before this can be merged?

Tyriar:
@anishathalye here's what I think needs to be done on this one:

- Fix the cursor style. There is meant to be a `text` cursor style when mouse mode is off and a `default` (pointer) cursor style when mouse mode is on. Note that the trivial fix here of `.terminal .xterm-viewport { cursor: text; }` is not ideal as it makes the cursor a text cursor when mousing over the scroll bar. This happens because `.xterm-viewport` is wider than `.xterm-rows`.
- Verify in each of the browsers supported. Chrome and safari seem to work.

Tyriar:
Closing this as it's pretty out of date and no plans to fix any time soon. It's always available for reference if someone wants to pick it up.

jdmayfield:
@Tyriar, I have a working prototype for touch-scrolling that can switch between xterm "mouse-wheel" scrolling in mouse-mode, or arrow-keys and regular scrolling not in mouse-mode.  I was able to disable regular scrolling when in mouse-mode by setting a touchmove listener on the parent div and calling event.stopPropagation() in it's handler.  It can detect touchmove in different "regions" by calculating xy position so theres a difference between an "edge-scroll" and  scrolling over the main area.  It detects mouse-mode on/off by intercepting the escapes for mouse-mode on/off before reaching the terminal data listener.  This seems to work pretty good, and allows scrolling like the Android Termux app, plus different behaviors for different places on the screen, yet still lets other touch events through (including touchmove over the scrollbar outside of mouse-mode).  Would this be something you guys are interested in, perhaps as an "addon"?

Tyriar:
@jdmayfield I'd like to make sure ballistic scroll is covered before we add more options like that. The main question I have is how best to expose these new option(s), little worried too much bloat in the options recently.

Related: https://github.com/xtermjs/xterm.js/issues/1101

PR #749, opened 2017-07-02
Fix issue with small scrollback value

### LINKED ISSUE #506: Prevent scrollback from being less than the terminal's row count
Downstream issue: https://github.com/Microsoft/vscode/issues/19193

Setting a scrollback smaller than the row count causes unexpected results since our scrollback is not the *actual* scrollback but the size of the buffer.

irokas:
Closes #506

### DIFF
--- a/src/test/test.js
+++ b/src/test/test.js
@@ -55,6 +55,11 @@ describe('xterm.js', function() {
     it('should throw when setting a non-existant option', function() {
       assert.throws(xterm.setOption.bind(xterm, 'fake', true));
     });
+    it('should not allow scrollback less than number of rows', function() {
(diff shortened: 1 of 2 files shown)


Tyriar:
I think we also need to cover the case where scrollback is the minimum and then the number of rows increases.

parisk:
Is it OK, if we

1. reject increasing the number of rows above the current `scrollback` value
2. return `false`
3. print a similar warning message

Tyriar:
Either that or setting to the scrollback to max(currentScrollback, rows)

Tyriar:
Fixed in https://github.com/sourcelair/xterm.js/pull/750

PR #750, opened 2017-07-02
Set minimum scrollback value to number of rows if needed on resize

irokas:
I was not able to implement a test because I could not get jsdom to work in test.js.

@Tyriar Any ideas on how I can write this test?


Tyriar:
@irokas yeah there aren't any tests for `resize` yet, hopefully we can get around to better testing on the `Terminal` class when the TypeScript migration finishes.

PR #772, opened 2017-07-10
Add search.exclude setting for vscode workspace

mjbvz:
Adds some simple vscode `search.exclude` settings for the workspace


Tyriar:
@parisk this would be very useful to add for users developing in VS Code.

Before:

image

After:

image

PR #773, opened 2017-07-10
Extract colors to ts

mjbvz:
Extracts the color constants to a ts file


Tyriar:
I actually think all the color-related code is going to be removed in https://github.com/sourcelair/xterm.js/pull/756 as in JS-land we no longer care about matching colors since we will either color using the 256 color keys (when explicitly set) or truecolor (which is set via `style=""`).

Issue #720, opened 2017-06-19
Selection should not change text color


Tyriar:
https://github.com/sourcelair/xterm.js/pull/719 made it so that the selection is on top in order to be able to show the selection on characters with their background set. Ideally we shouldn't use a transparent selection though and instead draw the characters on top of the selection. Perhaps we should pull out the rendering of the background colors so that rendering occurs like this:

1. background colors
2. selection
3. text

mofux:
I think we may be able to do that without splitting the dom into three layers - by using a css trick 🤓. I've created a simple example here:
https://jsfiddle.net/p6abybku/3/

The idea is to wrap the text inside an inner span which has `z-index: 1` and `position: relative`. The outer span will get the background color, and the inner span the foreground color.

I believe this will also only have a minimal impact on the performance, compared to rendering and keeping separate background / foreground layers.

Tyriar:
@mofux I didn't think about this trick :smile: we would need to wait for https://github.com/sourcelair/xterm.js/issues/654 if we wanted to try this out.

On the performance side, I'm thinking making them separate layers might actually end up being faster. The spans wrapping the text would need to be recreated every time the line changes, whereas for the new layer approach we could have x background elements that move/hide depending on the state. They would need to be present and have the layer trick even when there is no selection as the renderer doesn't have any knowledge of the selection.

Another perk of separation is that we could delay background layer positioning/rendering like how the evaluating links is delayed (say 50ms, ~3 frames?). This means that for a command that spits out a bunch of output (with/without bg colors) we would only need to recalc and render the background when there is a small break in rendering, meaning frames are rendered faster and therefore output is processed faster.

Tyriar:
I may be overthinking things though :wink:

mofux:
I think the biggest advantage with the `css` approach is that we would not have to manually set the size of the background layer spans - which we otherwise would have to set manually by measuring the text layer span's. Mmmh and maybe the outer bg span would only have to be drawn if there is actually a background set 🤔.

If I read the code correctly, we probably would have to optimise the `spanElementObjectPool` to have the wrapped fg/bg spans cached, otherwise we may end up with a similar unoptimised code path that is used for wide characters?
https://github.com/sourcelair/xterm.js/blob/master/src/Renderer.ts#L270

Anyway, awesome work on the virtual selection! Just wanted to point out this alternate approach 😊

Tyriar:
[...]

PR #776, opened 2017-07-11
Draw selection between background and foreground

mofux:
This is an attempt to have the selection always underneath the foreground, but above the background, see #720.

selection

**How it works**

~~If there is a background color involved (and only in that case), an additional `<span class="xterm-bg ..."></span>` is added to the span that contains the characters. This span in positioned absolutely and fills the space of the parent span. It has a negative z-index to get below the parent span's text.~~

To have the selection between background and foreground, I had the move the `.xterm-selection` container inside the `.xterm-rows` container. Although this works great, it doesn't belong there semantically 🤧. 

As the code changes are very small, this could be a temporary solution to #720 before a more efficient solution is introduced.

**Update**
I have found a much better way that will have no performance impact if selection is inactive, and only minor performance impact if selection is active. Instead of drawing the background in a separate span all the time, we create a `:before` pseudo element if selection is active (`.terminal.selection-active`). The `:before` element will fill the whole space of the span, draw the background and position itself underneath the text. I had to extend the `xterm-bg-...` css rules for this change.

Notice: Once truecolor support is implemented, this method will not work any longer. For truecolor we will likely have to create a real element for the background color, but for the time being this should get the job done. 🤓

### DIFF
--- a/src/Renderer.ts
+++ b/src/Renderer.ts
@@ -161,7 +161,6 @@ export class Renderer {
       // Return the row's spans to the pool
       while (this._terminal.children[y].children.length) {
         const child = this._terminal.children[y].children[0];
-        console.log('pool:release', child);
         this._terminal.children[y].removeChild(child);
         this._spanElementObjectPool.release(<HTMLElement>child);
       }
--- a/src/Renderer.ts
+++ b/src/Renderer.ts
@@ -257,7 +257,8 @@ export class Renderer {
               }
 
               if (bg < 256) {
-                innerHTML += `<span class="xterm-bg xterm-bg-color-${bg}"></span>`;
+                currentElement.classList.add('xterm-bg');
+                currentElement.classList.add(`xterm-bg-color-${bg}`);
               }
 
               if (fg < 256) {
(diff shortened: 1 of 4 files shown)


mofux:
**Update**
I have found a much better way that will have no performance impact if selection is inactive, and only minor performance impact if selection is active. Instead of drawing the background in a separate span all the time, we create a `:before` pseudo element if selection is active (`.terminal.selection-active`). The `:before` element will fill the whole space of the span, draw the background and position itself underneath the text. I had to extend the `xterm-bg-...` css rules for this change.

Notice: Once truecolor support is implemented, this method will not work any longer. For truecolor we will likely have to create a real element for the background color, but for the time being this should get the job done. 🤓

Tyriar:
> Notice: Once truecolor support is implemented, this method will not work any longer.

Truecolor https://github.com/sourcelair/xterm.js/pull/756 will hopefully go into either the next release or the one after, so we probably shouldn't move forward with this only to remove it before it's released or regress in the future.

mofux:
@Tyriar 👍 I'd like to keep this PR open and try to refine it once truecolor support has shipped.

Tyriar:
@mofux sounds good. While you're thinking about this, try think of ways that we could support selection inverting as well (as an option). I've always thought it was a really great feature because it looks cool and does a good job of ensuring the text readable https://github.com/sourcelair/xterm.js/issues/692

mofux:
I'm closing the as we will get this in using the new canvas renderer.

PR #779, opened 2017-07-11
Add SSHy as real-world use

stuicey:


Tyriar:
Thanks @stuicey :smiley:

PR #780, opened 2017-07-11
[SelectionManager] Use event.detail instead for counting the mousedown event

mofux:
This PR introduces the usage of MouseEvent.detail to get the number of consecutive clicks. 

See:
https://www.w3schools.com/jsref/event_detail.asp

This does not only reduce the complexity of the code base, but also respects the OS double click speed setting of the user.

### DIFF
--- a/src/SelectionManager.ts
+++ b/src/SelectionManager.ts
@@ -87,23 +87,6 @@ export class SelectionManager extends EventEmitter {
   private _dragScrollAmount: number;
 
   /**
-   * The last time the mousedown event fired, this is used to track double and
-   * triple clicks.
-   */
-  private _lastMouseDownTime: number;
-
-  /**
-   * The last position the mouse was clicked [x, y].
-   */
-  private _lastMousePosition: [number, number];
-
-  /**
-   * The number of clicks of the mousedown event. This is used to keep track of
-   * double and triple clicks.
-   */
-  private _clickCount: number;
-
-  /**
    * The current selection mode.
    */
   private _activeSelectionMode: SelectionMode;
@@ -135,7 +118,6 @@ export class SelectionManager extends EventEmitter {
     this.enable();
 
     this._model = new SelectionModel(_terminal);
-    this._lastMouseDownTime = 0;
     this._activeSelectionMode = SelectionMode.NORMAL;
   }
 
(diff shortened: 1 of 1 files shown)


Tyriar:
This is pretty cool, I'm a little concerned about browser support though as I don't trust w3schools and mdn says `?` for most browsers https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail

@rebornix you wouldn't happen to know why monaco doesn't use `UIEvent.detail` do you?

src/SelectionManager.ts
```diff
@@ -396,16 +378,14 @@ export class SelectionManager extends EventEmitter {
...
       this._onShiftClick(event);
     } else {
-      if (this._clickCount === 1) {
+      if (event.detail === 1) {
           this._onSingleClick(event);
-      } else if (this._clickCount === 2) {
+      } else if (event.detail === 2) {
           this._onDoubleClick(event);
-      } else if (this._clickCount === 3) {
+      } else if (event.detail === 3) {
```
Tyriar: Should this be >= to maintain current behavior?

mofux:
@Tyriar As far as I can tell, UIEvent.detail is well supported. I have tested it on Safari, Chrome, Firefox and IE10 a while back, all working. It is not very well documented, and I believe most people don't even know it exists. Actually, I have implemented a very similar solution to yours a few months ago, and then I've seen that cloud9 can handle the same thing, but with respecting the double click setting of the os, so I was wondering how they do it and found out that they were using evt.detail 😉

src/SelectionManager.ts
mofux: I think you are right, more than three clicks should yield the same behaviour as three clicks.

src/SelectionManager.ts
```diff
@@ -385,7 +385,7 @@ export class SelectionManager extends EventEmitter {
           this._onSingleClick(event);
       } else if (event.detail === 2) {
           this._onDoubleClick(event);
-      } else if (event.detail === 3) {
+      } else if (event.detail >= 3) {
```
Tyriar: Actually looking closer at the code it was correct before :sweat_smile:, we shouldn't need to re-perform triple click as we don't drop the selection between clicks 3 and 4.

Tyriar:
Just revert that change and ready to go, pending the tests. I love deleting code :fireworks:

mofux:
Ready for take-off 🚀

Tyriar:
@parisk is it possible to make the mac lint/test Travis jobs non-block to merge? There is quite a queue for Mac builds apparently.

parisk:
@Tyriar, yes of course.

PR #781, opened 2017-07-11
Retain the color of a character for underline and bar cursor styles

### LINKED ISSUE #774: Retain the color of a character with the cursor on it, for underline and bar cursor styles
It makes sense to override the color for a 'block' cursor, but it would look better if the color was kept for the other cursor styles.

jul-10-2017 17-50-54

### LINKED ISSUE #789: Terminal focus class is lost under certain conditions
The `.terminal.focus` class is unset if a program that supports mouse pointer integration is active, and the terminal receives a `mousedown` event. This causes the terminal cursor to render in an unfocused state, even if the terminal still has focus.

**How to reproduce**
Try to open `mc` or `nvim` - they support mouse pointer integration by default. Inspect the `.terminal` element classes. Once you click on the terminal (sometimes it needs multiple attempts), the `focus` class will be unset, although the terminal still has focus.

### LINKED ISSUE #790: Terminal is initialised with the wrong cursor style
When creating the terminal, and passing the `cursorStyle` through `options`, the cursor will still be `block`. One has to trigger `setOption` before the cursor style is picked up. This is because the class for the cursor style `xterm-cursor-style-...` is not set when creating the terminal.

### LINKED ISSUE #820: Underline and blink cannot be rendered in parallel
With the current css, if a character has the `underline` **and** `blink` flags, only the blink style will be rendered. This is because there are conflicting css rules that either set `text-decoration: blink` **OR** `text-decoration: underline`, but there is no rule for both conditions.

mofux:
Fixes #774
Fixes #789
Fixes #790
Fixes #820
Related #783

⚠️ **Warning**
I had to change the cursor related `css` to get this feature in, and to resolve styling related issues. Consumers that have overwritten the default styles may have to adopt their stylesheet.

🚀 **Improvements**
The`.terminal` element will always have a class with the cursor style, previously, only `bar` and `underline` cursors had a class.

🐞 **Fixes**
- #774, #783: `bar` and `underline` cursors now render correctly and do not alter the foreground color of the text, 
- #774: `blink` is now working correctly if the block cursor is over a background, 
- #789: `focus` class is kept when interacting with programs that support mouse pointer integration
- #790: The cursor style passed through options is now respected
- #820: Allow blink and underline flags to be rendered in parallel

🎮 **Testing**
I found it very useful to `nvim README.md` in the xterm.js demo because the Markdown syntax highlighter of `nvim` will create almost all visual test cases (background, foreground, underline, bold).

### DIFF
--- a/src/Renderer.ts
+++ b/src/Renderer.ts
@@ -170,16 +170,13 @@ export class Renderer {
         let data: any = line[i][0];
         const ch = line[i][1];
         const ch_width: any = line[i][2];
+        const isCursor: boolean = i === x;
         if (!ch_width) {
           continue;
         }
 
-        if (i === x) {
-          data = -1;
-        }
-
-        if (data !== attr) {
-          if (attr !== this._terminal.defAttr) {
+        if (data !== attr || isCursor) {
+          if (attr !== this._terminal.defAttr && !isCursor) {
             if (innerHTML) {
               currentElement.innerHTML = innerHTML;
               innerHTML = '';
@@ -187,7 +184,7 @@ export class Renderer {
             documentFragment.appendChild(currentElement);
             currentElement = null;
           }
-          if (data !== this._terminal.defAttr) {
+          if (data !== this._terminal.defAttr || isCursor) {
             if (innerHTML && !currentElement) {
               currentElement = this._spanElementObjectPool.acquire();
             }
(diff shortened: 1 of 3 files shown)


Tyriar:
@mofux we don't want to use the same color when it's a block cursor, instead we want to use the terminal's background color. Is this case covered by this?

mofux:
@Tyriar I think you mean we want the foreground color, not the background color? Yes, the css rule takes precedence.

mofux:
I've just stumbled over a couple of issues while testing:

1. In the unfocused state, the block outline is always drawn, no matter which cursor style is selected. What should be the preferred look of the those cursor styles in the unfocused state?

2. The line and underline cursors also need the background color, otherwise there will be a black square whenever the cursor is over a background (fix is on the way)

3. When I open midnight command `mc` or `nvim` and perform a click, the `focus` class on the terminal is removed for some reason. It only comes back after I close the app. I believe the always happens if the program supports mouse pointer integration. I'll try to investigate...

Tyriar:
@mofux 

1. Yeah I knew about this but I don't think there is an issue for it. gnome-terminal just shows the line/underline solid when unfocused, Terminal.app acts as it is now. I think aligning with gnome-terminal here is ideal. Created https://github.com/sourcelair/xterm.js/issues/783
2. :+1:
3. Does it affect the cursor though? Most styles are applied to `:focus` and `.focus`

mofux:
This PR is ready for review. Please see the PR description for updated details.

mofux:
@Tyriar I can see this running into conflicts with #756. I propose to hold this back until #756 is merged. Once it's in I'll update this PR.

If you want, you could already include the change made here (https://github.com/sourcelair/xterm.js/pull/781/commits/804051e789e3f0605b18a14a0015a89110b83f00) into #756, which should solve #774.

Tyriar:
@mofux fill hopefully get to this in the next couple of days, sorry about the delay

src/Renderer.ts
```diff
@@ -199,14 +201,22 @@ export class Renderer {
               documentFragment.appendChild(currentElement);
             }
             currentElement = this._spanElementObjectPool.acquire();
+
             if (data === -1) {
+              let cursorBg = cursorData[0];
+              let cursorFg = cursorData[1];
               currentElement.classList.add('reverse-video');
               currentElement.classList.add('terminal-cursor');
+              if (cursorFg >= 0 && cursorFg < 256) {
```
Tyriar: I think we actually want to do more here as well, bold, underline, invisible (maybe?). Perhaps the cursor part should try share more code with the regular part?

Some things don't make sense here: blink, inverse

And some things need to be done differently: bg/fg application

src/Renderer.ts
```diff
@@ -168,13 +168,15 @@ export class Renderer {
...
         let data: any = line[i][0];
+        let cursorData;
         const ch = line[i][1];
         const ch_width: any = line[i][2];
         if (!ch_width) {
           continue;
         }
 
         if (i === x) {
+          cursorData = [data & 0x1ff, (data >> 9) & 0x1ff];
```
Tyriar: On the topic of sharing code, it would be good to reuse this line:

```ts
let bg
let fg = (data >> 9) & 0x1ff;
```

src/Renderer.ts
```diff
@@ -168,13 +168,15 @@ export class Renderer {
       for (let i = 0; i < width; i++) {
         // TODO: Could data be a more specific type?
         let data: any = line[i][0];
+        let cursorData;
```
Tyriar: Perhaps this should go outside the look and instead be `let isCursor = false;`

src/xterm.js
```diff
@@ -1044,6 +1045,9 @@ Terminal.prototype.bindMouse = function() {
   }
 
   on(el, 'mousedown', function(ev) {
+    // prevent the focus on the textarea from getting lost
+    ev.preventDefault();
```
Tyriar: I think this might need to go after the `self.mouseEvents` check? Can you verify by running tmux, vim, etc. in mouse mode and see if they get the events?

mofux: It's good to have it at the start, because we will want to prevent the focus of the textarea from being lost in either case. I have tested it in `nvim` and `mc` and it works as expected. I've also made sure it doesn't cause a regression for the selection manager and linkifier.

mofux:
@Tyriar I have merged the rendering of the cursor with the existing logic, which avoids code duplication and allows for all styles to applied to the cursor too (with the execption of `xterm-hidden` which I believe is something we don't want for the cursor). 

I have also fixed #820 on the way.

Check the updated PR head notes for running some tests (using nvim).

mofux:
This is ready for re-review 😁. The diff for Renderer.ts is a little confusing because I have removed a nested if, so the whole inner block has shifted one indentation to the left now.

mofux:
@parisk  It respects the decisions made in #783 - which means showing an outlined block cursor in the unfocused state, no matter what cursor style is currently active (basically the same behaviour as we had before). When I first started this PR it seemed that #783 would go into the direction of showing the selected cursor in unfocused state.

See commit https://github.com/sourcelair/xterm.js/pull/781/commits/e5158adb611a5136e362066db66967b18ac97d71 for details.

Tyriar:
Sorry haven't had time to get to this yet. I've been swamped with other tasks plus I was pretty ill all last week 🙁

parisk:
@mofux can you please rebase with master? It would be great to include this in 2.9.0 😄.

mofux:
@parisk I have just rebased the branch 😅

src/xterm.js
mofux: I think we can and should be removed once #828 is merged, because #828 adds a focus() at the top of the handler.

mofux:
Sorry, I was wrong, just tested with your changes from #828 and `ev.preventDefault()` is still required to prevent the focus from getting lost when in mouse interactive mode. I have put it back now.

mofux:
I have updated this PR to resolve the merge conflict from #828. Ready for merge if @Tyriar has no objections 😊

PR #788, opened 2017-07-12
Fix regression. Existing tests work again.

AndrienkoAleksandr:
Target-branch: buffer-set. 
Related issue: https://github.com/sourcelair/xterm.js/pull/717

Signed-off-by: Aleksandr Andriienko <oandriie@redhat.com>


AndrienkoAleksandr:
@parisk, review please. I'm sorry for regression - after reset Terminal didn't clean up lines in the buffers, so test src/test/test.js fails on checking surrogate pairs, because correct "reset" was condition to pass this test.

PR #795, opened 2017-07-14
Remove unused constants from the SelectionManager

mofux:
Removes two, now unused `const` declarations that I've missed in PR #780 😓


mofux:
😤 I am only removing code in this PR, how can this decrease the code coverage? 🤔

Tyriar:
It goes down because those lines were "covered" before; they were run when the file was loaded. Simplified example: 3/4 lines = 75% coverage, 1/2 lines = 50% coverage 😉

PR #797, opened 2017-07-14
Guard public selection api

### LINKED ISSUE #796: Guard Terminal.prototype methods that reach out for this.selectionManager
The selection manager `term.selectionManager` is only available after `term.open()` was called, before it will be `null`. Unfortunately, there are methods on Terminal.prototype (`hasSelection`, `getSelection`, `clearSelection`, `selectAll`) that operate on `term.selectionManager` and assume that it is available. We should either guard these methods, or make `term.selectionManager` available before `open` is called.

mofux:
Fixes #796 by soft-failing if selection manager is not yet available (term.open not yet called). I believe this is better than throwing an error. This behaviour lines up with the behaviour of the linkifier, that only throws for sensitive api calls (like registering a link matcher for example), but otherwise soft-fails too.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -1437,7 +1437,7 @@ Terminal.prototype.deregisterLinkMatcher = function(matcherId) {
  * Gets whether the terminal has an active selection.
  */
 Terminal.prototype.hasSelection = function() {
-  return this.selectionManager.hasSelection;
+  return !!(this.selectionManager && this.selectionManager.hasSelection);
 };
 
 /**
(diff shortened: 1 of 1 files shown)


src/xterm.js
```diff
@@ -1437,29 +1437,33 @@ Terminal.prototype.deregisterLinkMatcher = function(matcherId) {
  * Gets whether the terminal has an active selection.
  */
 Terminal.prototype.hasSelection = function() {
-  return this.selectionManager.hasSelection;
+  return !!(this.selectionManager && this.selectionManager.hasSelection);
```
Tyriar: Let's simplify this so it's easier to read:

```js
if (!this.selectionManager) {
  return false;
}
return this.selectionManager.hasSelection;
```

PR #798, opened 2017-07-16
faster wcwidth with lookup table

jerch:
This PR optimizes `wcwidth` by using a lookup table for BMP characters. For a BMP char it boils down to a simple table lookup with some shifting magic to pack the data into 16k bytes. Speedup is at least 10 times. Equality tests included.


src/InputHandler.ts
```diff
@@ -1588,9 +1590,32 @@ const wcwidth = (function(opts) {
...
+        num |= (wcwidthBMP(j + 2));
+        num <<= 2;
+        num |= (wcwidthBMP(j + 1));
+        num <<= 2;
+        num |= (wcwidthBMP(j));
+        table[i] = num;
+    }
+    return function (num) {
+        num = num | 0;
+        return (num < 65536) ? (table[num >> 2] >> ((num & 3) * 2)) & 3 : wcwidthHigh(num);
```
Tyriar: Should this have an early return for ascii characters as they are the vast majority, or would that not boost perf much at this point?

src/InputHandler.ts
```diff
@@ -1588,9 +1590,32 @@ const wcwidth = (function(opts) {
...
+    function wcwidthHigh(ucs) {
+        if (bisearch(ucs, COMBINING_HIGH))
+            return 0;
+        if ((ucs >= 0x20000 && ucs <= 0x2fffd) || (ucs >= 0x30000 && ucs <= 0x3fffd)) {
+            return 2;
+        }
+        return 1;
+    }
+    let table = new Uint8Array(16384);
+    for (let i = 0; i < 16384; ++i) {
```
Tyriar: Does this increase startup in a noticeable way?

src/InputHandler.ts
Tyriar: Could do with a comment explaining the bit shifting part

src/InputHandler.ts
```diff
@@ -1588,9 +1590,32 @@ const wcwidth = (function(opts) {
...
+    }
+    function wcwidthHigh(ucs) {
+        if (bisearch(ucs, COMBINING_HIGH))
+            return 0;
+        if ((ucs >= 0x20000 && ucs <= 0x2fffd) || (ucs >= 0x30000 && ucs <= 0x3fffd)) {
+            return 2;
+        }
+        return 1;
+    }
+    let table = new Uint8Array(16384);
```
Tyriar: A named variable/comment explaining 16384 would be good ((65536-4)/4+1?).

Tyriar: Also this needs to be conditional and fallback to a regular array for browsers without support

src/InputHandler.ts
jerch: @Tyriar Well it gives 5% to exclude the ascii range from the shifting & lookup. So yeah gonna change it.

src/InputHandler.ts
jerch: It add 10 to 16 ms.

Tyriar: Is that 10-16ms when the first terminal is launched or when the script is loaded?

jerch: When the script loads, it is done during import.

Tyriar: Created a follow up :smiley: https://github.com/sourcelair/xterm.js/issues/799

jerch:
@Tyriar Changed the table to `Uint32Array`, which speeds up the creation time (down to 6ms), it slightly slows down the reading by 20% compared to `Uint8Array`, dunno why. About the lazy init - I dont know where to hook it into. Is there a sumthing like a terminal-init event?

Tyriar:
@jerch this would mean doing it on either terminal creation (`new Terminal()`) or on first request of character width (might cause async issues with the way writing works?).

jerch:
Addressed the lazy loading with the last commit, the lookup table is created at the first call with a char > 127. Benchmarks with node V8 on my machine are now:
- old wcwidth ASCII range (51200 times): 6.5MB in 120ms
- new wcwidth ASCII range (512000 times): 6.5MB in 60ms
- old wcwidth 0-65536 range (100 times): 6.5MB in 850ms
- new wcwidth 0-65536 range (100 times): 6.5MB in 100ms

src/InputHandler.test.ts
```diff
@@ -47,3 +48,138 @@ describe('InputHandler', () => {
...
+        (ucs >= 0xff00 && ucs <= 0xff60) ||    // Fullwidth Forms
+        (ucs >= 0xffe0 && ucs <= 0xffe6) ||
+        (ucs >= 0x20000 && ucs <= 0x2fffd) ||
+        (ucs >= 0x30000 && ucs <= 0x3fffd)));
+    }
+    return wcwidth;
+})({nul: 0, control: 0});  // configurable options
+
+describe('wcwidth', () => {
+  it('same as old implementation for BMP and individual higher', () => {
```
Tyriar: Can't get more confident about no functional changes than this 😉

Tyriar:
@jerch CI failed on macOS:

```
  181 passing (5s)
  1 failing
  1) wcwidth same as old implementation for BMP and individual higher:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

Might need to increase the timeout of that test? https://travis-ci.org/sourcelair/xterm.js/jobs/254546569

src/InputHandler.test.ts
jerch: haha :D

jerch:
Hmm yeah seems the travis OSX boxes are pretty busy from time to time and tend timeout longer running tests. The test runs in under a second here...

PR #810, opened 2017-07-21
Make SelectionManager more resilient

### LINKED ISSUE #809: Add null check to SelectionManager._onSingleClick
Exception report: https://github.com/Microsoft/vscode/issues/31184

Tyriar:
Fixes #809


blink1073:
Something we have done in [JupyterLab](https://github.com/jupyterlab/jupyterlab/pull/2670) is temporarily turn on `strictNullChecks` and fix as many local errors as we can, and then turn the flag off.  We have a couple of type system restrictions that prevent us from turning it on permanently.  One annoying thing is that `getComputedStyle` attributes are all possibly null, leading to `style.width!` (notice the null assertion operator).

Tyriar:
It would be good to be as strict as possible eventually, can't wait until https://github.com/sourcelair/xterm.js/issues/335 is done

PR #813, opened 2017-07-21
[WIP] Fix reset terminal to initial state: http://vt100.net/docs/vt510-rm/RIS.html

AndrienkoAleksandr:
Fix reset terminal to initial state: http://vt100.net/docs/vt510-rm/RIS.html . Clear all buffers: alt and normal to inital state.


AndrienkoAleksandr:
WIP. Don't merge.

PR #816, opened 2017-07-24
Alt key should force selection

joaomoreno:
From Microsoft/vscode#28863

### DIFF
--- a/src/SelectionManager.ts
+++ b/src/SelectionManager.ts
@@ -317,7 +317,8 @@ export class SelectionManager extends EventEmitter {
    * @param event The mousedown event.
    */
   private _onMouseDown(event: MouseEvent) {
-    // If we have selection, we want the context menu on right click
+    // If we have selection, we want the context menu on right click even if the
+    // terminal is in mouse mode.
     if (event.button === 2 && this.hasSelection) {
       event.stopPropagation();
       return;
@@ -330,7 +331,7 @@ export class SelectionManager extends EventEmitter {
 
     // Allow selection when using a specific modifier key, even when disabled
     if (!this._enabled) {
-      const shouldForceSelection = Browser.isMac ? event.altKey : event.shiftKey;
+      const shouldForceSelection = Browser.isMac && event.altKey;
 
       if (!shouldForceSelection) {
         return;
--- a/src/SelectionManager.ts
+++ b/src/SelectionManager.ts
@@ -317,20 +317,27 @@ export class SelectionManager extends EventEmitter {
    * @param event The mousedown event.
    */
   private _onMouseDown(event: MouseEvent) {
-    // Ignore this event when selection is disabled
-    if (!this._enabled) {
-      if (!(Browser.isLinux ? event.shiftKey : event.altKey)) {
-        return;
-      } else {
-
-        // Don't send the mouse down event to the current process
-        event.stopPropagation();
-      }
+    // If we have selection, we want the context menu on right click
+    if (event.button === 2 && this.hasSelection) {
+      event.stopPropagation();
+      return;
     }
 
     // Only action the primary button
     if (event.button !== 0) {
       return;
+    }
+
+    // Allow selection when using a specific modifier key, even when disabled
+    if (!this._enabled) {
+      const shouldForceSelection = Browser.isMac ? event.altKey : event.shiftKey;
+
+      if (!shouldForceSelection) {
+        return;
+      }
+
+      // Don't send the mouse down event to the current process, we want to select
+      event.stopPropagation();
     }
 
     // Tell the browser not to start a regular selection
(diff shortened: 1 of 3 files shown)


joaomoreno:
@Tyriar This is not yet done, still a WIP.

I've changed what `enable()` and `disable()` do in the SelectionManager and made them more lightweight. That way, the <kbd>Alt</kbd> can still force the SelectionManager to do his business, even though it is disabled. We probably need different analogy other than enabled/disabled...

I'll still need to address these questions:

> - Does this work the same on Linux?
> - What happens if alt is dropped half way through the selection?
> - Does shift+alt click expand the selection?
> - Should it be an [option](https://github.com/sourcelair/xterm.js/blob/48dab494db8b4e1eabcbc656501f8f538cd104cc/src/xterm.js#L338)?

As for an option, I'd say no. I couldn't find one for iTerm.

src/xterm.js
```diff
@@ -517,7 +517,7 @@ Terminal.prototype.initGlobal = function() {
   on(this.element, 'copy', event => {
     // If mouse events are active it means the selection manager is disabled and
     // copy should be handled by the host program.
-    if (this.mouseEvents) {
+    if (!this.selectionManager.hasSelection) {
```
mofux: I think you should either guard this call or use `this.hasSelection()` because `this.selectionManager` might not yet be available at this stage (it is only available after `term.open` was called).

Tyriar: Now that you say this, `Terminal.hasSelection` should probably have a null check in it as consumers could call the API before `Terminal.open`.

mofux: It has since #797 😉

Tyriar: Opps, was looking at the code from GitHub and I guess it wasn't master 😕

joaomoreno:
Fixed the NPE.

Answering the questions:

- Does this work the same on Linux?

In the Gnome terminal, the <kbd>Shift</kbd> key is used to escape this Mouse mode. We should make this platform dependent since in many Linux systems it seems <kbd>Alt</kbd> moves the window. 

- What happens if alt is dropped half way through the selection?

In the Gnome terminal, nothing really happens, the selection continues to be modified as the mouse moves. Seems like we are OK there.

- Does shift+alt click expand the selection?

In the Gnome terminal, in such a mode, <kbd>Shift</kbd> no longer allows to extend a selection; every selection is a new one. This is fine by me and I propose to implement this.

- Should it be an option?

I wouldn't make it an option.

@Tyriar @mofux OK to go ahead with these assumptions?

joaomoreno:
Also, sorry about the auto formatting. Let me know if you prefer a clean commit which doesn't format untouched code.

mofux:
Your assumptions sound reasonable. My only slight concern is if we implement an 'altIsMeta' option in the future, this might conflict with the alt key being used for selection... but since it's only mouse interaction that shouldn't really be a problem...

joaomoreno:
Alright guys. I've pushed more changes. Tests should go green soon. Feel free to properly review and/or merge. :+1:

src/SelectionManager.ts
```diff
@@ -314,6 +317,17 @@ export class SelectionManager extends EventEmitter {
...
    */
   private _onMouseDown(event: MouseEvent) {
+    // Ignore this event when selection is disabled
+    if (!this._enabled) {
+      if (!(Browser.isLinux ? event.shiftKey : event.altKey)) {
+        return;
+      } else {
+
+        // Don't send the mouse down event to the current process
+        event.stopPropagation();
```
Tyriar: Does this break the context menu on right click? Maybe the enabled check should be below the primary button check?

src/SelectionManager.ts
```diff
@@ -314,6 +317,17 @@ export class SelectionManager extends EventEmitter {
    * @param event The mousedown event.
    */
   private _onMouseDown(event: MouseEvent) {
+    // Ignore this event when selection is disabled
+    if (!this._enabled) {
+      if (!(Browser.isLinux ? event.shiftKey : event.altKey)) {
```
Tyriar: Shouldn't this be isMac? Also the inverted ternary is pretty hard to read

src/SelectionManager.ts
joaomoreno: Great catch!

joaomoreno:
Refactored a bit to accommodate for the fact that we always want the context menu to show up when there is a selection in the editor.

joaomoreno:
Pushed!

Tyriar:
Looks good, I made a minor change to disable it on Linux/Windows as it will probably lead to weird behavior being on shift a that performs an incremental not a single click.

joaomoreno:
@Tyriar I started selfhosting on Linux last month. Without that change, I still won't be able to select text in the terminal. :cry: 

Note that the shift key hijacking would only happen in a program which enabled mouse mode, thus disabling selection.

Tyriar:
After checking now this is how it works on gnome-terminal, we should bring it back with shift for the next version. I don't think I had a Linux box with me when I was merging, sorry :slightly_frowning_face:. Tracking in #869

PR #817, opened 2017-07-24
Add lineFeed event

### LINKED ISSUE #819: Expose lineFeed event
Currently the only way to listen to things happening inside the terminal is either on refresh (fired even on prompt input) or by listening to the data stream (a huge amount of data, inefficient to parse to check for \n) being fed into xterm.js. It would be useful to have some more granularity here for example if someone wants to know when there is a new line feed added. This would allow programs that integrate with the terminal a greater understanding of what's happening.

Lixire:
- this is useful for checking when a cmd is entered or when a cmd finishes
- also handy to have for accessibility for output

Fixes #819

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -132,6 +132,12 @@ export class InputHandler implements IInputHandler {
     if (this._terminal.buffer.x >= this._terminal.cols) {
       this._terminal.buffer.x--;
     }
+  /**
+   * This event is emitted whenever the terminal outputs a LF or NL.
+   *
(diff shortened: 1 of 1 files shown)


src/InputHandler.ts
```diff
@@ -132,6 +132,12 @@ export class InputHandler implements IInputHandler {
     if (this._terminal.buffer.x >= this._terminal.cols) {
       this._terminal.buffer.x--;
     }
+  /**
```
Tyriar: Indentation is a bit off

PR #822, opened 2017-07-27
Fix save and restore cursor behavior

### LINKED ISSUE #818: Output is all messed up when running apt install/remove
Seems related to the progress bar is uses.

https://github.com/Microsoft/vscode/issues/31339

Tyriar:
This broke in #717.

Fixes #818

### DIFF
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -15,6 +15,9 @@ import { CircularList } from './utils/CircularList';
 export class Buffer {
   public lines: CircularList<[number, string, number][]>;
 
+  public savedY: number;
+  public savedX: number;
+
   /**
    * Create a new Buffer.
    * @param {Terminal} terminal - The terminal the Buffer will belong to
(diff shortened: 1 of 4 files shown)


AndrienkoAleksandr:
Ok, merge it, but seems that's not a fix, this is alternative of real solution...

AndrienkoAleksandr:
I am working on some tests on this topic, https://github.com/AndrienkoAleksandr/xterm.js/blob/fixFullResetTerminal/src/test/test.js#L1025
So, @Tyriar thank you for a "new testcase", I will think about how to implement some solution to cover all known testcases...

Tyriar:
@AndrienkoAleksandr are you sure we want to save/restore the cursor when switching to/from the alt buffer? I assume it was originally commented out for a reason and I'm not sure xterm does this https://github.com/joejulian/xterm/blob/defc6dd5684a12dc8e56cb6973ef973e7a32caa3/misc.c#L4172

Tyriar:
@AndrienkoAleksandr clarification on what's wrong with this would be good too, is it just the save/restore on buffer change?

Issue #406, opened 2016-12-13
Deprecate and finally drop Bower support


parisk:
I think it's time to deprecate support for Bower gracefully and eventually drop it completely in the next major release (3.0.0).

This is a topic that has been discussed a little bit in the past, but I always argued against it since we had users that preferred to use xterm.js via Bower.

Eventually I changed my mind after seeing the weird things that can happen with the built files checked out in the repo and reading https://github.com/codemirror/CodeMirror/issues/4328#issuecomment-255110880, which actually states the fundamental design issue of Bower: **it confuses source repositories (like this) with distribution mechanisms**.

To the point, I propose taking the following steps to make sure that this goes as gracefully as possible for our users:

1. Create a new **xtermjs/bower** repository, where we will point the `xterm.js` Bower package and push the built files of each release there (starting with the current one).
2. Remove `dist` from this repository completely, ignore it from Git and include it in NPM distributions
3. Stop updating the **xtermjs/bower** repository starting from version 3.0.0

/cc @Tyriar

Tyriar:
:fireworks: :tada: :sparkler:

kvanberendonck:
Just pulled xterm using bower, is this deprecation necessary? Probably wouldn't use it if it isn't available on bower.

Tyriar:
@kvanberendonck bower puts additional burden on package maintainers (namely keeping compiled files in the tree and additional release steps) is the main reason for dropping. Any reason you don't use the npm package and bundle it with your bundler of choice?

kvanberendonck:
We generally don't use npm except for build system dependencies. The build system spans across multiple related parts of the project, and each has its own set of bower dependencies and versions

parisk:
@kvanberendonck the main issue with Bower it that it has adopted a practice which is problematic for projects like ours; it has equated the repository with the distribution of the package.

This means what @Tyriar said, maintaining compiled files in tree (which is quite confusing for new contributors), adding additional release steps (extra commit) and be always careful not to commit a new file in `dist` (as it cannot be `.gitignore`d).

What we will do at first is create a new repository, which will replace the old one in bower (this requires no change by you).

Next from version **3.0.0**, we will stop maintaining this repository completely, so you will have to change the way you download xterm.js.

What we could do though, is publish the build artifact as to our GitHub Releases, so you can use the [tar/zip URL format in your dependencies](https://github.com/bower/spec/blob/master/json.md#dependencies).

Tyriar:
[...]

PR #823, opened 2017-07-28
Issue #821 - Prevent TypeError in resize()

### LINKED ISSUE #821: TypeError calling resize() via fit add-on
I'm not sure how I'm getting into this state as the application is pretty complex but I have a situation where, on resizing the terminal element, in the resize() code that pads out each line in the terminal I'm ending up with lines that are undefined. This causes the code to fail when it queries for the line length:

```
TypeError: Cannot read property 'length' of undefined 
message:"Cannot read property 'length' of undefined"
stack:"TypeError: Cannot read property 'length' of undefined\n    at Terminal.19.Terminal.resize (http://localhost:3000/utc/bower_components/xterm.js/dist/xterm.js:4637:37)\n    at Object.exports.fit (http://localhost:3000/utc/bower_components/xterm.js/dist/addons/fit/fit.js:73:12)\n    at Terminal.Xterm.fit (http://localhost:3000/utc/bower_components/xterm.js/dist/addons/fit/fit.js:82:20)\n    at http://localhost:3000/utc/scripts/utccomponents/utcterminal.js:31:19\n    at m.$emit (http://localhost:3000/utc/bower_components/angular/angular.min.js:151:429)\n    at broadcastLeftRightResize (http://localhost:3000/utc/scripts/framework/stratosresizer/resizer.js:96:28)\n    at HTMLDocument.handleMouseUp (http://localhost:3000/utc/scripts/framework/stratosresizer/resizer.js:133:25)"
```

I'm using an event from the Angular resizer to call fit() but this is not called continuously while the dragging is happening, only once when the resizing is completed.

I can 'fix' it by updating the resize method in xterm.js like this:

```
        ch = [this.defAttr, ' ', 1];
        i = this.lines.length;
        while (i--) {
            ////////////////////////////////////////
            if (this.lines.get(i) === undefined) {
                this.lines.set(i, this.blankLine());
            }
            ////////////////////////////////////////
            while (this.lines.get(i).length < x) {
                this.lines.get(i).push(ch);
            }
        }

```
But my concern is that, if no-one else is having this problem, it may be an indication of a bigger issue somewhere else. 

- Have you seen anything like this before and if so do you have any idea of what might cause it to happen?
- Is this a reasonable way to fix it without causing more problems downstream?

## Details
- Browser and browser version:  Chrome 59.0.3071.115 (Official Build) (64-bit)
- OS version:  Windows 7
- xterm.js version:  2.8.1

jpmasters:
Updated Terminal resize() to fill undefined rows before resizing them.

Closes #821.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -1941,6 +1941,9 @@ Terminal.prototype.resize = function(x, y) {
     ch = [this.defAttr, ' ', 1]; // does xterm use the default attr?
     i = this.buffer.lines.length;
     while (i--) {
+      if (this.buffer.lines.get(i) === undefined) {
+        this.buffer.lines.set(i, this.blankLine());
+      }
       while (this.buffer.lines.get(i).length < x) {
         this.buffer.lines.get(i).push(ch);
       }
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ npm-debug.log
 .env
 build/
 .vscode/
+.DS_Store
 
 # Directories needed for code coverage
 /coverage/
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,6 +55,12 @@
       "integrity": "sha1-uB3l9ydOxOdW15fNg08wNkJyTl0=",
       "dev": true
     },
+    "acorn": {
+      "version": "4.0.13",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-4.0.13.tgz",
+      "integrity": "sha1-EFSVrlNh1pe9GVyCUZLhrX8lN4c=",
+      "dev": true
+    },
     "acorn-globals": {
       "version": "https://registry.npmjs.org/acorn-globals/-/acorn-globals-3.1.0.tgz",
       "integrity": "sha1-/YJw9x+7SZawBPqIDuXUZXOnMb8=",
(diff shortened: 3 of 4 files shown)


parisk:
@jpmasters can you please revert the changes in all files in `dist/` and squash the commits please?

parisk:
Again, thanks for this PR @jpmasters! I was able to remove the checked out `dist/*` files myself, without bothering you 😄.

jpmasters:
Awesome, thanks @parisk, I wasn't sure whether the /dist files needed to be updated or not. Does this mean that gulp needs to be run after xterm.js is added to a project via a bower install or is this done automatically?

parisk:
@jpmasters the `dist/*` files are updated only on new releases and they are there only for Bower.

This will go away thankfully though on 3.0: #406.

src/xterm.js
```diff
@@ -1941,6 +1941,9 @@ Terminal.prototype.resize = function(x, y) {
     ch = [this.defAttr, ' ', 1]; // does xterm use the default attr?
     i = this.buffer.lines.length;
     while (i--) {
+      if (this.buffer.lines.get(i) === undefined) {
```
Tyriar: The resize rows section below is meant to handle this row resizing, if this is actually happening we should figure out the root cause. `this.buffer.lines.get(i)` should always return an array when `i < this.buffer.lines.length` which is guaranteed by line 1942.

jpmasters: Underneath the CircularList is just a straight Javascript array so I'm assuming that somehow 'undefined' got pushed into the array. Could that not give the same symptoms? I'm wondering if perhaps there was some kind of problem parsing the data from the server that resulted in an undefined object getting into the circular list.

jpmasters: Looking at what is happening in the debugger, we're making a lot of use of the RI escape sequence to implement a 'scrolling' area of the terminal screen. This seems to result in calls to shiftElements that in turn calls set() in the circular buffer code which is bringing undefined data from the unused 1000 circular buffer array entries into the part being used by xterm. 

I think the root cause might be line 2282 in xterm.js (Terminal.prototype.reverseIndex). 
```
    this.buffer.lines.shiftElements(this.buffer.y + this.buffer.ybase, this.rows - 1, 1);
```
When I look at the prototype for shiftElements the parameters are start, **count** and offset so instead of setting count to this.rows - 1 should it not also subtract (this.buffer.y + this.buffer.ybase)? Otherwise when this.set is called in shiftElements (line 174 in CirlculstList.ts), it could be bringing in undefined data from the unused array area in the circular list buffer.

Tyriar: Nice investigation 😄 could you come up with a script or something that can reproduce the problem and put into https://github.com/sourcelair/xterm.js/issues/824? The best way I've found to do these is to open devtools console in the demo and type `term.write('\x1b[blah...')` containing the escape sequences that cause the issue.

PR #828, opened 2017-08-02
Always focus terminal on click

### LINKED ISSUE #827: Cannot focus terminal when text elsewhere on the page is selected
If I select text somewhere else on the page, I cannot focus the terminal by clicking inside it. It works fine after I clear the selection.

screencast 2017-08-01 13-27-32

## Details
- Browser and browser version: Chrome Version 59.0.3071.115 (Official Build) (64-bit)
- OS version: macOS Sierra 10.12.6 
- xterm.js version: 2.8.0 (and current master)

### Steps to reproduce

1. npm install
2. npm start
3. Open http://0.0.0.0:3000 in a browser
4. Select some text on the page outside of the terminal
5. Try to click inside the terminal

Problem: The terminal cannot be focused.

### LINKED ISSUE #830: Focus terminal on mousedown
At the moment, we focus the terminal on `click`, which means the terminal will only be in focused state after the mouse button comes back up (`mousedown -> mouseup -> click`):
[xterm.js#L750](https://github.com/sourcelair/xterm.js/blob/49c1ca3f57699a824ac79cf5cbd3d4a338b2b8aa/src/xterm.js#L750)

The natural behaviour of inputs is to focus on `mousedown`, which would also allow us to draw the focused state while performing a selection when the terminal was previously unfocused.

Maybe we should remove the current `click` handler completely, and then move the focus() call in the mousedown handler to the very top of the handler here:
[xterm.js#L1023](https://github.com/sourcelair/xterm.js/blob/49c1ca3f57699a824ac79cf5cbd3d4a338b2b8aa/src/xterm.js#L1023)

Any thoughts?

parisk:
This PR closes #827 and also closes #830.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -746,11 +746,6 @@ Terminal.prototype.open = function(parent, focus) {
     this.focus();
   }
 
-  // Focus terminal when it gets clicked
-  on(this.element, 'click', function() {
-    self.focus();
-  });
-
   // Listen for mouse events and translate
   // them into terminal mouse protocols.
   this.bindMouse();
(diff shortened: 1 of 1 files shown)


mofux:
@parisk Looking into the change, I am not sure if we need the click handler at all because imo we should already focus on `mousedown`. I have created #830 to catch up on that.

parisk:
That's right 😄.

PR #834, opened 2017-08-03
Prevent Webpack warning in 2.9.0

blink1073:
This warning came along with the upgrade to `2.9.0`:

```
WARNING in ../~/xterm/lib/addons/search/search.js
10:31-45 Critical dependency: the request of a dependency is an expression
```

We have a `require('xterm')` in our code, using WebPack 2.7.0.   This change prevents the warning.

### DIFF
--- a/src/addons/search/search.ts
+++ b/src/addons/search/search.ts
@@ -20,8 +20,7 @@ declare var window: any;
     /**
      * CommonJS environment
      */
-    const xterm = '../../xterm';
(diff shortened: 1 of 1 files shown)


Tyriar:
I think this was breaking either TypeScript or browserify in the build by merging all of the xterm lib into the search addon which isn't what we want.

parisk:
This looks OK for me. Gave it a try. Merging since it's fixing an error.

PR #855, opened 2017-08-07
Make buffer length equal rows + scrollback, make alt buffer always 0 scrollback

### LINKED ISSUE #802: alternate screen buffer has a bad scrollback experience
1. Use an application like vim that uses the alternate screen buffer
2. Use the arrow keys or page up / down so that content is added to the alt screen's scrollback buffer
3. Use the scrollbar to view the scrollback content

You will end up seeing duplicated content from the file you were using in vim.  This experience is confusing and not consistent with the two typical alt screen scrollback behaviors I have seen in other terminals.

Screenshot from vscode, highlighted sections are the duplicated content caused by the scrollback:

screen shot 2017-07-18 at 9 08 01 am

First noticed on 2.8.0 but the issue also exists on master.

Two typical scrollback experiences I have seen for alt screen:

1. The main window contains the alt screen content.  The scrollback buffer still contains the scrollback content from the main screen. Example of this is how mac's Terminal behaves.
2. When in the alt screen the scrollback buffer is disabled / set to 0

I think either of these experiences would be less confusing than the current behavior.

### LINKED ISSUE #847: The scrollback option should define the number of rows in the buffer above the viewport
Right now `buffersize = scrollback` when really it should be `buffersize = rows + scrollback`. This is how other terminals seems to act, it will also make supporting https://github.com/sourcelair/xterm.js/issues/802 easier and allow scrollback to equal 0 without needing to [warn like we currently do](https://github.com/sourcelair/xterm.js/blob/687a5e2efb808977ecebb78eb526cb01838c5ac4/src/Terminal.ts#L443).

Because of the way that the circular list is structured, resizing rows should resize the buffer to `rows + scrollback`.

Tyriar:
Fixes #847
Fixes #802


Tyriar:
@parisk I'm not sure that's better though, right now the "alt buffer scrollback is 0" logic is contained in a single place (as well as the comment explaining it). If we had `BufferSet` manage it, then we would need a `scrollback` parameter on the `Buffer.constructor`, `Buffer.clear` and `Buffer.resize`.

If preferable we could change the `_isAltBuffer` property to something like `_hasScrollback` to be more generic?

Tyriar:
Ping @parisk

parisk:
> we would need a scrollback parameter on the `Buffer.constructor`, `Buffer.clear` and `Buffer.resize`

Couldn't we just pass it only in the constructor and have `clear` and `resize` sourcing it from the buffer's attributes?

Renaming `_isAltBuffer` to `_hasScrollback` is 👌 too.

Tyriar:
@parisk but they change and the values are needed every time these functions are called. We would either have to pipe through the scrollback (or total height) every time from BufferSet or just have Buffer manage it itself.

PR #858, opened 2017-08-07
No longer focus the terminal by default

### LINKED ISSUE #646: Stop focusing automatically when opening the terminal in a DOM node
When using `open` to expose an xterm.js terminal in the DOM, the terminal gets focus automatically.

This may be useful for very simple use cases (like the demo of xterm.js), but it can make it difficult for more complex applications to control where the focus is.

The best thing to do here is stop focusing on the terminal inside `open` and let the user decide when to focus.

Initiated from #640.

Tyriar:
Fixes #646 

---

@parisk since we're breaking things, what are your thoughts on just removing focus all together as open can be followed by a `term.focus()` call?

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -759,20 +759,6 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     // Initialize global actions that need to be taken on the document.
     this.initGlobal();
 
-    /**
-     * Automatic focus functionality.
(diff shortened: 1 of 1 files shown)


parisk:
Yes, I am OK with removing it, but I would like xterm.js to offer some auto-focus functionality.

Not sure how this should be implemented though, so let's just remove the `focus` option for now.

PR #870, opened 2017-08-08
Revert "Disable selection override on Windows/Linux"

### LINKED ISSUE #869: Force selection manager using shift key on Linux/Windows
This was pulled last minute from #825/#816 but I think that was a mistake. Shift forces regular selection on gnome-terminal as well so I think it's a good idea.

Tyriar:
This reverts commit 3479614c732305b4174bf438e1ac3022aa3b3aa4.

Fixes #869

---

Tested on Linux and works nicely.


Tyriar:
FYI @joaomoreno, will pull this into VS Code probably early next week.

PR #874, opened 2017-08-09
Fix exception when resizing both dimensions

### LINKED ISSUE #860: Error after resizing both row and column at the same time.
In the case the terminal row and column are increased at the same time (calling term.resize() once), only the existing rows have their number of characters increased. The new rows are created with the previous number of columns. It seems to be a regression compared to xterm.js 2.9.0.

At the first refresh of the terminal, the error reported in the browser console is:
Uncaught TypeError: Cannot read property '0' of undefined
    at Renderer.10.Renderer._refresh (Renderer.ts:170)
    at Renderer.10.Renderer._refreshLoop (Renderer.ts:99)

## Details
- Browser and browser version: Chrome 59.0.3071.115
- OS version: Windows
- xterm.js version: 2.9.1

### Steps to reproduce

1.  Create a term object with 1 row and 1 column (for example)
2.  Resize the terminal to 24 rows and 80 columns. 
3.  Check the first row is resized to 80 columns (80 characters), but the second row is defined as a line with only one character (array of one element).

Tyriar:
Pulling https://github.com/sourcelair/xterm.js/pull/873 into v3, fixing conflicts. I didn't see much benefit moving blankLine into Buffer as suggested in #873.

Fixes #860

### DIFF
--- a/src/Buffer.test.ts
+++ b/src/Buffer.test.ts
@@ -144,5 +144,16 @@ describe('Buffer', () => {
         });
       });
     });
+
+    describe('row and column increased', () => {
+      it('should resize properly', () => {
+        buffer.fillViewportRows();
+        buffer.resize(INIT_COLS + 5, INIT_ROWS + 5);
+        assert.equal(buffer.lines.length, INIT_ROWS + 5);
+        for (let i = 0; i < INIT_ROWS + 5; i++) {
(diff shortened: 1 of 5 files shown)


Tyriar:
Ping @parisk

PR #884, opened 2017-08-14
Change demo to improve testing of the terminal resize mechanizm.

AndrienkoAleksandr:
changeddemo

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -12,10 +12,24 @@
         <script src="/build/addons/fit/fit.js" ></script>
         <script src="/build/addons/fullscreen/fullscreen.js" ></script>
         <script src="/build/addons/search/search.js" ></script>
+        <!-- for split panel https://github.com/nathancahill/Split.js -->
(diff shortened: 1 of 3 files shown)


AndrienkoAleksandr:
@Tyriar, @parisk if you don't like it you can simply close this pull request.

Tyriar:
While the info on the right is useful, using `setInterval` will make it more difficult to do performance work because the timeline will be muddied by the new interval. Also that's a lot of new dependencies, increasing the complexity and weight of the page.

Would making width and height a slider instead of a numeric input solve the same problem as the splits?

AndrienkoAleksandr:
@Tyriar about setInterval, you are right, but we can use some checkbox and do this information optional. About slider, I agree with you, it's will be lighter solution, I will think about it. 
What do you think, maybe it would be nice to create some addon with embedded auto-resize mechanism? It can be combo feature :)

Tyriar:
I think adding splits to the demo is more effort than it's worth, also if devs want a split in their own app they would probably write/integrate their own. It would be tough to get it right, also it would probably be the only addon to date that wraps `Terminal.element`.

PR #896, opened 2017-08-18
Alt+click will move the prompt cursor to that position

### LINKED ISSUE #890: Alt+click should move the prompt cursor to that position
Related discussion: https://github.com/zeit/hyper/issues/316

Note that this should only happen when mouse events are off. https://github.com/sourcelair/xterm.js/blob/f3d375a7a8c4a078f9892b5983d700896eb13174/src/Terminal.ts#L219

npezza93:
Fixes #890 

So this is still kind of a work in progress...Issues I am having:

1. If you move the cursor backward, the buffer backs up to that spot as well. So clicking the right arrow doesn't move the cursor anywhere but it moves to the left the extra spaces(See attached gif). I tried sending the escape sequence to the handler but it just outputted to D or C to the terminal. Along with 
`this.emit('data', C0.ESC + '[' + (this.buffer.x - mouseCol) + 'D');`. Any ideas what might be going on here?
2. You can move the cursor outside the typed buffer on the right and the left. How do I find the first and last col of input buffer?

alt-click


mofux:
I think you have to take a different approach to make this work, because you can only guess which parts of the cols / rows are actually available for a cursor position. In the end the tty will decide. Maybe you can try to count the amount of "move cursor left" or "move cursor right" it would need to get to the desired position - relative from the current cursor position, and then send the key sequences to the tty. The tty will then process that input and either really go to that position, or stop where it thinks is a physical barrier.

npezza93:
Thanks @mofux! I got it working now. If you click outside the available positions in the row it will send those to the tty but it just won't do anything (end up at either the beginning or the end of the line). Let me know what you guys think!

mofux:
Looks good! Can you also make it work for multiple rows? For example, if you generate a lot of input on the input line, so that it starts to break to the next line, you can observe that you can only move in the current line. We also have to make sure this works for double-width characters like `工` or `你`.

Tyriar:
I think jumping over double width characters will just work (I might be wrong), the main problem you will probably have is alt+clicking on the right cell of a double width char.

Some tests would be good here, we can defer this though as there really any integration tests like this yet. If you want to do some tests they should probably live in `Terminal.integration.ts` and do the following:

1. Create a Terminal
2. Hook it up to an instance of node-pty
3. Wait until prompt is stable
4. Fabricate and send alt+click event
5. Check cursor x,y
6. Repeat 4-5 with edge cases (double width, multi-line, left/right edge of prompt, etc.)

Note that Terminal.open doesn't get called as we would need a DOM for that.

Tyriar:
Multi-line still doesn't seem to work on my for macOS/Chrome.

Also wide characters aren't working either. To test paste "汉语/漢語" into the terminal and alt click the characters. Clicking the same character multiple times can change the selection as well.

src/Terminal.ts
```diff
@@ -1057,6 +1057,18 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
...
+        ev, this.rowContainer, this.charMeasure, this.cols, this.rows
+      ).map((coordinate: number) => {
+        return coordinate - 1;
+      });
+
+      if (!this.mouseEvents && ev.altKey && this.buffer.writableRows.indexOf(mouseRow) > -1) {
+        this.handleAltClick(ev, mouseCol, mouseRow);
+      }
+
       if (!this.mouseEvents) return;
```
Tyriar: The alt click handling section should go below this line, that way we don't need to check !this.mouseEvents and it's also very clear what's happening since the `this.mouseEvents` check exits early at the top.

src/Terminal.ts
```diff
@@ -1057,6 +1057,18 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
       ev.preventDefault();
       this.focus();
 
+      let mouseRow;
+      let mouseCol;
+      [mouseCol, mouseRow] = Mouse.getCoords(
```
Tyriar: Maybe move this into handleAltClick since that's the only place it's used?

src/Terminal.ts
```diff
@@ -2306,6 +2320,30 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
       this.helperContainer.removeChild(this.bellAudioElement);
     }
   }
+
+  private handleAltClick(ev: MouseEvent, mouseCol: number, mouseRow: number): void {
+    let movementString = '';
+    let positionsToMove = Math.abs(this.buffer.x - mouseCol);
```
Tyriar: I think the code could be simplified here by not doing Math.abs and handling negative values later on.

src/Terminal.ts
```diff
@@ -2306,6 +2320,30 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
...
+      (this.buffer.x > mouseCol && this.buffer.y < mouseRow);     // down/left
+
+    if (forwardCursorDirection) cursorDirection = 'C';
+    if (negativePositionMovment) positionsToMove *= -1;
+
+    for (let i = 0; i < (positionsToMove + rowPositions); i++) {
+      movementString += (C0.ESC + '[' + cursorDirection);
+    }
+
+    this.emit('data', movementString);
```
Tyriar: This should go through `Terminal.send`

npezza93:
@parisk:
> Can you please implement some tests (e.g. that Alt + Click sends the right movement string and that clicking without Alt does not send any movement string)
Yep! Still working on this.
> Please replace the usage of ev.code, as it is supported at all by neither Internet Explorer, nor Edge: KeyboardEvent.code Browser Compatibility
Yep, I can do that. 

@Tyriar 
> Multi-line still doesn't seem to work on my for macOS/Chrome.
Hmmmm interesting. I was testing on macOS/Chrome and it works how it should.

> Also wide characters aren't working either. To test paste "汉语/漢語" into the terminal and alt click the characters. Clicking the same character multiple times can change the selection as well.
Yep! Still working this.

src/InputHandler.ts
```diff
@@ -56,6 +56,9 @@ export class InputHandler implements IInputHandler {
         if (this._terminal.wraparoundMode) {
           this._terminal.buffer.x = 0;
           this._terminal.buffer.y++;
+          if (this._terminal.buffer.writableRows.indexOf(this._terminal.buffer.y) < 0 ) {
+            this._terminal.buffer.writableRows.push(this._terminal.buffer.y);
```
npezza93: Hmm I was testing on MacOS and it did...And we need it so when some clicks we can tell if they are clicking in a row that can be typed in. 

> Since telling the shell to move left when it's at the bounds couldn't we just send the instructions for multiple lines without even verifying this?

I guess we could, I was just trying to get the least amount of extra arrow clicks for performance reasons. Want me to go that way?

src/Terminal.ts
npezza93: Yea, when working on double width chars I found that this.

meastes:
I don't believe this approach will work in text editors like nano. I tested it out using nano in Chrome, and alt-clicking doesn't move the cursor at all.

Tyriar:
> I don't believe this approach will work in text editors like nano. I tested it out using nano in Chrome, and alt-clicking doesn't move the cursor at all.

I notice in Terminal.app it works in vim, it would be cool if we did support this.

npezza93:
Alright, so alt clicking now supports the normal and alt buffer, along with n wide characters. I tried making some tests but couldn't get it working. Let me know what you guys think about the code.

Tyriar:
I just tested this in vim and it seems to almost works, but when multiple lines are encountered it seems to jump a little too much. Anyone know why this would not be working?

mofux:
I can image why it is not working. We are basically manipulating the cursor by sending the key events to move it. Now in vim and other editors, reaching the end of a line (text) will automatically send you to the next row, even if the actual end of the screen is not yet reached. I think this behaviour is what is breaking the current implementation. To make it work one probably has to send the keys, then wait for the echo and check if the cursor is really in that position. Would be interesting to capture the tty data that `Terminal.app` sends in that situation, because they seem to have figured out a reliable way to do it.

npezza93:
@Tyriar Mind trying again? I found a bug when alt-clicking inside the alt buffer that is now resolved. I found if you don't click squarely in the middle of the character it will sometimes go one character too far. Let me know if you find any more issues!

Tyriar:
Still seeing some issues:

kapture 2017-10-23 at 13 43 16

npezza93:
Good find @Tyriar. What was happening was you were going from a long line to a short line. In vim if you arrow up from a long line to a short line it places your cursor at the end of the line instead of staying in the same column. I updated it so that it does the horizontal movement first then the vertical movement in that scenario.

Tyriar:
Still doesn't work when vim contains wrapped lines. I notice now that Terminal.app doesn't work perfectly either and normally needs 2 alt+clicks to get to the right spot when lines are wrapped, that's still a bit better than our implementation which may not get to the right line with multiple alt+clicks:

alt-click

There is also a remaining issue where clicking on the right half of a character will go to the following character:

offset

src/Terminal.ts
```diff
@@ -943,7 +944,10 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
       ev.preventDefault();
       this.focus();
 
-      if (!this.mouseEvents) return;
+      if (!this.mouseEvents) {
```
Tyriar: Not sure if we've discussed this yet but alt+click already has the functionality of forcing regular selection, even in mouse events mode. Currently this causes issues as both are applied at once:

force-select

@parisk @mofux what are your thoughts on this? Maybe we should have an option that defined the alt+click behavior? I also noticed that Terminal.app does box selection with alt+click.

src/handlers/AltClickHandler.ts
```diff
@@ -0,0 +1,136 @@
...
+      let horizontalChars = this.buildArrowSequence(this.altHorizontalCharCount(), this.horizontalCursorCommand(this.altMoveForward()));
+
+      if (this.altMoveForward()) {
+        keyboardArrows = verticalChars + horizontalChars;
+      } else {
+        keyboardArrows = horizontalChars + verticalChars;
+      }
+    }
+
+    this._terminal.send(keyboardArrows);
```
Tyriar: I wonder if doing multiple `.send` passes is feasible, the only issue I see is that it may be difficult to know when the pty comes back with the final result (especially in remote server scenarios). It would certainly be better for reliably positioning the cursor in apps like vim.

src/Terminal.ts
mofux: IMHO both features can live in parallel (as they do in Terminal.app). I notice that we currently trigger the alt+click cursor move on `mousedown`, which is what interferes with the selection alt+drag behaviour. If we were listening for the `click` event only, that should make it work conflict-free.

npezza93:
Hey @Tyriar! Sorry about the delay on this. Going to try to fix it up this weekend. I started looking into: 
> Still doesn't work when vim contains wrapped lines. I notice now that Terminal.app doesn't work perfectly either and normally needs 2 alt+clicks to get to the right spot when lines are wrapped, that's still a bit better than our implementation which may not get to the right line with multiple alt+clicks:

Let's take your example, the readme, the third line through the ninth is represented by one line in node-pty and is displayed as 6 separate buffer lines in xterm. So if you alt click on the fourth line from the first line of the readme `getCoords` says that the mouse is on the fourth line and that we have to travel three lines down to get there. But in reality, that fourth line is actually the third line and we should only be traveling two. Do we have any way to distinguish if a line is a wrapped line? Kind of stuck how to solve this problem, so I'm hoping you might have an idea to approach this :).

>I wonder if doing multiple .send passes is feasible, the only issue I see is that it may be difficult to know when the pty comes back with the final result (especially in remote server scenarios). It would certainly be better for reliably positioning the cursor in apps like vim.

I tried this with not much luck since this ("difficult to know when the pty comes back with the final result") exact issue. Say you alt-clicked like three cols to the right of the cursor. Initially, I had wanted to send one right arrow command at a time and see if a) we reached our destination, or b) did the cursor move at all, before sending the next cursor. But I couldn't get it to work so I scrapped it.

CodeTheory:
Hey @npezza93! Any update on this? 🙏 
Would be a great enhancement to tick off in Hyper

npezza93:
@timothyis Hey! Currently working on finishing it up. Hopefully will have it finished up by the weekend.

npezza93:
> Still doesn't work when vim contains wrapped lines. I notice now that Terminal.app doesn't work perfectly either and normally needs 2 alt+clicks to get to the right spot when lines are wrapped, that's still a bit better than our implementation which may not get to the right line with multiple alt+clicks:

Got this working better than terminal.app 😄. Works with only one click in both the normal and alt buffer and with wide chars. 

> There is also a remaining issue where clicking on the right half of a character will go to the following character:

Fixed.

> IMHO both features can live in parallel (as they do in Terminal.app). I notice that we currently trigger the alt+click cursor move on mousedown, which is what interferes with the selection alt+drag behaviour. If we were listening for the click event only, that should make it work conflict-free.

I tried moving the alt click inside the click handler and it still had conflicts. Any ideas?

Let me know what you think! Thanks :)

src/handlers/AltClickHandler.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  /**
+   * Arrow escape sequences
+   */
+
+  /**
+   * Constructs the escape sequence for the left or right arrow
+   * @param forward Right arrow or left arrow
+   */
+  private _colSequence(forward: boolean): string {
```
Tyriar: This should be nicer merged with `_rowSequence`  where you pass in an enum instead:

```ts
enum Direction {
  Up,
  Down,
  Right,
  Left
}
```

You could also stick the characters in a map const at the top to make the function super simple:

```
const directionToSequence = {
  Direction.Up: A,
  Direction.Down: B,
  Direction.Right: C,
  Direction.Left: D
}

...

private _sequence(direction: Direction): string {
  const mod = this._terminal.applicationCursor ? 'O' : '[';
  return C0.ESC + mod + directionToSequence[direction];
}
```

src/handlers/AltClickHandler.ts
```diff
@@ -0,0 +1,282 @@
+/**
+ * Alt+Click handler module: exports methods for handling all alt+click-related
+ * events in the terminal.
+ * @module xterm/handlers/AltClickHandler
+ * @license MIT
+ */
+
+import { Terminal } from '../Terminal';
```
Tyriar: Can you use the interfaces `ITerminal` & `ICircularList` instead? That prevents a circular dependency which can be a pain to manage.

src/handlers/AltClickHandler.ts
```diff
@@ -0,0 +1,282 @@
...
+export class AltClickHandler {
+  private _terminal: Terminal;
+  private _startRow: number;
+  private _startCol: number;
+  private _endRow: number;
+  private _endCol: number;
+  private _lines: CircularList<LineData>;
+  private _mouseEvent: MouseEvent;
+
+  constructor(mouseEvent: MouseEvent, terminal: Terminal) {
```
Tyriar: Handy TS feature can save a few lines:

```ts
constructor(
  private _mouseEvent: MouseEvent,
  private _terminal: Terminal
) {
```

You then don't need to do the assignment or declaration.

src/handlers/AltClickHandler.ts
```diff
@@ -0,0 +1,282 @@
...
+    this._terminal = terminal;
+    this._lines = terminal.buffer.lines;
+    this._startCol = this._terminal.buffer.x;
+    this._startRow = this._terminal.buffer.y;
+
+    [this._endCol, this._endRow] = this._terminal.mouseHelper.getCoords(
+      (this._mouseEvent = mouseEvent),
```
Tyriar: Avoid assignments inside other statements, doing the constructor suggestion above should fix this.

Tyriar:
> I tried moving the alt click inside the click handler and it still had conflicts. Any ideas?

How about we trigger the alt click stuff from inside the SelectionManager? The condition on whether to do it could be something like if the selection is <=1 character long and < 500ms have elapsed since mousedown?

npezza93:
Thanks for the review, @Tyriar! I think I've addressed all your feedback and got Travis green.

Tyriar:
I'll wait for a review from @parisk

mofux:
Works even better than Terminal.app, great!
Time to run `nc rya.nc 1987` in our favourite web terminal and celebrate 😅

parisk:
Code looks 👍  and it also works consistently well 😄 .

Issue #897, opened 2017-08-18
Remove dead options from DEFAULT_OPTIONS


Tyriar:
Some of these do nothing (eg. focusKeys), we should tidy it up:

https://github.com/sourcelair/xterm.js/blob/ae0d5e4ffbde93c2d77e4c52fa07e5cab0eb3c2e/src/Terminal.ts#L143-L161

❤️  embedded snippets

PR #899, opened 2017-08-19
Add explicit types

### LINKED ISSUE #898: Explicitly create type declaration files
Recently on the v3 branch I turned on `"declaration": true`. This generates typings for all objects which isn't really what we want. I also notice that it doesn't declare it as a module (`declare module "xterm"`) which might cause problems.

I suggest we handcraft these typings, this will also be a good way of enforcing what is in the public API and what is not.

### LINKED ISSUE #901: Proposal: Remove 'open' event
The `Terminal.open` call is synchronous so the open event seems a bit useless.

### LINKED ISSUE #903: Proposal: slim down refresh and resize events
The resize event handler gives the Terminal object in it's data object, the refresh event handler gives the element in it's data object. Both can just be grabbed off the terminal during the `on` call.

Tyriar:
Fixes #898
Fixes #901 
Fixes #903

---

This change brings in manual typings, previous the automatically generated typings via the `declaration` property in tsconfig.json contained everything which is not what we want to expose to consumers. After this goes in, `typings/xterm.d.ts` contains **the complete API** (excluding addons currently).

There are also simple tests in `fixtures/typings-test/` that ensure that TS code compiled against the typing files compiles as expected. I found a bug while writing this which is always a good sign that it's useful 😄 

---

TODO list:

- [x] Only allow the actual event types in `on` ('a' | 'b')
- [x] Provide types in `on` event listeners https://xtermjs.org/docs/api/terminal/#events
- [x] Only allow the actual addon types in `loadAddon` ('a' | 'b')
- ~~Ensure comments work for functions with overloads (getOption/setOption)~~
- [x] Add a comment for each function with an overload
- [x] Remove general function declarations when there are overloads?
- ~~Consider adding addon interfaces to the .d.ts~~ (maybe later?)
- [x] Add the link matcher API with experimental notice
- [x] Can we test this? We want CI to reject things like `Terminal.scrollDisp` getting removed or renamed
- [x] Should we remove the overloads in getOption, etc. from src/ so they're not duplicated?
- [x] Add tests for bellSound/bellStyle added in https://github.com/sourcelair/xterm.js/pull/878
- ~~Clean up and finalize valid options~~ #897


Tyriar:
Neat 😄 

screen shot 2017-08-18 at 5 31 17 pm

Tyriar:
screen shot 2017-08-18 at 5 45 54 pm

Tyriar:
Ready for review, updated top comment to explain the change more.

Tyriar:
@parisk 

1. :+1:
2. I was starting to go through the breaking changes but thought maybe I should stop as the PR is already huge.

Tyriar:
I allowed null for cursorStyle as well since it's similar to bellStyle.

PR #906, opened 2017-08-19
When bellSound changes, update the bellAudioElements src

### LINKED ISSUE #904: Changing bellSound will not apply until bellStyle is changed to 'none' and back
FYI @npezza93

npezza93:
Fixes #904

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -485,8 +485,8 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
         this.viewport.syncScrollArea();
         break;
       case 'tabStopWidth': this.setupStops(); break;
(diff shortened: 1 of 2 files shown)


src/Terminal.ts
```diff
@@ -2303,6 +2304,12 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
       this.helperContainer.removeChild(this.bellAudioElement);
     }
   }
+
+  private syncBellSound(): void {
```
Tyriar: Let's merge this functionality into preloadBellSound, and let's call is syncBellSound since it also removes.

Tyriar:
Thanks for fixing this 😃

PR #921, opened 2017-08-25
Do not browserify xterm again inside search addon. Switch check in um…

ficristo:
…d wrapper

The change to the umd wrapper is because seems to fix https://github.com/sourcelair/xterm.js/issues/881

I done the PR on top of master in the hope of a new patch release.

### DIFF
--- a/src/addons/search/search.ts
+++ b/src/addons/search/search.ts
@@ -11,21 +11,25 @@ declare var require: any;
 declare var window: any;
 
 (function (addon) {
-  if ('Terminal' in window) {
-    /**
-     * Plain browser environment
+  // One of the most prominent consumer is VSCode which uses a CommonJS environment and has global
+  // 'define' function exposed through its loader (https://github.com/Microsoft/vscode-loader).
+  // Checking only define first would break it, so we check for an AMD environment with 'define.amd'
+  // which is compliant with RequireJS.
+  if (typeof define === 'function' && define.amd) {
+    /*
+     * AMD environment
      */
-    addon(window.Terminal);
+    define(['../../xterm'], addon);
   } else if (typeof exports === 'object' && typeof module === 'object') {
-    /**
+    /*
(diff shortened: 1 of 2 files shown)


src/addons/search/search.ts
```diff
@@ -11,21 +11,21 @@ declare var require: any;
...
 
 (function (addon) {
-  if ('Terminal' in window) {
-    /**
-     * Plain browser environment
-     */
-    addon(window.Terminal);
-  } else if (typeof exports === 'object' && typeof module === 'object') {
-    /**
+  if (typeof exports === 'object' && typeof module === 'object' && typeof define !== 'function') {
```
Tyriar: This would break for VS Code since define is a global (non-requirejs module loader) but we need to use commonjs for this.

ficristo: Does VSCode also have `define.amd` ?
Otherwise I could check first something like
`if (typeof define === 'function' && define.amd)`

Tyriar: Seems fine 
screen shot 2017-08-27 at 12 38 07 pm

Tyriar: If we're going with this it should have a comment mentioning 'vscode-loader' https://github.com/Microsoft/vscode-loader

ficristo:
I switched the check to verify AMD first and added a comment, I hope it is good enough.

ficristo:
@Tyriar does this work for you? If so I'll open a new PR with only the addon change.

Tyriar:
@ficristo looks like it should work, it's a bit tough to test based on the master branch. Could you add the change based on the `v3` branch which is where we're doing new dev currently?

Tyriar:
I'm consolidating all bundling/loadAddon related issues into https://github.com/sourcelair/xterm.js/issues/1018, please comment on that issue if you have thoughts :smiley:

PR #926, opened 2017-08-26
Do not browserify xterm again inside search addon.

ficristo:


Tyriar:
Hmm, even after this change it seems xterm.js is being pulled in to the search addon?

Tyriar:
Ah, it's happening because https://github.com/sourcelair/xterm.js/pull/834 was merged. That's the reason I kept it separate; as it breaks the ./build/ output.

/cc @parisk @blink1073

Tyriar:
Created https://github.com/sourcelair/xterm.js/issues/927 to discuss.

ficristo:
It seems to work for me. The resulting search file drops from ~200KB to ~5KB.

If it's about v3, I think you need to change this:
`.external(path.join(outDir, 'xterm.js'))`
to this
`.external(path.join(outDir, 'Terminal.js'))`

Tyriar:
Ah you're right, that was the problem 😄

Tyriar:
I'll work on an adaption for `v3`

PR #670, opened 2017-05-21
Reimplement selection in the terminal

Tyriar:
This change re-implements the selection model within the terminal, taking the responsibility away from the browser and into xterm.js. This may seem like a lot of work to replace something we already get for free, but selection in the browser was never meant to work with the unique rendering model of xterm.js and other quirks with terminals. As such, this implementation will fix a bunch of bugs, probably most notably is the ability to select and copy more than a single viewport worth of data.

## Checklist

- [x] Selection
  - [x] Drag selection works within viewport
    - [x] Works with wide chars
  - [x] Drag selection on bounds of viewport scrolls the buffer
  - [x] Double click
    - [x] Selects a word
    - [x] Selects whitespace
    - [x] Works with wide chars
  - [x] Triple click selects a line
    - [x] Works with wide chars
    - [x] Dragging selects multiple full lines
  - [x] Double and triple clicks allow the selection to be expanded
  - [x] Selection is disabled when mouse events are enabled
  - [x] Ensure the selection is being cleared when appropriate (eg. if a line is removed in the middle of the selection)
  - [x] Select all
    - [x] Basic implementation
    - [x] cmd+a works
  - [x] Zoom
    - [x] Selection is drawn in correct position (fixed after https://github.com/sourcelair/xterm.js/pull/680 merged)
  - [x] Shift+click should expand selection when there is already a selection (send to terminal as normal if there is no selection)
- [x] Copy
  - [x] Selection can be copied to clipboard via cmd/ctrl+c
  - [x] Selection can be copied to clipboard via context menu
  - [x] Works with wide chars
  - [x] Alt buffer
    - [x] Copying in the alt buffer will grab from the alt buffer, not the normal buffer
- [x] Paste via context menu
- [x] API
  - [x] There is sufficient API to reimplement copy outside of xterm.js (eg. for custom hotkeys in Electron)
- [ ] Testing
  - [x] Ensure there is sufficient test coverage
  - [x] Ensure selection works within tmux
  - [x] Ensure copy works within tmux
  - [x] Do some stress testing
  - [ ] Test several browsers
    - [x] Chrome v58, Linux
    - [ ] Firefox, Linux
    - [x] Chrome, Mac
    - [x] Safari v10, Mac
    - [x] Chrome v59, Windows
    - [x] Edge v40, Windows
    - [x] ~~IE v11, Windows~~ (demo broken atm https://github.com/sourcelair/xterm.js/issues/684)
- [ ] Polish/clean up
[...]

PR #938, opened 2017-09-04
Render using canvas

### LINKED ISSUE #475: Line Spacing Issues
There are some odd line spacing issues.  The line spacing is set at 21px in the demo, but this makes spaces between the lines.  Setting it to 18px will get rid of the spaces, but now mouse coordinates are all off since they are based on the 21px line height.  Worse, some apps (like htop - see screenshot) display the status bar half-over the bottom of the terminal.   I think these are related.

## Details
- Chrome 53
- OS version: Antergos(Arch) Linux 
- xterm.js version: git ... 2.2.3-99-gfc528f6

Here's HTOP hanging over the edge
screenshot from 2017-01-13 02-17-13

Here is Midnight Commander looking awful (spaces between lines):
screenshot from 2017-01-13 02-44-36

And the expected behavior of Midnight Commander (shown in shellinabox, colors set to linux terminal):
screenshot from 2017-01-13 02-44-59

### LINKED ISSUE #681: Improve focus/blur state
Currently clicking in the terminal will cause a blur on mousedown and a focus on mouseup (provided there is no selection). This results in the cursor showing a non-focused state even though technically the terminal is still focused and keyboard input will go to the shell.

Notice the selection and the cursor state:

image

We should put a layer in between the browser textarea blur/focus events and expose our own events, that way we can cover these mouse-related edge cases without affecting the cursor state.

### LINKED ISSUE #720: Selection should not change text color

### LINKED ISSUE #730: xterm-bold is being added when bold font widget differs
See https://github.com/Microsoft/vscode/issues/29402#issuecomment-311137534

It seems that if the bold broken logic runs when the terminal is detected from the DOM or not visible it will break it for the life of the terminal as it's only measured once. We should do a better job of re-evaluating broken bold state when the font changes and to redo it later if the terminal is not in the right state.

### LINKED ISSUE #919: Error: Could not release an element not yet acquired
Seems like we ran into this error in production in SourceLair and looks like it's breaking `refresh`.

@Tyriar do you have any idea if we can handle this in the object reusing pool or what would be a good practice of handling it downstream?

https://github.com/sourcelair/xterm.js/blob/7645bdd418b37a8bd176a69179c06178c8cb1130/src/utils/DomElementObjectPool.ts#L49-L51

---
## Details
- Sentry Error: https://sentry.io/share/issue/35323433302e333334303630393535/
- Browser and browser version: Chrome 60
- OS version: Windows 8
- xterm.js version: 2.9.2

### LINKED ISSUE #920: NotFoundError: Failed to execute 'removeChild' on 'Node'
We also ran into this issue on SourceLair, at the same time with #919 (same user).

Seems like `rowContainer` can be detached from `term.element` at some points 🤔.

https://github.com/sourcelair/xterm.js/blob/6e9d60d76d6cfb3d279f9e58499ba62ca6af479c/src/Renderer.ts#L129-L131
---
## Details
- Sentry Error: https://sentry.io/share/issue/35323433302e333334303630393631/
- Browser and browser version: Chrome 60
- OS version: Windows 8
- xterm.js version: 2.9.2

### LINKED ISSUE #935: Implement a canvas renderer
A work in progress is already well under way. The performance benefits seem to far outweigh anything that can be accomplished in the DOM.

The main drawbacks of using canvas over the DOM:

- We lose functionality provided by text and node in the DOM, namely:
  - Selection & context menus, this doesn't matter as we already reimplemented them to work around then DOM getting in the way
  - Links just work in the DOM; create an `<a>` and just put it somewhere. This needs more handling code to work in a canvas.
- Zooming in will blur the text, so it's best used at 100% page scale

I'll list most of the improvements and bug fixes that canvas provided in the PR which should be out soon 😄 

Marking as breaking as we need new API around colors and font styles which will be managed in TS.

### LINKED ISSUE #936: Actual default FG/BG aren't used when defaults are inversed
Repro:

```
echo -e '\x1b[7mabc\x1b[m'
```

ANSI black and white are used, rather than the default fg and bg colors.

### LINKED ISSUE #937: vtop features misaligned lines
Related to floating point widths and the braille characters.

image

Tyriar:
@parisk @mofux I don't expect the code to change much now, it's mainly just tests, documentation and any bugs that pop up at this point. Please review, I know it's *big* but most of the changes are in new files inside `src/renderer` :smile:

Resolves #935
Fixes #936
Fixes #681
Fixes #730
Fixes #720
Fixes #475
Fixes #937
Fixes #920 
Fixes #919

---

## New Stuff!

- Ascii characters with their foreground, 8 color and 8 bright/bold colors are pre-rendered and stored in an `ImageBitmap`. This is used as a texture atlas to draw to canvas really fast.

  image
- Multiple canvases are used to layer different things onto the terminal, this method is fast and allows for a nice separation of logic.

  image
- Render layers keep track of what they have drawn (when it makes sense) and as such will diff the new contents and only draw when necessary. This turns single character changes in lines to typically take < 0.25ms, down from ~6ms.

  image
- Better support for line height, font size and font family. As they're all managed within `Terminal` now, just `setOption` and run the `fit` addon if necessary.

  image
- There is now a *very* clean separation from the model and everything rendering, pretty mcuh everything that renders is within `./src/renderer` (viewport being the only exception, which needs a background color set). We should move towards doing similar with the model and the DOM in Terminal to simplify and make testing easier.
- Link handlers are getting reworked quite a bit as previously the `validationCallback` exposes the `HTMLElement` of the link to allow tooltips to be attached. There is now a `hoverCallback` which achieves a similar goal.
- This fixes a bunch of bugs (see the link of "Fixes #x" above), a notable bug fixes is line characters being misaligned in `vtop` https://github.com/sourcelair/xterm.js/issues/937.

### API Changes

See complete list of changes in the [xterm.d.ts diff](https://github.com/sourcelair/xterm.js/pull/938/files#diff-3cf93a85b230e197db3cdf45b888018f).

## Limitations & regressions

- Zooming the browser window will zoom the canvas, meaning at > 100% the text may be a little blurry.
- Blink character attribute support was dropped. This is a little difficult to implement and not really worth the effort given its usage. https://github.com/sourcelair/xterm.js/issues/944
- Link underlines are no longer the same color as the fg color https://github.com/sourcelair/xterm.js/issues/957
- Ligatures are no longer supported https://github.com/sourcelair/xterm.js/issues/958

## Deferred

- Add new features to demo https://github.com/sourcelair/xterm.js/issues/939
- The selection render layer should do incremental draws, currently it clears and redraws the entire selection https://github.com/sourcelair/xterm.js/issues/941
- Window resizes that change devicePixelRatio is expensive (60ms+) https://github.com/sourcelair/xterm.js/issues/955

## Benchmarks

This table contains the average time it takes to render when running various actions or programs.

| Render time | `2.9.2` | `canvas_render` | Improvement
|---|---|---|---
| Typing a single letter | 4-5ms | 0.4-0.5ms | **10x faster**
| `vtop` idle | 22-73ms\* | 0.8-1.6ms | **27-46x faster**
| `ls -lR large_folder` | 17-25ms\* | 2.3-4.6ms | **5-7x faster**
| `yes` | 17-30ms\* | 1-1.3ms | **17-23x faster**

\* *Rendering took multiple frames*

**Setup**

- Measured on Chrome 60, Macbook Pro.
- Compared https://github.com/sourcelair/xterm.js/commit/cc7c310eee72b8b43d1b5372d4482b4ca54d5b91 (master, 2.9.2) and https://github.com/Tyriar/xterm.js/commit/2d1a1bfb76974219e6a9401e709c2acad501dd15
- The times are all rendering-related chunks exposed in devtools added up (render script, rendering, paint, composite layers).
- I tried to take the smallest and largest samples for the range.

### In pictures

Here's one of my favorites; comparing `vtop`'s performance 🔥.

**Before**

screen shot 2017-09-03 at 9 52 10 am

**After**

screen shot 2017-09-03 at 9 53 53 am

---

## TODO List

- [x] API: Move setting of theme into `Terminal.setOption`
- [x] Better emoji support
- [x] Add selection color to `ITheme`
- [x] Resolve/accept `TODO`'s in code
- [x] Document important classes/methods
- [x] Test
  - [x] Add unit tests
  - [x] Chrome
    - [x] macOS
    - [x] Linux
    - [x] Windows
  - [x] Safari/macOS - Does not support `ImageBitmap` so falls back to `HTMLCanvas`
  - [x] Firefox - Buggy and bad performance drawing from `ImageBitmap`, explicit fallback to `HTMLCanvas`
    - [x] Linux
    - [x] macOS
    - [x] Windows
  - [x] Edge/Windows - seems just as good as Chrome
  - [x] ~~IE11/Windows~~ - 💣 
- [x] Redraw terminal when devicePixelRatio changes (causes blurry text)
- [x] Support floating point devicePixelRatio (causes jagged text)
- [x] Fix right padding getting larger when resizing
- [x] Investigate support for ctrl only clicks in the linkifier (a render layer in an addon?)


parisk:
@Tyriar I think that this is also fixing #920 and #919, right?

src/renderer/BaseRenderLayer.ts
```diff
@@ -0,0 +1,190 @@
...
+    const result = acquireCharAtlas(terminal, this.colors);
+    if (result instanceof HTMLCanvasElement) {
+      this._charAtlas = result;
+    } else {
+      result.then(bitmap => this._charAtlas = bitmap);
+    }
+  }
+
+  public resize(terminal: ITerminal, canvasWidth: number, canvasHeight: number, charSizeChanged: boolean): void {
+    this.scaledCharWidth = terminal.charMeasure.width * window.devicePixelRatio;
```
mofux: I think we have to measure the width and height of a char using the canvas now. Still using the HTML measure method is likely prone to inaccurate results, especially when changing font-size or family, no?

This could be helpful:
https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText

src/renderer/SelectionRenderLayer.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    // No need to draw the selection
+    if (viewportCappedStartRow >= terminal.rows || viewportCappedEndRow < 0) {
+      return;
+    }
+
+    // Draw first row
+    const startCol = viewportStartRow === viewportCappedStartRow ? start[0] : 0;
+    const startRowEndCol = viewportCappedStartRow === viewportCappedEndRow ? end[0] : terminal.cols;
+    this._ctx.fillStyle = 'rgba(255,255,255,0.3)';
```
mofux: The selection color should be configurable, maybe through the theme interface? This value might look okay for white fg on black bg, but it doesn't work for black fg on white bg.

src/renderer/CursorRenderLayer.ts
```diff
@@ -0,0 +1,346 @@
+import { IColorSet } from './Interfaces';
```
mofux: After giving it some thoughts, I would prefer to still render the cursor element as HTML. It's the only really dynamic terminal element that people will likely want to style differently in different states. For example, having the line cursor 2 px wide, or changing the visual style of the inactive block cursor would currently require to modify or overwrite this code, which is not optimal. Another perk of having it as HTML is that we could use smooth transitions between the cursor states (focused, unfocused, blink, bell), which is hard to do with the canvas.

src/renderer/BaseRenderLayer.ts
```diff
@@ -0,0 +1,190 @@
...
+      this._ctx.fillStyle = this.colors.foreground;
+    }
+
+    // Since uncached characters are not coming off the char atlas with source
+    // coordinates, it means that text drawn to the canvas (particularly '_')
+    // can bleed into other cells. This code will clip the following fillText,
+    // ensuring that its contents don't go beyond the cell bounds.
+    this._ctx.beginPath();
+    this._ctx.rect(x * this.scaledCharWidth, y * this.scaledLineHeight + this.scaledLineDrawY, width * this.scaledCharWidth, this.scaledCharHeight);
+    this._ctx.clip();
```
mofux: I think this will mess up emojis. They are rendered as single-width characters, but normally consume the space of 2 cols. Although most terminals render them as single width characters, they will allow emojis to overlap the next col. That col will then draw over the overlapped parts of the emoji. I think we should behave the same.

src/renderer/CursorRenderLayer.ts
```diff
@@ -0,0 +1,346 @@
...
+      this._clearCursor();
+      return;
+    }
+
+    const charData = terminal.buffer.lines.get(cursorY)[terminal.buffer.x];
+
+    if (!terminal.isFocused) {
+      this._clearCursor();
+      this._ctx.save();
+      this._ctx.fillStyle = this.colors.ansi[COLOR_CODES.WHITE];
```
mofux: Maybe we should make the unfocused cursor color configurable?

LucaT1:
What about supporting italic font-style?

Tyriar:
> I think that this is also fixing #920 and #919, right?

@parisk It does indeed.

> What about supporting italic font-style?

@LucaT1 I missed this, will make a note to check it out.

src/renderer/BaseRenderLayer.ts
Tyriar: I did this initially but it turns out that support for getting the proper height dimensions we need is severely lacking across browsers https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics

src/renderer/BaseRenderLayer.ts
Tyriar: Added to the todo list

src/renderer/SelectionRenderLayer.ts
Tyriar: Will add to the ITheme, added to the todo list 👍

src/renderer/CursorRenderLayer.ts
Tyriar: We still needed to manage the animation in JS due to a performance bug in Chrome eaten CPU when using a low FPS animation, for the sake of performance I think we should do it in canvas.

On customizability, I think exposing the component api https://github.com/sourcelair/xterm.js/issues/808 will solve this. For example, components could be written to replace the cursor completely (either allowing to disable the regular cursor render layer or via `.xterm-cursor-layer { display:none}`.

src/renderer/CursorRenderLayer.ts
Tyriar: Opps, this should be `this.colors.CURSOR`, good find 😄

Tyriar: https://github.com/sourcelair/xterm.js/pull/938/commits/0bfdff0b39010b8a49a2396aa1e389de132c3d35

src/renderer/SelectionRenderLayer.ts
Tyriar: https://github.com/sourcelair/xterm.js/pull/938/commits/ff31f345516bd6b6a7b09e3251ba0c7c1047331f

Tyriar:
> What about supporting italic font-style?

@LucaT1 just realised this isn't a regression https://github.com/sourcelair/xterm.js/issues/580. I won't add this to this PR to reduce complexity, it is very much accomplishable though as I checked and italic chars in the font seem to stay in the cells:

screen shot 2017-09-04 at 8 21 11 am

Just need to keep track of the flag against the chars and adding `'italic '` to `this._ctx.font`.

src/renderer/BaseRenderLayer.ts
Tyriar: So Emoji support is a little tricky. Currently all characters "own" their cell, and they're the only ones that can print to it. In the case of regular wide characters, they also own the cell to the right and "null chars" that are next to them (where char=`''`, width=`0`) give it up.

I did a little research and it looks like putting a space immediately after the emoji is what people tend to do:

- https://superuser.com/a/538065
- http://osxdaily.com/2013/04/08/add-emoji-command-line-bash-prompt/

I propose a compromise: Emoji's will continue to own their own cell and they will only own the cell to the right if it's the space char `' '` (code 32). Let me know if you see any issues with this.

Tyriar:
Emoji support!

screen shot 2017-09-04 at 9 24 08 am

The bottom line shows off most states. Here's the cursor on top of the RHS of an emoji:

screen shot 2017-09-04 at 9 25 48 am

src/renderer/BaseRenderLayer.ts
Tyriar: https://github.com/sourcelair/xterm.js/pull/938/commits/a17950290c9951dde0af2862e4eb530c639b0097, https://github.com/sourcelair/xterm.js/pull/938/commits/1c3028c59f9afb4623c93d142aedb9ffc49b4cd6

LucaT1:
Tested the new emoji feature! :tada:
Despite the fact that I'm not able to write any emoji with the MacOS prompt(cmd+ctrl+space), this is how pasting "👍🏻 👈🏿" one of these results. Do you have any idea why skin modifiers aren't supported? They should since it's a web environment.

Platform: OSX
Browser: Chrome 60.0.3112.113
Branch: Latest commit from this PR(9506c01)



screen shot 2017-09-04 at 19 40 01

Tyriar:
@LucaT1 it looks like this is not a regression, it happens on master. It looks like we're defining the modifier as a separate character https://github.com/sourcelair/xterm.js/issues/942

Tyriar:
Tested on some more browsers, Firefox and IE don't work well. I'd be happy dropped IE support from the lib (Windows developers should be using Edge or another alternative at this point), not Firefox though. Will probably have to put the old renderer back and come up with a consistent interface. Investigating..

Tyriar:
Firefox is fine when I force it to use `HTMLCanvas` instead of `ImageBitmap` for the `CharAtlas`. This was the fallback I added to support Safari anyway. :tada:

mofux:
I have encountered a visual glitch with the canvas branch that I was unable to reproduce in master. In `mc`, when editing a file and scrolling, the first row of the screen does not seem to get updated properly. Usually, the first row contains a static menu bar with cyan background that shows the current row and cols the cursor is at. Here's a picture that shows the bug:

mcedit

You can install `mc` via `brew install mc` and then `mcedit some/file.js` to reproduce the bug.

mofux:
Maybe there is something wrong with the cols / rows calculated. As you can see in the screen capture, the most right cell is always empty, just as if the renderer renders cols - 1...

src/renderer/Renderer.ts
```diff
@@ -0,0 +1,133 @@
...
+  private _refreshRowsQueue: {start: number, end: number}[] = [];
+  private _refreshAnimationFrame = null;
+
+  private _renderLayers: IRenderLayer[];
+
+  private _colorManager: ColorManager;
+
+  constructor(private _terminal: ITerminal) {
+    this._colorManager = new ColorManager();
+    this._renderLayers = [
```
mofux: Can we make this a little more generic, so plugins / third party code can more easily register their own layers? Maybe adding a public API to the renderer like this:

```typescript
// register a layer (does not instantiate it)
registerLayer(id: string, zIndex: number, Layer: IRenderLayer): void,

// destroys the current instance of the layer (if existent)
disableLayer(id: string): void,

// creates a new instance of that layer (if not yet existent)
enableLayer(id: string): void
```

Then, we could already register the built-in layers and enable them by default, but could also make the active layers configurable through a config option: 

```javascript
// set the layers that should be enabled by default
term.setOption('renderLayers', ['foreground', 'background', 'selection'])

// ...or via api
term.renderer.disableLayer('cursor');
term.renderer.registerLayer('myCustomCursor', 4, CustomCursorRenderer);
term.renderer.enableLayer('myCustomCursor');
```

To make this work, IRenderLayer probably has to implement some `onDestroy` method that is called when a layer is disabled. There, the layer can clean up any DOM nodes and listeners.

Tyriar:
@mofux looks like the mc issue happens before this PR, probably fallout from the 0 scrollback alt buffer PR https://github.com/sourcelair/xterm.js/pull/855, please create a new issue 😃

Tyriar:
> Maybe there is something wrong with the cols / rows calculated. As you can see in the screen capture, the most right cell is always empty, just as if the renderer renders cols - 1...

Are you sure that's not just the space provided for the scroll bar?

Tyriar:
Something is definitely going wrong when I resize columns down.

screen shot 2017-09-05 at 7 37 48 am

README.md
```diff
@@ -97,12 +97,10 @@ xterm.fit();
 
 Since xterm.js is typically implemented as a developer tool, only modern browsers are supported officially. Here is a list of the versions we aim to support:
 
-- Chrome 48+
-- Edge 13+
-- Firefox 44+
-- Internet Explorer 11+
-- Opera 35+
-- Safari 8+
+- Chrome latest
```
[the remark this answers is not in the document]
Tyriar: Breakages from going up versions are not likely to happen on browsers and they deprecate and change behavior very slowly. The main point this is trying to get across is that this is for developers which should be on updated browsers. I certainly don't go back and check my features on Chrome 48, I do make sure they work on Blink, Gecko, etc. though.

src/renderer/Renderer.ts
Tyriar: This is definitely where I want us to go, I made `IRenderLayer` generic so it would be easy to do that once when want to fiddle with the API.

README.md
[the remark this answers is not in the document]
mofux: But does that also imply that we can start using a browser feature (like IntersectionObserver, ResizeObserver, async / await etc.) once that feature is supported by the aforementioned list of supported browsers?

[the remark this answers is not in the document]
Tyriar: Yes, and we absolutely should imo. `window.createImageBitmap` for example is implemented using progressive enhancement, falling back to a HTMLElement because Firefox and Safari are lacking here. It provides a 2x speed up when I measured it on Chrome though.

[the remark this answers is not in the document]
Tyriar: Things like async/await have the potential to break an awful lot of older browsers so we should be mindful of that, we're not trying to kill old browser support but rather guarantee it works on the latest (unless it becomes a nuisance).

src/renderer/LinkRenderLayer.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  private _clearCurrentLink(): void {
+    if (this._state) {
+      this.clearCells(this._state.x, this._state.y, this._state.length, 1);
+      this._state = null;
+    }
+  }
+
+  private _onLinkHover(e: LinkHoverEvent): void {
+    this._ctx.fillStyle = this.colors.foreground;
```
mofux: We should either make the underline color static but configurable - similar to the selection color, or (preferred) have the underline color match the foreground color of the cell. That would provide a more consistent look, and would also make sure that the underline stroke is always visible (in situations where the cell background matches the foreground):

link

src/renderer/BaseRenderLayer.ts
```diff
@@ -0,0 +1,274 @@
...
+    } else {
+      // If default color and bold
+      if (bold) {
+        colorIndex = 1;
+      }
+    }
+    const isAscii = code < 256;
+    const isBasicColor = (colorIndex > 1 && fg < 16);
+    const isDefaultColor = fg >= 256;
+    if (isAscii && (isBasicColor || isDefaultColor)) {
```
mofux: We still seem to have a problem with at least the `_` character that is only 1px high (whereas I would expect it to be 2px on a retina display, or equal the height of one of the `=` strokes). See the `__proto__` here:

link

mofux: This problem might actually be related to the 15px font size in the demo. Using 14px or any of the font sizes that are supported in `Terminal.app` gives good looking results. I am wondering if that's the reason why `Terminal.app` restricts the font-sizes to 9,10,11,12,13,14,18,24,... 🤔

mofux:
@Tyriar Support for ligatures seems to be broken (works in v3 and master). I think we should check if there is a way to support it with canvas - I'm afraid there are lots of people that use it (and some like [powerline](https://github.com/powerline/powerline) users can hardly live without it).

src/renderer/BaseRenderLayer.ts
Tyriar: I don't think we should include this in this PR but one was we could fix this is by have 2 render layers/canvases for the foreground, one for odd rows, one for even. That would allow characters like `_` to overlap cells while still being able to clear them without interfering with other cells.

Tyriar:
> Support for ligatures seems to be broken (works in v3 and master)

I didn't think powerline used ligatures? https://github.com/powerline/fonts/issues/42 Do you have a screenshot of the problem?

mofux:
@Tyriar Just double-checked, I thought some of the glyphs that are used by powerline are coming from the ligatures, but I was wrong, sorry 😓

src/renderer/BaseRenderLayer.ts
mofux: I've just installed oh-my-zsh, and it doesn't play well with unicode characters that are overlapping. Maybe we need a different way to allow overlapping characters, not just emojis?

xterm.js canvas
xterm

Terminal.app
terminal

src/renderer/LinkRenderLayer.ts
Tyriar: Good point with visibility on different backgrounds

Tyriar: Let's defer this one and pull out the color determination of the fg render layer in another PR. https://github.com/sourcelair/xterm.js/issues/957

parisk:
🎉

src/CompositionHelper.ts
```diff
@@ -63,6 +64,7 @@ export class CompositionHelper {
    * @param {CompositionEvent} ev The event.
    */
   public compositionupdate(ev: CompositionEvent): void {
+  console.log('compositionupdate');
```
Krinkle: Left a logger here I think?

Tyriar: Nice catch 😅 https://github.com/sourcelair/xterm.js/pull/983

arcanis:
I'm curious, does this renderer supports copy/pasting?

weihanglo:
Hi @arcanis.
I am not from the Xterm.js team. But after reading the sources, I think the selection of canvas renderer was built almost as same as the custom selection logic shown in #670. There is a [separate render layer](https://github.com/sourcelair/xterm.js/blob/b06eb44/src/renderer/Renderer.ts#L36) to make selections work. You can also run the [demo](https://github.com/sourcelair/xterm.js#demo) and experience its performance directly.

Tyriar:
@arcanis copy and paste was barely touched at all, only the drawing of the selection and the events that copy/paste listened to were tweaked.

src/renderer/CharAtlas.ts
```diff
@@ -0,0 +1,186 @@
...
+    // performance (tested on v55).
+    if (!('createImageBitmap' in window) || isFirefox) {
+      // Regenerate canvas and context as they are now owned by the char atlas
+      const result = this._canvas;
+      this._canvas = this._document.createElement('canvas');
+      this._ctx = this._canvas.getContext('2d');
+      this._ctx.scale(window.devicePixelRatio, window.devicePixelRatio);
+      return result;
+    }
+
```
weihanglo: Hi. I am curious about why we chose `createImageBitmap` over passing canvas out directly in `CharAtlas`. I thought `ImageBitmap` is for worker thread or [image decoding on background thread](https://bugs.chromium.org/p/chromium/issues/detail?id=580202). Does this gains any performance improvement?

Tyriar: There's a comment that explains this:

```js
// ImageBitmap's draw about twice as fast as from a canvas
```

This is when I initially measured it. I also found that `ImageBitmap` was cripplingly slow on Firefox and wasn't supported on Safari, these fall back to drawing from `HTMLCanvas`.

weihanglo: Got it.`ImageBitmap` is for fast accessing and drawing. Thanks for the reply!

tbodt:
Will selection and copy/paste still work on iOS?

Tyriar:
@tbodt probably not, I'm thinking we should probably bring back the old renderer anyway for environment edge cases. Some computers take much longer to render using canvas which I think it something to do with GPU rendering being disabled or something.

PR #952, opened 2017-09-07
Fix CSI Ps M to delete the correct line

### LINKED ISSUE #947: First row of alt buffer is drawn incorrectly in some situations
Editing a file in `mc` and scrolling (either with scroll wheel or using the arrow keys) renders the first row of the screen incorrectly:
mcedit

This bug was likely introduced by #855 and tried to fix by #923 without success. Note It does only happen on the `v3` branch, not master.

To reproduce, install mc `brew install mc`, then in the demo `mcedit lib/Terminal.js` and scroll using the wheel or `alt+down` and `alt+up`

mofux:
Fixes #947 

This fixes an issue in the v3 branch where CSI Ps M was clearing the wrong line. This bug was introduced by https://github.com/sourcelair/xterm.js/commit/a57f55dcbf7962662d8ba3e6da75bbea96eb331b#diff-5eb84354a70d0df188b3bbb5566120bbL493

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -467,7 +467,7 @@ export class InputHandler implements IInputHandler {
     while (param--) {
       // test: echo -e '\e[44m\e[1M\e[0m'
       // blankLine(true) - xterm/linux behavior
-      this._terminal.buffer.lines.splice(row - 1, 1);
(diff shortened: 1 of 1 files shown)


parisk:
🎉

PR #980, opened 2017-09-13
Draw text using decimal numbers

### LINKED ISSUE #972: Bad letter spacing with canvas
Depending on the font and font size, the letters are spaced inaccurately. For example, using the `monaco` font with a font size of `12px` will result in a real character width of `7.203...px` but charMeasure does `Match.ceil` it to `8px`. This creates a visible `0.8px` additional gap between characters, which causes the font to look somewhat stretched:

Current (with `Math.ceil`):
wide

Expected (without `Math.ceil`):
correct

@Tyriar Do we have to Math.ceil the measured dimensions for some reason?

Tyriar:
Fixes #972

### DIFF
--- a/src/renderer/BaseRenderLayer.ts
+++ b/src/renderer/BaseRenderLayer.ts
@@ -79,6 +79,16 @@ export abstract class BaseRenderLayer implements IRenderLayer {
   public abstract reset(terminal: ITerminal): void;
 
   /**
+   * Gets the left position of a cell. Since character width is stored as a
+   * float in order to prevent bad letter spacing, drawing shapes in the cell
+   * need to be rounded.
+   * @param x The column of the cell.
+   */
+  private _getCellLeft(x: number): number {
+    return Math.round(x * this.scaledCharWidth);
+  }
+
+  /**
    * Fills 1+ cells completely. This uses the existing fillStyle on the context.
    * @param x The column to start at.
    * @param y The row to start at
@@ -86,7 +96,12 @@ export abstract class BaseRenderLayer implements IRenderLayer {
    * @param height The number of rows to fill.
    */
   protected fillCells(x: number, y: number, width: number, height: number): void {
-    this._ctx.fillRect(x * this.scaledCharWidth, y * this.scaledLineHeight, width * this.scaledCharWidth, height * this.scaledLineHeight);
+    const cellLeft = this._getCellLeft(x);
+    this._ctx.fillRect(
+      cellLeft,
+      y * this.scaledLineHeight,
+      this._getCellLeft(x + width) - cellLeft,
+      height * this.scaledLineHeight);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


Tyriar:
Looks a lot better now. The top one is after the change, both windows are using the same font/zoom level:

screen shot 2017-09-12 at 7 07 46 pm

src/renderer/BaseRenderLayer.ts
```diff
@@ -79,14 +79,29 @@ export abstract class BaseRenderLayer implements IRenderLayer {
...
    * @param width The number of columns to fill.
    * @param height The number of rows to fill.
    */
   protected fillCells(x: number, y: number, width: number, height: number): void {
-    this._ctx.fillRect(x * this.scaledCharWidth, y * this.scaledLineHeight, width * this.scaledCharWidth, height * this.scaledLineHeight);
+    const cellLeft = this._getCellLeft(x);
+    this._ctx.fillRect(
+      cellLeft,
+      y * this.scaledLineHeight,
+      this._getCellLeft(x + width) - cellLeft,
```
mofux: How about adding a `this._getCellWidth(x, cells)` function that does does this calculation for us? Would make the code a little easier to read.

src/renderer/BaseRenderLayer.ts
```diff
@@ -96,10 +111,11 @@ export abstract class BaseRenderLayer implements IRenderLayer {
    * @param y The row to fill.
    */
   protected fillBottomLineAtCells(x: number, y: number, width: number = 1): void {
+    const cellLeft = this._getCellLeft(x);
     this._ctx.fillRect(
-        x * this.scaledCharWidth,
+        cellLeft,
         (y + 1) * this.scaledLineHeight - window.devicePixelRatio - 1 /* Ensure it's drawn within the cell */,
-        width * this.scaledCharWidth,
+        this._getCellLeft(x + width) - cellLeft,
```
mofux: Could re-use `this._getCellWidth(x, cells)`

src/renderer/BaseRenderLayer.ts
```diff
@@ -124,11 +140,12 @@ export abstract class BaseRenderLayer implements IRenderLayer {
...
    */
   protected strokeRectAtCell(x: number, y: number, width: number, height: number): void {
+    const cellLeft = this._getCellLeft(x);
     this._ctx.lineWidth = window.devicePixelRatio;
     this._ctx.strokeRect(
-        x * this.scaledCharWidth + window.devicePixelRatio / 2,
+        cellLeft + window.devicePixelRatio / 2,
         y * this.scaledLineHeight + (window.devicePixelRatio / 2),
-        (width * this.scaledCharWidth) - window.devicePixelRatio,
+        this._getCellLeft(x + width) - cellLeft - window.devicePixelRatio,
```
mofux: Could re-use `this._getCellWidth(x, cells)`

src/renderer/BaseRenderLayer.ts
```diff
@@ -147,7 +164,12 @@ export abstract class BaseRenderLayer implements IRenderLayer {
    * @param height The number of rows to clear.
    */
   protected clearCells(x: number, y: number, width: number, height: number): void {
-    this._ctx.clearRect(x * this.scaledCharWidth, y * this.scaledLineHeight, width * this.scaledCharWidth, height * this.scaledLineHeight);
+    const cellLeft = this._getCellLeft(x);
+    this._ctx.clearRect(
+        cellLeft,
+        y * this.scaledLineHeight,
+        this._getCellLeft(x + width) - cellLeft,
```
mofux: Could re-use `this._getCellWidth(x, cells)`

src/renderer/BaseRenderLayer.ts
Tyriar: Sounds good :smiley:

mofux: Sounds reasonable 👍 Guess we're good to go then!

PR #981, opened 2017-09-13
Treat \u279C arrow as ambiguous width char (like emoji)

### LINKED ISSUE #974: Popular "wide" characters are getting cropped
https://github.com/Microsoft/vscode/issues/34138

`➜`

image

Tyriar:
Fixes #974

---

Works fine now when a space is next to it:

screen shot 2017-09-12 at 7 24 09 pm

### DIFF
--- a/src/renderer/ForegroundRenderLayer.ts
+++ b/src/renderer/ForegroundRenderLayer.ts
@@ -159,6 +159,11 @@ export class ForegroundRenderLayer extends BaseRenderLayer {
    * @param char The character to search.
    */
   private _isEmoji(char: string): boolean {
+    // Check special ambiguous width characters
+    if (char === '➜') {
+      return true;
+    }
+    // Check emoji unicode range
     return char.search(/([\uD800-\uDBFF][\uDC00-\uDFFF])/g) >= 0;
   }
 
--- a/src/renderer/ForegroundRenderLayer.ts
+++ b/src/renderer/ForegroundRenderLayer.ts
@@ -159,6 +159,7 @@ export class ForegroundRenderLayer extends BaseRenderLayer {
    * @param char The character to search.
    */
   private _isEmoji(char: string): boolean {
+    // TODO: We need a generic solution for handling characters like this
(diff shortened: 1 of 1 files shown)


mofux:
I think we should tackle this whole character overlapping issue differently to be honest. We can never really know which character is overlapping without measuring its width individually. As far as I know, depending on the chosen font, if a character is not available in that particular font, it will fall back to the next font in the chain until it finds one that provides that character. 

Or the other way around, the \u279C character might fit into a single cell just fine if the chosen monospace font would provide it.

mofux:
`Terminal.app` and `iTerm2` simply allow overlapping characters, even if they are not followed by a space. I think we should mimic that behaviour:

iterm

terminal

mofux:
Here are some very nice insight to using multiple canvas elements to create layers, and why we probably don't want to do it (memory overhead):

https://www.ibm.com/developerworks/library/wa-canvashtml5layering/

mofux:
Maybe one way to tackle this problem is to dirty flag a cell if 
- it was modified
- is an overlapping cell and is followed by the modified cell

To make this work, we would have to abstract all the drawing behind an API that is used by the layers (it's already there in the BaseRenderLayer I guess), and then create a mental model of the cell changes, without explicitly drawing them yet. Then we would know which cells are dirty because they have been updated, and which cells would have to be redrawn because the are overlapping.

Finally the layer renderer has to tell us when it's finished, so we can do the work and plot the changes...

Tyriar:
@mofux 

> I think we should tackle this whole character overlapping issue differently

How about we merge this as it's the major problem case and keep the issue open until a more general fix is put in place? I won't have time to work on it for the next couple of weeks as I'm doing some travel.

> insight to using multiple canvas elements to create layers

I've been tracking GPU memory while developing and it doesn't seem to go above 25mb in VS Code, that's with the rest of the application as well.

mofux:
@Tyriar Well fine, maybe we can add a TODO comment to that particular part of the code, so we don't forget to clean up later on 🤓

PR #991, opened 2017-09-14
Support sub-pixel anti-aliasing and fix blurry text issues

### LINKED ISSUE #973: Support sub-pixel anti-aliasing
VS Code issue: https://github.com/Microsoft/vscode/issues/34119

Text needs to be drawn on a canvas with the background color in order to enable sub-pixel anti-aliasing. May need to rethink the layers, possibly merg background, foreground and selection layers?

### LINKED ISSUE #985: Text rendering is a little blurry still
Notice "daniel" gets blurry a few rows down when compared to rows above. Changing the canvas height from 500 -> 501 seems to fix it so it's probably a problem with the `IRenderDimensions` generation.

image

### LINKED ISSUE #987: _drawUncachedChar does not draw bold characters
```
_drawUncachedChar(terminal: ITerminal, char: string, width: number, fg: number, x: number, y: number): void
```

### LINKED ISSUE #988: Canvas is not cleared correctly
Since https://github.com/sourcelair/xterm.js/pull/980 the canvas does not seem to be cleared correctly.

Reproduce:
- Start the xterm.js demo
- run `ls -alh`
- scroll all the way up and back down
- type `clear`
- see the artifacts that remain

clear issue

screen shot 2017-09-13 at 22 26 12

Tyriar:
This PR does the following:

- Characters are now drawn to an integer grid, but cell width is now *floored*. Originally it was ceiled which made the text look odd, and then a floating point number was used which makes the CharAtlas optimizations look blurry. Now characters look great whether they're cached or uncached.
- Background and foreground layers are now merged, this was necessary to support *sub-pixel anti-aliasing* as the effect only applies when drawn only an opaque background.
- Bold is now supported for uncached text.

Fixes #988
Fixes #987
Fixes #973

---

### Browser zoom 100%

Before:

screen shot 2017-09-13 at 11 37 55 pm

After:

screen shot 2017-09-13 at 11 34 11 pm

### Browser zoom 100%, magnified

Before zoomed in:

screen shot 2017-09-13 at 11 39 18 pm

After zoomed in:

screen shot 2017-09-13 at 11 35 20 pm

### Browser zoom 150%

Before:

screen shot 2017-09-13 at 11 39 50 pm

After:

screen shot 2017-09-13 at 11 36 48 pm


Tyriar:
I just covered the following cases:

- Draw background on text layer before the canvas is shown to prevent a black flash when the background isn't #000
- Draw background on char atlas to properly support non-#000 backgrounds
- Redraw background when the canvas is resized

screen shot 2017-09-13 at 11 56 13 pm


screen shot 2017-09-13 at 11 57 54 pm

Tyriar:
The paint flashing in 2 isn't actually re-drawing the whole line, Chrome's just saying "things in this rectangle are changing"

PR #994, opened 2017-09-15
[search addon] Change umd check to be compatible with Require.js (v3)

ficristo:
@Tyriar if it does work for you, would you accept a backport to 2.9?

### DIFF
--- a/src/addons/search/search.ts
+++ b/src/addons/search/search.ts
@@ -12,21 +12,25 @@ declare var require: any;
 declare var window: any;
 
 (function (addon) {
-  if (typeof window !== 'undefined' && 'Terminal' in window) {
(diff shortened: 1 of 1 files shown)


parisk:
Thanks for your contribution @ficristo.

We try avoiding backporting things to 2.9 now that 3.0 is close (about a month from now).

ficristo:
I'm using xterm for a Brackets extension, and Brackets still runs on an old version of CEF (something like Chrome 49). I don't have any ETA of when they will upgrade their shell.
So, unfortunately, I don't think it is an option for me to use v3 anytime soon.

Tyriar:
I'm consolidating all bundling/loadAddon related issues into https://github.com/sourcelair/xterm.js/issues/1018, please comment on that issue if you have thoughts :smiley:

PR #996, opened 2017-09-15
Fix inverted cells, make sure inverted cells are cleared correctly

### LINKED ISSUE #995: Nano does not render correctly
In the current v3 branch with canvas, `nano` does not render correctly anymore:

screen shot 2017-09-15 at 14 58 39

vs master:

screen shot 2017-09-15 at 14 59 34

mofux:
Fixes #995 

Addresses two issues:
1) Previously, the foreground was not correctly swapped, getting the same color as the background when inversed.

2) Removes a render optimisation that caused the white top bar in nano not to render and also caused an issue when redrawing color inverted cells that contained a space.

### DIFF
--- a/src/renderer/TextRenderLayer.ts
+++ b/src/renderer/TextRenderLayer.ts
@@ -121,7 +121,7 @@ export class TextRenderLayer extends BaseRenderLayer {
         let fg = (attr >> 9) & 0x1ff;
 
         // If inverse flag is on, the foreground should become the background.
-				if (flags & FLAGS.INVERSE) {
+        if (flags & FLAGS.INVERSE) {
           const temp = bg;
           bg = fg;
           fg = temp;
--- a/src/renderer/TextRenderLayer.ts
+++ b/src/renderer/TextRenderLayer.ts
@@ -83,7 +83,7 @@ export class TextRenderLayer extends BaseRenderLayer {
         }
 
         // Clear the old character was not a space with the default background
-        const wasInverted = state && state[CHAR_DATA_ATTR_INDEX & FLAGS.INVERSE];
+        const wasInverted = !!(state && state[CHAR_DATA_ATTR_INDEX] && state[CHAR_DATA_ATTR_INDEX] >> 18 & FLAGS.INVERSE);
         if (state && !(state[CHAR_DATA_CODE_INDEX] === 32 /*' '*/ && (state[CHAR_DATA_ATTR_INDEX] & 0x1ff) >= 256 && !wasInverted)) {
           this._clearChar(x, y);
         }
--- a/src/renderer/TextRenderLayer.ts
+++ b/src/renderer/TextRenderLayer.ts
@@ -94,7 +94,7 @@ export class TextRenderLayer extends BaseRenderLayer {
         // Skip rendering if the character is invisible
         const isDefaultBackground = bg >= 256;
         const isInvisible = flags & FLAGS.INVISIBLE;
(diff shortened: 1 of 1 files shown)


src/renderer/TextRenderLayer.ts
```diff
@@ -124,7 +124,7 @@ export class TextRenderLayer extends BaseRenderLayer {
         if (flags & FLAGS.INVERSE) {
           const temp = bg;
           bg = fg;
-          fg = bg;
```
Tyriar: Oppsy 😄

src/renderer/TextRenderLayer.ts
```diff
@@ -94,7 +94,7 @@ export class TextRenderLayer extends BaseRenderLayer {
         // Skip rendering if the character is invisible
         const isDefaultBackground = bg >= 256;
         const isInvisible = flags & FLAGS.INVISIBLE;
-        if (!code || (code === 32 /*' '*/ && isDefaultBackground) || isInvisible) {
```
Tyriar: I think this optimization actually saves heaps of time, especially when the viewport is very wide as we ignore space cells very early in the process. Should this instead be:

if space char AND is default bg AND is not inverted?

mofux: I've tried that first, and it restores the white top bar in nano, but it fails when nano is closed:

```javascript
const isInverted = flags & FLAGS.INVERSE;
if (!code || (code === 32 /*' '*/ && isDefaultBackground && !isInverted) || isInvisible) {
    continue;
}
```
screen shot 2017-09-15 at 18 00 09

Tyriar: Oh, you need to pull this off the _state, ie. *was* it inverted. Similar to what's happening above with `(state[CHAR_DATA_ATTR_INDEX] & 0x1ff) >= 256`, this means "was default background color". I think you need to add it to the condition above:

```
// Clear the old character was not a space with the default background
if (state && !(state[CHAR_DATA_CODE_INDEX] === 32 /*' '*/ && (state[CHAR_DATA_ATTR_INDEX] & 0x1ff) >= 256) && <!wasInverted>) 
```

Pulling `state[CHAR_DATA_ATTR_INDEX]` out into an `oldAttr` var will help.

mofux: I think I've managed to fix it 😅

src/renderer/TextRenderLayer.ts
```diff
@@ -83,7 +83,8 @@ export class TextRenderLayer extends BaseRenderLayer {
         }
 
         // Clear the old character was not a space with the default background
-        if (state && !(state[CHAR_DATA_CODE_INDEX] === 32 /*' '*/ && (state[CHAR_DATA_ATTR_INDEX] & 0x1ff) >= 256)) {
+        const wasInverted = state && state[CHAR_DATA_ATTR_INDEX & FLAGS.INVERSE];
```
Tyriar: It should actually be:

```js
const wasInverted = state && ((state[CHAR_DATA_ATTR_INDEX] >> 18) & FLAGS.INVERSE);
```

`CHAR_DATA_ATTR_INDEX & FLAGS.INVERSE` is equivalent to `0 & 8` which equals 0, so currently it will return the character code.

Broken down this is:

```js
const attr = state[CHAR_DATA_ATTR_INDEX];
const flags = attr >> 18; // Bit shift right 18 bits
const wasInverted = !!(flags & FLAGS.INVERSE); // true if flags contains the bit for 8 (0b1000)
```

PR #997, opened 2017-09-15
Allow all overflowing characters to extend to the next cell if followed by a space (Emojis etc)

### LINKED ISSUE #974: Popular "wide" characters are getting cropped

mofux:
Fixes #974

This PR tries to provide a more generic way of detecting overlapping characters and allowing them to be drawn to the next cell if that cell is empty (space). The first time a character is seen, it will measure the width of that character and then check if the character is overlapping. It will then cache that result, so subsequent character appearances don't have to be measured again. The cache will be cleared if the font, or the measured character width changes.

### DIFF
--- a/src/renderer/TextRenderLayer.ts
+++ b/src/renderer/TextRenderLayer.ts
@@ -31,7 +31,7 @@ export class TextRenderLayer extends BaseRenderLayer {
 
   public resize(terminal: ITerminal, dim: IRenderDimensions, charSizeChanged: boolean): void {
     super.resize(terminal, dim, charSizeChanged);
-		
+
     // Clear the character width cache if the font or width has changed
     const terminalFont = `${terminal.options.fontSize * window.devicePixelRatio}px ${terminal.options.fontFamily}`;
     if (this._characterWidth !== dim.scaledCharWidth || this._characterFont !== terminalFont) {
@@ -123,7 +123,7 @@ export class TextRenderLayer extends BaseRenderLayer {
             width = 2;
             this._clearChar(x + 1, y);
             // The overlapping char's char data will force a clear and render when the
-            // overlapping char is no longer to the left of the character and also when 
+            // overlapping char is no longer to the left of the character and also when
             // the space changes to another character.
             this._state.cache[x + 1][y] = OVERLAP_OWNED_CHAR_DATA;
           }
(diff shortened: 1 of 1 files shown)


src/renderer/TextRenderLayer.ts
```diff
@@ -168,20 +179,37 @@ export class TextRenderLayer extends BaseRenderLayer {
...
-    }
-    // Check emoji unicode range
-    return char.search(/([\uD800-\uDBFF][\uDC00-\uDFFF])/g) >= 0;
-  }
+	
+	/**
+	 * Whether a character is overlapping to the
+	 * next cell.
+	 */
+	private _isOverlapping(char: string): boolean {
```
Tyriar: Let's also pass in char `code`, maybe the `CharData` object would be easier since we need 2 of its props?

src/renderer/TextRenderLayer.ts
```diff
@@ -168,20 +179,37 @@ export class TextRenderLayer extends BaseRenderLayer {
...
+
```
Tyriar: The common case:
```
if (code < 256) {
  return false;
}
```

It should be safe to assume this.

src/renderer/TextRenderLayer.ts
```diff
@@ -168,20 +179,37 @@ export class TextRenderLayer extends BaseRenderLayer {
...
+		// Measure the width of the character, but Math.floor it
+		// because that is what the renderer does when it calculates
+		// the character dimensions we are comparing against
+		const overlaps = Math.floor(this._ctx.measureText(char).width) > this._characterWidth;
+		
+		// Restore the original font to the context
+		if (font) this._ctx.font = font;
+		
+		// Cache and return
+		return this._characterOverlapCache[char] = overlaps;
```
Tyriar: Let's make this 2 lines, I always have to have a second look as doing this behaves very differently in different languages.

src/renderer/TextRenderLayer.ts
```diff
@@ -168,20 +179,37 @@ export class TextRenderLayer extends BaseRenderLayer {
...
+		let font;
+		if (this._ctx.font !== this._characterFont) {
+			font = this._ctx.font;
+			this._ctx.font = this._characterFont;
+		}
+		
+		// Measure the width of the character, but Math.floor it
+		// because that is what the renderer does when it calculates
+		// the character dimensions we are comparing against
+		const overlaps = Math.floor(this._ctx.measureText(char).width) > this._characterWidth;
```
Tyriar: This function seems a lot faster than I was expecting 😄

PR #1001, opened 2017-09-20
Fix mouse coordinates

### LINKED ISSUE #1000: Selection has an incorrect offset the further right the cursor goes
VS Code issue: https://github.com/Microsoft/vscode/issues/34184 (the horiziontal issue is an xterm.js problem, not the vertical issue)

Tyriar:
Fixes #1000


Tyriar:
Added additional changes to this PR that fixes mouse coords when zoom level is not 100%.

PR #1009, opened 2017-09-28
Test whether chars are emojis before measuring

### LINKED ISSUE #1008: Emojis can sometimes be classified as not overlapping
Not sure why this happens but we can work around the problem by having a more strict emoji test like before.

image

See https://github.com/Microsoft/vscode/issues/35102

Tyriar:
Fixes #1008


Tyriar:
This was caused by not being on latest v3, probably not worth trying to optimize this

PR #1023, opened 2017-10-02
Update README.md

ricktbaker:
Who would have thought that a name could have so much controversy.   I've renamed due to a bunch of flack about devops not logging into servers :)


Tyriar:
No worries, I like the new name more anyway :smile:

PR #1027, opened 2017-10-04
remove import file extension

jgillich:
Really no reason to have this, and it breaks the build when importing src/


Tyriar:
Yeah that should not be there, thanks @jgillich 😄

PR #1028, opened 2017-10-05
Normalize tsconfig outDir

### LINKED ISSUE #1024: Running the demo app fails on npm install
I have been trying to run the demo application locally, but have run into issues after running npm install because of the gulp build process. The error is related to this block of code:

```
// Under some environments like TravisCI, this comes out at absolute which can
// break the build. This ensures that the outDir is absolute.
if (outDir.indexOf(__dirname) !== 0) {
  outDir = `${__dirname}/${outDir}`;
}
```

The gulp build fails with a message such as:

Error: C:\Users\foobar\Downloads\xterm.js/C:/Users/foobar/Downloads/xterm.js/lib contains invalid WIN32 path characters.

I have commented this code out and everything works as expected, but I am not sure if I should PR this fix because according to the comment above, the code may break in other situations?

## Details
- Browser and browser version: -
- OS version: Windows 7
- xterm.js version: Cloned the project from the v3 branch

### Steps to reproduce

1. Follow the steps to run the demo app in the README

Tyriar:
It was coming out with the wrap path separators on Windows

Fixes #1024

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,8 +25,8 @@ let outDir = tsProject.config.compilerOptions.outDir;
 
 // Under some environments like TravisCI, this comes out at absolute which can
 // break the build. This ensures that the outDir is absolute.
-if (outDir.indexOf(__dirname) !== 0) {
-  outDir = `${__dirname}/${outDir}`;
+if (path.normalize(outDir).indexOf(__dirname) !== 0) {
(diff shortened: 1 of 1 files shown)


mofux:
LGTM 👍

Tyriar:
Test failure is wcwidth flakiness, merging

PR #1038, opened 2017-10-06
Rendering improvements

### LINKED ISSUE #1031: Background is not being drawn for second cell in overlapping char
VS Code issue: https://github.com/Microsoft/vscode/issues/35649

### LINKED ISSUE #1032: Sub-pixel anti-aliasing isn't working for uncached text
image

### LINKED ISSUE #1035: Wide characters are not cleared properly when removed from the screen
VS Code issue: https://github.com/Microsoft/vscode/issues/35660

### LINKED ISSUE #1036: Emojis can be cut off on Windows
VS Code issue: https://github.com/Microsoft/vscode/issues/35656

### LINKED ISSUE #1037: Right side of character is being cut off
Turns out this is really obvious with certain platform/font combinations.

VS Code issue: https://github.com/Microsoft/vscode/issues/35188

Tyriar:
Fixes #1032 
Fixes #1031
Fixes #1035
Fixes #1036
Fixes #1032
Fixes #1037

---

See individual commits for what was actually done. It turns out drawing the entire line like this is fairly fast, typically ranging from 0.5ms to 4ms on my macbook. The boost in stability/accuracy that this change brings is a must for now. It would be good to go back to intelligently diffing before drawing eventually (when we're certain the characters aren't overlapping).

### DIFF
--- a/src/renderer/ColorManager.ts
+++ b/src/renderer/ColorManager.ts
@@ -86,7 +86,7 @@ export class ColorManager implements IColorManager {
    */
   public setTheme(theme: ITheme): void {
     this.colors.foreground = theme.foreground || DEFAULT_FOREGROUND;
-    this.colors.background = theme.background || DEFAULT_BACKGROUND;
+    this.colors.background = this._validateColor(theme.background, DEFAULT_BACKGROUND);
     this.colors.cursor = theme.cursor || DEFAULT_CURSOR;
     this.colors.cursorAccent = theme.cursorAccent || DEFAULT_CURSOR_ACCENT;
     this.colors.selection = theme.selection || DEFAULT_SELECTION;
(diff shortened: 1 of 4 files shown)


src/renderer/ColorManager.ts
```diff
@@ -107,4 +107,17 @@ export class ColorManager implements IColorManager {
     this.colors.ansi[14] = theme.brightCyan || DEFAULT_ANSI_COLORS[14];
     this.colors.ansi[15] = theme.brightWhite || DEFAULT_ANSI_COLORS[15];
   }
+
+  private _validateColor(color: string, fallback: string): string {
```
mofux: Doesn't this mean that theme.background can only be a hex color, not rgb(a) - because otherwise it would just use the fallback color?

Tyriar: It does for background, validate could be changed to accommodate rgb/rgba/etc. though.

This was needed to be able to do this:
 
```
 +    const r = parseInt(background.substr(1, 2), 16);
 +    const g = parseInt(background.substr(3, 2), 16);
 +    const b = parseInt(background.substr(5, 2), 16);
```

PR #1040, opened 2017-10-09
Added contenteditable attribute to terminal to support paste in Firefox

joegaffey:


joegaffey:
Demonstration
https://joegaffey-xtermjs.glitch.me/

joegaffey:
This fix address https://github.com/sourcelair/xterm.js/issues/1039

mofux:
@joegaffey Thanks for this PR. 

It feels a little bit like a hack, but it seems to work on Chrome and Firefox without an issue. I think the real problem that causes this bug is that if one clicks on the terminal, the "invisible" textarea is losing focus for a brief moment until the `mousedown` handler on the terminal gives the focus back to the textarea. In Firefox, the check if it can paste seems to happen in between this short timeframe while the textarea is still unfocused. Having the terminal element `contenteditable` hence gives the terminal DOM element focus for that brief moment, allowing Firefox to think that it can paste.

I would really like to tackle this problem some other way, but I have no idea how, so your approach seems to be a good solution for now. Maybe you could add a comment above the line of code to explain why we do it, and then I wouldn't be objected to merge this PR. I'm unable to test it on IE though, maybe someone with IE can confirm it doesn't break there?

@Tyriar @parisk  Any thoughts?

joegaffey:
Tested in IE11. Seems to work ok.
ie11

joegaffey:
Added code comment on addition.

joegaffey:
Hi @mofux. There were some build errors after I added the code comment. I think these were environment related. Do I need to rebuild? If so, how?

mofux:
@joegaffey One error was somehow environment related, but the linting tests are failing because of a trailing whitespace character in the comment line:

src/Terminal.ts[582, 1]: trailing whitespace

joegaffey:
Thanks @mofux ! That's fixed now.

mofux:
Thanks! I'd like to wait for @Tyriar approval before we can get this merged. He's busy traveling these days, so let's have some patience ☺️

Tyriar:
I'm a little concerned that this will introduce some other weird bug. It might be good initially to indentify why moving the textarea under the cursor to trigger the context menu does not work for Firefox, if that problem cannot be resolved then perhaps we could detect it before adding `contenteditable`?

This is the code that does the move for the context menu: https://github.com/sourcelair/xterm.js/blob/b06eb44c4552b46d1af3db4a0ff59d2ef1bfdc5f/src/handlers/Clipboard.ts#L108

joegaffey:
Struggling to dig further into this at the moment due to a node-gyp install issue on Windows 10. 
Have the demo running on Ubuntu in a VM but paste is working fine there in Firefox (55) without setting contenteditable.

mofux:
Just tested the v3 branch on Firefox on Mac, copy & paste works flawlessly without this workaround. Is this only a Windows issue? 

@joegaffey Have you tried if the v3 branch exposes the same problem in Firefox for you?

joegaffey:
@mofux Yes, the issue was on the v3 branch. Seems to be a Windows only issue.

parisk:
Should we move forward with this?

joegaffey:
@parisk It would be appreciated. We may have to run the fork in production otherwise.

Tyriar:
I think it would be better to understand why this is happening on Firefox and report to Firefox if it's an issue there, rather than pushing in a weird edge case that could very well introduce more issues (now or in the future).

vincentwoo:
I second that idea - contenteditable is one of the most finicky things out there.

joegaffey:
I agree it would be best to get to the root cause of this.
However, in the meantime, would it make sense to include the Firefox fix in the Firefox specific code section of this file?
It could be further limited to Windows only execution thus reducing the scope of impact to the environment where the bug is occurring.
I can try this when I get back in front of my PC in a few days.

Tyriar:
> However, in the meantime, would it make sense to include the Firefox fix in the Firefox specific code section of this file?

What's more likely to happen is that this workaround will remain there forever and cause issues down the line as no one will bother following up as it's no longer a user facing issue (yet). Adding user agent-specific workarounds is really ugly and we should only resort to it when we absolutely must.

parisk:
OK, I moved on with testing things out on `v3` and Firefox (latest) on Windows 10.

Right-click paste worked fine for me (see attached screenshot), so this does not look like a platform issue.

@joegaffey can you please try testing this in a Private Window (without add-ons) or on a clean system?

I think there is nothing we can do on our side for this, so I will move one with closing this PR, but feel free to continue the conversation here.

## Screenshot
image

joegaffey:
I just tested the link in your screenshot in Firefox ESR, Stable and Developer Edition (all latest) in private mode on Windows 7 64-bit.

I am still seeing this issue in all tested Firefox versions. 

See my screenshots below.

**ESR**
image

**Stable**
image

**Developer Edition**
image

joegaffey:
Same deal in Firefox Stable latest and Developer Edition latest on Windows 10 Home 64-bit.

**Stable** 
image

**Developer Edition**
image

runner-mei:
why close the pull rquest?  This bug is happen on my firefox57/windows 
xterm

joegaffey:
@runner-mei I am also still experiencing this issue. While I understand the concern raised by the maintainers regarding avoid potential issues in the future, I would prefer to fix a known bug.
There is already a Firefox workaround in the file which uses browser sniffing to only execute in Firefox. This fix could be moved to that code section so that it could not affect other browsers.

PR #1041, opened 2017-10-09
Cleanup renderer code

mofux:
Since we are now always redrawing the whole line if it was changed, we can remove all the overlapping code that was put in for the optimisations. This yields another very nice effect, because overlapping characters can now overlap to the next cell without us having to worry about a space character after them.

render


Tyriar:
I think we should hang onto this code for the time being, I'd like to reintroduce partial line clearing/redrawing by looking for *words* that are dirty. Will put a write up of the idea in an issue soon.

mofux:
👍

PR #1043, opened 2017-10-09
Add support for dimmed characters (faint)

### LINKED ISSUE #1042: Emulation: bold off broken and dim not implemented?
Shutting off (20 + attribute) **bold** is not working, although others implemented (uline and reverse) do, i.e.,
```
echo -e '\x1B[33m yellow \x1B[1m bold \x1B[21m normal \x1B[m off'
echo -e '\x1B[33m yellow \x1B[4m uline \x1B[24m normal \x1B[m off'
echo -e '\x1B[33m yellow \x1B[7m reverse \x1B[27m normal \x1B[m off'
````

Oddly enough, changing **21** (bold off) to **22** (dim off) works:
`echo -e '\x1B[33m yellow \x1B[1m bold \x1B[22m normal \x1B[m off'`

... but bold -> dim _enable_ keeps bold _on_ -- I thought bold/dim are implemented as mutually exclusive, unlike underline, blink, and reverse that can be added or taken away with either bold or dim?  At the very least if dim is not implemented, it should shutoff bold, but it does not:
`echo -e '\x1B[33m yellow \x1B[1m bold \x1B[2m dim \x1B[m off'`

Two related questions:
1. I get **blink** (at least for now ;) not implemented, but how about **dim** and _italic_?
2. Can a custom color palette be constructed?  The one in v3 branch (below) is more palatable to me.

## Details
- Browser and browser version: Firefox 56 (64-bit)
- OS version: Fedora 26 (linux)
- xterm.js version: 2.9.2 and latest v3 branch

### Steps to reproduce
```
$ cat bin/ansi-test.sh
#!/bin/sh

alias print='echo -en'
alias println='echo -e'
csi="\\x1B["
color=( black red green yellow blue magenta cyan white )
attr=( reset bright dim italic uline blink 6 reverse )

for bg in `seq 0 7`; do
        print "(${csi}m${bg}) BG ${csi}0;1;$((30+${bg}))m ${color[${bg}]}${csi}m"
        for fg in `seq 0 7`; do
                print "\n    FG ${csi}0;1;$((30+${fg}))m ${color[${fg}]}${csi}m\t"
                print "${csi}$((40+${bg}))m"
                print "${csi}$((30+${fg}))m"
                for ca in `seq 1 7`; do
                        print "${csi}${ca}m  ${attr[${ca}]}  ${csi}$((20+${ca}))m"
                done
                print "${csi}m"
        done
        println "${csi}m"
done
```
**Showing Terminology - xterm.js - v3 branch:**
terminology xterm v3

### PR DESCRIPTION
**Pull request by mofux**
Fixes #1042 

This PR adds support to render dimmed characters.

screen shot 2017-10-09 at 17 13 23

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -1113,7 +1113,6 @@ export class InputHandler implements IInputHandler {
    *     Ps = 5  -> Blink (appears as Bold).
    *     Ps = 7  -> Inverse.
    *     Ps = 8  -> Invisible, i.e., hidden (VT300).
-   *     Ps = 2 1  -> Bold off or Double Underline.
    *     Ps = 2 2  -> Normal (neither bold nor faint).
    *     Ps = 2 4  -> Not underlined.
    *     Ps = 2 5  -> Steady (not blinking).
@@ -1226,9 +1225,6 @@ export class InputHandler implements IInputHandler {
       } else if (p === 2) {
         // dimmed text
         flags |= FLAGS.DIM;
-      } else if (p === 21) {
-        // bold off not widely supported; double underline hardly ever supported.
-        flags &= ~FLAGS.BOLD;
       } else if (p === 22) {
         // not bold nor faint
         flags &= ~FLAGS.BOLD;
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -1107,7 +1107,7 @@ export class InputHandler implements IInputHandler {
    * CSI Pm m  Character Attributes (SGR).
    *     Ps = 0  -> Normal (default).
    *     Ps = 1  -> Bold.
-	 *     Ps = 2  -> Faint, decreased intensity (ISO 6429).
+   *     Ps = 2  -> Faint, decreased intensity (ISO 6429).
    *     Ps = 4  -> Underlined.
    *     Ps = 5  -> Blink (appears as Bold).
    *     Ps = 7  -> Inverse.
@@ -1221,10 +1221,10 @@ export class InputHandler implements IInputHandler {
       } else if (p === 8) {
         // invisible
         flags |= 16;
-			} else if (p === 2) {
-				// dimmed text
-				flags |= 32;
-			} else if (p === 22) {
+      } else if (p === 2) {
+        // dimmed text
+        flags |= 32;
+      } else if (p === 22) {
         // not bold
         flags &= ~1;
       } else if (p === 24) {
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -1113,6 +1113,7 @@ export class InputHandler implements IInputHandler {
    *     Ps = 5  -> Blink (appears as Bold).
    *     Ps = 7  -> Inverse.
    *     Ps = 8  -> Invisible, i.e., hidden (VT300).
+   *     Ps = 2 1  -> 	Bold off or Double Underline.
    *     Ps = 2 2  -> Normal (neither bold nor faint).
    *     Ps = 2 4  -> Not underlined.
    *     Ps = 2 5  -> Steady (not blinking).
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/InputHandler.ts [hunk 1]
```diff
@@ -1220,6 +1221,9 @@ export class InputHandler implements IInputHandler {
       } else if (p === 8) {
         // invisible
         flags |= 16;
+      } else if (p === 2) {
+        // dimmed text
+        flags |= 32;
```
Tyriar: Let's import FLAGS from Types.ts here

src/renderer/BaseRenderLayer.ts
```diff
@@ -241,11 +241,15 @@ export abstract class BaseRenderLayer implements IRenderLayer {
       // ImageBitmap's draw about twice as fast as from a canvas
       const charAtlasCellWidth = this._scaledCharWidth + CHAR_ATLAS_CELL_SPACING;
       const charAtlasCellHeight = this._scaledCharHeight + CHAR_ATLAS_CELL_SPACING;
+      // Apply alpha to dim the character
+      if (dim) {
+        this._ctx.globalAlpha = 0.5;
```
Tyriar: Let's make the 0.5 a constant up the top. How does 0.5 compare to other terminals that implement this?

src/InputHandler.ts
mofux: Done. I have also replaced the other numeric references with the according flag.

src/renderer/BaseRenderLayer.ts
mofux: 0.5 was the value that matched the dimmed characters from `Terminal.app` best. I have made this a constant now

theflyingape:
I see inclusion for (p === **22**) to cancel both bold + dim, perfect, thank you.
While "not widely supported", there are legacy apps that can send a cancel bold only SGR, could this `else if(p === 21) flags &= ~FLAGS.BOLD;` be added?
As noted here: [Select Graphic Rendition (SGR)](https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_.28Select_Graphic_Rendition.29_parameters)

mofux:
@theflyingape Thanks for your feedback! I've added SGR 21 to clear bold.

Tyriar:
@theflyingape normally we use http://invisible-island.net/xterm/ctlseqs/ctlseqs.html as the "spec" as it's maintained by a developer of xterm. Do several popular terminal emulators implement this clear bold?

mofux:
`Terminal.app` and `iTerm` do not I'm afraid (they don't seem to implement it at all, neither double underline nor cancel bold):

```bash
echo -e '\x1B[33m yellow \x1B[1m bo\x1B[21mld \x1B[2m dim \x1B[m off'
```

screen shot 2017-10-10 at 13 34 19

Not sure about other Terminal apps though...

@Tyriar Shall I remove the support for it (at least in this PR, so we can have a separate discussion about this feature in a separate issue)?

theflyingape:
@Tyriar, Linux-based X terminals including its physical VGA console indeed consume a clear bold sequence (<csi>21m) and turn only that character attribute off; and Windows commercial terminal emulators, sigh, like what we still use at our hospital for many classic apps that transmit bold on / bold off in this manner.

Tyriar:
I see gnome-terminal supports it but actual xterm does not:

screen shot 2017-10-10 at 7 41 25 am

Split it into a separate issue might be good so we can merge this and have a discussion.

theflyingape:
Right, it's not widely implemented and it's mostly ignored.  Thanks for the work on dim and please do split cancel bold into a separate non-blocking issue; I do not see any downside for adding it in.  For reference, **hterm** also [lists it](https://chromium.googlesource.com/apps/libapps/+/master/hterm/doc/ControlSequences.md#SGR) as ignored but TBD.

mofux:
I have reverted the SGR 21 support, let's track it in a separate issue. PR should be good to merge now.

PR #1045, opened 2017-10-10
More rendering improvements

### LINKED ISSUE #1044: Support letter spacing
Related: https://github.com/Microsoft/vscode/issues/35695

Tyriar:
- Adds `letterSpacing` option
- Rows are now clipped for uncached draws using the *cell* height, not the *char* height. This means that larger characters when not drawn from the char atlas will consume the additional space provided by the `lineHeight`

Fixes #1044


mofux:
@Tyriar Does this fix #1044 by manually setting the new `letterSpacing` option?

Tyriar:
It fixes #1044 as that's just a feature for letterSpacing, it provides a workaround for https://github.com/Microsoft/vscode/issues/35695 for the time being. The reason they don't use "standard spacing" is so they align with the grid such that the charatlas can be utilized. Using the charatlas on floating point numbers looks OK, but not perfect.

Tyriar:
Ping

mofux:
LGTM 👍

PR #1051, opened 2017-10-11
Draw from charatlas without changing scale

### LINKED ISSUE #1046: Characters are condensed
Since #1038 I am noticing that characters aren't as sharp as before. They also look not wide enough. Here is a direct comparision:

(expected) before #1038 
screen shot 2017-10-10 at 20 47 43

(current) since #1038 
screen shot 2017-10-10 at 20 49 17

I think only cached characters have this problem.

### LINKED ISSUE #1050: Characters are blurry on low dpi screens
VS Code issue https://github.com/Microsoft/vscode/issues/35991

May be related to https://github.com/sourcelair/xterm.js/issues/1046

Tyriar:
Fixes #1050 
Fixes #1046

---

Before
screen shot 2017-10-10 at 10 11 51 pm

After
screen shot 2017-10-10 at 10 13 54 pm


mofux:
Great finding, seems to also fix #1046, added it to the PR description 👍

PR #1064, opened 2017-10-15
Apple iPad - Smart Keyboard Arrows

exsilium:
Hi! It was a pleasant surprise to find out that xterm.js mostly works out of the box with Apple's Smart Keyboard using iPad iOS 11. However, the arrows don't work. Let me know if the main code is not a place for such platform specific mappings. 🍻 

- Arrows trigger `keydown` event
- `keyCode == 0`, always.
- The actual key pressed is identified via `key` strings:
    - `UIKeyInputUpArrow`
    - `UIKeyInputLeftArrow`
    - `UIKeyInputRightArrow`
    - `UIKeyInputDownArrow`

I'm not sure how Android does it and if this is truly iPad specific or not. I'm also not sure if this truly should be part of the main xterm code, or just handled separately via addon/application layer when useragent is identified as iPad.

### DIFF
--- a/src/xterm.js
+++ b/src/xterm.js
@@ -1511,6 +1511,36 @@ Terminal.prototype.evaluateKeyEscapeSequence = function(ev) {
   };
   var modifiers = ev.shiftKey << 0 | ev.altKey << 1 | ev.ctrlKey << 2 | ev.metaKey << 3;
   switch (ev.keyCode) {
+    case 0:
+      if (ev.key === "UIKeyInputUpArrow") {
+        if (this.applicationCursor) {
+          result.key = C0.ESC + 'OA';
+        } else {
+          result.key = C0.ESC + '[A';
+        }
+      }
+      else if (ev.key === "UIKeyInputLeftArrow") {
+        if (this.applicationCursor) {
+          result.key = C0.ESC + 'OD';
+        } else {
+          result.key = C0.ESC + '[D';
+        }
+      }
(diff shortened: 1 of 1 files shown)


parisk:
@exsilium can you please rebase the PR to use [`v3`](https://github.com/sourcelair/xterm.js/tree/v3) as the base branch? This is the current branch where we introduce new functionality.

Then I will give it a go on my device.

Also, do you have any public documentation available for these key names in Safari?

exsilium:
@parisk, thanks, will do! In fact, I did try v3 to begin with, but there seems that there are some interface changes comparing with v2 as it broke my integration with cloud9. I need to [look into that first](https://github.com/exsilium/cloud9/blob/development/plugins-client/ext.terminal/tty.js#L39).

As for the documentation - I have yet to find the official source. I was looking via console to see what (if any) keydown/keypress event is triggered by the arrows. I also noticed that the key strings passed match with the ones defined in [UIKit](https://developer.apple.com/documentation/uikit/uikeycommand/input_strings_for_special_keys) documentation, except they also define one string for escape key, a key which is lacking on the smart keyboard 😲 (maybe for the future?)

For a full event example being sent by the arrows, see [this comment](https://github.com/exsilium/cloud9/issues/33#issuecomment-336733830)

exsilium:
I resubmitted this against v3 as #1065 , feel free to close this one, as v2 is feature-locked.

PR #1065, opened 2017-10-16
Apple iPad - Smart Keyboard Arrows - v3

exsilium:
A duplicate of #1064 against the v3 branch. Please test and provide feedback.

- Arrows trigger `keydown` event
- `keyCode == 0`, always.
- The actual key pressed is identified via `key` strings:
    - `UIKeyInputUpArrow`
    - `UIKeyInputLeftArrow`
    - `UIKeyInputRightArrow`
    - `UIKeyInputDownArrow`

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -1415,6 +1415,36 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     };
     const modifiers = (ev.shiftKey ? 1 : 0) | (ev.altKey ? 2 : 0) | (ev.ctrlKey ? 4 : 0) | (ev.metaKey ? 8 : 0);
     switch (ev.keyCode) {
+      case 0:
+        if (ev.key === "UIKeyInputUpArrow") {
+          if (this.applicationCursor) {
+            result.key = C0.ESC + 'OA';
+          } else {
+            result.key = C0.ESC + '[A';
+          }
+        }
+        else if (ev.key === "UIKeyInputLeftArrow") {
+          if (this.applicationCursor) {
+            result.key = C0.ESC + 'OD';
+          } else {
+            result.key = C0.ESC + '[D';
+          }
+        }
+        else if (ev.key === "UIKeyInputRightArrow") {
+          if (this.applicationCursor) {
+            result.key = C0.ESC + 'OC';
+          } else {
+            result.key = C0.ESC + '[C';
+          }
+        }
+        else if (ev.key === "UIKeyInputDownArrow") {
+          if (this.applicationCursor) {
+            result.key = C0.ESC + 'OB';
+          } else {
+            result.key = C0.ESC + '[B';
+          }
+        }
+        break;
       case 8:
         // backspace
         if (ev.shiftKey) {
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -1416,28 +1416,28 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     const modifiers = (ev.shiftKey ? 1 : 0) | (ev.altKey ? 2 : 0) | (ev.ctrlKey ? 4 : 0) | (ev.metaKey ? 8 : 0);
     switch (ev.keyCode) {
       case 0:
-        if (ev.key === "UIKeyInputUpArrow") {
+        if (ev.key === 'UIKeyInputUpArrow') {
           if (this.applicationCursor) {
             result.key = C0.ESC + 'OA';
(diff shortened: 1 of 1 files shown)


parisk:
@exsilium code looks OK, but I can't get this working on my iPad Pro, with the smart keyboard on.

Do you have any debugging steps that I have to take to see what might be going wrong?

parisk:
btw, I did the following quick hack to see what might be going wrong and seems like it never gets into the `keyCode == 0` case:

```javascript
    switch (ev.keyCode) {
      case 0:
        alert('Zero keyCode!')
        alert(ev.key);
        if (ev.key === 'UIKeyInputUpArrow') {
          if (this.applicationCursor) {
            result.key = C0.ESC + 'OA';
          } else {
            result.key = C0.ESC + '[A';
          }
        }
```

exsilium:
Sad to hear.. do you get *any* keydown event triggered when using the arrows? e.g. pull those alert()'s before the keyCode check happens?

parisk:
I am on iOS 11.0.3 as well, using Safari (iPad Pro 10.9.

Seems like the arrows do not fire any of the keydown, keypress, keyup events, but they work fine on both ProseMirror (http://prosemirror.net/) and CodeMirror (http://codemirror.net/).

Both of these libraries though do not seem to handle the above events (e.g. `UIKeyInputUpArrow`) anywhere in their codebases so, is there any chance we have to handle this with a different event type?

exsilium:
I'm on 12.9 pro (2017) also on Safari, so it's unclear why I'm getting those events triggered - is it possible, that the bigger Smart Keyboard works differently? 😕 . The text editor examples rely on composition change to detect the arrow keys. Here's [how ace does it](https://github.com/ajaxorg/ace/blob/eb2437badf4013893e15cb873057607c09cc95ae/lib/ace/keyboard/textinput_ios.js#L482).

exsilium:
@parisk - a simple test, does your iPad keyboard give any keycode entry when you press an arrow key at this site: http://keycode.info

exsilium:
This seems to be specific to the demo app, I'm unable to get those events triggered while using the xterm.js bundled demo. However, it works when I'm using the xterm within https://github.com/exsilium/cloud9/tree/development (feel free to try this yourself)

parisk:
OK, I think I found the root cause of this.

For some reason it seems that the smart keyboard events are being fired only when listening to events on `document.body`.

Can you give it a try here: http://qizvas6t.apps.lair.io?

If this is the case, we should reconsider the way we are listening to events (at least on mobile devices).

Tyriar:
Will this issue go away when we move to listening to the `input` event? We need to do this anyway to cover other IMEs such as the emoji picker on Windows and Mac.

exsilium:
@parisk I might have been late to the party, the lair link states that the application "**has an error and cannot start**" :(

parisk:
@exsilium sorry about that, we shutdown projects after a few hours of inactivity 😅.

@Tyriar not sure. Only if this works out of the box, because the [`input` event](https://developer.mozilla.org/en-US/docs/Web/Events/input) does not seem to provide any information about the `key` pressed.

We could use a mix of all different events and we could try listening on `document.body` and making sure that the event target/source is the terminal before we act.

exsilium:
@parisk ok, let me know if it ever comes back. i'll try to be quicker about it to test it :) Cheers! 🍻

parisk:
@exsilium if you are in the mood of testing, please drop me a ping at https://gitter.im/sourcelair/xterm.js to put it online. Maybe we will be able to hash out a solution in a few minutes also 😄 .

parisk:
@exsilium I propose the following:

1. We detect the platform via [`src/utils/Browser.ts`](https://github.com/sourcelair/xterm.js/blob/v3/src/utils/Browser.ts)
2. If the platform is `iPad`, we listen to the `keydown` event
3. If `keyCode` is equal to `0` we forward the event to [`_keyDown`](https://github.com/sourcelair/xterm.js/blob/v3/src/Terminal.ts#L1355)

How does that sound?

exsilium:
Hi @parisk ! This sounds good 👍 Will you be making the needed changes and get this merged?

parisk:
@exsilium sure.

parisk:
Well, my plan didn't work at all 😅. I spent the last days trying to debug this but had no luck with it.

Let's merge this as-is to solve at least some cases and have a guide for the future.

I also created an issue to follow-up on iOS keyboard support: https://github.com/sourcelair/xterm.js/issues/1101

PR #1069, opened 2017-10-19
Use .xterm over .terminal in CSS

### LINKED ISSUE #1068: .terminal selector in xterm.css is too generic
I've wanted to change this for a long time, it's causing issues in VS Code now as there are multiple elements that use a `terminal` class as it's the name of the panel.

Tyriar:
Fixes #1068

### DIFF
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -35,7 +35,7 @@
  *  Default styles for xterm.js
  */
 
-.terminal {
+.xterm {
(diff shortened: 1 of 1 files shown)


mofux:
LGTM 👍 Also a perfect candidate for the 3.0 release because it's kind of a breaking change for those with custom css.

Tyriar:
I'll add breaking change label to the issue so we mention it in the release notes

PR #1073, opened 2017-10-21
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that xterm.js gets 874k hits per month on jsDelivr, so I thought you might like this badge.


parisk:
Of course we do! That's awesome. Thanks 😄.

PR #1075, opened 2017-10-23
`window` not defined error

veeramarni:
When trying to use with Server Side rendering, we hitting with `window not defined error`. This PR will fix the problem.


parisk:
@veeramarni can you please re-open this PR against the [`v3`](https://github.com/sourcelair/xterm.js/tree/v3) branch?

I overlooked it and merged it into master, which we do not update until we merge [`v3`](https://github.com/sourcelair/xterm.js/tree/v3) back.

PR #1076, opened 2017-10-23
Zmodem addon

FGasper:
Hi all,

Please look this over, and let me know what you think. If you’re happy with it I’ll squash and make another PR.

Thanks!


package.json
```diff
@@ -60,13 +60,15 @@
     "jsdoc": "3.4.3",
     "jsdom": "^11.1.0",
     "merge-stream": "^1.0.1",
-    "node-pty": "^0.4.1",
+    "node-pty": "0.6.4",
```
Tyriar: We cannot change this currently as it will cause tests to break.

demo/main.js
```diff
@@ -109,14 +111,231 @@ function createTerminal() {
         socket.onopen = runRealTerminal;
         socket.onclose = runFakeTerminal;
         socket.onerror = runFakeTerminal;
+
+        term.zmodemAttach(socket);
```
Tyriar: I'm concerned about:

1. Why is so much code in the demo? Shouldn't this be in the addon?
2. This functionality being on the demo at all, the demo needs to be kept slim as it's what development/benchmarking is done against.

demo/main.js
```diff
@@ -109,14 +111,231 @@ function createTerminal() {
...
+function _update_progress(xfer) {
+    var total_in = xfer.get_offset();
+
+    document.getElementById("bytes_received").textContent = total_in;
+
+    var percent_received = 100 * total_in / xfer.get_details().size;
+    document.getElementById("percent_received").textContent = percent_received.toFixed(2);
+}
+
+function _hide_progress() {
```
Tyriar: Try follow existing styles, this should be:

```js
function hideProgress
```

Or even better:

```js
function zmodemHideProgress
```

demo/js/zmodem.devel.js
```diff
@@ -0,0 +1 @@
+../../node_modules/zmodem.js/dist/zmodem.devel.js
```
Tyriar: This dependency should be linked to from within the addon, the search addon provides an example of how to do this. Search is written in TS which is ideally what addons going forward should be written in, you should be able to use `require` if you're not familiar with TS though.

demo/app.js
```diff
@@ -21,10 +21,22 @@ app.get('/main.js', function(req, res){
   res.sendFile(__dirname + '/main.js');
 });
 
+app.get('/js/*.js', function(req, res){
+    res.sendFile(__dirname + req.url);
+});
+
+const SHELL_ARGS = {
+    bash: ['--login'],
```
Tyriar: Why is this needed?

FGasper: It’s not, but starting bash as a login shell made it more colorful. I’ll revert it here. Is it a change you’d be interested in if submitted separately?

package.json
FGasper: Do you know of any way to get node-pty 0.4.1 to send binary?

demo/app.js
Tyriar: If you're on a mac and the user agent is checked for that, sure we could add it. On Linux it should be running in a login session already.

package.json
Tyriar: Unfortunately support was only added for that recently in node-pty. The real fix it to resolve the issue with the tests, I'm hesitant to make the `.native` call that it's complaining about public API in node-pty though.

FGasper:
OK, I think I’ve addressed the points raised:

- The zmodem.js base library is now included directly in the xterm.js addon.
- Tests are passing now (using the latest node-pty).
- Extra changes (e.g., use of the `-l` flag to bash) are removed.
- The UI changes in the demo from upstream are now the minimum required for the addon to work. Note that this does have the disadvantage that anything that looks like a ZMODEM initialization string will trigger the addon, and the user has no control to prevent that.

demo/app.js
```diff
@@ -25,6 +25,7 @@ app.post('/terminals', function (req, res) {
   var cols = parseInt(req.query.cols),
       rows = parseInt(req.query.rows),
       term = pty.spawn(process.platform === 'win32' ? 'cmd.exe' : 'bash', [], {
+        encoding: null,
```
Tyriar: I'd still prefer if this was more separate from the standard demo, this line and the ones in index.html/main.js concern me - they could introduce issues while developing.

@parisk Maybe we should pull out more specialized addon demos to live somewhere else?

parisk: @Tyriar yep.

We are already doing this in [`src/addons/attach/index.html`](https://github.com/sourcelair/xterm.js/blob/v3/src/addons/attach/index.html).

I agree with @Tyriar on this.

FGasper: Sounds good … I’ll create a separate demo for the zmodem addon.

Tyriar: @parisk oh cool, didn't know about that :smile:

FGasper:
OK. The zmodem addon’s demo is now separate from the main demo.

Anything else?

src/addons/zmodem/demo/index.html
```diff
@@ -0,0 +1,127 @@
...
+        <title>xterm.js demo</title>
+        <link rel="stylesheet" href="/build/xterm.css" />
+        <link rel="stylesheet" href="/build/addons/fullscreen/fullscreen.css" />
+        <link rel="stylesheet" href="style.css" />
+      	<script src="https://cdnjs.cloudflare.com/ajax/libs/es6-promise/4.1.1/es6-promise.auto.min.js"></script>
+      	<script src="https://cdnjs.cloudflare.com/ajax/libs/fetch/1.0.0/fetch.min.js"></script>
+        <script src="/build/xterm.js" ></script>
+
+        <script src="/build/addons/attach/attach.js" ></script>
+        <script src="/build/addons/zmodem/zmodem.js" ></script>
```
Tyriar: Could we link to `./node_modules/zmodem.js/dist/zmodem.js` directly here? This would avoid us bundling anything for zmodem, most websites will have their own bundlers anyway so we shouldn't need to get into this.

src/addons/zmodem/demo/style.css
```diff
@@ -0,0 +1 @@
+../../../../demo/style.css
```
Tyriar: Just linking to the this stylesheet from the html page will save us creating this file

src/addons/zmodem/demo/index.html
FGasper: Done.

src/addons/zmodem/demo/style.css
FGasper: Done.

FGasper:
@parisk asked:

> Should we handle the attach and terminado add-on deactivation via the zmodem add-on?

That seemed to me a bit less than ideal in terms of separation of concerns.

parisk:
Yep, looks great, thanks @FGasper!

PR #1083, opened 2017-10-25
Made SelectionManager use the terminal's buffer so it's buffer can not become stale

agronick:
Fixes issue with copy in VIM and other programs not working because the buffer was not updated. Instead of SelectionManager using it's own buffer, it works with what is being displayed on the screen. 

Fixes /Microsoft/vscode/issues/25000

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -1122,7 +1122,6 @@ export class InputHandler implements IInputHandler {
           // if (params[0] === 1049) {
           //   this.restoreCursor(params);
           // }
-          this._terminal.selectionManager.setBuffer(this._terminal.buffer.lines);
           this._terminal.refresh(0, this._terminal.rows - 1);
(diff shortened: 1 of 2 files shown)


Tyriar:
We're no longer accepting PRs to the master branch (v3 is the current main dev branch), @agronick I didn't think this reproduced in VS Code anymore? I recall fixing this in the v3 branch quite a while ago, are you on VS Code 1.17.2?

agronick:
I've integrated the master branch into my own project. I have an SSH session connected directly to the websocket so all escape codes are going through. This is the issue. Open up a file in VIM. Copy some stuff. When VIM starts the selection manager is using the old buffer and you'll be copying what was on the screen before you started VIM.

I can try the new stuff. I did try the canvas based version and had too many issues to use in prod.

Tyriar:
Closing as this is fixed in the v3 branch, we should release soon.

PR #1088, opened 2017-10-27
Upgrade node-pty to 0.7.2

Tyriar:
I exposed the missing API again in node-pty so that we can upgrade, this greatly improves Windows support in the demo and also supports binary output. See https://github.com/Tyriar/node-pty/issues/153

/cc @FGasper


FGasper:
Cool! :-)

PR #1094, opened 2017-10-29
Clarify that parent needs to be visible when calling open

### LINKED ISSUE #1089: setOption throws DOMException when setting theme
I'm using the following code to initialize xterm.js:
```
let term = new Terminal();
term.open(document.getElementById('xterm-system'));
term.setOption('fontFamily', 'Roboto Mono');
term.setOption('theme', LIGHT_THEME);
```

It throws the following exception when being run:
```
DOMException: Failed to execute 'getImageData' on 'CanvasRenderingContext2D': The source width is 0.
    at xterm/lib/renderer/CharAtlas.js.CharAtlasGenerator.generate
    at Object.acquireCharAtlas
    at xterm/lib/renderer/BaseRenderLayer.js.BaseRenderLayer._refreshCharAtlas 
    at xterm/lib/renderer/BaseRenderLayer.js.BaseRenderLayer.onThemeChanged 
    at Array.forEach (<anonymous>)
    at xterm/lib/renderer/Renderer.js.Renderer.setTheme
    at xterm/lib/Terminal.js.Terminal._setTheme
    at xterm/lib/Terminal.js.Terminal.setOption
```

I picked the relevant line from `CharAtlasGenerator.generate`:
https://github.com/sourcelair/xterm.js/blob/v3/src/renderer/CharAtlas.ts#L199
```
 var charAtlasImageData = this._ctx.getImageData(0, 0, this._canvas.width, this._canvas.height);
```
where `this.canvas.width` and `this._canvas.height` are zero.

This is because `scaledCharWidth` and `scaledCharHeight` have the value `undefined` (see https://github.com/sourcelair/xterm.js/blob/v3/src/renderer/CharAtlas.ts#L130)

Chrome 62, Linux
XTerm.js v3 branch, commit `1d76cc921a53f16a0131b99a7a4dd407e343bfbc`

Tyriar:
Fixes #1089

---

Related https://github.com/xtermjs/xtermjs.org/pull/33

### DIFF
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -290,7 +290,9 @@ declare module 'xterm' {
 
     /**
      * Opens the terminal within an element.
-     * @param parent The element to create the terminal within.
+     * @param parent The element to create the terminal within. This element
+     * must be visible (have dimensions) when `open` is called as several DOM-
+     * based measurements need to be performed when this function is called.
      */
     open(parent: HTMLElement): void;
 


Tyriar:
We could check if the dimensions come out as 0, this is already being done and drawing is just prevented in these cases currently.

PR #1105, opened 2017-11-05
Export interfaces in typings

### LINKED ISSUE #1080: Export interfaces in Typescript typings
I am using xterm.js from a Typescript program.  The interfaces defined in xterm.d.ts are not marked as "export", so it's not possible to create a variable of those types.  I think it would be useful if they were exported, in order to create intermediate variables of those types, e.g.:

```
let opts: ITerminalOptions = {
  cursorBlink: true,
};

let term = new Term(opts);
```

Tyriar:
Fixes #1080

### DIFF
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -11,7 +11,7 @@ declare module 'xterm' {
   /**
    * An object containing start up options for the terminal.
    */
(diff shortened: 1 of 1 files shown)


mofux:
LGTM 👍

PR #1106, opened 2017-11-05
Fix deleteChars not updating the view

martin31821:
When deleting a character in a line and moving the cursor to
another line afterwards, the deletion is not rendered.

This commit adds the line where the character was deleted
to the refresh range to enforce an update.

Fixes https://github.com/xtermjs/xterm.js/issues/1110

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -497,6 +497,7 @@ export class InputHandler implements IInputHandler {
       this._terminal.buffer.lines.get(row).splice(this._terminal.buffer.x, 1);
       this._terminal.buffer.lines.get(row).push(ch);
     }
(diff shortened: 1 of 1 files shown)


Tyriar:
Is there a bug that this fixes? Could you provide some repro steps if so?

martin31821:
There is no issue for this bug open at the moment. However it's perfectly reproducible:

* Send 2 lines of text to the terminal using .write
* Move the cursor into the first line
* Send the following escape sequence: `\e[B\e[P\e[A`
* As you can see, the terminal won't update the second line which is clearly a bug.

I'll prepare a fiddle to reproduce it.

Tyriar:
Repro'd and created https://github.com/xtermjs/xterm.js/issues/1110

PR #1108, opened 2017-11-05
Fix #948

### LINKED ISSUE #948: Infinite scrollback
In https://github.com/sourcelair/xterm.js/issues/518 it was mentioned that a large number should be given. In most APIs, if we want to specify such an infinite number, we would either pass `-1` or `0`.

`xterm` causes an error for:
1. `scrollback: -1`, produces:  "Invalid array length"
2. `scrollback: 0`, produces: "cannot read index 0 of undefined"
3. `scrollback: Number.MAX_SAFE_INTEGER`, produces "Invalid array length"

It appears that JS arrays are limited to an unsigned 32 bit value.

A better approach would be to detect `scrollback <= 0 || scrollback > 4294967295` and assign it the max upper value of `4294967295`

bmf-ribeiro:
Fix suggested in #948

This is my first pull request, so all feedback is welcome.

### DIFF
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -11,6 +11,7 @@ export const CHAR_DATA_ATTR_INDEX = 0;
 export const CHAR_DATA_CHAR_INDEX = 1;
 export const CHAR_DATA_WIDTH_INDEX = 2;
 export const CHAR_DATA_CODE_INDEX = 3;
+export const MAX_BUFFER_SIZE = -1 >>> 0; // 2^32 - 1
 
 /**
  * This class represents a terminal buffer (an internal state of the terminal), where the
@@ -68,7 +69,10 @@ export class Buffer implements IBuffer {
     if (!this._hasScrollback) {
       return rows;
     }
-    return rows + this._terminal.options.scrollback;
+
+    const correctBuferLength = rows + this._terminal.options.scrollback;
+
+    return correctBuferLength > MAX_BUFFER_SIZE ? MAX_BUFFER_SIZE : correctBuferLength;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/Buffer.ts
```diff
@@ -11,6 +11,7 @@ export const CHAR_DATA_ATTR_INDEX = 0;
 export const CHAR_DATA_CHAR_INDEX = 1;
 export const CHAR_DATA_WIDTH_INDEX = 2;
 export const CHAR_DATA_CODE_INDEX = 3;
+export const MAX_BUFFER_SIZE = -1 >>> 0; // 2^32 - 1
```
Tyriar: Are you sure this works consistently across all browsers? Doing an unsigned bit operation on a signed number seems dangerous. Might be better to just stick the number in there.

src/Buffer.ts
```diff
@@ -68,7 +69,10 @@ export class Buffer implements IBuffer {
     if (!this._hasScrollback) {
       return rows;
     }
-    return rows + this._terminal.options.scrollback;
+
+    const correctBuferLength = rows + this._terminal.options.scrollback;
```
Tyriar: `correctBuferLength` &rarr; `correctBufferLength`

Also, is this here only to cover the case where scrollback is set from the `Terminal` constructor? If so it might be good to validate at in the Terminal constructor instead of every time `_getCorrectBufferLength` is requested

src/Terminal.ts
```diff
@@ -384,6 +384,8 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
         }
         break;
       case 'scrollback':
+        value = value > 0 && value <= MAX_BUFFER_SIZE ? value : MAX_BUFFER_SIZE;
```
Tyriar: `Math.min` would be preferable to this ternary, I was surprised how fast it was when I looked into it.

src/Buffer.ts
bmf-ribeiro: When i first bumped into this was via terminal.setOption(), and if my understanding is correct we could also trigger this by resizing the terminal, and increasing the number of rows if the buffer size was MAX_BUFFER_SIZE.

Tyriar:
Thanks for the contribution and quick follow up 🎉

PR #1109, opened 2017-11-05
Fix last column selection problems

### LINKED ISSUE #1029: Selection is rendered incorrectly
In the v3 branch, we cannot select the last cell of a line via dragging:

selection

When uncommenting https://github.com/sourcelair/xterm.js/blob/v3/src/utils/MouseHelper.ts#L66 it works, but I think it is not really the solution to the problem...

Tyriar:
Fixes #1029


mofux:
LGTM 👍

Issue #1093, opened 2017-10-28
Expose a selection event


Tyriar:
In order to support features like copy on select an event needs to be exposed which is fired whenever there is a new selection finalized (on mouse up). There is currently a `newselection` event to support middle click on Linux but this probably needs to be separate.

It's probably best not to pass the selection text with the event but instead allow consumers to call `selectionText`. This will save the text being calculated when it's not used.

For those interested in doing a PR checkout https://github.com/sourcelair/xterm.js/blob/v3/src/SelectionManager.ts

PR #1113, opened 2017-11-06
Expose a selection event

bmf-ribeiro:
Implementation of the feature requested in #1093.

If there is additional work to be done on this feature feel free to ask.

### DIFF
--- a/src/SelectionManager.ts
+++ b/src/SelectionManager.ts
@@ -255,6 +255,7 @@ export class SelectionManager extends EventEmitter implements ISelectionManager 
   public selectAll(): void {
     this._model.isSelectAllActive = true;
     this.refresh();
+    this.emit('selection');
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/SelectionManager.ts
```diff
@@ -541,6 +541,9 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
    */
   private _onMouseUp(event: MouseEvent): void {
     this._removeMouseDownListeners();
+
+    if (this.hasSelection)
+      this.emit('selectionended');
```
Tyriar: I think just `'selection'` would be better as this will be public API. In addition to the actual change here, if we want this to expose this as an API it needs to be added here:

- Typings: https://github.com/xtermjs/xterm.js/blob/v3/typings/xterm.d.ts#L221
- Typings test: https://github.com/xtermjs/xterm.js/blob/v3/fixtures/typings-test/typings-test.ts#L66

Tyriar: I think this should fire on `selectAll` as well. https://github.com/Microsoft/vscode/issues/36271 is the main use case for this new event, I think that's all the cases we want to cover.

Issue #778, opened 2017-07-11
[SelectionManager] Select the word underneath the cursor on contextmenu


mofux:
If a user right clicks on a word or url in the terminal, the word should be automatically selected, so it can be copied without having to select it first. This would mimic the default behaviour of Terminal.app, Chrome and most other UIs that allow text selection. I can try to create a PR with this feature if desired.

Tyriar:
I think it would be good to support this for macOS only, a PR would be awesome! One thing to keep in mind is that it generally doesn't seem to select whitespace in the programs I tested, currently [`SelectionManager._selectWordAt`](https://github.com/sourcelair/xterm.js/blob/8fab079ad6cbd8ea8ce3004c4ba4b338ec76aae8/src/SelectionManager.ts#L686) will select whitespace so you'll need to add an extra parameter to that function or something.

mofux:
I was under the assumption that this would be the default behaviour for Windows and Linux too. How about making it platform dependant by default, but configurable through a config flag?

Tyriar:
@mofux sounds good if it's default on macOS only. It would be cool to have opt-in support on other platforms.

Tyriar:
Is it really blocked? I would have thought you'd check MouseEvent for whether it was a right click and/or listen to `contextmenu`?

mofux:
@Tyriar I thought I would just hook into the mousedown handler of the selection manager (the part that is changed by #780), check if options.rightClickSelectsWord is true, selection is currently empty and evt.button is 2 (right mouse button) and basically trigger the same action that is done for double-click (which should select the word underneath the cursor). Then, term.defaults.rightClickSelectsWord would only be true if Browser.isMac

Tyriar:
Right now the right click logic is contained within Clipboard.ts which is called by listening to contextmenu or mousedown depending on the browser. This should probably go inside rightClickHandler which already has the `SelectionManager` passed in.

Tyriar:
Fixed with https://github.com/xtermjs/xterm.js/pull/1114

PR #1114, opened 2017-11-06
Select the word underneath the cursor on contextmenu

nikonso:
Implements feature requested in #778.

Working prototype, although I'm not 100% happy with the new _**selectWordAtCursor**_ function.

Since this is a more ambitious contribution all feedback is welcome!  

Thanks in advance for the review! 😄


typings/xterm.d.ts
```diff
@@ -409,7 +409,7 @@ declare module 'xterm' {
      * Retrieves an option's value from the terminal.
      * @param key The option key.
      */
-    getOption(key: 'cancelEvents' | 'convertEol' | 'cursorBlink' | 'debug' | 'disableStdin' | 'enableBold' | 'popOnBell' | 'screenKeys' | 'useFlowControl' | 'visualBell'): boolean;
+    getOption(key: 'cancelEvents' | 'convertEol' | 'cursorBlink' | 'debug' | 'disableStdin' | 'enableBold' | 'popOnBell' | 'screenKeys' | 'useFlowControl' | 'visualBell' | 'rightClickSelectsWord'): boolean;
```
Tyriar: A matching call for new APIs should also be added to typings-test.ts as well

src/handlers/Clipboard.ts
```diff
@@ -103,10 +103,14 @@ export function moveTextAreaUnderMouseCursor(ev: MouseEvent, textarea: HTMLTextA
  * @param ev The original right click event to be handled.
  * @param textarea The terminal's textarea.
  * @param selectionManager The terminal's selection manager.
+ * @param shouldSelectWord If true and there is no selection the current word will be selected
  */
-export function rightClickHandler(ev: MouseEvent, textarea: HTMLTextAreaElement, selectionManager: ISelectionManager): void {
+export function rightClickHandler(ev: MouseEvent, textarea: HTMLTextAreaElement, selectionManager: ISelectionManager, shouldSelectWord: boolean): void {
   moveTextAreaUnderMouseCursor(ev, textarea);
 
+  if (shouldSelectWord && !selectionManager.hasSelection)
```
Tyriar: Terminal.app and iTerm2 seem to do the selection even if there is already a selection.

src/SelectionManager.ts
```diff
@@ -684,15 +696,19 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
         - leftLongCharOffset // The number of additional chars left of the initial char added by columns with strings longer than 1 (emojis)
         - rightLongCharOffset); // The number of additional chars right of the initial char (inclusive) added by columns with strings longer than 1 (emojis)
 
+    if (!selectWhiteSpace && line.slice(startIndex, endIndex).trim() === '')
```
Tyriar: I like to wrap all blocks in `{}` for readability. Should probably be a tslint rule 😄 https://github.com/xtermjs/xterm.js/issues/1115

src/SelectionManager.ts
```diff
@@ -578,7 +590,7 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
    * Gets positional information for the word at the coordinated specified.
    * @param coords The coordinates to get the word at.
    */
-  private _getWordAt(coords: [number, number]): IWordPosition {
+  private _getWordAt(coords: [number, number], selectWhiteSpace: boolean): IWordPosition {
```
Tyriar: Let's change the variable to be more descriptive, it's long but something like `allowWhitespaceOnlySelection` would be better imo (in other places it's used too).

src/SelectionManager.ts
```diff
@@ -250,6 +250,18 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
   }
 
   /**
+   * Selects word at the current mouse event coordenates.
```
Tyriar: Typo at end

src/handlers/Clipboard.ts
bmf-ribeiro: @Tyriar  are you sure we want to do that as well? Because in that scenario a user can't select right click when this option is enabled without losing the previous selection...

Tyriar: Oh I see what you mean, they only do this when the right click is not on the old selection. We can probably do that as well but feel free to defer it and create a new issue.

src/SelectionManager.ts
```diff
@@ -693,15 +726,20 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
...
   }
 
   /**
    * Selects the word at the coordinates specified.
    * @param coords The coordinates to get the word at.
+   * @param allowWhitespaceOnlySelection If whitespace should be selected
    */
-  protected _selectWordAt(coords: [number, number]): void {
-    const wordPosition = this._getWordAt(coords);
+  protected _selectWordAt(coords: [number, number], allowWhitespaceOnlySelection: boolean): void {
```
Tyriar: I forgot the reason we wanted this allowWhitespaceOnlySelection, were we trying to be consistent with some other terminal?

nikonso: image

I introduced the variable to prevent the selection of empty lines, as shown in the image, as per your suggestion in the Issue thread...

> I think it would be good to support this for macOS only, a PR would be awesome! One thing to keep in mind is that it generally **doesn't seem to select whitespace in the programs I tested**, currently SelectionManager._selectWordAt will select whitespace **so you'll need to add an extra parameter** to that function or something.

Tyriar: Ah that's right. It looks like Terminal.app does select whitespace and iTerm2 doesn't.

src/SelectionManager.ts
```diff
@@ -244,6 +244,39 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
...
+    return (start[1] < coords[1] && end[1] > coords[1]) || (start[1] === coords[1] && coords[0] > start[0]) || (end[1] === coords[1] && coords[0] < end[0]);
+  }
+
+  /**
+   * Selects word at the current mouse event coordinates.
+   * @param event The mouse event.
+   */
+  public selectWordAtCursor(event: MouseEvent): void {
+    const coords = this._getMouseBufferCoords(event);
+    if (coords) {
```
Tyriar: Once we get the coords shouldn't this function defer to `this.selectWordAt`? (and refresh if needed)

nikonso: Because the function `this.selectWordAt` doesn't set `this._model.selectionEnd` the selection is incorrect, that's why I set the both start and the end of the selection in the  `selectWordAtCursor` function.

Because you are more familiar with the code than I am, if you see another way to make this simpler, and without duplicating code I'm open to suggestions.

Tyriar: If I recall correctly you shouldn't need to set `this._model.selectionEnd`, that's the idea behind finalSelectionEnd, if there is nothing set for `selectionEnd`, it will check `selectionStartLength` instead and calculate it:

https://github.com/xtermjs/xterm.js/blob/62d6c242fce8cc4bc983821d914348392833ee83/src/SelectionModel.ts#L79-L82

nikonso: Yes you are correct... I did some more tests and it seems to work, nevertheless sometimes the word selection spans multiple lines when I click certain areas. 😕 

I need so spend some more time to find why that is... 😃

Tyriar: Strange, I would expect it to work basically like double click to select the word now. Let me know if you get stuck and I can take a look 😃

nikonso: Not sure what to do, or if I have to do something, but I think I figured it out, sometimes, on mouse down, when we set the event listeners for dragging, they get in a weird state, and that's why sometimes I was seeing weird behaviors.

xterm js demo

Tyriar: Is that after triggering the context menu? Maybe the mouseup event never fires which drops the mousemove listener?

nikonso: Yes, it is probably is event related, and it **doesn't happen on IE**, haven't tested on firefox yet, also i found simpler repro steps.



I just did, right click, then left click.

Tyriar: Have you tried setting `selectionEnd` to null inside selectWordAtCursor?

Tyriar: That fixed it, pushed the fix

Tyriar:
@nikonso this is great, thanks for your hard work 😃

Issue #1117, opened 2017-11-07
Support setting font weight


Tyriar:
VS Code issue: https://github.com/Microsoft/vscode/issues/29887

Some users would like to be able to set their standard font weight to something other than "normal". This needs to be supported by the new renderer.

Tyriar:
Open to PRs, this will involve adding a new option https://github.com/xtermjs/xterm.js/blob/a6899c74d838e68bc58fb41f718fd7796b9240b2/typings/xterm.d.ts#L14 and making `BaseRenderLayer` aware of it (that's the class that draws the text).

bmf-ribeiro:
I'd like to grab this 😄 

One quick question tough, in the VS Code issue there's the following comment:

> This setting would conflict with the "terminal.integrated.enableBold" setting that currently exists. Should it override that setting if the font weight is set to bold or a higher weight?

What should we do in this case?

Tyriar:
@bmf-ribeiro cool :smiley:

Thinking about it some more, they should probably live together:

- enableBold: Makes bold text act like non-bold text
- fontWeight: Sets the boldness of non-bold text, eg. 100, 200, bold, etc. (the formats accepted by [CanvasRenderingContext2D.font](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/font)

So when enableBold is false and fontWeight is set to "bold", that means that *all* text is bold.

The only thing I'm not sure about right now is what sort of values we should accept, ie. what level of validation should we use.

mofux:
Well, the most reasonable way to get this done would be to make both, the `normal` and `bold` font weights configurable (normal, bold, bolder, 100, 200, ...).

if enableBold is `false`, we would always use the font weight that is configured for normal...

Tyriar:
> make both, the normal and bold font weights configurable

I considered this in the VS Code issue, setting the normal part is the main requirement on the VS Code side to align with the editor setting. Apparently boldFontWeight was not considers when the `editor.fontWeight` setting was added though (as bold is not used as much as in the terminal). Anyway, 👍 to doing both 😄.

For ordering reasons flipping the wording might be a good idea; `fontWeight` (`fontWeightNormal`?), `fontWeightBold`

If we're doing both then maybe `enableBold` should be removed and if people want to implement it they can do it by setting `fontWeightNormal` and `fontWeightBold`. VS Code should be the only consumer right now.

theflyingape:
[...]

PR #1119, opened 2017-11-08
Support setting font weight

nikonso:
Implementation of the feature requested in #1117 based on the comments in the issue.

It seems to be working, but I'm not sure if i missed something.

Should I add typings test for all the available options?


Tyriar:
Bold doesn't seem to be working on the demo initially:

image

Expected (`v3`):

image

src/Terminal.ts
```diff
@@ -414,14 +415,19 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
...
         break;
-      case 'enableBold':
       case 'letterSpacing':
       case 'lineHeight':
         // When the font changes the size of the cells may change which requires a renderer clear
         this.renderer.clear();
         this.renderer.onResize(this.cols, this.rows, false);
         this.refresh(0, this.rows - 1);
         // this.charMeasure.measure(this.options);
+        case 'fontWeight':
```
Tyriar: Fix indentation

typings/xterm.d.ts
```diff
@@ -443,6 +443,12 @@ declare module 'xterm' {
      */
     setOption(key: 'fontFamily' | 'termName' | 'bellSound', value: string): void;
     /**
+    * Sets an option on the terminal.
+    * @param key The option key.
+    * @param value The option value.
+    */
+    setOption(key: 'fontWeight' | 'fontWeightBold', value: null | 'normal' | 'bold' | 'bolder' | 'lighter' | '100' | '200' | '300' | '400' | '500' | '600' | '700' | '800' | '900'): void;
```
Tyriar: This is great :smile:

I don't think null is a valid value? Maybe we need to validate it in `setOption`?

src/Terminal.ts
```diff
@@ -414,14 +415,19 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
...
+        case 'fontWeightBold':
+          // When the font weight changes the size of the cells may change which requires a renderer clear
```
Tyriar: We could merge this and the lineHeight/letterSpacing case and do something like this:

```js
const didCharSizeChange = (key === 'fontWeight' || key === 'fontWeightBold');
this.renderer.onResize(this.cols, this.rows, didCharSizeChange);
```

nikonso:
Locally, and on windows, it seems to be working for me.

image

Do you have any clues as to how i could reproduce the issue? 😕

typings/xterm.d.ts
mofux: @Tyriar From reading the code, passing `null` will use the default font weight, which seems to be a valid scenario?

src/renderer/CharAtlas.ts
```diff
@@ -137,7 +143,7 @@ class CharAtlasGenerator {
 
     this._ctx.save();
     this._ctx.fillStyle = foreground;
-    this._ctx.font = `${fontSize * window.devicePixelRatio}px ${fontFamily}`;
+    this._ctx.font = `${fontWeight} ${fontSize * window.devicePixelRatio}px ${fontFamily}`;
```
mofux: Would be nice if we would reuse the _getFont method introduced in BaseRenderLayer.

src/renderer/CharAtlas.ts
```diff
@@ -151,7 +157,7 @@ class CharAtlasGenerator {
     }
     // Default color bold
     this._ctx.save();
-    this._ctx.font = `bold ${this._ctx.font}`;
+    this._ctx.font = `${fontWeightBold} ${fontSize * window.devicePixelRatio}px ${fontFamily}`;
```
mofux: Would be nice if we would reuse the _getFont method introduced in BaseRenderLayer.

src/renderer/CharAtlas.ts
```diff
@@ -163,11 +169,11 @@ class CharAtlasGenerator {
     this._ctx.restore();
 
     // Colors 0-15
-    this._ctx.font = `${fontSize * window.devicePixelRatio}px ${fontFamily}`;
+    this._ctx.font = `${fontWeight} ${fontSize * window.devicePixelRatio}px ${fontFamily}`;
```
mofux: Would be nice if we would reuse the _getFont method introduced in BaseRenderLayer.

src/renderer/CharAtlas.ts
```diff
@@ -163,11 +169,11 @@ class CharAtlasGenerator {
...
     for (let colorIndex = 0; colorIndex < 16; colorIndex++) {
       // colors 8-15 are bold
       if (colorIndex === 8) {
-        this._ctx.font = `bold ${this._ctx.font}`;
+        this._ctx.font = `${fontWeightBold} ${fontSize * window.devicePixelRatio}px ${fontFamily}`;
```
mofux: Would be nice if we would reuse the _getFont method introduced in BaseRenderLayer.

src/renderer/TextRenderLayer.ts
```diff
@@ -33,7 +33,7 @@ export class TextRenderLayer extends BaseRenderLayer {
     super.resize(terminal, dim, charSizeChanged);
 
     // Clear the character width cache if the font or width has changed
-    const terminalFont = `${terminal.options.fontSize * window.devicePixelRatio}px ${terminal.options.fontFamily}`;
+    const terminalFont = `${terminal.options.fontWeight} ${terminal.options.fontSize * window.devicePixelRatio}px ${terminal.options.fontFamily}`;
```
mofux: Would be nice if we would reuse the _getFont method introduced in BaseRenderLayer.

mofux:
> Bold doesn't seem to be working on the demo initially

Are we seeing a bug here? If I read this comment correctly, bold is only working for normal colored chars, but not for colored chars, which also explains why @nikonso does see it working (white bold text), but @Tyriar is not (colored bold prompt)...

https://github.com/xtermjs/xterm.js/issues/1117#issuecomment-343180815

Tyriar:
@mofux yeah looks like non-white colors in the char atlas are not being drawn as bold:

screen shot 2017-11-09 at 10 17 15 am

nikonso:
Ok, I'l be away from the computer during the weekend, so I'll try to fix
that when I get back on Sunday :wink:

Tyriar:
@nikonso no problems, thanks for all the help recently 😃

Tyriar:
Sorry about the delay, looks good to push in now :+1:

PR #1123, opened 2017-11-09
Support setting a padding on the .xterm element

### LINKED ISSUE #946: Add an option to set the terminal padding
I would like to come up with a solution for the following issue:

Say you have a tab that renders the terminal, say it's 300x200. Now if we put the terminal into that tab, we can either make the terminal container to fill it 100%, but in that case the terminal text will touch the edges of the tab at the left, right, top and bottom, which might look weird in some situations. So an easy solution would be to add a 5px margin to the tab container, which does the job, but now you have the viewport scrollbar drawn at a weird position. I think this problem cannot be solved through css (correct me if I'm wrong) because the viewport has to be made aware of the margin to still work in sync. I would therefore propose to make the margin (or better call it padding) configurable through a config option, and then make xterm aware of it:

```javascript
// 5px padding to all sides
term.setOption('padding', 5);

// individual paddings
term.setOption('padding', {
  top: 5, 
  left: 0,
  right: 10,
  bottom: 5
});
```

### The problem in pictures

Fit the tab
clash
5px margin to the tab, weird looking scrollbar
margin bad
5px margin, but scrollbar matching the tab container
margin right

### LINKED ISSUE #1081: Wrong canvas width calculation
I want to integrate a terminal in a fixed width element and have xterm.js figure out how many columns fit in this space.  I think that the calculation of the number of columns (and therefore the canvas width) has a problem, resulting in less columns than there should be, and therefore some wasted space on the right.

First, a screenshot:

xterm

The red portion represents the canvas elements, the black is the viewport element.

What happens is the following:

* In [`fit.js::proposeGeometry`](https://github.com/sourcelair/xterm.js/blob/c2b3853aa7aace1353844c844a09c6a4fce8c1cd/src/addons/fit/fit.js#L36), the parent element is 983px wide and term.charMeasure.width is 7.8125px.  proposeGeometry therefore returns 125 columns.
* In [`Renderer.ts::onResize`](https://github.com/sourcelair/xterm.js/blob/c2b3853aa7aace1353844c844a09c6a4fce8c1cd/src/renderer/Renderer.ts#L89) the character width is floored, which makes it 7px.  The total canvas size is then computed as the number of columns times the character width, which is `7 px * 125 cols = 875 px`.
* The canvas is drawn 875 px wide, even though the available area is of 983 px.

I suppose the solution would be include using the same character width in both computation.

## Details
- Browser and browser version: Chromium Version 61.0.3163.100
- OS version: Ubuntu 16.04
- xterm.js version: v3 (c2b3853aa7aace1353844c844a09c6a4fce8c1cd)

### Steps to reproduce

1. Check out the following branch: https://github.com/simark/xterm.js/tree/repro-canvas-size
2. Build and run the xterm.js demo

It's possible for the character width to vary between platforms, and therefore you might see something a bit different than me.  For me, the default font gives a character width of ~9.01 px, so you don't really notice the bug since it doesn't make a big difference when it gets rounded to 9 px.  With a font size of 13 px, the width is 7.8125 px, which makes quite a big difference between when it gets rounded to 7.

mofux:
Fixes #946 
Fixes #1081 

This PR moves the canvas layers into their own div element, .xterm-screen. This allows us to apply a css padding to the .xterm element and to have the scrollbar still fill the whole height of the .xterm element. 

The fit() addon has also been updated to support this new feature.

fit


src/Viewport.ts
```diff
@@ -48,13 +50,9 @@ export class Viewport implements IViewport {
   private refresh(): void {
     if (this.charMeasure.height > 0) {
       this.currentRowHeight = this.terminal.renderer.dimensions.scaledCellHeight / window.devicePixelRatio;
+      this.lastRecordedViewportHeight = this.viewportElement.offsetHeight;
 
-      if (this.lastRecordedViewportHeight !== this.terminal.renderer.dimensions.canvasHeight) {
-        this.lastRecordedViewportHeight = this.terminal.renderer.dimensions.canvasHeight;
-        this.viewportElement.style.height = this.lastRecordedViewportHeight + 'px';
```
Tyriar: Why do we not need to set this anymore?

mofux:
The scrollbar is still there, but OSX draws it black, so it is really hard to see:

screen shot 2017-11-19 at 02 00 24

OSX draws a light or dark scrollbar, depending on the background color of the scrolling container. Unfortunately we cannot set the background color on the scrolling container because that would hide the content (canvas) underneath it 😟  Setting a opacity on the scroll container also applies that opacity to the scrollbar, so that doesn't work.  The only thing that is working is to set the terminal background color on the scroll container with a very low alpha value e.g. `background-color: rgba(0,0,0,0.01)` if the terminal background was `rgb(0,0,0)`, but that would mean that we would have to calculate that very opaque color from the currently used background color of the theme, and it would also add a little performance impact to the rendering pipeline to draw the (likely invisible) semi-transparent background. Not sure how to go on with this really...

src/Viewport.ts
mofux: Well the new .xterm-screen element is positioned relative to the .xterm parent and has a the same height as the canvas elements inside. Its width equals the width of the canvas + the width of the scrollbar on the viewport. If you add a padding to the .xterm element, that means that the .xterm element will grow by that padding. The viewport is positioned absolute to always match the size of the .xterm element. For that reason, we don't have to manually set the height of the viewport element any longer.

mofux:
I have added a workaround for the overlay scrollbar being barely visible on OSX. It sets the background color of the viewport element by taking the background color from the theme (which by the current design [must be](https://github.com/xtermjs/xterm.js/blob/8ece306dbe9c7aee13cce9c60fae74202f8660ee/src/renderer/ColorManager.ts#L89) a 7 digit hex value, e.g. `#000000`) and adding a very low alpha (01) to it (`#00000001`). This background color on the viewport is not perceptible by the human eye, but it fools OSX into calculating the scrollbar thumb color based on that background color, ignoring the alpha. This solution adds another alpha pass to the browser's rendering pipeline which will slow it down a little bit. Unfortunately I couldn't find any other workaround, so I made sure this workaround is only applied on OSX, and if an overlay scrollbar is used (which results in a calculated scrollbar width of 0).

mofux:
The approach above doesn't work on all browsers (tested with the demo):
- Chrome: Draws a white scrollbar thumb 👍
- Firefox: Draws a black scrollbar thumb, works correctly if no background color is applied to the viewport element 👍/👎
- Safari: Draws a black scrollbar thumb, only works when setting a background color alpha > 0.9 👎

**Concusion**:
We can't get the overlay scrollbar to work reliably on all browsers.

**Solutions:**
1. Account some extra space for the scrollbar if the scrollbar width is 0 (OSX overlay scrollbar). This will waste some space and make the terminal look weird if not currently being scrolled, but I think that is what we did before.
2. Overwrite the scrollbar style with custom css. This will force OSX to always draw the scrollbar, but unfortunately we cannot re-use the system scrollbar style 😡
3. Introduce the workarounds for Chrome and Firefox and fallback to 1. for the other browsers.
4. Have somebody smarter than me figure out a css trick that works across all these browsers.

**Workarounds I've tried:**
- Set opacity on the viewport element: Also applies that opacity to the scrollbar
- Use the css `clip` property to clip away the background: Also clips away the scrollbar
- Use background color with alpha: See above
- Use `transform: scaleX` to scale down the width of the viewport element: Also applies that scale to the scrollbar
- Make the viewport container only 1px wide but set `overflow-x: visible` on it: Only shows 1px of the scrollbar

Tyriar:
I don't think the `01` alpha trick is ideal as that's extra a bunch of extra processing that needs to happen when compositing. I looked into a possible solution and got something that kind of works, the basic outline is:

- Move the `.xterm-screen` element after `.xterm-viewport` so that it is drawn on top
- If the OS is macOS and scroll bar width is 0, assume it's `x` pixels large
- Make room for the scroll bar in `.xterm-viewport`, not `.xterm-screen`

The dark scroll bar still happens on Safari but that was there before.

Here's a very messy diff as a proof of concept:

```
diff --git a/src/Terminal.ts b/src/Terminal.ts
index c5a7575..a101473 100644
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -589,10 +589,6 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
 
     this.element.setAttribute('tabindex', '0');
 
-    this.screenElement = document.createElement('div');
-    this.screenElement.classList.add('xterm-screen');
-    this.element.appendChild(this.screenElement);
-
     this.viewportElement = document.createElement('div');
     this.viewportElement.classList.add('xterm-viewport');
     this.element.appendChild(this.viewportElement);
@@ -600,6 +596,10 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this.viewportScrollArea.classList.add('xterm-scroll-area');
     this.viewportElement.appendChild(this.viewportScrollArea);
 
+    this.screenElement = document.createElement('div');
+    this.screenElement.classList.add('xterm-screen');
+    this.element.appendChild(this.screenElement);
+
     // preload audio
     this.syncBellSound();
 
diff --git a/src/Viewport.ts b/src/Viewport.ts
index 0581445..07fcbcb 100644
--- a/src/Viewport.ts
+++ b/src/Viewport.ts
@@ -33,7 +33,7 @@ export class Viewport implements IViewport {
     private scrollArea: HTMLElement,
     private charMeasure: CharMeasure
   ) {
-    this.scrollBarWidth = this.viewportElement.offsetWidth - this.scrollArea.offsetWidth;
+    this.scrollBarWidth = (this.viewportElement.offsetWidth - this.scrollArea.offsetWidth) || 10;
     this.viewportElement.addEventListener('scroll', this.onScroll.bind(this));
 
     // Perform this async to ensure the CharMeasure is ready.
@@ -47,9 +47,10 @@ export class Viewport implements IViewport {
     // calculates the scrollbar thumb color from the background color of the viewport element.
     // Since the viewport element has to be transparent in order to see the underlying canvas,
     // we use the terminal background color from the theme and add a very low alpha value to it.
-    if (isMac && isChrome && this.scrollBarWidth === 0 && colors.background.length === 7) {
-      this.viewportElement.style.backgroundColor = `${colors.background}01`;
-    }
+    // if (isMac && isChrome && this.scrollBarWidth === 0 && colors.background.length === 7) {
+
+    this.viewportElement.style.backgroundColor = colors.background;
+    // }
   }
 
   /**
@@ -57,6 +58,8 @@ export class Viewport implements IViewport {
    * necessary.
    */
   private refresh(): void {
+    console.log('scroll', this.scrollBarWidth);
+    this.viewportElement.style.width = `${this.terminal.renderer.dimensions.canvasWidth + this.scrollBarWidth}px`;
     if (this.charMeasure.height > 0) {
       this.currentRowHeight = this.terminal.renderer.dimensions.scaledCellHeight / window.devicePixelRatio;
       this.lastRecordedViewportHeight = this.viewportElement.offsetHeight;
diff --git a/src/renderer/Renderer.ts b/src/renderer/Renderer.ts
index b6a5c68..a7f6c0b 100644
--- a/src/renderer/Renderer.ts
+++ b/src/renderer/Renderer.ts
@@ -136,7 +136,7 @@ export class Renderer extends EventEmitter implements IRenderer {
     this._renderLayers.forEach(l => l.resize(this._terminal, this.dimensions, didCharSizeChange));
 
 		// Resize the screen
-    this._terminal.screenElement.style.width = `${this.dimensions.canvasWidth + this._terminal.viewport.scrollBarWidth}px`;
+    this._terminal.screenElement.style.width = `${this.dimensions.canvasWidth}px`;
     this._terminal.screenElement.style.height = `${this.dimensions.canvasHeight}px`;
 
     // Force a refresh
```

mofux:
@Tyriar I'm tempted to not use a workaround for this at all. I agree that the hex value hack is bad and shouldn't go into production. 

I think that having empty space beside the screen area is even worse than having a barely visible overlay scrollbar. Ultimately, this bug should be solved by Chrome and Safari - Firefox seems to do the best job here. Last but not least, implementors can always provide their own scrollbar style via css (like vscode does), which will work just fine. @parisk @Tyriar Any thoughts?

Tyriar:
Let's defer this as we need to get v3 out, @mofux could you pull out the fix for https://github.com/xtermjs/xterm.js/issues/1081 into it's own PR? We should be able to merge something like that easily.

mofux:
@Tyriar no problem. I'll get to it tomorrow, hope that's soon enough.

Tyriar:
@mofux sounds good :ok_hand:

mofux:
@Tyriar @parisk I have updated the PR to account additional 15px as the scrollbar width if the actual scrollbar width is 0 (as suggested by @Tyriar above). I think this is a good compromise for now (and it mirrors the current behaviour). Please review again, hope it's good to merge now. 🧐

Tyriar:
`Terminal.scrollToBottom` doesn't seem to work properly when you have padding. On macOS, I scroll up to top then press a key (which triggers `scrollToBottom`) and the terminal scrolls to the second last line.

There is also a test failure on Linux (`xterm output comparison t0500-bash_long_line.in`  https://travis-ci.org/xtermjs/xterm.js/jobs/315932024) and a compile problem :`src\Viewport.ts(9,17): error TS2305: Module '"C:/Users/Daniel/Documents/dev/mofux/xterm.js/src/utils/Browser"' has no exported member 'isChrome'.`

Tyriar:
Pulled the `fit` sizing fix into https://github.com/xtermjs/xterm.js/pull/1154

cancan101:
Any updates on this PR?

mofux:
Closing this PR in favour of #1208

Issue #1128, opened 2017-11-16
Custom addons


feamsr00:
Hello all!

I'm trying to create a new addon and the documentation doesn't seem to be terribly clear on what it takes to go from 0-60. 

For example, 1) exactly how does the API for add-ons differ from the nominal Terminal API? Does it at all? Will it?* Additionally, 2) is directly modifying Terminals' prototype the most appropriate way for add-ons to register functionality? It seems to be just asking for collisions. Is there any other namespace registration or dic facility? (I suppose maybe even just adding the addon object eg Terminal.MyAddon.method() for the simplest way, but surely Terminal.addon('MyAddon').method() is much more sound). 3) Also, it doesn't seem to be clear how to actually add a third party addon, as the names have all been hardcoded... (I have taken to extending Terminal and widening loadAddon (`static loadAddon(String): void;`)) I am making assumptions, since the docs don't say if loadAddon is only for private (non 3rd party) use.

I have reviewed some of the existing add-ons, but they don't seem to have the most consistent implementations.

*I'm creating a couple add-ons (at least one for custom escape codes) because I wish to add discrete functionality to xTerm.js. It seems far more architecturally sound to do that instead of just making a big abstraction layer on *top* of xTerm.

## Details
- tsc.exe version: 2.6.1
- xterm.js version: 3

jerch:
somewhat offtopic: For custom escape codes there is imho no way to hook into the default parser atm. You can still get those custom codes working with your own parser beforehand. The ANSI spec allows custom escape codes for OSC and DCS commands. Unless you know what you are doing - stick with those to stay compatible with the default parser.

feamsr00:
@jerch Thanks for the insight. I was worried about that. This kinda goes back to the API question. Even without a special API for add-ons per se, It's not clear how one could merely call a "Terminal.setParser()" (especially considering no such method seems to exist) to set that dependency. Moreover, it would be helpful if there was some interface contract for Parser. Are there any custom parsers in the wild, or other examples on how to move forward?
Might I suggest a feature to at least emit parser/ctrl events?

Tyriar:
@feamsr00 currently there are no real external addons, everything is built in. The main reason we have not actively encouraged the community building addons right now is all they would be able to do is call the standard API of which you can't really do much interesting stuff that would call for an addon.

An addon could of course call into private members but it will likely break in the future as we don't commit to keeping private APIs stable. The addons packaged with the repo have a better chance of being stable at any given time.

[...]

PR #1130, opened 2017-11-17
New addon system: `applyAddon`

### LINKED ISSUE #1018: Master issue for webpack/loadAddon issues
I've seen a bunch of issues recently come in around how loadAddon works and I wonder if we're over complicating things. In today's world of bundlers and source maps, it appears very difficult to get this right in a way that will work across all build systems.

## Goals

- We need an easy way to load addons without forcing the user to download and parse/run the script. The core should be as slim as possible, leaving out optional components.
- It should work with webpack, browserify and vscode's module system (and more?)
- It should work nicely with TypeScript

### Current Situation

- We wrap addons in UMD which allows handling of the addon script in a plan browser environment, CommonJS environment and require.js environment. A problem here is that we needed to reorder the UMD definition as VS Code defines the `define` global, messing with the typical format and introducing issues for people eg. https://github.com/sourcelair/xterm.js/pull/994
- In a CommonJS environment the addon script attempts to load in the `Terminal` object via `require`, this causes issues for webpack https://github.com/sourcelair/xterm.js/issues/927
- `Terminal` inside addons is not typed currently as the definition for `Terminal` lies outside of the addon folder

## Potential Solutions

I'm not attempting to solve the problem right now, here are some ideas though:

- Have `loadAddon` pull in the script addon object and apply it to `Terminal`, as opposed to leaving it up to the addon to load `Terminal` (see https://github.com/sourcelair/xterm.js/issues/927#issuecomment-325210992)

  ```js
  public static loadAddon(addon: string, callback?: Function): boolean | any {
      // TODO: Improve return type and documentation
      if (typeof exports === 'object' && typeof module === 'object') {
        // CommonJS
        return require('./addons/' + addon + '/' + addon)(Terminal);
      } else if (typeof define === 'function') {
  ...
  ```
- Require bundling or concatenation when using addons?

If anyone has more insights into how this sort of dynamic loading at runtime issue is typically handled, I'd love to hear it :smiley: 

CCing people from related issues: @parisk @mofux @ficristo @dgriffen @ebertmi @rdicroce @jclangst @LucaT1 @jan0e @KagamiChan @vincentwoo

parisk:
This Pull Request introduces a new API for using xterm.js add-ons that does not mess up with importing stuff, along with a handful of breaking changes to standardize and simplify how add-ons work and how they should be built.

I also created a repository, where you can test that this branch works fine with Webpack: https://github.com/parisk/xterm-webpack.

## 🆕 API

Now, to use an add-on in xterm.js the developer has to import it first and then to *apply* it on the `Terminal` constructor.

```javascript
import { Terminal } from xterm;
import * as fit from 'xterm/lib/addons/fit/fit';


Terminal.applyAddon(fit);

var xterm = new Terminal();  // Instantiate the terminal
xterm.fit();                 // Use the `fit` method, provided by the `fit` addon
```

## ⚠️ Breaking changes

- `loadAddon` does not exist any more
- Add-ons do not get imported automatically any more
- Add-ons export only an `apply` method

This PR closes #1018.


parisk:
@Tyriar, @mofux I am requesting a review from both of view, mostly because I want you to be aware of the changes coming up, as this is a significant API breakage and hear any comments or feedback on it 😄.

mofux:
Looks very good to me! The only shortcoming I can see at the moment is that one has no chance to initialise an addon right after instantiation. Do you think it would make sense to provide an api for that as well?

```javascript
const middleware = (term) => {
  console.log('Ready!');
}

Terminal.use(middleware);
const term = new Terminal(); // logs "Ready!" 
```

If the middleware needs to accept some configuration, it could be done like this:

```javascript
const middleware = (opts) => {
  return (term) => {
    console.log('Ready!', opts.foo);
  }
}

Terminal.use(middleware({ foo: 'bar' }));
const term = new Terminal(); // logs "Ready! bar" 
```

coderaiser:
Good idea to use `use` instead of `applyAddon` like [expressjs does](https://expressjs.com/en/4x/api.html#router.use) and allow chaining :).

mofux:
@coderaiser I think there are two use-cases for a addon api. The `Terminal.use` case is helpful for addons that automatically add behaviour to the terminal. For example, an addon that parses the contents of the prompt line and emits a custom event, say `prompt`, whenever it detects that the prompt has changed.

The other use-case is for addons like the `fit` or `attach` addon, that have to be explicitly called to perform an action. For example, you must call `term.fit()` in order for the fit addon to do anything. One question that is still under discussion is if these addons should really extend the Terminal prototype, because they could easily work by passing the terminal instance to them:

``` javascript
import {fit} from 'xterm/lib/addons/fit/fit';
const term = new Terminal();
fit(term);
```

Tyriar:
I touched on this on Slack but I'll expand on it here. I prefer @mofux's suggestion:

```ts
import {fit} from 'xterm/lib/addons/fit/fit';
const term = new Terminal();
fit(term);
```

The most important reason I believe this is I don't think we can provide first class TypeScript support if we're modifying the prototype. Currently we need to add the types of all built-in addons to the [xterm.d.ts typings file](https://github.com/xtermjs/xterm.js/blob/v3/typings/xterm.d.ts) in order for TypeScript to pick them up. This is what I do in vscode; I fork the xterm.d.ts file and add the addon methods I use: 

https://github.com/Microsoft/vscode/blob/1f040a1258c13f39938bb39d296b86db5df34152/src/typings/xterm.d.ts#L529-L545

As mentioned previously, we could add all the addon types to the xterm.d.ts file, but the problem here is that then the available methods on `Terminal` becomes cluttered and filled with methods that may not actually do anything because the addon is not loaded.

Even ignoring the problem with bundled addons, we should consider what happens when custom addons https://github.com/xtermjs/xterm.js/issues/1128 are a thing. If the way addons work is by modifying the prototype then they will not have types as we certainly do not want to add external custom addon typings in xterm.d.ts. That will force users to use custom addons like this:

```ts
import { Terminal } from xterm;
import * as xtermCustomFit from 'xterm-custom-fit';
import * as xtermFind from 'xterm-find';

Terminal.applyAddon(xtermCustomFit);
Terminal.applyAddon(xtermFind);

var xterm = new Terminal();
(<any>xterm).customFit(); // 😐
(<any>xterm).findNext();  // 😑

```

The alternative is:

```ts
import { Terminal } from xterm;
import { customFit } from 'xterm-custom-fit';
import { findNext, findPrevious } from 'xterm-find';

var xterm = new Terminal();
customFit(xterm);
findNext(xterm);
// Types work if the addon defines them 🙂
```

The other reasons I believe we should drop loadAddon/applyAddon are:

- `applyAddon` is a static methods and also tampers with the prototype, both of which should be avoided in genral
- `applyAddon` makes testing a bit harder
- Addons seldom have constructors, currently the `applyAddon` functions don't do much other than modifying the prototype. The other way avoids this call all together, instead you call the addon when it's ready. This could be immediately after `new Terminal` if you needed the addon to activate on construction, or when it's needed (eg. `findNext`)

parisk:
OK, I pushed an update that I think can make most of us happy with few trade-offs.

Before explaining the update I just pushed, I want to state that I will insist on keeping the prototype extension. Let me explain the reasons why:

1. Until now, the prototype extension has not been the source of any issues (the main issue with `loadAddon` was the automatic importing. So there is nothing to "fix" here.
2. The prototype extension provides a better user experience for those who want to use it. For example, for us at SourceLair, the prototype extension means:
    - Setting up the class with additional capabilities once
    - Using less imports wherever we need the terminal
    - Passing less arguments to terminal addon methods (no `term` argument needed)
3. Static/class methods and class/prototype extensions are features of programming languages that are meant to be used (and are all being used by trustworthy projects like CodeMirror, Django), so I do not believe that they are bad practice -- we just have to use them right

Saying that, I exported static add-on methods wherever possible to allow also seamless usage for people that do not prefer `applyAddon` and for better TypeScript integration.

I wasn't able to do that for the zmodem add-on, which is using the ZModem global variable.

Personally, I am OK with supporting both `applyAddon` and the standalone addon modules.

src/addons/fit/fit.ts
```diff
@@ -0,0 +1,60 @@
...
+      if (term.rows !== geometry.rows || term.cols !== geometry.cols) {
+        term.renderer.clear();
+        term.resize(geometry.cols, geometry.rows);
+      }
+    }
+  }, 0);
+};
+
+export function apply(terminalConstructor) {
+  terminalConstructor.prototype.proposeGeometry = function () {
```
Tyriar: This should work and be even more compact:

```
terminalConstructor.prototype.proposeGeometry = proposeGeometry.bind(this, this)
```

src/addons/attach/attach.ts
```diff
@@ -0,0 +1,119 @@
...
+   *
+   * @param {WebSocket} socket - The socket to attach the current terminal.
+   * @param {boolean} bidirectional - Whether the terminal should send data
+   *                                  to the socket as well.
+   * @param {boolean} buffered - Whether the rendering of incoming data
+   *                             should happen instantly or at a maximum
+   *                             frequency of 1 rendering per 10ms.
+   */
+  terminalConstructor.prototype.attach = function(socket, bidirectional, buffered) {
+    return exports.attach(this, socket, bidirectional, buffered);
```
Tyriar: I think we can omit the `exports.` here and make it consistent with other calls.

src/addons/zmodem/zmodem.ts
```diff
@@ -0,0 +1,89 @@
...
+ * **IMPORTANT:** After you confirm() a zmodem.js Detection, if you have
+ *  used the `attach` or `terminado` addons, you’ll need to suspend their
+ *  operation for the duration of the ZMODEM session. (The demo does this
+ *  via `detach()` and a re-`attach()`.)
+ */
+
+export function apply(terminalConstructor) {
+  let Zmodem = (typeof window == 'object') ? (<any>window).ZModem : {Browser: null};  // Nullify browser for tests
+
+  terminalConstructor.prototype.zmodemAttach = function(ws, opts) {
```
Tyriar: I think we can still pull this `zmodemAttach` function out even though it relies on `window.ZModem`?

parisk:
@Tyriar fixed comments. As soon as tests pass, this should be GTG.

## A few more changes introduced

### Webpack
Now we need to bundle the JavaScript files of the demo, since `applyAddon` requires "importing/requiring" the add-on module.

I went with Webpack, instead of Browserify for two reasons:

1. Browserify gets too complex when requiring already bundled files (e.g. `build/xterm.js`)
2. I thought it would be a good idea to have a built-in webpack example in the repo

## Slightly updated repo structure

I removed and updated some descriptor files (e.g. removed `Procfile.dev`) and I removed `nodemod` for the time being until we find a more elegant way to use it in the demo. The problem with it is that it was looping perpetually, because of Webpack changing files.

Tyriar:
I get this error when I run the demo (via `npm start`):

image

Tyriar:
I also can't seem to access `term` in the console anymore?

parisk:
Woo! 🎉

simark:
@parisk, following this PR, I think the static method name should be changed in `typings/xterm.d.js`, from loadAddon to applyAddon.

Tyriar:
@simark thanks for pointing that out, created https://github.com/xtermjs/xterm.js/pull/1163

PR #1137, opened 2017-12-03
Fix parser not resetting state for some escape sequences

damz:
Test case:

```
$ echo -e '\e|test'
est
```

And an _Unknown ESC control: t._ error is generated in debug mode.


Tyriar:
Does this fix make it behave the same as other terminal emulators?

image

Tyriar:
Thanks @damz! Sorry about the delay, we only just got v3 out

PR #1144, opened 2017-12-11
Use IntersectionObserver to turn off renderer when hidden

### LINKED ISSUE #880: Performance: pause / resume rendering
At the moment, the renderer will always draw incoming data (manipulate the dom), no matter if the terminal is actually visible to the user. Bigger applications like hyper or vscode, that may render multiple terminals in tabs would benefit a lot (performance and battery drain) if they could suspend the rendering while the terminal element is visually hidden. I think we should offer an API that allows the implementor to pause and resume the renderer. 

```
term.pauseRenderer();
term.resumeRenderer();
term.rendererPaused;
```

Once the renderer is resumed, it has to immediately redraw to catch up the new state. 
Any thoughts? I would be tempted to provide a PR 😉

Tyriar:
Fixes #880 

---

@mofux seems to work nicely in Chrome, what do you think? This can save like 5ms a frame for heavy operations :fire:


Tyriar:
Here's an annotated timeline of the change within VS Code doing the following:

1. Open a terminal
2. Run `ls -lR` in large folder
3. Create new terminal (hide busy terminal)

image

To give an idea of the time saved, the draws on the busy terminal before the switch are about 5ms, 5ms and 2ms every available animation frame.

Tyriar:
Linux tests are passing, merging this in (mac agents are backed up)

Issue #965, opened 2017-09-10
Canvas renderer cursor do not hide


tsl0922:
Cursor may not hided sometimes with the canvas renderer.

## Details
- Browser and browser version: Chrome 60.0.3112.113
- OS version: MacOS Sierra 10.12.6
- xterm.js version: v3 HEAD

### Steps to reproduce

1. run `htop` (`brew install htop`)
2. or run `sl` (`brew install sl`)

### Screenshots

![image](https://user-images.githubusercontent.com/1680515/30250687-735905ec-9685-11e7-80a0-80bc7ac88bb4.png)
![image](https://user-images.githubusercontent.com/1680515/30250700-9f1bd056-9685-11e7-95be-89da58af2aa5.png)

Tyriar:
Thanks for the report, I can reproduce.

LoganDark:
`brew` is a macOS-specific package manager.

tsl0922:
This does not happen on the latest master branch anymore, closing.

PR #1146, opened 2017-12-12
Fix cursor on focus

chabou:
In case of multiple terms in same document, cursor is not refreshed when focusing.
To fix that, this PR adds same refresh mechanism as blur.


chabou:
🤔  It seems to fix a side effect when term is badly reinjected in DOM.
You can reject this PR. (But it could be a good and free protection)

Tyriar:
You're not seeing https://github.com/xtermjs/xterm.js/issues/965 are you? It would be good to understand more about the issue you're seeing as the current v3 should always draw regardless of whether the element is on the DOM or not.

Tyriar:
Closing this until there is a more solid repro :smiley:

PR #1148, opened 2017-12-14
Add allowTransparency option

### LINKED ISSUE #1004: Support transparent backgrounds
It looks as though starting with 365661704424f9a33b3a2cb4b85ad144948a86a9 setting the themes background to 'none' does not work. It just sets the background to black. If it helps at all I found changing this line: https://github.com/sourcelair/xterm.js/blob/c4da8325ace6e4bde77822377bf994d389a7c75a/src/renderer/TextRenderLayer.ts#L104 to `let bg = this._colors.background` fixed the issue but not sure if that is the correct solution to the problem.

chabou:
Fixes #1004 

I didn't notice any performance loss (measuring frame rendering duration for `vtop` on electron).


src/renderer/ColorManager.ts
```diff
@@ -107,20 +107,4 @@ export class ColorManager implements IColorManager {
     this.colors.ansi[14] = theme.brightCyan || DEFAULT_ANSI_COLORS[14];
     this.colors.ansi[15] = theme.brightWhite || DEFAULT_ANSI_COLORS[15];
   }
-
-  private _validateColor(color: string, fallback: string): string {
```
Tyriar: We still need this to support `#rgb` notation as it doesn't work in canvas, I think just expanding it to allow `#rrggbbaa` will be fine

src/renderer/Renderer.ts
```diff
@@ -25,14 +25,15 @@ export class Renderer extends EventEmitter implements IRenderer {
   public colorManager: ColorManager;
   public dimensions: IRenderDimensions;
 
-  constructor(private _terminal: ITerminal, theme: ITheme) {
+  constructor(private _terminal: ITerminal, theme: ITheme, alpha: boolean) {
```
Tyriar: Let's just reference `this._terminal.options.alpha` here instead of adding another param

src/renderer/TextRenderLayer.ts
```diff
@@ -24,8 +24,8 @@ export class TextRenderLayer extends BaseRenderLayer {
   private _characterFont: string;
   private _characterOverlapCache: { [key: string]: boolean } = {};
 
-  constructor(container: HTMLElement, zIndex: number, colors: IColorSet) {
-    super(container, 'text', zIndex, false, colors);
+  constructor(container: HTMLElement, zIndex: number, alpha: boolean, colors: IColorSet) {
```
Tyriar: Let's move `alpha` to the end of the param list to be consistent with `LinkRenderLayer`

src/renderer/ColorManager.ts
chabou: I can expand #rgb to #rrggbb if detected.
But if we want to support `rgba(r,g,b,a)` and `hsl(h,s,l,a)` or `transparent`, we should not fallback to DEFAULT_BACKGROUND if `#rgb` and `#rrggbb` are not detected.
Is it ok for you ?

src/renderer/Renderer.ts
chabou: 👍

src/renderer/TextRenderLayer.ts
chabou: 👍

src/renderer/ColorManager.ts
chabou: Actually, I just tested #rgb notation and it works great 🤔

Tyriar:
Closing in favor of https://github.com/xtermjs/xterm.js/pull/1236

PR #1153, opened 2017-12-18
Add null check to SelectionManager.selectionText

### LINKED ISSUE #1152: Null pointer exception in SelectionManager
VS Code issue: https://github.com/Microsoft/vscode/issues/38292

Failing line:

https://github.com/xtermjs/xterm.js/blob/eeda380b6776320d54bd6d8a34169b66f00acae0/src/SelectionManager.ts#L197

Tyriar:
Fixes #1152


Tyriar:
Found the root cause, see https://github.com/xtermjs/xterm.js/pull/1168

PR #1157, opened 2017-12-18
Fix fit() sizing and also support letterSpacing, enhanced

mofux:
Improves #1154 by letting CharMeasure and Renderer measure their dimensions synchronously. This makes sure that canvas and character sizes are available right after calling `term.open()`, enabling `term.fit()` to be called synchronously right after it. This PR also fixes a bug in the demo where resize didn't work anymore because `pid` was undefined due to a javascript scope difference introduced by the new Webpack bundler.


src/Terminal.ts
```diff
@@ -578,8 +578,8 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this.element = this.document.createElement('div');
     this.element.classList.add('terminal');
     this.element.classList.add('xterm');
-
     this.element.setAttribute('tabindex', '0');
+    this.parent.appendChild(this.element);
```
Tyriar: `viewport` and `helperContainer` should probably be added to a `DocumentFragment` first now to prevent multiple layouts from `element.appendChild`

src/utils/CharMeasure.ts
```diff
@@ -33,24 +41,6 @@ export class CharMeasure extends EventEmitter implements ICharMeasure {
...
-      this._measureElement = this._document.createElement('span');
-      this._measureElement.style.position = 'absolute';
-      this._measureElement.style.top = '0';
-      this._measureElement.style.left = '-9999em';
-      this._measureElement.style.lineHeight = 'normal';
-      this._measureElement.textContent = 'W';
-      this._measureElement.setAttribute('aria-hidden', 'true');
-      this._parentElement.appendChild(this._measureElement);
-      // Perform _doMeasure async if the element was just attached as sometimes
-      // getBoundingClientRect does not return accurate values without this.
```
Tyriar: Did you observe this? Could this have been from element not being in the DOM yet?

mofux: Yes I tested it in Chrome, Firefox and Safari. Once terminal element is attached to the dom (which is one of the first things we do on term.open), this will deliver accurate measurements.

src/Terminal.ts
mofux: I see what you mean. Shall we open a separate issue for this to not block the v3 release on this?

mofux: viewport and helperContainer are now build using a `DocumentFragment` to avoid multiple layouts

src/addons/fit/fit.ts
```diff
@@ -26,27 +26,22 @@ export function proposeGeometry(term) {
   var availableHeight = parentElementHeight - elementPaddingVer;
   var availableWidth = parentElementWidth - elementPaddingHor;
   var geometry = {
-    cols: Math.floor(availableWidth / term.charMeasure.width),
-    rows: Math.floor(availableHeight / Math.floor(term.charMeasure.height * term.getOption('lineHeight')))
+    cols: Math.floor(availableWidth / (term.renderer.dimensions.actualCellWidth + term.getOption('letterSpacing'))),
```
Tyriar: Opps this is my bad, `letterSpacing` is already accounted for in `actualCellWidth`

PR #1169, opened 2017-12-27
Move selection event emission to the terminal object

npezza93:
Currently, the 'selection' is fired on the SelectionManager instance so to listen to the selection event you have to do `term.selectionManager.on('selection')`. Seems like this event should be emitted by the terminal instead. Feel free to close if I am incorrect.


npezza93:
Ahh whoops. Should be good now

PR #1171, opened 2017-12-29
Improved key mapping

ztmr:
After some experimenting, I ended up doing this. Is it something you'd like to use? Tests should cover the original code behaviour, only mapping is much easier and can be done as follows (may need some documentation later):
```
  term = new Terminal({
    debug: true,
    cursorBlink: true,
    cols: 80,
    rows: 24,
    scrollback: 2048,
    tabStopWidth: 10,
    fontSize: 18,
    keyMap: {
      /**
       * A mapping typical for DEC VT4xx/VT5xx friendly applications
       * Map value can be string, IKeyHandlerResult-compatible object or function (only for non-modified keys -- such a function would handle modifiers and other context specific behaviour)
       * Key format is case insensitive, uses '-' or '+' delimiters and the following are equivalents C=Ctrl=Control, A=Alt, M=Meta, S=Shift.
       */
      'Home'        : '\x1b[1~',   // Find
      'End'         : '\x1b[4~',   // Select
      'Esc'         : '\x1b[20~',
      'Insert'      : '\x1b[2~',
      'F1'          : '\x1b[28~',  // Help
      'F11'         : '\x1bOP',    // PF1
      'F12'         : '\x1b[4~',   // Select
      'C-F3'        : '\x1b[25~',  // F13
      'C-F4'        : '\x1b[26~',  // F14
      'C-F5'        : '\x1b[28~',  // F15
      'C-F6'        : '\x1b[29~',  // F16
      'C-F7'        : '\x1b[31~',  // F17
      'C-F8'        : '\x1b[32~',  // F18
      'C-F9'        : '\x1b[33~',  // F19
      'C-F10'       : '\x1b[34~',  // F20
      'KP_Divide'   : '\x1bOP',
      'KP_Multiply' : '\x1b[29~',
      'KP_Enter'    : '\x1bOM',
      'KP_Subtract' : '\x1bOm',
      'KP_Add'      : '\x1bOl',
      'KP_Decimal'  : '\x1bOn',
      'KP_0'        : '\x1bOp',
      'KP_1'        : '\x1bOq',
      'KP_2'        : '\x1bOr',
      'KP_3'        : '\x1bOs',
      'KP_4'        : '\x1bOt',
      'KP_5'        : '\x1bOu'
    }
  });
```

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -89,13 +89,7 @@ function createTerminal() {
   term = new Terminal({
     cursorBlink: optionElements.cursorBlink.checked,
     scrollback: parseInt(optionElements.scrollback.value, 10),
-    tabStopWidth: parseInt(optionElements.tabstopwidth.value, 10),
-    translations: {
-      // XXX: only these are supported now
-      Home  : '\x1b[1~',
-      End   : '\x1b[4~',
-      Esc   : '\x1b[20~'
-    }
+    tabStopWidth: parseInt(optionElements.tabstopwidth.value, 10)
   });
   window.term = term;  // Expose `term` to window for debugging purposes
   term.on('resize', function (size) {
--- a/demo/main.js
+++ b/demo/main.js
@@ -89,7 +89,13 @@ function createTerminal() {
   term = new Terminal({
     cursorBlink: optionElements.cursorBlink.checked,
     scrollback: parseInt(optionElements.scrollback.value, 10),
-    tabStopWidth: parseInt(optionElements.tabstopwidth.value, 10)
+    tabStopWidth: parseInt(optionElements.tabstopwidth.value, 10),
+    translations: {
+      // XXX: only these are supported now
+      Home  : '\x1b[1~',
+      End   : '\x1b[4~',
+      Esc   : '\x1b[20~'
+    }
   });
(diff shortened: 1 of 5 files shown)


Tyriar:
The issue tracking this is here which you can see a discussion on the topic https://github.com/xtermjs/xterm.js/issues/487

Recently I've become more of the mindset that we shouldn't put this into xterm.js as pretty much all consumers of the library are going to have their own keybindings system so in the end it's just going to be more code to load for little gain.

mofux:
Looking through this PR, I first have to say that I do appreciate the effort you have put into this. I actually quite like it, especially the fact that the key handling code has been sourced out to a separate file. On the other hand, I don't really see the use-case for this 😔. I would expect these bindings to managed by the hosting shell, and not by xterm.js as the client. Could you provide your use-case? I feel I might looking at it from the wrong perspective...

Tyriar:
I'm going to close this off to clean up the PR list, currently the recommended way to customize is by hooking into the key events via:

https://github.com/xtermjs/xterm.js/blob/404a285eaf5cb98471f17cf7fac6a97699431b1b/typings/xterm.d.ts#L386

PR #1172, opened 2017-12-29
Update charatlas when monitor DPI changes

### LINKED ISSUE #1118: Switching from a high DPI monitor to a low DPI monitor needs to refresh the texture atlas
VS Code issue: https://github.com/Microsoft/vscode/issues/34320

This is one of the bigger problems to come out of the canvas renderer. There is no event to listen to for monitor DPI changes so the solution will likely involve checking `window.devicePixelRatio` because using the char atlas, if it's changed then queue a refresh of the char atlas (this will suck until `OffscreenCanvas` is available) and revert to using uncached drawing.

image
image

Tyriar:
Fixes #1118

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -637,7 +637,11 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this.on('resize', () => this.renderer.onResize(this.cols, this.rows, false));
     this.on('blur', () => this.renderer.onBlur());
     this.on('focus', () => this.renderer.onFocus());
-    window.addEventListener('resize', () => this.renderer.onWindowResize(window.devicePixelRatio));
+    const rendererResizeListener = () => this.renderer.onWindowResize(window.devicePixelRatio);
+    // Listen for window zoom
+    window.addEventListener('resize', () => rendererResizeListener);
+    // Listen for monitor DPI change
+    window.matchMedia('screen and (-webkit-min-device-pixel-ratio: 1.5)').addListener(rendererResizeListener);
     this.charMeasure.on('charsizechanged', () => this.renderer.onResize(this.cols, this.rows, true));
     this.renderer.on('resize', (dimensions) => this.viewport.syncScrollArea());
 
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -637,11 +637,6 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this.on('resize', () => this.renderer.onResize(this.cols, this.rows, false));
     this.on('blur', () => this.renderer.onBlur());
     this.on('focus', () => this.renderer.onFocus());
-    const rendererResizeListener = () => this.renderer.onWindowResize(window.devicePixelRatio);
-    // Listen for window zoom
-    window.addEventListener('resize', () => rendererResizeListener);
-    // Listen for monitor DPI change
-    window.matchMedia('screen and (-webkit-min-device-pixel-ratio: 1.5)').addListener(rendererResizeListener);
     this.charMeasure.on('charsizechanged', () => this.renderer.onResize(this.cols, this.rows, true));
     this.renderer.on('resize', (dimensions) => this.viewport.syncScrollArea());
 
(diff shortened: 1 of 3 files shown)


src/Terminal.ts
```diff
@@ -637,7 +637,11 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this.on('resize', () => this.renderer.onResize(this.cols, this.rows, false));
     this.on('blur', () => this.renderer.onBlur());
     this.on('focus', () => this.renderer.onFocus());
-    window.addEventListener('resize', () => this.renderer.onWindowResize(window.devicePixelRatio));
+    const rendererResizeListener = () => this.renderer.onWindowResize(window.devicePixelRatio);
+    // Listen for window zoom
+    window.addEventListener('resize', () => rendererResizeListener);
+    // Listen for monitor DPI change
+    window.matchMedia('screen and (-webkit-min-device-pixel-ratio: 1.5)').addListener(rendererResizeListener);
```
mofux: Having a look at the `-webkit-min-device-pixel-ratio` media query, we should be able to replace it with the more standardised `min-resolution` clause as mentioned in the examples section here: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-webkit-device-pixel-ratio. I have not tested it yet, but I think it's worth to give it a try as it would also make it work for other browsers.

```
screen and (-webkit-min-device-pixel-ratio: 1.5)
vs.
screen and (min-resolution: 1.5dppx)
```
Btw, do you know if using any other value than 1.5 does make a difference?

Tyriar: I updated this to use a more robust solution that fixes the 1.5 problem 😃 

Can try `min`/`max-resolution` as well

src/utils/ScreenDprMonitor.ts
```diff
@@ -0,0 +1,62 @@
...
+    // Clear listeners for old DPR
+    if (this._minMediaMatchList) {
+      this._minMediaMatchList.removeListener(this._outerListener);
+    }
+    if (this._maxMediaMatchList) {
+      this._maxMediaMatchList.removeListener(this._outerListener);
+    }
+    // Add listeners for new DPR
+    this._currentDevicePixelRatio = window.devicePixelRatio;
+    this._minMediaMatchList = window.matchMedia(`screen and (-webkit-min-device-pixel-ratio: ${window.devicePixelRatio})`);
```
mofux: I've just tested using this code:

```javascript
window.matchMedia(`screen and (resolution: ${window.devicePixelRatio}dppx)`).addListener(() => {
  console.log('dpi changed')
});
```

It fires whenever that media query match becomes either true or false (good to test by changing the browser zoom level). This means we can listen for the `resolution` change without having to listen for `min` and `max` separately.

Tyriar: Way better! :smiley:

src/utils/ScreenDprMonitor.ts
```diff
@@ -0,0 +1,56 @@
...
+  private _listener: ScreenDprListener;
+  private _resolutionMediaMatchList: MediaQueryList;
+
+  public setListener(listener: ScreenDprListener): void {
+    if (this._listener) {
+      this.clearListener();
+    }
+    this._listener = listener;
+    this._outerListener = () => {
+      console.log('change!');
```
mofux: Look what I've found 😅

mofux:
In general, what do you think about initialising the dpr monitor from within the xterm constructor and then have it emit an event on the terminal instance whenever the screen size changes? The renderer would then listen for that event on the terminal instance instead. This would give the advantage that external addons could also listen for screen size changes without having to create a separate ScreenDprMonitor instance.

src/utils/ScreenDprMonitor.ts
Tyriar: That should be *changed!*

Tyriar:
I'll refactor as necessary when it's needed. The screen reader support PR needs to listen to this as well so I'll fiddle with it pretty fast https://github.com/xtermjs/xterm.js/pull/1182

Tyriar:
Pushing this in as tests pass locally and review requests are done. I need to build upon `ScreenDprMonitor` in other PRs so merging this in will make that much easier to manage.

PR #1180, opened 2018-01-02
Allow alt+backspace to delete a word

### LINKED ISSUE #1179: alt+backspace does not delete word
VS Code report: https://github.com/Microsoft/vscode/issues/40444

alt+backspace is meant to delete a word in bash. Note that it behaves differently to ctrl+w.

Tyriar:
Fixes #1179

---

Figured out the right thing to send by running this:

```
❯ bind -p | grep backward-kill-word
"\e\C-h": backward-kill-word
"\e\C-?": backward-kill-word
# shell-backward-kill-word (not bound)
```


Tyriar:
Doubt didn't check whether this one was signed off, let me know if there's a problem with the change and I'll fix it up

Issue #731, opened 2017-06-27
Support screen readers


Tyriar:
See https://github.com/Microsoft/vscode/issues/8339

jluk:
+1 to this! Is this on track for the August release?

Tyriar:
@jluk yep, haven't started yet though so not 100% on how long it will take.

Tyriar:
Reading the letter back when it's deleted from the prompt does not seem trivial due as the prompt line gets re-written when typing:

```
[?25l
C:\Users\daimms\Documents\dev\Tyriar\xterm.js>l[0K
[48G[?25h
[?25l
C:\Users\daimms\Documents\dev\Tyriar\xterm.js>ls[0K
[49G[?25h
[?25l
C:\Users\daimms\Documents\dev\Tyriar\xterm.js>l[0K
[48G[?25h
```

Tyriar:
The following work in NVDA:

- Hovering a line to read it out
- Typing reading out the letter that was added

The following does not work:

- Backspace reading out the letter that was deleted
- Moving left/right in the prompt reading the letter on the cursor
- Running a command reading the output (nothing happens)

Tyriar:
Fixed in 1182

SiulSalas:
Mouse hovering on the Visual Studio Code terminal stopped working with NVDA

Tyriar:
@SiulSalas you're better off creating an issue in the vscode repo

PR #1182, opened 2018-01-03
Screen reader support

Tyriar:
FYI @jluk

This PR adds support for screen readers (targeting NVDA on Windows initially). Screen reader support is disabled by default and must be opt in (`screenReaderMode` setting), this is so we can avoid all the DOM manipulation when we don't need it for performance reasons.

### Currently supported

- Terminal is announced when focused
- Characters entered into the terminal via keyboard are announced
  - Character and word echo settings are respected
- Read command output upon execution (up to 20 rows to prevent DOM from growing)
- Support screen reader navigation within current viewport
  - NVDA can read out row content on mouse over
  - NVDA/VoiceOver navigation mode works with rows

### Known issues

- Characters are not read out when deleted from the prompt
- Reading out history when navigating shells using up/down sometimes doesn't work
- Holding down up/down while navigating rows in NVDA can skip over the terminal

Part of #731


camlorn:
Does this respect my keyboard echo settings?  If not, that's amazingly annoying.

I mention it because there is an anti-pattern in i.e. google docs where you get the screen readers to read via echoing to a live region, and this both doesn't respect settings and doesn't work out with punctuation (some is silent because it's trying to read it as words).

As someone who very intentionally doesn't use character or word echo, being forced into it would be suboptimal.  It's distracting and not very useful when i.e. on conference calls or whatever, because either it's pulling you away from whatever else is important or everyone else has to hear it if you're unlucky enough not to be able to have headphones.

Tyriar:
@camlorn character echo settings are respected, word echo does not work currently but it could potentially later on (I made a note in the PR).

The live region is used to read out the command output, if you have anything I need to watch out for here I'd love to hear it. The live region seemed to handle perfectly what I wanted with the web APIs I have access to; to allow reading out of a stream of text (that may be incomplete) and stop the screen reader upon typing or when there is too much output.

Tyriar:
I was planning on allowing calls to scrollPage/scrollToTop/scrollToBottom to scroll and focus in navigation mode, unfortunately screen readers take complete control of the keyboard in these modes and I cannot intercept the needed keyboard strokes.

In case it's useful later, here's the tougher bit of code that did the scrolling and focusing:

```ts
  /**
   * Moves the focus of the terminal relatively by a number of rows.
   * @param amount The amount of rows to scroll
   */
  public moveRowFocus(amount: number): void {
    console.log('moveRowFocus', amount);
    // Do nothing if not navigating rows or the amount to navigate is 0
    if (!this.isNavigatingRows || amount === 0) {
      return;
    }

    // Find and validate the new absolute row position
    const buffer = this._terminal.buffer;
    const oldRelativeRow = this._rowElements.indexOf(<HTMLElement>document.activeElement);
    const oldAbsoluteRow = buffer.ydisp + oldRelativeRow;
    const newAbsoluteRow = Math.max(Math.min(oldAbsoluteRow + amount, buffer.lines.length - 1), 0);
    if (oldAbsoluteRow === newAbsoluteRow) {
      return;
    }

    // Find the new relative row, this cannot be on a boundary element unless the focused row is the
    // top-most or bottom-most row in the buffer.
    let newRelativeRow: number;
    if (newAbsoluteRow === 0) {
      newRelativeRow = 0;
    } else if (newAbsoluteRow === buffer.lines.length - 1) {
      newRelativeRow = this._terminal.rows - 1;
    } else {
      newRelativeRow = Math.max(Math.min(oldRelativeRow + amount, this._terminal.rows - 2), 1);
    }

    // Find the new viewport position
    let newYDisp: number;
    if (newAbsoluteRow === 0) {
      // Start of buffer
      newYDisp = 0;
    } else if (newAbsoluteRow === buffer.lines.length - 1) {
      // End of buffer
      newYDisp = buffer.lines.length - 1 - this._terminal.rows;
    } else if (newAbsoluteRow > buffer.ydisp && newAbsoluteRow < buffer.ydisp + this._terminal.rows - 1) {
      // No scrolling necessary
      newYDisp = buffer.ydisp;
    } else if (newAbsoluteRow < oldAbsoluteRow) {
      // Scrolling up needed
      newYDisp = newAbsoluteRow - 1;
    } else { // if (newAbsoluteRow > oldAbsoluteRow)
      // Scrolling down needed
      newYDisp = newAbsoluteRow + this._terminal.rows - 2;
    }

    // Perform scroll
    const scrollAmount = newYDisp - buffer.ydisp;
    this._terminal.scrollLines(scrollAmount);

    // Focus new row
    this._rowElements[newRelativeRow].focus();
  }
```

Tyriar:
This is ready for review

Tyriar:
Pushing to v3.2.0 as this is a big PR and we have a bunch out for review in v3.1.0.

PR #1191, opened 2018-01-05
Add Kubebox to real-world uses

astefanutti:
Online version using Xterm.js is deployed at http://kube.sh.

Thanks for your awesome project!


Tyriar:
Cool 😃

PR #1197, opened 2018-01-06
Ship the declaration files

### LINKED ISSUE #1196: Add typings for addons?
We currently have to separately declare local typings for addons, e.g.

```TypeScript
declare module 'xterm/lib/addons/fit' {
    export function fit(term: any): void;
}
```

However, if the above were included in `typings/term.d.ts`, we would not need the separate declaration. I verified this locally.

## Details
- Browser and browser version: Firefox 57
- OS version: macOS 10.13
- xterm.js version: 3.0.1

blink1073:
Fixes #1196.

I verified locally that the following works from a TypeScript consumer using the `.d.ts` files produced by this PR:

```javascript
import {
  fit
} from 'xterm/lib/addons/fit';
```

cf https://github.com/jupyterlab/jupyterlab/pull/3553 if you're interested in the JupyterLab migration from 2.x -> 3.x

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -51,9 +51,14 @@ gulp.task('tsc', function () {
 
     let tsProjectAddon = ts.createProject(`./src/addons/${addon}/tsconfig.json`);
     let tsResultAddon = tsProjectAddon.src().pipe(sourcemaps.init()).pipe(tsProjectAddon());
-    let tscAddon = tsResultAddon.js
-      .pipe(sourcemaps.write('.', {includeContent: false, sourceRoot: ''}))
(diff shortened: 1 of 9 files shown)


blink1073:
Given this PR, I think it makes sense to suggest this pattern: using the addon functions directly, passing in the terminal instance.   Using `apply` would require manual handling of the `addon` typings by either the `xterm` library authors or consumers.

Tyriar:
Looks good, created https://github.com/xtermjs/xterm.js/issues/1219 to follow up shipping hand crafted typings https://github.com/xtermjs/xterm.js/issues/1219

Issue #1187, opened 2018-01-05
v3: is requirejs still supported?


ficristo:
I'm only looking inside the addon folder that inside build and lib folders.
No addons has an umd wrapper.

## Details
- Browser and browser version: -
- OS version: Windows 10
- xterm.js version: v3

### Steps to reproduce

1. Just look at build/addons/fit.js for example

Tyriar:
See https://github.com/xtermjs/xterm.js/pull/1130 for the changes/discussion. @parisk can confirm

ficristo:
I don't see any mention about requirejs... is there?
It seems to me that PR talks only about modules and commonjs way to import files.

ficristo:
It seems to me that also the browser usage for the addons is broken.

FGasper:
Please LMK if there’s anything I can do to help resolve this more quickly, as it’s a direct impediment to adoption of v3 for our team.

tsl0922:
It seems the [zmodem addon demo](https://github.com/xtermjs/xterm.js/tree/master/src/addons/zmodem/demo) is broken with errors like https://github.com/xtermjs/xterm.js/pull/1130#pullrequestreview-78576803.

jpmasters:
We're having the same problem. We're using the control inside an Angular 1.6 component so we include the fit addon from \dist in an html script tag. This no longer works because the UMD header is missing. I've made a temp workaround by creating my own version of the fit addon but obviously it would be great to be able to use the real thing.

parisk:
Just got on top of this. This should be fixed after https://github.com/xtermjs/xterm.js/pull/1199, since add-ons are compiled in UMD format. I think it's a good idea to publish another bugfix release for this.

Issue #881, opened 2017-08-12
Require search addon


ficristo:
I use npm to install my deps also for frontend and requirejs to include them.
When I include addons like fit or attach the following works, but when I try to require the search addon, it seems to work but the `findNext` and `findPrevious` aren't available.
I tryed to use `Terminal.loadAddon` but it doesn't work with any addon.

```js
define(function (require, exports, module) {
    var Terminal = require("node_modules/xterm/dist/xterm"),

    require([
        "node_modules/xterm/dist/addons/fit/fit",
        "node_modules/xterm/dist/addons/search/search"
    ]);

    var term = new Terminal({});
});
```

## Details
- Browser and browser version: I'm using it in a Brackets (a code editor) extension
- OS version: Windows
- xterm.js version: 2.9.2

### Steps to reproduce

1. Use npm
2. Use requirejs
3. Include search addon

parisk:
Hi @ficristo, can you provide any logs from your browser's developer tools console (you can toggle it by hitting <kbd>F12</kbd>).

Tyriar:
I'm assuming I broke requirejs on addons as I didn't test when changing it, @parisk are you familiar with requirejs?

parisk:
Yep, I am. I can take a look next week though.

ficristo:
In the developer tools integrated in Brackets there isn't any error.

If I try to load an existing addon I get:
```js
Failed to load resource: net::ERR_FILE_NOT_FOUND
require.js:141 Uncaught Error: Script error for: node_modules/xterm/dist/addons/foo/foo
http://requirejs.org/docs/errors.html#scripterror
```

Tyriar:
@ficristo does it really say foo or did you replace that?

ficristo:
Sorry, I meant a **non** exisisting addon....

ficristo:
With https://github.com/sourcelair/xterm.js/pull/926 now I see `Uncaught Error: Module name "../../xterm" has not been loaded yet for context: _. Use require([])`

Tyriar:
Maybe a revision of how addons works as suggested in #927 would be required. It would certainly simplify things if there wasn't a cyclic dependency.

Tyriar:
I'm consolidating all bundling/loadAddon related issues into https://github.com/sourcelair/xterm.js/issues/1018, please comment on that issue if you have thoughts :smiley:

PR #1199, opened 2018-01-07
Make addons UMD compatible

ficristo:
This is a possible fix for https://github.com/xtermjs/xterm.js/issues/1187
In the following examples the logs print the addon functions.

Requirejs usage:
```js
define(function (require, module, exports) {
    var xterm = require("node_modules/xterm/dist/xterm");
    var fit = require("node_modules/xterm/dist/addons/fit/fit");
    var search = require("node_modules/xterm/dist/addons/search/search");

    console.log(fit);
    console.log(search);

    xterm.applyAddon(fit);
    xterm.applyAddon(search);
    var term = new xterm();

    console.log(term.fit);
    console.log(term.findNext);
});

```

Browser usage
```html
<!doctype html>
<html>
  <head>
    <link rel="stylesheet" href="node_modules/xterm/dist/xterm.css" />
    <script src="node_modules/xterm/dist/xterm.js"></script>
    <script src="node_modules/xterm/dist/addons/fit/fit.js"></script>
  </head>
  <body>
    <div id="terminal"></div>
    <script>
      Terminal.applyAddon(fit);
      var term = new Terminal();

      console.log(term.fit);
    </script>
  </body>
</html>
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -18,10 +18,12 @@ const ts = require('gulp-typescript');
 const util = require('gulp-util');
 const webpack = require('webpack-stream');
 
-let buildDir = process.env.BUILD_DIR || 'build';
-let tsProject = ts.createProject('tsconfig.json');
-let srcDir = tsProject.config.compilerOptions.rootDir;
+const buildDir = process.env.BUILD_DIR || 'build';
+const tsProject = ts.createProject('tsconfig.json');
+const srcDir = tsProject.config.compilerOptions.rootDir;
 let outDir = tsProject.config.compilerOptions.outDir;
+
+const addons = ['attach', 'fit', 'fullscreen', 'search', 'terminado', 'winptyCompat', 'zmodem'];
 
 // Under some environments like TravisCI, this comes out at absolute which can
 // break the build. This ensures that the outDir is absolute.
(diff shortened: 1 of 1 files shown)


mofux:
Works great for me, thanks! The code is also more generic and cleaner than before. @parisk Any thoughts?

Tyriar:
@ficristo was search and winptyCompat working fine before and this PR just makes all other addons build like them?

ficristo:
The difference now is all addons provide a global variable for browser usage (see the `standalone` field) which is necessary to create an UMD wrapper.
@Tyriar A part this and the copy code, yes the code is almost equal.

search [didn't work](https://github.com/xtermjs/xterm.js/issues/881) with requirejs in 2.9.2 (and winptyCompat didn't exist)

I don't think any addons works with requirejs or in the browser since 3.0.0.
They are browserified but the result is not copied in the build folder, it is instead overwritten with what there is in the lib folder. https://github.com/xtermjs/xterm.js/blob/be18507df64d6f79113d216572921ddb912b8196/gulpfile.js#L61-L64

I'm not sure if useful for you, but you can see the result without whitespace changes:
https://github.com/xtermjs/xterm.js/pull/1199/files?w=1

Issue #824, opened 2017-07-28
this.buffer.lines.get(i) returns undefined


Tyriar:
Original issue: https://github.com/sourcelair/xterm.js/issues/821
PR for quick fix: https://github.com/sourcelair/xterm.js/pull/823

> The resize rows section below is meant to handle this row resizing, if this is actually happening we should figure out the root cause. `this.buffer.lines.get(i)` should always return an array when `i < this.buffer.lines.length` which is guaranteed by line 1942.

@jpmasters any more details on how you're using xterm.js?

jpmasters:
Hi @Tyriar I can reproduce it in the demo using the following after refreshing the demo page:

```
// create a scroll region from lines 13 to 19
term.write('\x1b[13;19r');
// move the cursor to the top of the scroll region
term.write('\x1b[13;0H');
// RI
term.write('\x1bM');
// now the last entries are undefined
console.log(term.buffer.lines.length);
console.log(term.buffer.lines[term.buffer.lines.length - 5]);
```
If you now call term.resize() with a rows value > the current rows it'll throw a type error if the terminal has had the chance to render in the meantime so try the resize() as a new command.

```
term.resize(50, 50);
```
The key to reproducing it seems to be to get (Terminal.buffer.y === Terminal.buffer.scrollTop) &&((Terminal.buffer.y +  Terminal.buffer.ybase) > 0) then calling reverseIndex. We managed to do that by using a scroll region but I don't know if that's the only way it can happen.  The code as it is at the moment assumes a RI will only need to insert a new line when buffer.y === 0.

Does that make sense?

parisk:
I confirm that this is happening on SourceLair as well. Here is a public Sentry event:

https://sentry.io/share/issue/35323433302e333139343137383232/

amejia1:
Tracked this down to a problem with the 'buffer' property [here](https://github.com/xtermjs/xterm.js/blob/be18507df64d6f79113d216572921ddb912b8196/src/Terminal.ts#L296). Seems now with the latest version of xterm.js, this keeps whatever buffer value was set to it initially or on activate. As I resize the terminal, I can see the buffers in the bufferset get updated, but it doesn't seem to get reflected in this property.

Changing this property to be implemented through a getter resolves the resizing issue for me but now I get a problem with selecting content to be copied from the terminal. In any case, this is the diff I have so far.

```
diff --git a/src/Buffer.ts b/src/Buffer.ts
index 623c385..3e7d612 100644
[...]

PR #1206, opened 2018-01-13
Fix issue with resizing terminal.

amejia1:
Need to ensure the active buffer is always being used.
Fixes https://github.com/xtermjs/xterm.js/issues/824 .

### DIFF
--- a/src/SelectionManager.ts
+++ b/src/SelectionManager.ts
@@ -119,11 +119,27 @@ export class SelectionManager extends EventEmitter implements ISelectionManager 
     this._mouseMoveListener = event => this._onMouseMove(<MouseEvent>event);
     this._mouseUpListener = event => this._onMouseUp(<MouseEvent>event);
 
-    // Only adjust the selection on trim, shiftElements is rarely used (only in
-    // reverseIndex) and delete in a splice is only ever used when the same
-    // number of elements was just added. Given this is could actually be
-    // beneficial to leave the selection as is for these cases.
-    this._buffer.lines.on('trim', (amount: number) => this._onTrim(amount));
+    this._updateBufferSetHandlers();
+    this._terminal.on('setup', () => {
+        this._onSetupHandler();
+    });
+  }
+
+  private _onSetupHandler(): void {
+      this.clearSelection();
+      this._updateBufferSetHandlers();
+  }
+
+  private _updateBufferSetHandlers(): void {
+      this._terminal.buffers.on('activate', (buffer: IBuffer) => {
+          this.clearSelection();
+      });
+      // Only adjust the selection on trim, shiftElements is rarely used (only in
+      // reverseIndex) and delete in a splice is only ever used when the same
+      // number of elements was just added. Given this is could actually be
+      // beneficial to leave the selection as is for these cases.
+      this._terminal.buffers.normal.lines.on('trim', (amount: number) => this._onTrim(amount));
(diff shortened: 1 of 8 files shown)


amejia1:
As far as I know, the change here will fix the issue with resizing the terminal as well as ensure no other problems arise and continues to pass the test suite. Works for me on Windows and Linux.

src/SelectionManager.ts
```diff
@@ -139,16 +142,6 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
...
   }
 
-  /**
-   * Sets the active buffer, this should be called when the alt buffer is
-   * switched in or out.
-   * @param buffer The active buffer.
-   */
-  public setBuffer(buffer: IBuffer): void {
-    this._buffer = buffer;
-    this.clearSelection();
```
Tyriar: It looks like selection isn't going to be cleared when the terminal buffer switches, which feels awkward as random text is then selected.

amejia1: Alright, added a new commit to fix this.

Tyriar:
I made a few changes in https://github.com/xtermjs/xterm.js/pull/1206/commits/b5aff0be7e9b7af13c3488b0c6f05e54a9c5c24b which I think are correct, @amejia1 please double check they're good then we can push this in.

amejia1:
New changes fail the test suite. Just check on my Windows box. The Travis CI tests are already failing too.

amejia1:
So yes, the old BufferSet can't be kept when resetting the terminal, otherwise the test suite fails and it doesn't actually clear anything when actually resetting a terminal. Also, since the BufferSet has to be changed, the selection manager has to be cleared, otherwise it throws errors about not being able to read properties of `undefined`.

src/Terminal.ts
```diff
@@ -292,7 +292,6 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
 
     // Create the terminal's buffers and set the current buffer
     this.buffers = new BufferSet(this);
-    this.emit('setup');
```
amejia1: There's going to be a need for some kind of notification when the BufferSet changes. Best way I could think of was to use the event emitter. Should I reimplement this? Maybe indicate it as a private event with a name like '_changebufferset'?

src/Terminal.ts
```diff
@@ -2073,9 +2072,11 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this.options.cols = this.cols;
     const customKeyEventHandler = this.customKeyEventHandler;
     const inputHandler = this.inputHandler;
+    const buffers = this.buffers;
```
amejia1: Was trying to mention this earlier. This causes the test suite to fail.

Tyriar:
> Ah. This causes the test suite to fail. The old BufferSet can't be kept otherwise resetting the terminal won't do the right thing.

@amejia1 I was a little confused by this, do you know why `term.reset()` works (clears the buffer) when the `BufferSet` is being reset? I think @parisk did that part.

screen shot 2018-01-16 at 12 07 41 pm

> Should this clear selection when the terminal resets?

Oh it wasn't clear that's what this was doing. Maybe for now let's just have a specific call to SelectionManager in reset?

amejia1:
@Tyriar Ok, so clearing the selection manager on reset turns out to be much cleaner. This passes the test suite and I ensured that resetting the terminal does clear the terminal and also any pending selection.

Tyriar:
Thanks heaps @amejia1, sorry about the delay 😃

PR #1214, opened 2018-01-17
Fix lint, update tslint and add more tslint rules

Tyriar:
Also includes package-lock.json which is generated by npm 5 and the recommendation is to commit it (so builds are more reproducible when deps update)


Tyriar:
Merging this as nothing changed functionally and it's blocking further work and the build on some platforms.

PR #1215, opened 2018-01-18
Add support for modifiers for PageUp / PageDown keys.

markus-oberhumer:
This allows using Ctrl-PageUp/Down and Alt-PageUp/Down.

[ Originally submitted as Tyriar/xterm.js#2 ]


Tyriar:
Thanks 😄

PR #1217, opened 2018-01-18
Fix cursor lagging behind when blink is on

### LINKED ISSUE #1216: Cursor with blink turned on is lagging behind
## Details
- Browser and browser version: `Electron 1.7.10` (chrome)
- OS version: macOS `10.13.2 (17C88)`
- xterm.js version: `3.0.2`

### Steps to reproduce

1. Open a new terminal
2. Press enter a couple times
3. Press `CMD+K` (clear) when the cursor appears
4. The cursor will not move into position until after it goes black and comes back again

Demo:

kapture 2018-01-18 at 13 19 35

I think that even tough blink is active, the cursor should still move like a non-blinking cursor. Maybe even reset the blink states to "SOLID" when the cursor moves?

Tyriar:
Fixes #1216

### DIFF
--- a/src/renderer/CursorRenderLayer.ts
+++ b/src/renderer/CursorRenderLayer.ts
@@ -108,9 +108,10 @@ export class CursorRenderLayer extends BaseRenderLayer {
   }
 
   public onGridChanged(terminal: ITerminal, startRow: number, endRow: number): void {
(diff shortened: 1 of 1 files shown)


mofux:
LGTM 👍

PR #1218, opened 2018-01-18
Remove OSX from travis

Tyriar:
Builds take too long on free tier


mofux:
LGTM 👍. It slows us down more than it helps.

PR #1223, opened 2018-01-19
Add rtty to the real-world uses section of README

zhaojh329:
A reverse proxy WebTTY. It is composed of the client and the server.

Signed-off-by: Jianhui Zhao <jianhuizhao329@gmail.com>


Tyriar:
Thanks!

Issue #1221, opened 2018-01-18
Share Terminal interface between core and addons


Tyriar:
Both core and the addons should be able to point at the .d.ts file that holds our public API. We can also base our internal ITerminal interface off the `Terminal` class declaration in xterm.d.ts.

This is blocked on https://github.com/xtermjs/xterm.js/issues/1220

Tyriar:
Already done

PR #1224, opened 2018-01-19
Move ITerminal and all its dependents into typings/

Tyriar:
This PR essentially does the following:

- Has `ITerminal` extend the public API for `Terminal` defined in `./typings/xterm.d.ts`
- Pulls `ITerminal` from `./src/Interfaces.ts` into `./typings/xterm-internal.d.ts`

What this will allow us to do (after https://github.com/xtermjs/xterm.js/pull/1222) is share the `ITerminal` interface with addons. This should make working on addons much better as there will be reliable type checking for properties on `Terminal` that are relied on like `buffer` and `selectionManager`. This also means that if private APIs change, the build will break.

Part of #1221


Tyriar:
I've had a while to brew on this change and I'm going to make the following additional changes:

- Merge `Interfaces.ts` and `Types.ts` into Types.**d**.ts, this will enforce the fact that we don't want any TypeScript code to be used here
- Move the `xterm-internal.d.ts` types back into the corresponding Types.d.ts
- Add a note in the README that importing from `Types.d.ts` should only be done at a higher level or from addons. What I mean here is that `src/renderer/Types.d.ts` should only be accessed by files within `src/renderer` (or addons). If `src/blah.ts` needs to access it, the type should be moved to `src/Types.d.ts`

These changes will also enable addons to import whatever (because they're .d.ts files) while still keeping the nice encapsulation of the types and getting rid of the unnecessary distinction between interfaces and types.

PR #1225, opened 2018-01-19
Mac OSX: altIsMeta support

saamalik:
* Support for altIsMeta option
* Instead of providing third-level shift characters
* Sample demo application using new altIsMeta option

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -24,6 +24,9 @@
             <label><input type="checkbox" id="option-cursor-blink"> cursorBlink</label>
           </p>
           <p>
+            <label><input type="checkbox" id="option-mac-option-is-meta"> macOptionIsMeta</label>
+          </p>
+          <p>
             <label>
               cursorStyle
               <select id="option-cursor-style">
--- a/demo/main.js
+++ b/demo/main.js
@@ -27,6 +27,7 @@ var terminalContainer = document.getElementById('terminal-container'),
     optionElements = {
       cursorBlink: document.querySelector('#option-cursor-blink'),
       cursorStyle: document.querySelector('#option-cursor-style'),
+      macOptionIsMeta: document.querySelector('#option-mac-option-is-meta'),
       scrollback: document.querySelector('#option-scrollback'),
       tabstopwidth: document.querySelector('#option-tabstopwidth'),
       bellStyle: document.querySelector('#option-bell-style')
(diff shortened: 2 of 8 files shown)


saamalik:
Fixes #487

saamalik:
@Tyriar let me know if you'd like me to make any additional changes.

Tyriar:
@saamalik thanks for the PR 😃. What are some good keybindings for me to play around with to test this out?

saamalik:
The common ones I use are: 

```
alt+f  move forward a word
alt+b  move backwards
alt+.  yank last argument
```

A whole list of useful functions are available here: https://www.skorks.com/2009/09/bash-shortcuts-for-maximum-productivity/

Tyriar:
Thanks @saamalik, works great! I did a bit of clean up and this is good to go after CI passes. Thanks for looking into it 😃

saamalik:
@Tyriar cool man! cleanup looks great too!

timjs:
Is it possible to have a similar option for the editor itself?

An extension like [Dance](https://github.com/71/dance) to emulate Kakoune modal editing makes extensively use of meta mappings. For example, `meta+e` is equivalent to Vim's `E` "move to end of word including punctuation".

On macOS typing `option+e` enters a dead key mode to enter characters like é, á etc. and therefore is not usable as a shortcut for this motion. Using Kakoune in the terminal is no problem, because of the "Use Option-key as meta-key" in the Terminal settings.

Tyriar:
@timjs you mean in VS Code? option+<letter> on mac is an OS-level feature that maybe you can turn off?

timjs:
Sorry @Tyriar, I just overlooked and thought this is an issue on the VSCode repo, but it isn't. Think I'll have to find or create an issue in the right repo 😝 

Sadly you can't turn it of OS wide, only by hand crafting new keyboard layouts. Only [the macOS Terminal app has a setting to do this](http://osxdaily.com/2013/02/01/use-option-as-meta-key-in-mac-os-x-terminal/).

Tyriar:
@timjs doesn't using alt in keybindings override it though?

timjs:
I’m afraid not. In VSCode you're simply not able to bind anything to option+e/u/i/n because it produces a dead character. I.e. to type é in any text field on macOS, you first hit option+e which gives you the ´, and then e. The OS merges the two into é.

Tyriar:
This keybinding correctly focused the terminal when the editor is focused for me? It ends up sending adding a ` char to the terminal which I think is already captured in some vscode issue.

```
{"key": "alt+e", "command": "workbench.action.terminal.toggleTerminal" }
```

timjs:
It indeed works...

In the [Dance extension](https://github.com/71/dance) @71 is working on, then `when` clause contains `editorTextFocus`. Removing it does the trick! 

Thanks for sharing your ideas @Tyriar! Helped to find a solution!

Issue #145, opened 2016-06-24
broken escape sequence parser states


jerch:
Please have a look at http://www.vt100.net/emu/dec_ansi_parser. In the state diagram there you can see, that "execute" characters (like '\n', '\r' ...) don't break or reset the actual parser state - they will just get executed for most of the possible states.
Problem is that some sequence parser states allow nesting other escape sequences while others dont. That is not reflected by the actual state handling in xterm.js
This is partly tested by vttest under "1. Test of cursor movements" (press 5 times 'enter')

IMHO this is a major issue, since it affects the flow/state logic in `write` which is kinda the terminal core.

jerch:
You can test this with: `echo -e '###\x1b[\nC###'`

expected result:
```
###
 ###
```

xterm.js result:
```
###
   C###
```
The cursor movement sequence is cancelled and the 'C' is treated as printable. There seems to be another problem since the '\n' doesnt move the cursor to position 1 on the next line (have not investigated this).

parisk:
@jerch thanks for reporting this! Could you please mention a use case where this bug will come up and what will be the outcome for the end user?

parisk:
Disregard my previous comment.

thomasjm:
I think I have an example of this affecting a real user... I'm using zsh in xterm.js. On a blank line, I type `emacs` and then hit tab to trigger tab completion. The backend responds with this on the websocket:

```
[0memacs                emacs24-x            emacsclient.emacs24
[Jemacs24              [Jemacsclient          [J                   [A[A[m[m[m
[7Cemacs[K
```

When xterm.js renders this, the string at the cursor instantly turns into `ememacs`, and if I backspace 5 times I can't delete the `em` -- hitting backspace a lot makes the cursor stay like this: `prompt > em` 

Seems like broken escape sequence parsing, apologies in advance if it's actually a separate issue.

Rodeoclash:
I'm seeing the problem with parsing `\n` characters as well. While the `writeLn` function works great, new lines move the cursor to the next row but the same column. Here's an example of rendering the output from a WebPack build to the console:

![screen shot 2017-03-12 at 5 54 01 pm](https://cloud.githubusercontent.com/assets/63807/23829694/ed574a20-074c-11e7-8ba0-19b029cd5e9b.png)

jerch:
[...]

PR #1226, opened 2018-01-20
Solution to issue #145: broken escape sequence parser states

yaoyu-wang:
There were some problems in ESCAPED and CSI state, as the issue mentioned, \n (or other normal state characters) should not make the ESCAPED state quit. 

Just add some normalStateHandler judgements can solve this problem.


src/Parser.ts
```diff
@@ -236,6 +236,10 @@ export class Parser {
           }
           break;
         case ParserState.ESCAPED:
+	  if (ch in normalStateHandler) {
```
Tyriar: This looks bad, the normal state handler is only meant to handle things when the state is `ParserState.NORMAL` by design, same with the CSI handler. I think a better approach would be finding the "problem" characters and reverting back to the normal state in this case. We would need to understand how a bunch of other terminals behave with these character(s) though in order to ensure we're behaving as expected.

Tyriar:
I'll close this off as it can't land in its current form, feel free to put out another PR after exploring another approach as indicated in https://github.com/xtermjs/xterm.js/pull/1226#discussion_r163947965

PR #1234, opened 2018-01-22
Use explicit null check to ensure start/end row is valid

### LINKED ISSUE #1233: Links are sometimes not activated until scroll
https://github.com/Microsoft/vscode/issues/36072

Tyriar:
It looks like this could have caused the inconsistent behavior with
links not activating and then working again after scrolling

Fixes #1233

### DIFF
--- a/src/Linkifier.ts
+++ b/src/Linkifier.ts
@@ -82,12 +82,12 @@ export class Linkifier extends EventEmitter implements ILinkifier {
     }
 
     // Increase range to linkify
-    if (!this._rowsToLinkify.start) {
+    if (this._rowsToLinkify.start === null) {
       this._rowsToLinkify.start = start;
       this._rowsToLinkify.end = end;
     } else {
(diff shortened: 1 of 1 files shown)


src/Linkifier.ts
```diff
@@ -82,12 +82,12 @@ export class Linkifier extends EventEmitter implements ILinkifier {
     }
 
     // Increase range to linkify
-    if (!this._rowsToLinkify.start) {
+    if (this._rowsToLinkify.start === null) {
```
Tyriar: This is the only line that functionally changed, before this condition would fire when `this._rowsToLinkify.start` was 0, meaning it could have been linkifying all rows:

```
this._rowsToLinkify.start = 0
this._rowsToLinkify.end = term rows - 1
```

And then the next call to the function would override it, which means only one row would be checked if `start === end`.

mofux: 👍 Good catch!

PR #1236, opened 2018-01-23
 Add allowTransparency option

### LINKED ISSUE #1004: Support transparent backgrounds

chabou:
Fixes #1004 (and replaces #1148)

I didn't notice any performance loss (measuring frame rendering duration for vtop on electron).


src/renderer/CharAtlas.ts
```diff
@@ -116,19 +116,19 @@ let generator: CharAtlasGenerator;
  * Initializes the char atlas generator.
  * @param document The document.
  */
-export function initialize(document: Document): void {
+export function initialize(document: Document, alpha: boolean): void {
```
Tyriar: Looks like `setOption('allowTransparency', true)` will not work when set after the page's first `new Terminal()` call. We need a handler in `setOption` and for `CharAtlas.ts` to regenerate the canvas when `alpha` changes.

src/renderer/TextRenderLayer.ts
```diff
@@ -24,8 +24,8 @@ export class TextRenderLayer extends BaseRenderLayer {
   private _characterFont: string;
   private _characterOverlapCache: { [key: string]: boolean } = {};
 
-  constructor(container: HTMLElement, zIndex: number, colors: IColorSet) {
-    super(container, 'text', zIndex, false, colors);
+  constructor(container: HTMLElement, zIndex: number, colors: IColorSet, alpha: boolean) {
+    super(container, 'text', zIndex, alpha, colors);
```
Tyriar: Same here, changing after the first `new Terminal()`will not affect `TextRenderLayer`

typings/xterm.d.ts
```diff
@@ -409,7 +415,7 @@ declare module 'xterm' {
      * Retrieves an option's value from the terminal.
      * @param key The option key.
      */
-    getOption(key: 'cancelEvents' | 'convertEol' | 'cursorBlink' | 'debug' | 'disableStdin' | 'enableBold' | 'popOnBell' | 'screenKeys' | 'useFlowControl' | 'visualBell'): boolean;
+    getOption(key: 'cancelEvents' | 'convertEol' | 'cursorBlink' | 'debug' | 'disableStdin' | 'enableBold' | 'popOnBell' | 'screenKeys' | 'useFlowControl' | 'visualBell' | 'allowTransparency'): boolean;
```
Tyriar: Let's put the new option at the start

src/renderer/CharAtlas.ts
chabou: Will do that

Tyriar:
@chabou I resolved the conflicts, I also fixed the canvas alpha updating after the option changed when *requesting* the char atlas as there was some refactoring done that made that trivial, as well as the typing ordering.

Changes still needed:

- Make sure `setOption('allowTransparency', ...)` requests a new char atlas and updates the `TextRenderLayer` canvas properly

chabou:
@Tyriar My last commit https://github.com/xtermjs/xterm.js/pull/1236/commits/fe53e1a09136828cb32faab5326d48ec11456c1f clones canvas to obtain a new context with desired transparency capability.
This change needs to be done in BaseRenderer class (to access some private properties/methods), but only TextRenderer need to recreate its canvas and regenerate its char atlas.

I am not sure that is the best way to handle this and I'll be happy to update it to suit your needs.

Tyriar:
I made a few minor changes and it looks good. I removed the call to `this._ctx.scale` as it was breaking the scale of the terminal after `setOption('allowTransparency', true)` was called. I'm not sure it was doing anything anyway.

Tyriar:
🎉

src/renderer/BaseRenderLayer.ts
```diff
@@ -54,23 +57,22 @@ export abstract class BaseRenderLayer implements IRenderLayer {
...
       return;
     }
-    this._alpha = alpha
-    const oldCanvas = this._canvas
-    // Cloning preserves canvas properties and permits to acquire a new context 
+
+    // Create new canvas and replace old one
+    const oldCanvas = this._canvas;
+    this._alpha = alpha;
+    // Closing preserves properties
```
chabou: Litlle typo here: Closing -> Cloning

Tyriar: Whoops 😛

Tyriar: Fixed https://github.com/xtermjs/xterm.js/pull/1260

PR #1241, opened 2018-01-23
Don't publish this to npm

torn4dom4n:


Tyriar:
Most of these files aren't actually included due to the `files` property in package.json

screen shot 2018-01-23 at 2 29 50 pm

We can probably remove `.npmignore` all together now.

torn4dom4n:
@Tyriar I think we don'nt need glupfile.js, jsdoc.json, tsconfig.json...

torn4dom4n:
BTW, do you think remove `files` in `package.json` and use `.npmignore`?

Tyriar:
> @Tyriar I think we don'nt need glupfile.js, jsdoc.json, src, tsconfig.json, tslint.json and webpack.config.json.

tslint.json (linting), webpack.config.js (demo bundling) and jsdoc.json (not needed anymore) are definitely not needed, I'm hessitant to remove the others as I think with the others it's possible to modify a node module and rebuild.

> BTW, do you think remove files in package.json and use .npmignore?

I'm more in favor of removing `.npmignore` in favor of `package.json`

torn4dom4n:
I see this in VSCode:
image
I think we don't need a few things. For me, if someone wants to modify a node module and rebuild, they can clone the repo. Removing files (not need) can compact memory. Finally, using `.npmignore` can help `package.json` isn't trouble.

Tyriar:
@torn4dom4n it was different in VS Code because of the way that it was getting pulled in back then.

Tyriar:
I'm going to close this off as you're targeting VS Code stable's module as opposed to the npm module. We now use npm to manage vscode-xterm though so you can do another PR to clean up this:

image

I agree people probably shouldn't be reinstalling modules after an `npm install`, so specifically adding these to package.json's ignore:

```
gulpfile.js
jsdoc.json
tsconfig.json
tslint.json
webpack.config.js
```

We still want `src/` as the sourcemaps point there for easy debugging and we still want `dist/` as I think people rely on this.

PR #1243, opened 2018-01-25
Support triggering callbacks as soon as the mouse hovers over a link.

amejia1:
Although tooltip callback could also work, I'm in need of performing actions as soon as the mouse hovers over a link. The delay in calling tooltip makes it look as the the terminals don't provide a responsive experience.

With xterm.js v2 I was able to add on hover callbacks to the link anchors, therefore this change would bring back the ability to do the same but with xterm.js v3.


Tyriar:
@amejia1 I'm a little worried about the amount of callbacks here. Are you using `hoverCallback` for tooltips? Maybe a `hoverDelay` configurable option would be better?

amejia1:
If the hoverDelay can be set to 0 (i.e. no delay), then I can use that as well.

amejia1:
Also, I'm not displaying a tooltip with hoverCallback. I'm adding a class to the terminal container div in order for Atom to display two new menu items for opening and copying links.

Tyriar:
@amejia1 as in menu bar up the top? I don't totally understand the use case still, do you have a screenshot/gif?

Also keep in mind that the link matcher API was never declared stable as is likely to change in the future.

amejia1:
This is what I mean.
atom-xterm-right-click-links

This behavior is possible by appending a class to the div's classList that's holding the terminal on hover, and then removing it once it hovers away from links.

Tyriar:
@amejia1 can't this already be accomplished by using the main click handler and handling right click?

mofux:
AFAIK the context menu in atom works on css classes. When right clicking in atom, it propagates a DOM event that bubbles up the DOM and collects any context menu entries where the css selector for that context menu item matches a selector in that event DOM tree. This means that if you want to propagate context menu items for xterm.js, you have to add a class to the terminal element whenever a link is hovered, and then you can register your link specific context menu items on that class ('e.g. .xterm.link-hovered')

amejia1:
@Tyriar @mofux  Woah, just noticed the last two comments here. I'm going to respond to them at https://github.com/xtermjs/xterm.js/pull/1251 since that's my latest attempt at resolving this issue.

PR #1244, opened 2018-01-25
Remove legacy .gitignore rules

Tyriar:
@parisk we don't need the dist/ rule anymore right?


Tyriar:
Actually maybe we do need dist still

image

What task builds the dist folder?

.gitignore
```diff
@@ -19,10 +19,5 @@ fixtures/typings-test/*.js
 # Directories needed for code coverage
 /coverage/
 
-# Keep legacy files out of the repo, this can be removed when we merge v3 into master
-dist/
```
parisk: We need `dist/` to keep the distribution files  created by the following lines:

https://github.com/xtermjs/xterm.js/blob/6c54c6cded526c7c72fdd276bed818e3a4b5d8f0/bin/prepare-release#L17-L18

Tyriar: Ah, that's where it's done. Was there any reason we don't just move the build dir to dist?

Tyriar: Created https://github.com/xtermjs/xterm.js/issues/1254 to follow this question up

PR #1251, opened 2018-01-26
Support triggering tooltip callback using customizable delay.

amejia1:
This will support an option for defining the delay to trigger the tooltip callback, or a delay less than or equal to 0 which means the tooltip callback will be triggered immediately. I also verified that not defining the new `tooltipDelay` option will set it to the original default delay.

The below shows what I need this for.
atom-xterm-no-tooltip-delay

### DIFF
--- a/src/Linkifier.ts
+++ b/src/Linkifier.ts
@@ -5,7 +5,7 @@
 
 import { IMouseZoneManager } from './input/Types';
 import { ILinkHoverEvent, ILinkMatcher, LinkMatcherHandler, LinkMatcherValidationCallback, LineData, LinkHoverEventTypes, ILinkMatcherOptions, ITerminal, IBufferAccessor, ILinkifier, IElementAccessor } from './Types';
-import { MouseZone } from './input/MouseZoneManager';
+import { MouseZone, HOVER_DURATION } from './input/MouseZoneManager';
 import { EventEmitter } from './EventEmitter';
 
 const protocolClause = '(https?:\\/\\/)';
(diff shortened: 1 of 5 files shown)


amejia1:
@Tyriar Following up on the discussion at https://github.com/xtermjs/xterm.js/pull/1243 , I tried your suggestion just now. I'm afraid I can't use the main click handler on right click for Atom. The main click handler is never triggered on right click.

@mofux Yes, and that is why I need this change. I have the "Open Link" and "Copy Link Address" menu items appear only on link hover by adding/removing a class in the terminal container div. I'm using tooltipCallback and leaveCallback for this but I need the tooltipCallback to be triggered immediately upon hovering over a link.

Tyriar:
I actually just put out a PR that will enable the support of modifiers and I think you can leverage this to achieve your goal as the new `willLinkActivate` fires on `mousedown`, not `click` https://github.com/xtermjs/xterm.js/pull/1264. Please try it out and let me know how it goes.

Also looking into the future, there will likely be a big refactor to the link matcher API once custom parsers are supported https://github.com/xtermjs/xterm.js/issues/583, before we make it stable.

amejia1:
I need something that fires on hover, not mouse down.

Tyriar:
@amejia1 my thinking is that the mousedown would be hit before the contextmenu is, are you sure it doesn't?

amejia1:
@Tyriar I was concerned that the context menu would appear first before this callback is called on Linux (where the contextmenu appears on mousedown instead of mouseup), but that is not the case. PR https://github.com/xtermjs/xterm.js/pull/1264 will work for me.

Tyriar:
Great 👍

Issue #1039, opened 2017-10-06
The paste option is not shown in the context menu in Firefox


travisobregon:
In the latest version of Firefox, the option to paste is not there:

![image](https://user-images.githubusercontent.com/12074713/31281859-967c3c9c-aaa8-11e7-859f-acde85b50d77.png)

## Details
- Browser and browser version:  Firefox 56.0 (latest)
- OS version: Windows 7
- xterm.js version: 2.9.1 (v3 branch)

### Steps to reproduce

1. Follow the steps to run the demo app in the README

joegaffey:
I've submitted a pull request to address this issue:
https://github.com/sourcelair/xterm.js/pull/1040 

You can try it here:
https://joegaffey-xtermjs.glitch.me/

runner-mei:
This bug is happen on my firefox57/windows

Tyriar:
Duplicate https://github.com/xtermjs/xterm.js/issues/418

Issue #418, opened 2016-12-16
context menu does not offer paste in Firefox-on-Windows


risacher:
The right-click context menu does not offer 'paste' as an option when using Firefox 50.0, FirefoxPortable 50.0.2, and FirefoxPortable 50.1.0 on Windows 8.1.  Pasting works when the paste option is selected from the Firefox menu.

## Details
- Browser and browser version:  Firefox 50.0 and FirefoxPortable 50.0.2 and 50.1.0
- OS version:  Win 8.1
- xterm.js version:  github as of 2016-12-16

### Not affected: 
 - Firefox Nightly, Google Chrome, and Safari on OS X (El Capitan) 
 - Firefox 50.1.0 on Linux
 - Google Chrome on Win 8.1 
 - Internet Explorer 11 on Win 8.1 has a paste option, but it doesn't work. (Different problem)

### Steps to reproduce

1. Open an xterm.js Terminal in Firefox on Windows 
2. Select some text with the mouse
3. Right click and select 'Copy'
4. Right click again to bring up context menu; observe that there is no "Paste" option.

parisk:
Thanks for reporting this @risacher. Will take a look and post updates here.

Does this happen in the demo or in your consumer application? If the case is the latter there are a few chances it happens because one of the terminal's ancestors are using CSS transforms.

Oliboy50:
Same issue on Windows 10.

Any news about this? 

And not only the context menu, I just don't have any clue about how to paste in Firefox on Windows... :(

parisk:
@Oliboy50 I will try to get on top of this as soon as possible.

Are you aware if the host application of xterm.js that you are using, is using CSS transforms on xterm.js or any of it's parent elements?

risacher:
@parisk you can poke at a live version of my app at https://risacher.org/ttyx/ , the code for which is https://github.com/risacher/ttyx.  You shouldn't be able to log in, but you can test the terminal paste function without logging into the server.

In my case the only css that's loaded is xterm.css from xterm.js itself and https://risacher.org/ttyx/style.css, which is derived from chjj's tty.js style.css.  

I dunno about the demo app;  I'll try to check that.

risacher:
I just confirmed that the same thing happens in the demo app.

risacher:
@Oliboy50 I can paste in Firefox-on-Windows by selecting "paste" in the dropdown menu, which opens when I click on a symbol in the upper right of the browser window that looks like three horizontal bars.  

[...]

PR #1253, opened 2018-01-26
adjust timeout to ensure right-click event has fired

sitzmar:
As described in issues #418, #1039 and #1040, there appears to be a timing/race condition in the right-click handler that impacts Firefox on Windows. In the clipboard handler, the textarea element gets modified to be under the mouse cursor, and and then very quickly reset with a setTimeout method. 

In testing, a minimum value of 135ms consistently resolved the issue. However using 200ms gives a little extra buffer, and doesn't appear to have any negative impacts.

Tested on:
Firefox 58, Chrome 64, and Edge 41 on Windows 10
Firefox 58, Chrome 63, and Safari 11.0.2 on macOS HighSierra

### DIFF
--- a/src/handlers/Clipboard.ts
+++ b/src/handlers/Clipboard.ts
@@ -100,6 +100,7 @@ export function moveTextAreaUnderMouseCursor(ev: MouseEvent, textarea: HTMLTextA
   textarea.focus();
 
   // Reset the terminal textarea's styling
+  // Timeout needs to be long enough for click event to be handled.
   setTimeout(() => {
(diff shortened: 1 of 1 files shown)


mofux:
Good finding! @Tyriar do you think this could be a solution for now? I know it's not ideal and it would be better to understand and fix the root cause of this problem, but since this issue has been around for so long this might at least be a good workaround?

Tyriar:
Thanks for taking the time to fix this @sitzmar! 😃

sitzmar:
No problem!

PR #1256, opened 2018-01-28
Change render layer id to a class

### LINKED ISSUE #1255: canvas has static id attribute
the canvas has id such as `xterm-selection-layer`, when run multiple xterm inst in same page, it has a potential problem: can not select it by id, and it has no className. will you consider using class or unique id instead ?

## Details
- Browser and browser version: all
- OS version: all
- xterm.js version: 3.0.1

### Steps to reproduce

just use it.

Tyriar:
It's only used to identify them in DOM explorer currently

Fixes #1255


mofux:
LGTM 👍

PR #1264, opened 2018-02-01
Allow support of modifiers with links

### LINKED ISSUE #1021: Provide modifier option for links
VS Code only follows links when ctrl is held. Support for this inside xterm.js would be useful, specifically:

- If the modifier is set to ctrl, the link underline will only show up when ctrl is held
- Events are ignored unless ctrl is held

Also consider the following scenario:

1. `tmux`
2. `echo 'http://google.com'`

Ideally no modifier mousedown/click events on the link would be handed by tmux, but ctrl+mousedown/click events would be handled by Linkifier. Currently (after https://github.com/sourcelair/xterm.js/pull/1022) no modifier mousedown/click events on a link will never make it to tmux even when ctrl is required to follow the link.

Tyriar:
Fixes #1021

### DIFF
--- a/src/input/MouseZoneManager.ts
+++ b/src/input/MouseZoneManager.ts
@@ -151,10 +151,10 @@ export class MouseZoneManager implements IMouseZoneManager {
     // components from handling the mouse event.
     const zone = this._findZoneEventAt(e);
     if (zone) {
-      // TODO: When link modifier support is added, the event should only be
-      // cancelled when the modifier is held (see #1021)
-      e.preventDefault();
-      e.stopImmediatePropagation();
+      if (zone.willLinkActivate(e)) {
+        e.preventDefault();
+        e.stopImmediatePropagation();
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/input/MouseZoneManager.ts
```diff
@@ -151,10 +151,10 @@ export class MouseZoneManager implements IMouseZoneManager {
     // components from handling the mouse event.
     const zone = this._findZoneEventAt(e);
     if (zone) {
-      // TODO: When link modifier support is added, the event should only be
-      // cancelled when the modifier is held (see #1021)
-      e.preventDefault();
-      e.stopImmediatePropagation();
+      if (zone.willLinkActivate(e)) {
```
mofux: Is it safe to assume here that the `zone.willLinkActive` function exists?

Tyriar: It is there yeah, it's mandatory for `MouseZone`/`IMouseZone`, optional for `ILinkMatcherOptions`

PR #1271, opened 2018-02-05
Update README.md

krasimir:
In `Getting Started` section we read `npm install` only. Shouldn't be `npm install xterm`?


Tyriar:
You're right, thanks for the PR :smiley:

PR #1273, opened 2018-02-05
Add `watch` gulp task and npm script

### LINKED ISSUE #1232: We should bring back the npm script to watch and relaunch the dev server on changes
`npm run dev` used to rebuild on changes. We should bring it back.

We might be better off splitting it out to a `npm run watch` to build sources as well as a `npm run demo` to build the demo and launch the server. That way we can save a bunch of time if we don't need the server (eg. building the module for another application).

parisk:
This PR closes #1232 by introducing a `watch` gulp task and npm script.

---

## 🆕 Workflow for the Demo app

Instead of building xterm.js **before** running the demo server, xterm.js developers should run `npm run watch` in parallel with `npm start`.

This will serve the contents of the `demo/` directory, build xterm.js and watch for file system changes in `src/` to trigger the build again.

If developers work on a Procfile-based environment like SourceLair, both tasks (`npm start` and `npm run watch`) will be handled automatically by the development environment.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -192,8 +192,8 @@ gulp.task('webpack', ['build'], function() {
     .pipe(gulp.dest('demo/dist/'));
 });
 
-gulp.task('webpack:watch', ['webpack'], () => {
-  gulp.watch('./src/*', ['webpack']);
+gulp.task('watch', ['webpack'], () => {
+  gulp.watch(['./src/*', './src/**/*'], ['webpack']);
 });
 
 /**
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -192,6 +192,10 @@ gulp.task('webpack', ['build'], function() {
     .pipe(gulp.dest('demo/dist/'));
 });
 
+gulp.task('webpack:watch', ['webpack'], () => {
+  gulp.watch('./src/*', ['webpack']);
+});
+
 /**
  * Submit coverage results to coveralls.io
  */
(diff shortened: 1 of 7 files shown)


gulpfile.js
```diff
@@ -192,6 +192,10 @@ gulp.task('webpack', ['build'], function() {
     .pipe(gulp.dest('demo/dist/'));
 });
 
+gulp.task('webpack:watch', ['webpack'], () => {
+  gulp.watch('./src/*', ['webpack']);
```
Tyriar: This needs to be `['./src/*', './src/**/*']`

gulpfile.js
```diff
@@ -192,6 +192,10 @@ gulp.task('webpack', ['build'], function() {
     .pipe(gulp.dest('demo/dist/'));
 });
 
+gulp.task('webpack:watch', ['webpack'], () => {
```
Tyriar: Should we just call it `watch` instead?

parisk:
@Tyriar fixed comments and added demo instructions for all platforms possible (included SourceLair, Docker and foreman/Procfile runners). Lemme know.

parisk:
@Tyriar not sure what would be the most relevant to people.

I sorted the "demo workflows" per solution that has the least steps to set it up (managed → manual).

I want to keep it that way, because plain macOS, Linux, and Windows need at least 3 steps to get it working (npm install, npm start npm run watch), so if someone is using a more managed solution with 0, 1 or 2 steps this means less chances of things going wrong 😄.

Issue #1278, opened 2018-02-08
Xterm.js in headless mode yields TypeError on resize


johnpoth:
When running Xterm.js 3.0.2 in [headless mode](https://github.com/xtermjs/xterm.js/issues/266) ( i.e not calling the `open(container)` function) I get:

`kubebox.js:2633 TypeError: Cannot read property 'width' of undefined
    at Terminal.resize (xterm.js:4099)`

Which corresponds to this [line](https://github.com/xtermjs/xterm.js/blob/3.0.2/src/Terminal.ts#L1855).

Which is expected since open() wasn't called and so `this.charMeasure` is undefined. I'm not sure this is needed when running headless. I removed the line and it seems to be working.

If it **is** needed we could mock the DOM elements by using something like [JSDOM](https://github.com/jsdom/jsdom)?

Thanks!

Tyriar:
Yes it shouldn't be needed as it's measured when `open` is called.

Tyriar:
Fixed in https://github.com/xtermjs/xterm.js/pull/1280

PR #1280, opened 2018-02-12
Fix xterm.js in headless mode yields TypeError on resize

AndrienkoAleksandr:
Related issue: https://github.com/xtermjs/xterm.js/issues/1278
Fix xterm.js in headless mode yields TypeError on resize.


Tyriar:
Looks good :+1:

PR #1282, opened 2018-02-12
Added FreeMAN to list of real-world uses

matthew-matvei:
FreeMAN uses `xterm.js` to provide a front end for its integrated shell terminal.


Tyriar:
:+1:

PR #1288, opened 2018-02-16
fix terminal width calculation

### LINKED ISSUE #1284: Fit addon covers scrollbar
If you call `fit`, it seems to extend the text area over the scrollbar. The mouse has a text cursor and you can't click the scrollbar. Without `fit`, it seems fine. It looks like fit adds 1 or 2 too many columns. 

In the last release build (3.0.2) it had the text cursor, but clicking the scrollbar worked.

## Details
- Browser and browser version: Electron 1.8.2, Chrome 64.0.3282.140, FFx 58.0.1
- OS version: Windows 10
- xterm.js version: 3.1.0 (npm version) 

### Steps to reproduce

1. Create, open xterm instance
2. Call fit()
3. Mouse over scrollbar: shows text cursor
4. Click scrollbar: nothing
5. Remove fit call, works OK

### Example

```html
<!doctype html>
<html lang=en>
<head>
<meta charset=utf-8>
<title>Test Page</title>
<style>

  #terminal {
    border: 1px solid gray;
    width: 632px;
    height: 333px;
    margin: auto;
  }

</style>
<link rel="stylesheet" href="../node_modules/xterm/dist/xterm.css">
</head>
<body>

<div id='terminal'>

</div>

<script src="../node_modules/xterm/dist/xterm.js"></script>
<script src="../node_modules/xterm/dist/addons/fit/fit.js"></script>
<script>
  Terminal.applyAddon(fit);

  term = new Terminal({rows: 40, cols: 40});
  term.open(document.querySelector("#terminal"));
  term.focus();

  // fit fits properly, but breaks scrollbar
  term.fit();

  // remove the fit call and scollbar works OK

  for( let i = 0; i< 1000; i++ )
    term.write(`Line ${i}\r\n`);


</script>
</body>
</html>
```

dgriffen:
Fixes #1284 

The problem breaks down as follows:
- The regression was introduced in #1208 to support padding in the `.xterm` css class
- Padding `.xterm` is calculated from the right hand side of the scrollbar
- The PR attempted to make the effective area of the terminal match the padding exactly
- The addition of scrollBarWidth in the code segment below causes the effective interactive area of the terminal to be `(column_width * num_columns) + scrollbar_width`

`Renderer.ts`
```typescript
this._terminal.screenElement.style.width = `${this.dimensions.canvasWidth + this._terminal.viewport.scrollBarWidth}px`;
```

- This means that there is an area past where the terminal will actually have text that has all the interactions of a terminal, when the padding is less than the width of the scroll bar this will cause interactivity issues with the scroll bar element.

I believe the correct fix is to remove the addition of scrollBarWidth when setting the screenElement width. ~~This does mean that padding on the right side of the terminal will appear to be too large but the terminal already had the implicit padding of avoiding the scroll bar anyways.~~

**Edit**: After looking at the results more closely this will have the same visual layout as before, it just makes the clickable area match the actual terminal area.

A workaround for anyone affected by this would be to set the right padding of `.xterm` to be equal to the width of the scroll bar.

### DIFF
--- a/src/renderer/Renderer.ts
+++ b/src/renderer/Renderer.ts
@@ -121,7 +121,7 @@ export class Renderer extends EventEmitter implements IRenderer {
     }
 
     // Resize the screen
-    this._terminal.screenElement.style.width = `${this.dimensions.canvasWidth + this._terminal.viewport.scrollBarWidth}px`;
+    this._terminal.screenElement.style.width = `${this.dimensions.canvasWidth}px`;
     this._terminal.screenElement.style.height = `${this.dimensions.canvasHeight}px`;
 
     this.emit('resize', {


Tyriar:
This broke the scrollbar in VS Code as well when I reduced the horizontal margin between the xterm element and the scrollbar. The fix works 👍

PR #1293, opened 2018-02-19
Validate colors

### LINKED ISSUE #1195: Allow theme colors to be named colors?
Currently named colors are ignored by themes

## Details
- Browser and browser version: Firefox 57
- OS version: macOS 10.13
- xterm.js version: 3.0.1

### Steps to reproduce

1. Apply named colors like `'white'` to a theme
2. Colors are not handled by [_validateColor](https://github.com/xtermjs/xterm.js/blob/57c45712049e4fea67c28f614ec7fa5d6428e593/src/renderer/ColorManager.ts#L111) and fall back to the default

nikonso:
Closes #1195 

I did a few tests, and it seems to be working. 

Should I add tests to this feature, and if so, should I make the validate function public?


src/renderer/ColorManager.ts
```diff
@@ -66,6 +66,8 @@ function toPaddedHex(c: number): string {
  * Manages the source of truth for a terminal's colors.
  */
 export class ColorManager implements IColorManager {
+  private VALID_NON_NAMED_COLORS = new RegExp('^(#[0-9a-f]{3}|#(?:[0-9a-f]{2}){2,4}|(rgb|hsl)a?\((-?\d+%?[,\s]+){2,3}\s*[\d\.]+%?\))$', 'i');
```
Tyriar: Do you need this part given when we have `_validateNamedColor`?

nikonso:
@Tyriar you're right I don't need the Regex.

This method causes the mocha tests for the ColorManager to fail, because we don't have a document...

 `   const d = document.createElement('div');`

What should I do to fix this?

Tyriar:
@nikonso you can make document a constructor parameter and then mock it out with `.style.color`. It won't verify but it should keep the current testing behavior.

mofux:
LGTM 👍

PR #1296, opened 2018-02-23
Paste newlines as \r

olsonpm:
Fix issue where non-windows OS terminals don't paste newlines correctly
into misbehaving programs.  As noted in JetBrains/jediterm#136, the
most practical solution is for terminals to handle this in their paste
logic.  In a perfect world, all terminal programs would implement
bracketed paste mode properly, but that's a lot to ask.

- translate newlines to \r which is what terminal programs expect

### DIFF
--- a/src/handlers/Clipboard.test.ts
+++ b/src/handlers/Clipboard.test.ts
@@ -8,13 +8,19 @@ import * as Terminal from '../Terminal';
 import * as Clipboard from './Clipboard';
 
 describe('evaluatePastedTextProcessing', () => {
-  it('should replace carriage return + line feed with line feed on windows', () => {
-    const pastedText = 'foo\r\nbar\r\n';
-    const processedText = Clipboard.prepareTextForTerminal(pastedText, false);
-    const windowsProcessedText = Clipboard.prepareTextForTerminal(pastedText, true);
+  it('should replace carriage return and/or line feed with carriage return', () => {
+    const pastedText = {
+      unix: 'foo\nbar\n',
+      windows: 'foo\r\nbar\r\n'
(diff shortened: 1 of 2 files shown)


olsonpm:
I tested this branch on macOS 10.13, ubuntu 16.04, and my arch linux machine.  If there are any edge cases you'd like me to test then let me know, but I'm pretty confident in the fix after reading more than I ever thought I would about newline handling in terminals.

Tyriar:
@olsonpm I haven't noticed any issues with pasting in the terminal, do you have some steps that illustrate the bug?

olsonpm:
[originally encountered on hyper](https://github.com/zeit/hyper/issues/1448)

I'm able to reproduce it on a fresh ubuntu 16.04 by:

1. install vscode
2. enter integrated terminal
3. open nano
4. paste multiline text into nano

Those steps also reproduce the bug on my fresh macOS vm as well as my current arch box

As noted in the linked issue, more than just nano encounter this issue - any program which doesn't implement bracketed paste mode properly will encounter this issue.

If there's anything more I can provide let me know

Tyriar:
> This seems to be a bug in any terminal program which doesn't implement bracketed paste mode 

We do implement bracketed paste mode though? https://github.com/xtermjs/xterm.js/pull/1097

olsonpm:
but [nano doesn't](https://savannah.gnu.org/bugs/?40060)

Tyriar:
Ah I see

olsonpm:
awesome.  I appreciate the clear directions in the documentation for getting the demo up and running btw. Helped out a lot.

PR #1298, opened 2018-02-23
Pull web link support into an addon

### LINKED ISSUE #1297: Pull http link setup into an addon
It's getting pretty ugly handling http differently because it's built in. For example we expose `setHypertextLinkHandler` and `setHypertextValidationCallback` and need a `setWillLinkActivate` in order to fix https://github.com/Microsoft/vscode/issues/41415. It would be better if http was just another link handler but was nicely packaged in an addon.

Tyriar:
Fixes #1297

---

The main motivation behind this is to slim down the xterm.js API and move the complexity of customizing the web link handler into an addon. To customize it you can now do this:

```js
term.webLinksInit(
  /* handler */ console.log,
  /* options */ {
    validationCallback: (uri, cb) => { console.log('validate', arguments); cb(true); }
  }
);
```

A nice benefit of doing this is that it makes both the web links and the linkifier seem much simpler as they're now separates into different files.

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -3,6 +3,7 @@ import * as attach from '../build/addons/attach/attach';
 import * as fit from '../build/addons/fit/fit';
 import * as fullscreen from '../build/addons/fullscreen/fullscreen';
 import * as search from '../build/addons/search/search';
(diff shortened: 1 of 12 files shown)


mofux:
LGTM 👍

PR #1302, opened 2018-02-26
Enforce option value via types 

### LINKED ISSUE #889: Enforce option value via types
I think we can extend StringOption and similar to actually check the value at compile time, not just the type.

https://github.com/sourcelair/xterm.js/blob/39328d6e0c05b1f1f574303ad0288687ca6f813f/src/Types.ts#L33-L35

nikonso:
Closes #889 

If I missed some property, or if I should do something else, feel free to say something! 😄

### DIFF
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -487,12 +487,7 @@ declare module 'xterm' {
      * @param key The option key.
      */
     getOption(key: 'handler'): (data: string) => void;
-    /**
-     * Retrieves an option's value from the terminal.
-     * @param key The option key.
-     */
-    getOption(key: string): any;
-
+ 
     /**
      * Sets an option on the terminal.
      * @param key The option key.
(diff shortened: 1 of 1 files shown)


Tyriar:
The proposed change actually makes it worse I think as the typescript server will then present "BellStyle" as the type instead of showing all the strings which is more useful.

I think what I was getting at for #889 was to add more specific `setOption` calls, which is how the string values are already:

```ts
setOption(key: 'fontWeight' | 'fontWeightBold', value: null | 'normal' | 'bold' | '100' | '200' | '300' | '400' | '500' | '600' | '700' | '800' | '900'): void;
```

nikonso:
Yes, that makes complete sense. 

Then, the proper fix was to remove the  overloads with type any, as well as adding the missing call, correct?

Tyriar:
@nikonso the fix happened a while ago, back when there was only a single `setOption` for 'string' value types.

nikonso:
Yes but doesn't this pull request still make sense nevertheless as is?

Because of the **setOption** overload with the type any the _typings-test.ts_ could miss some cases, and it was by removing it that I found that the set option overload for _rows_ and _cols_ was missing..

Otherwise, feel free to close it 😉

Tyriar:
Oh I missed your new commit, the rows/cols thing is definitely good 👍 

I think I tried to remove `getOption(key: string): any` before but then I couldn't call it with generic string object as it was too strict. For example, I think we want to allow this:

```ts
const someString: string = 'foo';
term.getOption(someString)
```

When I tried a while ago this did not work, which would be a bad dev experience if you need to cast it to `<'rows'>someString`.

Tyriar:
Thanks again @nikonso! :smiley:

PR #1303, opened 2018-02-27
Get multi-line links working

### LINKED ISSUE #24: Link matchers do not work with multi-line links

Tyriar:
Fixes #24


Tyriar:
This is ready for review, multi line links seem to work great with the exception of echoing the link and then sizing the terminal down. This happens because of the workaround for reflow where we retain the text to the right of the terminal.

image

image

Since I plan on adding support for reflow in the next couple of months I don't think it's even worth working around this issue.

mofux:
Found a couple of issues by running `head README.md` from the demo. See the attached capture.

kapture 2018-03-09 at 22 02 16

Tyriar:
@mofux should be fixed now, added tests too 👍

mofux:
I found another bug that is also likely caused by wrapped lines. If you run `head README.md` twice you should have enough buffer filled to scroll. Scroll the whole way up and hover the upper links, it will work. Then scroll a few lines down so that the first link is outside the viewport. Notice that all the other links won't get detected anymore. I believe this is caused because all links are actually in one line that is wrapped:

kapture 2018-03-13 at 23 19 01

Tyriar:
@mofux I decided to ignore this edge case for now, but you found a case where it looks particularly bad 😛. It happens because we don't bother to backtrack and linkify lines that are outside of the viewport (only initial line (!isWrapped) are linkified). I removed the TODO and make a comment to that effect in https://github.com/xtermjs/xterm.js/pull/1303/commits/b44a11c42eaebf2115dd0b91be71ebc3f91544ff

I guess if the requested line is the first line and it's flagged as wrapped, we could backtrack and linkify. Then add handling to make sure mouse zones are not created outside the viewport.

Tyriar:
@mofux ok I think that fixes it

mofux:
Just tested, this is working nicely now! There's one issue left, but it might have to do with the link detection in general. Some links are too long (ending on a bracket), and others span muliple links at once - see the capture below.

kapture 2018-03-16 at 15 33 46

Tyriar:
@mofux yeah it's an issue with the link detection in general. Since we use regex we either have to decide whether to allow brackets (and quotes) and accept cases like this, or break heaps of links such as https://msdn.microsoft.com/en-us/library/windows/desktop/ms684839(v=vs.85).aspx

Tracked in https://github.com/xtermjs/xterm.js/issues/583

mofux:
I think we can merge this PR. Let's tackle the link detection issues separately.

Issue #1294, opened 2018-02-21
Use a dynamic cache for the character texture atlas


Tyriar:
This would significantly boost performance particularly on cells where the non-default background or non-ascii chars are used.

/cc @bgw

Downstream: https://github.com/zeit/hyper/issues/2592
Related: https://github.com/xtermjs/xterm.js/issues/1063

PR #1307, opened 2018-03-05
Split CharAtlas types and utility functions off

bgw:
I'm starting to pull some changes off of my large WIP branch/commit that adds an alternative dynamic character atlas (#1294). We're going to need to support multiple CharAtlas implementations, and this should make that easier.

### DIFF
--- a/src/renderer/CharAtlas.ts
+++ b/src/renderer/CharAtlas.ts
@@ -5,21 +5,12 @@
 
 import { ITerminal } from '../Types';
 import { IColorSet } from './Types';
+import { ICharAtlasConfig } from './atlas/Types';
 import { isFirefox } from '../shared/utils/Browser';
 import { generateCharAtlas, ICharAtlasRequest } from '../shared/CharAtlasGenerator';
+import { generateConfig, configEquals } from './atlas/CharAtlasUtils';
 
 export const CHAR_ATLAS_CELL_SPACING = 1;
-
-interface ICharAtlasConfig {
-  fontSize: number;
-  fontFamily: string;
-  fontWeight: string;
-  fontWeightBold: string;
-  scaledCharWidth: number;
-  scaledCharHeight: number;
-  allowTransparency: boolean;
-  colors: IColorSet;
-}
 
 interface ICharAtlasCacheEntry {
   bitmap: HTMLCanvasElement | Promise<ImageBitmap>;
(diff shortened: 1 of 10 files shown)


mofux:
Looks good, but I'd rather merge this as part of your final PR to be honest. I'm a little bit concerned this grows the code base without much gain otherwise. @Tyriar Any thoughts?

bgw:
> Looks good, but I'd rather merge this as part of your final PR to be honest.

If you'd like, I can close this and wait until the rest of my changes for #1294 are ready, and then have one PR with a bunch of commits.

I just figured it might be easier to review if I got a few small changes out of the way first. (this and #1308).

Tyriar:
It doesn't look like this would affect the overall size of the codebase, seems like it's mostly moving code around rather than adding. Splitting the big change into several smaller changes will make them much easier to merge in.

src/renderer/CharAtlas.ts
```diff
@@ -5,22 +5,13 @@
 
 import { ITerminal } from '../Types';
 import { IColorSet } from './Types';
+import { ICharAtlasConfig } from './atlas/Types';
 import { isFirefox } from '../shared/utils/Browser';
 import { generateCharAtlas, ICharAtlasRequest } from '../shared/CharAtlasGenerator';
+import { generateConfig, configEquals } from './atlas/CharAtlasUtils';
 
 export const CHAR_ATLAS_CELL_SPACING = 1;
```
Tyriar: Should this be removed?

Tyriar:
Also I like the idea of shipping 2 options to begin with, I caused a lot of problems when I did this change initially because I completely ripped out the DOM-based renderer. It would have been worth the extra overhead to have an option for users to switch back.

src/renderer/atlas/Types.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) 2017 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { FontWeight } from 'xterm';
+import { IColorSet } from '../Types';
+
+export const CHAR_ATLAS_CELL_SPACING = 1;
+export const INVERTED_DEFAULT_COLOR = -1;
```
Tyriar: It doesn't look like are being used anywhere.

bgw:
I've added a couple more commits that should address the comments.

On a somewhat related note: @Tyriar, how do you feel about using absolute imports everywhere?

Tyriar:
> On a somewhat related note: @Tyriar, how do you feel about using absolute imports everywhere?

I very much want this, created https://github.com/xtermjs/xterm.js/issues/1314 and made some notes in it.

Tyriar:
Thanks again @bgw, looking forward to more in this area 😄

PR #1308, opened 2018-03-05
Remove didCharSizeChange logic from renderer

bgw:
The only thing `didCharSizeChange` was used for was to figure out if we
need to refresh the character atlas or not. However, `acquireCharAtlas`
already inspects the atlas owned by this terminal, and if it matches, it
avoids generating a new atlas.

So if `didCharSizeChange` is true and it's not needed, it quickly bails
out as a no-op. But if `didCharSizeChange` is false, and it was needed
(possible given the complexity of testing all of these edge cases), it
could introduce a bug.

I think it makes sense to just get rid of `didCharSizeChange`, and just
always call `acquireCharAtlas`.

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -472,11 +472,9 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
       case 'lineHeight':
       case 'fontWeight':
       case 'fontWeightBold':
-        const didCharSizeChange = (key === 'fontWeight' || key === 'fontWeightBold' || key === 'enableBold');
-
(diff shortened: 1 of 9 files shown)


Tyriar:
@bgw I think this looks good, changing settings and resizing rarely happens anyway, the main concern was extra GCs that could be caused by `Renderer. onWindowResize` but that's still blocked behind a check for `window.devicePixelRatio`. And yes, we I think we have hit bugs of the char atlas not regenerating like you mention 😄 

I'll merge this in once 3.2.0 gets branched off.

PR #1309, opened 2018-03-05
Only move horizontally on alt click in normal buffer

### LINKED ISSUE #1305: Have alt+click only move horizontally when in the normal buffer
Current alt+click will move the cursor to that position, it was made to work with vim/nano/etc by moving left to the start of the line, moving up/down vertically and then moving right. The problem is that this can lead to weird behavior when in a shell as up moved through history. A nice solution to this is to only move horizontally when in a shell, we need new API such as a setting to define alt behavior for that though.

See https://github.com/Microsoft/vscode/issues/44642

/cc @npezza93

Tyriar:
Fixes #1305

### DIFF
--- a/src/Types.ts
+++ b/src/Types.ts
@@ -252,6 +252,7 @@ export interface IBuffer {
   tabs: any;
   scrollBottom: number;
   scrollTop: number;
+  hasScrollback: boolean;
   savedY: number;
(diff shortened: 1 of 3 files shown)


mofux:
Seems to do the job perfectly! LGTM 👍

Issue #1304, opened 2018-02-27
Mouse wheel translation to arrows doesn't feel right


Tyriar:
I think fast scrolls are being capped at 1 scroll per frame, so it's difficult to get from one end of a file to the other.

Repro:

```
vim somelongfile
```

Scroll using a touchpad very fast to the bottom

Tyriar:
This is talking about the feature being introduced in https://github.com/xtermjs/xterm.js/pull/1310

PR #1310, opened 2018-03-05
Send arrow events on mouse wheel in alt buffer

### LINKED ISSUE #426: Mouse scroll should move cursor in vim
### Steps to reproduce

1. `vim`
2. `i`
3. Fill file beyond number of rows
4. Scroll using mouse wheel, the vim cursor should respond

Tyriar:
Fixes #426

---

I guess I didn't put a PR out for this, it's been working great in VS Code


mofux:
Quite unusable with the Macbook touchpad (generates way too many scroll events with a single swipe), but works great with a mouse. Shall we handle the Macbook touchpad case?

mofux:
As a second note, what do you think about moving the cursor horizontally if the `shift` key is down while scrolling?

Tyriar:
@mofux it's been in VS Code for over a month, haven't heard any complaints. In fact I'm just doing https://github.com/xtermjs/xterm.js/issues/1304 now to send more events, based on the amount scrolled. So a quick flick can get you to the start to the end of the file.

Tyriar:
> As a second note, what do you think about moving the cursor horizontally if the shift key is down while scrolling?

I wouldn't want to add anything like this unless it's in multiple other terminals and used widely.

Tyriar:
@mofux pushed the change for #1304. This actually fixes scroll by line/page in the regular scroll bar, as well as allowing scrolling of _partial_ lines. By default scroll events with a touch pad scroll by pixel, so if you scroll 16 times by 1 pixel before it would scroll 16 lines, now it keeps track of partial lines and when you do the 16th single pixel scroll it will scroll the single line.

Tyriar:
I remember people saying that scrolling in Windows was all messed up, this was probably due to their specific mouse settings set to scroll by line.

mofux:
Just tested it, doesn't seem to work for me after you latest changes. Scroll works perfectly fine on the viewport, but if fails for the alt-buffer scroll (doesn't scroll, not errors). Logging `amount` on Terminal.ts:1061 gives `NaN`. Can you reproduce?

mofux:
Oh and one more thing while we're at it, we are listening twice for the wheel event (`on(el, 'wheel', (ev: WheelEvent) => {...})`), at `Terminal.ts:1056` and right after that block at `Terminal.ts:1085`. I think we should merge those listeners.

Tyriar:
@mofux it's probably a browser support problem, I only tested on Firefox Nightly/Linux. What browser were you on?

mofux:
@Tyriar I've tested on Chrome 64.0.3282.186. Also tried it on Firefox, which works when scrolling using the mouse (which fails on Chrome) - but it also fails in Firefox when scrolling using the integrated Macbook touchpad (also logs `NaN`).

mofux:
Fixed the problem on your branch, `_wheelPartialScroll` was initially `undefined` which caused the `NaN` problems. Initialising it to `0` makes this work on Chrome.

Tyriar:
@mofux I guess I didn't test thoroughly after a refactor, thanks for the fix 😃. I'll still need to test on IE/Edge/Safari when I get a chance.

Tyriar:
Works on Safari, ready for final review

PR #1313, opened 2018-03-07
Add xterm.js Docker image to documentation

### LINKED ISSUE #1295: Propose: create xterm.js organization on the dockerhub and push xterm.js image to the dockerhub.
It would be nice to have official docker image xterm.js on the dockerhub. After each release you can create new latest docker image with version tag. It's usefull for some new customers who want to try xterm.js without installation node, npm. They simply do "docker run" with xterm.js image and then tests xterm.js like docker application. For now you have dockerfile. So I don't think it's a problem to create xterm.js organization on the dockerhub and begin to do realese docker image after each realease xterm.js.

parisk:
Closes #1295

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,4 +16,4 @@ RUN npm install
 COPY . /usr/src/app
 
 # Run the tests and build, to make sure everything is working nicely
-RUN npm run build && npm run test
+RUN npm run build && npm run webpack && npm run test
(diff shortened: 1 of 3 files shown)


parisk:
@Tyriar I have set up [automated builds](https://docs.docker.com/docker-hub/github/) for the Docker image, so it pretty much happens on its own when we push to master and when we push a new tag.

Tyriar:
Nice 🙂

PR #1315, opened 2018-03-08
Rename all remaining private variables to use underscores

Tyriar:
Variables changed based on regex: `private [^(_|get)]`

I did the following to test and didn't see errors:

- Run long command
- Run script that uses all ansi colors
- Run find addon
- Open vim and move around a bit


Tyriar:
@mofux 👍 for notifying in the release notes

Tyriar:
I'll add the breaking change label to remind me to call it out.

PR #1316, opened 2018-03-08
Add noUnusedLocals tsconfig flag, fix issues

Tyriar:
Two parts were commented out:
- _writeStopped: This is because _xoffSentToCatchUp references this variable
  and the feature is sort of half implemented/disabled.
- _clearChar: This is because some comments still reference this, it's part
  of the work needed to do better line redrawing.


mofux:
LGTM 👍

PR #1317, opened 2018-03-09
Merge ICharAtlasRequest with ICharAtlasConfig

bgw:
`ICharAtlasRequest` and `ICharAtlasConfig` need almost exactly the same set of information, so it's simpler if we just merge the two types.

As an added bonus, this also adds `devicePixelRatio` to the `config`, which helps guarantee that we won't ever accidentally end up with an atlas using a different pixel ratio than we need.


Tyriar:
👍

Issue #1322, opened 2018-03-11
Options object is mutated when creating terminal


LucianBuzzo:
If I have some code like the following:

```js
const termConfig = {
	cols: 80,
	cursorBlink: false,
	rows: 24,
	theme: {
		background: '#343434',
	},
};
const tty = new Xterm(termConfig);
```

If I try to re-use the value of `termConfig` it will cause problems, as the value has been unexpectedly mutated. For example, the theme value will be `null`.
This could be resolved on this line https://github.com/xtermjs/xterm.js/blob/3.2.0/src/Terminal.ts#L260 by using `Object.assign` or a clone utility (if you have one available).

## Details
- Browser and browser version: whatsmybrowser.org/b/YWN4JTB
- OS version: OSX 10.13.3
- xterm.js version: 3.2.0

### Steps to reproduce

*See code example above*

Tyriar:
👍 open to PRs, we can't use `Object.assign` as we need to retain IE11 support for now.

LucianBuzzo:
Cool! I'll find a solution that will support IE11

Tyriar:
Fixed in https://github.com/xtermjs/xterm.js/pull/1326

PR #1326, opened 2018-03-11
Do not mutate options object used in constructor

LucianBuzzo:
Connects to #1322

change-type: patch

### DIFF
--- a/src/Terminal.test.ts
+++ b/src/Terminal.test.ts
@@ -20,12 +20,13 @@ class TestTerminal extends Terminal {
 
 describe('term.js addons', () => {
   let term: TestTerminal;
+  const termOptions = {
+    cols: INIT_COLS,
+    rows: INIT_ROWS
+  };
 
   beforeEach(() => {
-    term = new TestTerminal({
-      cols: INIT_COLS,
-      rows: INIT_ROWS
-    });
+    term = new TestTerminal(termOptions);
     term.refresh = () => {};
     (<any>term).renderer = new MockRenderer();
     term.viewport = new MockViewport();
@@ -41,6 +42,15 @@ describe('term.js addons', () => {
         remove: () => {}
       }
     };
+  });
+
+  it('should not mutate the options parameter', () => {
+    term.setOption('cols', 1000);
+
+    assert.deepEqual(termOptions, {
+      cols: INIT_COLS,
+      rows: INIT_ROWS
+    });
   });
 
   it('should apply addons with Terminal.applyAddon', () => {
--- a/src/utils/Clone.ts
+++ b/src/utils/Clone.ts
@@ -1,0 +1,28 @@
+/**
+ * Copyright (c) 2016 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+/*
+ * A simple utility for cloning values
+ */
+export const clone = <T>(val: T, depth: number = 5): T => {
+  if (typeof val !== 'object') {
+    return val;
+  }
+
+  // cloning null always returns null
+  if (val === null) {
+    return null;
+  }
+
+  // If we're cloning an array, use an array as the base, otherwise use an object
+  const clonedObject: any = Array.isArray(val) ? [] : {};
+
+  for (const key in val) {
+    // Recursively clone eack item unless we're at the maximum depth
+    clonedObject[key] = depth <= 1 ? val[key] : clone(val[key], depth - 1);
+  }
+
+  return clonedObject as T;
+};
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -40,6 +40,7 @@ import { CharMeasure } from './utils/CharMeasure';
 import * as Browser from './shared/utils/Browser';
 import * as Strings from './Strings';
 import { MouseHelper } from './utils/MouseHelper';
+import { clone } from './utils/Clone';
 import { DEFAULT_BELL_SOUND, SoundManager } from './SoundManager';
 import { DEFAULT_ANSI_COLORS } from './renderer/ColorManager';
 import { MouseZoneManager } from './input/MouseZoneManager';
@@ -245,7 +246,7 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     options: ITerminalOptions = {}
   ) {
     super();
-    this.options = options;
+    this.options = clone(options);
     this._setup();
   }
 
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -88,5 +89,7 @@
     "webpack": "gulp webpack",
     "watch": "gulp watch"
   },
-  "dependencies": {}
+  "dependencies": {
+    "clone": "^2.1.1"
```
Tyriar: We generally prefer to roll our own solution for non dev dependencies for performance and security reasons. [This is a lot of code](https://github.com/pvorb/clone/blob/master/clone.js) to include when we could have quite a small function to handle the same thing (clone properties, go an extra level deep for `theme`).

LucianBuzzo: Can do.
I think it would be worth adding the policy for 3rd party modules to https://github.com/xtermjs/xterm.js/blob/master/CONTRIBUTING.md

Tyriar: 👍 I'll handle in https://github.com/xtermjs/xterm.js/issues/1328

LucianBuzzo:
@Tyriar This PR is ready for another review when you have time

src/Terminal.test.ts
```diff
@@ -20,12 +20,13 @@ class TestTerminal extends Terminal {
 
 describe('term.js addons', () => {
   let term: TestTerminal;
+  let termOptions = {
```
Tyriar: const?

src/utils/Clone.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export const clone = <T>(val: T): T => {
+  if (typeof val !== 'object') {
+    return val;
+  }
+
+  const clonedObject: any = {};
+
+  for (const key in val) {
+    clonedObject[key] = clone(val[key]);
```
Tyriar: I think you can set the theme in the ctor as well so we need to go 2 levels deep

src/Terminal.test.ts
LucianBuzzo: Good catch

src/utils/Clone.ts
LucianBuzzo: I'll double check this and I'll add a test to ensure that values are cloned at that depth.

src/utils/Clone.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright (c) 2016 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+/*
+ * A simple utility for cloning values
+ */
+export const clone = <T>(val: T): T => {
+  if (typeof val !== 'object') {
```
mofux: This will allow `null` and `[]` to pass, e.g. `typeof null` and `typeof []` returns `object`. This means that `null` property values would get converted to `{}` which doesn't seem correct.

LucianBuzzo: I'll add a test case for this as well

LucianBuzzo:
@Tyriar @mofux I've addressed your review comment

src/utils/Clone.ts
```diff
@@ -11,9 +11,16 @@ export const clone = <T>(val: T): T => {
...
   }
 
-  const clonedObject: any = {};
+  // cloning null always returns null
+  if (val === null) {
+    return null;
+  }
+
+  // If we're cloning an array, use an array as the base, otherwise use an object
+  const clonedObject: any = Object.prototype.toString.call(val) === '[object Array]' ? [] : {};
```
mofux: Let's use `Array.isArray(val)` instead:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

LucianBuzzo: Ok sure, I'll push the change now

src/utils/Clone.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (val === null) {
+    return null;
+  }
+
+  // If we're cloning an array, use an array as the base, otherwise use an object
+  const clonedObject: any = Array.isArray(val) ? [] : {};
+
+  for (const key in val) {
+    // Recursively clone eack item
+    clonedObject[key] = clone(val[key]);
```
Tyriar: This will blow up with a recursive reference, an optional max depth param on the function should prevent this possible misuse.

LucianBuzzo: @Tyriar What do you recommend for max depth? It doesn't have to be crazy as we're only really using this for the options. Perhaps 5 levels?

Tyriar: I don't think we have any settings deeper than 2 so that's all we need at least in this case. Maybe make it a mandatory parameter and explicitly pass in 2 for the options case?

LucianBuzzo: My preference here would be to have an optional second parameter for depth, that defaults to `5`, which makes the utility nice and usable.

Tyriar: @LucianBuzzo sounds good 👍

LucianBuzzo: Cool, I've implemented this and added some additional tests

Tyriar:
screen shot 2018-03-23 at 4 11 53 pm

👌

src/Terminal.test.ts
```diff
@@ -43,6 +44,13 @@ describe('term.js addons', () => {
     };
   });
 
+  it('should not mutate the options parameter', () => {
+    assert.deepEqual(termOptions, {
+      cols: INIT_COLS,
```
Tyriar: I think you may need to call `term.setOption('cols')` or something here to add another level of verification?

LucianBuzzo: Good idea!

LucianBuzzo:
@Tyriar Oops, looks like my editor setup isn't checking tslint errors :(
 What do you think of running the `lint` script as part of the `test` script, so it becomes `npm run lint && gulp test` ?

LucianBuzzo:
@Tyriar I've added the `setOption` call to the the terminal test and rebased against master, it should be good to merge now 👍

Tyriar:
> What do you think of running the lint script as part of the test script, so it becomes npm run lint && gulp test ?

Yes, recent PRs have made it pretty clear we should be doing this. Using the && approach in `scripts` isn't good enough though as it won't work on Windows. Created https://github.com/xtermjs/xterm.js/issues/1353

Tyriar:
Looks like the other lint failure is unrelated.

PR #1327, opened 2018-03-12
Modularize the character atlas system, add a LRU-cache based dynamic character atlas

### LINKED ISSUE #1294: Use a dynamic cache for the character texture atlas

bgw:
I've split this into a series of small commits, which should hopefully make reviewing this easier.

This introduces a character atlas API (defined by BaseCharAtlas), and three implementations of that API (static, dynamic, and none). The static character atlas is default for now, but you can switch between implementations using the `experimentalCharAtlas` configuration option, or with a dropdown in the demo webpage.

The new DynamicCharAtlas implementation uses a fixed-size texture and an ordered map to maintain an LRU cache of single-width characters in different colors and with different backgrounds. After using it for a while, you end up with an atlas that looks something like this:

When scrolling through vim with the dynamic character atlas, we can see that almost no time is spent calling `fillText`, and that most of the time is spent in calls to `drawImage` and `fillRect`:

Whereas the opposite is generally true for the static char atlas:

There's more performance overhead involved in creating and managing the LRU cache, but in practice it should cover more glyphs, so performance will depend on what kind of text you're rendering.

I've got more optimizations I'm currently implementing and/or playing with that reduce the cost of drawing and the overhead of maintaining the cache, but I think the current implementation is usable and competitive with the static atlas for most use-cases.

This does add a dependency on `Map` (and an implementation of `Symbol.iterator` to traverse it), but that's only if you enable the dynamic atlas. If you don't enable the dynamic atlas, you shouldn't need `Map` (In theory; I haven't tested this claim).

I'm considering replacing `Map` with a custom ordered map implementation, since the act of deleting and setting entries on the Map with every cache read (to mark it as used and move it to the end of the Map) ends up being rather expensive, and appears to slow down over time (on V8 at least?). I think I can probably devise a data structure that works better for our access patterns.

Fixes #1294

### DIFF
--- a/src/renderer/atlas/BaseCharAtlas.ts
+++ b/src/renderer/atlas/BaseCharAtlas.ts
@@ -1,0 +1,46 @@
+/**
+ * Copyright (c) 2017 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { IGlyphIdentifier } from './Types';
+
+export default abstract class BaseCharAtlas {
+  private _didWarmUp: Promise<any>;
+
+  /**
+   * Perform any work needed to warm the cache before it can be used. May be called multiple times.
+   * Implement _doWarmUp instead if you only want to get called once.
+   */
+  public warmUp(): Promise<any> {
+    if (this._didWarmUp == null) {
+      this._didWarmUp = this._doWarmUp();
+    }
+    return this._didWarmUp;
+  }
+
+  /**
+   * Perform any work needed to warm the cache before it can be used. Used by the default
+   * implementation of warmUp(), and will only be called once.
+   */
+  protected _doWarmUp(): Promise<any> {
+    return Promise.resolve();
+  }
+
+  /**
+   * May be called before warmUp finishes, however it is okay for the implementation to
+   * do nothing and return false in that case.
+   *
+   * @param ctx Where to draw the character onto.
+   * @param glyph Information about what to draw
+   * @param x The position on the context to start drawing at
+   * @param y The position on the context to start drawing at
+   * @returns The success state. True if we drew the character.
+   */
+  public abstract draw(
+    ctx: CanvasRenderingContext2D,
+    glyph: IGlyphIdentifier,
+    x: number,
+    y: number,
+  ): boolean;
+}
--- a/src/renderer/atlas/BaseCharAtlas.ts
+++ b/src/renderer/atlas/BaseCharAtlas.ts
@@ -6,26 +6,24 @@
 import { IGlyphIdentifier } from './Types';
 
 export default abstract class BaseCharAtlas {
-  private _didWarmUp: Promise<any>;
+  private _didWarmUp: boolean = false;
 
   /**
    * Perform any work needed to warm the cache before it can be used. May be called multiple times.
    * Implement _doWarmUp instead if you only want to get called once.
    */
-  public warmUp(): Promise<any> {
-    if (this._didWarmUp == null) {
-      this._didWarmUp = this._doWarmUp();
+  public warmUp(): void {
+    if (!this._didWarmUp) {
+      this._doWarmUp();
+      this._didWarmUp = true;
     }
-    return this._didWarmUp;
   }
 
   /**
    * Perform any work needed to warm the cache before it can be used. Used by the default
    * implementation of warmUp(), and will only be called once.
    */
-  protected _doWarmUp(): Promise<any> {
-    return Promise.resolve();
-  }
+  protected _doWarmUp(): void { }
 
   /**
    * May be called before warmUp finishes, however it is okay for the implementation to
--- a/src/renderer/atlas/BaseCharAtlas.ts
+++ b/src/renderer/atlas/BaseCharAtlas.ts
@@ -26,6 +26,15 @@ export default abstract class BaseCharAtlas {
   protected _doWarmUp(): void { }
 
   /**
+   * Called when we start drawing a new frame.
+   *
+   * TODO: We rely on this getting called by TextRenderLayer. This should really be called by
+   * Renderer instead, but we need to make Renderer the source-of-truth for the char atlas, instead
+   * of BaseRenderLayer.
+   */
+  public beginFrame(): void { }
+
+  /**
    * May be called before warmUp finishes, however it is okay for the implementation to
    * do nothing and return false in that case.
    *
--- a/src/renderer/atlas/CharAtlasCache.ts
+++ b/src/renderer/atlas/CharAtlasCache.ts
@@ -8,12 +8,14 @@ import { IColorSet } from '../Types';
 import { ICharAtlasConfig } from '../../shared/atlas/Types';
 import { generateConfig, configEquals } from './CharAtlasUtils';
 import BaseCharAtlas from './BaseCharAtlas';
+import DynamicCharAtlas from './DynamicCharAtlas';
 import NoneCharAtlas from './NoneCharAtlas';
 import StaticCharAtlas from './StaticCharAtlas';
 
 const charAtlasImplementations = {
   'none': NoneCharAtlas,
   'static': StaticCharAtlas,
+  'dynamic': DynamicCharAtlas,
 };
 
 interface ICharAtlasCacheEntry {
--- a/src/renderer/atlas/CharAtlasCache.ts
+++ b/src/renderer/atlas/CharAtlasCache.ts
@@ -4,18 +4,21 @@
  */
 
 import { ITerminal } from '../../Types';
+import BaseCharAtlas from './BaseCharAtlas';
+import StaticCharAtlas from './StaticCharAtlas';
 import { IColorSet } from '../Types';
 import { ICharAtlasConfig } from '../../shared/atlas/Types';
-import { generateCharAtlas } from '../../shared/atlas/CharAtlasGenerator';
 import { generateConfig, configEquals } from './CharAtlasUtils';
 
 interface ICharAtlasCacheEntry {
-  bitmap: HTMLCanvasElement | Promise<ImageBitmap>;
+  atlas: BaseCharAtlas;
   config: ICharAtlasConfig;
+  // N.B. This implementation potentially holds onto copies of the terminal forever, so
+  // this may cause memory leaks.
   ownedBy: ITerminal[];
 }
 
-let charAtlasCache: ICharAtlasCacheEntry[] = [];
+const charAtlasCache: ICharAtlasCacheEntry[] = [];
 
 /**
  * Acquires a char atlas, either generating a new one or returning an existing
@@ -23,7 +26,12 @@ let charAtlasCache: ICharAtlasCacheEntry[] = [];
  * @param terminal The terminal.
  * @param colors The colors to use.
  */
-export function acquireCharAtlas(terminal: ITerminal, colors: IColorSet, scaledCharWidth: number, scaledCharHeight: number): HTMLCanvasElement | Promise<ImageBitmap> {
+export function acquireCharAtlas(
+  terminal: ITerminal,
+  colors: IColorSet,
+  scaledCharWidth: number,
+  scaledCharHeight: number,
+): BaseCharAtlas {
   const newConfig = generateConfig(scaledCharWidth, scaledCharHeight, terminal, colors);
 
   // Check to see if the terminal already owns this config
@@ -32,7 +40,7 @@ export function acquireCharAtlas(terminal: ITerminal, colors: IColorSet, scaledC
     const ownedByIndex = entry.ownedBy.indexOf(terminal);
     if (ownedByIndex >= 0) {
       if (configEquals(entry.config, newConfig)) {
-        return entry.bitmap;
+        return entry.atlas;
       } else {
         // The configs differ, release the terminal from the entry
         if (entry.ownedBy.length === 1) {
@@ -51,22 +59,18 @@ export function acquireCharAtlas(terminal: ITerminal, colors: IColorSet, scaledC
     if (configEquals(entry.config, newConfig)) {
       // Add the terminal to the cache entry and return
       entry.ownedBy.push(terminal);
-      return entry.bitmap;
+      return entry.atlas;
     }
   }
 
-  const canvasFactory = (width: number, height: number) => {
-    const canvas = document.createElement('canvas');
-    canvas.width = width;
-    canvas.height = height;
-    return canvas;
-  };
-
   const newEntry: ICharAtlasCacheEntry = {
-    bitmap: generateCharAtlas(window, canvasFactory, newConfig),
+    atlas: new StaticCharAtlas(
+      document,
+      newConfig,
+    ),
     config: newConfig,
-    ownedBy: [terminal]
+    ownedBy: [terminal],
   };
   charAtlasCache.push(newEntry);
-  return newEntry.bitmap;
+  return newEntry.atlas;
 }
--- a/src/renderer/atlas/CharAtlasCache.ts
+++ b/src/renderer/atlas/CharAtlasCache.ts
@@ -4,11 +4,17 @@
  */
 
 import { ITerminal } from '../../Types';
-import BaseCharAtlas from './BaseCharAtlas';
-import StaticCharAtlas from './StaticCharAtlas';
 import { IColorSet } from '../Types';
 import { ICharAtlasConfig } from '../../shared/atlas/Types';
 import { generateConfig, configEquals } from './CharAtlasUtils';
+import BaseCharAtlas from './BaseCharAtlas';
+import NoneCharAtlas from './NoneCharAtlas';
+import StaticCharAtlas from './StaticCharAtlas';
+
+const charAtlasImplementations = {
+  'none': NoneCharAtlas,
+  'static': StaticCharAtlas,
+};
 
 interface ICharAtlasCacheEntry {
   atlas: BaseCharAtlas;
@@ -64,7 +70,7 @@ export function acquireCharAtlas(
   }
 
   const newEntry: ICharAtlasCacheEntry = {
-    atlas: new StaticCharAtlas(
+    atlas: new charAtlasImplementations[terminal.options.experimentalCharAtlas](
       document,
       newConfig,
     ),
--- a/src/renderer/atlas/CharAtlasUtils.ts
+++ b/src/renderer/atlas/CharAtlasUtils.ts
@@ -17,6 +17,7 @@ export function generateConfig(scaledCharWidth: number, scaledCharHeight: number
     ansi: colors.ansi.slice(0, 16)
   };
   return {
+    type: terminal.options.experimentalCharAtlas,
     devicePixelRatio: window.devicePixelRatio,
     scaledCharWidth,
     scaledCharHeight,
@@ -35,7 +36,8 @@ export function configEquals(a: ICharAtlasConfig, b: ICharAtlasConfig): boolean 
       return false;
     }
   }
-  return a.devicePixelRatio === b.devicePixelRatio &&
+  return a.type === b.type &&
+      a.devicePixelRatio === b.devicePixelRatio &&
       a.fontFamily === b.fontFamily &&
       a.fontSize === b.fontSize &&
       a.fontWeight === b.fontWeight &&
--- a/src/renderer/atlas/CharAtlasUtils.ts
+++ b/src/renderer/atlas/CharAtlasUtils.ts
@@ -8,13 +8,16 @@ import { IColorSet } from '../Types';
 import { ICharAtlasConfig } from '../../shared/atlas/Types';
 
 export function generateConfig(scaledCharWidth: number, scaledCharHeight: number, terminal: ITerminal, colors: IColorSet): ICharAtlasConfig {
+  // null out some fields that don't matter
   const clonedColors = {
     foreground: colors.foreground,
     background: colors.background,
     cursor: null,
     cursorAccent: null,
     selection: null,
-    ansi: colors.ansi.slice(0, 16)
+    // For the static char atlas, we only use the first 16 colors, but we need all 256 for the
+    // dynamic character atlas.
+    ansi: colors.ansi,
   };
   return {
     type: terminal.options.experimentalCharAtlas,
--- a/src/renderer/atlas/CharAtlasUtils.ts
+++ b/src/renderer/atlas/CharAtlasUtils.ts
@@ -17,7 +17,7 @@ export function generateConfig(scaledCharWidth: number, scaledCharHeight: number
     selection: null,
     // For the static char atlas, we only use the first 16 colors, but we need all 256 for the
     // dynamic character atlas.
-    ansi: colors.ansi,
+    ansi: colors.ansi.slice(0, 16),
   };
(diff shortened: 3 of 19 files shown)


src/renderer/atlas/StaticCharAtlas.ts
```diff
@@ -0,0 +1,100 @@
...
+    // ImageBitmap's draw about twice as fast as from a canvas
+    const charAtlasCellWidth = this._config.scaledCharWidth + CHAR_ATLAS_CELL_SPACING;
+    const charAtlasCellHeight = this._config.scaledCharHeight + CHAR_ATLAS_CELL_SPACING;
+
+    // Apply alpha to dim the character
+    if (glyph.dim) {
+      ctx.globalAlpha = DIM_OPACITY;
+    }
+
+    // Draw the non-bold version of the same color if bold is not enabled
```
bgw: I don't know what we should do here. The old implementation did this, but it's a bit tricker for me to do it here, since I don't have a reference to `terminal`; I'd need to add it to the atlas config.

It doesn't look like the `_drawUncachedChar` codepath supports this, so as-is, this rendering behavior is inconsistent. Either we should make sure it's consistent everywhere, or get rid of it.

IMO, it doesn't seem worth the effort to implement since this is a deprecated option.

src/renderer/atlas/BaseCharAtlas.ts
```diff
@@ -0,0 +1,46 @@
...
+import { IGlyphIdentifier } from './Types';
+
+export default abstract class BaseCharAtlas {
+  private _didWarmUp: Promise<any>;
+
+  /**
+   * Perform any work needed to warm the cache before it can be used. May be called multiple times.
+   * Implement _doWarmUp instead if you only want to get called once.
+   */
+  public warmUp(): Promise<any> {
```
Tyriar: For IE11 can a native `Promise` polyfill just be loaded before xterm.js is?

src/renderer/atlas/CharAtlasCache.ts
```diff
@@ -51,22 +67,18 @@ export function acquireCharAtlas(terminal: ITerminal, colors: IColorSet, scaledC
...
-  const canvasFactory = (width: number, height: number) => {
-    const canvas = document.createElement('canvas');
-    canvas.width = width;
-    canvas.height = height;
-    return canvas;
-  };
-
   const newEntry: ICharAtlasCacheEntry = {
-    bitmap: generateCharAtlas(window, canvasFactory, newConfig),
+    atlas: new charAtlasImplementations[terminal.options.experimentalCharAtlas](
```
Tyriar: What do you think about pulling the dynamic cache into an addon for now and monkeypatching some factory method or something to inject while it's still being evaluated?

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -0,0 +1,209 @@
...
+}
+
+function getGlyphCacheKey(glyph: IGlyphIdentifier): GlyphCacheKey {
+  return `${glyph.bg}_${glyph.fg}_${glyph.bold ? 0 : 1}${glyph.dim ? 0 : 1}${glyph.char}`;
+}
+
+export default class DynamicCharAtlas extends BaseCharAtlas {
+  // An ordered map that we're using to keep track of where each glyph is in the atlas texture.
+  // It's ordered so that we can determine when to remove the old entries.
+  private _cacheMap: Map<GlyphCacheKey, IGlyphCacheValue> = new Map();
```
Tyriar: Do you know if you are using any `Map` APIs that aren't supported in IE11's implementation? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -0,0 +1,209 @@
...
+ * Copyright (c) 2017 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { DIM_OPACITY, IGlyphIdentifier, INVERTED_DEFAULT_COLOR } from './Types';
+import { ICharAtlasConfig } from '../../shared/atlas/Types';
+import BaseCharAtlas from './BaseCharAtlas';
+import { clearColor } from '../../shared/atlas/CharAtlasGenerator';
+
+// In practice we're probably never going to exhaust a texture this large. For debugging purposes,
```
Tyriar: I would have thought exhausting the texture is a real concern? Some things that should put more load on the cache:

- CJK characters - I noticed reading through that only ascii is cached, was this just done to simplify the step to the dynamic cache or is there a reason we shouldn't do this? Any strategy for supporting wide chars like CJK/emoji?
- Background colors (are these cached currently?)
- Truecolor (in the future https://github.com/xtermjs/xterm.js/issues/484) 
- High DPI screen (devicePixelRatio of 2 will x4 the pixels in the glyph

src/renderer/atlas/StaticCharAtlas.ts
Tyriar: I guess we'll mark moving to the dynamic cache as the default as v3 to allow breaking changes?

tsconfig.json
```diff
@@ -2,6 +2,14 @@
   "compilerOptions": {
     "module": "commonjs",
     "target": "es5",
+    "lib": [
+      "DOM",
+      "ES5",
+      "ScriptHost",
```
Tyriar: What does ScriptHost and DOM do? We were referencing DOM stuff before, no?

src/renderer/atlas/StaticCharAtlas.ts
bgw: This isn't an issue with the dynamic atlas. It's an bug with the drawing behavior of the static atlas being inconsistent with `_drawUncachedChar`, and I think this is a bug on master.

So I don't know if I should fix the bug, or if so, how I should fix the bug. My proposal is that we should delete the logic and make `StaticCharAtlas` behave like `_drawUncachedChar` does.

src/renderer/atlas/BaseCharAtlas.ts
bgw: As I understood it, xterm already required a Promise polyfill to be loaded.

https://github.com/xtermjs/xterm.js/blob/ed7741e00610747194ecd6e11224d03e864f8563/src/shared/atlas/CharAtlasGenerator.ts#L96

src/renderer/atlas/CharAtlasCache.ts
bgw: Seems reasonable. We'll still probably need/want to add the BaseCharAtlas stuff though, so we have a sane way to hook into the atlas/drawing system.

src/renderer/atlas/DynamicCharAtlas.ts
bgw: I have to use iterators to extract the first element from the map. It's theoretically possible to use `.forEach` for that, but it involves throwing and catching an exception to avoid iterating over the entire map, which is both awkward and messy.

I'm going to take a stab at replacing Map with a custom ordered map implementation that fits our use-case better. Map is eating up about 10% of each frame's execution time, and I think a large part of that is the awkward ways that we need to use it's API to manipulate insertion order.

src/renderer/atlas/DynamicCharAtlas.ts
bgw: > I would have thought exhausting the texture is a real concern?

Exhausting the texture shouldn't generally be much of a concern. If we do exhaust it, we can evict stale entries.

You need at least enough space to store each unique glyph that's currently drawn on the screen once, plus whatever is likely to be rendered on the next frame. Most people are only going to be using a small set of characters (usually Latin characters) in a small set of colors at the same time. Even if you have truecolor support, your editor/tool is only going to use so many colors at once.

Most time spent drawing from the cache is probably going to be spent either re-drawing parts of the screen that haven't changed (TextRenderLayer is redrawing too much stuff right now, but that's out of scope here), or dealing with scrolling of some content.

So yes, there are contrived ways to exhaust the cache every frame (use a different bg/fg combination for every cell on screen), and yes, we'd start thrashing and pay a large performance penalty for that, but I think it's actually pretty hard to do in practice.

It might make sense to re-evaluate this at some point, and make it larger, configurable, or maybe adjust it based on how frequently we're evicting elements, but a magic number keeps this simpler for the initial version.

> - CJK characters - I noticed reading through that only ascii is cached, was this just done to simplify the step to the dynamic cache or is there a reason we shouldn't do this? Any strategy for supporting wide chars like CJK/emoji?

Yes, I'm very intentionally only supporting single-width characters for now, and ascii is a convenient approximation of that. We can tackle double-width characters later, but that seemed like a lot more work, and is a lower priority for me personally. At least this shouldn't regress in performance on those characters.

There's a couple different ways to pack glyphs onto a texture. One is to pack the glyphs tightly using a bin-packing algorithm:



This can make writes to the atlas slower (bin-packing), and more importantly, there's no clear strategy for evicting/replacing glyphs when we run out of space later on. This might be a good approach for the static character atlas to take at some point.

The approach I took was to pack each glyph on a fixed grid. This makes cache eviction and replacement easy (just draw over and re-use that cell). However, this means that I need a defined cell width and height. I could make every cell double-width, but that wastes half of the texture space for most glyphs.

If we want to cache double-width characters, I think it makes sense to have two textures, one for single-width characters and another for double-width characters.

It's also possible to implement an LRU-like cache with two textures:
1. When drawing a glyph, cache it to the first texture, until we run out of space. Once we're out of space,
2. When drawing a glyph, see if it's in the first texture but not the second texture. If it's not in the second texture, copy it from the first texture to the second texture. Repeat this until we're out of space in the second texture.
3. Swap the first and second texture. Clear the new "second" texture. Repeat from step 2.

Old content is always kept in the old texture, and it isn't re-used soon enough, it gets wiped with the rest of that old texture.

I personally like this algorithm ([It's kind of like a copying garbage collector!](http://www.cs.cornell.edu/courses/cs312/2003fa/lectures/sec24.htm)), and it doesn't require overwriting glyphs in the texture, so this could support variable-width glyphs cleanly, but it also require a lot more copying of data between textures, so I opted not to follow this approach for v1. I might change my mind later (especially with the `Map` performance issue I've run into).

> - Background colors (are these cached currently?)

Yes, we draw the background, and then the foreground so that sub-pixel AA works, before removing most of the background color.

If any attribute of `IGlyphIdentifier` changes, that results in a new cache entry.

> - Truecolor (in the future #484)

Our performance is relative to how many colors you're using on the screen at once, not how many you could theoretically draw on the screen at once. My goal is to make normal use-cases fast, and most people use true color for things like vim's syntax highlighting. I don't care about optimizing for playing video content with VLC through AALib.

> - High DPI screen (devicePixelRatio of 2 will x4 the pixels in the glyph)

Yes, worth considering, but I had no performance issues using vim and tmux with a 512x512 atlas on my 1440p screen, so I figured I'd double that, and we'd be fine.

---

TL;DR: If you want to bump this to 2048x2048 or something for now, I'd be fine with that, but we need to pick some number.

tsconfig.json
bgw: These are just the default libraries for the ES5 compilation target.

> Note: If `--lib` is not specified a default list of librares are injected. The default libraries injected are:
> ► For `--target ES5`: `DOM,ES5,ScriptHost`
> ► For `--target ES6`: `DOM,ES6,DOM.Iterable,ScriptHost`
>
> — https://www.typescriptlang.org/docs/handbook/compiler-options.html

I might actually be able to omit these. I assumed I couldn't, but I should check.

src/renderer/atlas/BaseCharAtlas.ts
Tyriar: I don't think it will get to that point on IE as it doesn't support `createImageBitmap`

src/renderer/atlas/CharAtlasCache.ts
Tyriar: 👍

src/renderer/atlas/DynamicCharAtlas.ts
Tyriar: What I was thinking is if we can maybe advise IE11 users to load a map polyfill themselves as it's well supported except for IE.

chabou:
@bgw Just tried `Hyper` with your work (commit 7c3a30f).
This is really amazing 😮 
Large `vim` with some background color is really faster than before...

Thank you so much for your hard work 🙏

Tyriar:
I found a regression with the dynamic cache that I'd like to understand. The following script runs much slower with the dynamic cache turned on:

```bash
for clbg in {40..47} {100..107} 49 ; do
  #Foreground
  for clfg in {30..37} {90..97} 39 ; do
    #Formatting
    for attr in 0 1 2 4 5 7 ; do
      #Print the result
      echo -en "\x1b[${attr};${clbg};${clfg}m ^[${attr};${clbg};${clfg}m \x1b[0m"
    done
    echo #Newline
  done
done
```

Dynamic cache cold:

screen shot 2018-03-18 at 7 23 08 am

Dynamic cache warm:

screen shot 2018-03-18 at 7 23 32 am

Static cache:

screen shot 2018-03-18 at 7 24 46 am

I notice that the warm dynamic cache keeps drawing to the cache. Upon further inspection it appears to be happening because the cache is full:

image

Some thoughts:

- Maybe we should treat the space character special so it doesn't consume a cell in the atlas for every single color (could waste up to 256 cells currently, more when truecolor is supported).
- Even using a texture of 2000x2000 doesn't fit all the characters printed by that script, I think we need some way of falling back to uncached when this happens as the continual draw to and frame the cache calls degrade FPS/scrolling so much.

chabou:
Another regression that you may already know (but blocking hyper integration as is): It doesn't work with non-opaque background colors.

Opaque black:
capture d ecran 2018-03-18 a 18 23 51

Black with 0.1 alpha
capture d ecran 2018-03-18 a 18 23 38

bgw:
@Tyriar Yes, this is a pathological worst-case. I can look into heuristics to disable the dynamic atlas if it's thrashing, but I don't think this is a common use-case, so I was intentionally was ignoring it.

I'd rather not special-case certain characters, because that's making an assumption about the font you're using (though maybe it's not an unreasonable assumption), and that only partially solves this problem.

@chabou Good catch, should be easy to fix.

bgw:
@chabou The transparency issues should be fixed by 1e68f8b4f7e0ab17548389c4ec9659c27c05bf1d.

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -0,0 +1,226 @@
...
+  // The texture that the atlas is drawn to
+  private _cacheCanvas: HTMLCanvasElement;
+  private _cacheCtx: CanvasRenderingContext2D;
+
+  // A couple temporary canvases that glyphs are drawn to before being transfered to the atlas.
+  //
+  // We'll use the a ctx without alpha when possible, because that will preserve subpixel RGB
+  // anti-aliasing, and we'll fall back to a canvas with alpha when we have to.
+  private _tmpCtx: CanvasRenderingContext2D;
+  private _tmpCtxWithAlpha: CanvasRenderingContext2D;
```
Tyriar: Won't we only ever need one of the other unless options change? Elsewhere we only maintain a single canvas based on the `allowTransparency` option

src/renderer/atlas/CharAtlasCache.ts
bgw: I've spent a little bit of time trying to pull DynamicCharAtlas off into a separate addon, but it's a bit painful given the number of shared types and utility functions, I don't feel like it provides much benefit since we'd still need to leave a large amount of the changed code in place to decouple the renderer and atlas, and it would make it more difficult for us to make DynamicCharAtlas the primary atlas later on, since we'd need to re-integrate it.

How important is this to you?

src/renderer/atlas/DynamicCharAtlas.ts
bgw: Yes, if `allowTransparency` is false, we'll never need `_tmpCtxWithAlpha`. However, if `allowTransparency` is true, I'd like to hold onto two ctx objects so that we can draw characters on an opaque background with subpixel AA when possible. This is, however, inconsistent with the `_drawUncachedChar` codepath, which likely won't use subpixel AA anywhere.

Allocating both in the `allowTransparency: false` case is wasteful, but it's pretty small, and it kept the code simpler, so I didn't think it mattered.

If you'd like me to use a single canvas, I can do that pretty easily, as long as you're aware of the tradeoff this makes.

(I don't have much personal preference, since I don't use `allowTransparency`)

chabou:
@bgw I tried with you last commit (d00f2b4). Chars are ok, but they have an opaque background:
capture d ecran 2018-03-20 a 10 08 18

bgw:
Figured out the issue: hyper is passing in [`'transparent'`](https://github.com/zeit/hyper/blob/269bb8e65feb68df248eb7f9d8b3d5337e1c31a2/lib/components/term.js#L22) as the background color.

I assumed that because [`clearColor`](https://github.com/xtermjs/xterm.js/blob/4f82cf26f055a3b13cd7b9030203ecf67c339771/src/shared/atlas/CharAtlasGenerator.ts#L114) assumes hex-formatted `#rrbbgg` colors, that I should use `#rrbbggaa` to represent alpha colors.

It's really just a fluke that `transparent` works with the existing character atlas, since it parses to a bunch of `NaN` values, which causes `clearColor` to be a no-op. I'll see if I can make xterm's color parsing either stricter or more robust, but in the meantime, you can try making hyper use `#00000000` as the background color instead of `transparent`.

chabou:
Ok I found the cause.
In fact, `#RRGGBBAA` hex notation is not compatible with canvas.
You have to put `tansparent` or `rgba(0,0,0,0)` here: https://github.com/bgw/xterm.js/blob/d00f2b42988971fcd30d9d070121f78ce40c51cd/src/renderer/atlas/DynamicCharAtlas.ts#L168

I tested and it works great 🎉

src/renderer/atlas/DynamicCharAtlas.ts
Tyriar: > I'd like to hold onto two ctx objects so that we can draw characters on an opaque background with subpixel AA when possible.

I actually liked that `allowTransparency` was also a switch to enable or disable SPAA completely. I think if we try to do SPAA for some cells and not others then the font might look inconsistent and weird?

bgw:
Ah, it looks like `#RRGGBBAA` got added to Chrome 63, but Hyper is based on 59. Regardless, I don't think IE11 supports this, so it's also not a good choice because of that. :tired_face:

I'm reworking how some of the color parsing works, so that we parse and store colors in 32-bit uints from arbitrary CSS color strings (instead of #RRBBGG format) which should make this stuff saner. Hopefully I'll be able to get that PR up tonight.

Since @Tyriar wants `allowTransparency` to disable SPAA universally, that'll also probably fix this specific issue, since I shouldn't need to parse color values anymore.

bgw:
@Tyriar 83cf279dae291dd72db47b9835fa053939ecf3cd should mostly fix the performance regression you found.

I'll rebase or merge this onto master once #1346 lands. I think the current set of comments have been addressed with the exception of:

- Should we remove bold colors from bold glyphs when `enableBold` is turned off? [See this comment thread.](https://github.com/xtermjs/xterm.js/pull/1327/files/be810f87aaf42f6bef958bd4258f6fed3d1c47c6#r173680499) I believe the the current implementation in master is inconsistent between the cached and uncached codepaths.

- How important is it to split this off into an addon? I started working towards that, and it appears that it would require a lot of work for little gain. [See this comment thread](https://github.com/xtermjs/xterm.js/pull/1327/files/be810f87aaf42f6bef958bd4258f6fed3d1c47c6#r173900745).

chabou:
FYI: we released a new Hyper canary release (v2.0.0-canary.15) with this PR merged (9e54b8646e8e1fbfd351e6483bddf63d6b6c569a) 🎉

Tyriar:
@bgw Thanks! I'll check it out as soon as I can, we're in the test/release phase of VS Code so I have limited time for the next few days.

Tyriar:
I notice that running `tree` on Windows will continually draw the pipe chars uncached:

image

image

Log at the start of `_drawUncachedChar`:

image

Any idea why these are not being cached?

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    // copy the data from imageData to _cacheCanvas
+    const [x, y] = this._toCoordinates(index);
+    // putImageData doesn't do any blending, so it will overwrite any existing cache entry for us
+    this._cacheCtx.putImageData(imageData, x, y);
+
+    return {
+      index,
+      isEmpty,
```
Tyriar: Could we not store anything in the canvas at all if this is the case? Probably a good candidate for deferring until after merged. Let's create an issue if this is the plan.

src/renderer/atlas/CharAtlasUtils.ts
```diff
@@ -8,15 +8,19 @@ import { IColorSet } from '../Types';
...
   const clonedColors = {
     foreground: colors.foreground,
     background: colors.background,
     cursor: null,
     cursorAccent: null,
     selection: null,
-    ansi: colors.ansi.slice(0, 16)
+    // For the static char atlas, we only use the first 16 colors, but we need all 256 for the
+    // dynamic character atlas.
+    ansi: colors.ansi,
```
Tyriar: This property looks like it's being shallow cloned now, so it will change with the theme. I'd prefer not to slice an array this big every time we need a config, perhaps extending `ICharAtlasConfig` for the dynamic cache to add the extra colors is the answer? Also remember the 256 color codes are constant and will never change.

src/renderer/atlas/DynamicCharAtlas.ts
bgw: Yeah, let's defer this. It'd be a small win (most glyphs aren't empty), and I'm not entirely sure it would be worth the potential complexity. I've opened #1389.

chabou:
Is it possible that this PR disables `fontWeight` and `fontWeightBold` options?

Tyriar:
@chabou it looks like fontWeight is still being used and wasn't really touched:

https://github.com/xtermjs/xterm.js/pull/1327/files#diff-0f9aebb75af9d3b431640c5def0cfe48L124

Tyriar:
It looks like this did indeed break `fontWeight`. Running `ls` followed by `term.setOption('fontWeight', 'bold')` turned the ls output bold before, not on this branch.

bgw:
:+1: Good catch. Should be easy-ish to fix.

bgw:
@chabou I've fixed the font weight issue, and merged the latest version of master in.

@Tyriar I'm going to wait until #1391 merges, and then update this branch so that I'm the one to deal with the messiness of that merge, and so that I can make sure the new option works with the dynamic atlas. I think I've addressed all of your comments; do you want to review anything else?

chabou:
This is fantastic! Thank you so much for your hard work :pray:

Issue #1319, opened 2018-03-09
Enable strictNullChecks


Tyriar:
A lot of errors when the flag is enabled in tsconfig, we should slowly chip away at them so we can turn the setting on :slightly_smiling_face:

Tyriar:
Closing this off in favor of #1507

PR #1329, opened 2018-03-12
Use bang operator for array methods

Tyriar:
Part of #1319

---

We can use the bang operator `!` to get rid of the `| undefined` if we're sure that the type is not undefined.


mofux:
LGTM 👍

Issue #1330, opened 2018-03-12
Hard-coded CSS mangles RTL layouts


FGasper:
#701 documents ongoing work toward full support for RTL; however, a separate, related issue is that `xterm.css` includes hard-coded `left` and `right` rules which don’t display correctly on RTL pages:

![image](https://user-images.githubusercontent.com/844867/37313477-c4040944-2625-11e8-83c1-feb492273676.png)

## Details
- Browser and browser version: FF/Mac and Chrome/Mac, current versions
- OS version: macOS High Sierra
- xterm.js version: 3.2.0

### Steps to reproduce

1. Set `dir="rtl"` on the demo’s index.html.
2. Load the demo.

OBSERVE: The page appears to be blank because everything is -9999em to the left.
EXPECTED: The CSS shouldn’t assume left-to-right document layout.

FGasper:
Is there any motion on this? It’s going to be an impediment for our implementation. We can do a local patch, but I would think the fix is desirable outside our own code.

Thanks!

Tyriar:
@FGasper just responded, this should make it into 3.3. I had a huge backlog of notifications that I'm nearly through now.

Tyriar:
Fixed in https://github.com/xtermjs/xterm.js/pull/1350

FGasper:
@Tyriar This is broken again in v3.3.0.

![image](https://user-images.githubusercontent.com/844867/38815986-1fbdccb8-4163-11e8-849f-e5c7ac4574c3.png)

FGasper:
Looks like the fix was bad … :-/ I repeated the testing with b27d5bb08dba9589721d37fc34b04a1fa5437d75 and still see the issue.

FGasper:
@Tyriar Rather than setting -9999px for the various elements that get that treatment, what about a combination of:

color: transparent;
background-color: transparent;
opacity: 0;
height: 0;
width: 0;
z-index: -9999

??

Tyriar:
Do you mean this? https://github.com/xtermjs/xterm.js/blob/a3ce881897818d1c61476620e7d037717ca9a68f/src/xterm.css#L125

FGasper:
@Tyriar That’s what I would have thought, but there’s no class on the element; the styles are directly on the element.

Tyriar:
@FGasper there should be:

[...]

Issue #701, opened 2017-06-13
Support RTL languages


Tyriar:
Downstream issue: https://github.com/Microsoft/vscode/issues/28571

When we enforced unicode character width in https://github.com/sourcelair/xterm.js/issues/467 this broke RTL language characters as they are now rendered in reverse (LTR). We could revert that for RTL character ranges only but we should do the right fix and reverse the strings so they're actually on the character grid as the new selection model relies on all characters lining up perfectly on the grid https://github.com/sourcelair/xterm.js/pull/670

Ideally line reflow https://github.com/sourcelair/xterm.js/issues/622 would be done before this so it's easier to change the contents of multiple lines.

Terminal.app:

![image](https://user-images.githubusercontent.com/2193314/27095747-a3a94b9c-5023-11e7-9541-8673f9fd42f1.png)

VS Code 1.13 (notice sentences are reversed):

![image](https://user-images.githubusercontent.com/2193314/27095742-9fa9139c-5023-11e7-8194-c8dfe4ad3f7e.png)

@mostafa69d @CherryDT a little info on the languages in question would be handy:

1. Where should the strings be flipped.for Hebrew/Arabic/Persian, do I reverse entire continuous sequences of characters in-between ascii characters?
2. How are the characters meant to interact with characters like 0-9 or punctuation?

Useful references:

- http://www.i18nguy.com/temp/rtl.html
- https://en.wikipedia.org/wiki/Arabic_script_in_Unicode
- https://en.wikipedia.org/wiki/Syriac_(Unicode_block)
- https://en.wikipedia.org/wiki/Hebrew_(Unicode_block)

CherryDT:
It is actually a whole lot more complicated and includes statefulness and even mirroring certain characters. I'd say it's a science of its own.

See also:
https://en.wikipedia.org/wiki/Bi-directional_text
https://www.w3.org/International/articles/inline-bidi-markup/uba-basics
https://www.w3.org/International/tutorials/svg-tiny-bidi/ (the initial premise is not related but it explains a few things better than the previous link)
https://github.com/fevangelou/doctype-mirror/tree/master/bidihowto/bidi-support-in-a-ui

mostafa69d:
@Tyriar 
First of all I'm gonna give you a very brief perspective of Arabic and Persian language maybe it help you(I'm not sure if the Hebrew is the same).
[...]

PR #1331, opened 2018-03-13
Make terminal display correctly in RTL layouts.

FGasper:
Issue #1330

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -620,10 +620,13 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this._screenDprMonitor = new ScreenDprMonitor();
     this._screenDprMonitor.setListener(() => this.emit('dprchange', window.devicePixelRatio));
 
+    let isRtl = (getComputedStyle(this._parent).direction === 'rtl');
+
     // Create main element container
     this.element = this._document.createElement('div');
     this.element.classList.add('terminal');
     this.element.classList.add('xterm');
+    this.element.classList.add( isRtl ? 'xterm-rtl' : 'xterm-ltr' );
     this.element.setAttribute('tabindex', '0');
     this._parent.appendChild(this.element);
 
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -66,7 +66,6 @@
      */
     position: absolute;
     opacity: 0;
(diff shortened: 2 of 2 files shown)


Tyriar:
I'm confused, does rtl directional text inside a container reverse the container's positioning? 😕

FGasper:
@Tyriar I’m not sure of the exact mechanics of all of this, but generally I’ve found that RTL just has to reverse left/right positioning directives. So, the `left` CSS styles need to be `right`, and vice-versa.

src/Terminal.ts
```diff
@@ -620,10 +620,13 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this._screenDprMonitor = new ScreenDprMonitor();
     this._screenDprMonitor.setListener(() => this.emit('dprchange', window.devicePixelRatio));
 
+    let isRtl = (getComputedStyle(this._parent).direction === 'rtl');
```
Tyriar: I think this will cause a layout which isn't desirable. Would it be possible to target the styles using a media query like this https://stackoverflow.com/a/26796889/1156119?

FGasper: `:dir(rtl)` only works in Firefox, according to MDN. I’ve not found a reliable, cross-browser CSS-based solution for detecting an individual element’s active directionality.

There is `[dir=rtl]`, but that will match something like:
```
<html dir="rtl">
    <div dir="ltr">…</div>
</html>
```
… which isn’t going to work.

Can you describe your layout concerns in more detail? The JS change as-is on the branch shouldn’t affect the layout since all it does is assign a class.

Tyriar: Could we enforce ltr? The terminal doesn't support rtl text anyway and that might simplify the fix?

Tyriar: By layout I mean a reflow which can cause performance problems

FGasper: Forcing LTR works: issue_1331_rtl_layout

Tyriar: Nice, let's go with that and reevaluate when we tackle https://github.com/xtermjs/xterm.js/issues/701

Tyriar:
Superseded by https://github.com/xtermjs/xterm.js/pull/1350

Issue #1335, opened 2018-03-16
Avoid inline-styling to meet tighter content-security-policy


mareklibra:
If strict content-security-policy is set by a project consuming `xterm.js`,
the use of inline-style is not permitted.

To make use of the `xterm.js` even in projects with higher security demands, like the Cockpit (www.cockpit-project.org), modifications/use of `style` HTML element attribute must be avoided.

This can be partially achieved via use of CSS `class`. 
Or other solution shall be found in case of `class` is is not an option.

Since behavior without `style` will not be probably perfect all the time, this feature is optional while keeping former behavior by default.
The `Terminal` component will be enriched for new `avoidInlineStyling` boolean attribute (`false` by default) which will disable use of `style` attribute within processing of `Terminal` or it's inner code.

## Details
- Browser and browser version:  any
- OS version: any newer supporting CSP
- hosting application enforces use of strict CSP, especially disabling insecure inline-styling

mareklibra:
Preliminary work on this issue has already started, will provide a PR for early feedback shortly.

mareklibra:
If good-enough, the xterm.js is planed to be used in patternfly-react (https://github.com/patternfly/patternfly-react/pull/160), so transitively in multiple other projects, like https://github.com/cockpit-project/cockpit

PR #1336, opened 2018-03-16
Replace inline-styles by external CSS

mareklibra:
In applications using tight Content Security Policy, the use of
inline-styles can be disabled.

This patch introduces new optional `avoidInlineStyling` attribute of
the `Terminal` component, set to `false` by default.

If set to `true`, inline-styles are not applied and replaced by CSS classes
instead whereever possible.
Not all features can be available with this option set, so inline-styles
are still used by default.

Fixes: https://github.com/xtermjs/xterm.js/issues/1335

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -118,7 +118,7 @@ const DEFAULT_OPTIONS: ITerminalOptions = {
   allowTransparency: false,
   tabStopWidth: 8,
   theme: null,
-  rightClickSelectsWord: Browser.isMac
+  rightClickSelectsWord: Browser.isMac,
   // programFeatures: false,
   // focusKeys: false,
 };
@@ -138,7 +138,6 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
   private _viewportElement: HTMLElement;
   private _helperContainer: HTMLElement;
   private _compositionView: HTMLElement;
-  private _charSizeStyleElement: HTMLStyleElement;
 
   private _visualBellTimer: number;
 
@@ -668,8 +667,6 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this._compositionHelper = new CompositionHelper(this.textarea, this._compositionView, this);
     this._helperContainer.appendChild(this._compositionView);
 
-    this._charSizeStyleElement = document.createElement('style');
-    this._helperContainer.appendChild(this._charSizeStyleElement);
     this.charMeasure = new CharMeasure(document, this._helperContainer);
 
     // Performance: Add viewport and helper elements from the fragment
--- a/src/utils/CharMeasure.ts
+++ b/src/utils/CharMeasure.ts
@@ -23,10 +23,7 @@ export class CharMeasure extends EventEmitter implements ICharMeasure {
     this._document = document;
     this._parentElement = parentElement;
     this._measureElement = this._document.createElement('span');
-    this._measureElement.style.position = 'absolute';
-    this._measureElement.style.top = '0';
-    this._measureElement.style.left = '-9999em';
-    this._measureElement.style.lineHeight = 'normal';
+    this._measureElement.classList.add('xterm-char-measure-element');
     this._measureElement.textContent = 'W';
     this._measureElement.setAttribute('aria-hidden', 'true');
     this._parentElement.appendChild(this._measureElement);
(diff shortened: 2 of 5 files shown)


mareklibra:
So far WIP but looking for early feed back, if possible.

src/utils/CharMeasure.ts
```diff
@@ -41,8 +38,12 @@ export class CharMeasure extends EventEmitter implements ICharMeasure {
   }
 
   public measure(options: ITerminalOptions): void {
-    this._measureElement.style.fontFamily = options.fontFamily;
-    this._measureElement.style.fontSize = `${options.fontSize}px`;
+    if (!options.avoidInlineStyling) {
+      this._measureElement.style.fontFamily = options.fontFamily;
+      this._measureElement.style.fontSize = `${options.fontSize}px`;
+    } else {
+      window.console.debug('CharMeasure: use of inline-styling is disabled. Font set by a parent element is used instead of: ', options.fontFamily, ', size: ', options.fontSize);
```
mofux: Is there a particular reason to reference `window` directly?

src/Terminal.ts
```diff
@@ -668,8 +673,10 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this._compositionHelper = new CompositionHelper(this.textarea, this._compositionView, this);
     this._helperContainer.appendChild(this._compositionView);
 
-    this._charSizeStyleElement = document.createElement('style');
-    this._helperContainer.appendChild(this._charSizeStyleElement);
+    if (!this.options.avoidInlineStyling) {
+      this._charSizeStyleElement = document.createElement('style');
+      this._helperContainer.appendChild(this._charSizeStyleElement);
```
mofux: Can you please explain why we need this code? Looks like it is creating an empty style tag that is never used again 🤔

mareklibra: I don't see any other use of the `_helperContainer` than in `CharMeasure` to set font prior measurement.

So, in case the `avoidInlineStyling` is true, the `style` element is not used but is causing CSP failure.

For completeness, if `avoidInlineStyling` is true, the font is expected to be set by an element parenting the `Terminal`. Flexibility is lower, but it works.

src/utils/CharMeasure.ts
mareklibra: nope, removing

mareklibra:
@Tyriar , far from to be comprehensive but some of the issues related to inline-styles: https://www.owasp.org/index.php/Testing_for_CSS_Injection_(OTG-CLIENT-005)

I understand, these changes increase complexity of `xterm.js` codebase. Anyway, they will probably be not so often and the changes might be localized (at least from recent findings).
If this gets in, it would be great to have the CSP aspect on mind when introducing new functionality in the future.

Generally speaking, if the application has high security demands, it's good practice to
- use CSP
- and keep even the inline-styles disabled, not only `script-src` or so.

Considering the focus of `xterm.js` is to provide terminal access to a remote system, security shall be the concern.
There's no doubt, the consuming application plays the key role in enforcing the security requirements.
Anyway its components shall be aligned with the restrictions.

I believe, the CSP is nothing unusual these days, since it has been standardized for (quite) a long time.

mareklibra:
The motivation behind this change is the `Cockpit` project - web-based linux server management tool.
In this application, the javascript is granted with same access rights to the server as if the user is connected via `ssh`. The CSP is one of the building blocks for ensuring the security there.

src/xterm.css
```diff
@@ -117,11 +120,11 @@
     visibility: hidden;
 }
 
-.xterm .xterm-char-measure-element {
-    display: inline-block;
-    visibility: hidden;
```
Tyriar: Are you sure this isn't needed?

Tyriar:
I still don't really buy that this is needed, if you're defining the trusted domains using `script-src` then only trusted code will be able to inject CSS? I think setting trusted domains should prevent all the attacks mentioned in the article you linked to.

I suggest the following instead:

- Use an iframe as mentioned by @mofux to isolate the terminal from the rest of the page
- Maintain a fork for this so the burden of no inline styles doesn't impede the core devs

The only part of the PR I think we should take in is the static char-measure-element styles moving to CSS.

mareklibra:
@mofux , @Tyriar , thanks for your comments. Recent changes seem to be good-enough even for the Cockpit flow, especially after removing the unused `<style>` element.

src/Terminal.ts
```diff
@@ -668,8 +667,6 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this._compositionHelper = new CompositionHelper(this.textarea, this._compositionView, this);
     this._helperContainer.appendChild(this._compositionView);
 
-    this._charSizeStyleElement = document.createElement('style');
```
Tyriar: Thanks for spotting this 👍

PR #1338, opened 2018-03-17
Implement Markers API

### LINKED ISSUE #1325: Marker API
In order for consumers to provide some cool integrations such as activating different link matchers in different parts of the body or jumping to some arbitrary marked line we need a way to track lines in the normal buffer/scrollback.

## Goals

- Provide an API that allows consumers to add "markers" to arbitrary lines that are then tracked within the buffer.
- Don't slow down the terminal; minimize manipulations/GC of adding and maintaining markers.
- Should be an optional feature, it will have minimal O(1) overhead for unless consumers opt in.
- Be well tested.
- Markers will only work in the normal buffer.
- Link matchers will eventually adopt scoping which will use markers, this will allow link matchers to activate only on a subset of the buffer.
- (Future) Allow consumers to define how they visualize markers (or not show them at all) - this will likely involve standardizing a render layer API.

## API

```ts
class Terminal {
    /**
     * Adds a marker on cursor line, allowing it to be tracked.
     * @param cursorOffset The offset from the cursor of the line to be marked.
     */
    addMarker(cursorOffset?: number): Marker;
    
    /**
     * Scrolls to a line within the buffer.
     * @param line The 0-based line index to scroll to.
     */
    scrollToLine(line: number): void;

    /**
     * Selects text in the buffer between 2 lines.
     * @param start The 0-based line index to select from (inclusive).
     * @param end The 0-based line index to select to (inclusive).
     */
    selectLines(start: number, end: number): void;

    /**
     * Gets a list of active markers, generally you should request this every time you need to use
     * this array and not keep a reference as markers are disposed when they reach the end of the
     * scrollback.
     */
    markers: Marker[];
}

interface Marker {
    /**
     * The ID of the marker, this is a unique number generated for each marker and can be used to
     * differentiate markers.
     */
    id: number;

    /**
     * The absolute line of the marker.
     */ 
    line: number;

    /**
     * Whether the marker is disposed.
     */
    isDisposed: boolean;

    /**
     * Disposes the marker.
     */
    dispose(): void;
}
```

## How it works

Trim/splice listeners are setup on the buffer and adjust the position of the marker when a change occurs. Deleting a line will dispose the marker, this can happen through a splice call or from being trimmed off the top of the buffer. Consumers may still have a reference to the marker but it will be marked as `isDisposed` when this happens.

Tyriar:
Fixes #1325


mofux:
Hi @Tyriar! Is this PR still WIP, or is it ready for review? I can see some TODO comments and console.log's in it 🧐. There is also nothing but the test suite that currently uses that API. Do you plan on moving stuff over to the new API later on?

Tyriar:
@mofux ready for review. Initially I'm using this API in VS Code, it would be good for link matchers to adopt it eventually such that sections of the buffer can use different patterns but that's not my focus at the moment.

mofux:
LGTM 👍
As a side note: I like the extensions to the SelectionManager public API (although they don't have to do anything with the MarkerAPI). The MarkerAPI itself, well, I don't like the fact that we provide an experimental API that we don't use internally (yet) - that could easily be written as an addon by listening to the `trim` event of the buffer lines. But I want to leave that decision up to you 😅

Tyriar:
@mofux I considered making it an addon, I think we'll eventually move to relying on markers when we identify lines internally though. The experimental part is to make sure there's no major issues with it that I discover while using it 😉

PR #1340, opened 2018-03-19
Remove unused function

Tyriar:


mofux:
Even if we don't use this method internally, I think it is worth keeping. CircularList is not iterable directly, so having the forEach methods makes sense to me.

Tyriar:
@mofux this is clean up in preparation for https://github.com/xtermjs/xterm.js/issues/622, no need keeping it around if it's not going to be used. The only thing that should be doing something over the entire buffer is reflow, and that was just going to be a loop over length since the list won't change and to make sure it's as fast as possible.

mofux:
Happy to get this merged, we can bring it back if we really have to 👍

Issue #1247, opened 2018-01-25
Improve DOM-based integration tests


Tyriar:
Right now we're on jsdom which has a lot of problems and as a result we don't write that many integration tests. https://github.com/GoogleChrome/puppeteer

bgw:
At `$dayjob`, we use a combination of JSDom and Puppeteer.

Puppeteer is definitely a lot more powerful, but it's also slower, and it's harder to get tests to run reliably. Puppeteer has an async API (it's easy to introduce race conditions) and more stuff that can go wrong. As a result, we've only built tests using Puppeteer for our most important surfaces.

I'd recommend using a combination of both, or just sticking with JSDom.

Tyriar:
@bgw I think your idea of adding node-canvas sounds like the most attractive first step atm

PR #1346, opened 2018-03-22
Allow theming with arbitrary CSS color strings

bgw:
This mostly worked before, but `clearColor` made the assumption that the colors were always in `#RRGGBB` format. See the discussion here: https://github.com/xtermjs/xterm.js/pull/1327#issuecomment-374812842

This changes the internal representation of a color so that we hold onto into an object containing the original css string along with an RGBA value encoded as a 32-bit uint. `clearColor` can then use that RGBA representation.

Additionally, this deduplicates some code between `ColorManager` and `Terminal`. `Terminal` was generating the 256 ansi colors the same way `ColorManager` was, so this just makes `Terminal` use the same constant.

256 colors still works:

and so does the 24-bit color approximation:

The static char atlas looks sane, so `clearColor` is working:

We can change the theme using arbitrary css strings:

And `clearColor` still works properly, but now using a pink background:


mofux:
Really nice PR, this is addressing one of the bigger areas of concern, thanks! 

If I'm reading the code correctly, passing alpha values like `rgba(255,255,255,0.5)` will result in a color with the alpha removed, no? This shouldn't be a big problem in general, except for the selection color that needs some alpha at the moment to not hide the characters underneath.

There's also one test failling:

```
Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package) undefined
      1) "before each" hook for "should fill all colors with values"
```

src/Terminal.ts
```diff
@@ -2233,38 +2233,6 @@ function wasMondifierKeyOnlyEvent(ev: KeyboardEvent): boolean {
  * ANSI color code.
  */
 
-// Colors 0-15 + 16-255
-// Much thanks to TooTallNate for writing this.
-const vcolors: number[][] = (function(): number[][] {
```
Tyriar: I've wanted this gone for a long time 😛

Tyriar:
This looks great, I haven't tested the code but if everything works fine and the tests pass then I'm 👍

bgw:
I've fixed the test, and amended the commit!

> If I'm reading the code correctly, passing alpha values like rgba(255,255,255,0.5) will result in a color with the alpha removed, no?

This parses alpha channels.



> This shouldn't be a big problem in general, except for the selection color that needs some alpha at the moment to not hide the characters underneath.

Selections work as expected. They're partially transparent.

src/renderer/ColorManager.test.ts
```diff
@@ -17,261 +17,272 @@ describe('ColorManager', () => {
     dom = new jsdom.JSDOM('');
     window = dom.window;
     document = window.document;
+    (<any>window).HTMLCanvasElement.prototype.getContext = () => ({
```
bgw: It might make sense to pull in [node-canvas](https://github.com/Automattic/node-canvas) (which JSDom integrates with) as a dev dependency if we want to test more rendering-related codepaths, but this works for now.

Tyriar: I want to improve the testing situation, I was thinking about puppeteer but I'm not sure if that's the right tool for the job thinking about it some more. https://github.com/xtermjs/xterm.js/issues/1247

bgw:
I merged bgw/xterm.js#2 (Block transparent colors without allowTransparency) into this PR, and I'm not aware of any outstanding issues, so I think this is ready to be merged.

I'll do the node-canvas stuff as a separate PR once this stuff lands.

Tyriar:
Thanks @bgw 👍

PR #1350, opened 2018-03-23
Force LTR for the terminal element

FGasper:
Issue #1330: This prevents RTL layouts from mishandlin the
intended layout of the terminal container and children.


Tyriar:
@FGasper cheers 😃

PR #1355, opened 2018-03-27
Handle blob in attach Addon

AlexandrePetit-Pas:
I have added blob handler for the attach Addon.
This allows to handle Docker with ws (works on Docker API v1.35)
I have also added a callback to handle asynchronous FileReader.


IDerr:
Hmm, I think the build failure is not linked to this PR, 
src/Terminal.ts[625, 35]: comment must start with a space
but this PR didn't touch this file

Thanks for your work @AlexandrePetit-Pas

Issue #1164, opened 2017-12-22
Webfont rendering on v3?


vincentwoo:
Playing with v3, it seems that on some browser hard refereshes, xterm doesn't seem to pick up the custom fontFamily I set and renders oddly:

![screenshot 2017-12-21 18 04 02](https://user-images.githubusercontent.com/613320/34282331-5de967bc-e679-11e7-918e-a79e328d06f7.png)

On a normal refresh after:

![screenshot 2017-12-21 18 04 35](https://user-images.githubusercontent.com/613320/34282340-6b04e282-e679-11e7-901d-4364954f7f80.png)

Is there a recommended flow for loading xterm with webfonts?

mofux:
@vincentwoo Good point. Since we use the canvas for rendering, there is no automatic swap-out of fonts anymore. In general you have to make sure that the font is available when your terminal creation code runs. To do that, make sure the terminal creation code is loaded in head or body after the stylesheet that loads the font (see https://stackoverflow.com/questions/9899372/pure-javascript-equivalent-of-jquerys-ready-how-to-call-a-function-when-t). You could also force xterm.js into refreshing its char atlas by setting the fontFamily on the theme once you know that the font is available:

```javascript
term.setOption('theme', { fontFamily: 'YourFontFamily' })
```

@Tyriar Do you think we should handle this case automatically (refreshing the char atlas once the window finished loading)?

vincentwoo:
It's tricky because even if your scripts occur at the end of `<body>` you're not guaranteed that all webfonts have loaded - the browser delays rendering until a character of a given font family is deemed needed. Ideally, xterm would use fallback fonts and refresh as necessary.

Tyriar:
@vincentwoo I think you'll want to call open after the web fonts have finished loading, that would prevent flickering of a different font initially.

Anyone know if there an easy way to tell whether a font has been loaded/changed?

vincentwoo:
It might be useful to have xterm incorporate one of the polyfills mentioned here: https://stackoverflow.com/questions/5680013/how-to-be-notified-once-a-web-font-has-loaded. Right now, anyone who wants to use a webfont with xterm will have to go through something similar, so it might have good utility for users as an addon. I can look into it if you don't have the time.

Tyriar:
[...]

PR #1368, opened 2018-04-04
Addon that refreshes xterm upon webfont load

vincentwoo:
Addresses https://github.com/xtermjs/xterm.js/issues/1164

I'm new to TypeScript, so I'm not sure if I've done everything the right way. I've opted to monkeypatch _setup, because we need to get into the creation flow as soon as possible (and also because only _setup references what the default font options are). It's possible to do something like require a manual call to `loadWebfonts` but I've opted for a more seamless approach.

The addon also doesn't add any more configuration - it assumes that any font families passed in will be webfonts. If they aren't, there may be a momentary flicker but all should continue to work.

### DIFF
--- a/src/addons/webfont/webfont.ts
+++ b/src/addons/webfont/webfont.ts
@@ -1,0 +1,19 @@
+/// <reference path="../../../typings/xterm.d.ts"/>
+
+import { Terminal } from 'xterm';
+
+export function apply(terminalConstructor: typeof Terminal): void {
+  (<any>terminalConstructor.prototype).loadWebfontAndOpen = function (element: HTMLElement): void {
+    const FontFaceObserver = (typeof window === 'object' && (<any>window).FontFaceObserver);
+    if (!FontFaceObserver) {
+      console.warn('FontFaceObserver not available, opening xterm normally!');
+      return this.open(element);
+    }
+    const regular = new FontFaceObserver(this.options.fontFamily).load();
+    const bold = new FontFaceObserver(this.options.fontFamily, { weight: 'bold' }).load();
+
+    regular.constructor.all([regular, bold]).then(() => {
+      this.open(element);
+    });
+  };
+}
--- a/src/addons/webfont/webfont.ts
+++ b/src/addons/webfont/webfont.ts
@@ -12,8 +12,9 @@ export function apply(terminalConstructor: typeof Terminal): void {
     const regular = new FontFaceObserver(this.options.fontFamily).load();
     const bold = new FontFaceObserver(this.options.fontFamily, { weight: 'bold' }).load();
 
-    regular.constructor.all([regular, bold]).then(() => {
+    return regular.constructor.all([regular, bold]).then(() => {
       this.open(element);
+      return this;
     });
   };
 }
--- a/demo/webfont/index.html
+++ b/demo/webfont/index.html
@@ -1,0 +1,20 @@
+<html>
+  <head>
+    <link href="https://fonts.googleapis.com/css?family=Roboto+Mono:400,700" rel="stylesheet">
+    <link rel="stylesheet" href="../../build/xterm.css" />
+    <script src="../../build/xterm.js"></script>
+    <script src="../../build/addons/webfont/webfont.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/fontfaceobserver/2.0.13/fontfaceobserver.standalone.js"></script>
+  </head>
+
+  <body>
+    <div id="term"></div>
+
+    <script>
+      Terminal.applyAddon(webfont)
+      const term = new Terminal({ fontFamily: 'Roboto Mono', lineHeight: 0.9, fontSize: 15 })
+      term.loadWebfontAndOpen(document.getElementById('term'));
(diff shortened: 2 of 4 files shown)


src/addons/webfont/webfont.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const loadedFonts = {}
+
+export function apply(terminalConstructor: typeof Terminal): void {
+  const oldSetup = (<any>terminalConstructor.prototype)._setup;
+
+  (<any>terminalConstructor.prototype)._setup = function (): void {
+    const originalFont = this.options.fontFamily;
+    if (loadedFonts[originalFont]) return oldSetup.call(this);
+    delete this.options.fontFamily;
```
Tyriar: A more explicit `loadWebfonts.load(fontFamily: string)` would be preferable imo and align better with how the other addons work.

Also the font is only ever used after `open` is called. So what if this function was shaped more like this, rather than monkeypatching?

```ts
loadAndOpen = function (fontFamily: string, element: HTMLElement): void {
  Promise.all(...).then(() => this.open(element));
};
```

src/addons/webfont/webfont.ts
```diff
@@ -0,0 +1,26 @@
...
+    oldSetup.call(this);
+
+    const regular = new FontFaceObserver(originalFont).load();
+    const bold = new FontFaceObserver(originalFont, { weight: 'bold' }).load();
+
+    regular.constructor.all([regular, bold]).then(() => {
```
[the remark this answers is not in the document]
vincentwoo: Yes, but without relying on importing a specific Promise library.

package.json
```diff
@@ -88,5 +88,7 @@
     "webpack": "gulp webpack",
     "watch": "gulp watch"
   },
-  "dependencies": {}
+  "dependencies": {
+    "fontfaceobserver": "*"
```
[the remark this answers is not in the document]
vincentwoo: doable. what should behavior be if you try to run the addon and the dependency is not present?

src/addons/webfont/webfont.ts
vincentwoo: One rationale for doing it on _setup is that you may create a term, write to it, and then later render it. If we know you're going to need a webfont, we can get to it before open is actually called and avoid the resize flash.

If that's not as valuable to you as having a less magical thing, I can go ahead and change the surface.

vincentwoo: Also, if we're going to call the endpoint "load and open" we might as well just delay opening the terminal until fonts come in to avoid the reflash as well.

package.json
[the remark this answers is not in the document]
Tyriar: I guess console.warn and open normally?

src/addons/webfont/webfont.ts
Tyriar: > Also, if we're going to call the endpoint "load and open" we might as well just delay opening the terminal until fonts come in to avoid the reflash as well.

That's what my proposal does, it doesn't touch the DOM until the font load promises are finished?

vincentwoo: I'm down. Hang tight.

vincentwoo:
Okay, I've updated the PR. Two things:

1. I've made the surface `loadWebfontAndOpen = function (element: HTMLElement)`, opting not to pass the font as a separate option to the addon. I think it's a bit dangerous confusing for users to potentially specify font in two places.

2. I want to push for a better package story than "require the user add dependencies to the global window object". Can we add a dependency to just the addon somehow? I want a user's build system to simply import the addon and have the right thing happen. Usability of addons go down if you require the user to manually pollute the global object just to use them. Complicating this, FontFaceObserver is not fully ES6 yet and does require a `require` to work.

vincentwoo:
@Tyriar get a chance to see this?

Tyriar:
> I want to push for a better package story than "require the user add dependencies to the global window object". Can we add a dependency to just the addon somehow? I want a user's build system to simply import the addon and have the right thing happen. Usability of addons go down if you require the user to manually pollute the global object just to use them. Complicating this, FontFaceObserver is not fully ES6 yet and does require a require to work.

I don't want consumers who don't need zmodem or fontfaceobserver to need to download them when installing the package and also to remove them manually when they don't need them.

How about you pass it into the function then? That seems much better than relying on a global:

```ts
loadWebfontAndOpen = function (element: HTMLElement, ffo: FontFaceObserver)
```

```ts
webfont.loadWebfontAndOpen(element, require('fontfaceobserver'));
```

Tyriar:
That's the hope eventually (I just commented about this https://github.com/xtermjs/xterm.js/issues/1128#issuecomment-380575037), in which can it could be moved to a regular dependency.

vincentwoo:
I can kill this PR and just release this on my own if that's what you
prefer.

Tyriar:
Discoverability is then the main problem with third party addons at the moment as we don't really have a good story for that. It's up to you whether you want to do that or merge it in. Eventually this would be best to live as a third party addon not maintained by the team though.

mofux:
Couldn't we simply add a section to our README that lists 3rd Party Addons, similar to the Real-world uses?

Tyriar:
@mofux we could, I don't think it's a long term solution though. The readme is very long already.

vincentwoo:
I've split out this PR as its own package: https://www.npmjs.com/package/xterm-webfont. I'll leave it up to you on whether to link to it somewhere or not. I think this is a good middle ground and long-term place to be wrt addons.

Tyriar:
https://github.com/xtermjs/xterm.js/pull/1390

Issue #1341, opened 2018-03-20
Terminals are leaking memory


hansonw:
By using Chrome's heap profiler, it appears that all `Terminal` objects are being permanently retained at the moment (even after calling `destroy()`) via two code paths:

1) The `charAtlasCache`, which records the terminals that own each character atlas but doesn't delete them when the terminals go away. I think this one can be resolved by making `ownedBy` a `WeakSet` and then holding a counter separately (though this isn't ideal).

2) A permanent listener to the `'resize'` event on `window`: https://github.com/xtermjs/xterm.js/blob/master/src/Terminal.ts#L690. I guess we'd have to keep the callback somewhere and detach it on `destroy()`.

I can probably find time to submit PRs to fix these but getting open source approval at Facebook may be a bit messy... it may be more expedient for a maintainer to patch these. Thanks! :)

mofux:
@hansonw Thanks for figuring this out and posting it here. I found another case where listeners are never cleaned up. The ScreenDprMonitor provides a `clearListener` method, but I cannot see it ever being called:
https://github.com/xtermjs/xterm.js/blob/be18507df64d6f79113d216572921ddb912b8196/src/utils/ScreenDprMonitor.ts#L43

Tyriar:
I put out https://github.com/xtermjs/xterm.js/pull/1372 which I think fix those 2 cases. It's tough to test with the demo whether everything is cleared though due to a bunch of references in the demo code. I think we should refactor the demo as part of this to allow easier disposal of a terminal.

PR #1372, opened 2018-04-04
Remove terminal references on dispose

### LINKED ISSUE #1370: dispose/destroy are used in various places to mean the same thing
I suggest we standardize on "dispose" for such methods (and deprecate `Terminal.destroy` https://github.com/xtermjs/xterm.js/issues/1371).

Tyriar:
Part of #1341
Fixes #1370


src/renderer/atlas/CharAtlas.ts
```diff
@@ -70,3 +72,22 @@ export function acquireCharAtlas(terminal: ITerminal, colors: IColorSet, scaledC
...
+ * Removes a terminal reference from the cache, allowing its memory to be freed.
+ * @param terminal The terminal to remove.
+ */
+export function removeTerminalFromCache(terminal: ITerminal): void {
+  for (let i = 0; i < charAtlasCache.length; i++) {
+    const index = charAtlasCache[i].ownedBy.indexOf(terminal);
+    if (index !== -1) {
+      if (charAtlasCache[i].ownedBy.length === 1) {
+        // Remove the cache entry if it's the only terminal
+        charAtlasCache.splice(i, 1);
```
bgw: I think we can `break` here, because a terminal should only be able to own one config. If I'm mistaken about that, and a terminal can own multiple configs, then we should do `i--` because we've shifted the subsequent items in the array.

src/Terminal.ts
```diff
@@ -250,7 +254,28 @@ export class Terminal extends EventEmitter implements ITerminal, IInputHandlingT
     this._setup();
   }
 
+  public dispose(): void {
+    super.dispose();
+    this._disposables.forEach(d => d.dispose());
+    this._disposables.length = 0;
```
bgw: I'm not sure why this is required. Or why we need to clear out `handler` and `write` either (though you just moved it in this commit).

src/renderer/atlas/CharAtlas.ts
Tyriar: Good catch :+1:

src/Terminal.ts
Tyriar: Setting or handler and write has been there from the beginning, I assume it's to prevent additional data being written after disposal in either direction. If we removed we would need to keep an `_isDisposed` variable and check it on several methods. Would probably be a good change but would rather not do as part of this PR.

PR #1373, opened 2018-04-04
Do not trim wrapped lines when copying

### LINKED ISSUE #1286: Copying a multi line selection will trim the right whitespace on wrapped lines incorrectly
image

See https://github.com/Microsoft/vscode/issues/43696

Tyriar:
Fixes #1286


Tyriar:
This actually leads to some weird behavior due to the reflow workaround where we hang onto data to the right of the viewport when it's sized down. Let's wait on reflow https://github.com/xtermjs/xterm.js/issues/622 before merging this in.

PR #1376, opened 2018-04-04
Move zmodem demo to demo/

### LINKED ISSUE #1324: We're shipping the zmodem demo and it also breaks debugging
screen shot 2018-03-11 at 12 03 04 pm

It should probably live in a sub-directory of demo/ instead.

Tyriar:
Fixes #1324 

---

@FGasper I think the demo was broken before as a result of the applyAddon refactor https://github.com/xtermjs/xterm.js/pull/1130. I think it should work now but I didn't actually test it.


FGasper:
@Tyriar 
https://github.com/FGasper/xterm.js/commit/72e226923f60f001518ac4ab68d6367dd937d3b6
^^ That makes it work.

Tyriar:
Pulled it in, ready for review now.

PR #1384, opened 2018-04-11
Guard usages of DOM attached vars in setOption

### LINKED ISSUE #1383: Crash when you change options before calling open
Attempting to setOption for some options triggers a `renderer.clear`, but if the terminal has not opened yet, this will crash.

[CodePen repro](https://codepen.io/akanet/pen/wmOMoe)

relevant code: https://github.com/xtermjs/xterm.js/blob/master/src/Terminal.ts#L452

Tyriar:
Fixes #1383


Tyriar:
@jupe see the milestone

bgw:
LGTM

PR #1388, opened 2018-04-13
Improved test related npm scripts

Tyriar:
I added a section to the README explaining the main npm scripts. This change was mainly motivated because I wanted to improve test debugging as it recently broke without our knowledge. `npm run test-debug` now follows the same code path as `npm test` so breakages shouldn't occur, and it just works if you have an auto attaching debugger.

I added `npm run test-suite <file>` to make it easier to discover running single test suites, and coverage is now done separately to `npm test` so it runs much faster now


Tyriar:
Turns out the whole .vscode folder is no longer needed at all as search and the quick picker now use .gitignore by default 🎉

Tyriar:
This PR would need VSTS to update the test call to use test-coverage

Tyriar:
This is ready, will just need a change in the VSTS build script after it's merged to master

PR #1391, opened 2018-04-17
Separate bright and bold

LinusU:
Previously all bright colors were always rendered bold, and all bold text was always rendered using the bright colors. This patch separates that so that bold and bright can be controlled individually.

This is probably going to affect people who relied on this previously, so I'm working on an option that re-enables the old behavior. I will push that as a separate commit to this same branch.

Feedback appreciated! 🙌 

Fixes zeit/hyper#2836

### DIFF
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -55,6 +55,11 @@ declare module 'xterm' {
     disableStdin?: boolean;
 
     /**
+     * Whether to draw bold text in bright colors. The default is true.
+     */
+    drawBoldTextInBrightColors?: boolean;
+
+    /**
      * Whether to enable the rendering of bold text.
      *
      * @deprecated Use fontWeight and fontWeightBold instead.
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -103,6 +103,7 @@ const DEFAULT_OPTIONS: ITerminalOptions = {
   cursorStyle: 'block',
   bellSound: DEFAULT_BELL_SOUND,
   bellStyle: 'none',
+  drawBoldTextInBrightColors: true,
   enableBold: true,
   fontFamily: 'courier-new, courier, monospace',
   fontSize: 15,
--- a/src/renderer/BaseRenderLayer.ts
+++ b/src/renderer/BaseRenderLayer.ts
@@ -248,7 +248,7 @@ export abstract class BaseRenderLayer implements IRenderLayer {
     const isBasicColor = fg < 16;
     const isDefaultColor = fg >= 256;
     const isDefaultBackground = bg >= 256;
-    const drawInBrightColor = (terminal.options.drawBoldTextInBrightColors !== false && bold && fg < 8);
+    const drawInBrightColor = (terminal.options.drawBoldTextInBrightColors && bold && fg < 8);
     if (this._charAtlas && isAscii && (isBasicColor || isDefaultColor) && isDefaultBackground && !italic) {
       let colorIndex: number;
       if (isDefaultColor) {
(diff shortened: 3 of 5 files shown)


LinusU:
Thanks for the feedback.

One question, if we have a `drawBoldAsBright` option, should that also draw bright text bold? That seems confusing but that's how it worked before, since bright and bold were essentially the same thing.

We could also make it two options: `drawBoldAsBright` and `drawBrightAsBold`, but then `as` might be confusing as it's rather both. Maybe `drawBoldTextBright` and `drawBrightTextBold`? Both defaulting to true since that was how it was before.

Personally leaning most towards `drawBoldTextBright` and `drawBrightTextBold`...

mofux:
I don't think that drawing bright text as bold would make sense. Most terminals offer two options:
- enable / disable drawing bold text as bright (the text may still be bold)
- enable / disable drawing bold text as bold (the text may still be bright)

So we might end up with something like `drawBoldTextBright` and `drawBoldTextBold` that can be combined to get the desired flavour 😁

LinusU:
> I don't think that drawing bright text as bold would make sense.

I agree, but as far as I know, that is the current behaviour of xterm.js. So if we want to keep backwards compatibility, I think we need an option for that 🤔

mofux:
Hmmm I haven't tested it, but I don't think that the current default behaviour is to draw bright text as bold (I've also not found any code that would point towards that). Imo we should not support an option that draws bright text as bold as it is confusing -- just as you noted. @xtermjs/core Any thoughts?

mofux:
>Implement an additional option drawBoldTextInBrightColors that will draw bold characters in bright colors and will default to false

@parisk The default should be `true`, as drawing bold text in bright colors is the default behaviour at the moment.

parisk:
Nah, you are right 😅. Fixed it.

LinusU:
Hahahhaaha 🤦‍♂️ 🤦‍♂️ 🤦‍♂️ 

```js
const isAscii = code < 256;
// A color is basic if it is one of the standard normal or bold weight
// colors of the characters held in the char atlas. Note that this excludes
// the normal weight _light_ color characters.
const isBasicColor = (colorIndex > 1 && fg < 16) && (fg < 8 || bold);
const isDefaultColor = fg >= 256;
const isDefaultBackground = bg >= 256;
if (this._charAtlas && isAscii && (isBasicColor || isDefaultColor) && isDefaultBackground) {
    // draw from atlas
} else {
    // draw using ctx
}
```

LinusU:
Rebased on master, and added two new commits.

The first one fixed so that I'm actually utilizing the char atlas for all 4-bit ANSI colors, bold and non-bold. The second one adds the `drawBoldTextInBrightColors` option which defaults to true.

2f1ec3b, and fd185a5 with `drawBoldTextInBrightColors` set to `false`:

screen shot 2018-04-20 at 10 31 58

fd185a5, which is same as current master:

screen shot 2018-04-20 at 10 33 44

From my side, this is ready for merging now 🚀 let me know if anything else should be fixed 🙌

typings/xterm.d.ts
```diff
@@ -54,6 +54,11 @@ declare module 'xterm' {
      */
     disableStdin?: boolean;
 
+    /**
+     * Wheter to draw bold text in bright colors. The default is true.
```
saul: s/Wheter/Whether/

LinusU: Fixed 👌

LinusU:
Rebased to avoid conflicts with #1422 

ping @Tyriar @bgw @parisk, would love to get this merged ❤️

bgw:
I added this small commit to fix a minor bug: https://github.com/xtermjs/xterm.js/pull/1391/commits/9893338020cc66efb8180f8f4df6852a0880bfae

Otherwise, LGTM. I manually tested it, and it seemed to work well for me. I'll merge once Travis finishes. Thanks for your contribution, sorry it took me so long to merge it.

bgw:
Travis is failing, but the issue is unrelated. It looks like a recent change to one of our dependencies broke some types. I was able to reproduce them locally in master by wiping and re-installing my node_modules.

PR #1392, opened 2018-04-18
Add node-canvas help increase the testable surface

bgw:
This is at least a first step towards fixing #1247. This adds some complexity to the initial setup, since canvas needs to be built from source.

This won't be true once canvas 2.x is stabilized, because that version downloads (Automattic/node-canvas#992). JSDom doesn't support 2.x, and doesn't plan to (jsdom/jsdom#1964) until a stable release is cut.

We could use [canvas-prebuilt](https://github.com/node-gfx/node-canvas-prebuilt), which JSDom does appear to support. However, I'm not sure if that would cause pain for people with 32-bit operating systems (see the compatibility table on that page).

I'm not sure if this travis configuration will work; I'll iterate on it in the PR if it fails. I removed a bunch of hacks that were added in #690, since they don't look necessary anymore (the sleep module is gone).

(I previously had a copy of this PR up over at bgw/xterm.js#1)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,20 @@ os:
 node_js:
   - 6
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get -qq update ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get -qq install g++-4.8 ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then export CXX=g++-4.8 ; fi
   - npm install -g npm@5.1.0
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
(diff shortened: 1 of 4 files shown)


bgw:
I've realized this will probably break the docker and procfile setups, so I'm going to try to test/fix those before landing.

parisk:
@bgw do you believe we can enforce this on Travis CI, while making it opt-in for the developers?

In the test suite we can try to import node-canvas and in case it fails, just skip the test.

Also, I think it's pretty straightforward to include the node-canvas dependencies in the Docker image. I can help with that, if you would like.

.travis.yml
```diff
@@ -4,12 +4,20 @@ os:
 node_js:
   - 6
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test ; fi
```
Tyriar: :+1:

package.json
```diff
@@ -45,6 +45,7 @@
     "@types/node": "^6.0.41",
     "@types/text-encoding": "0.0.32",
     "browserify": "^13.3.0",
+    "canvas": "^1.6.10",
```
Tyriar: Does JSDom just pick this module up if it's available?

PR #1393, opened 2018-04-18
Separate foreground & background rendering passes

bgw:
I'm primarily interested in doing this because it'll allow me to optimize the background rendering in later diffs, but this may also make it possible to fix some minor rendering issues.

Right now, if we draw a single-width character that overflows its cell's bounds, and the character to the right of it has a background, that background may cover up the first character's foreground. By drawing the background in a separate pass, we can avoid those cases.

There's still some issues with how dirty regions are computed that makes rendering stuff like that flaky, but this at least gets us closer to "correct" rendering.

Other terminal emulators (e.g. alacritty) render the foreground and background in separate passes:
https://github.com/jwilm/alacritty/blob/1b7ffea/src/renderer/mod.rs#L766

### DIFF
--- a/src/renderer/TextRenderLayer.ts
+++ b/src/renderer/TextRenderLayer.ts
@@ -48,21 +48,24 @@ export class TextRenderLayer extends BaseRenderLayer {
     this.clearAll();
   }
 
-  public onGridChanged(terminal: ITerminal, startRow: number, endRow: number): void {
-    // Resize has not been called yet
-    if (this._state.cache.length === 0) {
-      return;
-    }
-
+  private _forEachCell(
+    terminal: ITerminal,
+    startRow: number,
+    endRow: number,
+    callback: (
+      code: number,
+      char: string,
+      width: number,
+      x: number,
+      y: number,
+      fg: number,
+      bg: number,
+      flags: number
+    ) => void
+  ): void {
     for (let y = startRow; y <= endRow; y++) {
       const row = y + terminal.buffer.ydisp;
       const line = terminal.buffer.lines.get(row);
-
-      this.clearCells(0, y, terminal.cols, 1);
-      // for (let x = 0; x < terminal.cols; x++) {
-      //   this._state.cache[x][y] = null;
-      // }
-
       for (let x = 0; x < terminal.cols; x++) {
         const charData = line[x];
         const code: number = <number>charData[CHAR_DATA_CODE_INDEX];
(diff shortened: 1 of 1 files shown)


src/renderer/TextRenderLayer.ts
```diff
@@ -150,45 +116,68 @@ export class TextRenderLayer extends BaseRenderLayer {
...
+    });
+  }
+
+  public onGridChanged(terminal: ITerminal, startRow: number, endRow: number): void {
+    // Resize has not been called yet
+    if (this._state.cache.length === 0) {
+      return;
     }
+
+    this.clearCells(0, startRow, terminal.cols, endRow - startRow + 1); // endRow is inclusive
```
Tyriar: We could remove the comment if the params were renamed to firstRow and lastRow?

src/renderer/TextRenderLayer.ts
```diff
@@ -150,45 +116,68 @@ export class TextRenderLayer extends BaseRenderLayer {
...
-          }
-          this.fillBottomLineAtCells(x, y);
-        }
+        callback(code, char, width, x, y, fg, bg, flags);
+      }
+    }
+  }
 
-        this.drawChar(terminal, char, code, width, x, y, fg, bg, !!(flags & FLAGS.BOLD), !!(flags & FLAGS.DIM));
+  private _drawBackground(terminal: ITerminal, startRow: number, endRow: number): void {
```
Tyriar: nit: This would be better structured imo as `_drawBackgroundCell`/`_drawForegroundCell` and called like this:

```ts
this._forEachCell(terminal, startRow, endRow, this._drawBackgroundCell);
this._forEachCell(terminal, startRow, endRow, this._drawForegroundCell);
```

Tyriar:
This is related to this but looks like it was caused by another commit. Do you know why this is what I see on the master branch in `vim README.md` when the cursor is on the `*` next to "Note":

image

And this is what I see in vscode (close to master)?

image

bgw:
> Do you know why this is what I see on the master branch in vim README.md when the cursor is on the * next to "Note"

Are you saying that the highlight only appears when you move the cursor over the `*`?

A lot of vim theme/syntax combinations will try to make text like this italic, and if the terminal emulator doesn't support italics ([sometimes this requires messing with terminfo to get it to work](https://sookocheff.com/post/vim/italics/)), it'll often fall back to something else, like inverting the output. I think there's a lot of reasons this could be happening for you, but I'm not sure what it is exactly. I expect that it's related to differences between the demo environment and vscode's environment, rather than a regression in xtermjs.

- How does this look in other terminal emulators?
- What's your `TERM` environment variable in the demo and in vscode?
- Can you try to [figure out what syntax highlighting group this is](http://vim.wikia.com/wiki/Identify_the_syntax_highlighting_group_used_at_the_cursor), and see what your theme does for that group?

Tyriar:
@bgw the major issue is that the cursor is not visible at all, this may be because the background color is the same color as the cursor. I'll check your other questions when I get back to my Linux box 😄

src/renderer/TextRenderLayer.ts
bgw: My plan is to keep track of the previously drawn background color as a variable scoped to the top-level of `_drawBackground`, and then minimize `fillRect` calls by only using `fillRect` once the background color changes or we've stopped iterating over cells.

While your API looks a little nicer, I don't know how I'd accomplish that optimization with it.

Tyriar:
@bgw did you consider splitting up the bg and fg into separate render layers? I can't remember why I merged them to begin with, but it may have been because I was drawing the background with the character to the cache initially, before this existed:

https://github.com/xtermjs/xterm.js/blob/a3ce881897818d1c61476620e7d037717ca9a68f/src/shared/atlas/CharAtlasGenerator.ts#L111-L122

That would make drawing minimal diffs on the background layer super easy as we can keep a `_state` cache like we used to.

bgw:
> did you consider splitting up the bg and fg into separate render layers?

This would be fine for characters drawn from the cache, but the foreground layer would need to be transparent for this to work, and uncached characters drawn directly to the foreground layer wouldn't get SPAA.

We could render uncached characters to an intermediate buffer, and use clearColor on it, but that'll make the uncached codepath that much slower, so I don't think it's good idea.

Tyriar:
> uncached characters drawn directly to the foreground layer wouldn't get SPAA.

@bgw ah yes, that was probably the reason i did it :+1:

Issue #614, opened 2017-03-17
Support 1337 escape sequence to show inline images


cristynkells:
## Details
- Browser and browser version: all
- OS version: all
- xterm.js version: 2.4.0 

### Feature request
Is there a way to view images from the command line?

by either clicking on the image (much like the linkify addon) or via a command:

$ cat vacation-beach-pic.png

Tyriar:
Clicking on the link to open the image is possible via the an API that's experimental at the moment. It's experimental because it's likely to change but here is the current syntax:

```js
// Add a link for a particular regex and handle when it's clicked.
xterm.registerLinkMatcher(/\w\.png/, function (mouseEvent, imageFile) => {
  // imageFile is the text of the link
});
```

cristynkells:
Hi Daniel, thank you for your speedy reply! Was wondering if there a way to display html on the fly through a set of control characters? Similar to how DomTerm uses “\e]72;<html>\a” sequences ? 

 

https://lwn.net/Articles/670078/

Tyriar:
@cristynkells there isn't right now, standard xterm doesn't support anything like that so if we did want to support it it would certainly need to be in an addon.

DonJayamanne:
@Tyriar 
>it would certainly need to be in an addon

Is this supported today? Right now I don't see how this can be done even with an add on. Basically I'm trying to investigate what it would take to build something with inline plots (something we already have in other IDE, PTVS, RTVS, Spyder, and the like e.g. https://ipython.org/ipython-doc/3/interactive/qtconsole.html).

Tyriar:
@DonJayamanne since this issue I've become aware of a standard some terminals have implemented for displaying terminals. Because of this we could add it to the core library, it's probably a pretty hefty feature though in terms of making sure everything is correct, particularly relating to getting the scroll bar right as right now there's no such thing as variable height rows in the viewport.

DonJayamanne:
> implemented for displaying terminals

Yes, I found iTerm does this.

Thanks for getting back.

ebraminio:
Hi there.I opened [a related issue on VSCode](https://github.com/Microsoft/vscode/issues/44436) and redirected here I wonder supporting iTerm2 image feature a possibility it can be used as easy as this,

    printf "\n\033]1337;File=;inline=1:`cat a.png | base64`\a\n"`

[...]

Issue #576, opened 2017-03-02
Implement Shell Integration


ctwise:
iTerm2 has implemented (and documented) proprietary escape code for shell integration. This integration is used to provide command-line markers, jump between command-line entries in the terminal history, track successful commands, provide a terminal-level command history, track the current host and current directory. These escape codes originated with the (now defunct) FinalTerm project.

Please implement these escape codes - https://iterm2.com/documentation-escape-codes.html

Shell Integration/FinalTerm
iTerm2's Shell Integration feature is made possible by proprietary escape sequences pioneered by the FinalTerm emulator. FinalTerm is defunct, but the escape sequences are documented here.

Definitions
OSC stands for Operating System Command. In practice it refers to this sequence of two ASCII characters: 27, 93 (esc ]).
ST stands for String Terminator. It terminates an OSC sequence and consists either of two ASCII characters 27, 92 (esc \) or ASCII 7 (bel).
OSC sequences always begin with OSC, are followed by a sequence of characters, and are terminated with ST.

Most OSC codes begin with a number (one or more decimal digits), which we'll call the "command" in this document. If the command takes parameters it will be followed by a semicolon and the structure of the rest of the body of the OSC sequence is dependent on the command. Well-behaved terminal emulators ignore OSC codes with unrecognized commands.

Concepts
The goal of the FinalTerm escape sequences is to mark up a shell's output with semantic information about where the prompt begins, where the user-entered command begins, and where the command's output begins and ends.

[PROMPT]prompt% [COMMAND_START] ls -l
[COMMAND_EXECUTED]
-rw-r--r-- 1 user group 127 May 1 2016 filename
[COMMAND_FINISHED]

Escape Sequences
FinalTerm originally defined various escape sequences in its original spec that are not supported by iTerm2 and are not described in this document. The best remaining references to these codes are in iTerm2's source code.

FTCS_PROMPT

OSC 1 3 3 ; A ST

Sent just before start of shell prompt.

FTCS_COMMAND_START

OSC 1 3 3 ; B ST

Sent just after end of shell prompt, before the user-entered command.

FTCS_COMMAND_EXECUTED

OSC 1 3 3 ; C ST

[...]

PR #1399, opened 2018-04-21
new parser

jerch:
WIP - don't merge...

- basic transition to ansi complicant parser done
- speed gain ~30%

TODO:
- tests for glue class `ParserTerminal`
- refactoring with API for hooking custom escape codes and methods
- maybe merge `ParserTerminal` with `InputHandler`
- better decoupling of the higher level stuff


jerch:
This PR is up for a first technical and conceptual review:

About `EscapeSequenceParser`:
- implements the reference from https://vt100.net/emu/dec_ansi_parser as FSM and should therefore support all legacy escape sequences
- newer sequences like CSI with parameter element colon notation are not supported atm (were not part of the old DEC machines), can be added with some effort
- custom transition maps via parser argument
- custom transitions on the fly possible via `TransitionTable.add`
- FSM actions are hardcoded as switch statement in the parse method for performance reasons
- hot parse loop avoids memory allocations as much as possible (no string copies)
- supports error propagation (back propagation not implemented yet)
- fully tested

Interface to xterm.js:
- realized by class `ParserTerminal` implementing the action callbacks needed by `EscapeSequenceParser`
- no magic here - was straight forward to integrate with existing `InputHandler` methods
- good place to allow custom sequence handler hooks
- DCS not yet implemented
- no tests yet

Results so far:
- passes all tests, no awkward behavior during live tests
- 10% speed gain by switching to the new parser
- performance tests revealed `InputHandler.addChar` as main bottleneck (localizing `_terminal.buffer` gave 20% performance boost, seems the getters to the buffer implementation are very expensive - Should this be tracked in a new issue?)

There is still alot refactoring needed, several questions popped during coding (also see code comments):
- The Parser uses atm the global transition table by default. Should this be copied over to new instances instead to allow hot custom state patches without altering other terminals?
- Should all escape sequence types be customizable or only OSC and DCS?
- Should `ParserTerminal` be merged with `InputHandler`? If so this could be the base for a screen/input decoupled terminal class since it could handle all the escape sequence related stuff.
- There are still many references to the `_terminal` object. Are there any plans yet to decouple this further?
- The execute action does not define any method for XON or XOFF (copied over). Is this done by a different layer? Or does xterm.js currently not support XON/XOFF from slave side?
- The execute action falls back to `addChar` atm (again copied over). I think this needs to be revised since it might lead to weird output sometimes of characters that should not be printed at all.
- in ESC action: Does xterm.js support single G2/G3 shifts?
- in ESC action: I did not understand the code for `ESC /` - isnt this code [here](https://github.com/xtermjs/xterm.js/blob/master/src/Parser.ts#L382) unreachable?
- How to deal with custom escape sequence handlers? Registering/deregistering could be done in a typical event like fashion or by a more restrictive coupling to some API methods. Not sure what is better here, if the class would just emit escape sequence events people might get greedy with this and that, also performance will suffer. It would be more versatile though.

This very much at once, maybe someone finds some time to address some of the questions and basic layout ideas.

src/EscapeSequenceParser.ts
```diff
@@ -1,17 +1,62 @@
...
-    inst_p?: (s: string, start: number, end: number) => void;
-    inst_o?: (s: string) => void;
-    inst_x?: (flag: string) => void;
-    inst_c?: (collected: string, params: number[], flag: string) => void;
-    inst_e?: (collected: string, flag: string) => void;
-    inst_H?: (collected: string, params: number[], flag: string) => void;
-    inst_P?: (dcs: string) => void;
-    inst_U?: () => void;
-    inst_E?: () => void; // TODO: real signature
+    actionPrint?: (data: string, start: number, end: number) => void;
```
Tyriar: You should make this mandatory (remove `?`), otherwise you would need to check for its existence before calling every time

src/EscapeSequenceParser.ts
```diff
@@ -21,171 +66,195 @@ export function r(a: number, b: number): number[] {
...
+// default transition table points to global object
+// Q: Copy table to allow custom sequences w'o changing global object?
+export class EscapeSequenceParser {
     public initialState: number;
     public currentState: number;
     public transitions: TransitionTable;
     public osc: string;
     public params: number[];
     public collected: string;
     public term: any;
```
Tyriar: Is any just temporary?

src/EscapeSequenceParser.ts
```diff
@@ -21,171 +66,195 @@ export function r(a: number, b: number): number[] {
...
-    constructor(terminal: IParserTerminal) {
-        this.initialState = 0;
-        this.currentState = this.initialState | 0;
-        this.transitions = new TransitionTable(4095);
-        this.transitions.table.set(TRANSITION_TABLE.table);
+    constructor(
+        terminal?: IParserTerminal | any,
```
Tyriar: This could be:

```ts
constructor(
  public terminal: IParserTerminal
  ...
)
```

src/EscapeSequenceParser.ts
```diff
@@ -204,79 +273,81 @@ export class AnsiParser {
...
                     break;
-                case 1: // error
-                    // handle unicode chars in write buffers w'o state change
+                case ACTION.error:
+                    // chars higher than 0x9f are handled by this action to
+                    // keep the lookup table small
                     if (code > 0x9f) {
                         switch (currentState) {
-                            case 0: // GROUND -> add char to print string
+                            case STATE.GROUND:          // add char to print string
```
Tyriar: Avoid using whitespace for alignment, align with the `:` or put on a new line.

src/EscapeSequenceParser.ts
```diff
@@ -204,79 +273,81 @@ export class AnsiParser {
...
-                                    character: String.fromCharCode(code), // wrong character
-                                    state: currentState,   // in state
-                                    print: printed,         // print buffer
-                                    dcs: dcs,               // dcs buffer
-                                    osc: osc,               // osc buffer
+                                    pos: i,                 // position in string
+                                    code: code,             // actual character code
+                                    state: currentState,    // current state
+                                    print: printed,         // print buffer start index
+                                    dcs: dcs,               // dcs buffer start index
```
Tyriar: If key and value are the same you can just use one:

```ts
{
  pos: i,
  code,
  state: currentState,
  print,
  dcs,
  ...
}
```

src/EscapeSequenceParser.ts
```diff
@@ -1,17 +1,62 @@
...
+    DCS_ENTRY,
+    DCS_PARAM,
+    DCS_IGNORE,
+    DCS_INTERMEDIATE,
+    DCS_PASSTHROUGH
+}
+
+// FSM actions
+export const enum ACTION {
+    ignore = 0,
```
Tyriar: Let's use this naming convention for enums:

https://github.com/xtermjs/xterm.js/blob/716a8d592676ada264d9b9b7ebec2f1f3277a4c6/src/Parser.ts#L153-L162

src/EscapeSequenceParser.ts
```diff
@@ -21,171 +66,195 @@ export function r(a: number, b: number): number[] {
...
     public collected: string;
     public term: any;
-    constructor(terminal: IParserTerminal) {
-        this.initialState = 0;
-        this.currentState = this.initialState | 0;
-        this.transitions = new TransitionTable(4095);
-        this.transitions.table.set(TRANSITION_TABLE.table);
+    constructor(
+        terminal?: IParserTerminal | any,
+        transitions: TransitionTable = VT500_TRANSITION_TABLE)
```
Tyriar: Again you can add public before `transitions` and then remove the member above.

src/EscapeSequenceParser.ts
```diff
@@ -21,171 +66,195 @@ export function r(a: number, b: number): number[] {
...
+    table.addMany([0x1b, 0x9c], STATE.DCS_PASSTHROUGH, ACTION.dcs_unhook, STATE.GROUND);
+
+    return table;
 })();
 
-export class AnsiParser {
+
+// default transition table points to global object
+// Q: Copy table to allow custom sequences w'o changing global object?
+export class EscapeSequenceParser {
```
Tyriar: Is this going to hide behind an interface and use the readonly trick for properties that shouldn't be edited outside?

src/EscapeSequenceParser.ts
```diff
@@ -0,0 +1,785 @@
...
+            case C0.SI: return this._inputHandler.shiftIn();
+            default:
+                this._inputHandler.addChar(flag, flag.charCodeAt(0));
+        }
+        this._terminal.error('Unknown EXEC flag: %s.', flag);
+    }
+
+    actionCSI(collected: string, params: number[], flag: string): void {
+        this._terminal.prefix = collected;
+        switch (flag) {
```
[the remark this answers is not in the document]
jerch: Yep, gonna try to find out, how to jump with the lowest costs. I think emitting the charCode from `EscapeSequenceParser` and placing the function pointers directly into an object will be the fastest (at least in chrome numerical properties avoid the key hashing). As soon as string type properties are involved, a switch is faster until ~20 entries or so.
About the interface - since I am not that good with typescript declarations any more typescripty suggestion is welcome.

src/EscapeSequenceParser.ts
jerch: Ah nice shortcut.

src/EscapeSequenceParser.ts
jerch: Sure thing. Oh and the enums must be `const`, otherwise Typescript will place JS identifiers into the parse method with worse performance.

Tyriar: Oh wow, that's a huge difference! We must constify all the enums!

```ts
enum a {
	b,
	c,
    d
}

const enum A2 {
	B2,
	B3,
    B4
}

console.log(a.b);
console.log(A2.B2);
```

Compiles to:

```ts
var a;
(function (a) {
    a[a["b"] = 0] = "b";
    a[a["c"] = 1] = "c";
    a[a["d"] = 2] = "d";
})(a || (a = {}));
console.log(a.b);
console.log(0 /* B2 */);
```

bgw: It's probably not enough of an argument against it, but if we ever wanted to use babel to compile the typescript source directly, it's unable to handle `const enum`s, because they require type information.

src/EscapeSequenceParser.ts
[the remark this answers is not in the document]
jerch: Yes thats basically the decorator pattern I had in mind for the DCS part, should work for the others as well. :+1:

src/EscapeSequenceParser.ts
jerch: I am not sure either if you want enums to be const everywhere. I need this feature here to get the plain number at JS level avoiding costly deref at runtime, still I wish I had access to the enum definition (e.g. have the states named for hot patches) and already thought about workarounds (with no handy solution yet lol).

src/EscapeSequenceParser.ts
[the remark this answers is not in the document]
jerch: Hmm do we need support for multiple handlers for a single action? This would allow stacking functionality for addons on top of the the basic functionality. Not sure if maintaining an array of function pointers will hurt performance wise.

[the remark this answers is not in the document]
mofux: Maybe starting with support for a single handler is enough for the moment. With the suggested API it wouldn't be complicated to add support for multiple handlers later on if a use-case comes up for it. I'd think that emitting more higher-level events from the input handler (like "linefeed" or "cursorMove") is more appealing for most use-cases.

src/EscapeSequenceParser.ts
jerch: Kinda obsolete with the new jump tables.

src/EscapeSequenceParser.ts
jerch: Not sure on this one yet: the transition table takes 5kB as Uint8Array and much more as fallback Array implementation (> 40kB, not done yet). I tend to make the transition table private and give the parser a copy on write method for custom states to save memory for multiple terminals.
Not sure either whether to hide `EscapeSequenceParser` and give `ParserTerminal/InputHandler` the main control API.

jerch:
Made some progress towards a nicer integration. The current version does all callback registering directly in `InputHandler`. I decided to remove `ParserTerminal` since it mostly passed `InputHandler` functionality to the parser with almost no own business logic.
Still I dont quite understand the separation level of `InputHandler` vs. `Terminal` - some escape sequence actions directly access terminal internals while others do this via an InputHandler method. Maybe the direct accesses could also be managed by InputHandler?

Tyriar:
> Hmm do we need support for multiple handlers for a single action? This would allow stacking functionality for addons on top of the the basic functionality. Not sure if maintaining an array of function pointers will hurt performance wise.

@jerch It would be a map underneath not an array, so O(1) assign and lookup. This is what I had in mind:

```ts
private registerCsiHandler(char: string, callback: SomeType): void {
  this.csiHandlers[char] = callback;
}
```

Then action them more like `master` does now:

https://github.com/xtermjs/xterm.js/blob/381d3e14441209f605d047c1fee6bad8425ac185/src/Parser.ts#L484-L497

Tyriar:
> It's probably not enough of an argument against it, but if we ever wanted to use babel to compile the typescript source directly, it's unable to handle const enums, because they require type information.

Easy to remove const if we need to in the future.

> I am not sure either if you want enums to be const everywhere.

The tslint rule added in https://github.com/xtermjs/xterm.js/pull/1407 will force const on all non-exported enums, which don't need type info. Making exported ones const should also be OK, as long as they're not in `xterm.d.ts`.

src/InputHandler.ts
```diff
@@ -19,24 +20,254 @@ import { wcwidth } from './CharWidth';
...
+    // 113 - Reset mouse foreground color.
+    // 114 - Reset mouse background color.
+    // 115 - Reset Tektronix foreground color.
+    // 116 - Reset Tektronix background color.
+    // 117 - Reset highlight color.
+    // 118 - Reset Tektronix cursor color.
+    // 119 - Reset highlight foreground color.
+
+    // ESC handlers
+    this._parser.setEscHandler('', '7', this.saveCursor.bind(this));
```
Tyriar: I've found that .bind() loses all function type info so I always prefer `() => this.saveCursor()` now.

src/EscapeSequenceParser.ts
```diff
@@ -0,0 +1,544 @@
...
+  }
+  clearCsiHandler(flag: string): void {
+    if (this._csiHandlers[flag.charCodeAt(0)]) delete this._csiHandlers[flag.charCodeAt(0)];
+  }
+  setCsiHandlerFallback(callback: (...params: any[]) => void): void {
+    this._csiHandlerFb = callback;
+  }
+
+  setEscHandler(collect: string, flag: string, callback: IEscHandler): void {
+    this._escHandlers[collect + flag] = callback;
```
Tyriar: > The flag method used for CSI is much faster

If this is slow maybe it can be called with a single arg `collectAndFlag`? Do you know what's slowing it down?

src/EscapeSequenceParser.ts
```diff
@@ -0,0 +1,544 @@
...
+        case ParserAction.PARAM:
+          if (code === 0x3b) params.push(0);
+          else params[params.length - 1] = params[params.length - 1] * 10 + code - 48;
+          break;
+        case ParserAction.COLLECT:
+          collect += String.fromCharCode(code);
+          break;
+        case ParserAction.ESC_DISPATCH:
+          ident = collect + String.fromCharCode(code);
+          if (this._escHandlers[ident]) this._escHandlers[ident](collect, code);
```
Tyriar: > double lookup: if (<lookup>) <lookup>() though I think this gets optimized by the engine and is no biggie

Can be pretty easily fixed with an extra variable:

```ts
const handler = this._escHandlers[ident];
if (handler) handler(collect, code);
```

I don't expect the cost to be very great leaving it as is, but a variable should definitely be faster.

src/InputHandler.ts
```diff
@@ -19,24 +20,254 @@ import { wcwidth } from './CharWidth';
  * each function's header comment.
  */
 export class InputHandler implements IInputHandler {
```
Tyriar: > Still I dont quite understand the separation level of InputHandler vs. Terminal - some escape sequence actions directly access terminal internals while others do this via an InputHandler method. Maybe the direct accesses could also be managed by InputHandler?

The thinking here was mainly to pull a bunch of stuff out of `Terminal` to make it easier to work with, one of the big things I wanted to pull out were all the header comments like this so they were all centrally located:

```ts
  /**
   * CSI s
   *   Save cursor (ANSI.SYS).
   */
```
`IInputHandlingTerminal` ended up being added to expose member of `Terminal` specifically to `InputHandler` and nowhere else, to better leverage TypeScript. I have mixed feeling about it; on one hand it's great that `Terminal.ts` is not 4000 lines long, on the other hand `IInputHandlerTerminal` is large and ugly and the separation is not the best.

If I had more time to refine this I would probably do the following:

- Keep `InputHandler`
- Move `prefix`, `params`, `postfix`, `currentParam` into `Parser` (there was something blocking me from doing this earlier, forgot what it was)
- Add `surrogate_high` to `Parser` (it looks like it's already been removed from Terminal? Opps)
- Try move more of `Terminal` into `InputHandler` if functions are only used by `InputHandler` with the goal being to make both `Terminal.ts` and `InputHandlerTerminal` as lean as possible
- Try remove`Parser`'s dependency on `Terminal`
- Remove `Terminal.error` in favor of `console.log`

src/InputHandler.ts
jerch: Oh ok, gonna change it.

src/EscapeSequenceParser.ts
jerch: Yes a single register arg is better. :+1: 

The performance is worse for 2 reasons:

- in the parse loop I have to aggregate collect and flag to one key, this is done atm with a helper string with all its drawbacks (malloc & gc), maybe I can refactor collect internally to a typed array and just pass the reference around, btw same goes for the params array (biggest performance bummer in the parser before moving the callback lookups into the parser)
- numercial property lookups are faster than string lookups -  this is somewhat surprising since JS does all property lookups via the string representation (even for array indices). Maybe number to string conversion is much more efficient than eval the content of an arbitrary string, v8 also uses some internalizing tricks for often used short strings (boils down to a simple pointer comparison). See also https://jsperf.com/string-vs-integer-object-indices

This summed up the throughput for ESC/CSI heavy input is at ~60 MB/s while typical `ls` output with many print actions and only few CSI/EXECUTE actions is at ~120 MB/s. (measured with dummy callbacks to see plain parser performance).

src/EscapeSequenceParser.ts
jerch: Lol right. :blush:

src/InputHandler.ts
jerch: Thanks for the clarification.

Sounds reasonable to keep `InputHandler` as the main escape sequence input point. The parser in this PR does not distingish between `prefix` and `postfix`, instead it sums all intermediate characters into `collect`. I think this is mainly due to the fact that both do not appear in a single valid sequence (the author motivates this somewhere in the docs, cant remember exactly). I already moved `surrogate_high` to `InputHandler` in this PR, since it is only input related and not used anywhere else.

I can try to refine some of the input handling with decoupling in mind if you like (and more tests). But I can't do much at Terminal.ts itself since I have literally no clue about the other components like rendering and user interaction hooks. Though this might get important if a webworker separation shall be the goal in the future, but its definitely out of scope of this PR.

Offtopic:
Btw I am not sure if the character code in the cell buffer always contains the right value `[number, string, number, <this one>]`. What is it used for?

jerch:
@Tyriar With the last commits `InputHandler` should have all needed methods to control the terminal (if I did not miss a method from `Terminal`). I have not yet removed the original methods, not sure if this would break important parts of the API.

From here on it should be straight forward to polish `InputHandler` further:

- tests tests tests, there are only 2 tests atm lol
- revise `IInputHandlerTerminal`
- optimize code (least priority imho)

src/EscapeSequenceParser.ts
Tyriar: How about making `_escHandlers` a `{ [key: number]: { [key: number]: Handler }}`? That way you could access using `_escHandlers[collect.charCodeAt(0)][flag.charCodeAt(0)]` (with null checks).

src/InputHandler.ts
Tyriar: > I can try to refine some of the input handling with decoupling in mind if you like (and more tests)

I'd try to minimize other refactors in this PR, this is already turning into quite the beast that will be tricky to merge in a swift manner.

> Btw I am not sure if the character code in the cell buffer always contains the right value [number, string, number, <this one>]. What is it used for?

It's meant to always contain the character code, except for multi-byte chars (note sure what happens here 😕). I added that as `.charCodeAt(0)` seemed to be slowing things down, but I'm not confident I measured this right.

src/InputHandler.ts
```diff
@@ -428,7 +412,7 @@ export class InputHandler implements IInputHandler {
    * CSI Ps @
    * Insert Ps (Blank) Character(s) (default = 1) (ICH).
    */
-  public insertChars(params: number[]): void {
+  public insertChars(params: number[], collect?: string): void {
```
Tyriar: It's probably better to leave the `collect` param out if it's not being used.

src/EscapeSequenceParser.ts
jerch: This would not work properly, because `collect` must be able to store at least 2 chars. Also I would have to maintain a tree of handlers :(

I thought into this direction:
```js
> var key = new Uint32Array(1)
undefined
> var collect = new Uint8Array(a.buffer)
undefined
> key
Uint32Array [ 0 ]
> collect
Uint8Array [ 0, 0, 0, 0 ]
> key[0] = 1
1
> collect
Uint8Array [ 1, 0, 0, 0 ]
> key
Uint32Array [ 1 ]
> key[0] = key[0] << 8 | 1
257
> collect
Uint8Array [ 1, 1, 0, 0 ]
> key[0] = key[0] << 8 | 57
65849
> collect
Uint8Array [ 57, 1, 1, 0 ]
```
This would give me `collect` and `key` almost for free. Well works only with typed array support.

src/InputHandler.ts
jerch: Yes was wondering about multibyte chars, since it always contains the character code of the first byte only (and undefined for 2nd cell of wide char). If you need the codepoint code instead with the right code for multibyte as well it could be simplified (gets calculated anyways few lines above).

jerch: Correction: always contains the character code of the first 2 bytes (USC-2) and not the unicode codepoint, (which differs for surrogate chars)

src/EscapeSequenceParser.ts
jerch: I suggest not to optimize this in this PR. I think it needs some serious testing and the performance benefit is unclear until some benchmarks showed otherwise. Gonna change the API to a single arg only for now as you suggested.

jerch:
@Tyriar I think I have addressed all from above with the last commits. All thats left to be implemented is the DCS from your parser. And tests, of course :smile:

jerch:
Here some numbers for the parser with dummy handlers:
```
throughput     action                               payload: 50 MB of
197.86 MB/s    PRINT action                         a
 60.36 MB/s    EXECUTE action                       \n
 45.94 MB/s    ESC action with only a flag          ESC E
 19.34 MB/s    ESC action with collect and flag     ESC % G
 36.18 MB/s    CSI action with only a flag          CSI A
 46.43 MB/s    CSI action with collect and flag     CSI ? p
 42.88 MB/s    CSI action with params and flag      CSI 1;2 m
 18.04 MB/s    OSC action                           OSC 0;hi ST
```
The direct comparison is a bit unfair since EXECUTE has to trigger an action for every single byte while the PRINT literally does nothing for one byte (just increments a counter and compares few numbers). Note the two lame ducks, both contain direct string handling in the parser :open_mouth: Especially the difference for "ESC with collect and flag" compared to "ESC with only a flag" is weird since it boils down to
```js
this._escHandlers['<some char>' + String.fromCharCode(code)]
```
compared to
```js
this._escHandlers['' + String.fromCharCode(code)]
```
at code level. Seems the intermediate string creation doubles the runtime. For OSC the char by char string concatenation might be the root of all slowiness.
Well ESC and OSC actions are almost never in output heavy apps, though ESC and CSI actions are quite common in ncurses like apps.

With more realistic data it is not that bad at all:
```
157.47 MB/s    ls -lR /usr/lib (with color support)
 57.82 MB/s    truecolor image with SGR (CSI heavy with two CSI actions per cell)
 64.56 MB/s    midnight commander startup (ncurses like app)
```

jerch:
In a branch I tried to replace the params array with a typed array that is carried around to avoid malloc and gc. This gives a performance boost for CSI with params (runs at ~70 MB/s compared to ~43 MB/s), but involves ugly C-ish changes to `InputHandler` methods due to the borrowed reference to params and an explicit length argument. Those changes make `InputHandler` methods hard to use from outside. Also the speed benefit for real data is not that great (mc startup is at 70 MB/s vs. 64 MB/s before). Imho the code degradation surpasses the speed advantage - therefore not worth to be applied.

To not overcomplicate this PR I gonna stop the performance adjustments, any left could be issued separately.

jerch:
@Tyriar 
Well there is one major hole in testing - the fall-back to an array based transition table is not covered by test cases. I dont expect surprises here (tested it hardcoded at it passed the tests), still the test battery does not cover it automatically. Technically all parser tests should run a 2nd time with an array transition table (tbh I dont know how to achieve this with mocha without copying all the test cases lol).

Imho some of the FIXMEs need a decision before merging:
- C0 control codes without an explicit handler - what to do with those? To reassemble the old behavior those should call `InputHandler.print`. I still think printing those is faulty behavior, but maybe we should do to keep in line with old parser and address it later?
- C1 control codes - enable or disable? They were not part of xterm.js before, to reassemble old behavior should they be disabled?
- `DECRQSS`and the valid/invalid flag: I dont quite understand the old code for this but it seems the response always contained `0` as validation indicator. Problem here is the fact that DEC specifies `0` as valid and xterm spec as invalid (they flipped the meaning). I have no clue which one is right but would guess that most modern apps would follow the xterm meaning.

jerch:
JS array based transition table is now covered by the tests. All that remains to be decided is the list of FIXMEs above.

jerch:
@Tyriar I reread some of the "I want a custom escape sequence XY" issues and it got obvious that the OSC part of the parser should follow a similar API to the DCS part, because people want to use it with arbitrary lengthy data (see #614 and #576). Currently OSC works ok for short data but performance will suffer if people start to use the terminal for those "media enhancements".
I can rewrite the OSC API, proside is a more unique "look'n feel" of the API + ~5x speed gain for OSC, downside the rewrite and additional tests. What you think?

Tyriar:
> I can rewrite the OSC API, proside is a more unique "look'n feel" of the API + ~5x speed gain for OSC, downside the rewrite and additional tests. Since it would involve API changes it might be better to do it before this PR lands. What you think?

Let's definitely defer any more changes like this, I want to merge this ASAP to prevent it from becoming stale.

jerch:
> Let's address it later after we're sure there are no major issues with the changes. Shall we create a follow up issue?

Good idea for easier tracking. There is at least one issue that might be affected by this (cant find it atm, was something about a terminal game with weird control code output).

> Looks like this was fixed upstream after the fork chjj/term.js@36c1e5c

Yup, it follows the xterm meaning there too. This also contains the DCS sequence for custom keys and a tmux passthrough thing. Might be worth a closer look in later on.

Tyriar:
C0 explciit handler: https://github.com/xtermjs/xterm.js/issues/1461
Enable C1: https://github.com/xtermjs/xterm.js/issues/1462

And looks like the `DECRQSS` valid flag was fixed.

src/InputHandler.ts
```diff
@@ -4,12 +4,105 @@
...
+    this._terminal.send(C0.ESC + 'P0' + '+r' + this._data + C0.ESC + '\\');
+  }
+}
+
+/**
+ * DCS $ q Pt ST
+ *   DECRQSS (https://vt100.net/docs/vt510-rm/DECRQSS.html)
+ *   Request Status String (DECRQSS), VT420 and up.
+ *   Response: DECRPSS (https://vt100.net/docs/vt510-rm/DECRPSS.html)
+ *   FIXME: xterm and DEC flip P0 and P1 to indicate valid requests - which one to go with?
```
Tyriar: Let's fix this before merge

jerch: It returns atm 1 for a valid and 0 for an invalid request (xterm style). Should this be flipped? If not I can simply remove the comment.

Tyriar: Ah you're right it is already fixed 😄 let's remove the FIXME

src/InputHandler.ts
```diff
@@ -4,12 +4,105 @@
...
+
+  put(data: string, start: number, end: number): void {
+    this._data += data.substring(start, end);
+  }
+
+  unhook(): void {
+    switch (this._data) {
+      // valid: DCS 1 $ r Pt ST (xterm)
+      case '"q': // DECSCA
+        return this._terminal.send(C0.ESC + 'P1' + '$r' + '0"q' + C0.ESC + '\\');
```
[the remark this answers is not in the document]
jerch: Will that work 'ES5' as target? Thought this is ES6 style.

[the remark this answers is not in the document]
Tyriar: TS compiles it down. This in Renderer.ts:

```ts
this._terminal.screenElement.style.width = `${this.dimensions.canvasWidth}px`;
```

Becomes:

```js
this._terminal.screenElement.style.width = this.dimensions.canvasWidth + "px";
```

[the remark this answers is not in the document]
jerch: Oh sweet, gonna change it.

Tyriar:
@jerch I'll merge once I get another go ahead from you :smiley: 

I can pull this into vscode insiders in a few days, I just did a big update with heaps of rendering changes so want to give that some bake time first.

jerch:
Guess I am done with this PR.

Note that I already have some perf optimizations pending for `InputHandler.print`, I can get them PR'ed as soon as this PR has landed.

Tyriar:
:rocket: :fearful:

jerch:
:sweat:

PR #1404, opened 2018-04-23
Add no else return tslint rule

Tyriar:
This is a personal preference, any opposition?


mofux:
LGTM 👍I generally prefer early returns over if else statements.

PR #1407, opened 2018-04-24
Enforce upper case for public const/enum members and const

### LINKED ISSUE #1406: Some enums aren't using the right naming convention
We should use this convention everywhere:

https://github.com/xtermjs/xterm.js/blob/716a8d592676ada264d9b9b7ebec2f1f3277a4c6/src/Parser.ts#L153-L162

### LINKED ISSUE #1408: All enums should be const
There's a big difference in the compiled output, we never want the enums to change we they should be const.

```ts
enum a {
	b,
	c,
	d
}

const enum A2 {
	B2,
	B3,
	B4
}

console.log(a.b);
console.log(A2.B2);
```

Compiles to:

```ts
var a;
(function (a) {
    a[a["b"] = 0] = "b";
    a[a["c"] = 1] = "c";
    a[a["d"] = 2] = "d";
})(a || (a = {}));
console.log(a.b);
console.log(0 /* B2 */);
```

Tyriar:
Fixes #1406 
Fixes #1408

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "npm-run-all": "^4.1.2",
     "sorcery": "^0.10.0",
     "tslint": "^5.9.1",
+    "tslint-consistent-codestyle": "^1.13.0",
     "typescript": "~2.7.1",
     "vinyl-buffer": "^1.0.0",
     "vinyl-source-stream": "^1.1.0",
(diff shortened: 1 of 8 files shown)


Tyriar:
Created https://github.com/ajafff/tslint-consistent-codestyle/issues/67 to see why the rule ignores exports

Tyriar:
@bgw I'm terrified

PR #1413, opened 2018-04-25
Batch background draws together to reduce draws

bgw:
Adjacent cells on the same row sharing the same background color will be drawn using the same `fillCells` call. This should make drawing an application with a solid or mostly solid background color (e.g. vim) faster.

I set vim to draw the background color for a file, and started scrolling through it. Before, `_drawBackground` was taking around 6ms per frame. After this commit, it was taking around 0.5ms per frame.

Based on prior experience, I'd expect these results to be more drastic for larger terminal windows. The demo's window is pretty small.

### DIFF
--- a/src/renderer/BaseRenderLayer.ts
+++ b/src/renderer/BaseRenderLayer.ts
@@ -26,7 +26,7 @@ export abstract class BaseRenderLayer implements IRenderLayer {
     private _container: HTMLElement,
     id: string,
     zIndex: number,
-    private _alpha: boolean,
+    protected _alpha: boolean,
     protected _colors: IColorSet
   ) {
     this._canvas = document.createElement('canvas');
--- a/src/renderer/BaseRenderLayer.ts
+++ b/src/renderer/BaseRenderLayer.ts
@@ -26,7 +26,7 @@ export abstract class BaseRenderLayer implements IRenderLayer {
     private _container: HTMLElement,
     id: string,
     zIndex: number,
-    protected _alpha: boolean,
+    private _alpha: boolean,
     protected _colors: IColorSet
   ) {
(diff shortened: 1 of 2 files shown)


src/renderer/TextRenderLayer.ts
```diff
@@ -128,18 +128,58 @@ export class TextRenderLayer extends BaseRenderLayer {
...
+        nextFillStyle = this._colors.ansi[bg].css;
       }
+
+      if (prevFillStyle === null) {
+        // This is either the first iteration, or the default background was set. Either way, we
+        // don't need to draw anything.
+        startX = x;
+        startY = y;
+      } if (y !== startY) {
+        // our row changed, draw the previous row
```
Tyriar: I think this can be merged with the "flush the last color we encountered" if it's checking on the last column in the row instead.

bgw: You'd actually need to run

    this.fillCells(startX, startY, x - startX + width, 1);

whereas the fillstyle check runs

    this.fillCells(startX, startY, x - startX, 1);

Plus, this introduces issues if the last column has a different color than the column before it.

This is filling the previous cells we just iterated over, whereas filling on the last column involves filling the cell we're currently on, and all of the complexities that come with that.

src/renderer/BaseRenderLayer.ts
```diff
@@ -26,7 +26,7 @@ export abstract class BaseRenderLayer implements IRenderLayer {
     private _container: HTMLElement,
     id: string,
     zIndex: number,
-    private _alpha: boolean,
+    protected _alpha: boolean,
```
Tyriar: It doesn't look like this is being used

PR #1414, opened 2018-04-25
npm/npm#10074 - change prepublish to prepublishOnly

pro-src:
https://docs.npmjs.com/misc/scripts#prepublish-and-prepare

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "build:docs": "jsdoc -c jsdoc.json",
     "tsc": "tsc",
     "build": "gulp build",
-    "prepublish": "npm run build",
+    "prepublishOnly": "npm run build",
     "coveralls": "gulp coveralls",
     "webpack": "gulp webpack",
     "watch": "gulp watch"


Tyriar:
@pro-src so `prepare` runs on both `install` and `publish` and `prepublishOnly` runs only on `publish`? I think `prepublishOnly` might be even better then as currently `npm install && npm run watch` will build twice which wastes 5-10 seconds or so.

pro-src:
I think `prepublishOnly` is a good option for now that should remove the deprecation notice but only temporarily as the plan ultimately is to replace `prepublishOnly` with `prepublish` for npm@6? Leaving only the more inituitive `prepare`(npm install/upload) and `publish`(upload only) behaviors when the dust finally settles.

pro-src:
Feel free to close this if you don't mind the deprecation notice as it will fix itself eventually.

Tyriar:
Oh I see, well it looks like the eventual `prepublish` behavior is what we're after.

PR #1417, opened 2018-04-26
Stop publishing test files to npm - closes #1361

### LINKED ISSUE #1361: We're publishing test js files to npm
screen shot 2018-03-29 at 11 49 45 am

pro-src:
Changes `.npmignore`


pro-src:
This configuration has been tested using the following code with a test [repo](https://github.com/pro-src/xterm.js-packaging) and published as `@pro-src/xterm-packaging` for inspection.

```sh
# Created an empty repository to push at to
# https://github.com/pro-src/xterm.js-packaging

# Clone my fork
git clone https://github.com/pro-src/xterm.js xterm.js-packaging
cd xterm.js-packaging/

# Checkout my PR
git checkout packaging

# Add, commit, and push to the test repo
rm -rf .git/
git init
git remote add origin https://github.com/pro-src/xterm.js-packaging
git checkout -b master
git add .
git commit -m 'test'
git push -uf origin master

# Setup to publish using npm scopes
# package['name']       = '@pro-src/xterm-packaging'
# package['repository'] = 'https://github.com/pro-src/xterm.js-packaging'
# package['version']    = '3.3.1' # minor version bump
npm i # currently run build for us

# Publish the test release
# mv build/ dist/
# npm publish --access=public
bin/release 3.3.1

# Pull down and examine the resulting package files
cd ..
mkdir example-project
cd example-project
npm init -y
npm install --save --ignore-scripts @pro-src/xterm-packaging
tree .
```

pro-src:
The output of the tree command can be viewed [here](https://gist.github.com/pro-src/a1c40cc96395523c355ee7eac7b60ad0#file-files-md).

.npmignore
```diff
@@ -1,16 +1,49 @@
...
 docs/
-npm-debug.log
 /.idea/
-.env
+.vscode/
 build/
+fixtures/
+coverage/
+
+package-lock.json
```
Tyriar: This probably shouldn't be here, npm would probably expect this if we do ever want to use it.

.gitignore
```diff
@@ -1,25 +1,37 @@
...
-*.swp
-.lock-wscript
-lib/
-Makefile.gyp
-*.Makefile
-*.target.gyp.mk
-*.node
-example/*.log
+# Blacklist - exclude everything except files without extensions
+*.*
```
Tyriar: I'm not the biggest fan of blacklisting everything, especially for the `.gitignore`. Can we go back to the old model of blacklist only for `.gitignore` without impacting npm publish assets?

package.json
```diff
@@ -11,33 +11,6 @@
   "types": "typings/xterm.d.ts",
   "repository": "https://github.com/xtermjs/xterm.js",
   "license": "MIT",
-  "files": [
```
Tyriar: Glad to see package.json getting simplified 👍

pro-src:
No we can not :P
Joking... lol

The known file extensions aren't blacklisted.  It would only be a nuisance if a new/uncommon file type was being introduced but in that case `git add -f my-non-js.weridfile` would allow it.

I'll update this PR as soon as I hear back from you.

.gitignore
```diff
@@ -1,25 +1,37 @@
...
+
+# Whitelist - entries to be included must be negated with "!"
+!*.yml
+!*.md
+!*.js
+!*.js.map
+!*.json
+!*.css
+!*.ts
+!*.d.ts
```
pro-src: Files with these extensions and extension-less files are not blacklisted.

pro-src: @Tyriar should I go ahead and ammend with the requested changes?

.gitignore
pro-src: Yes

.npmignore
pro-src: Will be removed. Idk why I thought to put that in there... It's not being created per the `.npmrc`

pro-src:
I'm against the .gitignore blacklist now as I checked the test repo and both `.gitignore` and `.npmignore` are missing! They're both considered as having extensions even though they're basename-less and I didn't whitelist them... Lots of irony there but it wouldn't have been an issue if merged because git is already tracking those. Still, it alluded my attention. 😂

pro-src:
Don't merge... I found an issue. Understand I modeled after the original and my line of thinking has been to reproduce the original behavior and then add in the desired changes. The demo is being partially included. I think it would have been originally if it existed when publishing. As it's not in the current package on npm, should I blacklist it? Appending `demo/` to .npmignore will fix it.

Tyriar:
`./demo` should definitely not be published

pro-src:
@Tyriar `.gitignore` files aren't really that complicated, are they now?

Updated and ready to be merged :stuck_out_tongue_winking_eye:

package.json
LinusU: @pro-src there is a `ignore` section on lines 5 thru 9 in this file, that should probably go as well?

pro-src: That's a matter of preference. I think @Tyriar indirectly indicated that he'd prefer those entries, preferring not to have a whitelist (aka .gitallow), in the context of the .gitignore file which was removed from this PR.

Are those entries absolutely needed? No, removing them would not break anything important.

What is the behavior? Recursively includes some files (by extension) without having to modify .npmignore

I think those entries are currently in `package.json`.

So do we want a whitelist or a blacklist?

pro-src: I prefer the whitelist behavior of the unmodified PR as files won't be accidentally added to github or npm with the minor drawback(or feature) of having to explicitly allow any new uncommon files types. Considering it would have prevented test files from ever being published to npm in the first place. That's what it bulls down to... Preference. If the `.npmignore` should not behave as a whitelist then this PR should be modified to only use `.npmignore` for blacklisting files but then _why_ was `package.json` used to whitelist files???

IMHO, This solution is solid and achieves what was originally intended (by reason of the current .gitignore and package.json). All other issues are _other_ issues. Not that I mind modifying the PR.

pro-src:
The issue that this PR fixes is trivial and depends on the preference(s) of the core maintainer(s). Please feel free to remove the "help wanted" label from the associated issue.

package.json
Tyriar: I think @LinusU means https://github.com/xtermjs/xterm.js/pull/1417/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L5 might not be doing anything now with the extensive `.npmignore`. I can't even see that key in the package.json docs 😕 https://docs.npmjs.com/files/package.json

Tyriar:
@pro-src FYI I still want to merge this, been swamped and then MS Build happened. Currently working through a backlog of 190 GitHub notifications ☹️. We just need to resolve the `ignore` key thing

Tyriar:
Thanks @pro-src 😃

package.json
pro-src: @Tyriar, just going to open it back up, like _watch and learn boiiiiy!_ :joy:

I'm confused, your referring to which key? `files?`
I was referring to lines 15 and 16 of https://github.com/xtermjs/xterm.js/blob/master/package.json#L14

`.gitallow` was just a metaphor.

Tyriar: There was an `ignore` key which I think is now completely superseded by .npmignore, see https://github.com/xtermjs/xterm.js/pull/1417/commits/210a312fbff4a06f83f033ffdd3464cb4d1073c9

PR #1420, opened 2018-04-26
Update README.md

### LINKED ISSUE #1419: Consider changing the octal escapes in README.md to unicode escapes.
If the example on line 44 of README.md is used in the latest browsers version's strict mode, it will throw an error because octal escapes are deprecated in es2015. Thus the proper way to use ANSI escapes is via unicode escape character `'u001B['`. Using the Unicode escape will avoid confusion and point users in the right direction from the start.

Webpack, for example, adhering to the es-module loader spec will subtly add the `'use strict';` statement to each module definition if it encounter a synchronous `import` statement.

Meaning users which follow the example in conjunction with the `import` statement being used will be faced with this error without explicitly enabling strict mode in any environment which follows the spec.

pro-src:
Closes #1419 

Replace the deprecated octal escapes with hex escapes.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Deprecated_octal

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ To start using xterm.js on your browser, add the `xterm.js` and `xterm.css` to t
       <script>
       	var term = new Terminal();
         term.open(document.getElementById('terminal'));
-        term.write('Hello from \033[1;3;31mxterm.js\033[0m $ ')
+        term.write('Hello from \u001B[1;3;31mxterm.js\u001B[0m $ ')
       </script>
     </body>
   </html>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,7 +41,7 @@ To start using xterm.js on your browser, add the `xterm.js` and `xterm.css` to t
       <script>
       	var term = new Terminal();
         term.open(document.getElementById('terminal'));
-        term.write('Hello from \033[1;3;31mxterm.js\033[0m $ ')
+        term.write('Hello from \u001B[1;3;31mxterm.js\u001B[0m $ ')
```
Tyriar: Interesting, didn't know they were deprecated. I tend to prefer hex escapes as they're a little more compact, can we do that instead? `\x1b`

PR #1421, opened 2018-04-26
Add remaining naming conventions into tslint

Tyriar:
I think that's everything 😄 

The one that's commented out introduces a fair number of changes as it's inconsistent throughout the codebase. Let's defer that one for when there are less big PRs out.


mofux:
LGTM 👍

Issue #580, opened 2017-03-03
Support strikethrough text style


Tyriar:
I notice these aren't implemented:

- ~~Dim `\e[2m` -> `opacity: 50%`~~ (Done)
- ~~Italics `\e[3m` -> `font-style: italic`~~ (Done https://github.com/xtermjs/xterm.js/pull/1422)
- Strikethrough `\e[9m` -> `text-decoration: line-through`

LucaT1:
Is this implemented as of today?
The css is here, but for me it's not working inside of vim, macOS

Tyriar:
It's not implemented, what CSS are you referring to?

LucaT1:
Sorry for the CSS part, I thought to have seen it [there](https://github.com/sourcelair/xterm.js/blob/master/dist/xterm.css#L208-L222)

Are you planning on implementing this soon or is this a low priority task?

Tyriar:
Those are for the ones that are implemented.

This is low priority and is blocked by https://github.com/sourcelair/xterm.js/pull/756 which is reimplementing how the character attributes are stored.

mofux:
Here is a nice script to test the different styles:

https://gist.githubusercontent.com/mxmerz/92e97cd27857a9ba787b/raw/c16c216d11c3fa6e7f0254e2d5f485164cddfb96/ANSI-Test.sh

LucaT1:
Cool! 😍

flybayer:
Oh man, I want this so bad. 
It looks like this is no longer blocked by anything, correct?

felixfbecker:
Strike through would be really great. Git supports it as a diff style: `git config --global diff.color.old "red strike"`

PowerGit uses it for gone branches:

![image](https://user-images.githubusercontent.com/10532611/57560939-5da02180-7389-11e9-9306-06d8e3353b5a.png)

Tyriar:
If anyone wants to help out with this you will want to copy how underline works in the DOM (DomRendererRowFactory.ts) and canvas renderers (TextRenderLayer.ts), for the latter a new method like `fillBottomLineAtCells` will need to be done which draws a line through the middle.

Tyriar:
DOM and canvas renderer support is merged thanks to @silamon, webgl is still outstanding. It's a little harder there since we don't do underline yet either https://github.com/xtermjs/xterm.js/issues/2251

PR #1422, opened 2018-04-26
Add support for italic rendering

flybayer:
This adds support for italic rendering. 

You can test it with this command:
```bash
echo -e '\x1B[3m italic \x1B[m off'
```

Let me know if you'd like any changes or anything else!

Related to #580

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -1263,6 +1263,9 @@ export class InputHandler implements IInputHandler {
       } else if (p === 1) {
         // bold text
         flags |= FLAGS.BOLD;
+      } else if (p === 3) {
+        // italic text
+        flags |= FLAGS.ITALIC;
       } else if (p === 4) {
         // underlined text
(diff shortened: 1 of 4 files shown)


Tyriar:
I think the problem is the ordering when using `normal` as the weight before setting `.font` to this:

`normal italic 22.5px courier-new, courier, monospace`

Results in the fallback:

`10px sans-serif`

I'll look into fixing this.

Tyriar:
Should be fixed now, just needed to rearrange the font attributes.

Tyriar:
@bgw FYI I'll merge this in, conflicts should be minimal in your PR

ceoshikhar:
image
Still the italics aren't rendering for me. Any solution?

Tyriar:
Works for me:

image

Guessing it's your font config.

shivapoudel:
@Tyriar Help is appreciated for this Cascadia Code ticket https://github.com/microsoft/cascadia-code/issues/515

PR #1428, opened 2018-05-02
Call IRenderer.onCharSizeChanged at right time

Tyriar:
Even though Renderer just defers this to onResize, future IRenderer
implementations may take advantage of onCharSizeChanged specifically.


mofux:
LGTM 👍

Issue #1337, opened 2018-03-16
Use project references coming in TypeScript v3


Tyriar:
This is an issue for the future when TS 3 is released. With it are coming composite projects (ie. project references). This will allow us to section out our various layers and put hard references between them, so we could have a shared project instead of just a folder (and enforce compile/lint rules for that environment). Eventually maybe we would have a web worker and browser project as well.

Tyriar:
I have a WIP at https://github.com/Tyriar/xterm.js/tree/1337_project_refs_part_2, we need to wait until 3.1 before we adopt this.

Tyriar:
Apparently watch is much better in TS 3.3

PR #1432, opened 2018-05-04
Implement a fallback DOM renderer

### LINKED ISSUE #1360: Provide fallback DOM renderer
It's become clear over the last few months that certain combinations of GPUs/browsers just fail at drawing to canvas and the terminal slows to a crawl as a result. We should add an alternate DOM renderer as a fallback that can be switched to. On Electron you could then use [`app.getGPUFeatureStatus()`](https://electronjs.org/docs/api/app#appgetgpufeaturestatus) to determine if the hardware acceleration is disabled.

Tyriar:
This change is motivated in order to provide a fallback for environments that do not support or who have performance issues using GPU canvas acceleration.

Fixes #1360

### Notes

- A lot of the implementation is similar to the old renderer which can be viewed in #938
- I explored making this an addon but it depends on too many internal objects like `ColorManager`, `RenderDebouncer` and `EventEmitter`. Once TS 3 is released we should refactor out a "platform"/"base" project which can be shared easily depended on by the core and addons (and speed up builds since it only need to be rebuilt if files within it changes), see https://github.com/xtermjs/xterm.js/issues/1337. In the meantime I think keeping it in core as an option is the way to go


src/renderer/dom/DomRenderer.ts
```diff
@@ -0,0 +1,359 @@
...
+        ` background-color: ${this.colorManager.colors.selection.css};` +
+        `}`;
+    // Colors
+    this.colorManager.colors.ansi.forEach((c, i) => {
+      styles +=
+          `.xterm .${FG_CLASS_PREFIX}${i} { color: ${c.css}; }` +
+          `.xterm .${BG_CLASS_PREFIX}${i} { background-color: ${c.css}; }`;
+    });
+
+    this._themeStyleElement.innerHTML = styles;
```
mofux: I'm not sure if it is wise to create these styles for every single terminal instance. It will put quite some load on the CSS rendering engine with every new instance of the terminal spawned (because the classes created are not scoped to the DOM tree the style element is attached to). Shadow DOM might solve this problem...?

src/Terminal.ts
```diff
@@ -703,7 +705,11 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
     // Performance: Add viewport and helper elements from the fragment
     this.element.appendChild(fragment);
 
-    this.renderer = new Renderer(this, this.options.theme);
+    if (this.options.rendererType === 'canvas') {
```
mofux: The only known renderer type seems to be 'canvas', any other string will use the dom renderer. Shouldn't we expect that 'dom' is passed if the DOM renderer should be used, and throw an exception if an unsupported renderer name gets passed?

mofux:
In general I'm not a big fan of the current renderer API because it is too closely coupled. I'd much rather have a layer in the middle that holds abstract cell decoration instructions (foreground, background, bold, italic, underline, link, selected, etc...). Things like the linkifier or the selection manager would provide cell decorations instead of drawing directly - and then it would be up to the renderer to support and draw these decoration types. It will probably involve a lot of refactoring, but would open up xterm.js to more advanced plugins that could influence rendering more easily. Maybe we should iterate on something like that in a separate issue, potentially for a v4?

Tyriar:
> In general I'm not a big fan of the current renderer API because it is too closely coupled. I'd much rather have a layer in the middle that holds abstract cell decoration instructions (foreground, background, bold, italic, underline, link, selected, etc...). Things like the linkifier or the selection manager would provide cell decorations instead of drawing directly - and then it would be up to the renderer to support and draw these decoration types. It will probably involve a lot of refactoring, but would open up xterm.js to more advanced plugins that could influence rendering more easily. Maybe we should iterate on something like that in a separate issue, potentially for a v4?

More targeted PRs against those components would be a good in the future. We can also tweak `IRenderer` in the future, just take the DOM renderer along with the ride. I think the refactors you're proposing would also better enable us to pull the dom renderer into an addon which is where it should go in the end.

src/Terminal.ts
Tyriar: 👍 will fix this up

src/renderer/dom/DomRenderer.ts
Tyriar: This function is called sparingly though, only when the theme changes. I didn't want to spend too much time optimizing since this is just the fallback experience.

The way I was thinking we could fix this is by adding an ID (or a unique class) to the terminal element.

Tyriar:
@mofux feedback addressed, ready for another round

src/Terminal.ts
```diff
@@ -704,7 +706,11 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
     // Performance: Add viewport and helper elements from the fragment
     this.element.appendChild(fragment);
 
-    this.renderer = new Renderer(this, this.options.theme);
+    switch (this.options.rendererType) {
+      case 'canvas': this.renderer = new Renderer(this, this.options.theme); break;
+      case 'dom': this.renderer = new DomRenderer(this, this.options.theme); break;
+      default: throw new Error(`Unrecognized rendererType "${this.options.rendererType}"`);
```
odrzutowiec: maybe you can add open option to provide custom renderers?

```js
if (this.options.rendererType) {
  this.renderer = new this.options.rendererType(this, this.options.theme)
}
```

vincentwoo:
I'd like to submit a request for at least some (possibly undocumented way) to pull out a DOM render without necessarily attaching xterm to an open DOM element, etc.

src/Terminal.ts
Tyriar: @mofux and I will be improving how renderers attach soon, the DOM renderer will eventually be moved to an addon. Also I'm not sure about swapping out the renderer in its entirety in the API but I definitely want people to contribute layers that could sit on top.

mofux:
Just re-tested an I don't see any issues left. Good to merge from my side 👍

Tyriar:
I'll do another check of it as well now

Issue #1435, opened 2018-05-08
[bug]browser crash


kevinwon1985:
## Details
- Browser and browser version:  chrome 66.0.3359.139
- OS version: mac 10.13.4
- xterm.js version: 3.0, 3.1, 3.2, 3.3, master

### Steps to reproduce

1. use this crash demo commit 
`https://github.com/kevinwon1985/xterm.js/commit/c7dbb0c9808b6899bd6168ae96f4f3fa064eca97`
2. run the demo

kevinwon1985:
here is my pull request https://github.com/xtermjs/xterm.js/pull/1436

Tyriar:
I'd rather not expose API for this as it can very easily be handled on the consumer side. An alternative way to solve the problem is mentioned in https://github.com/xtermjs/xterm.js/issues/1416

Issue #1416, opened 2018-04-26
Browser crash related to fit addon returning geometry with "Infinity" sizes


vincentwoo:
![screenshot 2018-04-25 18 34 47](https://user-images.githubusercontent.com/613320/39281010-58dda396-48b7-11e8-890c-bd4e43fd7dec.png)

Appears to be caused by edge cases around offscreen xterms coming/going onscreen. With `term.renderer.dimensions.actualCellWidth` as `null`, fit returns `Infinity`. If you allow xterm to try to resize itself to Infinity, it consumes all available RAM and crashes the browser.

vincentwoo:
Here's a CodePen that mostly reproduces: https://codepen.io/akanet/pen/zjBgXa

ioquatix:
I think I'm running into the same issue.

ioquatix:
I just stuffed this in at the top:

```
var cellWidth = term.renderer.dimensions.actualCellWidth || 10;
var cellHeight = term.renderer.dimensions.actualCellHeight || 15;
```

and the appropriate adjustment to the bottom:

```
    var geometry = {
        cols: Math.floor(availableWidth / cellWidth),
        rows: Math.floor(availableHeight / cellHeight)
    };
```

Things starting to work a bit better, but obviously the fit is wrong.

ioquatix:
If you put this at the top of fit:

```
    if (!cellWidth || !cellHeight)
      return null;
```

and don't have defaults, it never renders correctly, and thus the renderer `actualCell*` fields never get a value.

vincentwoo:
I am mitigating currently by defaulting to some sane (but wrong) value in the case of Infinity, and re-running `fit` on a timeout. It's not ideal.

ioquatix:
Yeah, but I don't understand why it's necessary to default to the wrong value.... I'm just trying to figure it out.

ioquatix:
Somehow, if I call `terminal.resize(NaN, NaN)`, eventually it starts working. But if I call anything else, e.g. `terminal.resize(10, 10)`, it never works.

ioquatix:
```
		//this.terminal.fit();
		this.terminal.resize(NaN, NaN);
```

just seems so wrong.

ioquatix:
It literally doesn't work at all if I don't have that. The only other time fit is called is when the parent element changes size.

ioquatix:
Somehow, the renderer never computes values unless you do that.

ioquatix:
[...]

PR #1436, opened 2018-05-08
fix 1435: brower crash in some case

kevinwon1985:
- add maxCols and maxRows options, prevent too large rows and cols cause browser crash
- add dimensions check before.


Tyriar:
As mentioned in #1435 I don't think we should add API. This fix is very similar to what is proposed in https://github.com/xtermjs/xterm.js/issues/1416#issuecomment-386906673 though if you want to open another PR focused on that fix.

PR #1438, opened 2018-05-08
Add diag.ai to list of real-world uses 

ledbit:
We use xterm.js to provide terminal access to a local shell from the browser, while in the [app ](https://app.diag.ai)(lightweight local daemon required).


ledbit:
Hmm, unsure why the build failed as this PR modifies README.md

Tyriar:
@ledbit I think that was fixed in another PR 😄. Thanks, looks cool!

PR #1442, opened 2018-05-10
Save and restore the ctx for cached characters

### LINKED ISSUE #1424: Demo rendering appears to be broken
Loading and then pressing enter looks like it's drawing the text at 50% alpha or something.

screen shot 2018-04-30 at 3 11 50 pm

Reproing on my fork's master branch which is up to date with upstream.

bgw:
Drawing dim colors sets the globalAlpha to 0.5, so we need to
save/restore the ctx state. Otherwise, every time we draw some dim
content, our display will get progressively dimmer.

I tested this by running `echo '\u001b[2mfoo'` a few times in zsh.

Fixes #1424


Tyriar:
Thanks @bgw, we'll see if this is fixed when it goes out in VS Code Insiders 😄

PR #1444, opened 2018-05-10
Update how to use addons with Typescript

pengx17:


README.md
```diff
@@ -57,7 +57,18 @@ The proposed way to load xterm.js is via the ES6 module syntax.
 import { Terminal } from 'xterm';
 ```
 
-*Note: There are currently no typings for addons so you will need to upcast if using TypeScript, eg. `(<any>xterm).fit()`.*
+*Note: There are currently no typings for addons if they are accessed via extending Terminal prototype, so you will need to upcast if using TypeScript, eg. `(<any>xterm).fit()`.*
```
Tyriar: I think thus would be better called out briefly in the Addons section below. applyAddon should be recommended as an alternative imo, not the main way.

Tyriar:
Thanks 😄

PR #1446, opened 2018-05-11
Fix link cursor CSS

### LINKED ISSUE #1437: addon/webLinks: cursor when mouse hover a link cannot be pointer
When mouse hover a link, will add a class name "xterm-cursor-pointer" to terminal element.
However, there are some style who's priority level is higher than "xterm-cursor-pointer", and the cursor cannot be a pointer:
```
.xterm.enable-mouse-events {
    /* When mouse events are enabled (eg. tmux), revert to the standard pointer cursor */
    cursor: default;
}

.xterm:not(.enable-mouse-events) {
    cursor: text;
}

.xterm-cursor-pointer {
    cursor: pointer;
}
```
may be should change it like this?
```
.xterm.xterm-cursor-pointer {
    cursor: pointer;
}
```

Tyriar:
Ensure pointer overrides the default mouse events mode cursor.

Fixes #1437

### DIFF
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -135,6 +135,10 @@
     cursor: text;
 }
 
+.xterm.xterm-cursor-pointer {
+    cursor: pointer !important;
+}
+
 .xterm .xterm-accessibility,
 .xterm .xterm-message {
     position: absolute;
@@ -152,8 +156,4 @@
     width: 1px;
     height: 1px;
     overflow: hidden;
-}
-
-.xterm-cursor-pointer {
-    cursor: pointer;
 }
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -126,17 +126,17 @@
     line-height: normal;
 }
 
+.xterm {
+    cursor: text;
+}
+
 .xterm.enable-mouse-events {
     /* When mouse events are enabled (eg. tmux), revert to the standard pointer cursor */
     cursor: default;
 }
 
-.xterm:not(.enable-mouse-events) {
-    cursor: text;
-}
-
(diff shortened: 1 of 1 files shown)


src/xterm.css
```diff
@@ -135,6 +135,10 @@
     cursor: text;
 }
 
+.xterm.xterm-cursor-pointer {
```
bgw: I think we can do this a little cleaner if we:

- Remove `cursor: default;` from `.xterm`, since the combination of `.xterm.enable-mouse-events` and `.xterm:not(.enable-mouse-events)` would mean that it's always overridden.
- Replace the `.xterm:not(.enable-mouse-events)` selector with just a `.xterm` selector. `.xterm.enable-mouse-events` can override it, and that gives us the same behavior as the `:not` pseudo-selector would've, with a lower specificity.
- Remove `!important` from this cursor rule. It shouldn't be needed if we reduce the specificity needed for the `cursor: text` property.

Basically, what I'm saying is that now we have:

```css
.xterm {
  cursor: default;
}

.xterm.enable-mouse-events {
  cursor: default;
}

.xterm:not(.enable-mouse-events) {
  cursor: text;
}

.xterm.xterm-cursor-pointer {
  cursor: pointer !important;
}
```

But I think this is equivalent, simpler, and avoids increasing the specificity:

```css
.xterm {
  cursor: text;
}

.xterm.enable-mouse-events {
  cursor: default;
}

.xterm.xterm-cursor-pointer {
  cursor: pointer;
}
```

Tyriar: Yeah that's way better 😛

PR #1449, opened 2018-05-14
Bubble scroll events when scrolling beyond the bounds of the viewport

### LINKED ISSUE #1343: Improve Scrolling Experience
Right now Xterm.js captures all scroll events which leads to an inconvenient user experience in two related cases:

1. When Xterm is small and a user scrolls to the very bottom and is trying to scroll more, nothing will happen. I would expect the whole page to continue scrolling once Xterm reaches the very bottom.

2. When Xterm is large so there is no need for scrolling within Xterm. In this case Xterm captures all scrolling attempts and is stopping the whole page from scrolling. It's especially inconvenient if Xterm occupies most of the screen.

pro-src:
Closes #1343 

The changes in this PR _should_ allow the page to scroll when the terminal reaches it's top or bottom. If the terminal is scrolled then the events are canceled otherwise they bubble up.

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -1092,8 +1092,9 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
     // the shell for example
     on(el, 'wheel', (ev: WheelEvent) => {
       if (this.mouseEvents) return;
-      this.viewport.onWheel(ev);
-      return this.cancel(ev);
+      if (this.viewport.onWheel(ev)) {
+        return this.cancel(ev);
+      }
     });
 
     on(el, 'touchstart', (ev: TouchEvent) => {
@@ -1104,8 +1105,9 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
 
     on(el, 'touchmove', (ev: TouchEvent) => {
       if (this.mouseEvents) return;
-      this.viewport.onTouchMove(ev);
-      return this.cancel(ev);
+      if (this.viewport.onTouchMove(ev)) {
+        return this.cancel(ev);
+      }
     });
   }
 
--- a/src/Viewport.ts
+++ b/src/Viewport.ts
@@ -95,6 +95,28 @@ export class Viewport implements IViewport {
   }
 
   /**
+   * Adjusts the viewport's scrollTop and delegates the actual scrolling
+   * to `onScroll`.
+   * @param amount The amount to scroll
+   * @returns a boolean indicating whether the viewport was scrolled
+   */
+  private _scroll(amount: number): boolean {
+    if (amount === 0) {
+      return false;
+    }
+    if (amount < 0) {
+      if (this._viewportElement.scrollTop === 0) {
+        return false;
+      }
+    }
+    else if (this._viewportElement.scrollTop === this._viewportElement.scrollHeight) {
+      return false;
+    }
+    this._viewportElement.scrollTop += amount;
+    return true;
+  }
+
+  /**
    * Handles scroll events on the viewport, calculating the new viewport and requesting the
    * terminal to scroll to it.
    * @param ev The scroll event.
@@ -112,19 +134,12 @@ export class Viewport implements IViewport {
   }
 
   /**
-   * Handles mouse wheel events by adjusting the viewport's scrollTop and delegating the actual
-   * scrolling to `onScroll`, this event needs to be attached manually by the consumer of
-   * `Viewport`.
+   * This event needs to be attached manually by the consumer of `Viewport`.
    * @param ev The mouse wheel event.
    */
-  public onWheel(ev: WheelEvent): void {
+  public onWheel(ev: WheelEvent): boolean {
     const amount = this._getPixelsScrolled(ev);
-    if (amount === 0) {
-      return;
-    }
-    this._viewportElement.scrollTop += amount;
-    // Prevent the page from scrolling when the terminal scrolls
-    ev.preventDefault();
+    return this._scroll(amount);
   }
 
   private _getPixelsScrolled(ev: WheelEvent): number {
@@ -179,13 +194,9 @@ export class Viewport implements IViewport {
    * Handles the touchmove event, scrolling the viewport if the position shifted.
    * @param ev The touch event.
    */
-  public onTouchMove(ev: TouchEvent): void {
-    let deltaY = this._lastTouchY - ev.touches[0].pageY;
+  public onTouchMove(ev: TouchEvent): boolean {
+    const deltaY = this._lastTouchY - ev.touches[0].pageY;
     this._lastTouchY = ev.touches[0].pageY;
-    if (deltaY === 0) {
-      return;
-    }
-    this._viewportElement.scrollTop += deltaY;
-    ev.preventDefault();
+    return this._scroll(deltaY);
   }
 }
--- a/src/Viewport.test.ts
+++ b/src/Viewport.test.ts
@@ -1,0 +1,114 @@
+/**
+ * Copyright (c) 2016 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import jsdom = require('jsdom');
+import { expect } from 'chai';
+import { MockTerminal } from './utils/TestUtils.test';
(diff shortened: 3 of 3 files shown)


src/Terminal.ts
```diff
@@ -1104,8 +1105,9 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
 
     on(el, 'touchmove', (ev: TouchEvent) => {
       if (this.mouseEvents) return;
-      this.viewport.onTouchMove(ev);
-      return this.cancel(ev);
+      if (<any>this.viewport.onTouchMove(ev) === false) {
+        return this.cancel(ev);
```
pro-src: tsc wouldn't allow me to do a strict equality check without upcasting: cannot `void === boolean`. I'm not sure that this is the best way to handle it.

src/Viewport.ts
```diff
@@ -111,20 +119,35 @@ export class Viewport implements IViewport {
     this._terminal.scrollLines(diff, true);
   }
 
+  /**
+   * Calls onScroll synchronously and determines whether the terminal was
+   * actually scrolled.
+   */
+  private scrollSync(amount: number): boolean {
```
pro-src: This seems to be the best way to avoid code duplication as we can't really prevent the scroll event because changing `_viewportElement.scrollTop` is necessary? I don't really like the method name either, feedback please :confused:

src/Viewport.ts
```diff
@@ -111,20 +119,35 @@ export class Viewport implements IViewport {
...
     const amount = this._getPixelsScrolled(ev);
     if (amount === 0) {
       return;
     }
-    this._viewportElement.scrollTop += amount;
-    // Prevent the page from scrolling when the terminal scrolls
-    ev.preventDefault();
+    if (this.scrollSync(amount)) {
+      // Prevent the page from scrolling when the terminal scrolls
+      ev.preventDefault();
```
pro-src: Is the `ev.preventDefault()` statement meant for third party consumers? The `terminal.cancel` method was always being called after these handlers.

src/Terminal.ts
Tyriar: You would probably just want to changing onTouchMove/onWheel to always return a boolean

src/Viewport.ts
```diff
@@ -99,7 +102,12 @@ export class Viewport implements IViewport {
    * terminal to scroll to it.
    * @param ev The scroll event.
    */
-  private _onScroll(ev: Event): void {
+  private _onScroll(ev?: Event): void {
+    // Prevent duplicate event handling
```
Tyriar: Oh the double call is prevented here, this sort of state between async callbacks should be avoided at all costs imo. What's the duplicate event handling we're concerned about?

src/Viewport.ts
```diff
@@ -179,13 +202,16 @@ export class Viewport implements IViewport {
    * Handles the touchmove event, scrolling the viewport if the position shifted.
    * @param ev The touch event.
    */
-  public onTouchMove(ev: TouchEvent): void {
-    let deltaY = this._lastTouchY - ev.touches[0].pageY;
+  public onTouchMove(ev: TouchEvent): boolean | void {
```
Tyriar: Can't we just return a `boolean` on this and then leave event cancellation up to this part?

https://github.com/xtermjs/xterm.js/blob/7a9f291df9f25570731e6c79ca48b0b37db2591a/src/Terminal.ts#L1091-L1109

src/Viewport.ts
pro-src: OnTouchMove/onWheel both modify `_viewportElement.scrollTop` which causes the scroll event to fire on `_viewportElement` during the next event loop. If we call onScroll manually on the current event loop to determine whether the viewport was actually scrolled then the scroll event will still fire thus onScroll will get called a second time. We could remove this check if we don't modify `_viewportElement.scrollTop` but it's necessary?

pro-src: nvm that, I missed what you were saying and now it doesn't seem necessary to prevent the handling.

src/Viewport.ts
Tyriar: Yeah feel free to remove this as it's handled in `Terminal.cancel`.

src/Viewport.ts
```diff
@@ -111,20 +111,29 @@ export class Viewport implements IViewport {
...
   }
 
+  /**
+   * Calls onScroll synchronously and determines whether the terminal was
+   * actually scrolled.
+   */
+  private scrollSync(amount: number): boolean {
+    const oldYDisp = this._terminal.buffer.ydisp;
+    this._viewportElement.scrollTop += amount;
+    this._onScroll();
```
Tyriar: I don't think `this._onScroll()` shouldn't be needed in this function as it will fire when `this._viewportElement.scrollTop` changes?

pro-src: The event delegation is complicating things.

No, It wouldn't be needed if we could determine whether or not to cancel the events in onTouchMove/onWheel without it and without code duplication. The only other way that I can tell is to copy code from the `_onScroll` and `_terminal.scrollLines` functions to `scrollSync` but `onScroll` would still be called anyway.

The problem is that you can't check to see if `buffer.ydisp` has changed until after `onScroll` executes but that won't happen before returning from the original event handlers without calling it manually.

pro-src: And maybe I'm over complicating things :stuck_out_tongue_winking_eye: 

I haven't exactly tested to see if it `this._onScroll` is called immediately (during the current event loop) upon changing `_viewportElement.scrollTop`. However, I wouldn't think so... lol

I'll check the spec and get back to you on that.

pro-src: I didn't find exactly what I was looking for in the specs however if we're to believe [this mdn article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop) about the event loop then when the event is triggered by modifying `_viewportElement.scrollTop` a new message is added to the message queue but won't be processed until the current message is processed to it's completion which is the end of the current event loop. See "Run-to-completion" in the linked article.

Tyriar: Can we bounds check `this._viewportElement.scrollTop`? If it's 0 we don't need to scroll, I think there's a way to check if it's at the end as well.

pro-src: The bounds checking should be performed on the buffer and not `_viewportElement.scrollTop`? https://github.com/xtermjs/xterm.js/blob/master/src/Terminal.ts#L1213

We would be duplicating `terminal.scrollLines`?

pro-src:
@Tyriar updated.

src/Terminal.ts
```diff
@@ -1092,8 +1092,9 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
...
     on(el, 'wheel', (ev: WheelEvent) => {
       if (this.mouseEvents) return;
-      this.viewport.onWheel(ev);
-      return this.cancel(ev);
+      if (this.viewport.onWheel(ev)) {
+        return this.cancel(ev);
+      }
     });
 
     on(el, 'touchstart', (ev: TouchEvent) => {
```
pro-src: The `touchstart` event is still always canceled.

Tyriar:
I just tested the PR on Ubuntu 18.04/Chrome and all the scroll events are bubbling up, even when the terminal is being scrolled

pro-src:
Thanks for testing. I'll pull in changes from master and push an update when it's working. I'm writing tests for this as well.

pro-src:
I'm kinda stuck... Any ideas as to what's causing the issue with the PR? The tests pass but it still doesn't work.

Tyriar:
@pro-src I don't think there's much testing in this area. Can you repro the issue I was seeing? Could be your browser if not?

pro-src:
> @pro-src I don't think there's much testing in this area. Can you repro the issue I was seeing? Could be your browser if not?

@Tyriar, I am able to reproduce that issue but the only *in-browser* testing I've done has been on unsupported mobile browsers. The *node* tests that I wrote are passing though. I don't know that I'll revisit this anytime soon.

Tyriar:
Closing for housekeeping purposes, the PR is archived if we need the diff in the future.

Issue #1465, opened 2018-05-21
cleanup and performance optimizations for `InputHandler.print`


jerch:
`InputHandler.print` was copied over from `InputHandler.addChar`. Since it gets now a string instead of a single char it can be further optimized. And cleaned up abit.

PR #1466, opened 2018-05-21
cleanup and performance optimizations for InputHandler.print

jerch:
Fixes: #1465

changes in InputHandler.print:
- call updateRange only at start and end
- make all repeated terminal accesses local
- set full charCode (codepoint) in terminal buffer

@Tyriar 
For the last change I tried to spot all usages of the charCode in the code, all I found was the atlas entry key, which is capped anyways. Therefore I think it is safe to change it. Please double check.

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -318,10 +318,10 @@ export class InputHandler implements IInputHandler {
     const wraparoundMode: boolean = this._terminal.wraparoundMode;
     const insertMode: boolean = this._terminal.insertMode;
     const curAttr: number = this._terminal.curAttr;
-    let bufferRow: [number, string, number, number][] = buffer.lines.get(buffer.y + buffer.ybase);
+    let bufferRow = buffer.lines.get(buffer.y + buffer.ybase);
 
     this._terminal.updateRange(buffer.y);
-    for (let stringPosition: number = start; stringPosition < end; ++stringPosition) {
+    for (let stringPosition = start; stringPosition < end; ++stringPosition) {
       char = data.charAt(stringPosition);
       code = data.charCodeAt(stringPosition);
 
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -321,7 +321,7 @@ export class InputHandler implements IInputHandler {
     let bufferRow: [number, string, number, number][] = buffer.lines.get(buffer.y + buffer.ybase);
 
     this._terminal.updateRange(buffer.y);
-    for (let stringPosition = start; stringPosition < end; ++stringPosition) {
+    for (let stringPosition: number = start; stringPosition < end; ++stringPosition) {
       char = data.charAt(stringPosition);
       code = data.charCodeAt(stringPosition);
 
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -307,25 +307,36 @@ export class InputHandler implements IInputHandler {
   }
 
   public print(data: string, start: number, end: number): void {
-    let ch;
+    let char;
     let code;
     let low;
     let chWidth;
     const buffer = this._terminal.buffer;
+    const charset = this._terminal.charset;
+    const screenReaderMode = this._terminal.options.screenReaderMode;
+    const cols = this._terminal.cols;
+    const wraparoundMode = this._terminal.wraparoundMode;
+    const insertMode = this._terminal.insertMode;
+    const curAttr = this._terminal.curAttr;
+    let bufferRow = buffer.lines.get(buffer.y + buffer.ybase);
+
+    this._terminal.updateRange(buffer.y);
     for (let i = start; i < end; ++i) {
-      ch = data.charAt(i);
+      char = data.charAt(i);
       code = data.charCodeAt(i);
+
+      // surrogate pair handling
       if (0xD800 <= code && code <= 0xDBFF) {
         // we got a surrogate high
         // get surrogate low (next 2 bytes)
         low = data.charCodeAt(i + 1);
         if (isNaN(low)) {
           // end of data stream, save surrogate high
-          this._surrogateHigh = ch;
+          this._surrogateHigh = char;
           continue;
         }
         code = ((code - 0xD800) * 0x400) + (low - 0xDC00) + 0x10000;
-        ch += data.charAt(i + 1);
+        char += data.charAt(i + 1);
       }
       // surrogate low - already handled above
       if (0xDC00 <= code && code <= 0xDFFF) {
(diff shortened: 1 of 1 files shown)


Tyriar:
> For the last change I tried to spot all usages of the charCode in the code, all I found was the atlas entry key, which is capped anyways. Therefore I think it is safe to change it. Please double check.

Do you mean removing `CHAR_DATA_CODE_INDEX` all together? You can find all the references to it by searching for `[3]` and `CHAR_DATA_CODE_INDEX`. As long as you measure and perf impact is minimal it can be removed.

src/InputHandler.ts
```diff
@@ -307,25 +307,36 @@ export class InputHandler implements IInputHandler {
...
 
   public print(data: string, start: number, end: number): void {
-    let ch;
+    let char;
     let code;
     let low;
     let chWidth;
     const buffer = this._terminal.buffer;
+    const charset = this._terminal.charset;
+    const screenReaderMode = this._terminal.options.screenReaderMode;
```
Tyriar: Isn't this only important for when the properties are accessed a multiple times? Was the thinking here to better leverage locality of reference or something?

src/InputHandler.ts
```diff
@@ -307,25 +307,36 @@ export class InputHandler implements IInputHandler {
   }
 
   public print(data: string, start: number, end: number): void {
-    let ch;
+    let char;
```
Tyriar: Let's type these while we're here.

src/InputHandler.ts
```diff
@@ -336,41 +347,46 @@ export class InputHandler implements IInputHandler {
...
-        ch = this._terminal.charset[ch];
+      // get charset replacement character
+      // FIXME: Should code be replaced as well?
+      if (charset) {
+        char = charset[char] || char;
       }
 
-      if (this._terminal.options.screenReaderMode) {
-        this._terminal.emit('a11y.char', ch);
+      // FIXME: Do we have to trigger this for every single char?
```
Tyriar: Yep, this is the hook for the AccessibilityManager to add the character. Perf is expected to degrade when you're on a screen reader (it's also not as important since it takes time to read the text out).

jerch:
@Tyriar 
Well the impact is quite big on the runtime (~30%). What do you mean by "it can be removed"? Remove it completely?

src/InputHandler.ts
```diff
@@ -307,25 +307,36 @@ export class InputHandler implements IInputHandler {
...
+    const charset = this._terminal.charset;
+    const screenReaderMode = this._terminal.options.screenReaderMode;
+    const cols = this._terminal.cols;
+    const wraparoundMode = this._terminal.wraparoundMode;
+    const insertMode = this._terminal.insertMode;
+    const curAttr = this._terminal.curAttr;
+    let bufferRow = buffer.lines.get(buffer.y + buffer.ybase);
+
+    this._terminal.updateRange(buffer.y);
     for (let i = start; i < end; ++i) {
```
Tyriar: Maybe using `x` would be better than `i`? It kind of looks like this is iterating over the rows when read in isolation.

src/InputHandler.ts
```diff
@@ -379,44 +395,50 @@ export class InputHandler implements IInputHandler {
           } else {
             // The line already exists (eg. the initial viewport), mark it as a
             // wrapped line
+            // FIXME: Why is it only buffer.y here?
```
Tyriar: Do you mean how the original wrapped line is not marked? That's just how it's represented in the data atm. I want this improved eventually, probably when I get to https://github.com/xtermjs/xterm.js/issues/622.

Tyriar:
> Well the impact is quite big on the runtime (~30%). What do you mean by "it can be removed"? Remove it completely?
> 
> Edit:
> Yup CHAR_DATA_CODE_INDEX is only used in TextRenderLayer.ts, [3] only in Terminal.test.ts, which passes.

Impact in which direction? I'm not totally sure what your plans are with `CHAR_DATA_CODE_INDEX`, you might need to restate it :sweat_smile:

src/InputHandler.ts
jerch: Well making them local gives a huge performance boost, since all of them get tested for every single char in the string over and over.
Btw for the screen reader thingy im not sure if this has to be tested for every single char.

src/InputHandler.ts
jerch: Lol right.

src/InputHandler.ts
jerch: Wow ok, this will really degrade performance.

src/InputHandler.ts
jerch: Ah no it is not the row, it is actually the position in the original parse string. Maybe better name it `stringPosition` to avoid confusion?

src/InputHandler.ts
jerch: Ok.

jerch:
Well the original code always set this to the first charCode in the cell by doing this:
```
..., ch.charCodeAt(0)]
```
My change simply sets this to the full codepoint value of the character, which is the same for 0-65535, but differs for the higher ones (like emoijs).

The change makes the method ~30% faster (since I am calculating the codepoint value anyways above and just reuse it here).

src/InputHandler.ts
Tyriar: Doh, they are used multiple times lol

src/InputHandler.ts
Tyriar: It's the column the char is being added to isn't it? The row/col is referred to at y/x elsewhere in the codebase

jerch: Yes the char gets added to buffer.x, but is read from the input string at `stringPosition`. This does not start at 0 because the parser does not copy the string, it just returns the string of `parse(string)` with start/end positions of the printable chars.

jerch:
Up for another review, lemme know if something is missing or unclear.

Tyriar:
Thanks again @jerch 🙂

src/InputHandler.ts
```diff
@@ -321,7 +321,7 @@ export class InputHandler implements IInputHandler {
     let bufferRow: [number, string, number, number][] = buffer.lines.get(buffer.y + buffer.ybase);
 
     this._terminal.updateRange(buffer.y);
-    for (let stringPosition = start; stringPosition < end; ++stringPosition) {
+    for (let stringPosition: number = start; stringPosition < end; ++stringPosition) {
```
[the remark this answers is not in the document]
jerch: Ah ok, thx for fixing it.

PR #1469, opened 2018-05-24
Suppress Safari promise rejections.

tmpfs:
Currently Webkit implements createImageBitmap() but returns a promise
rejection.

Note that once Webkit implements createImageBitmap() correctly these
changes should be removed.


Tyriar:
Thanks @tmpfs, this is perfect. Travis still passes so all good 😃

Issue #1467, opened 2018-05-23
revamp wcwidth


jerch:
`wcwidth` needs several enhancements:

1. interface to configure width of ambiguous chars at runtime, maybe locale dependent (see https://chromium.googlesource.com/native_client/nacl-newlib/+/master-backup/newlib/libc/string/wcwidth.c). Maybe include this into 2.
2. interface to load widths from some kind of outer source, see https://github.com/xtermjs/xterm.js/issues/942#issuecomment-388406591
This would make it possible to overload the default width set with some custom values. Those values could be based on backend or frontend or elseone's needs.
3. _Still in doubt_ - maybe merge `wcwidth` with future grapheme handling. Most of the wcwidth rules are a subset of the grapheme clustering. Once we have working grapheme handling merging them might save a lot of runtime. This is somewhat doubtful because grapheme handling might not work at an early stage with current cell based buffer model and should be added at a later stage (closer to the renderer). See also #1468.

jerch:
While playing around with the custom setting in #1470 which addresses 2. from the list above it became clear that the emulator should primarily follow the wcwidth behavior of the backend system (where the pty is running) to avoid all kinds of weird cursor and eol behavior. This is due intelligent backend libs like libreadline that anticipate the terminal behavior and are used in almost every terminal program nowadays.

It can be solved with another node package that collects the wcwidth output from the server side. For not nodejs based server components it might need a small C helper to get those data.

jerch:
Kinda offtopic here but related to recent issues - it is kinda hard to get wcwidth to respect the backend situation in a platform independent fashion because of the following:

1. newer wcwidth impls are locale dependent. Plain nodejs does a bad job regarding locale settings (cant handle it at all in default builds lol). This can only be fixed by some custom C++ binding, bummer.
2. Given 1# the next question is - which wcwidth should we link? Due to poor update policy some OSes have an outdated wcwidth in their clib (OSX), Windows has none at all. cmdline apps fix those awkward env situations by providing their own wcwidth (as we did in xterm.js) with additional heuristics to get something useful shown. While linking against the clib's wcwidth should work correctly under most Linux and BSD distros, OSX support might need additional maps per OS version and appXY. For Windows it might need a fix in winpty to get the cursor movement after a char insert, no clue yet how WSL does play into this.
[...]

PR #1470, opened 2018-05-24
Revamp wcwidth

jerch:
This PR addresses 1. and 2. of issue #1467.

No tests yet.

@Tyriar I made as little as possible changes to the existing wcwidth implementation. To still be able to use the closured lookup table I simply extended the old factory like design. Therefore it looks a bit messy now, I can refactor it to a more readable flat design if needed.

Edit:
I was not sure where to put the set options method and added it to `InputHandler` for now.

Edit2:
The options method understands 2 settings:
- **ambiguous**: Can be used to override the width of all ambiguous characters at once (much like in iterm2). I inspected several other wcwidth implementations - since most of them alter the width for ambiguous chars all together I did not split them into separate segments either. If this is needed the **custom** setting can be used instead.
- **custom**: Can be used to adjust the wcwidth for specific characters in the form `{charCode: width}`. This setting overrides any other width rule. Not applicable for ASCII chars due a performance optimization (should never be an issue).

The factory method understands 2 more settings `nul` and `control`. I did not expose them since it is not a good idea to change their behavior.


src/Types.ts
```diff
@@ -110,6 +110,7 @@ export interface ICompositionHelper {
 export interface IInputHandler {
   parse(data: string): void;
   print(data: string, start: number, end: number): void;
+  setWcwidthOptions(opts: {ambiguous?: 0 | 1 | 2, custom?: {[key: number]: 0 | 1 | 2}}): void;
```
Tyriar: Is the idea to expose this as a setting? What other terminals have you checked that also do this?

jerch: Well the **ambiguous** setting is comparable to the setting in iterm2, although they only offer to set full width on or off (here 2 or 1, we still support "default" by omitting this setting). It might be useful to expose this so people can hot patch the width of ambiguous chars if they encounter issues. I think VTE also supports this by an env variable. There is even a test file, that will show the difference (see attachment 00example.txt here https://github.com/mintty/mintty/issues/88#issuecomment-108887796).
The **custom** setting might be helpful internally, if the systems wcwidth does not agree with our implementation - we could simply overload it with the system setting to avoid cursor and line ending problems. I would not expose this the user since it might break more than it will help. Still we could offer an addon or something similar to do the nasty low level stuff.

Local terminals are normally not affected by this since they can use the systems wcwidth (and will have automatically the same settings). We cant do that in a browser component since we have no access to the C land. In VSCode (and similar "local" terminals) this could be achieved by some additional node package with access the system wcwidth (some C++ binding), in the demo it gets more tricky to load those data, maybe by some addon with a server addition.

bgw: While it might be useful to allow someone to pass a custom dictionary, in the case where we want to match with the system's `wcwidth`, it might be better if we can instead pass a function, because that matches the API POSIX gives us.

So, I'd recommend that the option be supplied as `custom: (key: number) => 0 | 1 | 2 | undefined`, where `undefined` causes fall-through behavior. If someone wants to use an object, then they can implement it as a function stub (e.g. `(key) => {123: 2}[key]`).

jerch: @bgw Yupp thats better, gonna change it.

jerch:
Closing this for now since it needs some more thinking about the right way to cope with the unicode mess behind, see https://github.com/xtermjs/xterm.js/issues/1467#issuecomment-397032254 for further details.

PR #1474, opened 2018-05-27
Use incremental tsc builds

### LINKED ISSUE #1473: Build has gotten really slow
I need to wait 5+ seconds between compiles which really interrupts my flow. I have a WIP to fix.

Tyriar:
The build has become increasingly slow as the project has grown, this is because we're using `gulp.watch` to relaunch a full `tsc` on the lib and the addons. The solution in the PR is split the watch task into a bunch of `tsc -w` calls for the lib and the addons as well as a `gulp watch-demo` which watching `lib/` and `demo/` and builds the `build/` and `./demo/dist` dirs as before. This reduces iteration time from 5-10 seconds to about 1 second.

I ended up moving the watch-demo task into `npm start` which is best run after `npm run watch` is. This separates terminals for building `src/` and everything to do with the demo.

Fixes #1473 

---

Initial build:

image

Compile error:

image

An iteration:

peek 2018-05-27 07-55

A server error (run `npm start` twice):

image


Tyriar:
@parisk not sure what the best way to fix up the `Dockerfile` is since the demo build now happens in `npm start`

gulpfile.js
```diff
@@ -31,58 +30,11 @@ if (path.normalize(outDir).indexOf(__dirname) !== 0) {
...
-    return tscAddon;
-  });
-
-  // Copy all addons from ${srcDir}/ to ${outDir}/
-  let copyAddons = gulp.src([
-    `${srcDir}/addons/**/**`
-  ]).pipe(gulp.dest(`${outDir}/addons`));
-
-  // Copy stylesheets from ${srcDir}/ to ${outDir}/
-  let copyStylesheets = gulp.src(`${srcDir}/**/*.css`).pipe(gulp.dest(outDir));
```
parisk: The removal of this is the only thing breaking the demo in the Docker build.

parisk: The removal of this is the only thing breaking the demo in the Docker image.

## Screenshot
image

Tyriar: Good catch, are you sure moving webpack into `npm start` didn't break it? I was worried about this:

https://github.com/xtermjs/xterm.js/blob/b43ee08dd259c705373f7181f152f6f3f4c235d0/Dockerfile#L9

Tyriar: Fixed

mofux:
Works for me. @Tyriar Is there something we can do to stop it from logging two empty lines with every meaningful line logged?

Tyriar:
@mofux I don't think so unfortunately, I looked into that briefly. The good news is you will typically only get compilation errors from a single project at a time.

mofux:
I'd say let's merge then 👍

PR #1477, opened 2018-05-30
Prefer const vs let for variables that are only declared once

lgeiger:
Requires that variable declarations use `const` instead of `let` and `var` if possible: https://palantir.github.io/tslint/rules/prefer-const/

This is mostly for code readability, performance wins are debatable.


Tyriar:
Love it! 😄

PR #1478, opened 2018-05-30
grapheme support

jerch:
WIP - dont merge.


bin/create-graphemedata.js
```diff
@@ -0,0 +1,124 @@
...
+
+            // write to ts file
+            let final = '';
+            final += `// FIRST: 0 <= codepoint < 12443\n`;
+            final += `export const FIRST: string = '${first}';\n`;
+            final += `// SECOND: 42606 <= codepoint < 65536\n`;
+            final += `export const SECOND: string = '${second}';\n`;
+            final += `// THIRD: codepoint >= 65536\n`;
+            final += `export const THIRD: string = '${third}';\n`;
+            require('fs').writeFileSync(path, final);
```
Tyriar: Do we expect the data file to ever change? If not I'd prefer to do a one time conversion and not include conversion scripts/data files in the repo

jerch: It changes at least with every major release and the consortium updates the spec in a yearly fashion. In 2 months we should see unicode v11.0.
It changes address mostly new chars and updates of drafted specs, it seems the older a char the less likely is an update.

I am not a fan of the code generation script thingy either. It may introduce several problems due to inexpected changes in the spec - like the region_indicator from 9.0 to 10.0 where they even gave up the previous-next rule, which will not work correctly until explicitly handled in some corresponding code. Maybe we should tag a supported unicode version and stick to it for some time?

jerch: Edit: The data generator could be parked in some dev module to avoid rewriting it once we decide to upgrade to a newer unicode version in future.

Tyriar: `xtermjs/node-grapheme-cluster-break`?

jerch:
Closing for now, needs a proper way of dealing with different unicode versions that has come come. See also #1470.

PR #1487, opened 2018-06-03
Enable noImplicitAny compiler option

Tyriar:


mofux:
Haven't tested, but LGTM 👍

PR #1490, opened 2018-06-03
Fix Demo String Options

nikonso:
Fix Demo String Options

Because the number options were bundled with the string options, this was happening 

image


Tyriar:
Thanks for the super quick fix 😄

PR #1495, opened 2018-06-06
Explicitly  mark Event Listeners as not passive

### LINKED ISSUE #829: Violation warning about touch and wheel event listeners not being passive
It seems that Chrome does not like that we do not use passive event listeners for touch and wheel events.

More information on passive event listeners:

- https://developers.google.com/web/updates/2016/06/passive-event-listeners
- https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md
- https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners

## Details
- Browser and browser version: Chrome 59
- OS version: macOS Sierra
- xterm.js version: 2.8.1

### Steps to reproduce

1. Open demo
2. Open console to view logs

### Screenshot
image

nikonso:
Fixes #829


Tyriar:
So glad this will finally be fixed 😄 will merge after CI passes;  just merged a big CI change and want to make sure it works.

PR #1503, opened 2018-06-08
Nullify and remove mouse handlers when mouse move it exited

### LINKED ISSUE #998: Mouse coordinates are printed under certain circumstances
When closing (or killing) a program with mouse integration while a mouse button is down will print the mouse escape sequences to the prompt. 

nvim

Reproduce:
- open `nvim`
- type `:q`
- hold down the left mouse button
- hit `enter`
- move the mouse and see the coordinates being printed

Tyriar:
Fixes #998

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -1039,19 +1039,38 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
         return this.cancel(ev);
       }
 
+      // TODO: Seems dangerous to attach event listeners to document, could they be done on element?
+      // TODO: All mouse handling should be pulled into its own file.
+
       // bind events
-      if (this.normalMouse) on(this._document, 'mousemove', sendMove);
+      let moveHandler: (event: MouseEvent) => void;
+      if (this.normalMouse) {
+        moveHandler = (event: MouseEvent) => {
+          // Do nothing if normal mouse mode is on. This can happen if the mouse is held down when the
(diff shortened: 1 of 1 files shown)


src/Terminal.ts
```diff
@@ -1002,19 +1002,38 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
         return this.cancel(ev);
       }
 
+      // TODO: Seems dangerous to attach event listeners to document, could they be done on element?
```
parisk: I have seen a few use cases for attaching `mousemove` listeners to `document` or `window`. Usually it is being used in order to avoid too much strictness when moving the mouse 1px outside the boundaries of the element accidentally.

bgw: I don't think it's actually dangerous, as long as you assume nobody has overridden `document.addEventListener()` (are we're generally assuming that nobody is mutating globals). There doesn't appear to be an API for getting a list of the event listeners on an element, aside from [this Chrome API that only works in the developer console](https://developers.google.com/web/tools/chrome-devtools/console/command-line-reference#geteventlistenersobject), so the effects of setting an event listener on the document should be opaque to our callers.

If you listen for mouseup on an element, you won't get the event if we're no longer hovering over the element, so I don't think that's an option, though there probably is some stuff you can do with detecting the mouse if it moves back over the element, and checking for the mouse state then.

Issue #1507, opened 2018-06-10
Improved Layering and enabling TS strict


Tyriar:
## Background

While in Prague, @mofux and I got together to discuss a better architecture for xterm.js that could lead to things like pulling all DOM access away from the core, moving parts of xterm.js into a web worker, and swapping out the frontend completely. Notes from this discussion were captured in https://github.com/xtermjs/xterm.js/issues/1506

## Proposal

I propose we split the codebase into distinct parts with very specific roles. This could be done all in the VS Code repo and still result in only a single `xterm` module being published, while also not obstructing a future where we may want to allow swapping parts out.

Here is the breakdown:

- `src/` - Contains all source code as today
  - `base/` - Contains interfaces for core and frontend to communicate with each other, as well as code that is used between both of them
  - `core/` - This runs in both node.js and browser runtimes. This means far better and easier testing since everything is just node.js without worrying about the DOM
  - `ui/` - This code runs only in a browser, anything in xterm.js that interacts with the DOM today will move to here. That includes things like mouse wheel handling (and translating), composition helper, textarea, etc.
  - `public/` - This provides an implementation of the current day xterm.js API.

Dependencies between this folders are strictly enforced:

```
|------------------|
|       base       |
|------------------|
    ↑       ↑    ↑
|------| |----|  |
| core | | ui |  |
|------| |----|  |
    ↑       ↑    |
|------------------|
|      public      |
|------------------|
```

Notice that `ui` does not depend on `core`, any dependencies between `core` and `ui` are pulled into interfaces and placed into `base`.

## Example file structure

This tree will give you an idea of what lives where:

```
src/
  base/
    Types.ts (maybe this should be .d.ts?)
    Clone.ts
  core/
    Types.ts
    Core.ts
    components/
      Buffer.ts
      BufferSet.ts
      Parser.ts
  ui/
    Types.ts
    UserInterface.ts
    components/
      Accessibility.ts
      Composition.ts
  public/
    xterm.d.ts
    Terminal.ts
```

## Composite projects

In TypeScript 3 we will get the ability to build parts of projects independently and depend on them. This would lead to improved build times but also we get the layering enforcement for free as the `tsconfig.json` for `core` for example will not contain the `dom` APIs, so any use of them will fail compilation.

## Absolute imports

We may be able to use absolute imports when these changes happen, whenever I try to do it I run into issues with importing a file from the same folder as an absolute import. See https://github.com/xtermjs/xterm.js/issues/1314

## Plan

[...]

PR #1508, opened 2018-06-10
Start of improved layering

Tyriar:
Curious what people think of this change, I haven't gone and done a polish pass on this yet but the basic structure I think we should start with when considering #1507 is there.

- Created `IKeyboardEvent` in `common`
- Moved key event to escape sequence translation into `core/input/Keyboard.ts` and added some tests
- Introduced `public` which is the new API. This will allow us to make changes to `./src/Terminal` and eventually move it into `core`. This will break any integrations that access private API (which we don't guarantee).


mofux:
Can we safely merge this without breaking compatibility of the 3.x releases, or should this be targeted towards a 4.0 release? I'm a bit afraid the changes in this PR become stale if this hangs around for too long. I think as long as we maintain compatibility with the public 3.x api it should be okay to merge, no?

Tyriar:
I think this maintains the public API fine. It will break VS Code for example but that's because [`buffer` is touched](https://github.com/Microsoft/vscode/blob/bd5ee49f8899676bce2eabcae7dd781299973eef/src/typings/vscode-xterm.d.ts#L674).

I'll polish it up now as merging this will really help with pulling components into their correct spots.

Tyriar:
Ok I think it's ready for proper review(s)

mofux:
Still a long way to go, but it's a good start 😅Tested locally and works fine 👍

Tyriar:
Works in VS Code so merging in

PR #1513, opened 2018-06-12
Ensure drawBoldTextInBrightColors redraws screen

Tyriar:


bgw:
We spend a lot of effort trying to do the least amount of operations every time an option changes, however, options probably don't change frequently. I'm wondering how bad it would be for us to just redraw, remeasure, etc, every time *any* option changes.

Tyriar:
@bgw it's probably a good idea to just always do it. It might also be a good idea to allow setting an `ITerminalOptions` so multiple can be changed at once.

PR #1517, opened 2018-06-18
Sort asending for companies in readme and add Nutanix

JonKohler:
Sorted the company list in the readme asending and added Nutanix to the list of users


Tyriar:
The list is actually ascending based on the time they were added to the list intentionally, we can add yours to the end if you want?

JonKohler:
@Tyriar - I've pushed an updated patch that slims down LOC changes to just 2, adding our name and also adding a small note to the bottom of the list to instruct future PR's

Tyriar:
Thanks @JonKohler 😃

Issue #1521, opened 2018-06-20
'restore cursor' restoring position but not text attributes


7PH:
I had some issues working with some VT100 escape sequences:
- save cursor (ESC 7) (DECSC)
- restore cursor (ESC 8) (DECRC)

Usually, a [VT100 terminal should save the cursor position but also characters attributes when writing a save cursor escape sequence](https://vt100.net/docs/vt510-rm/DECSC.html)
The characters attributes should be restored when writing a restore cursor.

[Currently, the position of the cursor is saved and restored **but not the characters attributes**.](https://github.com/xtermjs/xterm.js/blob/master/src/InputHandler.ts#L1822)

I've tested the behavior of the save/restore escape sequences on other VT100 implementations to ensure they were handling it:
- gnome-terminal
- terminator
- [hterm](https://chromium.googlesource.com/apps/libapps/+/master/hterm)

I'll give some details below, and will start working on a solution since I need it for myself. 
Keep up the great work!

## Details
- Browser and browser version: **Chromium: Version 62.0.3202.89 (Developer Build) built on Debian buster/sid, running on Debian buster/sid (64-bit)**
- OS version: **debian buster/sid**
- xterm.js version: **3.4.1**

### Steps to reproduce

Here's the VT100 escape sequences used for testing:
```text
\x1b[0m (attributes off)
\x1b[1;1f (reset origin)
\x1b[2J (clear)
\x1b7 (save cursor state)
\x1b[7m (set negative color)
                        <colored> (just some output)
\x1b8 (restore cursor state)
<should not be colored>\n\r (another output - will be printed in negative color in xtermjs)
```

With xtermjs, "should not be colored" is colored
```typescript
xterm.write('\x1b[0m\x1b[1;1f\x1b[2J\x1b7\x1b[7m                        <colored>\x1b8<should not be colored>\n\r');
// will print:
// "<should not be colored> <colored>"
```

In a bash, "should not be colored" won't be colored
```bash
printf '\x1b[0m\x1b[1;1f\x1b[2J\x1b7\x1b[7m                        <colored>\x1b8<should not be colored>\n\r'
# will print
# "<should not be colored> <colored>"
```
### How to fix

Seems to me that text attributes should be saved and restored along with the position, [here](https://github.com/xtermjs/xterm.js/blob/master/src/InputHandler.ts#L1822).
[...]

PR #1522, opened 2018-06-20
#1521: VT100 escape codes DECSC and DECRC now handle character attributes

7PH:
**Following [issue #1521](https://github.com/xtermjs/xterm.js/issues/1521)**

Here is the list of the modification I've made:
- added an attribute 'savedCurAttr' used to save  the 'curAttr' attribute of 'Terminal' and 'MockInputHandlingTerminal' 
- edited the saveCursor and restoreCursor of the InputHandler class to effectively store and restore the characters attributes
- added unit tests (and checked they passed)

I tried to stay as close as possible to your code style.

Please let me know if there is anything that requires action on my side.


7PH:
Hmm, i don't get why the **Windows Job / Initialize phase** is failing, I doubt my changes made that fail.

Let me know if there is anything I can do to help.

Tyriar:
> Hmm, i don't get why the Windows Job / Initialize phase is failing

Yeah this wouldn't be related to your changes, it will probably pass next time.

src/InputHandler.ts
```diff
@@ -1834,6 +1835,7 @@ export class InputHandler implements IInputHandler {
   public restoreCursor(params: number[]): void {
     this._terminal.buffer.x = this._terminal.buffer.savedX || 0;
     this._terminal.buffer.y = this._terminal.buffer.savedY || 0;
+    this._terminal.curAttr = this._terminal.savedCurAttr || 0;
```
Tyriar: Should this be `|| DEFAULT_ATTR`?

Tyriar:
Thanks a lot @7PH!

PR #1525, opened 2018-06-21
Ensure all references are released when Terminal.dispose is called

### LINKED ISSUE #1518: Prevent memory leaks when Terminal.dispose is called and make it easy to test in the demo
It's hard to test for memory leaks in the demo because so many things have a reference to the Terminal object, including the Ui and addons.

Also it looks DOM listeners on document after `Terminal.dispose` is triggered are keeping the `Terminal` object in memory.

Tyriar:
This PR adds the following:

- A new abstract class `Disposable` that is extended by a number of classes and provides a `register` function to register disposables and a `dispose` function which disposes them
- Moves `util/Dom.ts` to `ui/Lifecycle.ts`
- Registered disposables for all DOM event and EventEmitter listeners (excluding some mouse ones which need some more work)
- Added a dispose terminal button to the bottom of the demo for easier testing of leaked references

After these changes, when a heap snapshot is taken it will no longer show `Terminal` in the heap. The following was taken after filling the buffer and disposing the terminal:

screen shot 2018-06-21 at 1 32 55 pm

This is what it looked like before (with demo and internal Terminal references holding onto the `Terminal`):

screen shot 2018-06-21 at 1 33 57 pm

Fixes #1518


src/AccessibilityManager.ts
```diff
@@ -72,29 +71,28 @@ export class AccessibilityManager implements IDisposable {
 
     this._terminal.element.insertAdjacentElement('afterbegin', this._accessibilityTreeRoot);
 
-    this._disposables.push(this._renderRowsDebouncer);
-    this._disposables.push(this._terminal.addDisposableListener('resize', data => this._onResize(data.cols, data.rows)));
-    this._disposables.push(this._terminal.addDisposableListener('refresh', data => this._refreshRows(data.start, data.end)));
-    this._disposables.push(this._terminal.addDisposableListener('scroll', data => this._refreshRows()));
+    this.register(this._renderRowsDebouncer);
```
parisk: Just an idea; would it make sense to replace `this.register(...` with [TypeScript decorators](http://www.typescriptlang.org/docs/handbook/decorators.html)? This way we could keep _disposable_ properties declared outside the constructor.

Tyriar: You mean something like this? (if it works)

```ts
function register(target: any, key: string) {
	let _val = this[key];
	
    const getter = () => _val;
    const setter = (newVal) => {
		if (_val) {
			_val.dispose();
		}
		target.register(newVal);
		_val = newVal;
    };
    Object.defineProperty(target, key, {
        get: getter,
        set: setter,
        enumerable: true,
        configurable: true 
    });
}
```

src/common/Lifecycle.ts
```diff
@@ -0,0 +1,33 @@
...
+  public dispose(): void {
+    this._disposables.forEach(d => d.dispose());
+    this._disposables.length = 0;
+  }
+
+  /**
+   * Registers a disposable object.
+   * @param d The disposable to register.
+   */
+  public register<T extends IDisposable>(d: T): void {
```
mofux: I think `registerDisposable` or `addDisposable` would be better, especially because classes that inherit from Disposable might want to use `register` with a totally different meaning. Say we have an `AddonRegistry` class that extends `Disposable`, it would certainly get very confusing what `AddonRegistry.register` would do 😅

mofux: What do you think about `register` taking a list of arguments `...disposables`, so we could simply register multiple disposables at once, e.g.:
```javascript
this._terminal.core.register(
  this._terminal.addDisposableListener('blur', () => this._clearLiveRegion()),
  this._terminal.addDisposableListener('linefeed', () => this._lineFeed()),
  this._terminal.addDisposableListener('focus', () => this._focus())
)
```

src/AccessibilityManager.ts
Tyriar: I looked into this and I don't think we can do it. For starters decorators are experimental, I switched it on and tried to get it working but it seemed to be acting not as I expected. Maybe we can do this in the future?

demo/main.js
```diff
@@ -139,11 +148,11 @@ function runFakeTerminal() {
     } else if (printable) {
       term.write(key);
     }
-  });
+  }));
 
-  term.on('paste', function (data, ev) {
+  term._core,register(term.addDisposableListener('paste', function (data, ev) {
```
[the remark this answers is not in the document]
Tyriar: Fixed

src/common/Lifecycle.ts
mofux: Fine with me, your call 👍

mofux:
@Tyriar Should I give this a final test / review?

Tyriar:
@mofux that would be good, also I think tests are passing it's just that https://github.com/xtermjs/xterm.js/pull/1531 broke everything

src/Terminal.ts
```diff
@@ -1002,30 +1002,37 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
...
       // bind events
-      if (this.normalMouse) on(this._document, 'mousemove', sendMove);
+      if (this.normalMouse) {
+        this._document.addEventListener('mousemove', sendMove);
+      }
 
       // x10 compatibility mode can't send button releases
       if (!this.x10Mouse) {
         const handler = (ev: MouseEvent) => {
           sendButton(ev);
```
Tyriar: test comment, ignore me

Tyriar: test comment, ignore me

src/Terminal.ts
```diff
@@ -1002,30 +1002,37 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
         return this.cancel(ev);
       }
 
+      // TODO: these event listeners should be managed by the disposable, the Terminal reference may
+      // be kept aroud if Terminal.dispose is fired when the mouse is down
+
       // bind events
-      if (this.normalMouse) on(this._document, 'mousemove', sendMove);
+      if (this.normalMouse) {
+        this._document.addEventListener('mousemove', sendMove);
```
Tyriar: test

Tyriar:
@mofux I think this is ready for final review, CI failure was unrelated.

mofux:
Works like a charm, ready to merge from my side 👍

Tyriar:
:tada:

Issue #791, opened 2017-07-13
Buffer performance improvements


Tyriar:
## Problem

**Memory**

Right now our buffer is taking up too much memory, particularly for an application that launches multiple terminals with large scrollbacks set. For example, the demo using a 160x24 terminal with 5000 scrollback filled takes around 34mb memory (see https://github.com/Microsoft/vscode/issues/29840#issuecomment-314539964), remember that's just a single terminal and 1080p monitors would likely use wider terminals. Also, in order to support truecolor (https://github.com/sourcelair/xterm.js/issues/484), each character will need to store 2 additional `number` types which will almost double the current memory consumption of the buffer.

**Slow fetching of a row's text**

There is the other problem of needing to fetch the actual text of a line swiftly. The reason this is slow is due to the way that the data is laid out; a line contains an array of characters, each having a single character string. So we will construct the string and then it will be up for garbage collection immediately afterwards. Previously we didn't need to do this at all because the text is pulled from the line buffer (in order) and rendered to the DOM. However, this is becoming an increasingly useful thing to do though as we improve xterm.js further, features like the selection and links both pull this data. Again using the 160x24/5000 scrollback example, it takes 30-60ms to copy the entire buffer on a Mid-2014 Macbook Pro.

**Supporting the future**

Another potential problem in the future is when we look at introducing some view model which may need to duplicate some or all of the data in the buffer, this sort of thing will be needed to implement reflow (https://github.com/sourcelair/xterm.js/issues/622) properly (https://github.com/sourcelair/xterm.js/pull/644#issuecomment-298058556) and *maybe* also needed to properly support screen readers (https://github.com/sourcelair/xterm.js/issues/731). It would certainly be good to have some wiggle room when it comes to memory.

This discussion started in https://github.com/sourcelair/xterm.js/issues/484, this goes into more detail and proposes some additional solution.

I'm leaning towards solution 3 and moving towards solution 5 if there is time and it shows a marked improvement. Would love any feedback! /cc @jerch, @mofux, @rauchg, @parisk 

## 1. Simple solution

This is basically what we're doing now, just with truecolor fg and bg added.

```ts
// [0]: charIndex
// [1]: width
// [2]: attributes
// [3]: truecolor bg
// [4]: truecolor fg
type CharData = [string, number, number, number, number];

type LineData = CharData[];
```

**Pros**

- Very simple

**Cons**

- Too much memory consumed, would nearly double our current memory usage which is already too high.

[...]

PR #1528, opened 2018-06-24
Testbed: text attribute atlas & buffer redesign thoughts

jerch:
This is just a testbed, do not merge. It adds some thoughts and implementation ideas to the discussion from this issue #791.

The PR illustrates the usage of a ref counting text attribute atlas. The code works, but is not well integrated in the codebase, I simply extended the obvious code parts with the needed ref handling and slot indirection. Also not all refs are released atm, this would need a tight integration with `Buffer` and `CircularList` to catch all derefs. And the atlas will need some index lookup magic to avoid searching in O(n) through existing slot entries (though not an issue in the tests, taken slots were always < 64).

So what is this good for? It shows a way to save memory by using indexed SGR attributes instead of holding them at every single cell, which will come handy once we support true color. It might also be a small step into a new buffer layout. Real memory usage is a tough topic with JS due to the dynamic way the engines act, I try a rough estimation:

- memory usage with current buffer layout with true color per cell
  ```
  CharData = [
    number (attr),
    string (content),
    number (wcwidth),
    number (charCode),
    number (foreground),  // needed for true color
    number (background)   // needed for true color
  ];
  ```
  this is roughly: 4 + 8 + 4 + 4 + 4 + 4 = 28 bytes, for 10000 lines à 100 cells it is ~28 MB of data, real usage is much bigger due to the JS objects and the way the engine decides how to store the primitives in the JS array (can be pointered, which will add 8 bytes per data point --> ~76 MB).
- with text attribute atlas per cell
  ```
  CharData = [
    number (index to attr slot),
    string (content),
    number (wcwidth),
    number (charCode)
  ];
  AtlasSlot = [
    uint32 (ref counter),
    uint32 (attr),
    uint32 (foreground),
    uint32 (background)
  ];
  ```
  which is 4 + 8 + 4 + 4 = 20 bytes per cell + the atlas size, for 10000 lines à 100 cells it is ~20 MB of data (or ~52 MB pointered) plus the atlas size. Atlas size is hard to determine, for my benchmark with `ls -lR /usr/lib` it was only ~800 bytes, worst case is 10000 * 100 * 4 * 4 = 16 MB, which is slightly worse than the direct approach (expected since an atlas cant play its advantages if all cells differ). Runtime is not much affected, the atlas indirections, lookups and management takes less than 0.1% runtime.

What do the calculations/tests show?
- we have no control/clue about the real mem usage due to the dynamics of JS objects
- adding data entries to the cell buffer directly is a bad idea, it adds much more than the actual data size, total mem consumption grows linearly on a bigger scale
- an attribute atlas can help to keep this low by reusing them
- almost no runtime penalty for the indirections

Imho the points have implications on a possible buffer redesign. The buffer contains two orthogonal data parts atm, that are changed independently:
- content related: the actual string, wcwidth and the charCode for the renderer atlas, set in `InputHandler.print`
- SGR attributes: only the flags number atm (to be extended with true color foreground and background), changed in `InputHandler.charAttributes` terminal wide and applied to a cell at different occations (most in `InputHandler.print` as well)

To really save memory at this point I would propose a step by step redesign of the cell buffer:
1. build a separate ref counting cell content storage to hold string data, wcwidth and charCode, maybe TypedArray based to avoid JS objects overhead
2. refine the attribute atlas with better integration
3. change `CharData` to `[number (attr pointer), number (content pointer)]`
4. change `CircularList` and `Buffer` to operate on a preallocated TypedArray

Also the steps 1. and 4. are likely to give a better runtime performance by avoiding unnecessary object creations and GC runs (GC is at 20% runtime atm).

### DIFF
--- a/src/AttributeAtlas.ts
+++ b/src/AttributeAtlas.ts
@@ -1,0 +1,92 @@
+interface ITextAttributes {
+  flags: number;
+  foreground: number;
+  background: number;
+}
+
+export const enum AtlasEntry {
+  FLAGS = 1,
+  FOREGROUND = 2,
+  BACKGROUND = 3
+}
+
+export class TextAttributeAtlas {
+  /** data storage */
+  public data: Uint32Array;
+  constructor(size: number) {
+    this.data = new Uint32Array(size * 4);
+  }
+  private _setData(idx: number, attributes: ITextAttributes): void {
+    this.data[idx] = 0;
+    this.data[idx + AtlasEntry.FLAGS] = attributes.flags;
+    this.data[idx + AtlasEntry.FOREGROUND] = attributes.foreground;
+    this.data[idx + AtlasEntry.BACKGROUND] = attributes.background;
+  }
+
+  /**
+   * convenient method to inspect attributes at slot `idx`.
+   * For better performance atlas idx and AtlasEntry
+   * should be used directly to avoid number conversions.
+   * @param {number} idx
+   * @return {ITextAttributes}
+   */
+  getAttributes(idx: number): ITextAttributes {
+    return {
+      flags: this.data[idx + AtlasEntry.FLAGS],
+      foreground: this.data[idx + AtlasEntry.FOREGROUND],
+      background: this.data[idx + AtlasEntry.BACKGROUND]
+    };
+  }
+
+  /**
+   * Returns a slot index in the atlas for the given text attributes.
+   * To be called upon attributes changes, e.g. by SGR.
+   * NOTE: The ref counter is set to 0 for a new slot index, thus
+   * values will get overwritten if not referenced in between.
+   * @param {ITextAttributes} attributes
+   * @return {number}
+   */
+  getSlot(attributes: ITextAttributes): number {
+    // find equal slot or free
+    for (let i = 0; i < this.data.length; i += 4) {
+      if (!this.data[i]) {
+        this._setData(i, attributes);
+        return i;
+      }
+      if (
+          this.data[i + AtlasEntry.FLAGS] === attributes.flags
+          && this.data[i + AtlasEntry.FOREGROUND] === attributes.foreground
+          && this.data[i + AtlasEntry.BACKGROUND] === attributes.background
+      ) {
+          return i;
+      }
+    }
+    // could not find a valid slot --> resize storage
+    const data = new Uint32Array(this.data.length * 2);
+    for (let i = 0; i < this.data.length; ++i) data[i] = this.data[i];
+    const idx = this.data.length;
+    this.data = data;
+    this._setData(idx, attributes);
+    return idx;
+  }
+
+  /**
+   * Increment ref counter.
+   * To be called for every terminal cell, that holds `idx` as text attributes.
+   * @param {number} idx
+   */
+  ref(idx: number): void {
+    this.data[idx]++;
+  }
+
+  /**
+   * Decrement ref counter. Once dropped to 0 the slot will be reused.
+   * To be called for every cell that gets removed or reused with another value.
+   * @param {number} idx
+   */
+  unref(idx: number): void {
+    this.data[idx]--;
+    if (this.data[idx] < 0)
+    this.data[idx] = 0;
+  }
+}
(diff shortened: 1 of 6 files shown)


mofux:
@jerch Thanks for this prototype, I really like your approach! Sharing the Attribute Altas between multiple terminal instances could safe even more memory, and I think it is totally doable with your current design. A potential Content Atlas could be a bit trickier to share if wcwidth differs between terminal instances though. Have you thought about sharing the atlas yet?

jerch:
Played abit with the buffer and removed the wcwidth entry by directly calling it where needed - this removes 0.5 MB from terminal object for the demo with 1000 lines scrollback with my default benchmark `ls -lR /usr/lib`. The runtime does not change at all - weird, thought it would suffer. Maybe chrome inlines the function, the function itself does only a table lookup and is not expensive.

Maybe we dont need wcwidth in the buffer at all?

jerch:
Done with the tests, closing the PR.

src/AttributeAtlas.ts
```diff
@@ -0,0 +1,92 @@
...
+   * Returns a slot index in the atlas for the given text attributes.
+   * To be called upon attributes changes, e.g. by SGR.
+   * NOTE: The ref counter is set to 0 for a new slot index, thus
+   * values will get overwritten if not referenced in between.
+   * @param {ITextAttributes} attributes
+   * @return {number}
+   */
+  getSlot(attributes: ITextAttributes): number {
+    // find equal slot or free
+    for (let i = 0; i < this.data.length; i += 4) {
```
Tyriar: Is the idea/hope here meant to be that there aren't that many unique attributes within a terminal? This could cause issues in programs that use a lot of attributes as it's called on every single printed character.

src/Buffer.ts
```diff
@@ -117,10 +117,11 @@ export class Buffer implements IBuffer {
     if (this.lines.length > 0) {
       // Deal with columns increasing (we don't do anything when columns reduce)
       if (this._terminal.cols < newCols) {
-        const ch: CharData = [DEFAULT_ATTR, ' ', 1, 32]; // does xterm use the default attr?
+        const ch: CharData = [this._terminal.attributeAtlas.getSlot({flags: DEFAULT_ATTR, foreground: 0, background: 0}), ' ', 1, 32]; // does xterm use the default attr?
```
Tyriar: It might be worth handling the default attribute specially, storing it in `data[0]` (or a const) and exposing some property to fetch it directly.

src/AttributeAtlas.ts
jerch: Ah yes - guess you found the catch lol. It will not be called on every char to print, but on every attribute change in between. This is far better (less likely to happen), still the runtime with O(n) worst case for n attribute entries sucks. It definitely will need some smart lookup code to speed it up. The attribute lookup for a char to be printed is a simple index access.

src/Buffer.ts
jerch: Yupp already thought about that, problem is - the default attrs could change while the old are still in the buffer, therefore it cant be loaded in a specific place easily. To optimize things here a second indirection might be handy to be able to reorganize the cache like in a real caching szenario, maybe be ref counts or something like that. Classical caching strategies wont apply here imho, since we have to ensure that all indices are still legal.

src/AttributeAtlas.ts
Tyriar: Instead of storing idx could we store a reference to the ITextAttributes in CharData? That would allow indexing a BST by say flags, then fg, then bg and binary search for that.

I guess that has issues with V8 deopts because it's then a [number, ITextAttributes]?

src/Buffer.ts
Tyriar: So have some `idxToActualIndex: {[idx: number]: number}` map to allow fiddling with the order of items?

src/AttributeAtlas.ts
jerch: Yes v8 is very picky about mixed array elements :cry:

src/Buffer.ts
jerch: Yup a pointer pointer, it gets more ugly but will make resorting/optimizing the store data structure possible. Like in C with **p.

src/AttributeAtlas.ts
jerch: Yeah I think a BST is the way to go, no clue yet, how to layout it correctly, somehow the flags-->fg-->bg thing makes me think of some radixsort like thingy or multiple BSTs. Lets do a fibonacci heap (just kidding). Maybe a red-black tree will do (as of runtime/memory tradeoff).

PR #1531, opened 2018-06-25
Pin node-pty to 0.7.6

Tyriar:
I broke the build when I released node-pty@0.7.5 :sweat_smile:


Tyriar:
Merging as this breaks the build and is failing PR builds

PR #1538, opened 2018-06-28
Support for column/box selection

### LINKED ISSUE #1300: Support for column/box selection
## Details

- OS version: macOS 10.13.2

### Steps to reproduce

In _Terminal_ app I could press _Alt_ key and make a vertical selection, similar in iTerm. E.g.

screen shot 2018-02-22 at 16 01 22

I expect similar behaviour for _xterm.js_.

Related: zeit/hyper/issues/2697

vsinha:
closes #1300


vsinha:
The final test in `SelectionManager.test.ts` is failing, would love some guidance on how to set that up. I left a comment in there which explains that the behavior in browser actually works as intended, but the test does not reflect that

src/SelectionManager.ts
```diff
@@ -80,7 +81,12 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
   /**
    * The current selection mode.
    */
-  private _activeSelectionMode: SelectionMode;
+  protected _activeSelectionMode: SelectionMode;
+
+  /**
+   * The modifier keys required to trigger block select mode with left click + drag
```
Tyriar: block -> column

src/SelectionManager.ts
```diff
@@ -507,6 +533,40 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
...
+    if (this._terminal.browser.isMac) {
+      return ['altKey'];
+    }
+
+    // Linux and Windows
+    return ['shiftKey'];
+  }
+
+  /**
+   * Begin a block selection
```
Tyriar: block -> column

src/SelectionManager.ts
```diff
@@ -507,6 +533,40 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
...
+   */
+  private _onColumnSelectSingleClick(event: MouseEvent): void {
+    this._onSingleClick(event); // Perform all the normal setup actions
+    this._activeSelectionMode = SelectionMode.COLUMN;
+  }
+
+  /**
+   * Checks if all required key modifiers are pressed in order to enable block
```
Tyriar: block -> column

src/xterm.css
```diff
@@ -139,6 +139,11 @@
     cursor: pointer;
 }
 
+.xterm.xterm-cursor-crosshair {
+    /* Block selection mode */
```
Tyriar: block -> column

src/SelectionManager.ts
```diff
@@ -182,30 +189,41 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
...
-      const bufferLine = this._buffer.lines.get(i);
-      const lineText = this._buffer.translateBufferLineToString(i, true);
-      if ((<any>bufferLine).isWrapped) {
-        result[result.length - 1] += lineText;
-      } else {
-        result.push(lineText);
+
+    if (this._activeSelectionMode === SelectionMode.COLUMN) {
+      // Ignore zero width selections
+      if (start[0] !== end[0]) {
```
Tyriar: We can exit early here and return `''`

src/SelectionManager.ts
```diff
@@ -254,7 +272,7 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
    */
   private _refresh(): void {
     this._refreshAnimationFrame = null;
-    this.emit('refresh', { start: this._model.finalSelectionStart, end: this._model.finalSelectionEnd });
+    this.emit('refresh', { start: this._model.finalSelectionStart, end: this._model.finalSelectionEnd, columnSelectMode: this._activeSelectionMode === SelectionMode.COLUMN });
```
Tyriar: Let's break this into a separate line for each property now since it's getting a bit long.

src/SelectionManager.ts
```diff
@@ -507,6 +533,40 @@ export class SelectionManager extends EventEmitter implements ISelectionManager
...
+    this._onSingleClick(event); // Perform all the normal setup actions
+    this._activeSelectionMode = SelectionMode.COLUMN;
+  }
+
+  /**
+   * Checks if all required key modifiers are pressed in order to enable block
+   * select mode
+   * @param event the mouse click event
+   */
+  public isColumnSelectMode(event: KeyboardEvent | MouseEvent): boolean {
```
Tyriar: I think this could be simplified by following the pattern of `shouldForceSelection` and not adding the code that enables customization. That will also let us get rid of `<any>`.

Also let's include `macOptionClickForcesSelection` in the condition just to make our intent explicit.

vsinha:
@Tyriar Updated :)

Issue #1551, opened 2018-07-06
avoid deoptimizations in v8


jerch:
There are several code parts that trigger alot of deoptimizations in v8. Run
`chrome --js-flags="--trace-opt --trace-deopt --trace"`
to get the trace info. For my typical benchmark with `ls -lR /usr/lib` the demo shows the following:

```
[deoptimizing (DEOPT eager): begin 0x27f83bd2d3b9 <JSFunction TextRenderLayer._forEachCell (sfi = 0x27f83bd1bf49)> (opt #2) @3, FP to SP delta: 96, caller sp: 0x7ffc6b031a28]
            ;;; deoptimize at <http://localhost:3000/dist/bundle.js:7140:36>, wrong map

[deoptimizing (DEOPT soft): begin 0x27f83bd2bea9 <JSFunction StaticCharAtlas.draw (sfi = 0x27f83bd2b189)> (opt #1) @19, FP to SP delta: 56, caller sp: 0x7ffc6b031748]
            ;;; deoptimize at <http://localhost:3000/dist/bundle.js:7740:36>, Insufficient type feedback for generic named access

[deoptimizing (DEOPT soft): begin 0x27f83bd19341 <JSFunction InputHandler.print (sfi = 0x27f83bd04fd1)> (opt #16) @53, FP to SP delta: 144, caller sp: 0x7ffc6b031dd0]
            ;;; deoptimize at <http://localhost:3000/dist/bundle.js:1999:30>, Insufficient type feedback for generic named access

[deoptimizing (DEOPT eager): begin 0x27f83bd19341 <JSFunction InputHandler.print (sfi = 0x27f83bd04fd1)> (opt #19) @21, FP to SP delta: 144, caller sp: 0x7ffc6b031e90]
            ;;; deoptimize at <http://localhost:3000/dist/bundle.js:2027:35>, wrong map

[deoptimizing (DEOPT lazy): begin 0x27f83bd4b819 <JSFunction Terminal.blankLine (sfi = 0x31d068a7c271)> (opt #28) @18, FP to SP delta: 72, caller sp: 0x7ffc6b031d00]

[deoptimizing (DEOPT eager): begin 0x27f83bd2d3b9 <JSFunction TextRenderLayer._forEachCell (sfi = 0x27f83bd1bf49)> (opt #7) @27, FP to SP delta: 96, caller sp: 0x7ffc6b031a28]
[...]

PR #1555, opened 2018-07-07
fix out of bound deopt in EscapeSequenceParser.parse

jerch:
Fixes the "out of bound" deopt of  EscapeSequenceParser.parse (see #1551).

### DIFF
--- a/src/EscapeSequenceParser.ts
+++ b/src/EscapeSequenceParser.ts
@@ -375,8 +375,7 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
       // shortcut for most chars (print action)
       if (currentState === ParserState.GROUND && code > 0x1f && code < 0x80) {
         print = (~print) ? print : i;
-        do code = data.charCodeAt(++i);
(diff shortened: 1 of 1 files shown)


src/EscapeSequenceParser.ts
```diff
@@ -375,8 +375,7 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
       // shortcut for most chars (print action)
       if (currentState === ParserState.GROUND && code > 0x1f && code < 0x80) {
         print = (~print) ? print : i;
-        do code = data.charCodeAt(++i);
-        while (i < l && code > 0x1f && code < 0x80);
+        while (++i < l && data.charCodeAt(i) > 0x1f && data.charCodeAt(i) < 0x80) {}
```
Tyriar: Let's change this back into a do while with `i++` in the middle, I think it's easier for most people (including myself) to read. I think this is equivalent:

```ts
do {
  i++;
} while (i < l && data.charCodeAt(i) > 0x1f && data.charCodeAt(i) < 0x80);
```

jerch:
Ok changed it.

PR #1563, opened 2018-07-11
Use yarn

### LINKED ISSUE #1562: Switch to yarn
It gives us better dependency management https://stackoverflow.com/questions/44897368/what-is-the-difference-between-yarn-lock-and-npms-package-lock/46352264#46352264

Tyriar:
Fixes #1562

The reset of the needed changes are in the wiki


mofux:
Maybe it's a stupid noobish question, but should we add the file to the `.npmignore`?

Tyriar:
https://github.com/xtermjs/xterm.js/wiki/Contributing/_compare/b4d291f35323dadfaa4f5367ba7a835a8940d781...a184391ade71babd07daa1122273045fbc0a250a

@mofux yeah we should, not a stupid noobish question 😉

Tyriar:
@mofux actually I think it's fine because we blacklist everything:

https://github.com/xtermjs/xterm.js/blob/a52aab46180a29f322f7ab6ff092c4ffdd04b2b1/.npmignore#L1-L3

mofux:
You are right, I think that should have it covered 👍

Issue #1505, opened 2018-06-09
Replace IEventEmitter in API with explicit handlers


Tyriar:
```ts
// Instead of this:
this._xterm.on('key', f);
// Do this:
this._xterm.onKey(f);
```

We could remove IEventEmitter from the API and just use that internally. This would prevent errors like:

```ts
this._xterm.on('ky', f);
```

And also make it harder to access private events.

PR #1568, opened 2018-07-13
Make Terminal.emit signature comply with that of EventEmitter.emit

astefanutti:
In some use cases, it is required to be able to pass multiple arguments to the emit method.


Tyriar:
This was actually intentional, going forward we only want a single event object to house the args. Also we're going to deprecate emit/on soon and replace them with individual events on the API, see https://github.com/xtermjs/xterm.js/issues/1505

PR #1569, opened 2018-07-14
Deprecate Terminal.emit

Tyriar:
Part of #1505

### DIFF
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -416,6 +416,13 @@ declare module 'xterm' {
      */
     off(type: 'blur' | 'focus' | 'linefeed' | 'selection' | 'data' | 'key' | 'keypress' | 'keydown' | 'refresh' | 'resize' | 'scroll' | 'title' | string, listener: (...args: any[]) => void): void;
 
+    /**
+     * Emits an event on the terminal.
+     * @param type The type of event
+     * @param data data associated with the event.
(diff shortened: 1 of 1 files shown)


Tyriar:
@mofux I think we should just remove all the event stuff except for `onKey`/etc. from the public API, that would allow us to change the underlying implementation (like moving to something like VS Code's [`Emitter` class](https://github.com/Microsoft/vscode/blob/fec5cf6f43406264a131fc8ee5e05b7918c5dee0/src/vs/base/common/event.ts#L56), which works really well with TS). xterm.js isn't an eventing library so I never thought emit/on should have even been in there to begin with. 

> I'd prefer to still expose the on and once methods as Public API, because I think they are very handy for addons (especially the once method).

I'd like to see some example usages of these 🙂

mofux:
> I'd like to see some example usages of these 🙂

One use-case I came across recently is prompt line / command detection, where I've used `once('linefeed')` to know when the command was actually submitted, but I'm sure there are other handy use-cases too. On the other hand, if the `onLineFeed()` listener would return a disposable, I could have also done it like this:

```javascript
const disposable = term.onLineFeed(() => {

  // stop listening for consecutive events
  disposable.dispose();

  // ...

});
```

So yeah, guess it's fine to remove the `on`, `once` and `emit` methods from the public API as long as  `onKey` and friends return a disposable.

PR #1570, opened 2018-07-15
Prevent Uncaught TypeError  - Cannot read property 'isWrapped' on Linkifier

### LINKED ISSUE #1509: Uncaught TypeError: Cannot read property 'isWrapped' of undefined
https://github.com/ioquatix/script-runner/issues/94

Not sure if this is an issue wiht xterm.js or script-runner. Any ideas?

nikonso:
Fixes #1509

### DIFF
--- a/src/Linkifier.ts
+++ b/src/Linkifier.ts
@@ -4,7 +4,7 @@
  */
 
 import { IMouseZoneManager } from './ui/Types';
-import { ILinkHoverEvent, ILinkMatcher, LinkMatcherHandler, LinkHoverEventTypes, ILinkMatcherOptions, ILinkifier, ITerminal } from './Types';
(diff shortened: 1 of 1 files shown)


src/Linkifier.ts
```diff
@@ -169,10 +169,18 @@ export class Linkifier extends EventEmitter implements ILinkifier {
         return;
       }
       // If the first row is wrapped, backtrack to find the origin row and linkify that
+      let line = (<any>this._terminal.buffer.lines.get(absoluteRowIndex));
```
Tyriar: I don't think the `get` call is needed here, just the definition of `line`?

nikonso: Yes, you're right... fixed 😄

PR #1577, opened 2018-07-23
Prototype of iTerm2 like image display

mme:
This is a prototype of adding iTerm2 like image display capabilities to xterm.js, as requested in several tickets:

https://github.com/xtermjs/xterm.js/issues/614
https://github.com/zeit/hyper/issues/101
https://github.com/Microsoft/vscode/issues/44436

Images can also be clicked to be opened in full size in a new window.

xterm-screenshot

It works by slicing the image into patches of ```scaledCellWidth```/```scaledCellHeight``` size and adding the resulting ```ImageData``` object directly to the ```Buffer``` object. When a cell in the buffer is overwritten, the previous ```ImageData``` is thrown away and the memory is freed - e.g. this is useful when showing animated charts by overwriting the previous image.

We are developing an app that would greatly benefit from a cross platform terminal with image display, so I would be quite happy if this feature could make it to xterm.js in one way or another.

Please let me know if I can be of further help to have this feature land in xterm.js!

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -4,6 +4,7 @@ import * as fit from '../build/addons/fit/fit';
 import * as fullscreen from '../build/addons/fullscreen/fullscreen';
 import * as search from '../build/addons/search/search';
 import * as webLinks from '../build/addons/webLinks/webLinks';
+import * as imageLinks from '../build/addons/imageLinks/imageLinks';
 import * as winptyCompat from '../build/addons/winptyCompat/winptyCompat';
 
 
@@ -12,6 +13,7 @@ Terminal.applyAddon(fit);
 Terminal.applyAddon(fullscreen);
 Terminal.applyAddon(search);
 Terminal.applyAddon(webLinks);
+Terminal.applyAddon(imageLinks);
 Terminal.applyAddon(winptyCompat);
 
 
@@ -75,6 +77,7 @@ function createTerminal() {
   term.open(terminalContainer);
   term.winptyCompatInit();
   term.webLinksInit();
+  term.imageLinksInit();
   term.fit();
   term.focus();
 
--- a/package.json
+++ b/package.json
@@ -55,11 +55,11 @@
     "mocha": "gulp test",
     "build:docs": "jsdoc -c jsdoc.json",
     "tsc": "tsc",
-    "prebuild": "concurrently --kill-others-on-fail --names \"lib,attach,fit,fullscreen,search,terminado,webLinks,winptyCompat,zmodem,css\" \"tsc\" \"tsc -p ./src/addons/attach\" \"tsc -p ./src/addons/fit\" \"tsc -p ./src/addons/fullscreen\" \"tsc -p ./src/addons/search\" \"tsc -p ./src/addons/terminado\" \"tsc -p ./src/addons/webLinks\" \"tsc -p ./src/addons/winptyCompat\" \"tsc -p ./src/addons/zmodem\" \"gulp css\"",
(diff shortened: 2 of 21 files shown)


Tyriar:
@mme cool! 

I have some high level questions before I jump in:

1. How did you get imgcat to work? [This one](https://github.com/eddieantonio/imgcat) just uses background colors and https://github.com/sindresorhus/term-img is hard coded to only support iterm2.
2. > It works by slicing the image into patches of scaledCellWidth/scaledCellHeight size and adding the resulting ImageData object directly to the Buffer object.

   Did you explore any other solutions to the problem like stretching a row to be the size of an image?
3. > this is useful when showing animated charts by overwriting the previous image.

   Do gifs work with this implementation?

Here are some notes on how iterm places the image: https://github.com/xtermjs/xterm.js/issues/614#issuecomment-407415127

@jerch you'll be interested in the change to `CharData`: 

https://github.com/xtermjs/xterm.js/pull/1577/files#diff-c59b761ad01bc77f0c593de25ec4d847R21

https://github.com/xtermjs/xterm.js/pull/1577/files#diff-c59b761ad01bc77f0c593de25ec4d847R15

I think this is going the opposite direction for memory of the overall buffer to what we want (an extra pointer for every character).

Tyriar:
Managed to get it working on Linux by commenting out the support if blocks in `term-cli`:

image

First bug spotted when changing font family:

image

Not sure there's an easy way to fix this with the way that it currently works, what if instead of each cell having a reference to an `IRenderable` there is a single link somewhere indicating a row has an image? That would allow changing the font size to be done and the number of rows to expand/contract dynamically?

jerch:
Yupp thats a really cool feature addon, also Im kinda surprised it worked this straight forward. Still I have 2 issues with the current impl:

- To be able to tell in the core lib, if a cell will be covered by image data, the image needs to be processed, which currently introduced a frontend dep to `Image`. This will make the core/frontend separation more complicated - could we get this done with some "delegation abstraction" (read as interface to get image metrics from internal API xy). This way we keep it flexible, e.g. a standalone offscreen nodejs impl could use it own image processing module while a fullstack impl with frontend part can use the browser's `Image` class.
- The additional data slot in the term buffer is a problem - it will raise memory consumption alot. Also I think it is a waste of memory and not needed here since most cell never hold image data. Imho it should not applied this way. Question is how to  store rarely occuring cell based data efficiently.

Tyriar:
> which currently introduced a dep to Image

@jerch Good point, `src/InputHandler.ts` shouldn't touch DOM APIs, I guess the the `core/` should store the base64 string and `ui/` should own the `HTMLImageElement`.

@mme to give you more context on this, we're moving towards this architecture:

https://github.com/xtermjs/xterm.js/issues/1507#issue-330992942
https://github.com/xtermjs/xterm.js/issues/1507#issuecomment-402304660

> Question is how to store rarely occuring cell based data efficiently.

@jerch Well the `IMarker` APIs were added to track rows so maybe there could be a map somewhere containing this info:

```ts
inlineImages: { [markerId: number]: SomeImage };
```

jerch:
> Is there a way to tell InputHandler to somehow pause processing until the asynchronous onload event is fired and we get the image parsed by the browser?

Eww, thats kinda impossible since down the road from `InputHandler.parse` to the fully applied state changes xterm.js is synchronous (most if not all events are sync too). You are kinda asking for a conditional WAIT statement within a big block of sync code, a thing that JS does not support by any means (since it would block the one and only main thread). To get this done the event loop would have to step in by promisifying or applying the async/await syntatic sugar to all code (entering async stuff is viral, all stuff will be pulled into this paradigm).

mme:
@jerch I see. This not a big deal since the current implementation deals with this already.

Tyriar:
Thanks for the effort on this but I don't think we'll be moving forward with this any time soon. I'd say we should hold off on supporting this until we get more of the current stories finished up, have a good plan on how to handle varying height rows and we're sure we want to commit to supporting this (is it a security issue for apps being able to add tracking images?).

j-f1:
> is it a security issue for apps being able to add tracking images?

I’d say no, since they could just make a tracking request using normal HTTP APIs.

PR #1578, opened 2018-07-23
Fix NPE in InputHandler.parse

### LINKED ISSUE #1567: Exception when disposing the terminal while parsing
Repro:

- Run `yes`
- Kill terminal

VS Code issue: https://github.com/Microsoft/vscode/issues/54132

Tyriar:
Fixes #1567

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -193,6 +193,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
 
   // Store if user went browsing history in scrollback
   private _userScrolling: boolean;
+  private _isDisposed: boolean = false;
 
   private _inputHandler: InputHandler;
   public soundManager: SoundManager;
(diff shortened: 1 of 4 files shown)


Fonger:
Could you merge this fix?

Tyriar:
@Fonger it should be merged for 3.6

src/Terminal.ts
```diff
@@ -193,6 +193,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
 
   // Store if user went browsing history in scrollback
   private _userScrolling: boolean;
+  private _isDisposed: boolean = false;
```
mofux: Wouldn't it be better to add this property to the EventEmitter class (the super class of Terminal), so it tracks the disposed status?

mofux:
I've left a comment above, apart from that I'm happy to get this merged.

src/Terminal.ts
Tyriar: Great idea :+1:

PR #1580, opened 2018-07-24
Remove StaticCharAtlas support (fixes #1576)

### LINKED ISSUE #1576: Promote dynamic texture atlas to default and remove static texture atlas
VS Code is moving to dynamic for the default, after we ship it in stable I think we should remove the 2 other caches (static and none) and make dynamic the only one for xterm.js (while still retaining renderer abstractions to allow experimentation in the future). Maybe 3.7.0 would be good timing for this?

/cc @bgw

bgw:
This defaults to the dynamic atlas and leaves in `NoneCharAtlas` and the `experimentalCharAtlas` option (with `'static'` removed).

This leaves me with a few questions:

1. Do we want to completely remove support for `NoneCharAtlas` and `experimentalCharAtlas`? `NoneCharAtlas` provides a nice example of the modularity of the atlas system, even though it doesn't make much sense to use.

2. If we want remove the `experimentalCharAtlas` option, should we leave the typedef there (ignoring any values passed) for backwards compatibility? I think semver requires this, though it's not a runtime concern.

3. If we want to keep `experimentalCharAtlas`, should we rename it to `charAtlas`? What should we do if someone passes in an `experimentalCharAtlas` option? Should we leave `'static'` in there as an alias to `'dynamic'` for compatibility? I think semver requires this.

4. Unfortunately I don't think we have any established standard for warning about deprecated features. Is it appropriate to use `console.warn` in a library context like this? My guess is no.


Tyriar:
@bgw 

1. I guess we can keep NoneCharAtlas as it's so small, plus it will be useful to debug whether issues are caused by the dynamic atlas.
2. We could add `@deprecated` to the comment and a `console.warn` when it's set for 3.6, anything labelled experimental doesn't need to follow semver. I think mainly just Hyper and VS Code are using dynamic which are easily fixed during the update.
3. Yeah let's rename to `charAtlas`.
4. I normally add `@deprecated` to the function in .d.ts, a console.warn can't hurt too.

src/shared/atlas/CharAtlasGenerator.ts
```diff
@@ -1,143 +0,0 @@
...
- */
-
-import { FontWeight } from 'xterm';
-import { CHAR_ATLAS_CELL_SPACING, ICharAtlasConfig } from './Types';
-import { IColor } from '../Types';
-import { isFirefox, isSafari } from '../utils/Browser';
-
-declare const Promise: any;
-
-export interface IOffscreenCanvas {
```
Tyriar: 👍 to removing this, the `shared/` idea has been replaced with https://github.com/xtermjs/xterm.js/issues/1507

PR #1585, opened 2018-07-26
Make double click select word aware of wrapped lines

### LINKED ISSUE #1571: Double click to select word does not check wrapped lines
Repro:

`echo 'somelongstringthatwraps'`

Double click on one of the lines and notice that the whole word is not selected.

Tyriar:
Fixes #1571


mofux:
Works nicely 👍

PR #1588, opened 2018-07-29
Use const for null cell

### LINKED ISSUE #1584: Replace `32` in code with a named constant
TS will inline the constants. We typically write `32 /* ' ' */`, it would be better as a const.

Tyriar:
Fixes #1584


mofux:
LGTM 👍

PR #1590, opened 2018-07-29
Allow renderer to be swapped in and out at runtime

### LINKED ISSUE #1589: Allow renderers to be swapped in and out at runtime
Related: https://github.com/Microsoft/vscode/issues/53274

Tyriar:
Fixes #1589


mofux:
LGTM 👍

PR #1594, opened 2018-07-31
Fix bug in column selection rendering

### LINKED ISSUE #1593: Column selection doesn't render correctly when the selection starts above the viewport
VS Code issue: https://github.com/Microsoft/vscode/issues/55488

Tyriar:
Fixes #1593


mofux:
This only solves part of the problem for me. It works correctly while moving the mouse inside the terminal viewport, but the selection is expanded to the edge if the mouse leaves the viewport:

kapture 2018-08-06 at 15 44 25

Tyriar:
Good catch, not sure how I missed that :confused:

Tyriar:
Should be fixed now. This also reveals the root cause of the flicker that happens when dragging the selecting below the viewport, the root cause there were the 2 ways of updating the selection; one in `_onMouseMove` that set the end x to cols (correct) and one in `_dragScroll` that set the end x to cols - 1 :tada:

mofux:
Well done, works nicely now! Good to merge.

PR #1604, opened 2018-08-09
Allow convertEol to be be modified with setOption

quackingduck:
The documentation says this should be set with `setOption`, but the code that checks this value checks `core.convertEol` not `core.options.convertEol` ... so calling `setOption` has no effect :/

For anyone else running into this issue, you can force the value to be set with:

```
xterm.setOption('convertEol', true);
xterm._core.convertEol = true; // workaround
```

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -493,6 +493,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
         }
         break;
       case 'tabStopWidth': this.buffers.setupTabStops(); break;
+      case 'convertEol': this.convertEol = value; break;
     }
     // Inform renderer of changes
     if (this.renderer) {


src/Terminal.ts
```diff
@@ -493,6 +493,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
         }
         break;
       case 'tabStopWidth': this.buffers.setupTabStops(); break;
+      case 'convertEol': this.convertEol = value; break;
```
Tyriar: Yes it looks like `convertEol` is broken currently. A better fix would be to remove Terminal.convertEol and instead use Terminal.getOption('convertEol') where it's used in `InputHandler`.

jerch:
@quackingduck Could you try #1645 and see if the problem persists?

quackingduck:
Yep. That works! Feel free to close this

PR #1608, opened 2018-08-09
[WIP] Pull mouse mode handling into its own module

### LINKED ISSUE #1504: Pull mouse mode handling into its own module
The xterm mouse handling code is fairly self contained and could be pulled into its own module:

https://github.com/xtermjs/xterm.js/blob/ddc94669169714bdd3b5468c5f3dd69ff5955d13/src/Terminal.ts#L783-L793

nikonso:
Fixes #1504 

Work in progress, I have a small question directly in the code...

Feedback is welcome 😄

### DIFF
--- a/src/MouseEventsHandler.ts
+++ b/src/MouseEventsHandler.ts
@@ -1,0 +1,366 @@
+import { IMouseEventsHandler } from './Types';
+import { addDisposableDomListener } from './ui/Lifecycle';
+import { C0 } from './common/data/EscapeSequences';
+
+/**
+ * Copyright (c) 2018 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+/**
+  * XTerm mouse events
+  * http://invisible-island.net/xterm/ctlseqs/ctlseqs.html#Mouse%20Tracking
+  * To better understand these
+  * the xterm code is very helpful:
+  * Relevant files:
+  *   button.c, charproc.c, misc.c
+  * Relevant functions in xterm/button.c:
+  *   BtnCode, EmitButtonCode, EditorButton, SendMousePosition
+  */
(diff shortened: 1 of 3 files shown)


src/MouseEventsHandler.ts
```diff
@@ -0,0 +1,366 @@
...
+  *   button.c, charproc.c, misc.c
+  * Relevant functions in xterm/button.c:
+  *   BtnCode, EmitButtonCode, EditorButton, SendMousePosition
+  */
+export class MouseEventsHandler implements IMouseEventsHandler {
+  private _decLocator: boolean; // This is unstable and never set
+  private _vt300Mouse: boolean; // This is unstable and never set
+
+  constructor(
+    private _terminal: any
```
nikonso: How should I go about typing this? Should I add the missing properties to the `IInputHandlingTerminal`, `ITerminal`, or should I create a new interface?

nikonso: How should I go about typing this? Should I add the missing properties to the IInputHandlingTerminal, ITerminal, or should I create a new interface?

Tyriar: The ideal situation is that `MouseEventsHandler` does not depend on Terminal at all, instead it only communicates back via events, this would help break the circular dependency.

Also eventually we should pull all the mouse-related properties of Terminal into `MouseEventsHandler`, and move the non-DOM dependent parts into `./core/input/Mouse.ts`, (like `./core/input/), and the DOM dependent parts to `./ui/Mouse.ts` (or `./ui/UserInterface`). This direction is explained in more detail in https://github.com/xtermjs/xterm.js/issues/1507

Tyriar:
Closing this off as stale, feel free to open a new PR after rebasing if you want to continue no this 😃

PR #1619, opened 2018-08-18
Bug with receive the result from FileReader

meefik:
The incorrect "this" variable must be replaced with "fileReader".


meefik:
After call event 'load' we must be get result data from FileReader instance, in this case 'fileReader.result'. Please see the example https://developer.mozilla.org/en-US/docs/Web/API/FileReader/result#Example

PR #1622, opened 2018-08-21
Prevent whole viewport from refreshing

### LINKED ISSUE #1620: Full viewport refreshing more often than necessary
| Type | Version |
| --- | --- |
| OS | Windows 10 |
| Browser | Electron 2.0.7 |
| Xterm | 3.6.0 |

---

From some profiling/investigating I've been doing, the renderer appears to be doing more work than needed when simple inputs are provided (such as individual characters typed at the input prompt).

To reproduce this, I ran the following:

1. Ran the [xterm-electron-sample](https://github.com/princjef/xterm-electron-sample) project
2. Made the window full screen (1080p screen in my case) and ensured every line had something printed on it by running `ls` until the screen was full.
3. Opened the dev tools (Ctrl+Shift+I) and started a profiling session
4. Started typing random ascii characters on the shell prompt and recorded the render times for each keystroke.

Under these conditions with xterm 3.6.0, I'm seeing an average of 45-50ms to render each frame, even though there is only one line that is actually changing with each keystroke. Having looked through the code, the render call is explicitly told which line range to render, and yet still seems to be re-rendering everything each time.

I did some more digging and I think the problem comes from a combination of these two sections:

https://github.com/xtermjs/xterm.js/blob/0e45909c7e79c83452493d2cd46d99c0a0bb585f/src/ui/RenderDebouncer.ts#L28-L29

https://github.com/xtermjs/xterm.js/blob/0e45909c7e79c83452493d2cd46d99c0a0bb585f/src/ui/RenderDebouncer.ts#L47-L48

The first is doing an explicit check against `undefined` to handle the case when there are no bounds already set, but the second resets the values back to `null` after each render cycle, thereby bypassing the check. (point of interest: `Math.min(null, 10) === 0`)

I was able to dramatically improve the render time by changing the assignments in the second block to assign `undefined` instead of `null`. With the same test case, I'm now seeing rendering times of 4-5ms, instead of 45-50ms.

I'm happy to shoot over a PR to change it but wanted to double check the intent first. It looks like the checks were changed from checking against null to checking against undefined [intentionally](https://github.com/xtermjs/xterm.js/commit/34eb08952c8e78b20ea7109403f244d82c63331c) so I wasn't sure if I'm just missing something.

Tyriar:
Fixes #1620

### DIFF
--- a/src/ui/RenderDebouncer.ts
+++ b/src/ui/RenderDebouncer.ts
@@ -22,11 +22,16 @@ export class RenderDebouncer implements IDisposable {
     }
   }
 
-  public refresh(rowStart?: number, rowEnd?: number): void {
+  public refresh(rowStart: number, rowEnd: number): void {
+    // Get the min/max row start/end for the arg values
     rowStart = rowStart || 0;
     rowEnd = rowEnd || this._terminal.rows - 1;
-    this._rowStart = this._rowStart !== undefined ? Math.min(this._rowStart, rowStart) : rowStart;
-    this._rowEnd = this._rowEnd !== undefined ? Math.max(this._rowEnd, rowEnd) : rowEnd;
+    // Check whether the row start/end values have already been set
+    const isRowStartSet = this._rowStart !== undefined && this._rowStart !== null;
+    const isRowEndSet = this._rowEnd !== undefined && this._rowEnd !== null;
+    // Set the properties to the updated values
+    this._rowStart = isRowStartSet ? Math.min(this._rowStart, rowStart) : rowStart;
+    this._rowEnd = isRowEndSet ? Math.max(this._rowEnd, rowEnd) : rowEnd;
 
     if (this._animationFrame) {
       return;


mofux:
LGTM 👌

On 21. Aug 2018, at 22:14, Daniel Imms <notifications@github.com<mailto:notifications@github.com>> wrote:


Fixes #1620<https://github.com/xtermjs/xterm.js/issues/1620>

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/xtermjs/xterm.js/pull/1622

Commit Summary

  *   Prevent whole viewport from refreshing

File Changes

  *   M src/ui/RenderDebouncer.ts<https://github.com/xtermjs/xterm.js/pull/1622/files#diff-0> (11)

Patch Links:

  *   https://github.com/xtermjs/xterm.js/pull/1622.patch
  *   https://github.com/xtermjs/xterm.js/pull/1622.diff

PR #1624, opened 2018-08-22
Only render first line when requested

### LINKED ISSUE #1623: Editing first line causes full viewport re-render
| Type | Version |
| --- | --- |
| OS | Windows 10 |
| Browser | Electron 2.0.7 |
| Xterm | 3.6.0 |

---

While verifying the fix for #1620, I noticed things were still on the slower side, but this time only when i was typing on the first line of the terminal. Sorry for piling on with another one of these 😞

Same general process as before:

1. Ran the [xterm-electron-sample](https://github.com/princjef/xterm-electron-sample) project
2. Made the window full screen (1080p screen in my case) and clear the output with Ctrl+L to make sure my input is on the first line.
3. Opened the dev tools (Ctrl+Shift+I) and started a profiling session
4. Started typing random ascii characters on the shell prompt and recorded the render times for each keystroke.

Looks like this one is caused by this (specifically the second line):

https://github.com/xtermjs/xterm.js/blob/9273e8175e446d29c26d94c891ad023ee16df377/src/ui/RenderDebouncer.ts#L27-L28

I had initially thought this one was fine because I had been remembering that the end line was exclusive, not inclusive. However, this is not the case, so when editing the first line and only the first line, `rowEnd || this._terminal.rows - 1` evaulates to `this._terminal.rows - 1` because `rowEnd` is falsy, triggering a full viewport refresh.

I've verified that tightening up the check fixes the slowdown for this case. Will send a PR shortly.

As a side note, I was surprised by the amount of time spent rendering the blank lines at the bottom of my terminal. This may be specific to the configuration of my normal setup (there's some transparency which results in compositing of the layers), but if it's not being done already, it might be worth trying to bypass foreground drawing for space characters altogether.

princjef:
Fixes #1623 

I also switched the neighboring checks from the form `this._rowStart !== undefined && this._rowStart !== null` to `this._rowStart != null` because the two are equivalent. Happy to flip the checks to the more explicit ones if desired.

cc: @Tyriar

### DIFF
--- a/src/ui/RenderDebouncer.ts
+++ b/src/ui/RenderDebouncer.ts
@@ -24,8 +24,8 @@ export class RenderDebouncer implements IDisposable {
 
   public refresh(rowStart: number, rowEnd: number): void {
     // Get the min/max row start/end for the arg values
-    rowStart = rowStart || 0;
-    rowEnd = rowEnd || this._terminal.rows - 1;
+    rowStart = rowStart !== null && rowStart !== undefined ? rowStart : 0;
+    rowEnd = rowEnd !== null && rowEnd !== undefined ? rowEnd : this._terminal.rows - 1;
     // Check whether the row start/end values have already been set
     const isRowStartSet = this._rowStart !== undefined && this._rowStart !== null;
     const isRowEndSet = this._rowEnd !== undefined && this._rowEnd !== null;


princjef:
Sounds good, updated.

> I'd like to turn on forcing ===/!== at one point.

@Tyriar any reason for holding off? I just did a quick scrub and there are only about a dozen places in the code right now that would need to be updated, and the'yre all of the `== null` or `!= null` variety so it's a pretty straightforward switch.

princjef:
Threw the linter check in a branch. Here's the diff if you're curious: https://github.com/xtermjs/xterm.js/compare/master...princjef:triple-equals

Tyriar:
@princjef just no one has gotten to it, another PR would be 👌

Tyriar:
Also if we ever get to https://github.com/xtermjs/xterm.js/issues/1319 we'll be able to remove one of the checks in most of these cases, shame we can't enable it conditionally for parts of the codebase.

Issue #957, opened 2017-09-08
Link underline should be the same as FG color 


Tyriar:
Regression introduced in https://github.com/sourcelair/xterm.js/pull/938

![image](https://user-images.githubusercontent.com/2193314/30223657-9bc3e2d8-9480-11e7-9e40-52ae4091420c.png)

peacefullatom:
I'd like to grab this :)

Tyriar:
@peacefullatom great! 😃 

The link underlines are rendered from this file: https://github.com/xtermjs/xterm.js/blob/master/src/renderer/LinkRenderLayer.ts

You can check this code for how to get the foreground color:

https://github.com/xtermjs/xterm.js/blob/0e45909c7e79c83452493d2cd46d99c0a0bb585f/src/renderer/TextRenderLayer.ts#L139

peacefullatom:
@Tyriar, thank you for hints, they were helpful!

This [comment](https://github.com/xtermjs/xterm.js/pull/938/files#r137778166) by @mofux was very informative as well.

If there is anything I can add/improve just tell me.

Cheers! :)

PR #1626, opened 2018-08-23
#957 Link underline should be the same as FG color

peacefullatom:
Fixes [#957](https://github.com/xtermjs/xterm.js/issues/957).

### DIFF
--- a/src/Linkifier.ts
+++ b/src/Linkifier.ts
@@ -7,6 +7,7 @@ import { IMouseZoneManager } from './ui/Types';
 import { ILinkHoverEvent, ILinkMatcher, LinkMatcherHandler, LinkHoverEventTypes, ILinkMatcherOptions, ILinkifier, ITerminal, LineData } from './Types';
 import { MouseZone } from './ui/MouseZoneManager';
 import { EventEmitter } from './EventEmitter';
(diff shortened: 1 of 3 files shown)


src/Linkifier.ts
```diff
@@ -187,7 +188,7 @@ export class Linkifier extends EventEmitter implements ILinkifier {
     let text = this._terminal.buffer.translateBufferLineToString(absoluteRowIndex, false);
     let currentIndex = absoluteRowIndex + 1;
     while (currentIndex < this._terminal.buffer.lines.length &&
-        (<any>this._terminal.buffer.lines.get(currentIndex)).isWrapped) {
```
Tyriar: This was intentionally like this as it makes it easier to see it's a wrapped line, not the first in the while loop.

peacefullatom: Need some clarifications here.

If this check will be removed then `text` variable will contain all lines from visible part of  buffer.

This in turn will crash code.

So, I need to know what we trying to achieve by this code change.

Tyriar: I just mean the 2 extra spaces that were removed 😃

peacefullatom: Ok. I merged latest changes to this branch.

Ready to go! 😃

Issue #1629, opened 2018-08-24
Crosshair cursor doesn't go away after hitting alt+tab


Tyriar:
https://github.com/Microsoft/vscode/issues/57141

![image](https://user-images.githubusercontent.com/2193314/44591465-a61f9c80-a772-11e8-8eff-22097fa5bfd4.png)

peacefullatom:
May I grab this? 😃

Tyriar:
@peacefullatom please 😃

PR #1630, opened 2018-08-25
#1629: Mouse cursor style is refreshed after focusing on terminal

apo34:
Fixes remaining crosshair cursor after alt-tabbing out of terminal


Tyriar:
Sorry for not seeing this, it got fixed in https://github.com/xtermjs/xterm.js/pull/1636

PR #1632, opened 2018-08-25
buffer redesign - 1.) buffer preparations

jerch:
~~WIP - dont merge...~~

**Goal:** The current buffer layout makes it hard to extend the terminal with new features like true color support and such due to the already very high memory footprint. There are typed arrays for the rescue as discussed in #791. The final goal shall be to get a working replacement of the current buffer based on typed arrays with good performance and much less memory consumption.

There are several steps to be taken for a proper buffer redesign. With a typed array based approach the main problem behind the scenes is the fact, that most improvements rely on explicit memory management within the typed arrays (much like in C with `malloc` and `free`). The terminal buffer is the perfect place for those references to memory locations repesenting a single cell, but to get this working their lifecycle needs to be tracked. To not to poison the whole project with explicit ref handling (which is kinda an antipattern in a GC world), the buffer classes will be changed to hide most of these impl details from the outer world.

A rough summary of things to do:
- cleanup code base regarding buffer usage and prepare for next steps
- ~~extend buffer classes with ref tracking~~ _obsolete_
- typed array memory implementation
- convenient API for text attribute access, remove the current scattered attr handling
- ~~typed array based storage for text attributes with true color support~~ _obsolete_
- API for cell string content (should contain wcwidth)
- ~~typed array based string storage~~ _obsolete_
- finally: typed array based buffer line implementation

Each of these steps are likely to have big impact on the core codebase, therefore they will be split into several upcoming isolated PRs to get a smooth transition without breaking functionality.

This first PR primarily addresses awkwardities all over the current codebase regarding the buffer. The goal of this PR is to cleanup the buffer interaction to remove/lower dependencies to buffer implementation details and prepare for the upcoming changes:

- [x] implement a full buffer encapsulation for lines and cells: _done with class `TerminalLine`_
- [x] remove low level and `<any>` typed accesses from the current core codebase: _all index access to cells and `<any>` line types removed, most length changing actions (push/pop/splice) internalized_
- [x] ~~apply new module separation rules to buffer parts~~: _in `src` for now_
- [x] perf changes, ~~maybe reallow low level access in perf critical core sections with proper documentation~~: _perf is the same, low level access not needed_
- [x] ~~implement ref tracking in buffer classes (as testable dummy first)~~: _should go into own PR_

cc: @Tyriar @mofux @bgw


jerch:
It seems to be safe to remove `Terminal.blankLine`, it is not exposed outside anywhere (last commit). If we still need it as a convenient access with prefilled term settings we can revert the last commit.

jerch:
I am done with the basic preparations of the buffer stuff. To keep this PR small I will do the ref tracking stuff with a later PR.
Since I did not know where to move the new class it ended up in src. Maybe move it next to `CircularList` maybe? Also I am not happy with the class name `TerminalLine`, suggestions are more than welcome. Maybe `BufferLine` or `BufferRow`?

src/TerminalLine.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export class TerminalLine {
+  static defaultCell: CharData = [0, NULL_CELL_CHAR, NULL_CELL_WIDTH, NULL_CELL_CODE];
+  static blankLine(cols: number, attr: number, isWrapped?: boolean): TerminalLine {
+    const ch: CharData = [attr, NULL_CELL_CHAR, NULL_CELL_WIDTH, NULL_CELL_CODE];
+    return new TerminalLine(cols, ch, isWrapped);
+  }
+  private _data: CharData[];
+  public isWrapped = false;
+  length: number;
```
Tyriar: Let's use access modifiers everywhere, I'd like to use the `member-access` tslint rule but it's a bit of a pain with the tests. Wonder if we can enable it conditionally in a nice way

src/TerminalLine.ts
```diff
@@ -0,0 +1,93 @@
...
+  constructor(cols?: number, ch?: CharData, isWrapped?: boolean) {
+    this._data = [];
+    this.length = this._data.length;
+    if (cols) {
+      if (!ch) ch = TerminalLine.defaultCell;
```
Tyriar: Prefer {} + new line for block statements, I don't want to add a tslint rule for this as it's a little restrictive but let's do this in most cases as it's easier to read

src/TerminalLine.ts
```diff
@@ -0,0 +1,93 @@
...
+ *    - move Buffer.translateBufferLineToString here?
+ *    - next steps towards typed array:
+ *        - create ITerminalLine interface w'o length methods
+ *        - resize method
+ *        - replace all external push/pop/splice accesses
+ *        - fixed length
+ *        - remove push/pop/splice
+ *        - implement typed array alternative once string is removed from CharData
+ */
+export class TerminalLine {
```
Tyriar: Missed one in the winptyCompat addon:

```
(anonymous) @ winptyCompat.ts:35
EventEmitter.emit @ EventEmitter.ts:73
InputHandler.lineFeed @ InputHandler.ts:497
(anonymous) @ InputHandler.ts:194
EscapeSequenceParser.parse @ EscapeSequenceParser.ts:402
InputHandler.parse @ InputHandler.ts:320
Terminal._innerWrite @ Terminal.ts:1353
(anonymous) @ Terminal.ts:1322
setTimeout (async)
Terminal.write @ Terminal.ts:1321
Terminal.write @ Terminal.ts:117
displayData @ attach.ts:79
addonTerminal.__getMessage @ attach.ts:62
```

Can't test it on Windows atm

src/renderer/CharacterJoinerRegistry.test.ts
```diff
@@ -257,8 +260,19 @@ describe('CharacterJoinerRegistry', () => {
   });
 });
 
-function lineData(line: string, attr: number = 0): LineData {
-  return line.split('').map<CharData>(char => [attr, char, 1, char.charCodeAt(0)]);
+interface IPartialLineData {
```
Tyriar: I think this can be expressed as a `type` as well which is a much more readable imo:

```ts
type IPartialLineData = [string, number?];
```

src/TerminalLine.ts
jerch: Ok, gonna change it.
@tests: No clue, maybe with a separate config file just for tests with less strict TS rules?

src/TerminalLine.ts
jerch: Ok gonna change 'em, there are a few of those, lol,

src/TerminalLine.ts
jerch: Oppsie...

src/TerminalLine.ts
Tyriar: Yeah it would kind of suck to have 2 though... I was looking at converting to use `tslint --project tsconfig.json` which is the preferred over a file pattern, maybe we could do a separate one for the tests when we get to that.

jerch: Fixed.

src/renderer/CharacterJoinerRegistry.test.ts
jerch: Got it working with:
```TS
type IPartialLineData = ([string] | [string, number]);
```

src/InputHandler.ts
```diff
@@ -546,20 +550,12 @@ export class InputHandler extends Disposable implements IInputHandler {
...
-    let j = buffer.x;
-    const ch: CharData = [this._terminal.eraseAttr(), NULL_CELL_CHAR, NULL_CELL_WIDTH, NULL_CELL_CODE]; // xterm
-
-    while (param-- && j < this._terminal.cols) {
-      buffer.lines.get(row).splice(j++, 0, ch);
-      buffer.lines.get(row).pop();
-    }
+    this._terminal.buffer.lines.get(this._terminal.buffer.y + this._terminal.buffer.ybase).insertCells(
+      this._terminal.buffer.x,
+      params[0] || 1,
```
Tyriar: I'm not sure this is correct. The old one is splicing from buffer.x to terminal.cols, the new one is from buffer.x to 0?

src/InputHandler.ts
```diff
@@ -780,15 +799,16 @@ export class InputHandler extends Disposable implements IInputHandler {
   public eraseInLine(params: number[]): void {
     switch (params[0]) {
       case 0:
-        this._terminal.eraseRight(this._terminal.buffer.x, this._terminal.buffer.y);
+        this._eraseInBufferLine(this._terminal.buffer.y, this._terminal.buffer.x, this._terminal.cols);
```
Tyriar: Regression tests for InputHandler would set my mind at ease, these have been messed up before and those bugs are tough to track down.

jerch: Yeah, lemme get some regression tests up for those changes in `InputHandler`. At least I checked it against the integrations tests, they fail with the same faulty output (still deactivated).

src/InputHandler.ts
jerch: This inserts cells at position `buffer.x`, `(param[0] || 1)` times. Imho thats the same the old code did. Regression test to come...

Tyriar:
> Also I am not happy with the class name TerminalLine, suggestions are more than welcome. Maybe BufferLine or BufferRow?

Maybe `BufferLine` would be best? It's not stated anywhere but I think normally "row" is used when it's within the viewport (or referring to the viewport size) and "line" is a generic line somewhere in the `Buffer`.

src/TerminalLine.ts
```diff
@@ -0,0 +1,111 @@
...
+    this.length = this._data.length;
+    // TODO: ref new, unref old
+    return removed;
+  }
+
+  /** to be called when a line gets removed */
+  public release(): void {
+    // TODO: unref here
+  }
+  public toArray(): CharData[] {
```
Tyriar: This should be exposed as `data` in `TestTerminalLine`, see `TestTerminal` for the pattern.

src/TerminalLine.ts
```diff
@@ -0,0 +1,111 @@
...
+  // to be removed for typed array
+  public splice(start: number, deleteCount: number, ...items: CharData[]): CharData[] {
+    const removed = this._data.splice(start, deleteCount, ...items);
+    this.length = this._data.length;
+    // TODO: ref new, unref old
+    return removed;
+  }
+
+  /** to be called when a line gets removed */
+  public release(): void {
```
Tyriar: Do we still need this if we're not doing `Memory.ts`? Should this be called `dispose`?

src/TerminalLine.ts
```diff
@@ -0,0 +1,111 @@
...
+ */
+import { CharData } from './Types';
+import { NULL_CELL_CODE, NULL_CELL_WIDTH, NULL_CELL_CHAR } from './Buffer';
+
+/**
+ * Class representing a terminal line.
+ * Currently the class is a thin proxy to `CharData[]`.
+ * Once the storages are in place it will proxy access to
+ * typed array based line data.
+ * TODO:
```
Tyriar: Are all these TODOs still relevant? We should move them out of the code and into an issue/PR imo.

src/Terminal.ts
```diff
@@ -1730,16 +1729,13 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
    * @param x The column from which to start erasing to the start of the line.
    * @param y The line in which to operate.
    */
+  // FIXME: decide whether to remove from Terminal
```
Tyriar: This is only used in tests, if we can move this to InputHandler we may as well.

src/Terminal.ts
```diff
@@ -1767,37 +1763,11 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
    * Erase all content in the given line
    * @param y The line to erase all of its contents.
    */
+  // FIXME: decide whether to remove from Terminal
```
Tyriar: This is only used in tests, if we can move this to InputHandler we may as well.

src/Terminal.ts
```diff
@@ -1713,15 +1714,13 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
    * @param x The column from which to start erasing to the end of the line.
    * @param y The line in which to operate.
    */
+  // FIXME: decide whether to remove from Terminal
```
Tyriar: This is only used in tests, if we can move this to InputHandler we may as well.

src/Types.ts
```diff
@@ -7,6 +7,7 @@ import { Terminal as PublicTerminal, ITerminalOptions as IPublicTerminalOptions,
 import { IColorSet, IRenderer } from './renderer/Types';
 import { IMouseZoneManager } from './ui/Types';
 import { ICharset } from './core/Types';
+import { TerminalLine } from './TerminalLine';
```
Tyriar: We cannot import implementations into Types.ts, we would need an `IBufferLine` interface here.

src/TerminalLine.ts
```diff
@@ -0,0 +1,114 @@
...
+    // TODO: ref new, unref old
+    return removed;
+  }
+
+  /** to be called when a line gets removed */
+  public release(): void {
+    // TODO: unref here
+  }
+
+  public toArray(): CharData[] {
```
Tyriar: This should be exposed as data in TestTerminalLine, see TestTerminal for the pattern.

Tyriar:
Looks good other than the above comments, let's get them fixed and have it merged

src/TerminalLine.ts
jerch: Dont think so, with the new approach the lines are self containing and GCable without hassles and we do not need to worry about releasing refs anymore. Goal is to keep them internal at all cost.

src/TerminalLine.ts
jerch: Leftover - deleted.

src/Terminal.ts
jerch: Done.

src/Types.ts
jerch: Done.

src/TerminalLine.ts
jerch: Done.

jerch:
Done with changes. I renamed the class and file to 'BufferLine', it still resides in `/src` though.

Tyriar:
@jerch all good, we can move it with `Buffer`

src/InputHandler.ts
```diff
@@ -1989,4 +1986,67 @@ export class InputHandler extends Disposable implements IInputHandler {
   public setgLevel(level: number): void {
     this._terminal.setgLevel(level);  // TODO: save to move from terminal?
   }
+
+
```
Tyriar: Lots of space 😄

src/InputHandler.ts
```diff
@@ -1989,4 +1986,67 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+
+
+    /**
+   * Erase in the identified line everything from "x" to the end of the line (right).
+   * @param x The column from which to start erasing to the end of the line.
+   * @param y The line in which to operate.
+   */
+  // FIXME: decide whether to remove from Terminal
```
Tyriar: Remove

src/InputHandler.ts
```diff
@@ -1989,4 +1986,67 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+  public eraseRight(x: number, y: number): void {
```
Tyriar: I'm not sure we need these anymore?

src/InputHandler.ts
jerch: Good to take a deep ... ah well removed.

src/InputHandler.ts
jerch: Done.

src/InputHandler.ts
jerch: The methods are redundant now with the direct ones based on the new methods on `BufferLine`. Can be removed, Id leave them for this PR for documentation since the regression tests are based on them.

src/BufferLine.ts
```diff
@@ -24,10 +24,10 @@ export class BufferLine implements IBufferLine {
     this.length = this._data.length;
     if (cols) {
       if (!ch) {
-        ch = DEFAULT_CELL;
+        ch = [DEFAULT_CELL[0], DEFAULT_CELL[1], DEFAULT_CELL[2], DEFAULT_CELL[3]];
```
Tyriar: Maybe this should just be `[0, NULL_CELL_CHAR, NULL_CELL_WIDTH, NULL_CELL_CODE]` instead?

jerch: Done.

jerch:
https://github.com/xtermjs/xterm.js/pull/1632/commit/999fce4e52ae3b235118a92afe40b8845f56f69f is the cleaner way than moving stuff around in the source files, thx.

Looks all good to me. :+1:

PR #1634, opened 2018-08-26
Add shellvault.io to README.md

angersson:
This change adds a link to [Shellvault](https://www.shellvault.io), a SaaS I made that uses xterm.js to offer cloud SSH terminals.


angersson:
Cool, thanks for the quick merge!

PR #1641, opened 2018-08-31
buffer redesign - 2. Typedarray based BufferLine

jerch:
This is part two of the buffer redesign with a first replacement of `BufferLine` with a typed array based implementation. It is not yet optimized, just a straight forward rewrite.

Here are some numbers for memory consumption and runtime with my benchmark `ls -lR /usr/lib` in the demo with 100000 lines scrollback (87 cols per line):

- Old JS Array based:
grafik
grafik
grafik

- New typed array based:
grafik
grafik
grafik

Imho the results speak for themselves, a few notes though:
- heap usage dropped to ~25%
- JS heap usage dropped to 10%
- new version is ~30% faster than the old one in this test, but this is mainly due to the high scrollback (imho Chrome got into memory issues), with 1000 lines scrollback the new version is slightly slower (~10%)

Note the typed array based version is not yet optimized, there is much room to do so. In the 2nd image under "New typed array base" the high runtime of `typedArrayConstructByLength` and `BufferLine.set` is caused by `BufferLine.blankLine` - the creation of new typed arrays seems expensive, this can be avoided by reusing them in `Terminal.scroll` / `CircularBuffer.push` (atm `Terminal.scroll` creates new lines from scratch). Also the methods still need some performance love.
Fun fact - to get the memory statistics for the old version I had to close several programs or the chrome tab would just die. With 16GB RAM lol.


src/BufferLine.ts
```diff
@@ -80,4 +80,133 @@ export class BufferLine implements IBufferLine {
...
+export class BufferLine implements IBufferLine {
+  static blankLine(cols: number, attr: number, isWrapped?: boolean): IBufferLine {
+    const ch: CharData = [attr, NULL_CELL_CHAR, NULL_CELL_WIDTH, NULL_CELL_CODE];
+    return new BufferLine(cols, ch, isWrapped);
+  }
+  protected _data: Uint32Array | null = null;
+  protected _combined: {[index: number]: string} = {};
+  public length: number;
+
+  constructor(cols?: number, ch?: CharData, public isWrapped: boolean = false) {
```
Tyriar: I think cols should be a mandatory param?

src/BufferLine.ts
```diff
@@ -80,4 +80,133 @@ export class BufferLine implements IBufferLine {
...
+    this.length = cols || 0;
+    if (cols) {
+      if (!ch) {
```
Tyriar: Let's call `ch`, `fillCharData`

src/BufferLine.ts
```diff
@@ -80,4 +80,133 @@ export class BufferLine implements IBufferLine {
...
+
+const enum Cell {
+  FLAGS = 0,
+  STRING = 1,
+  WIDTH = 2,
+  SIZE = 3
+}
+
+export class BufferLine implements IBufferLine {
+  static blankLine(cols: number, attr: number, isWrapped?: boolean): IBufferLine {
```
Tyriar: Maybe blankLine should live in Buffer instead, that way we don't need it to be static? That would make it easier to use the factory mentioned above.

src/BufferLine.ts
```diff
@@ -80,4 +80,133 @@ export class BufferLine implements IBufferLine {
...
+}
+
+const enum Cell {
+  FLAGS = 0,
+  STRING = 1,
+  WIDTH = 2,
+  SIZE = 3
+}
+
+export class BufferLine implements IBufferLine {
```
Tyriar: For the opt-in flag we will want some factory, probably in `Buffer` that describes how to crease a BufferLine, something like this:

```ts
class Buffer {
  private _bufferLineFactory: (cols: number, ch?: CharData, isWrapped?: boolean) => IBufferLine
}
```

And the flag will call into:

```ts
class Buffer {
  public setBufferLineFactory(type: BufferLineFactoryType): void {
    if (type === BufferLineFactoryType.JsArray) {
      this._bufferLineFactory = ...
    } else if (type === BufferLineFactoryType.TypedArray) {
      this._bufferLineFactory = ...
    } else {
      throw...
    }
  }
}
```

src/BufferLine.ts
jerch: :+1: Yes that will be part of the next optimization steps. Did not want to change to much from the last PR.

src/BufferLine.ts
jerch: Ok.

src/BufferLine.ts
jerch: Yeah the blankLine abstraction is likely to change due to the new creation in scroll. Not sure yet, how to get this done without pulling to much impl details into `CircularList`.

src/BufferLine.ts
jerch: Sweet thx.

src/BufferLine.ts
Tyriar: Basically you want to reuse lines inside CircularList, instead of trimming we just move the end item to the start of the list? In other words `_startIndex++`? I don't think too many changes will be needed, perhaps a `shiftStart(amount: number)` method on `CircularList` is all that's needed?

jerch: I am not that familiar with the code of `CircularList`, I think this will work once the buffer has reached max length. Before that the new empty line has to be created somewhere, imho this should be handled in `CircularList`. Currently `Terminal.scroll` always creates this new empty line even when the buffer is full and pushs it there. Thats the perf culprit.

jerch:
@Tyriar: First version with reusing old lines dropped from `CircularList`. The idea is to save old lines that get overwritten and reuse them in `Buffer.blankLine`. Only `CircularList.push` does this atm, the other places are still commented out since they trigger failures in test cases (see fe16930). For some reason the other code sections create duplicate line entries in the circular buffer if uncommented, which should not happen at all, weird. Any idea why?
Maybe you have a better idea how to hold the old lines and reuse them. I'd prefer a solution that does not span two classes, maybe we can put the whole old-line-stack-thingy into CircularList?

Btw the `BufferLine` option is ready for testing.

mofux:
Just tested it out, works very well for me 👍 I also like the way you've implemented it. LGTM from my side - but let's wait for @Tyriar to review before merging as he's more experienced with this part of the code 😉

Tyriar:
@jerch I'm seeing the new impl being slower, is this expected until we recycle the BufferLine when scrolling? Do recent changes explain why what I'm seeing is different to the images in the main PR comment?

## Before

screen shot 2018-09-09 at 8 05 18 am

screen shot 2018-09-09 at 8 09 14 am

## After

screen shot 2018-09-09 at 8 05 27 am

screen shot 2018-09-09 at 8 09 07 am

jerch:
Tyriar:
Several things are different in both of your runs:

- currently the typed array version is a bit slower than the JSArray version yeah, but not that much (~10%), this will change once we recycle the old lines
- you have different data input speeds in both runs - an indication for that is this weird big yellow part at the end of the JSArray graph. I see this alot while testing, it makes it hard to compare the results. One way to avoid it in the demo is prebuffering in server.js (its kinda smoothing the input data for better comparability).
- renderer takes more time - I have not counted the invocations but it seems the typed array version does more render calls (just compare the green frame boxes above the graphs). Not quite sure why, might be related to the GC being called less frequently in the input chain giving the renderer more time to do things. Once we recycle the old lines the GC will drop far below 3%, with typed arrays in the parser it will go below 1%. Alomst no nasty GC stops anymore.

jerch:
@Tyriar Up for another review. Imho the `blankLine` perf regression for the typed array version should be addressed by a later PR.

Tyriar:
@jerch I'm reviewing this now, let's aim for pushing this in early 3.9.0 so we get plenty of testing on it.

jerch:
@Tyriar Yupp 3.9 is fine, it prolly will give enough time to test and solve the reusing issue for old buffer lines that dropped out of the `CircularList`.
There are still some logical twists in this idea, I already tried a simple straight forward reuse of lines, that are about to get overwritten in `CircluarList`, but that is not sufficient as lines might end up multiple times in the ring buffer.

src/BufferLine.ts
```diff
@@ -9,75 +9,265 @@ import { NULL_CELL_CODE, NULL_CELL_WIDTH, NULL_CELL_CHAR } from './Buffer';
...
+    newLine.copyFrom(this);
+    return newLine;
+  }
+}
+
+const enum Cell {
+  FLAGS = 0,
+  STRING = 1,
+  WIDTH = 2,
+  SIZE = 3
```
Tyriar: `SIZE` should be a separate constant, it's confusing to mix the items in the array as well as the size of each cell.

PR #1645, opened 2018-08-31
fix convertEol option

jerch:
A fix for the problem described in #1604.


jerch:
CI is broken again?

Tyriar:
The yarnpkg registry was broken yesterday due to some npm dns issue, hopefully it's resolved now...

PR #1646, opened 2018-09-01
#900 Add tests for events on Terminal

### LINKED ISSUE #900: Add tests for events on Terminal
We should add some tests the events here https://xtermjs.org/docs/api/terminal/#events

peacefullatom:
Here is my version of tests for events.

Honestly, don't know how to implement tests for 'blur' | 'focus' | 'linefeed' | 'selection', because they are emitted from outer components.

Please, let me know if I can add/improve something. 😃 

Fixes #900


src/Terminal.test.ts
```diff
@@ -68,24 +68,128 @@ describe('term.js addons', () => {
...
+          assert.equal(typeof key, 'string');
+          expect(event).to.be.an.instanceof(Object);
+        });
+
+        term.on('key', (key, event) => {
+          assert.equal(typeof key, 'string');
+          expect(event).to.be.an.instanceof(Object);
+        });
+
+        term.keyDown(evKeyDown);
```
Tyriar: This test needs to be asynchronous and verify both events pass, as currently it will pass if key and keydown don't fire

src/Terminal.test.ts
```diff
@@ -68,24 +68,128 @@ describe('term.js addons', () => {
...
+          assert.equal(typeof key, 'string');
+          expect(event).to.be.an.instanceof(Object);
+        });
+
+        term.on('key', (key, event) => {
+          assert.equal(typeof key, 'string');
+          expect(event).to.be.an.instanceof(Object);
+        });
+
+        term.keyPress(evKeyPress);
```
Tyriar: This test needs to be asynchronous and verify both events pass, as currently it will pass if key and keypress don't fire

src/Terminal.test.ts
```diff
@@ -68,24 +68,128 @@ describe('term.js addons', () => {
...
+    });
+
+    describe('refresh', () => {
+      it('should receive an object: {start: number, end: number}', () => {
+        term.on('refresh', (data) => {
+          expect(data).to.have.keys(['start', 'end']);
+          assert.equal(typeof data.start, 'number');
+          assert.equal(typeof data.end, 'number');
+        });
+        term.refresh(0, term.rows - 1);
```
Tyriar: This test needs to be asynchronous and verify the event happens (add done to end of event listener)

src/Terminal.test.ts
```diff
@@ -68,24 +68,128 @@ describe('term.js addons', () => {
...
+    });
+
+    describe('resize', () => {
+      it('should receive an object: {cols: number, rows: number}', () => {
+        term.on('resize', (data) => {
+          expect(data).to.have.keys(['cols', 'rows']);
+          assert.equal(typeof data.cols, 'number');
+          assert.equal(typeof data.rows, 'number');
+        });
+        term.resize(1, 1);
```
Tyriar: Add done to end of event handler

src/Terminal.test.ts
```diff
@@ -68,24 +68,128 @@ describe('term.js addons', () => {
...
+      });
+    });
+
+    describe('scroll', () => {
+      it('should receive a number', () => {
+        term.on('scroll', (ydisp) => {
+          assert.equal(typeof ydisp, 'number');
+        });
+        term.scroll();
```
Tyriar: Add done to end of event handler

src/Terminal.test.ts
```diff
@@ -68,24 +68,128 @@ describe('term.js addons', () => {
...
+      });
+    });
+
+    describe('title', () => {
+      it('should receive a string', () => {
+        term.on('title', (title) => {
+          assert.equal(typeof title, 'string');
+        });
+        term.handleTitle('title');
```
Tyriar: Add done to end of event handler

src/Terminal.test.ts
peacefullatom: After additional investigation, I've found out that `refresh` is a dummy [function](https://github.com/xtermjs/xterm.js/blob/3a310f1597668236802ceec54f3b07f53950a9c5/src/Terminal.test.ts#L29).

Next thing is that the `MockRenderer` also does not implement any [functionality](https://github.com/xtermjs/xterm.js/blob/3a310f1597668236802ceec54f3b07f53950a9c5/src/utils/TestUtils.test.ts#L343).

There are two ways:

1. Extend mock functionality (because we're unable to use real renderers due to lack of `window` object, etc).
2. Skip implementation of this test completely or implement a dummy test.

peacefullatom:
List of questions I have right now.

1. What is a preferable way to implement a test for `refresh` event: skip the test, create a dummy test, extend mock functionality?

2. Is it possible to add a `sinon` library to a project (including `@types/sinon`)? Because it will enable us to write more elegant tests with an ability to show more meaningful error messages for `keypress` and `keydown` events.

Or, maybe, you can give me a hint to some other solution?

Thanks in advance 😃

Tyriar:
@peacefullatom sorry about the delay 😅 

1. Probably just don't bother with that one as it's fired from the renderer and we don't really have a good story for testing when the DOM is involved heavily (yet).
2. I think the error message is fine as it fairly accurately describes what happened (no event was fired, it timed out waiting). What are some other wins would the lib give? I don't think we should pull in another lib if the benefits are very limited as it complicates thing and it's another thing people need to learn.

peacefullatom:
No problem! 😃

PR #1648, opened 2018-09-02
Improve demo build task

Tyriar:
This PR does the following:

- Upgrade webpack to 4
- Moves the demo build from gulp to npm scripts
- Instead of relying on concurrently the server and watch tasks are now done within the node script demo/start.js
- Source maps are inlined in the webpack (dev) build, they don't seem to resolve to src/ yet but they resolve to lib/
- Client is now in TS and uses xterm.d.ts, we could do this for server in the future (it will complicate the output of the task though). client.ts has its own tsconfig.json but shares the tslint.json

Using webpack is probably the way to go for the core library as well.


mofux:
Tested and verified working. LGTM 👍

Issue #705, opened 2017-06-14
Improve search addon API


Tyriar:
The find API introduced in #704 can be expanded upon in the future to support:

- Using a different selection type so that multiple instances can be highlighted at once
- Find multiple instances on the same line
- ~~Find terms that are wrapped across 2 lines~~
- ~~Options: regex, case sensitivity, etc.~~
- Return the number of results (may impact performance)
- ~~Allow "find as you type", allowing the current selection to be expanded if it matches~~

mofux:
It would also be great to have an api that gets the total number of matched instances, and the index of the currently **active** instance, as well as a mechanism to get notified when the total number or active index changes. This would allow a consumer to draw a find UI and provide some more info to the user.

Tyriar:
@mofux this was one of the first thoughts I had, I left it out due to performance concerns as it's not the most efficient thing to convert a large buffer into searchable strings as each individual character is stored within it's own array. The buffer is basically structured like this: `[number, string, number][][]` so you need to iterate through the lines and then construct a string out of the line.

Definitely worth exploring later on whether my guesses around this are true though 👍.

mofux:
@Tyriar Interesting to see that a modern terminal emulator shares so many problems with code editors like `ace` or `monaco`, like line buffers, selection models, find api, tokenization. Actually, I remember that the cloud9 terminal does all the rendering through the Ace editor 😉

Tyriar:
@mofux yeah it's been useful to have people who work on monaco to ask about things. A lot of stuff poses similar problems. The fact that the buffer is continuously trimmed and that it's common to run a command that get a crazy amount of data streamed in are biggest unique parts.

FGasper:
It would be nice to have a browser-style “find-as-you-type” logic. I think it would work if a couple functions were added:

* `findCurrentOrNext(needle)`
* `findCurrentOrPrevious(needle)`

These would expand the current selection if the `needle` is present at the 0-index of the current selection; otherwise they’d do a normal `findNext` or `findPrevious`.

Thoughts?

Tyriar:
@FGasper yeah this sort of thing would be good, probably best tackled after this one: Find multiple instances on the same line

jerch:
Coming from #1649 I suggest to split up the functionality of the search addon into:

- search: just do something like `indexOf` and return a cols/rows positions, feeding this back gives the next position, as an advantage this will not need any DOM related stuff.
[...]

PR #1649, opened 2018-09-02
Fix search addon

### LINKED ISSUE #1647: Search addon broken from buffer refactors
Repro:

1. ls 
2. Add something to find next text box
3. Hit enter

image

image

jerch:
Fixes #1647.

A small test for `SearchHelper` is included, was not able to do this more generally since the addon relies on the selection manager with `window` defined. I suggest to split this addon into a search and a mark addon.

### DIFF
--- a/src/addons/search/search.test.ts
+++ b/src/addons/search/search.test.ts
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) 2018 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { assert, expect } from 'chai';
+import * as search from './search';
+import { SearchHelper } from './SearchHelper';
+import { ISearchHelper } from './Interfaces';
+
+
+class MockTerminalPlain {}
+
+class MockTerminal {
+  private _core: any;
+  public searchHelper: ISearchHelper;
+  constructor(options: any) {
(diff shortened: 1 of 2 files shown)


src/addons/search/search.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      const hello0 = (term.searchHelper as any)._findInLine('Hello', 0);
+      const hello1 = (term.searchHelper as any)._findInLine('Hello', 1);
+      const hello2 = (term.searchHelper as any)._findInLine('Hello', 2);
+      expect(hello0).eql({col: 0, row: 0, term: 'Hello'});
+      expect(hello1).eql(undefined);
+      expect(hello2).eql({col: 11, row: 2, term: 'Hello'});
+      done();
+    }, 100);
+  });
+  // selection stuff not testable - depends on window
```
Tyriar: This is an option after https://github.com/xtermjs/xterm.js/issues/1128, this is also related to https://github.com/xtermjs/xterm.js/issues/705. I think for now let's remove this comment as it makes the code a bit messy and comment on https://github.com/xtermjs/xterm.js/issues/705 instead?

jerch: Done.

PR #1651, opened 2018-09-02
Move tslint prefer-const with the other core rules

Tyriar:
The ones at the bottom are separated as they're part of tslint-consistent-codestyle


mofux:
LGTM 👍

PR #1652, opened 2018-09-04
Preparation for project references

Tyriar:
- Update to TypeScript 3
- Make common only reference things within common
- Fix strictNullChecks within common

Part of #1319 
Part of #1337


mofux:
I've verified it's still working. LGTM 👍

Tyriar:
Thanks @mofux will merge these after the release

PR #1661, opened 2018-09-06
Add support for block and underline style cursor in DOM rendering mode

### LINKED ISSUE #1640: Support cursorStyle in DOM renderer
Code pointers:

https://github.com/xtermjs/xterm.js/blob/f47f095b8928a6e50ce80f87d7fd11e50cf9a190/src/renderer/dom/DomRenderer.ts#L161-L170

https://github.com/xtermjs/xterm.js/blob/f47f095b8928a6e50ce80f87d7fd11e50cf9a190/src/renderer/dom/DomRendererRowFactory.ts#L49-L51

Hindol:
First PR, so suggestions welcome. Fixes #1640.

### DIFF
--- a/src/renderer/dom/DomRenderer.ts
+++ b/src/renderer/dom/DomRenderer.ts
@@ -325,7 +325,8 @@ export class DomRenderer extends EventEmitter implements IRenderer {
 
       const row = y + terminal.buffer.ydisp;
       const lineData = terminal.buffer.lines.get(row);
-      rowElement.appendChild(this._rowFactory.createRow(lineData, { isCursorRow: row === cursorAbsoluteY, cursorStyle: terminal.getOption('cursorStyle') }, cursorX, terminal.charMeasure.width, terminal.cols));
+      const cursorStyle = terminal.options.cursorStyle;
+      rowElement.appendChild(this._rowFactory.createRow(lineData, row === cursorAbsoluteY, cursorStyle, cursorX, terminal.charMeasure.width, terminal.cols));
     }
 
     this._terminal.emit('refresh', {start, end});
--- a/src/renderer/dom/DomRenderer.ts
+++ b/src/renderer/dom/DomRenderer.ts
@@ -9,7 +9,7 @@ import { ITheme } from 'xterm';
 import { EventEmitter } from '../../EventEmitter';
 import { ColorManager } from '../ColorManager';
 import { RenderDebouncer } from '../../ui/RenderDebouncer';
-import { BOLD_CLASS, ITALIC_CLASS, CURSOR_CLASS, DomRendererRowFactory } from './DomRendererRowFactory';
+import { BOLD_CLASS, ITALIC_CLASS, CURSOR_CLASS, CURSOR_STYLE_BLOCK_CLASS, CURSOR_STYLE_BAR_CLASS, CURSOR_STYLE_UNDERLINE_CLASS, DomRendererRowFactory } from './DomRendererRowFactory';
 
 const TERMINAL_CLASS_PREFIX = 'xterm-dom-renderer-owner-';
 const ROW_CONTAINER_CLASS = 'xterm-rows';
@@ -160,13 +160,19 @@ export class DomRenderer extends EventEmitter implements IRenderer {
         `}`;
     // Cursor
     styles +=
-        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}.${FOCUS_CLASS} .${CURSOR_CLASS} {` +
+        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}:not(.${FOCUS_CLASS}) .${CURSOR_CLASS} {` +
+        ` outline: 1px solid ${this.colorManager.colors.cursor.css};` +
+        ` outline-offset: -1px;` +
+        `}` +
+        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}.${FOCUS_CLASS} .${CURSOR_CLASS}.${CURSOR_STYLE_BLOCK_CLASS} {` +
         ` background-color: ${this.colorManager.colors.cursor.css};` +
         ` color: ${this.colorManager.colors.cursorAccent.css};` +
         `}` +
-        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}:not(.${FOCUS_CLASS}) .${CURSOR_CLASS} {` +
-        ` outline: 1px solid #fff;` +
-        ` outline-offset: -1px;` +
+        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}.${FOCUS_CLASS} .${CURSOR_CLASS}.${CURSOR_STYLE_BAR_CLASS} {` +
+        ` box-shadow: 1px 0 0 ${this.colorManager.colors.cursor.css} inset;` +
+        `}` +
+        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}.${FOCUS_CLASS} .${CURSOR_CLASS}.${CURSOR_STYLE_UNDERLINE_CLASS} {` +
+        ` box-shadow: 0 -1px 0 ${this.colorManager.colors.cursor.css} inset;` +
         `}`;
     // Selection
     styles +=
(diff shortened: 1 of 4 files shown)


src/renderer/dom/DomRenderer.ts
```diff
@@ -164,9 +164,15 @@ export class DomRenderer extends EventEmitter implements IRenderer {
         ` background-color: ${this.colorManager.colors.cursor.css};` +
         ` color: ${this.colorManager.colors.cursorAccent.css};` +
         `}` +
-        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}:not(.${FOCUS_CLASS}) .${CURSOR_CLASS} {` +
+        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}:not(.${FOCUS_CLASS}) .${CURSOR_CLASS}.${CURSOR_STYLE_BLOCK_CLASS} {` +
         ` outline: 1px solid #fff;` +
```
mofux: Let's start using the cursor color that was defined by the theme. I know this wasn't solved before for the DOM renderer, but let's use this PR to do it right this time. You should be able to access the cursor color via `this.colorManager.colors.cursor.css`. In case of the `block` cursor style, you can get the color of the text inside the block cursor via `this.colorManager.colors.cursorAccent.css`

src/renderer/dom/DomRenderer.ts
```diff
@@ -164,9 +164,15 @@ export class DomRenderer extends EventEmitter implements IRenderer {
...
         ` outline-offset: -1px;` +
+        `}` +
+        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}:not(.${FOCUS_CLASS}) .${CURSOR_CLASS}.${CURSOR_STYLE_BAR_CLASS} {` +
+        ` box-shadow: 1px 0 0 #fff inset;` +
```
mofux: Same as above, let's use `this.colorManager.colors.cursor.css` instead of using a hard-coded `#fff` cursor color

src/renderer/dom/DomRenderer.ts
```diff
@@ -164,9 +164,15 @@ export class DomRenderer extends EventEmitter implements IRenderer {
...
+        `}` +
+        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}:not(.${FOCUS_CLASS}) .${CURSOR_CLASS}.${CURSOR_STYLE_UNDERLINE_CLASS} {` +
+        ` box-shadow: 0 -1px 0 #fff inset;` +
```
mofux: Same as above, let's use `this.colorManager.colors.cursor.css` instead of using a hard-coded `#fff` cursor color

src/renderer/dom/DomRenderer.ts
```diff
@@ -164,9 +164,15 @@ export class DomRenderer extends EventEmitter implements IRenderer {
         ` background-color: ${this.colorManager.colors.cursor.css};` +
         ` color: ${this.colorManager.colors.cursorAccent.css};` +
         `}` +
-        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}:not(.${FOCUS_CLASS}) .${CURSOR_CLASS} {` +
+        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}:not(.${FOCUS_CLASS}) .${CURSOR_CLASS}.${CURSOR_STYLE_BLOCK_CLASS} {` +
         ` outline: 1px solid #fff;` +
         ` outline-offset: -1px;` +
+        `}` +
+        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}:not(.${FOCUS_CLASS}) .${CURSOR_CLASS}.${CURSOR_STYLE_BAR_CLASS} {` +
```
mofux: This selector is wrong. It will only show the bar cursor style if the terminal is unfocused - but what we really want is to only show the bar cursor if the terminal is focused, no? So the selector should be changed to `${this._terminalSelector} .${ROW_CONTAINER_CLASS}.${FOCUS_CLASS} .${CURSOR_CLASS}.${CURSOR_STYLE_BAR_CLASS}`

src/renderer/dom/DomRenderer.ts
```diff
@@ -164,9 +164,15 @@ export class DomRenderer extends EventEmitter implements IRenderer {
...
+        ` box-shadow: 1px 0 0 #fff inset;` +
+        `}` +
+        `${this._terminalSelector} .${ROW_CONTAINER_CLASS}:not(.${FOCUS_CLASS}) .${CURSOR_CLASS}.${CURSOR_STYLE_UNDERLINE_CLASS} {` +
```
mofux: Wrong selector, as above - this will only show the underline cursor while the terminal is unfocused

Hindol:
@mofux Thanks for the suggestions. Fixed!

mofux:
Thanks for addressing the remaining issues and helping out closing the related issue. This works fine for me now!

@Tyriar Please merge if you're okay with the changes (and want to have it in the upcoming 3.7.0 release).

src/renderer/dom/DomRenderer.ts
```diff
@@ -319,7 +325,7 @@ export class DomRenderer extends EventEmitter implements IRenderer {
 
       const row = y + terminal.buffer.ydisp;
       const lineData = terminal.buffer.lines.get(row);
-      rowElement.appendChild(this._rowFactory.createRow(lineData, row === cursorAbsoluteY, cursorX, terminal.charMeasure.width, terminal.cols));
+      rowElement.appendChild(this._rowFactory.createRow(lineData, { isCursorRow: row === cursorAbsoluteY, cursorStyle: terminal.getOption('cursorStyle') }, cursorX, terminal.charMeasure.width, terminal.cols));
```
Tyriar: Accessing terminal.options.cursorStyle should be faster and I think be stringly typed to `"block" | "bar" | "underscore"`

src/renderer/dom/DomRendererRowFactory.ts
```diff
@@ -5,19 +5,22 @@
...
+export const CURSOR_STYLE_UNDERLINE_CLASS = 'xterm-cursor-underline';
 
 export class DomRendererRowFactory {
   constructor(
     private _document: Document
   ) {
   }
 
-  public createRow(lineData: IBufferLine, isCursorRow: boolean, cursorX: number, cellWidth: number, cols: number): DocumentFragment {
+  public createRow(lineData: IBufferLine, cursorOptions: ICursorOptions, cursorX: number, cellWidth: number, cols: number): DocumentFragment {
```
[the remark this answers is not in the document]
Hindol: I don't like the idea of an **optional** parameter `cursorStyle` sitting in the middle. That is why I added the interface in the first place, but I agree speed is important here. If you have any suggestions around the function signature, I'm all ears.

src/renderer/dom/DomRendererRowFactory.ts
```diff
@@ -10,14 +10,17 @@ import { IBufferLine } from '../../Types';
...
+export const CURSOR_STYLE_UNDERLINE_CLASS = 'xterm-cursor-underline';
 
 export class DomRendererRowFactory {
   constructor(
     private _document: Document
   ) {
   }
 
-  public createRow(lineData: IBufferLine, isCursorRow: boolean, cursorX: number, cellWidth: number, cols: number): DocumentFragment {
+  public createRow(lineData: IBufferLine, isCursorRow: boolean, cursorStyle: string, cursorX: number, cellWidth: number, cols: number): DocumentFragment {
```
Tyriar: > I don't like the idea of an optional parameter cursorStyle sitting in the middle.

You're right it could be `undefined` here, let's make it a little more explicit then and make the type `string | undefined` (we will eventually turn on `strictNullChecks` that requires this) and instead of handling the `'block'` case in the switch, let's make that the `default:` case. This will mean it's handled similar to the canvas version:

https://github.com/xtermjs/xterm.js/blob/3a310f1597668236802ceec54f3b07f53950a9c5/src/renderer/CursorRenderLayer.ts#L168

Tyriar:
Thanks @Hindol! Merging 😃

Issue #1658, opened 2018-09-05
Search addon: Support "match whole word" option


Tyriar:
When true, a search for `"term"` would match `"a term b"` but not `"a terms b"`

Plan issue: #705

alexr00:
It seems like space wouldn't be the only delimiter that signifies the start/end of a word. I'm thinking that anything that isn't a letter or number should be used as a "whole word delimiter". 
Examples where "term" would **not** match:
"a 5term b" 
"a term2 b"
Examples where "term" **would** match:
"a %term b"
"a (term} b"

Tyriar:
@alexr00 👍 I think this is what monaco does as well.

Issue #1656, opened 2018-09-05
Search addon: Support regex option


Tyriar:
Plan issue: #705

PR #1662, opened 2018-09-06
Added support for searching accross wrapped lines 

### LINKED ISSUE #1655: Search addon: Find terms that are wrapped across 2 lines
Plan issue: #705

alexr00:
Fixes #1655

### DIFF
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -265,28 +265,6 @@ export class Buffer implements IBuffer {
     return lineString.substring(startIndex, endIndex);
   }
 
-  /**
-   * Translates a buffer line to a string, including subsequent lines if they are wraps.
-   * Wide characters will count as two columns in the resulting string. This
-   * function is useful for getting the actual text underneath the raw selection
-   * position.
-   * @param line The line being translated.
-   * @param trimRight Whether to trim whitespace to the right.
-   */
-  public translateBufferLineToStringWithWrap(lineIndex: number, trimRight: boolean): string {
-    let lineString = '';
-    let lineWrapsToNext: boolean;
-
-    do {
-      lineString += this.translateBufferLineToString(lineIndex, true);
-      lineIndex++;
-      const nextLine = this.lines.get(lineIndex);
-      lineWrapsToNext = nextLine ? this.lines.get(lineIndex).isWrapped : false;
-    } while (lineWrapsToNext);
-
-    return lineString;
-  }
-
   public getWrappedRangeForLine(y: number): { first: number, last: number } {
     let first = y;
     let last = y;
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -266,7 +266,7 @@ export class Buffer implements IBuffer {
   }
 
   /**
-   * Translates a buffer line to a string, with optional start and end columns.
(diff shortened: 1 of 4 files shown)


src/Buffer.ts
```diff
@@ -265,6 +265,28 @@ export class Buffer implements IBuffer {
...
 
+  /**
+   * Translates a buffer line to a string, including subsequent lines if they are wraps.
+   * Wide characters will count as two columns in the resulting string. This
+   * function is useful for getting the actual text underneath the raw selection
+   * position.
+   * @param line The line being translated.
+   * @param trimRight Whether to trim whitespace to the right.
+   */
+  public translateBufferLineToStringWithWrap(lineIndex: number, trimRight: boolean): string {
```
Tyriar: Let's move this into SearchHelper for now as nothing touches it in the core lib as the search addon is an optional component.

src/addons/search/SearchHelper.ts
```diff
@@ -111,11 +111,11 @@ export class SearchHelper implements ISearchHelper {
...
    */
   private _findInLine(term: string, y: number): ISearchResult {
-    const lowerStringLine = this._terminal._core.buffer.translateBufferLineToString(y, true).toLowerCase();
+    const lowerStringLine = this._terminal._core.buffer.translateBufferLineToStringWithWrap(y, true).toLowerCase();
     const lowerTerm = term.toLowerCase();
     let searchIndex = lowerStringLine.indexOf(lowerTerm);
-    if (searchIndex >= 0) {
-      const line = this._terminal._core.buffer.lines.get(y);
+    const line = this._terminal._core.buffer.lines.get(y);
+    if ((searchIndex >= 0) && (searchIndex < line.length)) {
```
Tyriar: I think this will search the same lines multiple times, consider the following scenario:

1. `_findInLine(term, 0)`: Searches a line that wraps over the next 2 lines, no result found
2. `_findInLine(term, 1)`: Searches a line that wraps over the next line, no result found
3. `_findInLine(term, 2)`: Searches a line, no result found

Another interesting case to think about is when we add regex #1656 and match whole word #1658 this complicates things further, because if we search a partial line the way match whole word and regex `^` interact with the string is different. For example for a buffer 4 column wide that wraps:

```
ab c
de
```

- `"cde"` is a whole word
- `"de"` is _not_ a whole word
- `/^de/` should _not_ match
- `/cde$/` should match

src/Buffer.ts
alexr00: Moved in SearchHelper

src/addons/search/SearchHelper.ts
alexr00: Updated to handle this and added a test case to ensure it does not regress! I agree that the text should be considered as typed, not as displayed in the terminal. i.e., a "text line" might span multiple "terminal lines"

Tyriar:
@alexr00 now that the regex PR is merged in I think that's most of the conflicts we will hit, let's fix the conflicts, make sure the `^` and `$` work as expected on wrapped line and then they should be good to go 😃

alexr00:
Rebased to pickup changes and tested that ^ and $ work as expected on wrapped lines.

PR #1670, opened 2018-09-08
Fallback on default color when linkifier can't find the character

### LINKED ISSUE #1669: Linkifier doesn't handle some unicode strings
I was this in VS Code today:

screen shot 2018-09-08 at 12 00 17 pm

Failing on 140:

screen shot 2018-09-08 at 12 01 41 pm

https://github.com/xtermjs/xterm.js/blob/bdb6647ae0e83bba07b2c1937c9371a02382bde9/src/Linkifier.ts#L223

I have had reports of some links not working, perhaps this is causing it (by failing on a row before?).

Tyriar:
This probably doesn't fix the underlying problem, but it will make it
obvious by using the white background color and more easily help identify
the problem case.

Fixes #1669

---

I suspect it's related translating the string index to the column index.

### DIFF
--- a/src/Linkifier.ts
+++ b/src/Linkifier.ts
@@ -220,8 +220,11 @@ export class Linkifier extends EventEmitter implements ILinkifier {
     // Get cell color
     const line = this._terminal.buffer.lines.get(this._terminal.buffer.ydisp + rowIndex);
     const char = line.get(index);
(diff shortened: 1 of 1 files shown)


jerch:
Guess we should add this to master for now to get a hold of those broken linkifier attempts. Wasn't able to repro this, all nasty inputs I tried got just silently swallowed.

Issue #1560, opened 2018-07-09
Make "debug mode" public API


Tyriar:
Right now you run `term._core.debug = true` (and `term.setOption('debug', true)`), we should probably just make this public API

https://github.com/xtermjs/xterm.js/wiki/Contributing#enabling-debug-mode

It might also be good to changing it to "logging level" instead of "debug mode".

Tyriar:
Need to update the wiki for how to configure log level in v4

Tyriar:
Doc update part moved to https://github.com/xtermjs/xterm.js/issues/2305

PR #1678, opened 2018-09-09
Improve unicode string handling in linkifier

### LINKED ISSUE #1669: Linkifier doesn't handle some unicode strings

jerch:
Had to do several changes to get the cell/length offset mess working. This PR shall fix #1669.

**Changes in `Buffer.ts`:**
- `BufferStringIterator`: class to get a hold of the buffer string content in a convenient way
- `Buffer.stringIndexToBufferIndex`: method to translate a string index from a string retrieved by `Buffer.translateBufferLineToString` to get the row and cols indices in the buffer. Atm this method assumes that the string started at col 0. ~(Can be changed easily to respect `startCol` as well.)~
- `Buffer.contents`: return a `BufferStringIterator`

**Changes in `Linkifier.ts`:**
- reimplemented `_linkifyRow` and `_doLinkifyRow`

**TODO:**

- [x] reimplement `Linkifier._addLink` to support unicode char offsets within a match (currently only pre/post match offsets are respected)
- [x] tests for unicode chars in linkifier

~@alexr00 The new methods in `Buffer` can also fix the highlight offset problems of the selection manager. Might be worth a look once this PR can land.~ ~_Seems to work now._~ issue created #1680 + #1686

### DIFF
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -374,7 +374,7 @@ export class Buffer implements IBuffer {
     this.markers.splice(this.markers.indexOf(marker), 1);
   }
 
-  public contents(trimRight: boolean, startIndex?: number, endIndex?: number): IBufferStringIterator {
+  public iterator(trimRight: boolean, startIndex?: number, endIndex?: number): IBufferStringIterator {
     return new BufferStringIterator(this, trimRight, startIndex, endIndex);
   }
 }
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -420,7 +420,7 @@ export class BufferStringIterator implements IBufferStringIterator {
     const range = this._buffer.getWrappedRangeForLine(this._current);
     let result = '';
     for (let i = range.first; i <= range.last; ++i) {
-      // TODO: always apply trimRight after fixing #1685 
+      // TODO: always apply trimRight after fixing #1685
       result += this._buffer.translateBufferLineToString(i, (this._trimRight) ? i === range.last : false);
     }
     this._current = range.last;
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -194,6 +194,26 @@ export class Buffer implements IBuffer {
     this.scrollBottom = newRows - 1;
   }
 
+  public stringIndexToBufferIndex(lineIndex: number, stringIndex: number): number[] {
+    if (!stringIndex) {
+      return [lineIndex, 0];
+    }
+    while (stringIndex) {
+      let line = this.lines.get(lineIndex);
+      if (!line) {
+        [-1, -1];
+      }
+      for (let i = 0; i < line.length; ++i) {
+        stringIndex -= line.get(i)[CHAR_DATA_CHAR_INDEX].length;
+        if (stringIndex < 0) {
+          return [lineIndex, i];
+        }
+      }
+      lineIndex++;
+    }
+    return [lineIndex, 0];
+  }
+
   /**
    * Translates a buffer line to a string, with optional start and end columns.
    * Wide characters will count as two columns in the resulting string. This
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -408,8 +408,8 @@ export class BufferStringIterator implements IBufferStringIterator {
     private _buffer: IBuffer,
     private _trimRight: boolean,
     private _startIndex: number = 0,
-    private _endIndex: number = _buffer.lines.length)
-  {
+    private _endIndex: number = _buffer.lines.length
+  ) {
     this._current = this._startIndex;
   }
 
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -199,7 +199,7 @@ export class Buffer implements IBuffer {
    * To get the correct buffer position the string must start at `startCol` 0
    * (default in translateBufferLineToString).
    * The method also works on wrapped line strings given rows were not trimmed.
-   * The method operates on the CharData width attribute, there are no
+   * The method operates on the CharData string length, there are no
    * additional content or boundary checks. Therefore the string and the buffer
    * should not be altered in between.
    * TODO: respect trim flag after fixing #1685
@@ -208,9 +208,6 @@ export class Buffer implements IBuffer {
    * @param startCol column offset the string was retrieved from
    */
   public stringIndexToBufferIndex(lineIndex: number, stringIndex: number): BufferIndex {
-    if (!stringIndex) {
-      return [lineIndex, 0];
-    }
     while (stringIndex) {
       const line = this.lines.get(lineIndex);
       if (!line) {
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -408,14 +408,17 @@ export class BufferStringIterator implements IBufferStringIterator {
   private _start: number;
   private _end: number;
   private _current: number;
+
   constructor (private _buffer: IBuffer, private _trimRight: boolean, startIndex?: number, endIndex?: number) {
     this._start = startIndex || 0;
     this._end = endIndex || this._buffer.lines.length;
     this._current = this._start;
   }
+
   public hasNext(): boolean {
     return this._current < this._end;
   }
+
   public next(withRanges: boolean = false): string | [{first: number, last: number}, string] {
     const range = this._buffer.getWrappedRangeForLine(this._current);
     let result = '';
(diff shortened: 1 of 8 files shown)


jerch:
Ready for review :smile_cat:

src/Buffer.ts
```diff
@@ -194,6 +194,39 @@ export class Buffer implements IBuffer {
...
+   * The method also works on wrapped line strings given rows were not trimmed.
+   * The method operates on the CharData width attribute, there are no
+   * additional content or boundary checks. Therefore the string and the buffer
+   * should not be altered in between.
+   * TODO: respect trim flag after fixing #1685
+   * @param lineIndex line index the string was retrieved from
+   * @param stringIndex index within the string
+   * @param startCol column offset the string was retrieved from
+   */
+  public stringIndexToBufferIndex(lineIndex: number, stringIndex: number): BufferIndex {
```
Tyriar: `SelectionManager._convertViewportColToCharacterIndex` and `SelectionManager._getWordAt` are closely related to this, not sure we can reuse any. Regardless, it's probably better not to do it in this PR.

src/CharWidth.ts
```diff
@@ -169,3 +169,26 @@ export const wcwidth = (function(opts: {nul: number, control: number}): (ucs: nu
       return wcwidthHigh(num);
     };
 })({nul: 0, control: 0});  // configurable options
+
+/**
+ * Get the terminal cell width for a string.
+ * @param s
```
Tyriar: Better to exclude the @param if it doesn't have docs

src/CharWidth.ts
```diff
@@ -169,3 +169,26 @@ export const wcwidth = (function(opts: {nul: number, control: number}): (ucs: nu
...
+ */
+export function stringWidth(s: string): number {
```
Tyriar: This could do with some tests.

Tyriar: `getStringCellWidth` is more explicit on what it does

src/Linkifier.ts
```diff
@@ -80,9 +81,24 @@ export class Linkifier extends EventEmitter implements ILinkifier {
    */
   private _linkifyRows(): void {
     this._rowsTimeoutId = null;
-    for (let i = this._rowsToLinkify.start; i <= this._rowsToLinkify.end; i++) {
-      this._linkifyRow(i);
+
+    const absoluteRowIndexStart = this._terminal.buffer.ydisp + this._rowsToLinkify.start;
+    if (absoluteRowIndexStart >= this._terminal.buffer.lines.length) {
```
Tyriar: Let's add the `// Ensure the row exists` back to help describe what's going on.

src/Linkifier.ts
```diff
@@ -153,99 +169,64 @@ export class Linkifier extends EventEmitter implements ILinkifier {
...
+    // clone regex to do a global search on text
+    const rex = new RegExp(matcher.regex.source, matcher.regex.flags + 'g');
+    let match;
+    let stringIndex = -1;
+    while ((match = rex.exec(text)) !== null) {
+      const uri = match[typeof matcher.matchIndex !== 'number' ? 0 : matcher.matchIndex];
+      if (!uri) {
+        // something matched but does not comply with the given matchIndex
+        // since this is most likely a bug the regex itself we simply do nothing here
+        // TODO: should this be logged for debugging?
```
Tyriar: Maybe we should `throw new Error('...')` here so it's caught and fixed fast if it happens?

src/Linkifier.ts
```diff
@@ -153,99 +169,64 @@ export class Linkifier extends EventEmitter implements ILinkifier {
...
+        break;
+      }
 
-    // Get index, match.index is for the outer match which includes negated chars
-    const index = text.indexOf(uri);
+      // due to complex regexes we cannot use match.index directly
```
Tyriar: I think my description on the left if a bit more clear, also we should remove the TODO as I don't think it's worth keeping in.

src/Buffer.ts
```diff
@@ -366,3 +403,38 @@ export class Marker extends EventEmitter implements IMarker {
...
+    let result = '';
+    for (let i = range.first; i <= range.last; ++i) {
+      // TODO: always apply trimRight after fixing #1685
+      result += this._buffer.translateBufferLineToString(i, (this._trimRight) ? i === range.last : false);
+    }
+    this._current = range.last + 1;
+    return (withRanges) ? [range, result] : result;
+  }
+
+  public toArray(): string[] {
```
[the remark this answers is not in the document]
jerch: Yeah, removed it completely and replaced it with `next` invocations in the tests.

src/Buffer.ts
```diff
@@ -366,3 +403,38 @@ export class Marker extends EventEmitter implements IMarker {
...
+    const result = [];
+    while (this.hasNext()) {
+      result.push(this.next() as string);
```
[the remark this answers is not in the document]
jerch: ǹext`is currently used in he linkifier, I created an interface for the iterator result for now (always contains the range) and removed the overloaded return type.
I was at first unsure how to put/integrate this, even thought about a separate range only iterator. Also the whole iterator tries to solve some of the issues you addressed with your proposal for the reflow buffer redesign, imho it will see further redesign in the future.

src/Buffer.ts
```diff
@@ -366,3 +403,38 @@ export class Marker extends EventEmitter implements IMarker {
     super.dispose();
   }
 }
+
+export class BufferStringIterator implements IBufferStringIterator {
+  private _start: number;
+  private _end: number;
+  private _current: number;
+
+  constructor (private _buffer: IBuffer, private _trimRight: boolean, startIndex?: number, endIndex?: number) {
```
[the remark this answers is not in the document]
jerch: Done.

src/Linkifier.ts
```diff
@@ -80,9 +81,24 @@ export class Linkifier extends EventEmitter implements ILinkifier {
...
+      return;
+    }
+
+    // iterate over the range of unwrapped content strings within start..end
+    // _doLinkifyRow gets full unwrapped lines with the start row as buffer offset for every matcher
+    // for wrapped content over several rows the iterator might return rows outside the viewport
+    // we skip those later in _doLinkifyRow
+    const linesIterator = this._terminal.buffer.contents(false, absoluteRowIndexStart, this._terminal.buffer.ydisp + this._rowsToLinkify.end + 1);
+    while (linesIterator.hasNext()) {
+      const lineData: any = linesIterator.next(true);
```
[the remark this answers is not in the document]
jerch: Must have been the cat :smile_cat: ....fixed.

src/Buffer.ts
jerch: Yeah figured that. I think we should consolidate this in the future and remove doubled implementations from the code base, maybe with resolving #1686. Might also be a starting for your reflow proposal afterwards (its kinda interlinked).

src/CharWidth.ts
jerch: Done.

src/CharWidth.ts
jerch: Done.

src/Linkifier.ts
jerch: Done.

src/Linkifier.ts
jerch: Eww a hard `throw` here will also stop any later linkifier matchers, means if the first matcher regex is broken none will be linkified at all. Thats imho a good thing in a debug session yeah, but in live code it is likely to create tons of issue reports for any integrator. Do we have some debug flag setup I could hide this behind? Do we still use/propagate the `terminal.debug` flag?

src/Linkifier.ts
jerch: Done.

src/CharWidth.ts
jerch: Kinda hard to test, since it is interlinked with the char handling in `InputHandler.print`. Can try to compare the results with the sum of buffer line wcwidth values.
Imho this and `wcwidth` above needs a better integration by some string handling class in the future, that does all the nasty unicode and wcwidth stuff and can be used by `InputHandler` and `Buffer` methods at any time.

jerch: Wrote some tests that compare the buffer wcwidths with the result of this function.

src/Linkifier.ts
jerch: Ah you already created an issue for the debug  settings (#1560). Added the error behind terminal.debug for now.

jerch:
@Tyriar I think I have addressed all review issues. Up for the next round :smile_cat: 

Could you also test the issue from https://github.com/xtermjs/xterm.js/issues/1669#issuecomment-419718223 I think it is not related to the linkifier at all and will need separate tracking before closing the issue.

PR #1690, opened 2018-09-14
Expose Viewport syncScrollArea() to Terminal and include force param

DBjelovuk:
To allow explicit invocation of scroll sync.

See: https://github.com/Microsoft/vscode/issues/58532

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -1287,6 +1287,10 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
     }
   }
 
+  public syncScrollArea(force?: boolean): void {
+    this.viewport.syncScrollArea(force);
+  }
+
   /**
    * Writes text to the terminal.
    * @param data The text to write to the terminal.
--- a/src/Types.ts
+++ b/src/Types.ts
@@ -91,7 +91,7 @@ export interface IInputHandlingTerminal extends IEventEmitter {
 
 export interface IViewport extends IDisposable {
   scrollBarWidth: number;
-  syncScrollArea(): void;
+  syncScrollArea(force?: boolean): void;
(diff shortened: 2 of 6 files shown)


Tyriar:
I notice in vscode that scroll to bottom/top commands don't do anything to the scrollbar, that seems wrong as all they do is forward calls to xterm.js. This also appears to be the case for scroll by line/page as well.

https://github.com/Microsoft/vscode/blob/7ccd93ed5ea3e21b5cf70db7ba73e3adfdd0d0d9/src/vs/workbench/parts/terminal/electron-browser/terminalInstance.ts#L684-L706

DBjelovuk:
It looks like `Viewport._refresh()` is the linchpin to get that scroll position synced up<sup>1</sup>, but it's contingent upon a delta of buffer length or element/height when invoked from `syncScrollArea()` <sup>2</sup>, neither of which is satisfied by a panel toggle or tab change action in vscode. As far as those `scroll`- functions, I can't trace any of them back to `Viewport._refresh()` either so it's looking to me like the missing piece of the puzzle.

<sup>1</sup> Simply replacing https://github.com/Microsoft/vscode/blob/7ccd93ed5ea3e21b5cf70db7ba73e3adfdd0d0d9/src/vs/workbench/parts/terminal/electron-browser/terminalInstance.ts#L666 with `this._xterm._core.viewport._refresh()` achieves the intended behaviour.

<sup>2</sup> Triggered from here after that 'scroll' emit in vscode from what I can make out:
https://github.com/xtermjs/xterm.js/blob/40cb057d3049abc472aa28aa19aabe36b93f591e/src/Terminal.ts#L727

Tyriar:
Fixed in https://github.com/xtermjs/xterm.js/pull/1698, thanks for the help 😃

Issue #1688, opened 2018-09-14
Improve testing and tracking of performance critical components


Tyriar:
Rendering performance has regressed over 100% since 3.3.0 https://github.com/xtermjs/xterm.js/issues/1677, we should improve how we test and track this. I'd love to hear from people on how we could go about doing this is a good way but this is what I think we want:

- A benchmark test suite which prints numbers for things like
  - Filled line render time
  - Empty line render time
  - Full viewport render time
  - Buffer write/read
  - Fill buffer
  - etc.
- Eventually a dashboard or some way to track this periodically

jerch:
We need an APM for this, lets stall xterm.js development for 3 ys and make the performance tooling first (or gather some bucks and at least 20 highly skilled C++/JS developer to get the job done). :laughing:

There are a few profiling tools that will cover parts of your list with reliable results (esp. components that could be tested in nodejs with a high res synchronous timer), as soon as the browser engine gets involved we are stuck with the nerfed timer due to Spectre. Since in the end all that counts is the user perceived performance the latter is still testable by doing "full runs" with typical actions (like my current `ls` benchmark) and comparing the numbers from the integrated profilers. Those numbers are less reliable though and contain noise, imho chrome and firefox use a statistical approach to get the numbers by peeking into the JS callstack periodically. This testing could be done in a selenium env, maybe electron allows there additional interaction. Last but not least chrome exhibits many debug switches that would help with tracing tasks.

Since you wrote this issue from the canvas perf regression perspective - imho this is even more tricky to test in a reliable manner, it heavily relies on system specifics like the OS, installed GPU and might even be driver version dependent. Under such circumstances an "once and for all" optimal solution does not exists.

**TL;DR**
- We could test core components isolated with standard nodejs tooling.
- We could test end user experience with typical actions in selenium envs or with electron and the built in profilers.
- No clue how to get reliable numbers for GPU driven stuff, lol.

jerch:
Here is a proof of concept perf tool, that gets the timeline data from chrome: https://github.com/jerch/perf-test. To run it, edit the options in `example.js` to your needs, start the xterm.js demo and run the example. It talks with chrome via the debugging protocol, I was not able to get the data with the webdriver (the timeline data was removed several versions ago from the selenium chromedriver).

Tyriar:
Current plan:

1. Improve chrome-timeline https://github.com/jerch/chrome-timeline/issues
[...]

PR #1692, opened 2018-09-14
Dynamic atlas: Reduce unnecessary objects generated and draw from ImageBitmap

### LINKED ISSUE #1677: Render performance degradation from 3.0.2 to 3.7.0
**Update:** This issue should be introduced between 3.3.0 and 3.4.0.

----------------------

Originally reported at: https://github.com/tsl0922/ttyd/issues/126

## Details

- Browser and browser version: Chrome69.0.3497.81
- OS version: macOS High Sierra 10.13.6
- xterm.js version: 3.7.0

### Steps to reproduce

1. Start the 3.7.0 demo.
2. Open the demo with chrome, and turn on the FPS meter.
2. Resize it to a larger size (155 cols / 40 rows).
3. Run: `vtop` (`npm install -g vtop`).
4. Press and hold the down arrow button.
5. Result: 4-14 PFS on 3.7.0 (not stable) and 30-35 FPS on 3.3.0 (stable).

It's much smoother on 3.0.2 (ttyd 1.4.0) than 3.7.0 (ttyd 1.4.1).

Tyriar:
This PR removed several of the in-between objects needed to render using the dynamic texture atlas. It's a bit difficult to measure this currently (https://github.com/xtermjs/xterm.js/issues/1688), the one piece that's very easy to measure is the impact on using a number instead of a string for the cache key:

Before:

screen shot 2018-09-14 at 12 11 50 pm

After:

screen shot 2018-09-14 at 12 12 02 pm

I couldn't find anything else that could contribute to improving #1677. I also experimented with drawing using an ImageBitmap but couldn't find big changes in performance so I left it out of the PR: https://github.com/Tyriar/xterm.js/commit/e8d06b677868e57887cd9b20b095984833cffe75

Fixes #1677


jerch:
@Tyriar 
Benchmarking with my `ls -lR /usr/lib` (average of 10 runs):

- master: ~850ms for `TextRenderLayer._forEachCell`
- this PR: ~750ms for `TextRenderLayer._forEachCell`
- your 1677_bitmap branch: ~550ms for `TextRenderLayer._forEachCell`

I see a really big difference with the ImageBitmap on Ubuntu with Chrome 68 and nvidia gc. Welcome to browser version + driver differences? Idk.

Tyriar:
@jerch oh yeah? I felt like I saw that earlier and then didn't... I can continue polishing that up with the fallback if it looks consistent.

jerch:
@Tyriar: Retested, its stable between 430 - 550 ms now :+1:

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -230,13 +290,76 @@ export default class DynamicCharAtlas extends BaseCharAtlas {
...
+      this._expandGlyphWaitingOnBitmapQueue();
+    }
+    this._glyphsWaitingOnBitmapQueue[this._glyphsWaitingOnBitmapCount++] = getGlyphCacheKey(code, fg, bg, bold, dim, italic);
+
+    // Check if bitmap generation timeout already exists
+    if (this._bitmapCommitTimeout !== null) {
+      return;
+    }
+
+    this._bitmapCommitTimeout = window.setTimeout(() => this._generateBitmap(), GLYPH_BITMAP_COMMIT_DELAY);
```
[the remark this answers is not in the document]
Tyriar: > You might see a win from just doing a setTimeout(..., 0), since it means that we're not trying to create the bitmap while also trying to finish drawing the current frame.

The reason for the delay is to batch the number of calls to `window.createImageBitmap`.

> With this change, should we make character atlases implement Disposable so that we can clearTimeout appropriately?

Good idea 😃

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -264,9 +275,47 @@ export default class DynamicCharAtlas extends BaseCharAtlas {
...
+    const countAtGeneration = this._glyphsWaitingOnBitmapCount;
+    // TODO: Fallback when createImageBitmap not supported
+    window.createImageBitmap(this._cacheCanvas).then(bitmap => {
+      // Set bitmap
+      this._bitmap = bitmap;
+
+      // Mark all new glyphs as in bitmap
+      for (let i = 0; i < countAtGeneration; i++) {
+        const key = this._glyphsWaitingOnBitmap[i];
+        this._cacheMap.get(key).inBitmap = true;
```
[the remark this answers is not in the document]
Tyriar: We could probably keep it this way with the below handed, working with array buffers is a lot faster than normal "holey" arrays.

> The cache value could've been evicted from the map by the time you try to read it.

Good point, I need to cover this case.

> .get() marks the entry as recently used and shifts it to the top of the queue. Writing the glyph to the bitmap shouldn't count as a use.

I would work around this by exposing a `peek()` function that gets without this shift? That way it's just a fast hash lookup `this._map[key]`.

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -29,14 +30,39 @@ const TRANSPARENT_COLOR = {
...
+  // Note that this only returns a valid key when code < 256
+  // Layout:
+  // 0b00000000000000000000000000000001: italic (1)
+  // 0b00000000000000000000000000000010: dim (1)
+  // 0b00000000000000000000000000000100: bold (1)
+  // 0b00000000000000000000111111111000: fg (9)
+  // 0b00000000000111111111000000000000: bg (9)
+  // 0b00011111111000000000000000000000: code (8)
+  // 0b11100000000000000000000000000000: unused (3)
+  return code << 21 | bg << 12 | fg << 3 | (bold ? 0 : 4) + (dim ? 0 : 2) + (italic ? 0 : 1);
```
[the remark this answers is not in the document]
Tyriar: Also keep in mind this complexity is encapsulated within this function, this is an easy win for the numbers imo:

screen shot 2018-09-17 at 7 21 21 am

src/renderer/atlas/Types.ts
```diff
@@ -5,13 +5,3 @@
 
 export const INVERTED_DEFAULT_COLOR = -1;
 export const DIM_OPACITY = 0.5;
-
-export interface IGlyphIdentifier {
```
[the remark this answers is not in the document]
Tyriar: Sounds good, I can undo this one. I might also keep an "active glyph identifier" object around and reuse that, best of both worlds?

src/renderer/atlas/LRUMap.ts
```diff
@@ -6,12 +6,12 @@
...
   prev: ILinkedListNode<T>;
   next: ILinkedListNode<T>;
-  key: string;
+  key: number;
   value: T;
 }
 
 export default class LRUMap<T> {
-  private _map: { [key: string]: ILinkedListNode<T> } = {};
+  private _map: { [key: number]: ILinkedListNode<T> } = {};
```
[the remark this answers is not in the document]
Tyriar: Even so, I think the numbers above fro LRUMap.get prove there's a pretty big impact in perf when moving to `number`

Tyriar:
@bgw thanks for the review! I marked all the conversations as resolved, let me know if you have more feedback (and feel free to unresolve if you don't think they are 😃).

src/renderer/BaseRenderLayer.ts
```diff
@@ -22,6 +22,19 @@ export abstract class BaseRenderLayer implements IRenderLayer {
 
   protected _charAtlas: BaseCharAtlas;
 
+  /**
+   * An object that's reused when drawing glyphs in order to reduce GC.
+   */
+  private _currentGlyphIdentifier: IGlyphIdentifier = {
```
bgw: I like this solution.

src/renderer/atlas/DynamicCharAtlas.ts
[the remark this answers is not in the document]
bgw: > Well we have `chars` already if we need it, can you clarify what for?

`chars` doesn't store the flag or color information.

The problem with a number type is that it restricts how many bits of information we could store in the future, though I do appreciate that the logic is mostly encapsulated.

`String.fromCharCode` would let us encode an arbitrary number of 16-bit integers. It's nearly as fast, and if we're assuming that LRUMap will cause an implicit coercion to a string later, it might even be cheaper: http://jsben.ch/yorNP

I think this solution is fine for now though. We can consider `String.fromCharCode` when we run out of space.

src/renderer/atlas/LRUMap.ts
```diff
@@ -6,12 +6,12 @@
 interface ILinkedListNode<T> {
   prev: ILinkedListNode<T>;
   next: ILinkedListNode<T>;
-  key: string;
+  key: number;
```
bgw: Can we make this `number | string`?

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -230,13 +282,73 @@ export default class DynamicCharAtlas extends BaseCharAtlas {
...
+  private _generateBitmap(): void {
+    const countAtGeneration = this._glyphsWaitingOnBitmapCount;
+    window.createImageBitmap(this._cacheCanvas).then(bitmap => {
+      // Set bitmap
+      this._bitmap = bitmap;
+
+      // Mark all new glyphs as in bitmap
+      for (let i = 0; i < countAtGeneration; i++) {
+        const key = this._glyphsWaitingOnBitmapQueue[i];
+        const value = this._cacheMap.peekValue(key);
```
bgw: Isn't it more expensive and complicated to:

1. Store an array of integers (with a typed array).
2. Look up an element from that array, and then look up it's value in a map.
3. Manually resize the array when you run out of space instead of letting the runtime do it with an otherwise slower array type.

Than to just:

1. Store an array of objects (basically an array of pointers).
2. Look up an element from that array.

You do incur some cost with the second approach because you're retaining references to the `IGlyphCacheValue`s a little longer, but it's only for a few frames. Occasionally updating `inBitmap` on a stale object shouldn't be very expensive either.

It seems like you're really going out of your way to avoid an `Array`, but I don't think it's worth it, because a simple Array would save you a ton of other operations.

src/renderer/atlas/DynamicCharAtlas.ts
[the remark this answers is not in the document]
Tyriar: Seems like a good direction 👍

src/renderer/atlas/DynamicCharAtlas.ts
Tyriar: I'll see if I can measure a perf impact of this and go with your suggestion if there's no obvious benefit. My thinking here is avoiding Array due to holey array deoptimizations and GC involved in their dynamic resize.

bgw: I'm more concerned about this in terms of code quality than I am in terms of performance. Using a typed array for this seems to add complexity, and I suspect it's slower anyways, so I'd like to just use a simple array.

> holey array deoptimizations

But the queue doesn't have any holes.

> GC involved in their dynamic resize.

For growing the array: You're paying that anyways when you manually grow the array.

When you wipe the array: It's probably faster (and cleaner) to let the whole array get collected and construct a new one than it is to zero it out.

Tyriar: > When you wipe the array: It's probably faster (and cleaner) to let the whole array get collected and construct a new one than it is to zero it out.

It's not zero'd out, it contains stale data at the end as we track the size of it in a number.

Also you've convinced me 😄

src/renderer/atlas/LRUMap.ts
Tyriar: I tried to do that but TS complained. Created https://github.com/Microsoft/TypeScript/issues/27219 to see their answer

Tyriar:
Ready for another round 😃

src/renderer/atlas/DynamicCharAtlas.ts
Tyriar: I looked into perf of that function and the amount of time spent in that function is so tiny compared to the rest of the code, classic case of premature optimization 😅

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -230,13 +272,59 @@ export default class DynamicCharAtlas extends BaseCharAtlas {
...
+  private _generateBitmap(): void {
+    let countAtGeneration = this._glyphsWaitingOnBitmapCount;
+    window.createImageBitmap(this._cacheCanvas).then(bitmap => {
+      // Set bitmap
+      this._bitmap = bitmap;
+
+      // Mark all new glyphs as in bitmap, excluding glyphs that came in after
+      // the bitmap was requested
+      while (countAtGeneration-- > 0) {
+        const value = this._glyphsWaitingOnBitmapQueue[0];
```
bgw: Shouldn't this be `this._glyphsWaitingOnBitmapQueue[countAtGeneration];`?

Also: It might be slightly cleaner to write this as a simple for loop.

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -230,13 +272,59 @@ export default class DynamicCharAtlas extends BaseCharAtlas {
...
+        // If the value has already been evicted, do nothing
+        if (value) {
```
bgw: I don't think this is nullable anymore. It's possible that the object is no longer in the LRUMap, but it's still fine to update it. It just means we'll hold onto some objects slightly longer, and their GC is slightly delayed.

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -57,6 +73,18 @@ export default class DynamicCharAtlas extends BaseCharAtlas {
 
   private _drawToCacheCount: number = 0;
 
+  // An array of glyph keys that are waiting on the bitmap to be generated.
+  private _glyphsWaitingOnBitmapQueue: IGlyphCacheValue[] = [];
+
+  // The number of glyphs keys waiting on the bitmap to be generated.
+  private _glyphsWaitingOnBitmapCount: number = 0;
```
bgw: Do we need to keep a separate counter from `_glyphsWaitingOnBitmapQueue.length`?

src/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -230,13 +272,59 @@ export default class DynamicCharAtlas extends BaseCharAtlas {
...
+
+      // Mark all new glyphs as in bitmap, excluding glyphs that came in after
+      // the bitmap was requested
+      while (countAtGeneration-- > 0) {
+        const value = this._glyphsWaitingOnBitmapQueue[0];
+        // If the value has already been evicted, do nothing
+        if (value) {
+          value.inBitmap = true;
+        }
+      }
```
bgw: I think you need to slice/splice `this._glyphsWaitingOnBitmapQueue` to remove the elements we just added to the bitmap.

src/renderer/atlas/DynamicCharAtlas.ts
Tyriar: Opps, this was meant to be a `shift`

src/renderer/atlas/DynamicCharAtlas.ts
Tyriar: I was planning on removing elements from the list using shift.

src/renderer/atlas/DynamicCharAtlas.ts
Tyriar: You're right, I can just grab the length at the time the bitmap is created

src/renderer/atlas/DynamicCharAtlas.ts
Tyriar: Changing the comment

src/renderer/atlas/DynamicCharAtlas.ts
bgw: A shift inside the loop is going to be O(n^2). An iteration and slice is O(n), so I'd recommend the later, especially since it's no less comprehensible.

That said, I also know that this function isn't called frequently, so I guess it doesn't matter too much.

Tyriar: @bgw well I assume JS engines have optimizations that make it such that it's not O(n^2), but I'll change it anyway as it's an assumption 😃

bgw:
LGTM, thanks. Did you test it with a small atlas size to verify that there's no obvious bugs when the atlas gets thrashed?

Tyriar:
@bgw yes, no errors when I went down to a 100x100 texture.

PR #1695, opened 2018-09-15
Fix coveralls integration

Tyriar:


Tyriar:
This is ready for review, looks like it's working. Check the bottom of https://coveralls.io/github/xtermjs/xterm.js

jerch:
@Tyriar Any idea why some of the commit messages show weird escaped html strings?

Tyriar:
@jerch hm, I didn't see that when I was testing it, I suspect it's a bug on the website.

Tyriar:
The next one worked:

screen shot 2018-09-21 at 8 16 20 am

PR #1698, opened 2018-09-17
Fix scroll APIs not affecting scroll bar

### LINKED ISSUE #1697: Scroll commands don't affect the scroll bar
They move the buffer ydisp but not the scroll bar.

Repro:

1. Fill buffer
2. `term.scrollLines(-10)`

Related: https://github.com/xtermjs/xterm.js/issues/291

Tyriar:
Fixes #1697
Related Microsoft/vscode#58532

---

While optimizing viewport layouts in https://github.com/xtermjs/xterm.js/commit/2127128ef58be350137fcf572e9b482342b56130 an edge case was missed, it looks like we have to query `viewportElement.scrollTop` in `syncScrollArea` in order to make sure that the scroll position of the scroll bar is correct.

### DIFF
--- a/src/Viewport.ts
+++ b/src/Viewport.ts
@@ -22,6 +22,7 @@ export class Viewport extends Disposable implements IViewport {
   private _lastRecordedViewportHeight: number = 0;
   private _lastRecordedBufferHeight: number = 0;
   private _lastTouchY: number;
+  private _lastScrollTop: number = 0;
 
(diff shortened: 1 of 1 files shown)


Tyriar:
I'll leave this for 3.8.0 as we're only a couple of weeks away from that anyway and no one has reported the issue against xterm.js

PR #1699, opened 2018-09-17
Enable strict null checks and no DOM access in common and core

Tyriar:
Part of #1507
Part of #1319

---

This change enables strict mode and prevents DOM access in `./src/common` and `./src/core` by adding a noEmit tsconfig.json that gets built as the `pretest` npm script task. 🍾


mofux:
Haven't tested, but LGTM in general 👍

PR #1703, opened 2018-09-20
Speculative fix for NPE in linkifier

### LINKED ISSUE #1702: TypeError: Cannot read property 'isWrapped' of undefined at Buffer.getWrappedRangeForLine
image

Tyriar:
Fixes #1702

---

I was not able to repro this in the demo, but I think this will fix it.

### DIFF
--- a/src/Linkifier.test.ts
+++ b/src/Linkifier.test.ts
@@ -42,6 +42,7 @@ describe('Linkifier', () => {
   beforeEach(() => {
     terminal = new MockTerminal();
     terminal.cols = 100;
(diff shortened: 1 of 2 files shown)


jerch:
@Tyriar Any idea why `.end` can contain a stale value? Is there a caller that does not get informed about resize events?

Tyriar:
@jerch my suspicion is it's happening because .end is only cleared at the end of that function, so if the resize happens during the `setTimeout`'s delay it could repro. I tried to reproduce in the demo by increasing the delay but couldn't.

PR #1705, opened 2018-09-21
Support link underlines in the DOM renderer

### LINKED ISSUE #1704: Support link underlines in the DOM renderer
Currently the DOM renderer doesn't support link underlines: 

https://github.com/xtermjs/xterm.js/blob/290c95ad71f8b1264139a7ed6f070c4741a8987e/typings/xterm.d.ts#L151

/cc @agurodriguez

agurodriguez:
Closes #1704

### DIFF
--- a/src/renderer/dom/DomRenderer.ts
+++ b/src/renderer/dom/DomRenderer.ts
@@ -10,7 +10,6 @@ import { EventEmitter } from '../../common/EventEmitter';
 import { ColorManager } from '../ColorManager';
 import { RenderDebouncer } from '../../ui/RenderDebouncer';
 import { BOLD_CLASS, ITALIC_CLASS, CURSOR_CLASS, CURSOR_STYLE_BLOCK_CLASS, CURSOR_STYLE_BAR_CLASS, CURSOR_STYLE_UNDERLINE_CLASS, DomRendererRowFactory } from './DomRendererRowFactory';
-import { INVERTED_DEFAULT_COLOR } from '../atlas/Types';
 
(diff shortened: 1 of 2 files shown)


Tyriar:
> I believe we should consider the border-bottom approach in this case. But of course, you've the last word 😊

I don't actually see that AA issue in Chrome 69, is that the browser you're using? I would prefer to use text-decoration as it simplified the logic a lot plus gives the nice perk of fancy underlines:

screen shot 2018-09-21 at 7 15 38 am

I removed `box-sizing: border-box`, are you still able to reproduce the AA problem on your PC?

agurodriguez:
I was testing on vscode itself. But it also happens on my Chrome:

screen shot 2018-09-21 at 11 56 36 am

I noticed also, that if you zoom the image you posted just before the gaps are also there:

screen shot 2018-09-21 at 11 56 23 am

agurodriguez:
Interesting fact: The AA problem disappears if the `span` width is a number without decimals (By default it computes to 7.234375px in my environment when using `font-size` 12).

screen shot 2018-09-21 at 12 28 15 pm

Tyriar:
Lol, it was there on mine, I didn't even notice. I suspect this issue isn't present on lower DPI screens and I say we go ahead with this as is as it's barely noticeable. Thoughts @agurodriguez?

Also, the character width has to be a decimal for the DOM renderer, this uses the width the font specifies.

agurodriguez:
Yeahp, I do agree it's barely noticeable. Let's go ahead.

PR #1711, opened 2018-09-27
Add new use case: Juno

pfitzseb:


pfitzseb:
Thanks for the quick merge!

PR #1714, opened 2018-09-28
multiple unicode version support

### LINKED ISSUE #1709: Unicode handling in xterm.js
Coming from #1707 it seems the correct unicode handling is more and more an issue for people due to emojis. Since we all love emojis this should get fixed ASAP :smile: 

Proposal:
Create a provider for different unicode versions, that is capable of hiding the version specific data and implementations behind a nice API. Currently we only need version dependent implementations for wcwidth, so a rough sketchup could look like this:

```TS
interface IUnicodeProvider {
  supportedVersions(): string[];
  getVersion(): string;
  setVersion(version?: string);  // version optional for fallback behavior
  wcwidth(ucs: number): number;
  getStringCellWidth(s: string): number;
  ... // more to come with support of other unicode features
}
```
Ideally the provider is self containing, thus the terminal just needs to deal with the interface methods and updates the version/locale when needed. The provider would have to deal with the low level stuff to provide the correct data sets so the methods just work as expected for a supported version.
Within the provider we then can decide whether the data is provided statically in the code base or even tries to create the data on the fly. First will have quite an impact on xterm.js' size, the second will raise async questions (remember - most of the core parts are synchronous atm). The whole unicode stuff could also be bundled into some addon like feature for version XY.

Up for discussion.
/cc @Tyriar, @bgw, @mofux, @dnfield

jerch:
First version of multiple unicode version support. Shall fix #1709, partly copied over from #1707 (thx to @dnfield).

The class `UnicodeProvider` currently holds registered versions at the global object. An instance of that class carries the active version info for a single terminal instance and returns the correct `wcwidth` impl. The statically provided versions currently reside under `unicode` in the codebase and get added in `UnicodeProvider.ts`. More versions can be added at runtime (maybe in the future by addons), there is a callback slot `onRegister` to get notified about newly registered versions.

Still very hacky, would be good to get some conceptual feedback first before doing tests and such.

Edit: Currently adds ~20kB to the package size uncompressed (now at ~385kB).

/cc @Tyriar, @mofux, @bgw, @dnfield


src/UnicodeProvider.ts
```diff
@@ -0,0 +1,147 @@
...
+   */
+  static registerVersion(impl: IUnicodeImplementation): void {
+    if (UnicodeProvider.versions[impl.version]) {
+      throw new Error(`unicode version "${impl.version}" already registered`);
+    }
+    UnicodeProvider.versions[impl.version] = impl;
+    UnicodeProvider._registerCallbacks.forEach(cb => cb(impl.version));
+  }
+
+  static registeredVersions(): number[] {
```
mofux: IMO we should rename the function to `getRegisteredVersions()`, the current name feels like a property name, but it is a function.

jerch:
@mofux: Thx! Im abit concerned about the package size, might try to get the new data in a more packed representation (The old one isnt even half as big, also zips further down).

src/UnicodeProvider.ts
jerch: Yepp gonna change it.

jerch:
Reminder to myself: A provider instance needs to be disposable from the callback register.

mofux:
I wouldn't bee too concerned about the extra 32KB 😅

jerch:
Yeah, also zipped the difference is 7.8 vs 2.5 kB, imho nothing to worry about.

jerch:
Made the provider disposable and added some tests.

Ready for serious reviews :smile_cat:

src/UnicodeProvider.ts
```diff
@@ -0,0 +1,188 @@
...
+    if (pos !== -1) {
+      UnicodeProvider.removeRegisterListener(this._registerCallbacks[pos][1]);
+      this._registerCallbacks.splice(pos, 1);
+    }
+  }
+
+  /**
+   * Get a list of currently registered unicode versions.
+   */
+  public getRegisteredVersions(): number[] {
```
Tyriar: I saw @mofux's early comment, maybe this is even better?

```ts
public static get registeredVersions(): number[]
```

src/UnicodeProvider.ts
```diff
@@ -0,0 +1,188 @@
...
+    return Object.getOwnPropertyNames(UnicodeProvider.versions).map(parseFloat).sort((a, b) => a - b);
+  }
+
+  /**
+   * Get the currently active unicode version.
+   */
+  public getActiveVersion(): number {
```
Tyriar:
```ts
get activeVersion(): number
```

src/UnicodeProvider.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  /**
+   * Activate a registered unicode version. By default the closest version will be activated
+   * (can be higher or lower). Setting `mode` to 'next' tries to get at least that version,
+   * 'previous' tries to get the closest lower version.
+   * Unless there is no version registered this method will always succeed.
+   * Returns the activated version number.
+   */
+  public setActiveVersion(version: number, mode?: 'exact' | 'closest' | 'next' | 'previous'): number {
```
Tyriar:
```ts
set activeVersion(..): number
```

Tyriar: Is `mode` needed? Would it not be better to just throw if the version wasn't there?

demo/client.ts
```diff
@@ -219,7 +220,9 @@ function initOptions(term: TerminalType): void {
         booleanOptions.push(o);
         break;
       case 'number':
-        numberOptions.push(o);
+        if (o !== 'unicodeVersion') {
```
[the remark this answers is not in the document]
jerch: This is because the option parsing defaults to a single input field for numbers, but I want the version numbers in a dropdown.

src/Linkifier.ts
```diff
@@ -256,7 +255,7 @@ export class Linkifier extends EventEmitter implements ILinkifier {
    * @param fg The link color for hover event.
    */
   private _addLink(x: number, y: number, uri: string, matcher: ILinkMatcher, fg: number): void {
-    const width = getStringCellWidth(uri);
+    const width = (this._terminal as any).unicodeProvider.getStringCellWidth(uri);
```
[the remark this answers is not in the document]
jerch: Oops...
Well this has to be replaced anyway by the `endIndex` calc as done in #1769 to support live version switching.

src/UnicodeProvider.ts
```diff
@@ -0,0 +1,188 @@
+/**
+ * Copyright (c) 2018 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+import { IUnicodeImplementation, IUnicodeProvider } from './Types';
+import { v6 } from './unicode/v6';
```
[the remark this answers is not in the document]
jerch: Yepp.

src/UnicodeProvider.ts
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * The version related implementations are stored statically
+ * to avoid recreating them for every single instance.
+ *
+ * An instance of this class is meant to serve unicode specific implementations
+ * for a single terminal instance. This way multiple terminals can have
+ * different unicode settings active while still referring to the
+ * same underlying implementations.
+ */
+export class UnicodeProvider extends Disposable implements IUnicodeProvider {
```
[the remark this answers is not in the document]
jerch: Ok gonna rename it.

src/UnicodeProvider.ts
```diff
@@ -0,0 +1,188 @@
...
+  public static versions: {[key: string]: IUnicodeImplementation} = {};
```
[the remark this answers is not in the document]
jerch: Ok.

src/UnicodeProvider.test.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    it('add version', function(): void {
+      UnicodeProvider.registerVersion(VERSION_DUMMY1);
+      assert.deepEqual(UnicodeProvider.getRegisteredVersions(), [6, 11, 15]);
+      delete UnicodeProvider.versions[15];
+    });
+
+    it('register callback', function(): void {
+      UnicodeProvider.addRegisterListener((version) => {
+        assert.equal(version, 15);
```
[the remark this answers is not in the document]
jerch: Oops, yeah the callback had the `done()` call before.

src/UnicodeProvider.ts
```diff
@@ -0,0 +1,188 @@
...
+    this._registerCallbacks = null;
+    this.wcwidth = null;
+  }
+
+  /**
+   * Callback to run when a version got registered.
+   * Gets the newly registered version and
+   * the `UnicodeProvider` instance as arguments.
+   */
+  public addRegisterListener(callback: (version: number, provider: UnicodeProvider) => void): void {
```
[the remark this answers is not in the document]
jerch: They are in preparation for adding unicode versions at runtime by addons or such. The listener will inform about new versions, so any terminal instance can switch if needed.

src/UnicodeProvider.ts
jerch: Yes can do those as real getter/setter (since they are hardly used they wont show their bad perf).

src/UnicodeProvider.ts
jerch: Same as above.

src/UnicodeProvider.ts
jerch: Same.

jerch: Guess thats kinda overengineered, changed it to a simple getter/setter.

jerch:
Up for next review.

jerch:
Ah thats weird, have to check if there is some high char hidden in the test, that might change width with a different unicode version.
And yes, I think we should wait with this after #1789.

jerch:
Currently blocked by #1789.

jerch:
@dnfield May I ask you to have a look at the new code, in particular here:
https://github.com/xtermjs/xterm.js/blob/a1257a1184ad9d79a8db1174c4259978ba9b8c9f/src/unicode/v11.ts#L335

As you can see I extracted the BMP plane codepoints from the others to create the lookup table from it and shortcut the bisect for higher abit too (no magic here, I have simply split your lists).

The problem: A closer look at your code revealed that you are not using the `WIDE_EASTASIAN_BMP` definitions from your extracted list, instead you used the same values from the v6 definition. Did you do this on purpose? Since the defintions for v6 and v11 are very different I am unsure which one to go with.

dnfield:
I didn't have any good test cases for the wide East Asian, and so I just left it as is, which I suppose is incorrect.

jerch:
@dnfield
Aww ok yeah same here, guess we need to get some test against a system with v11 up before we can decide this.
Thx for the fast response. :heart:

jerch:
The new iOS release and Mojave claim to be Unicode v11, Ubuntu still ships ICU v60 in 18.10 which is still v10. Ubuntu 19.04 has ICU v63 linked which is v11. Not sure if the latter is already useable, guess the tests have to be done on Mojave.

So I did a quick test in Mojave with the first codepoint that differs in v6 and v11: U+231A (watch)
their clib `wcwidth` returns 1 while the char is taking 2 cells in Terminal.app. Wth? Seems they use some other method to get the width. Does anyone know how get the correct width under macOS?
Btw Ubuntu 18.04 with unicode v10 returns 2. What a mess...

Tyriar:
🤷‍♂️

jerch:
Hmm it seems macOS' `wcwidth` is not updated/maintained. Only chance I see to get some numbers is to draw a char and get its bounding box. But then the numbers rely on the font glyphs and their relation/support for a specific unicode version @gnachman Maybe you know a way how to get wcwidth numbers more reliable on macOS?

Guess I gonna start with v10 on ubuntu to check which wide definitions can be pulled in.

Tyriar:
I had an idea yesterday which would probably be a good fix for this. Once we have addons, the unicode definitions can be "bundled" addons, so they will remain in xtermjs/xterm.js and ship in the `xterm` module, but loading them will be controlled via a new Terminal option. For example:

```ts
// Loads the default unicode addon version using dynamic imports such
// that all unicode definitions are not loaded when a terminal is created
const t1 = new Terminal();

// Loads only the v11 unicode addon via dynamic import
const t2 = new Terminal({
  unicodeVersion: 11
});
```

jerch:
@Tyriar Stumbled myself over the question what to do with at least 3 versions to support. For local applications (like electron based) 50kB more or less dont really hurt, for remote browser based scenarios this is nasty, since a user only needs one particular version.

My idea was to offer 2 ways to integrate the data:
- For local apps make the definition addons able to be prebundled (like your idea), this could also be used by browser solutions for an initial definition to start with or to include all if they want to go with a fat bundle.
- Optionally make other definition addons lazy loadable from some ressource if a user requests it. This would only need a small fetch callback that loads the new definition into the provider manager, prolly with a lazy unicode version loader addon.

Tyriar:
Closing as the eventual addon solution will look pretty different to this, see https://github.com/xtermjs/xterm.js/issues/1709#issuecomment-491335698

PR #1716, opened 2018-09-30
Get correct col/row coords from mouse event

Tyriar:
Before the y from getCoords could be -1, a ceil call was also causing
a single pixel offset in each dimension

Related Microsoft/vscode#50128


Tyriar:
I checked in a change to the tests to get the expected behavior but it still appears to be failing because `event.pageY - element.offsetTop` appears to equal `-1` for some reason. Does anyone know what's going on here?

Tyriar:
Can't figure it out, may revisit later.

PR #1724, opened 2018-10-03
fix font scaling for non dpi-aware browsers

dgriffen:
Certain browser scenarios (such as IE11 when hosted in WPF) are not fully DPI-aware. This creates odd scaling issues where the terminal font is far too small on HiDPI displays. This is due to the fact that as of version 3.0 font sizes were specified in pixel amounts instead of point sizes. This PR fixes the issue by changing the font sizes to use point instead of pixel. This does change the semantics slightly because px and pt are not the same scale but I would expect that using pt is actually more in line with what users expect.

Here is an example of the issue, the terminal window is at the bottom:
bad-scaling

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -251,7 +251,9 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
 
   private _setup(): void {
     Object.keys(DEFAULT_OPTIONS).forEach((key) => {
-      if (this.options[key] === null || this.options[key] === undefined) {
+      if (key === 'fontSize' && this.options[key] === null) {
+        return;
+      } else if (this.options[key] === null || this.options[key] === undefined) {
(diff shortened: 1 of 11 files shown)


Tyriar:
This would actually break the API since 12pt = 16px, any consumers setting the font will get extra large text if they pull in this change.

What if we allow fontSize to be a `number | string`, number is always `px` and a string needs to have a postfix of `em`, `pt`, `px` or `%`? I'm not sure if em and % would work when drawing to the canvas without running `window.getComputedStyle` to pull the font size from the container?

dgriffen:
That sounds like a good way to preserve the old behavior while allowing the new one. I'll start making those changes. I can test to see what happens with em and % in the canvas renderer since I have the repro scenario.

Tyriar:
Closing as the PR is a bit stale.

PR #1725, opened 2018-10-03
Centered logo and badges on the top of the README

gebeto:
Make it for prettier look :)


Tyriar:
Thanks for the PR but I'd prefer to keep it left aligned so the readme is simpler

PR #1731, opened 2018-10-08
first attempt to recycle buffer lines

jerch:
A first attempt to recycle buffer lines.

Can be tested by switching `experimentalBufferLineImpl` to `TypedArray`.

~Note: The recycling is only active for the typed array version to make it a fair comparison (the JS array version is much slower with it). Still buggy with faulty behavior in some tests (cannot be tested easily atm, the tests currently rely on the hardcoded JS array type). More to come...~

The recycling is now active for both buffer line versions.

### DIFF
--- a/src/BufferLine.ts
+++ b/src/BufferLine.ts
@@ -94,11 +94,10 @@ export class BufferLine implements IBufferLine {
     }
   }
 
-  public copyFrom(line: BufferLine): IBufferLine {
+  public copyFrom(line: BufferLine): void {
     this._data = line._data.slice(0);
     this.length = line.length;
     this.isWrapped = line.isWrapped;
-    return this;
   }
 
   public clone(): IBufferLine {
@@ -249,7 +248,7 @@ export class BufferLineTypedArray implements IBufferLine {
   }
 
   /** alter to a full copy of line  */
-  public copyFrom(line: BufferLineTypedArray): IBufferLine {
+  public copyFrom(line: BufferLineTypedArray): void {
     if (this.length !== line.length) {
       this._data = new Uint32Array(line._data);
     } else {
@@ -262,7 +261,6 @@ export class BufferLineTypedArray implements IBufferLine {
       this._combined[el] = line._combined[el];
     }
     this.isWrapped = line.isWrapped;
-    return this;
   }
 
   /** create a new clone */
--- a/src/BufferLine.ts
+++ b/src/BufferLine.ts
@@ -94,13 +94,11 @@ export class BufferLine implements IBufferLine {
     }
   }
 
-  public copyFrom(line: IBufferLine): void {
-    this._data = [];
-    for (let i = 0; i < line.length; ++i) {
-      this._push(line.get(i));
-    }
+  public copyFrom(line: BufferLine): IBufferLine {
+    this._data = line._data.slice(0);
     this.length = line.length;
     this.isWrapped = line.isWrapped;
+    return this;
   }
 
   public clone(): IBufferLine {
@@ -249,7 +247,7 @@ export class BufferLineTypedArray implements IBufferLine {
   }
 
   /** alter to a full copy of line  */
-  public copyFrom(line: BufferLineTypedArray): void {
+  public copyFrom(line: BufferLineTypedArray): IBufferLine {
     if (this.length !== line.length) {
       this._data = new Uint32Array(line._data);
     } else {
@@ -262,6 +260,7 @@ export class BufferLineTypedArray implements IBufferLine {
       this._combined[el] = line._combined[el];
     }
     this.isWrapped = line.isWrapped;
+    return this;
   }
 
   /** create a new clone */
--- a/src/BufferLine.ts
+++ b/src/BufferLine.ts
@@ -77,15 +77,15 @@ export class BufferLine implements IBufferLine {
 
   /** resize line to cols filling new cells with fill */
   public resize(cols: number, fillCharData: CharData, shrink: boolean = false): void {
+    while (this._data.length < cols) {
+      this._data.push(fillCharData);
+    }
     if (shrink) {
       while (this._data.length > cols) {
         this._data.pop();
       }
     }
-    while (this._data.length < cols) {
-      this._data.push(fillCharData);
-    }
-    this.length = cols;
+    this.length = this._data.length;
   }
 
   public fill(fillCharData: CharData): void {
@@ -139,11 +139,13 @@ export class BufferLineTypedArray implements IBufferLine {
     if (!fillCharData) {
       fillCharData = [0, NULL_CELL_CHAR, NULL_CELL_WIDTH, NULL_CELL_CODE];
     }
-    this._data = new Uint32Array(cols * CELL_SIZE);
-    for (let i = 0; i < cols; ++i) {
-      this.set(i, fillCharData);
+    if (cols) {
+      this._data = new Uint32Array(cols * CELL_SIZE);
+      for (let i = 0; i < cols; ++i) {
+        this.set(i, fillCharData);
+      }
     }
-    this.length = cols || 0;
+    this.length = cols;
   }
 
   public get(index: number): CharData {
@@ -210,12 +212,12 @@ export class BufferLineTypedArray implements IBufferLine {
   }
 
   public resize(cols: number, fillCharData: CharData, shrink: boolean = false): void {
-    if (cols === this.length) {
+    if (cols === this.length || (!shrink && cols < this.length)) {
       return;
     }
     if (cols > this.length) {
       const data = new Uint32Array(cols * CELL_SIZE);
-      if (this._data) {
+      if (this.length) {
         if (cols * CELL_SIZE < this._data.length) {
           data.set(this._data.subarray(0, cols * CELL_SIZE));
         } else {
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -107,7 +107,7 @@ const DEFAULT_OPTIONS: ITerminalOptions = {
   rightClickSelectsWord: Browser.isMac,
   rendererType: 'canvas',
   experimentalBufferLineImpl: 'JsArray',
-  experimentalPushRecycling: false
+  experimentalBufferLineRecycling: false
 };
 
 export class Terminal extends EventEmitter implements ITerminal, IDisposable, IInputHandlingTerminal {
@@ -1179,16 +1179,16 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
    * Scroll the terminal down 1 row, creating a blank line.
    * @param isWrapped Whether the new line is wrapped from the previous line.
    */
-  public scroll(isWrapped?: boolean): void {
+  public scroll(isWrapped: boolean = false): void {
     let newLine: IBufferLine;
-    const useRecycling = this.options.experimentalPushRecycling;
+    const useRecycling = this.options.experimentalBufferLineRecycling;
     if (useRecycling) {
       newLine = this._blankLine;
       if (!newLine || newLine.length !== this.cols || newLine.get(0)[CHAR_DATA_ATTR_INDEX] !== this.eraseAttr()) {
         newLine = this.buffer.getBlankLine(this.eraseAttr(), isWrapped);
         this._blankLine = newLine;
       }
-      newLine.isWrapped = !!(isWrapped);
+      newLine.isWrapped = isWrapped;
     } else {
       newLine = this.buffer.getBlankLine(this.eraseAttr(), isWrapped);
     }
@@ -1198,15 +1198,17 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
 
     if (this.buffer.scrollTop === 0) {
       // Determine whether the buffer is going to be trimmed after insertion.
-      const willBufferBeTrimmed = this.buffer.lines.length === this.buffer.lines.maxLength;
+      const willBufferBeTrimmed = this.buffer.lines.pushWouldTrim();
 
       // Insert the line using the fastest method
       if (bottomRow === this.buffer.lines.length - 1) {
         if (useRecycling) {
           if (willBufferBeTrimmed) {
-            // Warning: Never call .trimAndRecycle() without the
-            //          willBufferBeTrimmed guard!
-            this.buffer.lines.trimAndRecycle().copyFrom(newLine);
+            // push would trim the oldest line in the ringbuffer
+            // therefore we can recycle it here as the new line
+            const recycled = this.buffer.lines.get(0);
+            recycled.copyFrom(newLine);
+            this.buffer.lines.push(recycled);
           } else {
             this.buffer.lines.push(newLine.clone());
           }
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -346,7 +346,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
   }
 
   public get isFocused(): boolean {
-    return document.activeElement === this.textarea;
+    return document.activeElement === this.textarea && document.hasFocus();
   }
 
   /**
@@ -1179,6 +1179,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
    * @param isWrapped Whether the new line is wrapped from the previous line.
    */
   public scroll(isWrapped?: boolean): void {
+<<<<<<< HEAD
     let newLine: IBufferLine;
     const useRecycling = this.options.experimentalPushRecycling;
     if (useRecycling) {
@@ -1192,6 +1193,9 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
       newLine = this.buffer.getBlankLine(DEFAULT_ATTR, isWrapped);
     }
 
+=======
+    const newLine = this.buffer.getBlankLine(this.eraseAttr(), isWrapped);
+>>>>>>> master
     const topRow = this.buffer.ybase + this.buffer.scrollTop;
     const bottomRow = this.buffer.ybase + this.buffer.scrollBottom;
 
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -477,6 +477,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
         if (this._theme) {
           this.renderer.setTheme(this._theme);
         }
+        this.mouseHelper.setRenderer(this.renderer);
         break;
       case 'scrollback':
         this.buffers.resize(this.cols, this.rows);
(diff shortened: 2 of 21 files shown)


src/Buffer.ts
```diff
@@ -55,10 +55,13 @@ export class Buffer implements IBuffer {
   }
 
   public setBufferLineFactory(type: string): void {
+    this.lines.recycling = false;
+    this.lines.recycleHeap = [];
```
Tyriar: Instead of keeping a recycle array pool around, I was thinking more just handle things at a lower level in critical functions. I think `Terminal.scroll` is the main function we need to optimize, so changing this:

https://github.com/xtermjs/xterm.js/blob/8d16bb6ea2e90312b0efb28bee3f39d3995a906f/src/Terminal.ts#L1182-L1190

To something like this:

```ts
      if (willBufferBeTrimmed) {
        // NEW IMPL
        this.buffer.lines.trimStart(1);
        this.buffer.lines.length++; // Fix up what trimStart broke
        const lineToFix = this.buffer.lines.get(0);
        // blah
      } else {
        // OLD IMPL
        // Insert the line using the fastest method
        if (bottomRow === this.buffer.lines.length - 1) {
          this.buffer.lines.push(newLine);
        } else {
          this.buffer.lines.splice(bottomRow + 1, 0, newLine);
        }
      }
```

So instead of pushing and allowing the trimming mechanism to kick in, we handle it manually by incrementing startIndex and then change lines[0].

jerch:
@Tyriar Here comes the new approach: a callback version of push. It was the fastest thing I could come up with. I first tried your trim idea with another method, but it was much worse in runtime.

Still up for other ideas.

NB: I kinda screwed up the branch and had to reset it lol.

jerch:
Added an `experimentalPushRecycling` option, this way it can be tested in the demo with both bufferline implementations. Also made a small change to `BufferLine.copyFrom`, the JSArray version now profits from the recycling, too.

For my typical benchmark `ls -lR /usr/lib` I see the following runtime number (range from 5 runs):

- no recycling:
  - JSArray: 1900 - 2300 ms
  - TypedArray: 2200 - 2600 ms

- with recycling:
  - JSArray: 1600 - 2000 ms
  - TypedArray: 1700 - 2100 ms

With that small change in `BufferLine.copyFrom` the JSArray version is again the fastest. The TypedArray version show a greater benefit from recycling and is close behind.

jerch:
Here are the numbers for the input chain alone (done with this script https://gist.github.com/jerch/31f23538c5ca1a5079a78bbd627398ce, ./benchmark_data1 contains the output of `ls -lR /usr/lib`):
```
{ BufferLineType: 'JsArray',
  Recycling: false,
  Throughput: '10.45 MB/s',
  File: './benchmark_data1',
  Duration: 4596,
  Size: 50361113 }
{ BufferLineType: 'TypedArray',
  Recycling: false,
  Throughput: '11.44 MB/s',
  File: './benchmark_data1',
  Duration: 4199,
  Size: 50361113 }
{ BufferLineType: 'JsArray',
  Recycling: true,
  Throughput: '12.04 MB/s',
  File: './benchmark_data1',
  Duration: 3990,
  Size: 50361113 }
{ BufferLineType: 'TypedArray',
  Recycling: true,
  Throughput: '19.15 MB/s',
  File: './benchmark_data1',
  Duration: 2508,
  Size: 50361113 }
```
Seems the typed array + recycling doubles the throughput :smile:

src/common/CircularList.ts
```diff
@@ -100,6 +100,24 @@ export class CircularList<T> extends EventEmitter implements ICircularList<T> {
...
+   * Return the new value from the callback.
+   */
+  public pushRecycling(callback: (item: T | undefined) => T): void {
+    this._array[this._getCyclicIndex(this._length)] = callback(this._array[this._getCyclicIndex(this._length)]);
+    if (this._length === this._maxLength) {
+      this._startIndex++;
+      if (this._startIndex === this._maxLength) {
+        this._startIndex = 0;
+      }
+      this.emit('trim', 1);
```
Tyriar: This may need to be fired before the recycling happens?

src/common/CircularList.ts
```diff
@@ -100,6 +100,24 @@ export class CircularList<T> extends EventEmitter implements ICircularList<T> {
     }
   }
 
+  /**
+   * Recycling push variant with a callback.
+   * The callback gets the value at the current position to be overwritten.
+   * Return the new value from the callback.
+   */
+  public pushRecycling(callback: (item: T | undefined) => T): void {
```
Tyriar: Not a fan of a callback on a sync function, why not just return `this._array[this._getCyclicIndex(this._length)]`?

src/Terminal.ts
```diff
@@ -1184,9 +1201,13 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
 
       // Insert the line using the fastest method
       if (bottomRow === this.buffer.lines.length - 1) {
-        this.buffer.lines.push(newLine);
+        if (useRecycling) {
+          this.buffer.lines.pushRecycling((item) => (item) ? item.copyFrom(newLine) : newLine.clone());
```
Tyriar: We actually want to fill all cells the same, so could we avoid creating newLine here and instead pass in a single cell/numbers?

src/Terminal.ts
```diff
@@ -1184,9 +1201,13 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
...
+        } else {
+          this.buffer.lines.push(newLine);
+        }
       } else {
-        this.buffer.lines.splice(bottomRow + 1, 0, newLine);
+        this.buffer.lines.splice(bottomRow + 1, 0, (useRecycling) ? newLine.clone() : newLine);
```
Tyriar: Why is clone used here?

src/common/CircularList.ts
jerch: This basically resembles just the `.push` method, if we need this to fire beforehand, would this also apply to `.push`?

src/common/CircularList.ts
jerch: This way the callback can be used for any push, even if there is nothing to be recycled (early stages). The callback itself can determine whether to recycle or not. This would not be possible by returning the current item, it would limit the usage to the trim case.

src/Terminal.ts
jerch: `copyFrom` copies the cell entries over from `newLine` to `item`. `newLine` is the blueprint for an empty new line. If we are not yet in trim mode, we have to clone `newLine` to get a real instance.

src/Terminal.ts
jerch: `newLine` is not a new line anymore, but a blueprint for a new line that gets carried around by the terminal, thus we have to clone it here to get an independent instance. (Even the JSArray impl can benefit from this blueprint thing, for the TypedArray version I need it to have something to copy from).

jerch:
@Tyriar Did a version without the callback as you suggested (see last commit). Well, seems it has several flaws:
- needs a precheck on caller side whether trim would occur, imho thats bad API design as it leaks the trim functionality and even will lead to wrong results if the check was forgotten or is faulty
- its actually slower than the callback variant (14 vs. 17 MB/s)

I dont like the callback thing either here, its just I had no better idea, how to put it without exposing to many internals. Still up for other ideas.

jerch:
@Tyriar Removed the callback variant, `trimAndRecycle` is alot faster. Drawback is the precondition to use it only for a full circular list or hell will break loose. I commented it accordingly, so we should be on the safe side as long as we dont neglect the comments/doc.

Now it works as follows for recycling:
- a terminal instance holds a blankline blueprint (a buffer line instance with blankLine content)
- the blueprint is read upon entering `Terminal.scroll` and compared to `.cols` and current attrs, if they differ the blueprint is recreated
- any new line is fast cloned from this blueprint (`newLine = blueprint.clone();`)
- if the buffer is at max length, `trimAndRecycle` steps in, sets the old trimmed line as new active one and returns it, this line gets the cell content from the blueprint by fast copy (`trimAndRecycle().copyFrom(blueprint);`)
- done

The blueprint is needed to have something to copy the cell content from without recreating a line everytime. The blueprint alone already gives a nice speedup since the cached blankline is unlikely to change very often.
Overall the benchmark numbers for the input chain are:
- JSArray no recycling: 7 - 9 MB/s
- JSArray with recycling: 9 - 11 MB/s
- TypedArray no recycling: 10 - 12 MB/s
- TypedArray with recycling: 18 - 20 MB/s

Up for another review.

src/common/CircularList.ts
jerch: Obsolete now.

src/common/CircularList.ts
jerch: Obsolete.

src/Terminal.ts
jerch: Explained more detailed below.

src/Terminal.ts
jerch: See explanation below.

src/Terminal.ts
```diff
@@ -1184,9 +1201,19 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
 
       // Insert the line using the fastest method
       if (bottomRow === this.buffer.lines.length - 1) {
-        this.buffer.lines.push(newLine);
+        if (useRecycling) {
+          if (willBufferBeTrimmed) {
+            // Warning: Never call .trimAndRecycle() without the
```
Tyriar: We should move this to CircularList, this is the only usage of trimAndRecycle so it should be encapsulated inside Buffer. Maybe something like this?

```ts
class ICircularList<T> {
 push(value: T, allowRecycle?: boolean): void;
}
```

Nice and safe 😄

Tyriar: Hmm, actually we would need to tweak that a little, still not sure on the names:

```ts
class ICircularList<T> {
 push(valueOrCopyFrom: T, allowRecycle?: boolean): void;
}
```

Then this section could simply be:

```ts
this.buffer.lines.push(newLine, useRecycling);
```

typings/xterm.d.ts
```diff
@@ -112,6 +112,8 @@ declare module 'xterm' {
      */
     experimentalBufferLineImpl?: 'JsArray' | 'TypedArray';
 
+    experimentalPushRecycling?: boolean;
```
Tyriar: Needs a doc comment

src/Terminal.ts
```diff
@@ -1174,7 +1179,19 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
...
-    const newLine = this.buffer.getBlankLine(this.eraseAttr(), isWrapped);
+    let newLine: IBufferLine;
+    const useRecycling = this.options.experimentalPushRecycling;
+    if (useRecycling) {
+      newLine = this._blankLine;
+      if (!newLine || newLine.length !== this.cols || newLine.get(0)[CHAR_DATA_ATTR_INDEX] !== this.eraseAttr()) {
+        newLine = this.buffer.getBlankLine(this.eraseAttr(), isWrapped);
+        this._blankLine = newLine;
+      }
+      newLine.isWrapped = !!(isWrapped);
```
Tyriar: I think we can safely do away with this if we change scroll to:

```ts
scroll(isWrapped: boolean = false): void
```

src/BufferLine.ts
```diff
@@ -251,7 +249,7 @@ export class BufferLineTypedArray implements IBufferLine {
   }
 
   /** alter to a full copy of line  */
-  public copyFrom(line: BufferLineTypedArray): void {
+  public copyFrom(line: BufferLineTypedArray): IBufferLine {
```
Tyriar: I don't think the return value is used anywhere anymore?

jerch:
@Tyriar Something like this should work:
```ts
  public pushFromBlueprint(blueprint: IBufferLine, allowRecycle?: boolean): void {
    if (this._length === this._maxLength) {
      this._startIndex = ++this._startIndex % this._maxLength;
      this.emit('trim', 1);
      if (allowRecycle) {
        (this._array[this._getCyclicIndex(this._length - 1)] as unknown as IBufferLine).copyFrom(blueprint);
      } else {
        this._array[this._getCyclicIndex(this._length - 1)] = blueprint.clone() as unknown as T;
      }
    } else {
      this._array[this._getCyclicIndex(this._length)] = blueprint.clone() as unknown as T;
      this._length++;
    }
  }
```
But this has several issues:
- basically removes the type system
- BufferLine impl is pulled into CircularList
- not sure if `this.emit('trim', 1);` is at the right position (same goes for current `push`)

Note it is not possible to give `push` the final line object beforehand (cloned or recycled), since `push` has to decide first whether to recycle at all (the final object is either a clone or a copyFrom recycled object). To encapsulate this nicely we have imho only 2 options - either the callback variant or give up the generic `<T>` and do an inherited version of CircularList for `IBufferLine` to clean up the type system above.

typings/xterm.d.ts
jerch: Done.

src/Terminal.ts
jerch: Done.

jerch:
@Tyriar The last commit reverts to a cleaner precheck version and does the recycling explicitly in `Terminal.scroll`. This seems much cleaner to me than trying to merge it with `push`.

src/BufferLine.ts
jerch: Removed.

src/BufferLine.ts
```diff
@@ -94,11 +94,8 @@ export class BufferLine implements IBufferLine {
     }
   }
 
-  public copyFrom(line: IBufferLine): void {
-    this._data = [];
-    for (let i = 0; i < line.length; ++i) {
-      this._push(line.get(i));
-    }
+  public copyFrom(line: BufferLine): void {
+    this._data = line._data.slice(0);
```
Tyriar: This should be equivalent, right?

src/Terminal.ts
```diff
@@ -1174,20 +1179,44 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
...
       // Insert the line using the fastest method
       if (bottomRow === this.buffer.lines.length - 1) {
-        this.buffer.lines.push(newLine);
+        if (useRecycling) {
+          if (willBufferBeTrimmed) {
+            // push would trim the oldest line in the ringbuffer
+            // therefore we can recycle it here as the new line
+            const recycled = this.buffer.lines.get(0);
+            recycled.copyFrom(newLine);
+            this.buffer.lines.push(recycled);
```
Tyriar: Since `scroll` happens so often, maybe we should be more low level like before? The get is doing duplicate stuff for example. Another idea?

```ts
  public recycle(): T {
    if (this._length === this._maxLength) {
      this._startIndex++;
      if (this._startIndex === this._maxLength) {
        this._startIndex = 0;
      }
      this.emit('trim', 1);
    } else {
      this._length++;
    }
    // Is that the right one?
    return this._array[this._getCyclicIndex(this._length - 1)]!;
  }
```

```ts
this.buffer.lines.recycle().copyFrom(newLine)
```

src/BufferLine.ts
jerch: Yeah its equivalent, changed it to get rid of the runtime penalty with the loop to have a fair comparison.

src/Terminal.ts
jerch: Yeah thats better, it still does the `(this._length === this._maxLength)` check twice with the `pushWouldTrim` guard in place.

jerch:
@Tyriar The last commit is a compromise between speed and code safety. Was not able to remove the double checking of `(this._length === this._maxLength)` for the recycle control flow path. `recycle` now always works but returns `undefined` for non full ringbuffers, still the full precheck is faster than handling the undefined (due to deopt in the ringbuffer when reading out of bound).
Speed decreased only slightly (18.5 MB/s vs 17.5 MB/s seems neglectible).

src/Terminal.ts
jerch: Commit below mostly does this now, minor difference - the return value still can be undefined.

src/common/CircularList.ts
```diff
@@ -90,14 +90,33 @@ export class CircularList<T> extends EventEmitter implements ICircularList<T> {
...
+      this._length++;
+    }
+  }
+
+  /**
+   * Advance ringbuffer index and return current element for recycling.
+   * Note: If the ringbuffer is not full this method will return undefined,
+   * Either precheck with isFull() or handle the undefined return value accordingly.
+   */
+  public recycle(): T | undefined {
```
Tyriar: I don't think we want to allow `undefined` to return, we're using it in Terminal.ts like this:

```ts
this.buffer.lines.recycle().copyFrom(newLine);
```

Which would throw a NPE. 

I suggest instead we invert the if and start the function like this:

```ts
if (this._length !== this._maxLength) {
  throw new Error('Can only recycle when the buffer is full');
}
```

And then return with ! so TS knows it's a `T`:

```ts
return this._array[this._getCyclicIndex(this._length - 1)]!;
```

We could also add jsdoc to the interface with an explanation in a `@throws` line.

src/common/CircularList.ts
```diff
@@ -90,14 +90,33 @@ export class CircularList<T> extends EventEmitter implements ICircularList<T> {
...
     } else {
       this._length++;
     }
+    return this._array[this._getCyclicIndex(this._length - 1)];
+  }
+
+  /**
+   * Ringbuffer is at max length.
+   */
+  public isFull(): boolean {
```
Tyriar: I think a getters should be just as fast as a function and makes more sense for this since there's no side effects.

typings/xterm.d.ts
```diff
@@ -112,6 +112,13 @@ declare module 'xterm' {
      */
     experimentalBufferLineImpl?: 'JsArray' | 'TypedArray';
 
+    /**
+     * (EXPERIMENTAL) Enable recycling of buffer lines.
+     * 
+     * This option will be removed in the future.
+     */
+    experimentalBufferLineRecycling?: boolean;
```
Tyriar: Thoughts on merging this into `experimentalBufferLineImpl` when it's `'TypedArray'`? Too many experimental options makes thinks more complicated then they need to be, plus we want anyone testing `'TypedArray'` to test recycle as well.

src/common/CircularList.ts
jerch: Done.

src/common/CircularList.ts
jerch: Lol I hesitate to use property getters since they show sometimes really bad perf. But this one seems be the same, changed.

typings/xterm.d.ts
jerch: Ok merged it with TypedArray setting.

jerch:
Averages now to ~18.5 MB/s, this is a nice enhancement compared to ~7.5 MB/s in the current master.

src/common/CircularList.ts
Tyriar: I think the perf impact was in getters vs variables, not functions? Maybe I'm wrong though?

jerch: Well I did not investigate it further so not sure which "types" are/were affected, also I think they partly addressed it around May or so.

Issue #1700, opened 2018-09-18
Support lineHeight in DOM renderer


pelmers:
The line height configuration is only supported in the canvas renderer (as of 3.6.0). 
pointer: https://github.com/xtermjs/xterm.js/blob/e736f1223f5c42eb1b0674050ae55184f2215806/src/renderer/dom/DomRenderer.ts#L53

I thought I had seen an issue for this already but wasn't able to find it now.

Tyriar:
I don't think there's an issue but there's a comment calling this out.

https://github.com/xtermjs/xterm.js/blob/e736f1223f5c42eb1b0674050ae55184f2215806/typings/xterm.d.ts#L152

Open to PRs.

Tyriar:
You can see how this used to be handled by looking at the old DOM renderer in https://github.com/xtermjs/xterm.js/pull/938

leomoty:
Hello @Tyriar, I would like to tackle this one, is it just a matter of using lineHeight inside `DomRenderer.ts` and defaulting to normal otherwise?

Tyriar:
@leomoty great!

We want it to act like line height in the canvas renderer, which is used here:

https://github.com/xtermjs/xterm.js/blob/e7ae4be394d67ef0ca1b4f41e5251b42b9630999/src/renderer/Renderer.ts#L222-L230

Also the value is always set so you don't need to worry about using normal:

https://github.com/xtermjs/xterm.js/blob/e7ae4be394d67ef0ca1b4f41e5251b42b9630999/src/Terminal.ts#L94

You can see how we used to do it by looking at src/Renderer.ts in this PR:

https://github.com/xtermjs/xterm.js/pull/938

PR #1733, opened 2018-10-09
Support for lineHeight in DOM Renderer

leomoty:
Related to #1700. I've attempted to port the calculations from Canvas renderer, it seems to be a few pixels off.

Also, does xterm.js support zooming the browser officially? That is not working properly.

@Tyriar, can you please take a look.

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -473,6 +473,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
         if (this._theme) {
           this.renderer.setTheme(this._theme);
         }
+        this.mouseHelper = new MouseHelper(this.renderer);
         break;
       case 'scrollback':
         this.buffers.resize(this.cols, this.rows);
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -473,7 +473,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
         if (this._theme) {
           this.renderer.setTheme(this._theme);
         }
-        this.mouseHelper = new MouseHelper(this.renderer);
+        this.mouseHelper.setRenderer(this.renderer);
         break;
       case 'scrollback':
         this.buffers.resize(this.cols, this.rows);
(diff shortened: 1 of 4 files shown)


Tyriar:
> Also, does xterm.js support zooming the browser officially? That is not working properly.

@leomoty yes using the browser zoom should work.

Tyriar:
Hm, seems alright to me. Note that you're not aiming for exact parity with the canvas renderer, it's fine if they're a little different.

leomoty:
@Tyriar, updated `_createSelectionElement`, the selection height is working properly now, but the selection doesn't seem to match as you go down the lines. Any pointers?

Tyriar:
@leomoty here's the function that grabs the coordinates from the MouseEvent:

https://github.com/xtermjs/xterm.js/blob/b650db5b1ae230d4c7adeb39c433be28712dc66f/src/SelectionManager.ts#L485

leomoty:
@Tyriar I have been a bit busy with work, just noticed the status change, I also took another look, I might have found the issue. Unsure why it happens tho.

Following the code you gave me it goes here:

https://github.com/xtermjs/xterm.js/blob/master/src/utils/MouseHelper.ts#L51-L72

When I tried to inspect `this._renderer.dimensions` itself, it doesnt seem to account for the changes I did in: 

https://github.com/xtermjs/xterm.js/pull/1733/files#diff-0a9d734ca7737653ca5c980f3c663f7aR96

I.e. `actualCellHeight` is 17px instead of the expected 34px. I can confirm that inside `_updateDimensions` it is fine.

leomoty:
As a follow-up: `getCoords` gets lineHeight passed in, if I apply that to `actualCellHeight` it works as expected.

 I wonder if this might break the canvas renderer tho. I'll wait for your feedback.

leomoty:
@Tyriar Yeah, I didn't mean to change canvas renderer nor selection at all, when I noticed the `MouseHelper@getCoords` function, I checked the `this._renderer` object there.

What I noticed is:
- Canvas renderer works fine
- The changes I did to Dom Renderer are not applied

Inside `DomRenderer@_updateDimensions` with `lineHeight` = 2.0, I get 34px for `actualCellHeight`.
However `MouseHelper@getCoords` still accounts for 17px.

I have not been able to track down why it happens.

src/Terminal.ts
```diff
@@ -473,6 +473,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
         if (this._theme) {
           this.renderer.setTheme(this._theme);
         }
+        this.mouseHelper = new MouseHelper(this.renderer);
```
leomoty: @Tyriar when `lineHeight` is updated,  it falls through to `rendererType`, and thus destroying the renderer, in which mouseHelper is never updated.

I am not sure we want to destroy the current renderer when those props are updated.

Either way, selection works properly with this change.

PR #1739, opened 2018-10-09
Allow inverse colors to be stored in glyph keys, fix inverted colors in DOM renderer

### LINKED ISSUE #1737: Inverse characters are not printed correctly when used with dynamic texture atlas
Repro:

```
echo -e '\x1b[7mabc'
```

screen shot 2018-10-09 at 10 49 57 am

### LINKED ISSUE #1738: Inverted default fg color is not correct in DOM renderer
The actual background color (black) is not being used here:

screen shot 2018-10-09 at 11 38 16 am

Tyriar:
Fixes #1737

---

This issue was happening as we were assuming fg was a number >= 0 for the glyph key but it could also be `-1` to indicate an inverse default color. See the first commit for this fix.

The second commit cleans up the usage of 256/257 as the fg color by introducing constants. It also removes 257 as the "No BG color key" and instead uses 256 as that is the default color for the background, I don't think there's a reason to differentiate them.

The third commit fixes #1738 (not a regression).

---

## Before

**DOM Renderer**

screen shot 2018-10-09 at 11 41 42 am

**Canvas/static**

screen shot 2018-10-09 at 11 41 05 am

**Canvas/dynamic**

screen shot 2018-10-09 at 11 41 34 am

## After

**DOM Renderer**

screen shot 2018-10-09 at 11 42 47 am

**Canvas/static**

screen shot 2018-10-09 at 11 42 30 am

**Canvas/dynamic**

screen shot 2018-10-09 at 11 42 38 am


src/renderer/BaseRenderLayer.ts
```diff
@@ -298,7 +298,7 @@ export abstract class BaseRenderLayer implements IRenderLayer {
 
     if (fg === INVERTED_DEFAULT_COLOR) {
       this._ctx.fillStyle = this._colors.background.css;
-    } else if (fg < 256) {
+    } else if (fg < DEFAULT_COLOR) {
```
bgw: Doing a range check against what is essentially an enum value is a little confusing. Since we're doing this in a few places, maybe we should have a `isNormalColor` helper function?

Tyriar: Sounds good 😃

PR #1740, opened 2018-10-09
fix handling of length in BufferLine.resize

jerch:
Currently we do not trim buffer lines upon resizing. `BufferLine.resize` wrongly adjusted `.length` to the new shorter `cols` value, which makes the invisible content beyond `terminal.cols` non accessible for consumers like `Buffer.translateBufferLineToString`. This PR restores the behavior of v3.7.

### DIFF
--- a/src/BufferLine.test.ts
+++ b/src/BufferLine.test.ts
@@ -133,31 +133,71 @@ describe('BufferLine', function(): void {
     const line3 = line.clone();
     chai.expect(TestBufferLine.prototype.toArray.apply(line3)).eql(line.toArray());
   });
-  it('resize enlarge', function(): void {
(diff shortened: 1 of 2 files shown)


Tyriar:
@jerch I can see lines being chopped off in VS Code, but when I try shrink and restore on xterm.js master demo it seems to keep the content?

jerch:
@Tyriar No test written (can do), but you can test it in the demo with the linkifier:

Put some longish URL in the terminal, maybe with some parameters, shrink it so the URL matcher would still see it as a valid URL -- boom, the parameters are gone (matches wrongly the shorter one). Im not even sure what the right behavior here, tbh I like the WYSIWYG idea with terminals more than the hidden stuff idea, but you told me that vscode does some selection based things for tasks relying on that behavior, therefore this PR restores that behavior.

PR #1744, opened 2018-10-10
add tiny buffer to server part in demo

jerch:
This PR adds a tiny bit of buffering to the demo on the server side. This is mainly to get those unreliable numbers in browser perf tests somewhat more stable. A nice side effect is a major throughput boost for the demo app (~3x faster).


jerch:
Sweet, thx :smile_cat:

PR #1748, opened 2018-10-11
Fix NPE in DOM renderer underline

### LINKED ISSUE #1747: DOMRenderer Uncaught TypeError: Cannot read property 'children' of undefined
https://github.com/Microsoft/vscode/issues/60447

Stack 
TypeError: Cannot read property 'children' of undefined
at DomRenderer._setCellUnderline (\node_modules.asar\vscode-xterm\lib\renderer\dom\DomRenderer.js:280:45)
at DomRenderer._onLinkHover (\node_modules.asar\vscode-xterm\lib\renderer\dom\DomRenderer.js:273:14)
at Linkifier. (\node_modules.asar\vscode-xterm\lib\renderer\dom\DomRenderer.js:65:79)
at Linkifier.EventEmitter.emit (\node_modules.asar\vscode-xterm\lib\common\EventEmitter.js:70:20)
at MouseZone.hoverCallback (\node_modules.asar\vscode-xterm\lib\Linkifier.js:182:19)
at MouseZoneManager._onHover (\node_modules.asar\vscode-xterm\lib\ui\MouseZoneManager.js:106:18)
at MouseZoneManager._onMouseMove (\node_modules.asar\vscode-xterm\lib\ui\MouseZoneManager.js:84:18)
at HTMLDivElement._super.MouseZoneManager._this._mouseMoveListener (\node_modules.asar\vscode-xterm\lib\ui\MouseZoneManager.js:30:64)

Tyriar:
Fixes #1747

### DIFF
--- a/src/renderer/dom/DomRenderer.ts
+++ b/src/renderer/dom/DomRenderer.ts
@@ -352,7 +352,11 @@ export class DomRenderer extends EventEmitter implements IRenderer {
 
   private _setCellUnderline(x: number, x2: number, y: number, y2: number, cols: number, enabled: boolean): void {
     while (x !== x2 || y !== y2) {
-      const span = <HTMLElement>this._rowElements[y].children[x];
+      const row = this._rowElements[y];
+      if (!row) {
+        return;
+      }
+      const span = <HTMLElement>row.children[x];
       span.style.textDecoration = enabled ? 'underline' : 'none';
       x = (x + 1) % cols;
       if (x === 0) {


Tyriar:
@jerch I couldn't repro the problem unfortunately, this is just a sanity check to prevent the exception which does happen (which we will need when strict null checks are enabled anyway). If there is a bug, someone might report it someday by saying a link doesn't work and give us a repro case.

Issue #1749, opened 2018-10-11
The search addon searches from where was last clicked, not the top of the viewport when there is no selection


Tyriar:
Maybe this if is always true?

https://github.com/xtermjs/xterm.js/blob/b650db5b1ae230d4c7adeb39c433be28712dc66f/src/addons/search/SearchHelper.ts#L34-L37

VS Code issue: https://github.com/Microsoft/vscode/issues/59376

skprabhanjan:
@Tyriar , I changed it accordingly but I have one doubt 
` if (this._terminal.getSelection().length !== 0) {
        startRow = this._terminal._core.selectionManager.selectionEnd[1];
      }
      else if (this._terminal.getSelection().length === 0) {
        **startRow = X **
      }`
should this X be this._terminal._core.selectionManager.selectionStart[1] or 0 or something else ? 
Will send a PR to fix this if you confirm this , thanks :)

Tyriar:
@skprabhanjan ydisp contains the absolute row for the top of the viewport, so that would be the one to use when there is no selection 👍

PR #1755, opened 2018-10-15
#1749 The search addon searches from where was last clicked, not the …

Marina-Miranovich:
…top of the viewport when there is no selection


Marina-Miranovich:
From what I could see that doesn't clear selection if the user clicked only once. To say more accurate: after single click it clear the visual selection, but doesn't clear selectionEnd, selectionStart and etc. So add the clean method to the single click handler, so if the selection length == 0 (no selection) clear it.

Tyriar:
Looks like this was fixed in https://github.com/xtermjs/xterm.js/pull/1760, sorry for not seeing this ☹️

PR #1760, opened 2018-10-17
Search from top of viewport when there is no selection

### LINKED ISSUE #1749: The search addon searches from where was last clicked, not the top of the viewport when there is no selection

skprabhanjan:
@Tyriar , fixed it in two places to resolve issue #1749 .
The startRow is initialised to ydisp in both places, so just handled the case of changing it when selection length is not 0. 
Please review this , Thanks :)

Fixes #1749

### DIFF
--- a/src/addons/search/SearchHelper.ts
+++ b/src/addons/search/SearchHelper.ts
@@ -33,7 +33,9 @@ export class SearchHelper implements ISearchHelper {
     let startRow = this._terminal._core.buffer.ydisp;
     if (this._terminal._core.selectionManager.selectionEnd) {
       // Start from the selection end if there is a selection
-      startRow = this._terminal._core.selectionManager.selectionEnd[1];
+      if (this._terminal.getSelection().length !== 0) {
+        startRow = this._terminal._core.selectionManager.selectionEnd[1];
+      }
     }
 
     // Search from ydisp + 1 to end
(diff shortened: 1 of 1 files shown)


src/addons/search/SearchHelper.ts
```diff
@@ -108,7 +112,7 @@ export class SearchHelper implements ISearchHelper {
    */
   private _isWholeWord(searchIndex: number, line: string, term: string): boolean {
     return (((searchIndex === 0) || (nonWordCharacters.indexOf(line[searchIndex - 1]) !== -1)) &&
-         (((searchIndex + term.length) === line.length) || (nonWordCharacters.indexOf(line[searchIndex + term.length]) !== -1)));
```
Tyriar: This was intentional, it's not documented anywhere but double indents for wrapped lines helps differentiate new statements and wrapped. Example:

```
if (a &&
    b) {
  c();
}
```

vs

```
if (a &&
  b) {
  c();
}
```

skprabhanjan: Hey @Tyriar, sorry this was not intentional and got changed in the auto indentation.( I changed it back the first time but missed it this time )

PR #1764, opened 2018-10-20
Alternate buffer improvement

whydoubt:
Alternate buffer handling in xterm.js is lacking in a few areas compared with most other terminals.  One is how the cursor color and position is saved and restored.  Another is how clearing is done when entering the alternate buffer.

These changes help xterm.js better replicate the behavior of other terminals.

### DIFF
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -38,6 +38,7 @@ export class Buffer implements IBuffer {
   public tabs: any;
   public savedY: number;
   public savedX: number;
+  public savedCurAttr: number;
   public markers: Marker[] = [];
   private _bufferLineConstructor: IBufferLineConstructor;
 
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -114,11 +114,14 @@ export class Buffer implements IBuffer {
   /**
    * Fills the buffer's viewport with blank lines.
    */
-  public fillViewportRows(): void {
+  public fillViewportRows(fillAttr?: number): void {
     if (this.lines.length === 0) {
+      if (fillAttr === undefined) {
+        fillAttr = DEFAULT_ATTR;
+      }
       let i = this._terminal.rows;
       while (i--) {
-        this.lines.push(this.getBlankLine(DEFAULT_ATTR));
+        this.lines.push(this.getBlankLine(fillAttr));
       }
     }
   }
--- a/src/BufferSet.test.ts
+++ b/src/BufferSet.test.ts
@@ -48,4 +48,30 @@ describe('BufferSet', () => {
       assert.equal(bufferSet.active, bufferSet.alt);
     });
   });
+
+  describe('cursor handling when swapping buffers', () => {
+    beforeEach(() => {
+      bufferSet.normal.x = 0;
+      bufferSet.normal.y = 0;
+      bufferSet.alt.x = 0;
+      bufferSet.alt.y = 0;
+    });
+
+    it('should keep the cursor stationary when activating alt buffer', () => {
+      bufferSet.activateNormalBuffer();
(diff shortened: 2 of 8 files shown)


Tyriar:
Do you have an example program that illustrates the bad behavior inside xterm.js?

whydoubt:
[alt_test.txt](https://github.com/xtermjs/xterm.js/files/2498717/alt_test.txt)
Since there are a several variations I am checking for, I created a script that sleeps between each operation.  I open up all terminals I want to check, determine tty for each, then in a separate terminal run something like:
`. alt_test.txt > /dev/pts/7 & . alt_test.txt > /dev/pts/8`
...to make it run in all terminals practically simultaneously.  Each terminal treats it a bit different, but my changes most closely follow xterm.

whydoubt:
Some smaller tests...

### Fix cursor movement when swapping buffers
I made xterm.js follow xterm behavior in this test _on a cleared buffer_:
```shell
for x in 47 1047 1049 1048 9999; do
echo -e "\E[0m== ${x} ==\n\E[?${x}h\n\E[31mJUNK\E[?${x}lTEST\E[0m\n"; done
```
If 1049 comes back with TEST being red, it means that the attributes of the normal buffer are affected when I enter and exit vim.  The result of this can be unpleasant.

### Each buffer needs its own saved cursor attributes
On running the following in a bash shell, TEST should **not** be red.
 ```shell
echo -e '\E[0m\E[?1049h\E[31m\E[s\E[?1049lTEST\E[0m'
```

### Clear alt buffer with erase attributes upon entering
In most terminals, this will cause the entire buffer to flash green for a second:
```shell
echo -ne '\E[0m\E[42m\E[?1049h'; sleep 1; echo -ne '\E[?1049l\E[0m'
```

jerch:
@whydoubt Thanks alot for fixing this, LGTM :+1: . I think we should have a few test cases that cover the alt buffer switch behavior rather than relying on external scripts, maybe you could add those under `BufferSet.test.ts`?

whydoubt:
I have updated the commits to include relevant tests.

Tyriar:
The changes seem good, just want some clarification on what the first test is meant to do:

```bash
for x in 47 1047 1049 1048 9999; do echo -e "\033[0m== ${x} ==\n\033[?${x}h\n\033[31mJUNK\033[?${x}lTEST\033[0m\n"; done
```

This screenshot shows macOS programs in this order (left to right): This PR, iTerm2, Terminal.app, xterm.js based on master:

screen shot 2018-11-18 at 2 44 37 pm

whydoubt:
The key parts in pseudo-code: {print header} {enter alt} {print newline} {text red} {print 'JUNK'} {leave alt} {print TEST}.
- If you see JUNKTEST in red, it means the alt control code had no effect (9999 is the control case for that).
- Otherwise, if you see JUNK on the second line, it means the buffer did not clear on leaving the alt buffer (expected with 1048).
- If the cursor (position and attributes) is saved on entry / restored on exit, TEST should be in the upper-left corner with default attributes (1048 / 1049).
- If the cursor (position and attributes) are NOT restored on exit, TEST should be on the second line, indented, in red (47 / 1047).

whydoubt:
term_alt_tests
On Linux, left to right: Konsole, XTerm, Putty, GNOME Terminal (vte-based)

jerch:
@whydoubt Wow nice test cases! About correctness - only "spec" I can find about those higher set mode numbers is from xterm and this clearly states where to store/restore cursor attrs and where not. Guess the other emulators do it wrong? Maybe @egmontkob also wants to have a look at this.

Since it now works as stated in the xterm doc I am fine with adding it as it is.

PR #1765, opened 2018-10-22
Updated webLinks regex accept colons, fix vscode issue#60401

jrkong:
@Tyriar This PR should fix [issue #60401](https://github.com/Microsoft/vscode/issues/60401) for VS Code downstream.

I updated the `webLinks` `pathClause` regex https://github.com/xtermjs/xterm.js/blob/b9ab35dfe9f117e00133d6bb782cd298c1401d77/src/addons/webLinks/webLinks.ts#L17 to accept colons. So the new regex is now `(\\/[\\/\\w\\.\\-%~:]*)*`

I also updated the tests to test for the colon case in the path.

image

If there are any updates I should make please let me know.

### DIFF
--- a/src/addons/webLinks/webLinks.test.ts
+++ b/src/addons/webLinks/webLinks.test.ts
@@ -32,11 +32,11 @@ describe('webLinks addon', () => {
     const term = new MockTerminal();
     webLinks.webLinksInit(<any>term);
 
-    const row = '  http://foo.com/a~b#c~d?e~f  ';
+    const row = '  http://foo.com/a~b#c~d?e~f:  ';
 
     const match = row.match(term.regex);
     const uri = match[term.options.matchIndex];
 
-    assert.equal(uri, 'http://foo.com/a~b#c~d?e~f');
+    assert.equal(uri, 'http://foo.com/a~b#c~d?e~f:');
   });
 });
--- a/src/addons/webLinks/webLinks.test.ts
+++ b/src/addons/webLinks/webLinks.test.ts
@@ -32,11 +32,35 @@ describe('webLinks addon', () => {
     const term = new MockTerminal();
     webLinks.webLinksInit(<any>term);
 
-    const row = '  http://foo.com/a~b#c~d?e~f:  ';
+    const row = '  http://foo.com/a~b#c~d?e~f  ';
 
     const match = row.match(term.regex);
     const uri = match[term.options.matchIndex];
 
-    assert.equal(uri, 'http://foo.com/a~b#c~d?e~f:');
+    assert.equal(uri, 'http://foo.com/a~b#c~d?e~f');
+  });
+
+  it('should allow : character in URI path', () => {
+    const term = new MockTerminal();
+    webLinks.webLinksInit(<any>term);
+
+    const row = '  http://foo.com/colon:test  ';
+
+    const match = row.match(term.regex);
+    const uri = match[term.options.matchIndex];
+
+    assert.equal(uri, 'http://foo.com/colon:test');
+  });
+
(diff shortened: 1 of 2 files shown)


src/addons/webLinks/webLinks.test.ts
```diff
@@ -32,11 +32,11 @@ describe('webLinks addon', () => {
     const term = new MockTerminal();
     webLinks.webLinksInit(<any>term);
 
-    const row = '  http://foo.com/a~b#c~d?e~f  ';
+    const row = '  http://foo.com/a~b#c~d?e~f:  ';
```
Tyriar: Would it be possible to allow semi-colons except for the last character?

So this row:

```
const row = '  http://foo.com/a~b#c~d?e~:f:  ';
```

Resulted in this:

```
assert.equal(uri, 'http://foo.com/a~b#c~d?e~:f');
```

jrkong: @Tyriar I've been testing it out and I don't think it's doable, at least not through regex.

I was thinking of handling the case by processing the `uri` after it's processed by `strictUrlRegex`. Would that be a viable solution? If so how would I access the `uri` before it's returned? I couldn't to access it properly/safely during my tests.

If you have any other ideas on how to implement this please let me know.

Tyriar: The concern is that then we will break on this type of link https://github.com/Microsoft/vscode/issues/50426#issuecomment-432141250 (I know it's a file not a URL but this just popped up in my inbox). It's hard to say which is more common 😕

jrkong: @Tyriar is that the only case we need to be concerned with as I don't think it will be an issue. 

In VS Code the path regex is [managed by these lines in terminalLinkHandler.ts in VS Code](https://github.com/Microsoft/vscode/blob/496ebc6438e48ca6cc85836da946ab2323e378ab/src/vs/workbench/parts/terminal/electron-browser/terminalLinkHandler.ts#L19-L42).

This results in the regex 

```
/(((\.\.?|\~)|([^\0\s!$`&*()\[\]+'":;\\])+)?(\/([^\0\s!$`&*()\[\]+'":;\\])+)+)(((\S*)",[  ]line[  ]((\d+)([  ]column[  ](\d+))?))|((\S*)[  ]on[  ]line[  ]((\d+)(,[  ]column[  ](\d+))?))|((\S*):line[  ]((\d+)(,[  ]column[  ](\d+))?))|(([^\s\(\)]*)(\s?[\(\[](\d+)(,\s?(\d+))?)[\)\]])|(([^:\s\(\)<>'"\[\]]*)(:(\d+))?(:(\d+))?))/
``` 
in Javascript.
image

In comparison to 
```
/(?:^|[^\da-z\.-]+)((https?:\/\/)((([\da-z\.-]+)\.([a-z\.]{2,6}))|((\d{1,3}\.){3}\d{1,3})|(localhost))(:\d{1,5})?(\/[\/\w\.\-%~]*)*(\?[0-9\w\[\]\(\)\/\?\!#@$%&'*+,:;~\=\.\-]*)?(#[0-9\w\[\]\(\)\/\?\!#@$%&'*+,:;~\=\.\-]*)?)($|[^\/\w\.\-%]+)/
```
which is used when a web link is detected.

image

Are there any other cases that this change could break?

jrkong: @Tyriar fixed! `[^:\s]` worked. I completely overlooked the `\s` character as I was doing my tests! Thanks for pointing that out! I've added tests as well to verify the functionality of both cases.

PR #1769, opened 2018-10-25
fix linkifier after shrinking

jerch:
Fix for https://github.com/Microsoft/vscode/issues/61708.

~To get this working, I had to add a partial fix of the trim behavior of `Buffer.translateBufferLineToString` for lines that are longer than the current `Terminal.cols` and get new content (contains a generalized version of this commit 48bd634228e25fb7a47e554bd46134991c6160f5, thanks to @dnfield).~ Reverted since it introduces to many side effects.


src/Linkifier.test.ts
```diff
@@ -35,63 +34,29 @@ class TestMouseZoneManager implements IMouseZoneManager {
...
   let mouseZoneManager: TestMouseZoneManager;
 
   beforeEach(() => {
-    terminal = new MockTerminal();
-    terminal.cols = 100;
-    terminal.rows = 10;
-    terminal.buffer = new MockBuffer();
-    (<MockBuffer>terminal.buffer).setLines(new CircularList<IBufferLine>(20));
-    terminal.buffer.ydisp = 0;
+    terminal = new TestTerminal({cols: 100, rows: 10});
```
Tyriar: Much nicer 😃

src/Buffer.test.ts
```diff
@@ -500,19 +500,18 @@ describe('Buffer', () => {
...
 
-    it('multiline fullwidth chars with offset 1 (currently tests for broken behavior)', () => {
+    it('multiline fullwidth chars with offset 1', () => {
       const input = 'a１２３４５６７８９０１２３４５６７８９０';
       // the 'a' at the beginning moves all fullwidth chars one to the right
       // now the end of the line contains a dangling empty cell since
       // the next fullwidth char has to wrap early
       // the dangling last cell is wrongly added in the string
-      // --> fixable after resolving #1685
+      // --> fixable after resolving #1685, fixed in ....
```
Tyriar: fixed in ....?

src/Linkifier.ts
```diff
@@ -222,6 +221,30 @@ export class Linkifier extends EventEmitter implements ILinkifier {
...
+      endIndex[1] += this._terminal.buffer.lines.get(endIndex[0]).get(endIndex[1])[CHAR_DATA_WIDTH_INDEX];
+      if (endIndex[1] >= this._terminal.buffer.lines.get(endIndex[0]).length) {
+        endIndex[0]++;
+        endIndex[1] = 0;
+      }
+      // adjust end index to visible line length
+      if (endIndex[1] >= this._terminal.cols) {
+        endIndex[1] = this._terminal.cols - 1;
+      }
+      const visibleLength = (endIndex[0] - bufferIndex[0]) * this._terminal.cols - bufferIndex[1] + endIndex[1];
```
Tyriar: Does using visibleLength mean URLs will be broken if the viewport is shrunk?

src/Buffer.test.ts
jerch: lol yeah with this PR (had the number not yet)

src/Linkifier.ts
jerch: visibleLength is just there to give the marker the length to underline. This makes sure that the underline does not over- or underflow the real end of the url.

src/Buffer.test.ts
jerch: Fixed.

src/Linkifier.test.ts
jerch: Cleaned it up further.

jerch:
@Tyriar Can you test again? Now it gets very hacky, it also pulls in the problem of the empty cell representation. :unamused:

src/Linkifier.ts
```diff
@@ -221,7 +220,31 @@ export class Linkifier extends EventEmitter implements ILinkifier {
       rex.lastIndex = stringIndex + uri.length;
 
       // get the buffer index as [absolute row, col] for the match
-      const bufferIndex = this._terminal.buffer.stringIndexToBufferIndex(rowIndex, stringIndex);
+      const bufferIndex = this._terminal.buffer.stringIndexToBufferIndex(rowIndex, stringIndex, true);
+      // calculate buffer index of uri end
+      // we cannot directly use uri.length here since stringIndexToBufferIndex would
+      // skip empty cells and stop at the next cell with real content
+      // instead we fetch the index of the last char in uri and advance to the next cell
+      const endIndex = this._terminal.buffer.stringIndexToBufferIndex(rowIndex, stringIndex + uri.length - 1, true);
```
Tyriar: This is a problem case with always trimming:

screen shot 2018-10-26 at 10 13 44 am

What if we only trim the whitespace inside the shadowed content, so here the 2 spaces would still remain. This would open up a problem with echoing this and then sizing down twice:

screen shot 2018-10-26 at 10 17 17 am

But maybe that's a fine compromise until reflow as links without resizing will work then?

jerch:
@Tyriar Im not sure anymore if this is fixable in a sane way without altering the trimming behavior (this will touch the empty cell representation and the `isWrapped` handling as well). I am able to fix it with appending real empty cells with enlarging and not trimming lines under any circumstances (as done in the reverted commits, which partly implements #1685). But this comes with unknown side effects that are not covered by test cases which is really bad and therefore not feasible.
Imho your suggestion to only trim spaces in the shadowed content just moves the problem to the end of the shadowed line and shows the same awkward faulty behavior after shrinking. Enlarging wont work at all.
I suggest to stall this PR until we got reflow support since it will most likely also cover those trim awkwardities. Currently our buffer representation of wrapped stuff seems to be broken beyond repair to me. :unamused:
Another way to fix this with our current buffer representation of wrapped stuff is to get trimming based on #1685 working, I think thats what most other emulators do that dont support reflow.

NB: Also found another issue with `isWrapped` - if you skip through multiline commands at the shell the `isWrapped` flag never gets lifted again - any `Buffer.translateBufferLineToString` consumer **always** gets those lines as one line afterwards, even if they contain single short separate lines.

:unamused: :unamused:

jerch:
The last commit works regarding the linkifier as far as I have tested it, but contains those empty cells after enlarging, hmm. (Also a few tests dont pass since they test for the "faulty" behavior).

jerch:
Currently stalled by #1775.

jerch:
@Tyriar With #1775 the issue you described should be gone, also the original issue https://github.com/Microsoft/vscode/issues/61708 should be fixed now.

The line now sticks correctly to the first and the last cell of the url char, even for surrogates and combining chars. Only weird behavior still is the "space" extension to the right after enlarging the terminal. Note that this is only a visual glitch, c&p of the url correctly does not insert any spaces.

Up for review.

jerch:
Partially blocked by #1864.

Reminder: With reflow in place remove the hidden content workarounds as we dont need them anymore.

Tyriar:
Moved to https://github.com/xtermjs/xterm.js/issues/2115

PR #1772, opened 2018-10-29
Only show cross-hairs when terminal has focus

### LINKED ISSUE #1767: Column selection crosshair doesn't go away when you alt+tab
## Details
- OS version: Windows
- xterm.js version: 3.8.0

### Steps to reproduce

1. Launch demo
2. Put mouse over terminal
3. alt+tab to another program
4. Release alt
5. Mousing over the unfocused browser will incorrectly change to the crosshair cursor

Related: 

https://github.com/xtermjs/xterm.js/pull/1766

whydoubt:
Fixes #1767

### DIFF
--- a/src/xterm.css
+++ b/src/xterm.css
@@ -139,7 +139,7 @@
     cursor: pointer;
 }
 
-.xterm.xterm-cursor-crosshair {
+.xterm.column-select.focus {
     /* Column selection mode */
     cursor: crosshair;
 }
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -1154,9 +1154,9 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
    */
   public updateCursorStyle(ev: KeyboardEvent): void {
     if (this.selectionManager && this.selectionManager.shouldColumnSelect(ev)) {
-      this.element.classList.add('xterm-cursor-crosshair');
+      this.element.classList.add('column-select');
     } else {
-      this.element.classList.remove('xterm-cursor-crosshair');
+      this.element.classList.remove('column-select');
     }
   }
 


jerch:
@whydoubt This is not working for me, I dont get crosshair cursor at all now for column selection. (Ubuntu 16, tested with Firefxo and Chrome) Any clue what going on?

jerch:
Ah yepp my bad, the css file was not refreshed. It kinda partially solves the issue #1767 - after ALT+TAB with a different focus window in between it works, but it still shows the crosshair if I tab through all windows back to Chrome. Then I have to click in Chrome outside of the terminal div to get rid of the crosshair. Any idea how to solve this?

src/xterm.css
```diff
@@ -139,7 +139,7 @@
     cursor: pointer;
 }
 
-.xterm.xterm-cursor-crosshair {
```
Tyriar: `xterm-cursor-crosshair` aligns with the naming of `xterm-cursor-pointer`, not sure we want to change this?

whydoubt: The current alignment of the names re-enforces the false impression that the two are strongly correlated, and is part of why the name _should_ change. At some point `xterm-cursor-pointer` should be replaced with something more semantic as well (maybe something like `linkified`).

whydoubt:
@Tyriar : not sure what your point regarding VS Code is, because if I understand you correctly, it does not act that way now, and changing it would depend on changing VS Code.  Regarding clearing the `column-select` class on blur, I will look into that.
@jerch : Is this with the demo? With the dev tools, can you see what is happening in terms of the applied classes as I described in an earlier comment?

jerch:
@whydoubt Yepp the classes are applied correctly for tabbing in and out with a different focused window in between, but not if I tab through all back to chrome.

whydoubt:
@jerch It appears to affect all browsers I have on Fedora&mdash;with GNOME Shell handling window management.  I see that a keydown event for the first _Alt_ is received, but no further events.  The window manager ostensibly swallows all further keyups/keydowns, and no focus/blur events get fired either.  There may not be a solution to this one, unless we can find some event that does get fired.
BTW, pressing any key once you have returned should clear this, as this forces a keydown event.

jerch:
@whydoubt Yeah I think the skinda alwayskinda alwaysituation with the window manager cannot be improved by normal means here (means from within the browser). So the partial solution is good to go.

PR #1775, opened 2018-10-31
fix trimming of right side

jerch:
The current trimming of lines cannot distinguish between inserted spaces and empty cells since it trims spaces. This is related and discussed in #1685, also #1769 is affected by this.

This PR is a testbed to evaluate the empty cell representation with a different char than space to overcome the drawbacks. It currently inserts an empty string instead, for TypedArray this automatically falls back to '\x00' (treated as '').

Although doing an additional char replace this is even faster with TypedArray than the current impl (tested with xterm-benchmark with https://gist.github.com/jerch/acca11c68bfe5c8172c909a1cb9fa6db):

- master:
```
   Context "examples/terminal_translateToString.perf.js"
      Context "translateToString - JSArray"
         Case "#1" : 1 - runtime: 96.65 ms
         Case "#1" : 2 - runtime: 32.21 ms
         Case "#1" : 3 - runtime: 30.18 ms
         Case "#1" : 4 - runtime: 28.49 ms
         Case "#1" : 5 - runtime: 26.41 ms
         Case "#1" : 5 runs - average runtime: 42.79 ms
      Context "translateToString - TypedArray"
         Case "#1" : 1 - runtime: 64.41 ms
         Case "#1" : 2 - runtime: 65.35 ms
         Case "#1" : 3 - runtime: 42.41 ms
         Case "#1" : 4 - runtime: 43.92 ms
         Case "#1" : 5 - runtime: 37.30 ms
         Case "#1" : 5 runs - average runtime: 50.68 ms
```
- this PR:
```
   Context "examples/terminal_translateToString.perf.js"
      Context "translateToString - JSArray"
         Case "#1" : 1 - runtime: 93.65 ms
         Case "#1" : 2 - runtime: 30.70 ms
         Case "#1" : 3 - runtime: 37.63 ms
         Case "#1" : 4 - runtime: 34.29 ms
         Case "#1" : 5 - runtime: 32.54 ms
         Case "#1" : 5 runs - average runtime: 45.76 ms
      Context "translateToString - TypedArray"
         Case "#1" : 1 - runtime: 45.00 ms
         Case "#1" : 2 - runtime: 29.49 ms
         Case "#1" : 3 - runtime: 24.42 ms
         Case "#1" : 4 - runtime: 38.41 ms
         Case "#1" : 5 - runtime: 24.89 ms
         Case "#1" : 5 runs - average runtime: 32.44 ms
```

Why bothering at all, doesnt it work as it is? Well we have several issues open that boil down to this problem. All consumers of `Buffer.translateBufferLineToString` with trimming enabled are affected by this. It gets even worse when a consumer tries to unwrap lines, esp. if a resize was done in between (resizing itself is a different problem though that will be addressed with reflow).

~Edit: The canvas renderer seems to be faster too, the time for `./lib/renderer/TextRenderLayer.js.TextRenderLayer._forEachCell` dropped from ~ 350 ms to ~ 220 ms (tested in chrome, to lazy to write a xterm-benchmark case). Any idea why? (I am not familiar with the code)~ Was due to wrongly skipping empty cells...

Would be good to get some early feedback before going down the rabbit hole (yeah it is likely to affect several parts of the code base).


src/Buffer.test.ts
```diff
@@ -509,10 +509,10 @@ describe('Buffer', () => {
       // --> fixable after resolving #1685
       terminal.writeSync(input);
       // TODO: reenable after fix
-      // const s = terminal.buffer.contents(true).toArray()[0];
-      // assert.equal(input, s);
+      const s = terminal.buffer.iterator(true).next().content;
+      assert.equal(input, s);
       for (let i = 10; i < input.length; ++i) {
-        const bufferIndex = terminal.buffer.stringIndexToBufferIndex(0, i + 1); // TODO: remove +1 after fix
+        const bufferIndex = terminal.buffer.stringIndexToBufferIndex(0, i); // TODO: remove +1 after fix
```
Tyriar: Remove the todo

src/BufferLine.ts
```diff
@@ -279,4 +302,27 @@ export class BufferLineTypedArray implements IBufferLine {
     newLine.isWrapped = this.isWrapped;
     return newLine;
   }
+
+  public getTrimmedLength(): number {
```
Tyriar: I think VTE kept a single number with each buffer line for the length? If we did that we could keep stale data in the array beyond the length without concern and update the length in O(1).

src/BufferLine.ts
```diff
@@ -279,4 +302,27 @@ export class BufferLineTypedArray implements IBufferLine {
...
+
+  public translateToString(trimRight: boolean = false, startCol: number = 0, endCol: number = null): string {
+    let length = endCol || this.length;
+    if (trimRight) {
+      length = Math.min(length, this.getTrimmedLength());
+    }
+    let result = '';
+    while (startCol < length) {
+      const stringData = this._data[startCol * CELL_SIZE + Cell.STRING];
+      result += (stringData & 0x80000000) ? this._combined[startCol] : (stringData) ? String.fromCharCode(stringData) : ' ';
```
Tyriar: 0x80000000 should be a constant

src/InputHandler.test.ts
```diff
@@ -210,7 +210,7 @@ describe('InputHandler', () => {
   describe('regression tests', function(): void {
     function lineContent(line: IBufferLine): string {
       let content = '';
-      for (let i = 0; i < line.length; ++i) content += line.get(i)[CHAR_DATA_CHAR_INDEX];
```
Tyriar: When this PR is done will this be removed and all the tests fixed up to trim the null cells?

src/Buffer.test.ts
jerch: :+1:

src/BufferLine.ts
jerch: :+1:

src/InputHandler.test.ts
jerch: Yeah sounds good. If all goes well, we could even set trimming as default behavior for all string retrieving code from the buffer.

src/BufferLine.ts
jerch: Did some tests with a maintained `trimLength` property - for every write to the buffer line the property gets updated to the highest written cell index. Results:
- `translateBufferLineToString` is only ~5% faster (~10 ms saved for getting 86000 lines of `ls -lR /usr/lib` or 210 ms vs 200 ms)
- the property updates are very expensive during the input and add 50 - 70 ms in `InputHandler.print`

This could be tweaked further by omitting perma updates in the loop in `InputHandler.print`, still I am unsure if the overall benefit is worth the trouble as it would need several changes in `InputHandler` methods (all that alter the buffer content).

What concerns me more is the cost of `translateBufferLineToString` itself - it takes 200 ms to convert the buffer back to strings. Thats much better than before (350-400 ms), but it shows that querying strings from the buffer multiple times will be toxic for the runtime. Not sure if we have several "always on" listeners doing that, I am only aware of the linkifier.

src/InputHandler.test.ts
jerch: Changed the test cases (mostly cleanup since the regression tests against the old codebase are not needed anymore).

jerch:
Up for next review. Note that this also needs user testing before it can be added, since it is used at many places (not sure about the test coverage).

src/BufferLine.ts
```diff
@@ -120,6 +143,9 @@ const enum Cell {
   WIDTH = 2
 }
 
+/** single vs. combined char distinction */
+const COMBINED = 0x80000000;
```
Tyriar: `IS_COMBINED_BIT_MASK`?

src/Buffer.ts
```diff
@@ -16,9 +16,13 @@ export const CHAR_DATA_WIDTH_INDEX = 2;
 export const CHAR_DATA_CODE_INDEX = 3;
 export const MAX_BUFFER_SIZE = 4294967295; // 2^32 - 1
 
-export const NULL_CELL_CHAR = ' ';
+// export const NULL_CELL_CHAR = ' ';
```
Tyriar: Should these be kept as WHITESPACE_CELL_CHAR, WHITESPACE_CELL_CHAR and WHITESPACE_CELL_CHAR? You will need to make sure all usages of `NULL_CELL_*` are replaced with `WHITESPACE_CELL_*` when needed.

src/BufferLine.ts
jerch: Done.

src/Buffer.ts
jerch: Yupp, replaced all `|| ' '` with it.

src/Buffer.ts
```diff
@@ -272,64 +276,11 @@ export class Buffer implements IBuffer {
...
    */
   public translateBufferLineToString(lineIndex: number, trimRight: boolean, startCol: number = 0, endCol: number = null): string {
-    // Get full line
-    let lineString = '';
     const line = this.lines.get(lineIndex);
     if (!line) {
       return '';
     }
-
-    // Initialize column and index values. Column values represent the actual
```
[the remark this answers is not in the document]
jerch: Should we keep it for historical reason?

src/renderer/dom/DomRendererRowFactory.ts
```diff
@@ -41,9 +41,9 @@ export class DomRendererRowFactory {
 
     for (let x = 0; x < lineLength; x++) {
       const charData = lineData.get(x);
-      const char = charData[CHAR_DATA_CHAR_INDEX];
-      const attr = charData[CHAR_DATA_ATTR_INDEX];
-      const width = charData[CHAR_DATA_WIDTH_INDEX];
+      const char: string = charData[CHAR_DATA_CHAR_INDEX] || WHITESPACE_CELL_CHAR;
```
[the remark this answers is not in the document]
jerch: :+1:

src/addons/search/search.test.ts
```diff
@@ -116,10 +116,11 @@ describe('search addon', () => {
       expect(tilda2).eql({col: 0, row: 3, term: '~'});
     });
     it('should not select empty lines', () => {
+      // with addition of a real null char printing spaces is not considered empty anymore
```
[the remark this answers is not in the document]
jerch: Wth, its even twice there. The cat, it was the cat. :smile_cat:

PR #1784, opened 2018-11-09
Fixing isClickInSelection to cover missing cases

Tyriar:
The following cases weren't working:

- The first character on single line selections
- The first row on multi-line selection


Tyriar:
@jerch tests done, didn't need to simulate MouseEvents 👍

Issue #1785, opened 2018-11-11
initTable is slow, can we lazily load parts as they're needed?


Tyriar:
On my fairly fast gaming desktop it's over a frame:

![image](https://user-images.githubusercontent.com/2193314/48317850-0e7d5d80-e5ad-11e8-8090-5601994de2ea.png)

It looks like `initTable` could be made so only load chunks of `table` as necessary, splitting it into 32 parts (66636/32=2048 codepoints) would probably make it have minimal impact on frames following a load.

https://github.com/xtermjs/xterm.js/blob/cb97ab35cb753f66e79a3b3eaa9442fbd3747c4d/src/CharWidth.ts#L126-L131

jerch:
@skprabhanjan Its already highly optimized code, the returned function is used in the hot loop during input for every char. Therefore its not that easy to chunkify it without reintroducing a runtime penalty. Also it is not directly testable since v8 typically inlines the function. If you are still with me and still want to do it, these are the steps to be taken:
- See #798 for original PR of this with old impl and some benchmark numbers
- split the lookup table in either several tables or one table whose parts are filled independently with the first occurence of a char within a parts range
- extend the returned function with a condition, that inits the appropriate table (part) and fix the used table / offset
- compare the runtime of `InputHandler.print` before and after the change with some input data, either via chrome's timeline view (less reliable) or with https://github.com/xtermjs/xterm-benchmark (more reliable)

There are several tricks possible to further "hide" the workload (maybe use `requestIdleCallback` in the browser).
The original PR #798 saved ~10% runtime for ASCII chars and >80% for CJK chars in `InputHandler.print` compared to the old implementation (dont remember the exact values). An acceptable solution should not penalize the current runtime of `InputHandler.print` more than by 5-10% imho (must be tested with the different input char ranges).

Oh - and the lookup table is highly packed to avoid wasting memory. Would be good if this does not end up taking much more memory.

skprabhanjan:
@jerch , Thanks for the reply.
I would need more help from your side going forward with this.
I tried understanding this for quite some time and I shall try to understand this more deeply but for now i need clarification for this question

> split the lookup table in either several tables or one table whose parts are filled independently with the first occurence of a char within a parts range

If i choose to go with the second approach then what exactly does **first occurence of a char within a parts range** would mean ?
Thanks :)

jerch:
@skprabhanjan I think the most straight forward way is to do the following:

[...]

PR #1789, opened 2018-11-18
Optimisation for wcwidth

jerch:
Handles things discussed in #1785:

- optimized table creation (down to under 0.5 ms)
- runtime speedup
- fix slighty off surrogate pair handling

The PR comes with a regression test against the old implementation (can be removed with a later release). Also currently the table is created on lib loading, this should be changed to an explicit table creation by some managment code to avoid unused tables with multi unicode support. Imho this can be addressed in #1714 as well.


src/InputHandler.ts
```diff
@@ -327,7 +327,7 @@ export class InputHandler extends Disposable implements IInputHandler {
   public print(data: string, start: number, end: number): void {
     let char: string;
     let code: number;
-    let low: number;
+    let second: number;
```
Tyriar: Let's move the declaration lower and make it a `const`

src/CharWidth.ts
```diff
@@ -120,50 +88,49 @@ export const wcwidth = (function(opts: {nul: number, control: number}): (ucs: nu
...
-          num = (num << 2) | wcwidthBMP(CODEPOINTS_PER_ITEM * i + pos);
-        }
-        table[i] = num;
-      }
-      return table;
+
+    // create lookup table for BMP plane
+    // TODO: make callable/configurable from UnicodeManager
+    const table = new Uint8Array(65536);
+    table.fill(1);
```
Tyriar: While I think we should move to typed arrays only, we can't use fill (this impacts my upcoming PR too 😢) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray#Browser_compatibility

src/CharWidth.test.ts
```diff
@@ -77,3 +77,179 @@ describe('getStringCellWidth', function(): void {
   });
   // TODO: multiline tests once #1685 is resolved
 });
+
+describe('wcwidth regression', function(): void {
+  // TODO: remove with one of the next releases
```
Tyriar: Let's remove this TODO and change the test to:

```
it('should match all values from the old implementation'
```

No harm keeping it in for the long run.

Issue #1614, opened 2018-08-14
Dynamic texture atlas can contain cut off characters


Tyriar:
https://github.com/Microsoft/vscode/issues/55981

May be related to switching monitors with different DPI

![image](https://user-images.githubusercontent.com/2193314/44112452-fd08dcca-9fb9-11e8-85e6-7932851b2619.png)

Tyriar:
No plans to implement this for canvas renderer as we'll move towards the webgl addon that was just moved eventually

Issue #1389, opened 2018-04-16
In DynamicCharAtlas, don't store empty characters in the texture


bgw:
Discussion here: https://github.com/xtermjs/xterm.js/pull/1327#discussion_r179231675

We don't need to save anything to the atlas texture if a glyph is empty, we just need to keep track of the fact that it was empty. This may or may not be worth the added complexity, but we should try and see.

Tyriar:
No plans to implement this for canvas renderer as we'll move towards the webgl addon that was just moved eventually

Issue #955, opened 2017-09-08
Startup and canvas render window resize that updates devicePixelRatio is expensive


Tyriar:
After https://github.com/sourcelair/xterm.js/pull/938

Repro:

1. Launch xterm.js

---

1. Open demo
2. Start performance timeline 
3. Press cmd/ctrl and + 3 times

<img width="147" alt="screen shot 2017-09-07 at 9 57 09 am" src="https://user-images.githubusercontent.com/2193314/30194318-4a5ded40-9407-11e7-8496-875e203375e5.png">

The reason this happens is because it redraws the char atlas, that's also the reason the time gets greater the more cmd/ctrl + is pressed (more pixels). This could be done using `OffscreenCanvas` and web workers for browsers that support?

Tyriar:
The best solution to these two issues is to pull the CharAtlas generation into a web worker. This is blocked on OffscreenCanvas support so that we can draw on a worker thread.

The great thing about this solution is that the terminal can start drawing immediately just fine, the CharAtlas is just an optimization that can come in whenever and it will be picked up and used from that point on,.

parisk:
Sounds good to me.

Just one question. From what I understand, the CharAtlas needs to be generated every time the zoom factor changes.

If this is true, is the web worker going to be an one-off process or something like a daemon waiting to be requested to generate a new CharAtlas? I think an one-off process would introduce less memory overhead overall.

Tyriar:
One off process is what I'm thinking as it's generally a pretty infrequent thing. If they take a little while to spin up then we could add some time after it's stopped before terminating? (eg. ctrl++ multiple times)

Tyriar:
There's an example of generating the char atlas using a web worker at https://github.com/Tyriar/xterm.js/tree/955_worker. I'll probably leave it there for now and wait for Electron to get a version out that supports `OffscreenCanvas`.

mofux:
Nice! You can use `OffscreenCanvas` in `electron` already by enabling the `experimentalCanvasFeatures `:

```javascript
win = new BrowserWindow({
    width: 1200,
    height: 700,
    webPreferences: { experimentalCanvasFeatures: true }
})
```

https://stackoverflow.com/questions/47271240/can-experimental-features-like-offscreenanvas-be-enabled-in-electron

Tyriar:
[...]

Issue #941, opened 2017-09-04
Do incremental draws in the selection render layer


Tyriar:
Currently every mousemove the canvas is cleared and redrawn.

Tyriar:
No plans to implement this for canvas renderer as we'll move towards the webgl addon that was just moved eventually

PR #1790, opened 2018-11-20
WebGL Renderer

### LINKED ISSUE #720: Selection should not change text color

### LINKED ISSUE #1138: Tall characters have their top and/or bottom cut off
This applies to tall g's and some other characters on some fonts and also to diacritical marks in various languages.

VS Code issue: https://github.com/Microsoft/vscode/issues/35901

### LINKED ISSUE #1170: Default fonts are vertically offset on some Linux distributions
VS Code issue: https://github.com/Microsoft/vscode/issues/36322

image

Getting around this will probably involve drawing a character and checking the pixels to see where it was drawn in the cell and adjusting as necessary. From what I can see this just happens on some fonts which happen to be default on Kali and Debian.

Tyriar:
This PR introduces the `'webgl'` renderer which is basically canvas but much lower level and much faster. I see this eventually replacing the canvas renderer but it will need some time before that happens.

## But there's already a canvas renderer?

To illustrate the difference when drawing, the canvas renderer basically does this:

1. Tell the GPU to draw character `0,0` from texture `a`
2. Tell the GPU to draw character `1,0` from texture `a`
3. Tell the GPU to draw character `2,0` from texture `a`
4. and so on.

These individual instructions are costly, the webgl renderer instead builds a `Float32Array` containing all the data needed to draw and a "webgl program" (a vertex shader and a fragment shader) that knows how to draw things from the `Float32Array` is uploaded to the GPU which does the actual drawing. Of course the details are much more involved than this, but that's the basics of why the webgl renderer will be superior to the canvas renderer.

## Features

- It's super fast, it also scales much better with really large viewports.
- _All_ characters are cached in the atlas, including unicode, combined chars (emoji), etc.
- The atlas trims glyphs to their minimal rectangles and stores that in the texture, this means better space utilization of the texture. Here's a smaller 256x256 example illustrating this:
  image
- Selection is now drawn underneath the text (https://github.com/xtermjs/xterm.js/issues/720), using a uniform selection fg for accessibility reasons:
  screen shot 2018-11-20 at 8 34 18 am
- Since the buffer is redrawn every frame, characters are no longer trimmed at the top and the bottom (https://github.com/xtermjs/xterm.js/issues/1138).
  screen shot 2018-11-20 at 8 36 41 am
  screen shot 2018-11-20 at 8 37 00 am
- Renderer pausing when the canvas is offscreen is supported

## Notes

- WebGL2 is used, this reduces browser compatibility a little but allows using some niceties like vertex array objects and drawing using `drawElementsInstanced`.
- Render layers still work, the webgl renderer actually still utilizes the `CursorRenderLayer` and `LinkRenderLayer` in order to simplify the rendering logic (and because they're pretty fast anyway).
- When the texture atlas has reached its capacity, it will currently just clear itself and restart. This could be improved but this route was chosen due to the relatively inexpensive cost of starting over and so we don't need to juggle multiple textures.
- I experimented with disabling double buffering and drawing the minimal set of changes, it didn't appear to be worth the effort (at least given the time I had to work on this).

## Issues

🛑= merge blocker

- Currently you cannot customize the selection color. This is because cached glyphs do not currently support true color, this should be trivial to fix once true color is added.
- 🛑 Character joiners are not supported yet.
- 🛑 `allowTransparency` is not supported yet.
- Rendering seems to become offset on my macbook when rows are increased after 300x91, I'm guessing this is related to maximum canvas dimensions?

### Summary

This table shows the average time to draw the frame before and after, see below for more details.

| Benchmark | Canvas/dynamic (avg ms/frame) | WebGL (avg ms/frame) | Change (c/w-1)
|---|---|---|---
| Macbook 87x26 | 4.80 | 0.69 | **596% faster**
| Macbook 300x80 | 15.28 | 3.69 | **314% faster**
| Windows 87x26 | 7.31 | 0.73 | **901% faster**
| Windows 300x80 | 19.34 | 2.06 | **839% faster**
| Macbook 87x26 CJK | 14.63 | 5.93 | **147% faster**
| Macbook 87x26 Emoji | 27.47 | 19.28 | **42% faster**

### Methodology

<details>
The tests are done by running various commands in the build xterm.js repo directory and the following measuring code is injected to get the report:

```ts
(<any>window).frames = [];
(<any>window).report = () => {
  const frames = (<any>window).frames as number[];
  const average = Math.round(frames.reduce((p, c) => p + c, 0) / frames.length * 100) / 100;
  console.log(`...: frames ${frames.length}, average ${average}ms`);
};
```

Wrapping `_renderRows` of Renderer.ts (canvas) and WebglRenderer.ts:

```ts
const startTime = performance.now();
...
(<any>window).frames.push(performance.now() - startTime);
```

Parts of the chrome timelines are also presented. Note that the injected code is necessary because it's difficult to measure otherwise as the number of frames and average frame time isn't captured by the timeline.

The Macbook Pro is mid 2014 with an Intel Iris Pro 1536 MB, the Windows box has a GTX 760.

</details>

#### Canvas Renderer (dynamic atlas)

<details>
Canvas: frames 69, average 4.8ms

screen shot 2018-11-20 at 8 07 28 am
</details>

#### WebGL Renderer

<details>
WebGL: frames 80, average 0.69ms

screen shot 2018-11-20 at 8 05 24 am
</details>

#### Canvas Renderer (dynamic atlas)

<details>
Canvas: frames 30, average 15.28ms

screen shot 2018-11-20 at 8 09 28 am
</details>

#### WebGL Renderer

<details>
WebGL: frames 43, average 3.69ms

screen shot 2018-11-20 at 8 10 27 am
</details>

#### Canvas Renderer (dynamic atlas)

<details>
Canvas: frames 54, average 7.31ms

image
</details>

#### WebGL Renderer

<details>
WebGL: frames 60, average 0.73ms

image
</details>

#### Canvas Renderer (dynamic atlas)

<details>
Canvas: frames 59, average 19.34ms

image
</details>

#### WebGL Renderer

<details>
WebGL: frames 68, average 2.06ms

image
</details>

### Macbook Viewport 87x26, `cat zh_wcag_100.txt`

zh_wcag_100.txt contains this contents of https://www.w3.org/Translations/WCAG20-zh/ 100 times to help test throughput for wide characters.

#### Canvas Renderer (dynamic atlas)

<details>
Canvas: frames 40, average 14.63ms

screen shot 2018-11-19 at 5 06 10 pm
</details>

#### WebGL Renderer

<details>
WebGL: frames 45, average 5.93ms

screen shot 2018-11-20 at 8 13 02 am
</details>

### Macbook Viewport 87x26, `cat emoji_100.txt`

emoji_100.txt contains this contents of https://getemoji.com/ times to help test throughput for combined characters.

#### Canvas Renderer (dynamic atlas)

<details>
Canvas: frames 13, average 27.47ms

screen shot 2018-11-20 at 8 22 03 am
</details>

#### WebGL Renderer

<details>
WebGL: frames 16, average 19.28ms

screen shot 2018-11-20 at 8 20 22 am
</details>

---

Fixes #720
Fixes #1138 (Microsoft/vscode#35901)
Fixes #1170 (I think)
Obsoletes a bunch of issues related to the canvas renderer: #941, #955, #1389, #1614


jerch:
@Tyriar Have not yet looked at the code but could not resist to do a few tests on my own. Therefore first a few notes about the perf results.

I see about the same perf boost (rendering being 3-5x faster) on Linux with either intel graphics or nvidia GT520M (very old) and GTX 960. With my typical `ls -lR /usr/lib` benchmark I also hit the system throughput, means the terminal has to wait in between for the pty to deliver more data, and is idle the rest of the time (both systems with SSDs lol). That are really great results! 
Your timelines show weak performance for windows systems, really slow incoming data, hmm. This might be worth to be addressed upstream in win-pty/conpty.

Your results also show a big perf decrease on the MacBook for the bigger terminal view. I wonder how fullscreen would perform. Note that I cant resemble this decrease with Linux (even tried 500x100), I see only slightly worse runtime for intel or nvidia gc. Maybe thats a resolution problem (retina?) and could be tweaked further?

What I see now for my benchmark in the demo:
- total runtime: ~1800 ms (seems this cannot go faster due to system limits)
- ~ 120 ms render time (was 500 ms before) :heart_eyes:
- ~ 900 ms for the input chain (parser, buffer inserts, state handling) :unamused: 
- ~ 200 ms JS overhead (function calls, GC etc.)
- ~ 600 ms idle :heart_eyes:

Last but not least: since many peeps nowadays are on laptops - does anyone know if the gl renderer would drain more power than the canvas renderer? Note that the canvas renderer also relies on hardware acc, still unsure whether this might trigger a more costly power profile.

Tyriar:
On support, according to https://webglstats.com/ 68% of desktops support WebGL2. I looked into downgrading to WebGL1 (98% support) but there doesn't appear to be an extension for `layout`. I'm thinking we should push forward with replacing the canvas renderer with the WebGL2 implementation once stable and improve the DOM renderer a little so it has more reasonable performance, that way the majority of consumers will be on the best implementation, we don't muck the code up with WebGL1 messiness (+slowness?) and there is a viable alternative for the rest.

jerch:
@Tyriar Yeah seems only Safari cant do Webgl2, I am not sure if Safari will ever see this as Apple is pushing WebGPU based on Metal/Vulkan. Webgl1/2 is based on OpenGL which got deprecated on macOS.

Tyriar:
Currently experimenting with minimizing the amount of data uploaded to the GPU via `bufferData`, results looks favorable so far, especially when scrolling.

nojvek:
It makes me so happy to see xterm get even faster. 

I played with https://github.com/jwilm/alacritty a while back where the author claims it’s the fastest terminal due to it being written in rust and uses webgl for rendering. 

Terminals as a technilogy have been around for decades, but radically speeding them up, giving 256 bit color support, mouse support, accessibility e.t.c make it possible to target terminal as a platform for light weight apps with great UX.

src/renderer/webgl/GlyphRenderer.ts
```diff
@@ -0,0 +1,353 @@
...
+      }
+
+      let i = 0;
+      for (let y = 0; y < terminal.rows; y++) {
+        for (let x = 0; x < terminal.cols; x++) {
+          i += 8;
+          this._vertices.attributes[i++] = x / terminal.cols;
+          this._vertices.attributes[i++] = y / terminal.rows;
+        }
+      }
```
jerch: Eww I find this hard to read. It seems it adds at position 8 and 9 the col/row offset, then every 10th the same for the next cols/rows. But your vertex array definition is 12 elements long? (kinda same question as above)

src/renderer/webgl/GlyphRenderer.ts
```diff
@@ -0,0 +1,353 @@
...
+
+    const gl = this._gl;
+
+    gl.useProgram(this._program);
+    gl.bindVertexArray(this._vertexArrayObject);
+
+    this._activeBuffer = (this._activeBuffer + 1) % 2;
+    this._vertices.attributesBuffers[this._activeBuffer];
+    let bufferLength = 0;
+    for (let y = 0; y < renderModel.lineLengths.length; y++) {
```
jerch: Dont quite get what this loop used for - to merge the selection attrs onto the vertices?

src/renderer/webgl/GlyphRenderer.ts
```diff
@@ -0,0 +1,353 @@
...
+      const si = y * this._terminal.cols * INDICES_PER_CELL;
+      const sub = (isSelectionVisible ? this._vertices.selectionAttributes : this._vertices.attributes).subarray(si, si + renderModel.lineLengths[y] * INDICES_PER_CELL);
+      this._vertices.attributesBuffers[this._activeBuffer].set(sub, bufferLength);
+      bufferLength += sub.length;
+    }
+    gl.bindBuffer(gl.ARRAY_BUFFER, this._attributesBuffer);
+    const buffer = this._vertices.attributesBuffers[this._activeBuffer].subarray(0, bufferLength);
+    gl.bufferData(gl.ARRAY_BUFFER, buffer, gl.STREAM_DRAW);
```
jerch: Was hoping to get some speedup by using glBufferSubData but didnt change anything. And glMapBuffer does not work in webgl.

src/renderer/webgl/GlyphRenderer.ts
```diff
@@ -0,0 +1,353 @@
...
+out vec2 v_texcoord;
+
+void main() {
+  vec2 zeroToOne = (a_offset / u_resolution) + a_cellpos + (a_unitquad * a_size);
+  gl_Position = u_projection * vec4(zeroToOne, 0.0, 1.0);
+  v_texcoord = a_texcoord + a_unitquad * a_texsize;
+}`;
+
+const fragmentShaderSource = `#version 300 es
+precision lowp float;
```
jerch: Yeah low precision should be enough here, still we might have to test whether this leads to color artefacts on high res display with big font sizes.

src/renderer/atlas/CharAtlasCache.ts
```diff
@@ -54,6 +56,7 @@ export function acquireCharAtlas(
       }
       // The configs differ, release the terminal from the entry
       if (entry.ownedBy.length === 1) {
+        entry.atlas.dispose();
```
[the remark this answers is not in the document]
Tyriar: This should go away when I update the branch 👍

src/renderer/webgl/GlyphRenderer.ts
```diff
@@ -0,0 +1,353 @@
...
+}
+
+const enum VertexAttribLocations {
+  UNIT_QUAD = 0,
+  CELL_POSITION = 1,
+  OFFSET = 2,
+  SIZE = 3,
+  TEXCOORD = 4,
+  TEXSIZE = 5
+}
```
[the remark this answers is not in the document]
Tyriar: > Below you dont recalc CELL_POSITION for every char - could we remove this from the array?

Originally there were 2 buffers, one which had cell positions which was only changed and bound when the terminal resized. But this changed when I went from keeping a buffer of the entire buffer to only redrawing cells that needed it (basically trimming the right whitespace). I don't think we can keep this in a separate array anymore.

src/renderer/webgl/GlyphRenderer.ts
Tyriar: Seems to work fine from my minimal testing on the macbook

src/renderer/webgl/GlyphRenderer.ts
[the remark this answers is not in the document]
Tyriar: For size I need to look into the difference between a_size and a_texsize again as I think they might be able to be combined. But we would still need to keep them around as the textures have varying sizes due to the way they're packed into the atlas.

What I was explaining in my earlier comment was about CELL_POSITION changes very often now due to the way we pack the buffer to send over, for example for this:

```
ls
blah
```

We send only 6 cells, even if cols=100. Meaning CELL_POSITION won't line up if we had it in a different buffer that only changes on resize.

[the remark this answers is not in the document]
jerch: Ah ok, so the atlas glyphs vary in size. And about cell_pos - yeah I wondered about that in the render method, see my comment below about selection, now i get it :blush:.
Yeah so this seems much better memory wise than my idea with reducing the values per glyph (given that most lines contain alot of empty cells to the right). :+1:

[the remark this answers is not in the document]
Tyriar: Oh I remember, a_texsize is the texture size in 0-1 coords in the atlas texture, a_size is the size in 0-1 coords in the canvas texture. We could merge them into a single pixel size and then get the clipspace on the GPU. This would slightly reduce math on CPU, reduce the buffer size and increase math on GPU.

I think my original thinking was to reduce repetitive work on the GPU by calculating a_texsize once for each texture instead of doing it every single time the glyph is displayed. Ideally we would have a separate buffer that has the texture information pre-calculated on the GPU (as a uniform?) and that would be referenced in the other buffer. Seemed tough to do at the time.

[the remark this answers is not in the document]
jerch: Not sure what works here better, imho perfwise its already good as it is, only those green streaks are a bit bothersome, but I have no clue where they come from and I cannot repo them under Linux. From the workload perspective the shaders have not much to do atm (esp. the vertices operating on glyph ganularity). I only see some workload on the CPU, but its neglectible compared to the workload for the canvas renderer. So this might be a first world problem :laughing:

juancampa:
An issue I just found is when using `Terminal.setOption`, to change `rendererType` and `experimentalCharAtlas`: If you do it in the wrong order it will throw. Sounds like, for this case, we should be able to change two options atomically (i.e. without calling `Terminal.refresh` in between)

Tyriar:
@juancampa the plan's to remove the canvas renderer and removing experimentalCharAtlas all together, so I haven't bothered fixing this issue. The workaround on your end should work for now and that can be removed when this is merged.

vincentwoo:
Having now read all the threads on rendering issues recently, I'm pleased to see a solution in the works. I can't help but be a little wary of switching everything over to webgl with DOM rendering as a fallback, though. Webgl can be pretty finnicky, and browser support is at the point where we'd see half of users on the DOM implementation, which is quite a bit slower than canvas. Is implementing a more correct canvas implementation untenable?

Tyriar:
@vincentwoo well the webgl one is essentially the better canvas renderer, many of the issues can't be worked around in a good way for the old canvas renderer.

I did make the DOM renderer a lot faster by fixing https://github.com/xtermjs/xterm.js/issues/1609, I feel like it's good enough from a perf perspective now (many people on VS Code use it).

vincentwoo:
Okay, if you feel good about it, I feel good about it too. Any sense for when this might come out? I was unable to hack around the character clipping issues with the hack suggested here (https://github.com/xtermjs/xterm.js/issues/1138#issuecomment-410874958) on the latest release.

Tyriar:
@vincentwoo no ETA. This is a spare time project so it needs the mood to strike on the weekend 😛

src/renderer/webgl/GlyphRenderer.ts
```diff
@@ -0,0 +1,353 @@
...
+      rasterizedGlyph = this._atlas.getRasterizedGlyphCombinedChar(chars, attr, bg, fg, this._terminal.options.enableBold);
+    } else {
+      rasterizedGlyph = this._atlas.getRasterizedGlyph(code, attr, bg, fg, this._terminal.options.enableBold);
+    }
+
+    // Fill empty if no glyph was found
+    if (!rasterizedGlyph) {
+      fill(array, 0, i, i + INDICES_PER_CELL - 1 - CELL_POSITION_INDICES);
+      return;
+    }
```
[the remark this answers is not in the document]
Tyriar: Tweaked it to never return undefined for `rasterizedGlyph`, previously it could return undefined for code 32, but that case is handled above.

src/renderer/webgl/GlyphRenderer.ts
Tyriar: This one uses `INDICES_PER_CELL` which is set to 10, it certainly is a little confusing though, fixing.

src/renderer/webgl/GlyphRenderer.ts
Tyriar: Good point, added thorough comments https://github.com/xtermjs/xterm.js/pull/1790/commits/01b37b2fef2acfc9a1ef6f6e700d1823c4fbf993

src/renderer/webgl/GlyphRenderer.ts
Tyriar: Yeah I looked into using subdata but it seems to end up being slower in most cases.

j-f1:
Small bug with black text on a white background:

screen shot 2018-12-31 at 18 34 47

Tyriar:
@j-f1 that's @juancampa's workaround, it's not part of this PR. I think he's planning on doing a simpler solution to just invert to colors and do a PR to this branch.

Tyriar:
@j-f1 I mean the black border isn't in this PR (and won't be), it's in the Hyper fork, so this discussion is off-topic.

j-f1:
I’m talking about these issues, not the black border:

94d23e22-2c1f-45ad-b01d-30f571486ea4

Is that still an issue with the changes in Hyper?

Tyriar:
@j-f1 I think so, as it only happens within the border

src/renderer/webgl/WebglRenderer.ts
```diff
@@ -0,0 +1,435 @@
...
+    // and resume based on terminal visibility if so
+    if ('IntersectionObserver' in window) {
+      const observer = new IntersectionObserver(e => this.onIntersectionChange(e[0]), { threshold: 0 });
+      observer.observe(this._terminal.element);
+      this.register({ dispose: () => observer.disconnect() });
+    }
+  }
+
+  public dispose(): void {
+    this._renderLayers.forEach(l => l.dispose());
```
juancampa: missing `super.dispose()`

Tyriar: Indeed it is, good find!

Tyriar:
@jerch I think in general the DOM renderer should be good enough compared to the canvas software fallback, plus it won't have https://github.com/xtermjs/xterm.js/issues/1138

paperkettle:
A feature request: it would be tremendous if this design could allow for crispness at arbitrary levels of zoomability. I think it would see a lot of use in 3D libraries (threejs, a-frame, babylon, etc)

As a rudimentary proof of concept hack, I sent some of the current canvas renderer's contexts into three.js as textures & made an a-frame component. I hit the expected levels of unusable blurriness, but having a terminal inside ar & vr showed promise of livecoding the space from inside itself. Hot module swapping 🔥✨

I haven't fully wrapped my head around the sdf examples floating about, and I've just started looking at things like https://github.com/SeregPie/THREE.TextSprite, though I'm unsure if these would be useful given the current architecture, which I'm also starting to familiarize myself with.

If zoomy crispness meant a performance hit, perhaps it could be an optional mode if it didn't open a can of complexity worms? Or maybe just the renderer api could consider this as a hypothetical future renderer someone might help make. Thinking out loud / will do more research

Tyriar:
@paperkettle the textures for all characters are exactly as large as they need to be given the current `window.devicePixelRatio`. If you change the "canvas size" in any way that will result in blurriness, the textures would need to be much larger to get rid of most of the blurriness and that would a bunch more work on the CPU to create the textures and much a larger texture atlas being passing to the GPU.

As you've probably read though the plan is to move this into an addon and also its own repo, so you could fork and tweak as necessary at that point. There's still a few large issues blocking that from happening though (like https://github.com/xtermjs/xterm.js/issues/1128, https://github.com/xtermjs/xterm.js/issues/1994 and https://github.com/xtermjs/xterm.js/issues/2005).

PerBothner:
It would great if it were possible to mix WebGL and Dom rendering, so most "simple" lines are rendered using WebGL but "complex" lines could be rendered as DOM nodes.  Or some other way to add images and "raw HTML" to the output, including "non-standard height" lines as in issue #1901. Also issue #1140.

Just a wishlist item I hope you'll keep in mind.  It would make it tempting to switch DomTerm to using xterm.js, and I could work on fixing vttest and other little bugs.

Tyriar:
@PerBothner for decoration-type things like https://github.com/xtermjs/xterm.js/issues/1852, these would still be DOM-based as it's a little unreasonable to expect consumers to write canvas and and DOM instructions, also if we supported some way of displaying images it would have to be supported here too once we ship it.

Tyriar:
Underlines may not be working right? https://github.com/zeit/hyper/issues/3691

Eugeny:
I've been experimenting with it a bit, and it seems like every now and then it'll "skip" the FG color reset escape sequence. Can't pinpoint the exact case where this happens yet.

Tyriar:
@Eugeny that might be this issue which is still on this branch? https://github.com/xtermjs/xterm.js/issues/2196

Eugeny:
Looks like it. If you could rebase `webgl2` against master, I'll roll out a Terminus update with it.

Also, there have been reports of font/background misalignment: https://github.com/Eugeny/terminus/issues/1052

Tyriar:
@Eugeny will update the branch soon, maybe on the weekend if I get some time.

Tyriar:
Updated just now, seems to work.

TODO:

- [ ] Update imports to absolute
- [ ] Adopt new services in webgl component

Eugeny:
Brilliant, thanks!

Eugeny:
Judging by the cursor position, it's the foreground text that's offset to the top: 
image

(here, the triangles are characters with orange FG)

Tyriar:
I was just thinking about how we could move this PR forward as right now it's kind of stuck unless I miraculously get a bunch of time to devote to it. Here are the main issues:

- No one can contribute, well it's more difficult at least because you would need to fork my fork
- The PR breaks frequently and needs to get conflicts fixed regularly, particularly because of all the #1507 refactors
- It's seems to be hard to pull the WebGL into an addon because of so many dependencies internally
- The renderer API #2005 would end up being enormous

Here's my plan to fix the issues:

- Do not attempt to expose a renderer API as public API, instead renderers are only built by the core team
- [ ] Create `src/xterm-internals.d.ts` that `Terminal` (common) implements, this will be the mechanism for exposing private API such as `_renderService` to "internal" addons
- [ ] Convert Types.ts to .d.ts files, this will allow them to be imported and insure no code is pulled in as a result (this is something I wanted to do anyway)
- [ ] Pull constants required by the renderers such as `DEFAULT_COLOR` into separate files
- [ ] Move the WebGL addon into `addons/xterm-addon-webgl`
- [ ] Have the addon depend directly on certain self-contained files like TypedArrayUtils and Lifecycle (might need some work in path mappings/webpack to do this)
- [ ] Allow loading the WebGL addon in the demo
- [ ] Write some basic tests that activate the addon

Eugeny:
Sounds perfectly fine to me. While having a renderer API would be nice from the design standpoint, it's very unlikely that any third-party renderers would ever pop up.

Eugeny:
Moving the cursor seems to leave some ghosting behind in the shape of black scaled-up characters: 
image

Eugeny:
After a display DPI switch: 
image

Eugeny:
After a display DPI switch: 
image

Tyriar:
Apparently my recent merge really messed things up, thanks for pointing out the issues @Eugeny

Tyriar:
List of things needed by WebGL (excluding Lifecycle/TypedArrayUtils):

common/Types

```
DEFAULT_COLOR
ICellData
```

common/buffer/BufferLine

```
NULL_CELL_CODE,
WHITESPACE_CELL_CODE
CHAR_DATA_CHAR_INDEX,
CHAR_DATA_CODE_INDEX,
CHAR_DATA_ATTR_INDEX
CellData
DEFAULT_ATTR
AttributeData
```

src/renderer/Types

```
FLAGS
```

src/Types

```
ILinkifierAccessor
ILinkifierEvent
ITerminal
```

Tyriar:
I bisected the devicePixelRatio but and it happened between cf86760 and f42e53f.

Tyriar:
https://github.com/xtermjs/xterm.js/pull/2228 fixes thee canvas one but it still doesn't fix the webgl issue

Tyriar:
Ok the weird dpr thing is fixed in 2584de8

Tyriar:
The webgl renderer is now a proper addon in this PR, just need to work on minimizing the diff outside of the addons/ folder, which includes moving the webgl renderer over to CellData to get rid of the CharData compat stuff in BufferLine.ts.

Tyriar:
This is ready for review/merge, it's still a work in progress as these won't work:

- Underline attribute
- Selection color
- Character joiners
- `allowTransparency`
- Handle context loss when there are too many contexts (see https://github.com/Tyriar/xterm.js/pull/5)

However it's a standalone addon now, the only thing that affects the core is the very small diff outside of the `addons/` directory. Once merged in we'll get the build and tests in CI to prevent future breakages and we can create issues to finish off the above issues.

addons/xterm-addon-webgl/src/GlyphRenderer.ts
```diff
@@ -0,0 +1,367 @@
...
+
+    // Exit early if this is a null/space character
+    if (code === NULL_CELL_CODE || code === WHITESPACE_CELL_CODE || code === undefined/* This is used for the right side of wide chars */) {
+      fill(array, 0, i, i + INDICES_PER_CELL - 1 - CELL_POSITION_INDICES);
+      return;
+    }
+
+    let rasterizedGlyph: IRasterizedGlyph;
+    if (chars && chars.length > 1) {
+      rasterizedGlyph = this._atlas.getRasterizedGlyphCombinedChar(chars, attr, bg, fg);
```
jerch: Maybe we could do a similar workaround for RGB colors - just recreate them for every occurance? (Perf would go downhill, but at least it would show something with the right colors.)

addons/xterm-addon-webgl/src/Platform.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) 2016 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+const isNode = (typeof navigator === 'undefined') ? true : false;
+const userAgent = (isNode) ? 'node' : navigator.userAgent;
+
+export const isFirefox = !!~userAgent.indexOf('Firefox');
+export const isSafari = /^((?!chrome|android).)*safari/i.test(userAgent);
```
jerch: Is this used anywhere?

addons/xterm-addon-webgl/src/TypedArray.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  | Float32Array | Float64Array;
+
+function deepEquals(a: TypedArray, b: TypedArray): void {
+  assert.equal(a.length, b.length);
+  for (let i = 0; i < a.length; ++i) {
+    assert.equal(a[i], b[i]);
+  }
+}
+
+describe('polyfill conformance tests', function(): void {
```
jerch: Maybe we should move this and `sliceFallback into `TypedArrayUtils`?

addons/xterm-addon-webgl/src/TypedArray.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Copyright (c) 2018 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+export type TypedArray = Uint8Array | Uint16Array | Uint32Array | Uint8ClampedArray
+  | Int8Array | Int16Array | Int32Array
+  | Float32Array | Float64Array;
+
+export function slice<T extends TypedArray>(array: T, start?: number, end?: number): T {
```
jerch: We should move this to `TypedArrayUtils`, I think we gonna need this more often once I am done with removing the string interfaces from InputHandler.

addons/xterm-addon-webgl/src/WebglAddon.ts
```diff
@@ -0,0 +1,28 @@
...
+  constructor(
+    private _preserveDrawingBuffer?: boolean
+  ) {}
+
+  public activate(terminal: Terminal): void {
+    if (!terminal.element) {
+      throw new Error('Cannot activate WebglRendererAddon before Terminal.open');
+    }
+    const renderService: IRenderService =  (<any>terminal)._core._renderService;
+    const colors: IColorSet = (<any>terminal)._core._colorManager.colors;
```
jerch: Is there a possible race consition between these two assignments and `_core._renderService` and `_core._colorManager` not yet being defined? Like some lazy setups? Or is `activate` guaranteed to be executed only after those are set?

addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
```diff
@@ -0,0 +1,403 @@
...
+  private _getBackgroundColor(bg: number): IColor {
+    if (this._config.allowTransparency) {
+      // The background color might have some transparency, so we need to render it as fully
+      // transparent in the atlas. Otherwise we'd end up drawing the transparent background twice
+      // around the anti-aliased edges of the glyph, and it would look too dark.
+      return TRANSPARENT_COLOR;
+    } else if (bg === INVERTED_DEFAULT_COLOR) {
+      return this._config.colors.foreground;
+    } else if (is256Color(bg)) {
+      // TODO: Need to do a separate set for 16 color palette?
```
jerch: Well as far as I am aware no other emulator separates P16 colors, they kinda all link P16 into the lower P256 segment. Thus I think we should not do that.

addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
```diff
@@ -0,0 +1,403 @@
...
+      return this._getColorFromAnsiIndex(bg);
+    }
+    // TODO: Support true color
+    return this._config.colors.background;
```
jerch: Just for clarification - this will always return default fg/bg for any true color?

Eugeny:
Could you please republish the addon with the fixed typings file?

addons/xterm-addon-webgl/src/CharDataCompat.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Copyright (c) 2019 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { CellData } from 'common/buffer/CellData';
+import { FLAGS } from './Constants';
+import { IBufferLine } from 'common/Types';
+
+export function getCompatAttr(bufferLine: IBufferLine, index: number): number {
```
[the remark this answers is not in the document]
Tyriar: It doesn't yet, true color just results in weird colors for now. I added a fix to stop it crashing in https://github.com/xtermjs/xterm.js/pull/1790/commits/7148358acf1fe4345393b6d288ed931c17e6699c

addons/xterm-addon-webgl/src/GlyphRenderer.ts
Tyriar: :+1: yeah true color support should be relatively easy once the glyph map attribute key stuff is done

Tyriar:
@Eugeny done

Eugeny:
Looks like emojis are not considered to be double width chars anymore?
image

Tyriar:
@Eugeny I didn't think that ever worked because we don't detect unicode version we have to guess?

addons/xterm-addon-webgl/src/Platform.ts
Tyriar: That whole file isn't used apparently, good find 👍

addons/xterm-addon-webgl/src/TypedArray.test.ts
Tyriar: I actually just moved them into xterm-addon-webgl as it's only used there, we can always move them over to core if we need to.

addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
Tyriar: Removed comment

addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
Tyriar: Yes until true color gets supported, I'll be creating a bunch of issues for the webgl renderer and tracking them there.

addons/xterm-addon-webgl/src/TypedArray.ts
Tyriar: As said in the test file, we can always move it over when we need it, right now it's not used so it shouldn't be included in the core lib

addons/xterm-addon-webgl/src/WebglAddon.ts
Tyriar: Since we throw when `terminal.element` isn't around above this should always be defined. This is one of the few places where we do dodgy type stuff by pulling private members 😄

Tyriar:
Let's do this! 🚀 

I've created a bunch of issues tracking the remaining issues under the `addon/webgl` label. Community welcome to help finish it off 😃. If I missed something please create another issue.

PR #1796, opened 2018-11-23
support typed array in parser

jerch:
This PR moves the parser to typed array support instead of JS strings.
Changes the internal API for the `parse` method, the print handler and the DCS handler.

Parser speedup: runtime from 120 ms down to 45 ms (`ls -lR /use/lib`)

With a later change of the `BufferLine` API this enables the possibility to completely remove any string processing from `InputHandler.print`, which will give the input chain another nice speedup (>50%). This is not possible until we removed the JSArray variant of `BufferLine`.
It also opens the possibility to evaluate other array buffer based input encodings later on.

Part of #791.


Tyriar:
A very impressive freak out happens when you open vim:

kapture 2018-11-24 at 5 39 48

jerch:
@Tyriar At least it runs fast :blush:

jerch:
@Tyriar Should be fixed now, problem was that I mixed up the input data array with the result concat array.

src/common/TypedArrayUtils.ts
```diff
@@ -39,3 +38,28 @@ export function fillFallback<T extends TypedArray>(array: T, value: number, star
   }
   return array;
 }
+
+/**
+ * Concat two typed arrays `a` and `b`.
+ * Returns a new typed array.
+ */
+export function concat<T extends TypedArray>(a: T, b: T): T {
+  const result = new (a.constructor as any)(a.length + b.length);
```
Tyriar: Oh does `a.constructor` work? I was doing `new (<any>array).__proto__.constructor(` in https://github.com/xtermjs/xterm.js/pull/1790/files

src/common/TypedArrayUtils.ts
```diff
@@ -39,3 +38,28 @@ export function fillFallback<T extends TypedArray>(array: T, value: number, star
...
+/**
+ * Convert UTF16 char codes into JS string.
+ * Note the typed array is not limited to Uint16Array, make sure to align
+ * the values to 0-65535 integer for other typed array types, otherwise
+ * the conversion will fail.
+ */
+export function utf16ToString<T extends TypedArray>(data: T): string {
+  let s = '';
+  for (let i = 0; i < data.length; ++i) {
+    s += String.fromCharCode(data[i]);
```
Tyriar: Even better:

```
String.fromCharCode.apply(null, [60, 61, 62])
<=>
```

jerch: Yes this was overflowing in older JS engines. Since I cannot repro it with FF & Chrome I gonna change it. (Have not tested Edge/IE11/Safari).

jerch:
@Tyriar Yeah we can hold it back, also it will not show its real potential until we switched over to typed array only buffer.

jerch:
I think we should hold this back until 3.11, it does not make much sense with the JS Array based buffer lines.

Tyriar:
@jerch sounds good, try to periodically merge in changes so the branches don't diverge too much

jerch:
TODO before we can merge this:
- [ ] change buffer to 32 bit version layout in preparation for true color support
- [ ] establish UTF8 input chain

jerch:
@Tyriar Did some more cleanup, ready for review.

.gitignore
```diff
@@ -23,3 +23,7 @@ coverage/
 # Keep bundled code out of Git
 dist/
 demo/dist/
+
+# dont pullin other files from .vscode than launch.json
+.vscode/
```
Tyriar: Did you have some extension writing to .vscode/settings.json or something?

src/core/input/TextDecoder.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+  return String.fromCharCode(codePoint);
+}
+
+/**
+ * Convert UTF32 char codes into JS string.
+ * Basically the same as `stringFromCodePoint` but for multiple codepoints
+ * in a loop (which is a lot faster).
+ */
+export function utf32ToString<T extends TypedArray>(data: T, start: number = 0, end: number = data.length): string {
```
Tyriar: Should we be using TypedArray here? Isn't it only really valid for Uint32Array?

src/core/input/TextDecoder.ts
```diff
@@ -0,0 +1,104 @@
...
+  let result = '';
+  let cp;
```
Tyriar: `cp` should have an explicit type or it is an `any`

src/core/input/TextDecoder.ts
```diff
@@ -0,0 +1,104 @@
...
+  for (let i = start; i < end; ++i) {
+    if ((cp = data[i]) > 0xFFFF) {
```
Tyriar: Can we pull the assignment out of the conditional?

src/core/input/TextDecoder.ts
```diff
@@ -0,0 +1,104 @@
...
+      cp -= 0x10000;
+      result += String.fromCharCode((cp >> 10) + 0xD800) + String.fromCharCode((cp % 0x400) + 0xDC00);
```
Tyriar: This looks a bit like magic, maybe add a comment and if it makes sense make consts for the numbers?

.gitignore
jerch: Yeah seems 2 other extensions are placing stuff into this folder, although they are not even used in this project (a python and a rust extension).

src/core/input/TextDecoder.ts
jerch: Yeah might be better to avoid confusion, technically bytes will decode correctly, even Utf16 unless they are surrogates.
Gonna change it.

src/core/input/TextDecoder.ts
jerch: See next comment...

src/core/input/TextDecoder.ts
jerch: Hmm assignments in condition tend to be much faster, need to test the difference, this function is likely to be called very often from different places.

src/core/input/TextDecoder.ts
jerch: Well its the official UTF32 --> UTF16 trancoding rule, maybe I need to comment it abit more...

.gitignore
Tyriar: @jerch I normally report issues against the extensions if they do this, especially when the workspace 
has nothing to do with that language.

src/common/Types.ts
```diff
@@ -35,3 +35,7 @@ export interface ICircularList<T> extends IEventEmitter {
   trimStart(count: number): void;
   shiftElements(start: number, count: number, offset: number): void;
 }
+
+export type TypedArray = Uint8Array | Uint16Array | Uint32Array | Uint8ClampedArray
```
Tyriar: This can move back to TypedArrayUtils.ts now

.gitignore
jerch: Ok gonna remove it...

PR #1816, opened 2018-12-08
Remove one call to charCodeAt in a tight loop while parsing input

juancampa:
I've been profiling really long outputs on hyper and this is in the hot path. Not sure how much this helps (probably not a lot) but it doesn't hurt either.


Tyriar:
Thanks for the PR and investigation @juancampa, @jerch has this one covered though 😃

Issue #1811, opened 2018-12-04
Deprecate and remove JS array buffer version


jerch:
Reminder issue to deprecate and the remove the JS array based buffer implementation. Part of #791.

Currently the JS array based buffer line blocks several cleanups/features to come:
   - Edit: moved into separate tracking issue #1920

@Tyriar Can we make the typed array version the default in 3.9 yet, and deprecate the JS array thing? And maybe remove the JS array version afterwards? This way we have at least one release with both versions and the typed array as default. I feel abit uneasy about a quick removal of the JS array version, but its kinda needed to get some progress with the list above.

/cc @mofux  @bgw

jerch:
Relocated TODO list into tracking issue, closing issue.

PR #1817, opened 2018-12-08
Reduce garbage generation in InputHandler.print

juancampa:
## Problem

Profiler was showing Minor GC triggering because of `getCharAt`. Which means these 1-char are add GC pressure.

## Solution

Reuse 1-char strings by keeping them in a cache.

## Tests

I ran a few sanity checks and everything seems to be working fine. The profiler also doesn't show the GC firing to collect these anymore.

Gonna post profiler screenshots to illustrate the problem after the break :)

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -114,6 +114,7 @@ class DECRQSS implements IDcsHandler {
  */
 export class InputHandler extends Disposable implements IInputHandler {
   private _surrogateFirst: string;
+  private _charCache: { [s: number]: string };
 
   constructor(
       protected _terminal: IInputHandlingTerminal,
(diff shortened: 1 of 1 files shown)


juancampa:
Hmm :thinking: I can't seem to reproduce what I was seeing earlier so I'm gonna go ahead and close this until I can repro it

jerch:
@juancampa This highly depends if you are using the JSArray or TypedArray buffer:
- the JSArray holds the char in the buffer - long living object (expensive to collect for GC)
- the TypedArray has it only as stack variable, it does not outlive the function body (short living and easier to cleanup)

For both it adds GC pressure and for the TypedArray it is even a nonsense conversion and only there for API conformance with the JSArray buffer. I plan to address this once we got rid of the JSArray buffer version (see #1796 and #1811).

juancampa:
You're right, this will go away when your TypedArray PR goes in.

Additionally I'm seeing that at least V8 has an internal cache for one-byte strings: https://github.com/v8/v8/blob/master/src/code-stub-assembler.cc#L6661-L6664, though it's not clear if it's being used (e.g. I can't find the definition of `kSingleCharacterStringCache` in the V8 repo, so maybe it's dead code)

Okay then, I'm not bothering anymore with this PR. Thanks for the clarification @jerch

jerch:
Yeah v8 has the ability to "internalize" often used short strings, this turns a JS string basically into a char pointer under the hood (and makes things much more performant). I think they use some heuristics to determine whether to do the internalization, and I think we typically dont hit it during input flow (as the single char string change lot).

PR #1820, opened 2018-12-09
Add missing break in options code

Tyriar:
Atlas is reset as part of onResize, this should be safe

Extracted from https://github.com/xtermjs/xterm.js/pull/1790


mofux:
LGTM 👍

PR #1821, opened 2018-12-09
Call dispose when atlas is no longer used in cache

Tyriar:
Only runtime change is that this clears the bitmap commit timeout
in DynamicCharAtlas

Extracted from https://github.com/xtermjs/xterm.js/pull/1790


mofux:
LGTM 👍

PR #1826, opened 2018-12-10
 Fix #1660: Search as you type

### LINKED ISSUE #1660: Search addon: Support "find as you type"
> It would be nice to have a browser-style “find-as-you-type” logic. I think it would work if a couple functions were added:
> 
> * `findCurrentOrNext(needle)`
> * `findCurrentOrPrevious(needle)`
> 
> These would expand the current selection if the `needle` is present at the 0-index of the current selection; otherwise they’d do a normal `findNext` or `findPrevious`.

Plan issue: #705

nojvek:
- [x] Add incremental option to `ISearchOptions`
 - [x] DRY examples page
 - [x] Improve performance of translateBufferLineToStringWithWrap with lineCache

Fixes #1660


Tyriar:
@nojvek I'm happy to go forward with this without tests unless you had a good idea of how to test it. I had a look and it's a bit tricky as right now none of the tests really involve TerminalCore/SelectionManager as they are testing findInLine not findNext/Previous. Addon tests should be made easier as part of https://github.com/xtermjs/xterm.js/issues/1128 when we also pull these out of the core repo.

nojvek:
Hi Daniel,

I’m on vacation right now so won’t be able to get the tests out.

I’m happy with your best judgement.

Regards

PR #1828, opened 2018-12-11
Remove utils/ folder

Tyriar:
Part of #1507

Just a bunch of file moves and fixing strict null checks in Clone

utils/Clone* -> common/Clone*
utils/MouseHelper*-> ui/MouseHelper*
utils/TestUtils* -> ui/TestUtils*


mofux:
LGTM. There are still a couple of files in the root `src` directory that need moving to the right spot after this is done.

PR #1843, opened 2018-12-19
Fix translateToString endCol default arg

Tyriar:
@jerch this line breaks when endCol is 0:

```ts
let length = endCol || this.length;
```


Tyriar:
@jerch haha, need to tighten up compiler/lint rules 😃

Issue #1845, opened 2018-12-19
Weblinks within quotes incorrectly include end quote


linrock:
## Details
- OS version: macOS 10.14.2
- xterm.js version: HEAD

Ran into this issue using the latest Hyper terminal version (2.1.0). Original issue here: https://github.com/zeit/hyper/issues/3345

When a URL is in quotes (both single + double quotes), weblinks include the end quote when clicking the link, which opens an invalid link:

![image](https://user-images.githubusercontent.com/208617/50151388-1c3ca600-0304-11e9-8c4c-d540eda53ea9.png)

### Steps to reproduce

1. With webLinks addon enabled, click on a URL enclosed in quotes
2. An invalid URL is opened (should not include end quotes)

PR #1846, opened 2018-12-19
Weblinks should not allow quotes at end of urls enclosed in quotes

linrock:
Fixes https://github.com/xtermjs/xterm.js/issues/1845

Adds a few tests for the expected behavior. Not a comprehensive fix for enclosed URIs though. Could definitely use a better link parser as suggested in https://github.com/xtermjs/xterm.js/issues/583


jerch:
This seems to be a pretty complete grammar of the URL spec: https://stackoverflow.com/a/1856809 (have not read into the rfc). Jut in case someone wants to get the hands dirty...

PR #1849, opened 2018-12-20
Allow holding key on mac to send multiple keys to terminal

### LINKED ISSUE #265: Holding down a key only sends a single key on Mac
Original issue: https://github.com/Microsoft/vscode/issues/10398
Does not reproduce on Ubuntu 16.04
## Details
- Browser and browser version: vscode/electron (Chrome v49)
- OS version: OS X
- xterm.js version: 220828f
### Steps to reproduce
1. `npm start`
2. Focus terminal and hold the <kbd>a</kbd> key, only a single character is added to the prompt

epicfaace:
Fixes #265 - Holding down a key only sends a single key on Mac 
Also see https://github.com/Microsoft/vscode/issues/65436

### DIFF
--- a/src/core/input/Keyboard.test.ts
+++ b/src/core/input/Keyboard.test.ts
@@ -289,6 +289,7 @@ describe('Keyboard', () => {
 
     it('should handle uppercase letters', () => {
       assert.equal(testEvaluateKeyboardEvent({ shiftKey: true, keyCode: 65, key: 'A' }).key, 'A');
+      assert.equal(testEvaluateKeyboardEvent({ shiftKey: true, keyCode: 49, key: '!' }).key, '!');
     });
 
   });
--- a/src/core/input/Keyboard.test.ts
+++ b/src/core/input/Keyboard.test.ts
@@ -281,5 +281,15 @@ describe('Keyboard', () => {
       assert.equal(testEvaluateKeyboardEvent({ keyCode: 0, key: 'UIKeyInputDownArrow' }).key, '\x1b[B');
       assert.equal(testEvaluateKeyboardEvent({ keyCode: 0, key: 'UIKeyInputDownArrow' }, { applicationCursorMode: true }).key, '\x1bOB');
     });
+
+    it('should handle lowercase letters', () => {
+      assert.equal(testEvaluateKeyboardEvent({ keyCode: 65, key: 'a' }).key, 'a');
+      assert.equal(testEvaluateKeyboardEvent({ keyCode: 189, key: '-' }).key, '-');
+    });
+
+    it('should handle uppercase letters', () => {
+      assert.equal(testEvaluateKeyboardEvent({ shiftKey: true, keyCode: 65, key: 'A' }).key, 'A');
+    });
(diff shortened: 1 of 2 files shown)


Tyriar:
Seems to work, but all other editors I try to do this in does this:

screen shot 2018-12-20 at 12 18 44 pm

Isn't that the behavior we're after? Do I need to flip a switch somewhere to disable this popup?

epicfaace:
@Tyriar I'm not sure what you mean. We agree that this code changes the behavior from popup showing -> sending multiple keys, right? Are you seeing the popup when using xterm.js?

epicfaace:
@Tyriar additionally, for me, the popup shows up only when using a text editor, but not when using the Terminal window. I would think that xterm.js should emulate the behavior of Terminal by not showing the popup.

Tyriar:
> the popup shows up only when using a text editor, but not when using the Terminal window

You're right, silly me didn't check actual terminals 😅

Tyriar:
@mofux @jerch looks like this fixes it but it seems to make all the keypress code unnecessary. Do either of you know why keypress would be handled differently? It seems like it's been this way from the very beginning: https://github.com/chjj/term.js/blob/548af6e0a15c5978a2b62a882f7db03ec2ae2850/src/term.js#L2850

mofux:
The difference between `keypress` and `keydown` is that `keydown` will fire for any key that is down - even if it's only the `shift` key. `keypress` will only fire if the pressed key results in a character (the shift key alone won't trigger a keypress event). I'm not sure if the hook to keypress was done intentionally to leverage this behaviour. Maybe we can merge this PR and create a follow-up issue that deals with it?

jerch:
The key event handling dates back to the original impl from 2011. In those days the it was a nightmare to get a somewhat reliable key handling across browsers and platforms. For many keys and combinations you would have to deal with both events due to a lack of specification (http://unixpapa.com/js/key.html).

Not sure which event should be used these days, a summary can be found here https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode.

epicfaace:
I think we should be using https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key instead (and use a polyfill - https://github.com/cvan/keyboardevent-key-polyfill - if we need to support older browsers). Refactoring the code to do that is probably a separate issue unrelated to this one though.

epicfaace:
@Tyriar is this all set? Are there any changes you'd like to see before this PR can be merged?

Tyriar:
I just tried and that popup I mentioned earlier does appear now (I recently upgrades to Mojave):

screen shot 2018-12-27 at 9 37 01 am

Even cancelling the event doesn't seem to stop this unfortunately 🙁 

Also, I tried removing the keypress event on top of this change, but then space doesn't work.

Created https://github.com/xtermjs/xterm.js/issues/1862 for moving to `key` and `code`.

juancampa:
This causes an issue on Linux when the `Meta` key is pressed. I'll post a PR with a fix

PR #1854, opened 2018-12-21
Testbed: 32 bit memory layout for buffer

jerch:
Testbed PR to be able to identify differences between 32 and 16 bit memory layout for the buffer in runtime and memory usage.

Implements:
- 32 bit buffer layout as discussed in #791 
- utf8 input encoding

Else its based on current master, additionally it includes the webgl renderer as default (makes it easier to spot buffer related things in the timeline).
Simply run the demo app and compare the timeline results with the 16 bit variant (yet to come).

Note: To see the impact of the utf8 input handling alone, the timeline results can be compared to the webgl PR branch.

/cc @Tyriar, @mofux, @bgw


jerch:
First results:

grafik

grafik

grafik

grafik

jerch:
Results in xterm-benchmark:

```
            Case "#1" : 10 runs - average runtime: 1545.61 ms
            Case "#1" : 10 runs - average throughput: 31.27 MB/s
```

Tyriar:
I think this is definitely the way to go to start with, double the (small amount of) memory for a little faster runtime and it's way more simple especially since we don't have to deal with how attributes are stored and invalidated.

Tyriar:
@jerch let's move the memory layout discussion over to https://github.com/xtermjs/xterm.js/issues/484, I think we're in agreement for how the buffer should work initially and it's sounding like we can probably get true color going in 3.11 if we go with this lower risk approach?

jerch:
@Tyriar Yes, with this approach here the extended attrs are much easier and less error prone to implement, prolly can be done in one true color PR.

PR #1855, opened 2018-12-21
Testbed: 16 bit memory layout for buffer

jerch:
Counterpart to #1854 but with UTF16 codepoint and attr indirection.

Implements:
- 16 bit buffer layout as discussed in #791
- utf8 input encoding
- direct buffer access in InputHandler.print (therefore the JS array buffer will not work here)

Other than in the test described in https://github.com/xtermjs/xterm.js/issues/791#issuecomment-448216662 I changed the following things here:
- access is done by a single 32 bit value
- attr storage is now a simple JS array, matching is done by `indexOf` instead of a RB tree

The changes show much better runtime behavior than the old test, its still slower than the 32 bit version. Memory usage is the same as in the old test. The attr storage can be further optimized by using another typed array.

More detailed tests results will follow.

/cc @Tyriar, @mofux, @bgw


jerch:
First results:

grafik

grafik

grafik

grafik

jerch:
Results in xterm-benchmark:

```
            Case "#1" : 10 runs - average runtime: 1728.82 ms
            Case "#1" : 10 runs - average throughput: 27.63 MB/s
```

jerch:
The attr storage is still flawed as it will hog memory over time for many ongoing cell changes which will even survive a buffer roundtrip due to the recycling :blush:. No way around the ref counting I guess.

Also with js array the attr storage is already at 200 bytes for just 5 one single number entries (due to preallocing empty slots), while the line content typed array for 87 cols is just 532 bytes per line.

Much room for improvements still.

Tyriar:
> attr storage is now a simple JS array, matching is done by indexOf instead of a RB tree

Are they packed something like 24 bg, 24 fg, 5 remaining attributes? The fact that `number` is actually a float64 would probably also slow things down? The webgl renderer needs fast access to bg and fg as they're used as keys for the glyph cache.

jerch:
@Tyriar Currently the attr storage does not deal yet with RGB values, it just stores and compares the single attr number we currently use. With true color support the storage will have to match against 2 numbers. Imho its not possible to squeeze them into one number (we need around 58 bits, but a number can only store 53 exact bits).
Thus the final attr storage will have to do 2 number comparisons (slightly slower). This would change with Uint64 support, which is in draft (v8 supports this since summer).

PR #1856, opened 2018-12-22
Remove gitter and improve README

### LINKED ISSUE #1838: Should we shutdown the Gitter?
We direct people to [this Gitter](https://gitter.im/sourcelair/xterm.js?at=5c137777b4c74555ccc6efa0) from the question part of the issue template: https://github.com/xtermjs/xterm.js/issues/new/choose

It looks like it rarely sees activity and when it does, often things go unanswered for quite a while. I personally don't track the Gitter as I have enough channels I need to watch already. Thoughts on shutting down the Gitter and directing people to just Stack Overflow instead?

Tyriar:
Major changes:

- Remove links to the Gitter (see #1838), Stack Overflow is now where we direct questions
- Rewrote the features section to combine some points and focus more on things that make xterm.js shine
- Removed duplicate API section
- Removed third party addons section, it looks bad with only one, plus we kind of discourage distributing these until #1128 is handled
- Removed the jsdelivr badge, I don't think it's important and causes clutter, it's also not the recommended way to consume the lib

Fixes #1838

### DIFF
--- a/.github/ISSUE_TEMPLATE/question.md
+++ b/.github/ISSUE_TEMPLATE/question.md
@@ -1,8 +1,8 @@
 ---
 name: Question
-about: The issue tracker is not for questions. Please ask questions on https://stackoverflow.com/questions/tagged/xtermjs or https://gitter.im/sourcelair/xterm.js.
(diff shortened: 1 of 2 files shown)


Tyriar:
Created one to sync the website as well xtermjs/xtermjs.org#63

PR #1859, opened 2018-12-23
Fix text being top aligned in latest Chrome and Firefox

### LINKED ISSUE #1858: Cursor is taller than line.
xterm.js:
image

gnome-terminal:
image

Tyriar:
Chrome and Firefox textBaseline behavior has always differed, see:
- https://bugzilla.mozilla.org/show_bug.cgi?id=737852
- https://bugs.chromium.org/p/chromium/issues/detail?id=607053

Plus Chrome seems to have moved to Firefox's behavior in the latest version, moving to middle should fix the problem for both.

Fixes #1858 

---

After Firefox 64

screen shot 2018-12-23 at 10 52 54 am

After Chrome 71

screen shot 2018-12-23 at 10 50 47 am

Before Firefox 64

screen shot 2018-12-23 at 10 54 17 am

Before Chrome 71

screen shot 2018-12-23 at 10 53 51 am

### DIFF
--- a/src/renderer/BaseRenderLayer.ts
+++ b/src/renderer/BaseRenderLayer.ts
@@ -236,12 +236,12 @@ export abstract class BaseRenderLayer implements IRenderLayer {
    */
   protected fillCharTrueColor(terminal: ITerminal, charData: CharData, x: number, y: number): void {
     this._ctx.font = this._getFont(terminal, false, false);
-    this._ctx.textBaseline = 'top';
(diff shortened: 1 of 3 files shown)


Tyriar:
@jerch I don't think we need to get them to align exactly. They're both using the browser's definition of middle which is fine.

To be clear, the before screenshots are before the PR, on those versions, not < the version.

jerch:
Oops, got it, yeah much better now :+1:

Issue #1801, opened 2018-11-26
search addon doesn't work well with chinese


zjjEcho:
I found some related issues, like this [emoji issue](https://github.com/xtermjs/xterm.js/issues/1103)
But it doesn't seem to solve my problem.

## Details
- Browser and browser version:  chrome(52.0.2743.82  64-bit)
- OS version: Ubuntu 16.04 LTS
- xterm.js version:  3.8.0

### Steps to reproduce

1. generate test content with some chinese in it 
2. search some content，like this
![2](https://user-images.githubusercontent.com/5327847/49000178-be4cf080-f193-11e8-89ad-d8c849e331bc.jpg)

if there's no chinese before the search content, search addon works fine, like this
![1](https://user-images.githubusercontent.com/5327847/49000245-eb010800-f193-11e8-902a-173174f48c15.jpg)

jerch:
This is prolly related to #1686.

Issue #1686, opened 2018-09-13
Selection with search and unicode


jerch:
Combining, surrogate or fullwidth chars in the line and/or the search string lead to weird selection offset problems. Steps to repro:

- insert into demo: `echo -en 'combining: ééé\nfullwidth: ￥￥￥\nsurrogate: 𓂀𓂀𓂀\n'`
- search for 'ééé', '￥￥￥' and '𓂀𓂀𓂀'

The selection is kinda off for all 3 types, it gets even worse if the line contains any of these before their occurence. It seems the renderer and the selection manager do not agree on the chars widths and lengths.

Since I had a similar problem with the linkifier, it might be fixable the same way (#1678).

jerch:
@Tyriar Nope its not gone, still the same here. Maybe its a platform issue?

Looks like this atm:
![grafik](https://user-images.githubusercontent.com/6193135/45951100-82b27080-c002-11e8-803a-f2b09a25d11d.png)
![grafik](https://user-images.githubusercontent.com/6193135/45951013-38c98a80-c002-11e8-9714-66d88fe75b88.png)

Tyriar:
@jerch are you on Linux?

jerch:
Yes Ubuntu 16 here.

Tyriar:
I guess we need to have a setting for this stuff like you were suggesting before. Still not sure the best way of querying the platform for these character widths though, I doubt we can rely on all Linux distros being the same and macOS being a different case.

Tyriar:
Same underlying issue to https://github.com/xtermjs/xterm.js/issues/1059?

jerch:
Nope, this time its not wcwidth's fault, changing the argument I mentioned above fixes the problems (tested a few minutes ago)

jerch:
Currently blocked by #1707 and #1709.

jerch:
Some background on this:
The way the start and end pos of the selection is determined still does not work for all surrogates and fullwidth chars combinations - thus if there are any of those in the line before the match or in the match itself start and end offsets can occur.

This can be fixed the same way I had to fix the linkifier underlining in #1769, by mapping a string index back to the buffer index:
https://github.com/xtermjs/xterm.js/blob/c7fa89da8e97e907cdfb72b23eabf5c3a5d1bb9e/src/Linkifier.ts#L223

If done twice (match start and end) the selection will correctly point to the underlying cells.

Tyriar:
This still happens. This is the problem line:

[...]

PR #1866, opened 2018-12-29
Fix search addons: Fix problem of skipping some results and incorrect text selection after resize window

ntchjb:
When function `_findInLine` is run in reverse mode (find previous), it skipped the results that is wrapped, and also skipped the line that has `isWrapped === true` although there are other results in the line. When resizing the window and search again, it highlights incorrect position.

So, for skipping wrapped result problem, I added a condition to check whether start column index is in the range or not. If not, then `lastIndexOf` function should begin matching at the end of the string.

For skipping wrapped line, I think the part of code inside `_findInLine` that did a return when found properties `isWrapped === true` can be removed because we should also scan other results in wrapped line.

For resizing window, I saw that the value at `this._terminal._core.buffer.lines.get(y).length` didn't change when the window was resized, but for `this._terminal.cols`, the value is changed, so I decided to use `this._terminal.cols` to match with searching range.


src/addons/search/SearchHelper.ts
```diff
@@ -187,9 +187,6 @@ export class SearchHelper implements ISearchHelper {
    * @return The search result if it was found.
    */
   protected _findInLine(term: string, row: number, col: number, searchOptions: ISearchOptions = {}, isReverseSearch: boolean = false): ISearchResult {
-    if (this._terminal._core.buffer.lines.get(row).isWrapped) {
```
Tyriar: Testing ti locally on master and your branch it doesn't seem to work on either? :confused: 

The idea behind this `isWrapped` check is to ignore all rows flagged as wrapped and only take the first row (which always has `isWrapped = false`

ntchjb: Oh sorry, I misunderstand the idea. I will put back the return block and try to fix else where. The problem is that if the wrapped line has multiple matches, the search result will highlight only one result and skipped others then go to next line.

ntchjb:
I fixed the problem by controlling index of `col` before passing to `_findInLine`. When the function scans on unwrapped line, it should have the value of `col` high enough so that the function can scan through all wrapped lines below the unwrapped line.

ntchjb:
This PR might relate to #1686 and #1801

Tyriar:
> For the latest commit, I try to modify searching function to support wide characters and combined characters by mapping between buffer index and string index.

It might be better to revert the latest commit and do that in another one as this was so close. I'm happy to merge if we revert that, provided that "- Fix bug when search selection skipped some result" in the latest commit message is not something we need to worry about?

> I'm not sure why \n in write function results in several empty cells before 123....hello string.

That's happening because you did `\n` which moves the cursor to the following line, but you missed the `\r` which moves the cursor to the start of the line.

ntchjb:
OK, I reverted the latest commit to the previous commit. However, at the latest commit, there is a fix that relates to the previous commit, so I make an additional commit to complete the fix.

src/addons/search/SearchHelper.ts
```diff
@@ -71,11 +94,23 @@ export class SearchHelper implements ISearchHelper {
     // Search from the top to the startRow (search the whole startRow again in
     // case startCol > 0)
     if (!result) {
+      // Assume that The first line is always unwrapped line
+      let findingRow = 0;
+      // Scan at beginning of the line
+      let cumulativeCols = 0;
       for (let y = 0; y <= startRow; y++) {
-        result = this._findInLine(term, y, 0, searchOptions);
+        result = this._findInLine(term, findingRow, cumulativeCols, searchOptions);
```
Tyriar: It looks like this change is grabbing the `cumulativeCols` and calling _findInLine multiple times for a single wrapped row. This kind of goes against the design and _findInLines is meant to get the full unwrapped line and search it fully. I think the actual root cause is somewhere inside `_findInLine` and calling it multiple times on the same row is getting around it.

After this section `searchStringLine` is meant to contain the full wrapped line so we can search it in full:

https://github.com/xtermjs/xterm.js/blob/509ce5fa3a698ee7847419117e9dd6b979b105bf/src/addons/search/SearchHelper.ts#L190-L203

ntchjb: Oh, you were right. `_findInLine` should return the result at the first time, otherwise, go to the next unwrapped line.

ntchjb:
I'm sorry, but I couldn't reproduce the problem. I tested in demo server using `yarn start`. The search results found `df` keywords in both forward and backward which covers all matching words. Here is my snapshot.
Screen Shot 2562-03-15 at 17 49 52

I used column with width cols 87 and rows 26. Here is the text results
```
bash-3.2$ echo '
> lkmasdjasdfknajlsdlnkafdslnkasdflnjalsdfnkjdasfnlkjadsfnljksdajklsadfllfsdakafdsfadsnlkadf
> '

lkmasdjasdfknajlsdlnkafdslnkasdflnjalsdfnkjdasfnlkjadsfnljksdajklsadfllfsdakafdsfadsnlkadf

bash-3.2$ 
```

ntchjb:
If you mean that, after typed a bunch of the string and ran `findPrevious` search without any selection at the first time, and it didn't select matching string at the end of the first line, then, I changed the code in `findPrevious` to start scanning at the last row first if there is no text selection.

PR #1872, opened 2019-01-02
Setup continuous deployment beta builds

Tyriar:
This PR will build and publish a "beta" tagged build to npm whenever new commits arrive in master.

We will need to change our workflow to update the version number immediately after releasing instead of just before a release for this the names to be right.


bin/publish.js
```diff
@@ -0,0 +1,49 @@
...
+
+// Get the version
+const tag = 'beta'
+const nextVersion = getNextVersion(tag);
+console.log(`Publishing version: ${nextVersion}`);
+
+// Set the version in package.json
+const packageJsonFile = path.resolve(__dirname, '..', 'package.json');
+let packageJsonRaw = fs.readFileSync(packageJsonFile).toString();
+packageJsonRaw = packageJsonRaw.replace(/("version": ")3.10.0(")/, `$1${nextVersion}$2`);
```
mofux: I'm not pretending to understand this completely, but the hard-coded `3.10.0` looks wrong to me 🤔

jerch: Can't this be derived directly from some repo tag?

mofux: Another maybe cleaner and less error-prone approach:

```javascript
// requiring a .json file will result in a parsed json object
const packageJson = require(packageJsonFile);

// update the version
packageJson.version = nextVersion;

// write back as json (using 4 spaces for every indent level)
fs.writeFileSync(packageJsonFile, JSON.stringify(packageJson, null, 4));
```

Tyriar: I tried that using `JSON.stringify(packageJson)` earlier, will try with the extra args

mofux:
LGTM so far. Maybe adding a note about the beta channel to the README would be useful?

jerch:
> We will need to change our workflow to update the version number immediately after releasing instead of just before a release for this the names to be right.

Isn't this abit unconvenient/error prone to get done right for difference release cases, esp. patches? Ive never done the release things, still can image that one can easily mix up things here. Would it be easier to calculate the next version number from the existing, e.g. with the help of node-semver?

Tyriar:
> Isn't this abit unconvenient/error prone to get done right for different release cases, esp. patches? Ive never done the release things, still can image that one can easily mix up things here. Would it be easier to calculate the next version number from the existing, e.g. with the help of node-semver?

My suggestion aligned with how VS Code does it which works pretty well, if the "error" case does happen it doesn't really matter that much, it just means that `3.12.0-betaX` was released instead of `3.13.0-beta1`. We could add this to the releasing guide (and maybe automate it away eventually) https://github.com/xtermjs/xterm.js/wiki/Releasing

Tyriar:
I'm not sure if we will want to keep this long term, but it would be a really good way to test out the process for automating stable releases.

Tyriar:
> Isn't this abit unconvenient/error prone to get done right for different release cases, esp. patches? Ive never done the release things, still can image that one can easily mix up things here. Would it be easier to calculate the next version number from the existing, e.g. with the help of node-semver?

After thinking about it more I decided to calculate the next version based on the current version and bumping the minor version. That way we can keep the current system of committing the version number when it's being released.

PR #1874, opened 2019-01-03
stop AltGraph/Meta from being written to terminal

pfitzseb:
Fixes regression introduced in https://github.com/xtermjs/xterm.js/pull/1849 where pressing the `AltGraph` key (to e.g. input a `]` on a german keyboard) results in a literal `AltGraph` being input in the terminal. Same goes for `Meta` -- not sure if either of those are Linux-only problems though.

image


Tyriar:
Thanks for the PR, going to go with this fix which should hopefuly fix all the cases https://github.com/xtermjs/xterm.js/pull/1881

PR #1881, opened 2019-01-04
Don't print keys that print more than a single char

### LINKED ISSUE #1880: Issues with holding key down PR
PR: https://github.com/xtermjs/xterm.js/pull/1849

Problems:
- https://github.com/xtermjs/xterm.js/pull/1874
- https://github.com/Microsoft/vscode/issues/65949
- https://github.com/Microsoft/vscode/issues/65967

Maybe we should do a patch release?

Tyriar:
Fixes #1880

### DIFF
--- a/src/core/input/Keyboard.ts
+++ b/src/core/input/Keyboard.ts
@@ -349,9 +349,8 @@ export function evaluateKeyboardEvent(
         if (ev.keyCode === 65) { // cmd + a
           result.type = KeyboardResultType.SELECT_ALL;
         }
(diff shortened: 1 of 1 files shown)


pfitzseb:
Yeah, this is definitely better than my attempt at a fix.
Would it be possible to get this into a 3.10.1? I'd really like to use 3.10 because of the improved search API, but this issue is a blocker.

Tyriar:
Yeah I think we'll need to do a 3.10.1 for this based on the amount of issues I've seen come into the vscode report, we'll probably release it on the 8th.

PR #1885, opened 2019-01-06
Fix for Windows key printing "Meta" on Linux (and Windows?)

juancampa:
## Problem

On Linux, press the "Windows" key: The text "Meta" is sent to the terminal. Introduced in https://github.com/xtermjs/xterm.js/pull/1849

## Solution

Do not set `result.key` to "Meta" in `evaluateKeyboardEvent`

## Tests

 - [x] Doesn't send "Meta" anymore on Linux
 - [x] Key repeat still works
 - [ ] Sanity check MacOS

**only tested on Hyper**


Tyriar:
Already have a fix out which covers all cases: https://github.com/xtermjs/xterm.js/pull/1881

PR #1887, opened 2019-01-07
account empty cells in stringIndexToBufferIndex

jerch:
`stringIndexToBufferIndex` does not account empty cells correctly, which leads to wrong underline offsets in linkifier if the line contains empty cells:
- before
grafik

- after
grafik


jerch:
Note this is only partial fix for the linkifier for the start offset, the end offset might still be at the wrong position (needs changes from the other linkifier PR).

src/Buffer.ts
```diff
@@ -449,14 +449,17 @@ export class Buffer implements IBuffer {
...
       const line = this.lines.get(lineIndex);
       if (!line) {
         return [-1, -1];
       }
-      for (let i = 0; i < line.length; ++i) {
-        stringIndex -= line.get(i)[CHAR_DATA_CHAR_INDEX].length;
+      const length = (trimRight) ? line.getTrimmedLength() : line.length;
+      for (let i = 0; i < length; ++i) {
+        if (line.get(i)[CHAR_DATA_WIDTH_INDEX]) {
+          stringIndex -= line.get(i)[CHAR_DATA_CHAR_INDEX].length || 1; // WHITESPACE_CELL_CHAR.length
```
Tyriar: Let's add a comment about how we need 0 width nulls cells actually take 1 space

PR #1888, opened 2019-01-07
linedata cache

jerch:
This PR introduces line level cache for the trimmed length and the trimmed content string, which should greatly speedup any consumer of `translateToString`.
The caching is lazy, thus a value gets only saved upon requesting:
- trimmedLength
- trimmed full line string (important to start at col 0, this will also cache trimmed length as it relies on `getTrimmedLength()` internally)

@Tyriar I have no good real world test case for this. Could you test it with the lineFeed consumer in vscode and compare the runtime and memory consumption with and without this PR? Not sure yet whether its the right way to overcome the multiple string consumer problem.

### DIFF
--- a/src/BufferLine.ts
+++ b/src/BufferLine.ts
@@ -25,6 +25,8 @@ const IS_COMBINED_BIT_MASK = 0x80000000;
 export class BufferLine implements IBufferLine {
   protected _data: Uint32Array | null = null;
   protected _combined: {[index: number]: string} = {};
+  protected _cachedTrimmedLength: number = -1;
+  protected _cachedTrimmedString: string = null;
   public length: number;
 
   constructor(cols: number, fillCharData?: CharData, public isWrapped: boolean = false) {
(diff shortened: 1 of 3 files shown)


jerch:
Nope, results are worse - GC explodes from 30ms up to 750ms, total runtime goes up from 1200ms to ~2000ms for `translateToString` in a scroll handler. Interesting results, the only difference is the lifetime of the strings.
Not yet helpful...

Tyriar:
Idea we chatted about in Slack; having an optional cache parameter which does the caching.

This would let the the search and other addons to leverage caching from the core lib when caching is called for (if we are likely going to be wanting the line multiple times).

jerch:
@Tyriar I would hold this PR back until we are done with the basic buffer transitions to avoid to many interdependencies (reflow PR, typedarray PR, UTF32 PR). The addition of a cache flag is quite easy but would have to be applied all over the codebase.

Issue #469, opened 2017-01-11
Support emoji IMEs


Tyriar:
Currently inputting emojis doesn't work on all platforms (via mac: cmd+ctrl+space, windows: win+.).

Context: https://github.com/zeit/hyper/issues/1275#issuecomment-271679999
VS Code: https://github.com/Microsoft/vscode/issues/35266

Related: https://github.com/sourcelair/xterm.js/issues/467

parisk:
👍  I'll give this a try, because I love adding emojis in my squash messages.

rauchg:
@parisk the technique we use in Hyper (might be useful for you) is to place the contenteditable inside the actual caret, so that its position remains relevant.

Tyriar:
Yes we do something similar with our textarea. @parisk let me know if you need help understanding the `CompositionHelper`, there are two issues with it AFAIK: https://github.com/sourcelair/xterm.js/issues/273 https://github.com/Microsoft/vscode/issues/11334

rauchg:
> One way this issue could be fixed is by no longer hiding the textarea offscreen and instead keeping it in sync with the cursor

In our case I believe we always have it _inside_ the caret div, so no reposition is needed. @matheuss can confirm

Tyriar:
Our caret is an actual character on screen which may also contain text and it's reconstructed whenever the prompt line is refreshed.

matheuss:
Yeah @rauchg, we always have our own `contenteditable` inside the `hterm`'s carret (which is a floating `div` that happens to be _above_ the terminal).

parisk:
I think that we the best option for us is moving the hidden textarea to the caret's position on every stroke (similar to CodeMirror's mechanism for the hidden textarea).

What we have to investigate though is why the character is not being inserted in the terminal when click. I guess it's a focus issue, but I will get on top of it within the weekend to tell for sure.

Tyriar:
The reason we're not getting the emoji is because we don't listen to the `input` event. So I guess we need to listen to that to get the changes. We may want to consider reevaluating `keypress`/`keydown` if we're also listening to `input`.

parisk:
You mean ditching keypress/keydown completely for `input` or using it supplementary to the typical keyboard events?

Tyriar:
Not sure yet, maybe we don't need keydown anymore? I've also been told that the input eventis needed to respect repeating chars when holding down a key on macOS.

epicfaace:
[...]

PR #1890, opened 2019-01-07
Emoji fix

juancampa:
## Problems
There were multiple problems with emoji

 1. Opening the emoji dialog was possible but it appeared at a weird location. This was due to the textarea we use for input was not being kept in sync with the cursor position unless a composition (i.e. IME input) was underway.
 1. No emoji was received by the terminal.

## Solutions

 1. Now the textarea always moves with the cursor.
 1. Added handler for the `input` event, only for emoji handling. Also, the textarea needs to have a `lineheight > 0` for this event to fire, at least in Chromium.

## Testing

 - [ ] MacOS
   - [x] Emojis are inserted from the emoji input popup
   - [x] The popup shows at the right location
   - [x] Pasting still works fine. Including pasting emoji
   - [x] Chrome
   - [ ] Firefox
   - [x] Safari
   - [x] IMEs still work (jp/ko/ch) including inserting emoji from Pinyin IME
 - [ ] Windows
   - [ ] Emojis are inserted from the emoji input popup
   - [ ] The popup shows at the right location
   - [ ] Pasting still works fine. Including pasting emoji
   - [ ] Chrome
   - [ ] Firefox
   - [ ] Safari
   - [x] ~~IE (support dropped)~~
   - [ ] IMEs still work (jp/ko/ch) including inserting emoji from Pinyin IME
 - [x] Electron (via Hyper)

### DIFF
--- a/src/CompositionHelper.ts
+++ b/src/CompositionHelper.ts
@@ -4,6 +4,7 @@
  */
 
 import { ITerminal } from './Types';
+import { isEmoji } from './Emoji';
 
 interface IPosition {
   start: number;
@@ -125,7 +126,7 @@ export class CompositionHelper {
       // Cancel any delayed composition send requests and send the input immediately.
       this._isSendingComposition = false;
       const input = this._textarea.value.substring(this._compositionPosition.start, this._compositionPosition.end);
-      this._terminal.handler(input);
+      this._handler(input);
     } else {
       // Make a deep copy of the composition position here as a new compositionstart event may
       // fire before the setTimeout executes.
@@ -157,9 +158,15 @@ export class CompositionHelper {
             // (eg. 2) after a composition character.
             input = this._textarea.value.substring(currentCompositionPosition.start);
           }
-          this._terminal.handler(input);
+          this._handler(input);
         }
       }, 0);
+    }
+  }
+
+  private _handler(str: string): void {
+    if (!isEmoji(str)) {
+      this._terminal.handler(str);
     }
   }
 
@@ -177,7 +184,7 @@ export class CompositionHelper {
         const newValue = this._textarea.value;
         const diff = newValue.replace(oldValue, '');
         if (diff.length > 0) {
-          this._terminal.handler(diff);
+          this._handler(diff);
         }
       }
     }, 0);
@@ -190,10 +197,6 @@ export class CompositionHelper {
    *   necessary as the IME events across browsers are not consistently triggered.
    */
   public updateCompositionElements(dontRecurse?: boolean): void {
-    if (!this._isComposing) {
-      return;
-    }
-
     if (this._terminal.buffer.isCursorInViewport) {
       const cellHeight = Math.ceil(this._terminal.charMeasure.height * this._terminal.options.lineHeight);
       const cursorTop = this._terminal.buffer.y * cellHeight;
@@ -206,11 +209,13 @@ export class CompositionHelper {
       // Sync the textarea to the exact position of the composition view so the IME knows where the
       // text is.
       const compositionViewBounds = this._compositionView.getBoundingClientRect();
+      const width = Math.max(compositionViewBounds.width, this._terminal.charMeasure.width);
+      const height = Math.max(compositionViewBounds.height, this._terminal.charMeasure.height);
       this._textarea.style.left = cursorLeft + 'px';
       this._textarea.style.top = cursorTop + 'px';
-      this._textarea.style.width = compositionViewBounds.width + 'px';
-      this._textarea.style.height = compositionViewBounds.height + 'px';
-      this._textarea.style.lineHeight = compositionViewBounds.height + 'px';
+      this._textarea.style.width = width + 'px';
+      this._textarea.style.height = height + 'px';
+      this._textarea.style.lineHeight = height + 'px';
     }
 
     if (!dontRecurse) {
--- a/src/CompositionHelper.ts
+++ b/src/CompositionHelper.ts
@@ -4,7 +4,6 @@
  */
 
 import { ITerminal } from './Types';
-import { isEmoji } from './Emoji';
 
 interface IPosition {
   start: number;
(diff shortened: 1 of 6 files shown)


Tyriar:
The position is still broken on Windows:

win+.

image

Actual input doesn't seem to work either, but that's probably related to shell/pty problems.

image

yarn.lock
```diff
@@ -26,6 +26,11 @@
   resolved "https://registry.yarnpkg.com/@types/chai/-/chai-3.5.2.tgz#c11cd2817d3a401b7ba0f5a420f35c56139b1c1e"
   integrity sha1-wRzSgX06QBt7oPWkIPNcVhObHB4=
 
+"@types/dom-inputevent@^1.0.3":
```
Tyriar: This should be undone since we're not using anymore

src/common/Types.ts
```diff
@@ -21,6 +21,15 @@ export interface IKeyboardEvent {
   type: string;
 }
 
+/**
+ * There's a @types/dom-inputevent but at the moment it's outdated and missing
+ * some properties (e.g. inputType). So in the meantime we're using this instead
+ */
+export interface IInputEvent {
+  readonly data?: string;
+  readonly inputType?: string;
```
Tyriar: It's missing from the type probably because support is bad (no Firefox), do you know why it appears to work on Firefox despite this? https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/inputType#Browser_compatibility

src/Emoji.ts
```diff
@@ -0,0 +1,32 @@
+// This function returns whether the first code point of the provided string is
+// within a range that contains emoji. Some of the codepoints in these ranges
+// are unassigned so this test is an approximation to the real thing.
+// Ranges taken from:
+//   https://stackoverflow.com/questions/30757193/
+// For a complete list of all assigned codepoints, go to:
+//   https://unicode.org/Public/emoji/11.0/emoji-data.txt
+export function isEmoji(str: string): boolean {
```
Tyriar: Let's move this into CharWidth.ts, this will later be renamed to Unicode*.ts so i think it makes sense to live there.

juancampa:
> The position is still broken on Windows:

Unfortunately I don't have a good way to test on Windows at the moment. Trying to get access to a remote desktop machine.

> macOS/firefox

Looking into it, it seems to work sometimes though :thinking:

src/CompositionHelper.ts
```diff
@@ -190,10 +190,6 @@ export class CompositionHelper {
    *   necessary as the IME events across browsers are not consistently triggered.
    */
   public updateCompositionElements(dontRecurse?: boolean): void {
-    if (!this._isComposing) {
-      return;
-    }
-
     if (this._terminal.buffer.isCursorInViewport) {
```
Tyriar: Why did you need to remove this? I don't think it's meant to move when composing (after it becomes visible)

src/Emoji.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (str.codePointAt === undefined) {
```
Tyriar: Please add:

```
// TODO: Remove when IE11 support is dropped
```

src/common/Types.ts
```diff
@@ -21,6 +21,15 @@ export interface IKeyboardEvent {
   type: string;
 }
 
+/**
+ * There's a @types/dom-inputevent but at the moment it's outdated and missing
+ * some properties (e.g. inputType). So in the meantime we're using this instead
+ */
+export interface IInputEvent {
```
Tyriar: Let's call this `IDomInputEvent` to be clear as `IKeyboardEvent` above is an xterm.js thing

src/common/Types.ts
epicfaace: @Tyriar The `data` attribute is not supported on Firefox. As per my previous comment, inputting emojis somewhat works with Firefox even _without_ the code on this branch: https://github.com/xtermjs/xterm.js/issues/469#issuecomment-451789158

juancampa: Interesting. I'll run it with the debugger to understand what's happening on Firefox

src/CompositionHelper.ts
juancampa: Not sure I follow. If that were the case, shouldn't the `if` be `if (this._isComposing)`. My understanding is that this `if` avoids unnecessarily moving the `textarea` when is not needed (not composing anything), but I might be wrong :stuck_out_tongue_closed_eyes: 

The idea here is that we need a `textarea` to follow the cursor at all times so that the emoji popup shows at the right position (right under the text area).

juancampa: Thinking about this a bit more, it might be better to move the positioning of the `textarea` to `Terminal.ts` since it's not a composition thing anymore

src/common/Types.ts
juancampa: I did a bit of investigation on Firefox:

 - An alternative to `data` is to use `event.target.value` but only if we're clearing the `textarea` on every change
 - Firefox positioning of the Emoji popup is buggy in that, if you move the `textarea` around, the emoji popup's position (later, when it shows) will use the previous position of the textarea. A work-around I found is to manually set `textarea.value` to something (doesn't matter what), which seems to force firefox to recalculate the position used for emoji popup. Seems like a cache invalidation bug.
 - Firefox treats emoji input as composition while chrome doesn't :man_shrugging:

juancampa: - Oh, and finally, and this might be a deal breaker, there's no `inputType` on Firefox, which we're using to detect when text is inserted vs. when text is pasted. There might be a way to work around this, by treating it as a composition but I haven't looked deeper into it

epicfaace: If we can't easily do it reliably on Firefox, I think at least doing it on Chrome would definitely be an improvement.

Tyriar:
@jerch do you know why if you press space twice after inserting an emoji a `.` appears?

screen shot 2019-01-26 at 12 51 59 pm

This even happens on master when you paste an emoji

src/CompositionHelper.ts
Tyriar: This function is only meant to do anything after `compositionstart` and before `compositionend`, I think you're right that maybe this should be moved out.

Tyriar:
Ping @juancampa, you'll have difficulty keeping this page up to date in Hyper if it doesn't get merged in

Tyriar:
Closing off since this needed a bunch of changes and has diverged a lot. A new attempt at fixing should probably start on a fresh branch.

PR #1895, opened 2019-01-13
True color support

### LINKED ISSUE #484: Terminal does not render true color

jerch:
PR to finally get true color support. Fixes #484.

Steps to get this done:
- [x] rewrite terminal attributes to support RGB
- [x] use new buffer layout
- [x] enhance DOM renderer to deal with RGB
- [x] enhance canvas renderer to deal with RGB

With this PR xterm.js will support true color in the semicolon notation like CSI 38 ; 2 ; R ; G ; B m. 
To be more in line with the spec the parser has to be extended to support the colon notation. It will not be covered with this PR.

This PR relies on #1878, most of the commit are from there. The true color changes start with commit c444eebd89b448afd64be59f7206cc6febeaeb1e.


jerch:
DOM renderer has now RGB support.

- without RGB
grafik

- with RGB
grafik

jerch:
@Tyriar I am not sure about the canvas renderer whether it should be applied there at all, since it is a drop candidate. Any thoughts on this?

Tyriar:
@jerch 😍

I think we should, the canvas renderer will likely take months to phase out (iff there are no issues with webgl).

jerch:
canvas renderer has now preliminary RGB support. This only hacked in for now and bypasses the cache, several conditions like inverting fg/bg does not work yet.

jerch:
Transition of the canvas renderer is done. RGB is for now always uncached, not sure if we ever want to cache those, the sheer amount of colors will invalidate the cache entries often. Might still be worth to try since most use cases will still only use a small set of colors.
A minor optimization might be to allow cached drawing of palette foreground on RGB background, but we should test first if the possible glyph border artefacts are bearable.

Technically this PR is ready for review, but I think we should get the pending PRs it relies on done. In the meantime I gonna cleanup the new code and do more tests. The latter is cumbersome since we have no automated canvas tests at all. Maybe we get something rolling with node-canvas?

Tyriar:
@jerch some conflicts are stacking up on this one.

jerch:
@Tyriar This was expected, its due to #1878 and the reflow changes, both conflict in buffer changes.

Tyriar:
Fixed a lot of merge conflicts

src/Types.ts
```diff
@@ -521,11 +520,43 @@ export interface IEscapeSequenceParser extends IDisposable {
...
+  isInverse(): number;
+  isBold(): number;
+  isUnderline(): number;
+  isBlink(): number;
+  isInvisible(): number;
+  isItalic(): number;
+  isDim(): number;
+
+  // color modes
+  getFgColormode(): number;
```
Tyriar: Capital M?

PR #1897, opened 2019-01-15
Prevent charsizechanged from firing unnecessarily

juancampa:
## Problem

`charsizechanged` was being fired regardless of whether the actual char size changed or not. 

## Solution
 
Only fire the event when the "adjusted size" (rounded up) changes.

I ran into this bug because, when using the WebGL renderer, canvas resizes are very expensive on Chrome. This was causing the renderer to resize the canvas every time `measure` was called.

## Tests

 - [x] Terminal resizes up/down when changing the number of cols/rows
 - [x] Terminal resizes up/down when changing the font size

### DIFF
--- a/src/ui/CharMeasure.ts
+++ b/src/ui/CharMeasure.ts
@@ -46,9 +46,10 @@ export class CharMeasure extends EventEmitter implements ICharMeasure {
     if (geometry.width === 0 || geometry.height === 0) {
       return;
     }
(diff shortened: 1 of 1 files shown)


juancampa:
@Tyriar let me know if there's other functionality you'd like me to test. So far it seems like nothing breaks

Tyriar:
@juancampa you mentioned changing this caused some other bug, was that something in your local workspace?

PR #1899, opened 2019-01-17
Support RTL languages in the canvas renderer

Tyriar:
Part of #701 

Before:

screen shot 2019-01-17 at 9 15 13 am

After:

screen shot 2019-01-17 at 9 15 21 am

This does slow things down (~50% slower) but I think @jerch is working on improving the character joiner code to be faster, if the registerCharacterJoiner API works with codes instead of a big string it shouldn't make much difference.

DOM/WebGL renderer support will work when they support the character joiner API.

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -52,6 +52,7 @@ import { IKeyboardEvent } from './common/Types';
 import { evaluateKeyboardEvent } from './core/input/Keyboard';
 import { KeyboardResultType, ICharset } from './core/Types';
 import { clone } from './common/Clone';
+import { registerRtlCharacterJoiners } from './Unicode';
 
 // Let it work inside Node.js for automated testing purposes.
 const document = (typeof window !== 'undefined') ? window.document : null;
@@ -752,7 +753,6 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
     // Listen for mouse events and translate
     // them into terminal mouse protocols.
     this.bindMouse();
-
   }
 
   private _setupRenderer(): void {
@@ -762,6 +762,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
       default: throw new Error(`Unrecognized rendererType "${this.options.rendererType}"`);
     }
     this.register(this.renderer);
+    registerRtlCharacterJoiners(this);
   }
 
   /**
--- a/src/Unicode.test.ts
+++ b/src/Unicode.test.ts
@@ -1,0 +1,19 @@
+/**
+ * Copyright (c) 2017 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { assert } from 'chai';
+import { createUnicodeRangeJoiner } from './Unicode';
+
+describe('createUnicodeRangeJoiner', () => {
+  it('should join charcaters within the range', () => {
+    // Join a-z
+    const joiner = createUnicodeRangeJoiner(97, 122);
+    assert.deepEqual(joiner('ABCabc aBc abC Abc'), [
+      [3, 6],   // abc
+      [11, 13], // ab
+      [16, 18]  // bc
+    ]);
+  });
+});
--- a/src/Unicode.ts
+++ b/src/Unicode.ts
@@ -47,5 +47,5 @@ function createUnicodeRangeJoiner(rangeStart: number, rangeEnd: number): (text: 
       result.push([start, start + length + 1]);
     }
     return result;
-  }
+  };
 }
--- a/src/Unicode.ts
+++ b/src/Unicode.ts
@@ -25,7 +25,7 @@ export function registerRtlCharacterJoiners(terminal: Terminal): void {
   terminal.registerCharacterJoiner(createUnicodeRangeJoiner(0x1EE00, 0x1EEFF));
 }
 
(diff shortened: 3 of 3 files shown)


babakks:
It's good in single words (e.g., "اینجا" instead of "ا‌ج‌ن‌ی‌ا"), but still puts the words in reverse order. The correct sentence is: "اینجا را بخوانید" (not "بخوانید را اینجا").

(I'm a Persian native.)

Tyriar:
@babakks I included space as well, I think it looks right now.

Tyriar:
I guess it will all fall apart when a line wraps though

babakks:
> @babakks I included space as well, I think it looks right now.

I meant the order of the words:
image

Tyriar:
@babakks did you check the latest?

screen shot 2019-01-17 at 2 01 09 pm

babakks:
> @babakks did you check the latest?
> 
> screen shot 2019-01-17 at 2 01 09 pm

That's perfect. Good job!

jerch:
@Tyriar 
Wow nice job! Sadly I cannot test it myself. I think we should get native peeps involved that want to help testing it thoroughly, before going further down the rabbit hole. Bidi is complicated and full of edgecases, that might need to be addressed in a certain way to work with terminals. @babakks interested?

Tyriar:
I think this gets pretty close to macOS' default terminal where I'm pretty sure it all falls apart on wrapped lines

Terminal.app (اینجا را بخوانید wrapped):

screen shot 2019-01-21 at 10 48 07 am

Maybe this will be fine after optimizations to have very basic support (on non-wrapped lines).

Tyriar:
Closing this off, would probably be just as bad having it work sometimes and really mess things up other times. If anything it had some lessons learned 😃

HKalbasi:
@Tyriar this looks like a definite improvement, what it exactly messed up? If it doesn't make problem for ltr, it would be better than current situation.

Tyriar:
@HKalbasi words ended up in reverse order and there were likely other problems with wrapped words. I found I didn't understand the problem fully so I wasn't comfortable making a half fix.

HKalbasi:
I have an approximate knowledge of the bidi unicode algorithm, since I'm a native RTL speaker and this problem is so common. Let me try to explain it:

There are four kind of unicode characters in bidi:
* Strong LTR: alphabet of LTR lanugages, e.g. a, b, x, ....
* Strong RTL: alphabet of RTL languages, like ا ب ج ...
* Weak LTR: numbers like ۱ 1 ۲ 2 ...
* Neutral: white spaces, dot, comma, math signs and ...

The [original categories](https://www.unicode.org/reports/tr44/#Bidi_Class_Values) are a bit more verbose.

For rendering a text:
* Divide the sentence into blocks that consist of consecutive characters with the same type.
* Merge two blocks that there is only neutral and weak LTR blocks between them.
* Put blocks in the direction of first strong character of the text. Most terminals omit this step and always put blocks in LTR.
* Render each block per its type. Render weak LTR blocks inside RTL blocks as LTR.

An example:

```
some ltr SOME RTL 123 MORE, RTL some ltr
LLLLNLLLNRRRRNRRRNWWWNRRRRNNRRRNLLLLNLLL
LLLLLLLLNRRRRRRRRRRRRRRRRRRRRRRNLLLLLLLL
some ltr LTR ,EROM 123 LTR EMOS some ltr
```

From what I'm seeing here, it will become correct per the unicode spec if we join all RTLs and neutrals and weak LTRs.

About line wrapping, I don't see the problem, and how RTL differs from LTR in it. The [screen shot above](https://github.com/xtermjs/xterm.js/pull/1899#issuecomment-456168402) looks good to me.

Tyriar:
@HKalbasi thanks, the [spec for the algorithm](https://www.unicode.org/reports/tr9/) is quite long though, so I imagine reading through that and working through the subtleties will take quite a while. There is also this terminal-specific one which recommends a different approach and is also very long https://terminal-wg.pages.freedesktop.org/bidi/. After reading all that and understanding the problem fully I'm guessing it's probably possible via character joiners but I'm not certain which is what concerns me. I feel like it'll take a pretty significant amount of effort to do this right, the ambiguity also bugs me as I cannot truly test this as I don't know any of these languages.

Tyriar:
Joining the whole line would cause some pretty bad performance issues. All joined characters get their own texture, meaning essentially all repeat characters/words would be cache misses and need to get re-rendered, that would thrash the backing texture and cause it to get cleared frequently.

It seems like some combination of joined characters (for rtl words) and then the ability to reorder the words while keeping the word textures separate would be needed for it to function properly, right now the character joiner system isn't capable of this.

pourmand1376:
Hey @Tyriar,

Thanks for taking time to implement this great feature. 

As a native Persian speaker, I would say that current implementation is useless for any RTL language. 

In cases of need, we would switch to other terminals, or not use terminal for the task. So, I don't think a little performance drop would be significant as we are already not using it when we have to deal with RTL languages. 

From the images, I would be a lot happier in both cases that you have implemented (even reversed word order is readable enough). 

I would appreciate it if you could merge this PR. 🙏

Tyriar:
@pourmand1376 this PR didn't do it properly. It's an awful lot more difficult to do this than you might expect since terminals are built on ancient tech which never had this in mind. I won't have any time in the foreseeable future to work on this unfortunately.

PR #1904, opened 2019-01-20
Utf8 input

jerch:
This PR adds UTF8 input to the terminal.

API is as simple as this:
```TS
Terminal.writeUtf8(data: Uint8Array);
```
The typed array can be picked directly from the websocket (set binaryType to 'arraybuffer') or from `node-pty` for local apps. "Binary strings" are not supported as kinda all tooling in JS world expects buffers for raw UTF8 data.

To actually see this in action with the demo several changes are needed:
- server.js
  - set pty encoding in server.js to `null`
  - change input buffer to work on `Buffer` instead of strings (see here: https://github.com/jerch/xterm.js/blob/c08c298061bd9ef3cdedf1e893535cf4b4d0555a/demo/server.js#L62-L76)
- attach addon:
  - set `binaryType` of the websocket to 'arraybuffer'
  - invoke `.writeUtf8` instead of `.write`

PR relies on #1878, changes start at commit 6bdd380. 

Edit:
~~Note that the UTF8 input currently does no flow control, thus can easily been flooded and "stall" the terminal.~~ Fixed.


jerch:
The last commit (e6e5ecc) applies the changes to the demo. This will most likely not be part of this PR.

Here are some numbers:

- without UTF8
```
   Context "lib/xterm_perfcases/input.js"
      Context "Terminal: ls -lR /usr/lib"
         Case "#1" : 1 - runtime: 2063.41 ms
         Case "#1" : 1 - throughput: 23.36 MB/s
         Case "#1" : 2 - runtime: 1884.72 ms
         Case "#1" : 2 - throughput: 25.57 MB/s
         Case "#1" : 3 - runtime: 1853.08 ms
         Case "#1" : 3 - throughput: 26.01 MB/s
         Case "#1" : 3 runs - average runtime: 1933.73 ms
         Case "#1" : 3 runs - average throughput: 24.98 MB/s
```
- with UTF8
```
   Context "lib/xterm_perfcases/input_utf8.js"
      Context "Terminal: ls -lR /usr/lib"
         Context "input throughtput"
            Case "#1" : 1 - runtime: 1443.94 ms
            Case "#1" : 1 - throughput: 33.41 MB/s
            Case "#1" : 2 - runtime: 1301.88 ms
            Case "#1" : 2 - throughput: 37.05 MB/s
            Case "#1" : 3 - runtime: 1279.51 ms
            Case "#1" : 3 - throughput: 37.70 MB/s
            Case "#1" : 3 runs - average runtime: 1341.78 ms
            Case "#1" : 3 runs - average throughput: 36.06 MB/s
```

Tyriar:
Remaining for this:

- Merge conflict (probably just messed up imports)
- Adding time-based limiting to writeUtf8

jerch:
@Tyriar 
Now with input limit logic (taken from `Terminal.write`). Not yet happy with this, it doubles the buffering logic and makes it impossible to mix string and utf8 byte inputs during heavy input flow (while the buffers are not empty). Any idea how to solve this? A mixed type buffer would solve it but looks quite wrong to me.

src/core/input/TextDecoder.ts
```diff
@@ -73,37 +112,231 @@ export class StringToUtf32 {
...
+   */
+  decode(input: Uint8Array, target: Uint32Array): number {
+    const length = input.length;
+
+    if (!length) {
+      return 0;
+    }
+
+    let size = 0;
+    let byte1;
```
Tyriar: These are typed as `any` atm, can we set this to 0 here to make them a `number`?

src/core/input/TextDecoder.ts
```diff
@@ -73,37 +112,231 @@ export class StringToUtf32 {
...
+    let codepoint = 0;
+    let startPos = 0;
+
+    // handle leftover bytes
+    if (this.interim[0]) {
+      let discardInterim = false;
+      let cp = this.interim[0];
+      cp &= ((((cp & 0xE0) === 0xC0)) ? 0x1F : (((cp & 0xF0) === 0xE0)) ? 0x0F : 0x07);
+      let pos = 0;
+      let tmp;
```
Tyriar: Needs a type

src/core/input/TextDecoder.ts
jerch: Woops - fixed.

jerch:
@Tyriar Any clue why [this test](https://dev.azure.com/xtermjs/xterm.js/_build/results?buildId=1482&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&taskId=5b4cc83a-7bb0-5664-5bb1-588f7e4dc05b&lineStart=979&lineEnd=979&colStart=1&colEnd=53) times out? Happened yday too, but worked in the second run.

jerch:
Reminder to myself: Remove e6e5ecc before merging.

Tyriar:
> Any clue why this test times out? Happened yday too, but worked in the second run.

I saw that too, but rebuilding fixed it 🤔

PR #1905, opened 2019-01-21
fix mouse event listener change before term attached

pfitzseb:
Whenever a terminal would receive a `\x1b[?1003h` or `\x1b[?1003l` escape code (e.g. when opening a tmux session) before being attached with `terminal.open(element)` the following internal error would be thrown:
```
Uncaught TypeError: Cannot read property 'classList' of undefined
    at InputHandler.resetMode (InputHandler.ts:1482)
    at Array.<anonymous> (InputHandler.ts:161)
    at EscapeSequenceParser.parse (EscapeSequenceParser.ts:507)
    at InputHandler.parse (InputHandler.ts:299)
    at Terminal._innerWrite (Terminal.ts:1382)
    at Terminal.ts:1351
```

This PR fixes that issue and should still apply the correct classes at all times.

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -1291,7 +1291,9 @@ export class InputHandler extends Disposable implements IInputHandler {
           this._terminal.vt200Mouse = params[0] === 1000;
           this._terminal.normalMouse = params[0] > 1000;
           this._terminal.mouseEvents = true;
-          this._terminal.element.classList.add('enable-mouse-events');
+          if (this._terminal.element) {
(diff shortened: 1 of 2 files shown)


pfitzseb:
[I did.](https://github.com/xtermjs/xterm.js/pull/1905/files#diff-90c5b49ba0606ef9652e22e4529cc885R737) The second branch there should technically be pointless but can't hurt to have I think.

Tyriar:
@pfitzseb doh how did I miss that 😅

PR #1907, opened 2019-01-22
reference isn't commented out properly

MRHarrison:
Error is being thrown when installing xterm:

ERROR in node_modules/xterm/typings/xterm.d.ts(10,1): error TS1084: Invalid 'reference' directive syntax.

### DIFF
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -7,7 +7,7 @@
  * to be stable and consumed by external programs.
  */
 
-/// <reference lib="dom"/>
+// <reference lib="dom"/>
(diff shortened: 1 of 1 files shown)


Tyriar:
Maybe your typescript version is out of date? The CI failure is showing that adding the slash will stop the DOM reference from working https://dev.azure.com/xtermjs/xterm.js/_build/results?buildId=1110

cp1001:
/// <reference lib="..." /> is a  new reference directives from TypeScript 3.0.
just update typescript above version 3 will fix it.

PR #1916, opened 2019-01-30
Clear selection state if selection doesn't exists

skyway777:
Firstly, I've checked the issue of VSCode, https://github.com/Microsoft/vscode/issues/67438
While debugging, I've realised that is issue of xterm.js dependency.
Also, it is reproduced in demo mode (Chrome browser, mac os)
Problem is that previous selection state is saving, but not always clearing.
And while window is refreshing, new selection is comparing with previous state that not always correct.

### DIFF
--- a/src/renderer/SelectionRenderLayer.ts
+++ b/src/renderer/SelectionRenderLayer.ts
@@ -55,6 +55,7 @@ export class SelectionRenderLayer extends BaseRenderLayer {
 
     // Selection does not exist
     if (!start || !end) {
+      this._clearState();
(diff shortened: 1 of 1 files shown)


Tyriar:
@skyway777 I was thinking the issue here is that backspace wasn't triggering an incremental find? Was this not the case?

skyway777:
Yes, this case.
It is because after enter new chars (with not found result) and backspaces cached selection is exact like was previous one when did found.
It is why highlight logic doesn't trigger.

PR #1917, opened 2019-01-31
Fix various problems with reflow

### LINKED ISSUE #1910: Terminal can crash when reflowing with a lot of wide characters in the prompt/cursor line
It appears to be related to Buffer.y becoming < 0 and also Buffer.ybase becoming too large.

```
TypeError: Cannot read property 'set' of undefined
    at InputHandler.print (InputHandler.ts:434)
    at EscapeSequenceParser._printHandler (InputHandler.ts:128)
    at EscapeSequenceParser.parse (EscapeSequenceParser.ts:442)
    at InputHandler.parse (InputHandler.ts:299)
    at Terminal._innerWrite (Terminal.ts:1372)
    at Terminal.ts:1341
```

```
ERR Cannot read property '_data' of undefined: TypeError: Cannot read property '_data' of undefined
    at BufferLine.copyCellsFrom (/Users/daimms/dev/Microsoft/vscode/node_modules/vscode-xterm/lib/BufferLine.js:165:27)
    at Buffer._reflowSmaller (/Users/daimms/dev/Microsoft/vscode/node_modules/vscode-xterm/lib/Buffer.js:244:45)
    at Buffer._reflow (/Users/daimms/dev/Microsoft/vscode/node_modules/vscode-xterm/lib/Buffer.js:170:18)
    at Buffer.resize (/Users/daimms/dev/Microsoft/vscode/node_modules/vscode-xterm/lib/Buffer.js:152:18)
    at BufferSet.resize (/Users/daimms/dev/Microsoft/vscode/node_modules/vscode-xterm/lib/BufferSet.js:78:22)
    at Terminal.resize (/Users/daimms/dev/Microsoft/vscode/node_modules/vscode-xterm/lib/Terminal.js:1096:22)
    at Terminal.resize (/Users/daimms/dev/Microsoft/vscode/node_modules/vscode-xterm/lib/public/Terminal.js:53:20)
```

Tyriar:
- No longer reflow lines where the full unwrapped line contains the cursor
- Add guards to prevent y and ybase becoming invalid values

Part of Microsoft/vscode#67364
Fixes #1910

### DIFF
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -3,13 +3,13 @@
  * @license MIT
  */
 
-import { CircularList, IInsertEvent, IDeleteEvent } from './common/CircularList';
-import { CharData, ITerminal, IBuffer, IBufferLine, BufferIndex, IBufferStringIterator, IBufferStringIteratorResult } from './Types';
-import { EventEmitter } from './common/EventEmitter';
 import { IMarker } from 'xterm';
 import { BufferLine } from './BufferLine';
+import { reflowLargerApplyNewLayout, reflowLargerCreateNewLayout, reflowLargerGetLinesToRemove, reflowSmallerGetNewLineLengths } from './BufferReflow';
+import { CircularList, IDeleteEvent, IInsertEvent } from './common/CircularList';
+import { EventEmitter } from './common/EventEmitter';
 import { DEFAULT_COLOR } from './renderer/atlas/Types';
-import { reflowSmallerGetNewLineLengths, reflowLargerGetLinesToRemove, reflowLargerCreateNewLayout, reflowLargerApplyNewLayout } from './BufferReflow';
+import { BufferIndex, CharData, IBuffer, IBufferLine, IBufferStringIterator, IBufferStringIteratorResult, ITerminal } from './Types';
 
 export const DEFAULT_ATTR = (0 << 18) | (DEFAULT_COLOR << 9) | (256 << 0);
 export const CHAR_DATA_ATTR_INDEX = 0;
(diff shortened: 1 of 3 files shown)


src/Buffer.ts
```diff
@@ -285,6 +289,13 @@ export class Buffer implements IBuffer {
         wrappedLines.unshift(nextLine);
       }
 
+      // If these lines contain the cursor don't touch them, the program will handle fixing up
+      // wrapped lines with the cursor
+      const absoluteY = this.ybase + this.y;
+      if (absoluteY >= y && absoluteY < y + wrappedLines.length) {
```
jerch: What happens here, if the cursor is not in the last row (e.g. after the cursor jumped some lines backwards)? I guess `absoluteY >= y` will turn true for any line after the cursor?

src/Buffer.ts
```diff
@@ -357,10 +368,13 @@ export class Buffer implements IBuffer {
             this.ydisp++;
           }
         } else {
-          if (this.ybase === this.ydisp) {
-            this.ydisp++;
+          // Ensure ybase does not exceed its maximum value
```
jerch: Did you spot overflows here? I mean you do quite some calculations at the beginning to skip line content at the scroll buffer top that would be cut off anyway. Maybe there is an offset error that gets carried on?

Tyriar: There might be a calculation off somewhere, but this and the `this.y` check seemed to fix the majority of the issues. We also want these just in case as invalid values can brick the terminal and make it unusable.

Issue #782, opened 2017-07-11
Background is rendered improperly when specifying standard CSI code \033[40m


jmwilkinson:
Running the escape sequence `\033[40m` [should set the background to a standard black](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors). However, it results in a more grey background, which can be reset using `\033[0m`.

It looks as though the background color is being interpreted as "bright" rather than normal. 

It appears to be due to https://github.com/sourcelair/xterm.js/blob/master/src/xterm.css#L229

This could be replaced with:

```css
.terminal .xterm-bg-color-0 {
    background-color: #000000;
}
```

and it would behave as expected.

## Details
- Browser and browser version:  All browsers 
- OS version: All
- xterm.js version: 2.7.0, 2.8.1

### Steps to reproduce

[JS Fiddle](https://jsfiddle.net/6zr98d1c/)

```
  <!doctype html>
  <html>
    <head>
      <link rel="stylesheet" href="bower_components/xterm.js/dist/xterm.css" />
      <script src="bower_components/xterm.js/dist/xterm.js"></script>
    </head>
    <body>
      <div id="terminal"></div>
      <script>
      	var term = new Terminal();
        term.open(document.getElementById('#terminal'));
        term.write('I am \033[40mbroken\033[0m $ ')
      </script>
    </body>
  </html>
```
![image](https://user-images.githubusercontent.com/17836030/28087968-2764049c-6639-11e7-9b32-955a24d1bf00.png)

parisk:
Ha! Interesting. I am putting this on the milestone of the 3.0 release.

The best way to solve this is to start using a CSS preprocessor, in order to keep consistent colors across different points of the stylesheet.

Tyriar:
This was actually intentional so that you could see the black color on the background, I'm not opposed to changing it to #000 though. You would also want to change `.terminal .xterm-color-0`

Tyriar:
Deferring to unblock v3

Tyriar:
I'm closing this off to keep out issue count lower as it hasn't gained any traction, is stale or is not worth tracking.

jmwilkinson:
This is still an issue, and there is no longer an obvious workaround using CSS.

jmwilkinson:
Related, it is unclear why the color won't reset when switching between foreground blacks:

[...]

PR #1922, opened 2019-02-01
Fixes inconsistent CSI '40m' black color

jmwilkinson:
This fixes the issue described in https://github.com/xtermjs/xterm.js/issues/782.

Essentially, reset the default black used for resetting the terminal and the CSI `40m` black color are very different, and they should not be.


Tyriar:
Our behavior is correct:

```
            Ps = 4 0  -> Set background color to Black.
            Ps = 4 1  -> Set background color to Red.
            Ps = 4 2  -> Set background color to Green.
            Ps = 4 3  -> Set background color to Yellow.
            Ps = 4 4  -> Set background color to Blue.
            Ps = 4 5  -> Set background color to Magenta.
            Ps = 4 6  -> Set background color to Cyan.
            Ps = 4 7  -> Set background color to White.
            Ps = 4 9  -> Set background color to default, ECMA-48 3rd.
```

https://invisible-island.net/xterm/ctlseqs/ctlseqs.html

Black !== default

PR #1923, opened 2019-02-01
signal sequences was handled

### LINKED ISSUE #1879: fallback handler wrongly called for OSC and CSI
With the new add handler logic the fallback handler gets wrongly called, although a default handler is registered.

jerch:
Inputhandler methods did not yet follow the paradigm to return `true` if a sequence was finally handled, thus the fallback handler get called. Fixes #1879.

### DIFF
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -446,15 +446,16 @@ export class InputHandler extends Disposable implements IInputHandler {
    * BEL
    * Bell (Ctrl-G).
    */
-  public bell(): void {
+  public bell(): boolean {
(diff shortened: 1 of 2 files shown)


Tyriar:
What if we change it to `callback: (params: number[], collect: string) => boolean | undefined` and only when it's `false` (not undefined) will it be treated as not handled? Would save adding all this extra code to InputHandler.

jerch:
@Tyriar Hmm yeah, tbh I dont like adding all those return values either. Guess I'd do it your way.

jerch:
@Tyriar :+1:

PR #1927, opened 2019-02-02
*actually* fix mouse handler before term attached

pfitzseb:
This is a follow-up for https://github.com/xtermjs/xterm.js/pull/1905, which I apparently didn't test properly -- turns out `_terminal.selectionManager` isn't initialized either until `terminal.open()` is called...

With this additional fix opening a tmux session in an un-`open`ed terminal works. 

It would probably be a good idea to test this somewhere, but I'm not sure where to put this kind of integration test.


pfitzseb:
Argh, I messed this up again (got `disable` and `enable` the wrong way around)... Should be good to go now though.

pfitzseb:
Thanks for the style fix (I should really try to remember to run a linter locally before pushing) and merge! :)

Tyriar:
@pfitzseb np, CI made it easy for me to do 😃.

PR #1929, opened 2019-02-04
fix: Renderer: IntersectionObserver can produce more then 1 entry

coderaiser:
[IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) can produce more than 1 entry, and latest is the most true, then first.

I'm using `xterm.js` in [Cloud Commander](cloudcmd.io) as [gritty](https://github.com/cloudcmd/gritty) middleware in a [modal](https://github.com/cloudcmd/modal) window. And I have a bug which is reproduced in the latest version of chrome but not in firefox. The thing is firefox shows modal with a terminal and output information because it has exactly one entry, and chrome has two entries, and second has `intersectionRatio=1`, but first has `intersectionRatio=0` so it shows empty screen.(https://github.com/coderaiser/cloudcmd/issues/214).


coderaiser:
@Tyriar It is always returns an array, so it is definitely should be iterated. You can look at scroll example on the the link above: when the element you observing appears `intersectionRatio` becomes `1` and you can see percents change.

It should not brake other browsers (then `chrome` and `firefox`), because it relates only on output of information to terminal, so the reasoning is when terminal is loading - output nothing, finished loading - show everything.

In any case, last `IntersectionObserverEntry` is an actual result, all of them has `time` property, so last one has a bigger time, so it is what most closely is going on to current moment of a thing you observing.

And main difference of how it was before: you observed an older entry, which has lower time, and which is already done in the past so the `terminal` window in the `modal` of `Cloud Commander` is empty and needs restart.

If behavior I'm talking about is inappropriate would be great to make using `IntersectionObserver` optional so I do not need to remove it from `window` to avoid detection, maybe other libraries  also depend on it.

PR #1936, opened 2019-02-11
Remove font-family from .css file

### LINKED ISSUE #1935: We probably don't need to set font-family on the terminal anymore
Since the renderers own how text is drawn I don't think we need this line anymore:

https://github.com/xtermjs/xterm.js/blob/abf6a9fa58704de6f75184f8003fc233751053ce/src/xterm.css#L39

DOM renderer sets family here:

https://github.com/xtermjs/xterm.js/blob/abf6a9fa58704de6f75184f8003fc233751053ce/src/renderer/dom/DomRenderer.ts#L154

Canvas renderer sets family here:

https://github.com/xtermjs/xterm.js/blob/abf6a9fa58704de6f75184f8003fc233751053ce/src/renderer/BaseRenderLayer.ts#L347

Tyriar:
Fixes #1935

### DIFF
--- a/src/CompositionHelper.ts
+++ b/src/CompositionHelper.ts
@@ -204,6 +204,7 @@ export class CompositionHelper {
       this._compositionView.style.height = cellHeight + 'px';
       this._compositionView.style.lineHeight = cellHeight + 'px';
       this._compositionView.style.fontFamily = this._terminal.options.fontFamily;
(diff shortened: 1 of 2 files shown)


Tyriar:
@mofux fixed the IME issue by setting the font 👍 

Before:

image

After:

image

Issue #1901, opened 2019-01-18
support non-text BufferLine


PerBothner:
Suppose one wanted to support inline images (issue #614, pull request #1577) or a block of rich text (perhaps for help messages).  Ideally one wants to position these are anywhere, interleaving them with character cells. However, I think you'd get 90% of the benefit by restricting these areas to cover the entire line width.  This should be easier to implement, with fewer performance issues.

The contents of a Buffer is an array of BufferLine, which is (conceptually) an array of CellData. We can generalize that to an array of BufferSection, where a BufferSection can be a BufferLine - or other things:

    class BufferSection {
        yOffset: number;
        ...
    }
    class BufferLine extends BufferSection { ... }

    /** Arbitrary DOM element (HTML data) wrapped in a <div> */
    class BufferDivSection extends BufferSection {
         element: HTMLDivElement;
    }

    class BufferCanvasSection extends BufferSection {
         element: HTMLCanvasElement;
    }

Some previous prototypes have split an image into "bands" each the height of a line, using multiple BufferLines per image.  I think this would be a mistake.  One reason is it unnaturally forces each image/area to be an integer number of lines high. While it make be plausible to split an image into bands, this is no logical concept of bands for (say) BufferDivSection.

The implication is that BufferSection can have varying heights, which mildly complicates mapping from BufferSection to screen position and back (say for mouse clicks). But that is easy to solve: Just store the yOffset in each BufferSection.  To map (say) a mouse click to a BufferSection, just use binary search.

Note we do not *prohibit* splitting an image into multiple line-high bands.  That could be useful for sixel images or other images which are aligned to character cells.  However, I don't think that ability is a priority.

A minor bonus: We can easily handle double-height characters using a regular BufferLine.

I can work on this if it makes sense.

Tyriar:
@mofux had a nice idea (https://github.com/xtermjs/xterm.js/issues/791#issuecomment-448934018) of aligning more with how Monaco does it which is similar to your proposal. Instead of dealing with the largely unrelated `BufferLine` you would inject a view at some marker (a line that tracks where it is in the buffer). This is closely related to the line decorator proposal here https://github.com/xtermjs/xterm.js/issues/1852

Maybe something like this (heavily inspired by [Monaco's API](https://github.com/Microsoft/vscode/blob/f20fb4e7c907a5835f4bc9c5680614220efabf7c/src/vs/monaco.d.ts#L3514-L3520))?

```ts
class Terminal {
[...]

Issue #1140, opened 2017-12-11
Smooth scrolling


baukevanderlaan:
As someone who’s new to working with CLI’s I find it cognitively harder to view the line by line scrolling  used in terminals as opposed to the smooth scrolling of most other applications. I learned that it’s because in a terminal, the character and not the pixel is the smallest display unit.

Is it possible (and desirable) for xterm.js to work around this and incorporate a smooth pixel-based scrolling option?

I originally asked @albinekb the same thing for Hyper (https://github.com/zeit/hyper) and he referred me to this project since Hyper uses xterm.js.

Tyriar:
It's certainly possible to do but it would be quite a lot of work (+added complexity) for something that so few people would use. I'd prefer not to add this personally.

baukevanderlaan:
That's understandable and fair. Thanks for your work!

egmontkob:
FYI: VTE (GNOME Terminal and friends) [does this](https://bugzilla.gnome.org/show_bug.cgi?id=746690) since version 0.44. (Of course it only works in the scrollback of the terminal emulator, not when e.g. scrolling the file in a text editor.)

egmontkob:
: something that so few people would use

Everyone who scrolls with the touchpad uses it in VTE, whether they like it or not :D

Tyriar:
@egmontkob I have gnome terminal 3.18.3 here and it doesn't seem to work, the scrolling section in profile settings also doesn't seem to have a setting?

<img width="460" alt="screen shot 2017-12-18 at 6 57 19 am" src="https://user-images.githubusercontent.com/2193314/34112209-bb156140-e3c0-11e7-8f41-1f4e708eda04.png">

egmontkob:
g-t 3.18.3 most likely means you have vte 0.42, whereas this feature is new in 0.44.

No new setting (there's a pending [low-prio request](https://bugzilla.gnome.org/show_bug.cgi?id=786778) to add one), it just behaves different than before.

egmontkob:
Someone already expressed their dislike about this not being configurable in VTE. Of course it's totally your choice whether you implement this feature at all in xtermjs, and if you do, whether you make it configurable.

[...]

PR #1940, opened 2019-02-12
Experimental Sixel and embedded graphics support.

PerBothner:
This is an experimental prototype for Sixel images and potentially other embedded graphics.  This is definitely work-in-progress.  I've been testing this with the [DomTerm embedding](https://github.com/PerBothner/DomTerm/blob/master/README-xtermjs.md) but it should probably work for other embeddings. You can then `cat images/*.six` from [libsixel]( https://github.com/saitoha/libsixel).

This is logically two parts:

* Support for ElementInset objects, which attach an HtmlElement to the view. This is actually the more interesting and useful (and unfinished) part.

* Parsing Sixel graphics, using the [node-sixel](https://github.com/jerch/node-sixel) library written by @jerch , converting that to an Image, which is then embedded in an ElementInset. For this prototype I just copied node-sixel to the `src/sixel-image.ts` - it should probably go elsewhere.

Scrolling is handled by making the ElementInset element be children of the `_viewPortScrollArea`.

Likecycle management (disposing) and `trim` are missing.  There is also missing handling for inserting/deleting lines: Insert/delete before the image should adjust the position.  We also need to delete the image when the owning line(s) are deleted.  (That would allow sixel images to be over-written.)

Longer term I think this would be cleaner by allocating a single "tall line" for the image, instead of multiple regular lines - see discussion issue #1901.  However, that depends on various issues, including smooth scrolling (issue #1140).

Comments/suggestions appreciated.

### DIFF
--- a/src/Buffer.ts
+++ b/src/Buffer.ts
@@ -3,7 +3,7 @@
  * @license MIT
  */
 
-import { IMarker, IDisposable } from 'xterm';
+import { IMarker } from 'xterm';
 import { BufferLine } from './BufferLine';
 import { reflowLargerApplyNewLayout, reflowLargerCreateNewLayout, reflowLargerGetLinesToRemove, reflowSmallerGetNewLineLengths } from './BufferReflow';
 import { CircularList, IDeleteEvent, IInsertEvent } from './common/CircularList';
@@ -667,12 +667,13 @@ export class BufferStringIterator implements IBufferStringIterator {
   }
 }
 
-export class ElementInset implements IElementInset, IDisposable {
+export class ElementInset extends EventEmitter implements IElementInset {
     element: HTMLElement;
     wrapper: HTMLElement;
     owner: IBufferLine;
     marker: Marker;
     constructor(private _terminal: any, element: HTMLElement) {
+        super();
         this.element = element;
         const wrapper = document.createElement('div');
         wrapper.classList.add('inset-wrapper');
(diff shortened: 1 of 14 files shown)


PerBothner:
The prototype is getting close to usable now.

PerBothner:
Quake (using SDL-sixel) runs pretty well.

AndrienkoAleksandr:
> Quake (using SDL-sixel) runs pretty well.
I guess you mean Guake.

PerBothner:
_"I guess you mean Guake."_

No, I mean Quake, the game. Specifically [sdlquake](https://github.com/ahefner/sdlquake) using [sdl-sixel](https://github.com/saitoha/SDL1.2-SIXEL).

You can download a pre-built (Linux 64-bit) binary from [here](https://repl.it/@basicer/quake). (Select one of the other files. Then you get a 3-dots menu, and you can click "Download as zip".)  This works (at reasonable speed) with my patches (tested in my DomTerm+xterm.js prototype).

AndrienkoAleksandr:
@PerBothner It would be nice if you record video with this feature, convert to diff and attach to this pr.

PerBothner:
I implemented the DomTerm OSC '72' escape sequence to embed general HTML:

xterm-image

There is no "scrubbing" of the HTML, but of course that can and should be done.  (DomTerm has a scrubber that could be ported.)

The screenshot shows a timing glitch with the calculation of the image height: The first time the image is inserted the image height (clientHeight) is reported as 0, which means we don't allocate enough vertical space for it.  The second time we do get the correct height.  This is somewhat mysterious to me; I'm guessing adding some kind of timer may be needed.

I think the best fix will be to support variable-height lines (as in issue #1901), which is tied in to smooth scrolling (issue #1140) which is gated by webgl.  But once we have real variable-height lines then we can also support deferring calculating the height until render-time, and we can support the height changing (based on widow width and magnification).

Tyriar:
Sorry about the delay, I think the whole team has been busy with other stuff for the past month.

Ultimately I think sixel support is not a priority and should probably be part of an addon instead, this is a lot of code which so few people would actually utilize in the end. In the same vein we wouldn't want to ship OSC 72 as it's non-standard and not well supported by other terminals. While I appreciate the effort, I don't think this is the direction we want to go, at least for now.

Also, after the long discussion in https://github.com/xtermjs/xterm.js/issues/1901, the team and I developed a [roadmap](https://github.com/xtermjs/xterm.js/wiki/Roadmap) to help keep us focused on what's most important to focus on. The unfortunately truth is that reviewing and merging huge changes like this takes a lot of time that we may not have or that would be better spent on something else.

PerBothner:
I created an [issue](https://gitlab.freedesktop.org/terminal-wg/specifications/issues/12) on the freedesktop.org [terminal-wg](https://gitlab.freedesktop.org/terminal-wg) to try to standardize "simple image display".  I hope we can get agreement on a suitable semi-portable escape sequence, and that you'd be more willing to accept that.

Xeddius:
Disappointed to see this was passed over, I very much would like for this to become a standard feature as sixels allow for flexible terminal images and even allows for backwards compatibility with DEC sixels from the old dec terminals.

Tyriar:
@Xeddius @jerch has been experimenting with this direction some bit more in https://github.com/xtermjs/xterm.js/pull/2503

Issue #1946, opened 2019-02-16
unify trim/delete/insert events


PerBothner:
I believe various issues would be simplified if we unified `trim`, `delete`, and `insert` events on a Buffer's lines.  Most obviously, a `trim` is just a `delete` where the event `index` is 0.  Similarly, an `insert` can be treated as a `delete` with a negative `amount`.  I believe any code that needs to handle one of these events needs to handle them all - just look at at the code for `Buffer.addMarker`.

Also notice the ElementInsert.put method (in Buffer.ts) in my sixel pull request #1940.

Not sure whether to call the unified event 'delete' or come up with a more neutral name.

I noticed that the selection is discarded when the window is resized.  That is a symptom of not handling these events in a unified manner: The `SelectionManager` watches for `trim` events, but does not handle `delete` or `insert` events.  That should be fixed.

A related issue is that deleteLines/insertLines/eraseInDisplay should perhaps also signal a delete event.  I need something similar so images (sixel and otherwise) can get disposed when appropriate.

There will some compatibility problems. Addons that have `trim` handlers will have to be updated.  It seems possible to kludge it so that old trim handlers are called when the index is 0.

Tyriar:
> I noticed that the selection is discarded when the window is resized. That is a symptom of not handling these events in a unified manner: The SelectionManager watches for trim events, but does not handle delete or insert events. That should be fixed.

I can't think of a way to actually hit this as an end user, please create another issue if you have a repro.

I did hit this when trying to repro though https://github.com/xtermjs/xterm.js/issues/1963

PR #1948, opened 2019-02-19
Unify trim/delete/insert events as 'move' events.

PerBothner:
This solves GitHub issue #1946.

### DIFF
--- a/src/SelectionModel.ts
+++ b/src/SelectionModel.ts
@@ -4,6 +4,7 @@
  */
 
 import { ITerminal } from './Types';
+import { IMoveEvent } from './common/CircularList';
 
 /**
  * Represents a selection within the buffer. This model only cares about column
@@ -112,13 +113,15 @@ export class SelectionModel {
    * @param amount The amount the buffer is being trimmed.
    * @return Whether a refresh is necessary.
    */
-  public onTrim(amount: number): boolean {
+  public onMove(event: IMoveEvent): boolean {
     // Adjust the selection position based on the trimmed amount.
-    if (this.selectionStart) {
-      this.selectionStart[1] -= amount;
+    if (this.selectionStart && this.selectionStart[1] >= event.start
+        && (! event.end || this.selectionStart[1] < event.end)) {
+      this.selectionStart[1] -= event.amount;
     }
-    if (this.selectionEnd) {
-      this.selectionEnd[1] -= amount;
+    if (this.selectionEnd && this.selectionEnd[1] >= event.start
+        && (! event.end || this.selectionEnd[1] < event.end)) {
+      this.selectionEnd[1] -= event.amount;
     }
 
     // The selection has moved off the buffer, clear it.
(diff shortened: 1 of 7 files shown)


src/SelectionModel.ts
```diff
@@ -112,13 +113,15 @@ export class SelectionModel {
    * @param amount The amount the buffer is being trimmed.
    * @return Whether a refresh is necessary.
    */
-  public onTrim(amount: number): boolean {
+  public onMove(event: IMoveEvent): boolean {
     // Adjust the selection position based on the trimmed amount.
-    if (this.selectionStart) {
-      this.selectionStart[1] -= amount;
+    if (this.selectionStart && this.selectionStart[1] >= event.start
+        && (! event.end || this.selectionStart[1] < event.end)) {
```
Tyriar: Personally I think the current way is much more readable and explicit in explaining what is actually happening, the need to check for `end?` and handle it is one of the main reasons I think this.

Another good example:

```
// Where is it moving?
model.onMove({start: 0, amount: 1});
```

A negative amount is also a little confusing unless you go and read the definition of `IMoveEvent`.

Tyriar: Thoughts @jerch?

PerBothner:
I see the main reason for unifying them is that I believe anywhere you'd need a handler for one would need a handler for all three.  Having to write 3 handlers is both more verbose and more error-prone (likely to forget one or make a mistake in one).

src/SelectionModel.ts
jerch: Gonna answer below to stay in the thread flow...

jerch:
Well I dont quite the point of this idea, tbh. Is this just for API unification? Imho this is more a matter of taste (code style) than it shows real benefits. It also has drawbacks. A "one event fits all purposes" paradigm kinda defeats language features like strong typing (well for TS at least during compile time), you will have to read the docs to grasp the correct meaning of an event with parameters xy. With separate events you still need to know what a particular events means, but TS can assist you through several levels of invocations.
If you really need to deal with several events in one block of code, you can still register to all of the needed events. Again TS can help you in your code block with the type system.

Thats all not possible with a single event type, where you instead have to implement the meaning orthogonal to the type system. Much like redux's actions with TS, they kinda feel silly in TS as they cannot not leverage the type system well.

PerBothner:
I claim there is no functional difference between a 'trim' and a 'delete' event, and there is no place in the code (or possible addons) where you would want to handle one and not the other, or handle them differently: 'trim' is just a special case of 'delete'. When it comes to 'insert' vs 'delete', the only difference is one of sign, and I also believe there is no place where you would want to handle one and not the other.

This to me argues strongly for unifying these events.

Tyriar:
> 'trim' is just a special case of 'delete'

This is a good point and I could see us merging these, but it would only result in more GC. Right now I find things fairly readable and merging these events would mean a new object would need to be created for every trim as the trim event just passes in the amount with the event right now. Since thousands+ of these can fire a second that's a lot of `IDeleteEvent` (or `IMoveEvent`) objects needed to be created and GC'd.

We'll just have to be diligent in making sure we listen to everything 🙂

PR #1950, opened 2019-02-20
Disable reflow when winptyCompat is on

### LINKED ISSUE #1943: Reflow can still happen on Windows which can break the terminal
It's likely to do with flagging lines as wrapped via winptycompat but maybe some rows are getting flagged as wrapped incorrectly on resize? https://github.com/Microsoft/vscode/issues/68721

Tyriar:
Fixes #1943

---

This is a quick and dirty fix that should completely disable reflow when winptyCompat is on (ie. the terminal backend is Windows-based). This prevents a potential crash.


Tyriar:
@jerch I want to have this addon built in later as part of https://github.com/xtermjs/xterm.js/issues/1128, so the hack will go away at that point.

PR #1951, opened 2019-02-25
Fix emoji width

minvedacat:
## Problem
As described in https://github.com/xtermjs/xterm.js/issues/469#issuecomment-451789158, emoji width is 1 when added then displayed in half, and delete width is 2 when it is deleted which caused character in the left is deleted by mistake.

### Conditions
1. Most emoji are 2 in cell width like😂, whose unicode is `\u55357`+`\u56834`
2. Some symbol emoji like ❌ is single unicode `\u10060`.
3. The monstrous zero code joiner 👩‍❤️‍💋‍👩 is `\u{200D}`, so`'👩‍❤️‍💋‍👩'.length === 11`, when copy & paste in Chrome address, it becomes  👩‍ ❤️‍ 💋‍ 👩 separated by `\u8205`

This only fix emoji width, IME support is WIP here https://github.com/xtermjs/xterm.js/pull/1890

## Solution
Single 'string' width is calculated by `wcwidth` which doesn't consider emojis. 
By extending `getStringCellWidth`, it can calculate both single 'string' and normal string width, containing either `wcwidth` standing characters and unicode emojis. Terminal should use `getStringCellWidth` other than `wcwidth` to cover emoji width calculation.

## References
https://mathiasbynens.be/notes/javascript-unicode
https://blog.jonnew.com/posts/poo-dot-length-equals-two

## Testing
- [x] Normal double-width emoji as 😂
- [ ] Normal single-width emoji as ❌
- [ ] Emoji with zero code joiner as 👩‍❤️‍💋‍👩, which is 👩‍ ❤️‍ 💋‍ 👩
- [ ] Cell width break `Buffer stringIndexToBufferIndex fullwidth combining with emoji - match emoji cell`. 'Lots of ￥\u0301 make me 😃.'  is 3 lines in `terminal.buffer.lines.get(some_line_index)` like:
```
Lots of ￥́
 make me 
😃.
```

### DIFF
--- a/src/CharWidth.test.ts
+++ b/src/CharWidth.test.ts
@@ -76,6 +76,20 @@ describe('getStringCellWidth', function(): void {
     assert.equal(getStringCellWidth(s), sumWidths(terminal.buffer, 0, 1, '#'));
   });
   // TODO: multiline tests once #1685 is resolved
+  it('emojis', function(): void {
+    const input = '😁😁😁😁😁😁#';
+    terminal.writeSync(input);
+    const s = terminal.buffer.iterator(true).next().content;
+    assert.equal(input, s);
+    assert.equal(getStringCellWidth(s), sumWidths(terminal.buffer, 0, 1, '#'));
+  });
+  it('emojis with combining chars', function(): void {
+    const input = 'e\u0301e\u0301😁😁😁😁😁😁#';
+    terminal.writeSync(input);
+    const s = terminal.buffer.iterator(true).next().content;
+    assert.equal(input, s);
+    assert.equal(getStringCellWidth(s), sumWidths(terminal.buffer, 0, 1, '#'));
+  });
 });
 
 it('wcwidth should match all values from the old implementation', function(): void {
--- a/src/CharWidth.ts
+++ b/src/CharWidth.ts
@@ -138,6 +138,40 @@ export const wcwidth = (function(opts: {nul: number, control: number}): (ucs: nu
   };
 })({nul: 0, control: 0});  // configurable options
 
+// from another emoji fix https://github.com/xtermjs/xterm.js/pull/1890/files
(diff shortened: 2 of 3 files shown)


src/CharWidth.test.ts
```diff
@@ -76,6 +76,20 @@ describe('getStringCellWidth', function(): void {
     assert.equal(getStringCellWidth(s), sumWidths(terminal.buffer, 0, 1, '#'));
   });
   // TODO: multiline tests once #1685 is resolved
+  it('emojis', function(): void {
+    const input = '😁😁😁😁😁😁#';
```
epicfaace: I haven't seen emojis in code haha this is great

minvedacat: > I've never seen emojis in code before haha this is great

Actually there were a few and the failed test is one of them😂

Tyriar:
I think this is trying to solve the same problem as https://github.com/xtermjs/xterm.js/pull/1714? @jerch

minvedacat:
> I think this is trying to solve the same problem as #1714? @jerch 
This PL will not solve the root cause which is the unicode support.
Additional calculation is involved to get real displaying width and the result varies on different OS.

jerch:
Yes thats part of #1714. Unless we properly respect different Unicode versions we cannot make this work reliably. I have to vote against this patch here, as it only covers some chars while others are left out. Furthermore this will break with systems, that still use a lower Unicode version, like Ubuntu 16 or older OSX versions (before High Sierra).

Tyriar:
@jerch ah k that's what I suspected, @minvedacat thanks for the PR but unfortunately this needs to wait for #1714 to fix without regressing elsewhere (which is blocked on https://github.com/xtermjs/xterm.js/issues/1128).

PR #1953, opened 2019-03-01
 docs: style improvements

cherouvim:
- Added a trailing dot at the end of each list line.
- Consistent usage of `xterm.js` in "Real-world uses".

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,17 +7,17 @@ Xterm.js is a front-end component written in TypeScript that lets applications b
 
 ## Features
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -106,9 +106,9 @@ Note that some APIs are marked *experimental*, these are added to enable experim
 ## Real-world uses
 Xterm.js is used in several world-class applications to provide great terminal experiences.
 
-- [**SourceLair**](https://www.sourcelair.com/): In-browser IDE that provides its users with fully-featured Linux terminals based on xterm.js
```
Tyriar: Could we switch this back to just xterm.js from `xterm.js`? I think that's more fitting since it's not a file or code but a lib name

cherouvim: Correct. Done.

Issue #1954, opened 2019-03-03
webLinks not interpreted


ppot:
## Details
- Browser and browser version: 
- OS version: 
- xterm.js version: 

Referencing issue : https://github.com/zeit/hyper/issues/3503
### Steps to reproduce

1. Write `http://example.io`
2. mouse over and the link is not interpreted as valid link you can't click it.

Tyriar:
Code pointer: https://github.com/xtermjs/xterm.js/blob/8ea254bdc798ffcf72cb1b37385af40c747d3fa4/src/addons/webLinks/webLinks.ts#L12

PR #1957, opened 2019-03-06
Fixes issue #1954

trowacat:
See #1954
Links with tld of 2 chars is now being interpreted


src/addons/webLinks/webLinks.ts
```diff
@@ -9,7 +9,7 @@ const protocolClause = '(https?:\\/\\/)';
 const domainCharacterSet = '[\\da-z\\.-]+';
 const negatedDomainCharacterSet = '[^\\da-z\\.-]+';
 const domainBodyClause = '(' + domainCharacterSet + ')';
-const tldClause = '([a-z\\.]{2,6})';
+const tldClause = '([a-z\\.]{1,6})';
```
Tyriar: Can you explain this change as I'm a bit puzzled as to why it wasn't working before. hostClause starts with:

```
const hostClause = '((' + domainBodyClause + '\\.' + tldClause + ')
```

Isn't tldClause matching only the "com" part from ".com"? I would have thought {2,6} would capture things like "io" and "com.au"?

trowacat: @Tyriar I thought this was odd as well but upon testing it appeared to fix the issue. I've looked deeper into it and have learned it was the pathClause which was causing the real problem in marking "io" or "com.au" as false.

Tyriar: Hmm, if we change pathClause like this we'll probably regress in not including the quotes or colon at the end of a URL (see https://github.com/xtermjs/xterm.js/commit/9cd7bf2af1178c29d6be4254bd13ed0c191ae299)

Tyriar:
I looked into this a bit but I can't figure out how to do it without regressing in other important areas like supporting links like this without including the last semi-colon:

```
http://foo.bar:
http://foo.bar/:
http://foo.bar/foo:bar:
```

I think this is one of those issues that needs to get fixed when a non-regex link solution exists https://github.com/xtermjs/xterm.js/issues/583

Issue #1947, opened 2019-02-17
webLinks plugin issue. Wrong behaviour on selection


vitaliy-codeocean:
If i will finish selection in the middle of some link and release mouse button, link will be clicked.
So in general im just trying to copy part of the text but opening link in the end :)

## Details
- Browser and browser version: Google Chrome Version 72.0.3626.109
- OS version: macOS 10.14.3
- xterm.js version: 3.11.0

### Steps to reproduce

1. Start to select text
2. Stop when your cursor will be over some link
3. Link is opened

Tyriar:
I guess to prevent this we could make sure not selection change has occurred (or no selection exists?) when a link is about to be activated.

PR #1958, opened 2019-03-06
WebLinks plugin don't follow link on selection

nikonso:
This is not the correct solution, but it's something to get the ball rolling...

Feedback on what's the best approach is welcome, right now it is in a working state, but I believe we should be smarter about this... I have 2 questions regarding how to proceed:

1. Should this logic live in the plugin, or should it go to the `Linkifier`?
1. If I wanted to be smarter about the selection, what's the recommended way to access non public API's in the addon world?

Fixes https://github.com/xtermjs/xterm.js/issues/1947

### DIFF
--- a/src/addons/webLinks/webLinks.ts
+++ b/src/addons/webLinks/webLinks.ts
@@ -36,6 +36,13 @@ function handleLink(event: MouseEvent, uri: string): void {
  */
 export function webLinksInit(term: Terminal, handler: (event: MouseEvent, uri: string) => void = handleLink, options: ILinkMatcherOptions = {}): void {
   options.matchIndex = 1;
+
+  handler = (event, uri) => {
+    if (!term.hasSelection()) {
+      window.open(uri, '_blank');
+    }
+  };
+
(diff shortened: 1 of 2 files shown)


src/addons/webLinks/webLinks.ts
```diff
@@ -36,6 +36,13 @@ function handleLink(event: MouseEvent, uri: string): void {
  */
 export function webLinksInit(term: Terminal, handler: (event: MouseEvent, uri: string) => void = handleLink, options: ILinkMatcherOptions = {}): void {
   options.matchIndex = 1;
+
+  handler = (event, uri) => {
```
Tyriar: So we definitely can't be overriding `handler` here as that would break custom handlers, but just moving this into handleLink seems like a reasonable fix to me

nikonso:
That seems like a better approach to this problem.. I'll try to have a POC during the week! 😄

nikonso:
After some testing, the recommended approach seems to be working! Awaiting review 😃

Issue #1627, opened 2018-08-24
Remove all circular dependencies


Tyriar:
Once removed we could add this tslint rule to prevent future ones https://www.npmjs.com/package/tslint-no-circular-imports

Related: https://github.com/xtermjs/xterm.js/issues/1507

Tyriar:
This is happening in https://github.com/xtermjs/xterm.js/issues/1507

PR #1961, opened 2019-03-07
Fix #1773 - drawBoldTextInBrightColors is not taken into account with DOM renderer

### LINKED ISSUE #1773: drawBoldTextInBrightColors is not taken into account with DOM renderer
## Details
- Browser and browser version: Chrome 70
- OS version: Mac OS X
- xterm.js version: 3.8.0

### Steps to reproduce

1. Set `drawBoldTextInBrightColors` to `false` and `rendererType` to `dom`
2. Bold characters are still rendered brighter

It works as expected with the canvas renderer.

nikonso:
Fix #1773 by taking into account the terminal options

### DIFF
--- a/src/renderer/dom/DomRendererRowFactory.ts
+++ b/src/renderer/dom/DomRendererRowFactory.ts
@@ -5,7 +5,7 @@
 
 import { CHAR_DATA_CHAR_INDEX, CHAR_DATA_ATTR_INDEX, CHAR_DATA_WIDTH_INDEX, CHAR_DATA_CODE_INDEX, NULL_CELL_CODE, WHITESPACE_CELL_CHAR } from '../../Buffer';
 import { FLAGS } from '../Types';
-import { IBufferLine } from '../../Types';
+import { IBufferLine, ITerminal } from '../../Types';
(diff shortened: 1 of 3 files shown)


src/renderer/dom/DomRendererRowFactory.ts
```diff
@@ -17,6 +17,7 @@ export const CURSOR_STYLE_UNDERLINE_CLASS = 'xterm-cursor-underline';
 
 export class DomRendererRowFactory {
   constructor(
+    private _terminal: ITerminal,
```
Tyriar: I've been slowly trying to break this pattern of passing `ITerminal` around everywhere (https://github.com/xtermjs/xterm.js/issues/1627), I think passing in `ITerminalOptions` here would be preferable as we never reassign that on `Terminal`.

nikonso:
Makes sense, I'll make the changes Sunday when I'm back home ??

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

Tyriar:
Sounds good, thanks a bunch as always 😃

nikonso:
Circular dependency removed, and verified that it still works!

PR #1978, opened 2019-03-23
Let consumers decide whether winptyCompat should be active

Tyriar:


Tyriar:
This is a soft API breakage, some people may be expecting this to not work when called on non-Windows, we should call out in the release notes.

PR #1993, opened 2019-04-02
Replace winptyCompat addon with windowsMode option

Tyriar:
Part of #1128


Tyriar:
Removes an addon so let's note it in release notes.

PR #1995, opened 2019-04-03
Fix CSI scroll down handler

thekondr:
The issue has been introduced in cd8477a942e1fac85de167eab7e8c09c85f55255


Tyriar:
Thanks for this, looks right 😃 fyi @jerch

Issue #1994, opened 2019-04-02
Buffer API


Tyriar:
As part of https://github.com/xtermjs/xterm.js/issues/1128 we will be pulling the search addon out of the core repo and into a satellite repo to be published as its own npm module. This is exactly the sort of thing that would be cool for addons to be able to build so we should stabilize some API as right now it hooks into internals. Here's what it touches in buffer:

- ydisp
- ybase
- length
- translateBufferLineToString
- lines (for isWrapped, length, get width, get char to check if emoji)

Given this, I'm proposing the following API. It resembles the internal structure, but I tweaked a bit here and there to make it more consistent and easier to use (eg. `buffer.lines.get` -> `buffer.getLine`):

```ts
class Terminal {
    /**
     * The terminal's current buffer, note that this might be either the normal
     * buffer or the alt buffer depending on what's running in the terminal.
     */
    buffer: IBuffer;
}

interface IBuffer {

    /**
     * The y position of the cursor. This ranges between `0` (when the
     * cursor is at baseY) and `Terminal.rows - 1` (when the cursor is on the
     * last row).
     */
    readonly cursorY: number;

    /**
     * The x position of the cursor. This ranges between `0` (left side) and
     * `Terminal.cols - 1` (right side).
     */
    readonly cursorX: number;

    /**
     * The line within the buffer where the top of the viewport is.
     */
    readonly viewportY: number;

    /**
     * The line within the buffer where the top of the bottom page is (when
     * fully scrolled down);
     */
    readonly baseY: number;

    /**
     * The amount of lines in the buffer.
     */
    readonly length: number;

    /**
     * Gets a line from the buffer.
     * 
     * @param y The line index to get.
     */
    getLine(y: number): IBufferLine;
}

interface IBufferLine {
    /**
     * Whether the line is wrapped from the previous line.
     */
    readonly isWrapped: boolean;

    /**
     * Gets a cell from the line.
     * 
     * @param x The character index to get.
     */
    getCell(x: number): IBufferCell;

    /**
     * Gets the line
     */
    translateToString(trimRight: boolean, startColumn: number, endColumn: number): string;
}

interface IBufferCell {
    /**
     * The character within the cell.
     */
    readonly char: number;

    /**
     * The width of the character. Some examples:
     * 
     * - This is `1` for most cells.
     * - This is `2` for wide character like CJK glyphs.
     * - This is `0` for cells immediately following cells with a width of `2`.
     */
    readonly width: number;
}
```

PR #1997, opened 2019-04-04
Fix backspace on demo

### LINKED ISSUE #1989: Backspace event in the demo does not function properly (missing term.x)
The backspace event handler in the demo looks at `term.x` to decide whether or not to emit backspace characters.  Unfortunately, `term.x` does not exist, so the backspace key simply fails to work at all.

## Details
- Browser and browser version: N/A
- OS version: N/A
- xterm.js version: N/A

### Steps to reproduce

1.  Visit https://xtermjs.org/
2.  Type some text into the demo.
3.  Press backspace.

### Additional info

At:

https://xtermjs.org/js/demo.js

The backspace key is being detected but compares a non-existent variable against a positive integer (`term.x > 2`), which always evaluates to false.

Tyriar:
buffer.x will probably be public API soon (#1994)

Fixes #1989

### DIFF
--- a/demo/client.ts
+++ b/demo/client.ts
@@ -172,7 +172,7 @@ function runFakeTerminal(): void {
       term.prompt();
     } else if (ev.keyCode === 8) {
      // Do not delete the prompt
-      if (term.x > 2) {
+      if (term._core.buffer.x > 2) {
         term.write('\b \b');
       }
     } else if (printable) {


Tyriar:
PR for website: https://github.com/xtermjs/xtermjs.org/pull/75

PR #2002, opened 2019-04-07
Fix reflow case where wrapped line portions ended in \t

### LINKED ISSUE #1932: TypeError: Cannot read property '_data' of undefined
While resizing the terminal smaller, the following exception occurs. It's not clear exactly how to reproduce. I cat the content of a file and then resize the terminal to a small size:

```
Uncaught TypeError: Cannot read property '_data' of undefined
    at BufferLine.copyCellsFrom (xterm.js:887)
    at Buffer._reflowSmaller (xterm.js:468)
    at Buffer._reflow (xterm.js:388)
    at Buffer.resize (xterm.js:370)
    at BufferSet.resize (xterm.js:1140)
    at Terminal.resize (xterm.js:5078)
    at Terminal.resize (xterm.js:6615)
    at resize (kubebox.js:1351)
    at Item.run (kubebox.js:52341)
    at drainQueue (kubebox.js:52311)
```

## Details
- Browser and browser version: Chrome 69
- OS version: Mac OS X
- xterm.js version: 3.11.0

### Steps to reproduce

1. Resize the terminal smaller

Tyriar:
Fixes #1932

---

The tests explain the problem pretty well, the tricky thing here was fixing this issue while still keeping the correct behavior of moving wide characters to the following line which is why the fix is more complex than one would expect.

### DIFF
--- a/src/Buffer.test.ts
+++ b/src/Buffer.test.ts
@@ -518,6 +518,29 @@ describe('Buffer', () => {
         assert.equal(secondMarker.line, 1, 'second marker should be restored');
         assert.equal(thirdMarker.line, 2, 'third marker should be restored');
       });
+      it('should correctly reflow wrapped lines that end in null space (via tab char)', () => {
+        buffer.fillViewportRows();
+        buffer.resize(4, 10);
+        buffer.y = 2;
+        buffer.lines.get(0).set(0, [null, 'a', 1, 'a'.charCodeAt(0)]);
+        buffer.lines.get(0).set(1, [null, 'b', 1, 'b'.charCodeAt(0)]);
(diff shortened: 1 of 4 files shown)


akosyakov:
I still can reproduce issues with demo:
reflow

I'm resizing from 10 to 82 several times and getting strange ouput. Do i do something wrong?

Tyriar:
@akosyakov other terminals do things like that when you resize into the prompt when I tested, we're just doing what the shell is telling us. You're not seeing the exception that could brick the terminal though right?

akosyakov:
@Tyriar no errors in dev console

PR #2007, opened 2019-04-10
Convert imports to non-relative, format all code using TS Server

### LINKED ISSUE #1314: Use tsconfig baseUrl and absolute imports
I tried to do this last week but ran into issues with the way our build process works. I think the issue was `lib/` trying to find things that only existed in `src/` or something. Moving the tsconfig.json into the src/ directory might fix this.

Tyriar:
We have `./`, `../` and `../../` scattered all over out imports which makes it hard to tell at a glace which imports are coming from where. This PR converts everything over to non-relative via the `baseUrl` TS compiler option. Some other things of note:

- The `baseUrl` for all projects is the repo root, this means that all imports consistently start with `'src/...'` and are sorted nicely (plus we need to do this for files in `src/` to be non-relative)
- A VS Code settings file is added to give hints to the editor to use single quotes and to import non-relative, I think this is the only way to get it to work within VS Code.
- Format on save for VS Code is also turned on so spacing inconsistencies are a thing of the past, this breaks how we were spacing out code in some places but imo it's best to embrace how the TS formatter does things (and there's no setting for that).
- TS auto imports will still import from lib for project references, there's another issue still open for this.
- I've organized imports using TS Server for all files, it seemed to act a little weird for composite projects though as it will open the generated .d.ts file and make the changes there. I organized core/common manually. This will be a bit of a pain for PRs but it's a one time thing and the conflicts are fairly easy to fix since the compiler tells you what's missing.

Fixes #1314


Tyriar:
Need to fix up the browserify build

Tyriar:
Looks like this is a lot more complicated to get right, we would need to lay out the code like this:

- src
  - xterm
    - common
    - other files/folders

With imports like this `'xterm/common...'`. Then it would also run into issues when the `lib` version is used because of this TS issue https://github.com/Microsoft/TypeScript/issues/10866 (closed as designed). It could be done if we packaged up the lib version instead of having all the files split up.

PR #2008, opened 2019-04-10
Explicit event API

### LINKED ISSUE #1505: Replace IEventEmitter in API with explicit handlers

Tyriar:
This PR exposes the new explicit events API that will replace the old node-like `EventEmitter`. I've converted all internal events over to the new `EventEmitter2`, the only usage of `EventEmitter` remaining is within `Terminal` to refire the events so we don't break compatibility until v4.

Some of the events have changes a little or been removed, see below:

- blur -> Removed as you can just listen to the DOM event on Terminal.textarea
- focus -> Removed as you can just listen to the DOM event on Terminal.textarea
- linefeed -> `onLineFeed`
- selection -> `onSelectionChange`
- data -> `onData`
- key -> `onKey`
- keypress -> Removed in favor of a single onKey
- keydown -> Removed in favor of a single onKey
- refresh -> `onRender`
- resize -> `onResize`
- scroll -> `onScroll`
- title -> `onTitleChange`

New events being added:

- `onCursorMove`, this was already being used by the search addon and also within VS Code

If anyone sees events they rely upon removed please let me know how you're using it as `on` is now deprecated and will be removed soon.

Fixes #1505


vincentwoo:
If `.off` is deprecated, is there a new way to remove event listeners now?

Tyriar:
@vincentwoo onData, etc. return an IDisposable. Hold on to that and call `.dispose()` on it to remove the listener.

vincentwoo:
oh ho ho ho, fancy

Tyriar:
😎

Issue #1896, opened 2019-01-14
lets make the DOM renderer great again


jerch:
The DOM renderer needs some love. It kinda fell behind and still misses some flags like INVISIBLE, BLINK and DIM. Furthermore we might want to investigate if we can speedup things by less `<span>`ning.

Tyriar:
None of them support blink, too much work for something that is rarely used and an accessibility nightmare.

jerch:
@Tyriar  Tbh Im not aware of single terminal program in the last 20ys that uses blinking for some important feature. I think Turbo Pascal under DOS had some blinking animation after compiling or when an error was found, cannot remember exactly. :smile: 
Still for the cursor blinking it quite common, it can be done with a CSS animation quite easy.

jerch:
Hmm maybe we should hide it behind some option? I see a reason to be able to completely disable it, blinking often annoys ppl.

Tyriar:
Custom CSI handler + https://github.com/xtermjs/xterm.js/issues/1852 would probably make it possible.

Tyriar:
Most of this is done now thanks for community contributions 🎉, let's close this off and create issues for any remaining lack of functionality. Initially the DOM renderer was just the fallback but it will likely be the default in the future and users will be able to opt in to using GPU acceleration with the webgl renderer, so it should be just as solid as the others 💪

PR #2014, opened 2019-04-14
feat: add underline support to DOM renderer

starpit:
part of #1896


starpit:
doh! thanks for the good eyes, i forgot to commit that part.

Tyriar:
Thanks 🙂

PR #2017, opened 2019-04-17
fix selection not drawn when size changes

### LINKED ISSUE #1963: Selection doesn't always redraw when resizing the window (canvas renderer)
Repro:

1. Make a selection
2. Resize the window

Sometimes the selection will show, sometimes not.

khoitd1997:
this fixes #1963. This fix asks the program to redraw every time resize happens, there maybe a more efficient fix.


Tyriar:
This change doesn't seem to work for me.

After thinking about this a bit maybe it would be better for when a resize happens to trigger a selection draw from `Terminal`, just like how it does when the selection changes?

Tyriar:
This doesn't appear to be ready yet so closing off for now

PR #2035, opened 2019-04-29
cell representation for joined cells

### LINKED ISSUE #2028: Ligatures (possibly character joiners in general) broken in 3.13.x
Since 3.13.0-beta1, `xterm-addon-ligatures` produces weird spacing instead of actual joined chars - often in a different location too:

image

vs

image

`xterm-addon-ligatures`'s joiner implementation still returns correct indexes, so there must be something going on in xterm itself.

jerch:
Shall fix #2028.

@Eugeny I have no way to test it reliably with this repo and the demo. Could you try it out with your terminus build?


src/renderer/CharacterJoinerRegistry.ts
```diff
@@ -1,8 +1,51 @@
...
-import { CellData } from '../BufferLine';
+import { CellData, Content } from '../BufferLine';
 import { WHITESPACE_CELL_CHAR } from '../Buffer';
 
+export class JoinedCellData extends CellData implements ICellData {
+  private _width: number = 0;
+  private _code: number = 0x1FFFFF;  // highest allowed codepoint, meant as -1
+
+  public content: number = 0;
+  public fg: number = 0;
```
Tyriar: Can avoid setting these 3 and _width as they're done in the ctor

src/renderer/CharacterJoinerRegistry.ts
```diff
@@ -1,8 +1,51 @@
-import { ITerminal, IBufferLine } from '../Types';
+import { ITerminal, IBufferLine, ICellData, CharData } from '../Types';
 import { ICharacterJoinerRegistry, ICharacterJoiner } from './Types';
-import { CellData } from '../BufferLine';
+import { CellData, Content } from '../BufferLine';
 import { WHITESPACE_CELL_CHAR } from '../Buffer';
 
+export class JoinedCellData extends CellData implements ICellData {
+  private _width: number = 0;
+  private _code: number = 0x1FFFFF;  // highest allowed codepoint, meant as -1
```
Tyriar: Let's move into getCode

src/renderer/CharacterJoinerRegistry.ts
```diff
@@ -1,8 +1,51 @@
-import { ITerminal, IBufferLine } from '../Types';
+import { ITerminal, IBufferLine, ICellData, CharData } from '../Types';
 import { ICharacterJoinerRegistry, ICharacterJoiner } from './Types';
-import { CellData } from '../BufferLine';
+import { CellData, Content } from '../BufferLine';
 import { WHITESPACE_CELL_CHAR } from '../Buffer';
 
+export class JoinedCellData extends CellData implements ICellData {
```
Tyriar: Should this be extending? It looks like it implements basically everything already?

src/renderer/BaseRenderLayer.ts
```diff
@@ -260,8 +261,8 @@ export abstract class BaseRenderLayer implements IRenderLayer {
    */
   protected drawChars(terminal: ITerminal, cell: ICellData, x: number, y: number): void {
 
-    // skip cache right away if we draw in RGB
-    if (cell.isFgRGB() || cell.isBgRGB()) {
+    // skip cache right away if we draw in RGB or have joined cells
+    if (cell.isFgRGB() || cell.isBgRGB() || cell instanceof JoinedCellData) {
```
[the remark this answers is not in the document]
jerch: Correct. Gonna fix it.

src/renderer/CharacterJoinerRegistry.ts
jerch: True. To be fixed.

src/renderer/CharacterJoinerRegistry.ts
jerch: Like above.

src/renderer/CharacterJoinerRegistry.ts
jerch: Right, it is a needless safety thing.

Eugeny:
@jerch works nicely within Terminus

src/renderer/BaseRenderLayer.ts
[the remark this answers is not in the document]
jerch: Removed this, it will be handled anyway by the uncached method due to the high `charCode`.

Note for the performance book: the CellData <---> JoinedCellData overload might lead to heavy deopts here once a JoinedCellData appears. If this turns out to be the case we might have to refactor this again.

src/renderer/CharacterJoinerRegistry.ts
jerch: Changed this to `extends AttributeData` (needed for the color/flag primitives).

jerch:
@Eugeny Could you test this once more before we add it?

Eugeny:
Can't try this right now - upgraded to Node 12 (for Electron 5) and xterm's build system uses Gulp 3 which is long unsupported and is now broken

Eugeny:
Nevermind - works! Here's a quick and dirty patch to get the build working (sourcemaps are still broken tho) https://gist.github.com/Eugeny/89d3b7b9fd3625161b8c78a416a12560

jerch:
@Eugeny Thx for testing :heart:

PR #2037, opened 2019-05-03
feat: terminal.write returns a Promise<void>

### LINKED ISSUE #1650: writeSync method
I suggest a `.writeSync` method for `Terminal` and to document `.write` as asynchronous. Will help while writing tests and ppl to do sync stuff for small input.

starpit:
in order to allow clients to observe completion of writes, the `refresh` event is insufficient; that event allows clients to know when pending animation frames are done, but there is still a race window due to the setTimeout in Terminal.write. this PR allows clients to await on that setTimeout; combining an await on Terminal.write with listens on `refresh`, a client can observe full completion of pending writes

Fixes #1650

### DIFF
--- a/src/Terminal.test.ts
+++ b/src/Terminal.test.ts
@@ -34,6 +34,7 @@ describe('xterm.js', () => {
     term.write = (data) => {
       term.writeBuffer.push(data);
       (<any>term)._innerWrite();
+      return Promise.resolve();
     };
     (<any>term).element = {
       classList: {
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -184,6 +184,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
   // user input states
   public writeBuffer: string[];
   private _writeInProgress: boolean;
+  private _currentWritePromise: Promise<void>;
 
   /**
    * Whether _xterm.js_ sent XOFF in order to catch up with the pty process.
@@ -281,7 +282,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
     this._customKeyEventHandler = null;
     removeTerminalFromCache(this);
     this.handler = () => {};
-    this.write = () => {};
+    this.write = () => Promise.resolve();
     if (this.element && this.element.parentNode) {
       this.element.parentNode.removeChild(this.element);
     }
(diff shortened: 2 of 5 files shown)


Tyriar:
What's your use case and why isn't the refresh event sufficient? The reason I'm resistant to this is it adds a bunch of overhead by creating potentially thousands of promises a second that will typically go unused.

jerch:
@starpit Im not a fan of the promise idea either, care to explain your use case?

Note that my intention for #1650 comes from the wish to make synchronous actions possible for test cases, where you want to have a well define (and testable) state after some actions. but thats already well enough covered by the workaround described  there.

Typically promises are the first choice in JS to get back some strict synchronous execution. But as @Tyriar pointed out we cannot easily add them to the write method due to performance implications. Furthermore I am not sure that your promise is as useful as it should be for several reasons:
- The whole nature of the data flow is intended for streamline usage, it will apply whatever data it finds (if there is still enough time to do so, otherwise it gets buffered and will span several render cycles), this also applies to parts or even single bytes of the whole message. Thus its hard to explicitly mark some position in this data as "point of interest" -  you simply dont know for sure when to peek into. Worst case - the buffer gets drained two times - and you have to deal with 2 promises. Which one to go with?
- the terminal write endpoint is a sparce resource, any write from elsewhere will account into the buffering. If the terminal is connected to a backend pty and you "inject" some data on purpose the promise might cover to much data (backend data might get interleaved with your injected data).

There are some ways to overcome this "non deterministic aura" (from a point of terminal writer), still not convinced that we need this in a public general fashion.

starpit:
hey y'all! yes, i will close this one now. i wanted to let our code settle for a bit. 3.13's `refresh` flows seem to suit our needs for now!

at some point, it might be worth considering a lightweight use of Promises versus setTimeout --- no need to return them as does this PR, rather only employ them for asynchronous dispatch. my (very facile, at the moment) understanding is that the general trend in NodeJS is to optimize for the former; e.g. with highly optimized and lightweight "microtask" scheduling.

Tyriar:
@starpit we supported IE until very recently which is why everything is callback based, but yes I agree in general with moving to promises internally (for most things?).

Before we go and support something like this in the API though I'd definitely like to hear a valid use case where the `refresh` event (onRender in the next version) isn't sufficient

PR #2047, opened 2019-05-09
Base the next beta version on the correct current beta

### LINKED ISSUE #2044: CD doesn't work once there are 10 beta builds
localCompare is bad, need to grab the number and sort with that.

https://github.com/xtermjs/xterm.js/blob/d6da010c7e583f60394fef565331ef9df9fd8450/bin/publish.js#L47

Tyriar:
Fixes #2044


Tyriar:
Merging this in now to get the release ready, the next beta build should publish with it.

Issue #2005, opened 2019-04-08
Renderer API


Tyriar:
An API for renderers is necessary if we want to include the WebGL renderer as an optional component. A lot of this issue is about refactoring the renderer code to share a lot of the core functionality of all the renderers to make the actual `IRenderer` API interface slimmer for addons to eventually implement. Here are the sub-goals:

- [x] Move `ColorManager` to live under `Terminal`, pass `IColorSet` to the renderer
- [x] Implement the renderer "orchestrator" that does the debouncing/pausing etc. that the current renderers do.
- [ ] Look into whether `CharMeasure` can live only in the DOM renderer code (canvas renderers can use `onOptionsChange`)
- [ ] ~~Break renderer dependence on Terminal by passing in relevant objects to handlers~~ Renderer addons will have access to the xterm.js API
- [ ] Expose the renderer API as experimental

### API proposal

*Note that the inline proposal will change as progress is made.*

```ts
class Terminal {
  /**
   * Gets or sets the renderer to use for the `Terminal`, the renderer is
   * responsible for displaying all terminal data to the user, including the
   * mouse selection.
   */ 
  renderer: IRenderer;
}

/**
 * An interface that implements a renderer capable of rendering terminal data.
 * Note that if the browser supports `IntersectionObserver` it will not call to
 * render frames when the terminal is completely hidden. The best way to
 * implement this interface is to keep a 
 */
interface IRenderer extends IDisposable {
  /**
   * The dimensions that the renderer will use for things like canvas, cell and
   * character size.
   */
  readonly dimensions: IRenderDimensions;

  /**
   * A method that is called when `window.devicePixelRatio` changes.
   */
  onDevicePixelRatioChange(): void;

  /**
   * A method that is called when the dimenisons of the termnial change.
   */
  onResize(cols: number, rows: number): void;

  /**
   * A method that is called when the terminal loses focus.
   */
  onBlur(): void;

  /**
   * A method that is called when the terminal gains focus.
   */
  onFocus(): void;

  /**
   * A method that is called when the terminal cursor moves, this may be useful
   * for managing the state of a blinking cursor.
   */
  onCursorMove(): void;
  
  /**
   * A method that is called when terminal options change. Note that changes to
   * the theme should be handled with `IRenderer.onThemeChange`.
   */
  onOptionsChange(): void;

  /**
   * A method that is called when the theme changes.
   */
  onThemeChange(colors: IColorSet): void;
  
  /**
   * Clears the viewport.
[...]

PR #2065, opened 2019-05-11
New addon API

### LINKED ISSUE #1128: Custom addons

Tyriar:
I added back the old addons so we can ship the new API in v3, I think we should move ahead with this so that we can start converting the other addons over to the new API incrementally.

Fixes #1128
Related #2076

## The API

The final API ended up being extremely simple:

```ts
export class Terminal {
    /**
     * (EXPERIMENTAL) Loads an addon into this instance of xterm.js.
     * @param addon The addon to load.
     */
    loadAddon(addon: ITerminalAddon): void;
}

export interface ITerminalAddon extends IDisposable {
    /**
     * (EXPERIMENTAL) This is called when the addon is activated within xterm.js.
     */
    activate(terminal: Terminal): void;
}
```

In fact the only interesting part about the new addon API is how the addon's `dispose` method gets wrapped once loaded so that it can unload itself from the Terminal when called.

## Terminal events

The idea of exposing a more rich API with events that get fired on the addon itself came up, but I'm leaning towards having addons handle that themselves, at least for now. For example if the addon only works after open it could check `Terminal.element`:

```ts
const addon = {
  activate: (term: Terminal) => {
    if (term.element) {
      // it's already open, init now
    } else {
      // otherwise handle it when it happens
      term.onOpen(() => ...);
    }
  },
  dispose: () => { }
}
```

## Example usage

The usage from the demo is close to this:

```ts
term.loadAddon(new WebLinksAddon());
attachAddon = new AttachAddon();
term.loadAddon(attachAddon);
```

Notice we keep a reference to `attachAddon` as we might want to dispose it later (detach).

## Addon implementations

- https://github.com/xtermjs/xterm-addon-attach
- https://github.com/xtermjs/xterm-addon-web-links


src/ui/AddonManager.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export class AddonManager implements IDisposable {
+  protected _addons: ILoadedAddon[] = [];
+
+  constructor() {
+  }
+
+  public dispose(): void {
+    for (let i = this._addons.length - 1; i >= 0; i--) {
```
jerch: Reversed disposing - should do the trick for most linear dependencies  :+1: 
(Guess we will never see cycling deps, they are a nightmare on its own)

src/ui/AddonManager.ts
```diff
@@ -0,0 +1,57 @@
...
+      instance,
+      dispose: instance.dispose,
+      isDisposed: false
+    };
+    this._addons.push(loadedAddon);
+    instance.dispose = () => this._wrappedAddonDispose(loadedAddon);
+    instance.activate(<any>terminal);
+  }
+
+  private _wrappedAddonDispose(loadedAddon: ILoadedAddon): void {
```
jerch: Guess Ive overlooked it - where are the addon disposals hooked into the terminal dispose chain?
Also does the API allow to do nasty things like `terminal.dispose()` within `addon.dispose`?

Tyriar:
The fullscreen addon will be removed when we transition over as it does so little, for fit it can be converted after the renderer API is done in https://github.com/xtermjs/xterm.js/issues/2005

src/ui/AddonManager.ts
Tyriar: Yeah this is something we can look into supporting as it comes up, not supported for now

src/ui/AddonManager.ts
Tyriar: > Guess Ive overlooked it - where are the addon disposals hooked into the terminal dispose chain?

It just moved to dispose in `./public/Terminal` as the addon manager now lives there which is much nicer (as they build on the API, not `TerminalCore`).

> Also does the API allow to do nasty things like terminal.dispose() within addon.dispose?

Well you could as `Terminal.dispose` is part of the API, no way around that really nor should we try imo. I see a possible infinite loop because of that, I think just moving isDisposed = true above the addon.dispose call should fix that.

PR #2067, opened 2019-05-11
fix: don't use Math.floor in DomRenderer charWidth computation

### LINKED ISSUE #2063: DomRenderer.dimensions.scaledCharWidth is floored
## Details
- Browser and browser version: any
- OS version: any
- xterm.js version: 3.13

### Steps to reproduce

1. have a client emit some text to the dom (not using xterm)
2. emit the same text to the dom (using xterm)

Expected result: the width of the two elements should be equal
Actual result: the width of the xterm element is always <= the normal dom element

Cause: in DomRenderer.ts, the `scaledCharWidth` attribute is taken as the floor of the measured character width, as follows:

```
this.dimensions.scaledCharWidth = Math.floor(this._terminal.charMeasure.width * window.devicePixelRatio);
```

For example, on macOS with `fontFamily:"SF Mono"` and `fontSize:14px`, the measured character width is 8.66px (rounded to two decimal places); xterm.js instead uses a character width of 8px, due to the `floor` calculation.

As a consequence, xtermjs elements render as width-compressed compared to the natural character spacing of the chosen font.

I assume that there was a reason for the `floor`? I am reporting it here, as it causes poor font rendering for our use case.

starpit:
Fixes #2063

### DIFF
--- a/src/renderer/dom/DomRenderer.ts
+++ b/src/renderer/dom/DomRenderer.ts
@@ -101,7 +101,7 @@ export class DomRenderer extends Disposable implements IRenderer {
   }
 
   private _updateDimensions(): void {
-    this.dimensions.scaledCharWidth = Math.floor(this._terminal.charMeasure.width * window.devicePixelRatio);
+    this.dimensions.scaledCharWidth = this._terminal.charMeasure.width * window.devicePixelRatio;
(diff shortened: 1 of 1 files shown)


Tyriar:
I believe this caused https://github.com/microsoft/vscode/issues/74183 so I'm reverting it. I can't actually reproduce the problem but I've got a flood of reports over the past 2 days since updating xterm in vscode.

PR #2071, opened 2019-05-11
Adding TypeScript formatting using prettier along with TSLint

luxcium:
adding TypeScript formatting using prettier
tslint-config-prettier disables conflicting rules


luxcium:
I am here to work on the selection background/foreground issue in fact but I like to work with prettier so I hope my changes will be in line with the project philosophy so it can be merged

luxcium:
Ok I do understand, first the editor.formatOnSave/Paste might have been a little far fetch, and I am sorry, but the formatting introduced by prettier let you do that stuff on autopilot (I am never indenting, adding semicolons, and better I don't have to care about trailing spaces). I knew it had good chances to be rejected but I did it anyway maybe it's like a cleaning lady reorganizing the book on your shelf or doing stuff she was not asked to do so I am sorry. But more than that  Prettier is maybe also as a religion on its own and is not something one should impose on others I should have to think better but I will try to learn something from this ...

PR #2074, opened 2019-05-11
Buffer API

### LINKED ISSUE #1994: Buffer API

Tyriar:
This exposes the primary parts of the buffer to APIs, see changes to xterm.d.ts for the actual public API. It works by having a "view" object in public/ that owns the actual object and exposes friendly aliases for the properties, for example `viewportY` instead of `ydisp`, `cursorX` over `x`, `getLine` over `lines.get(x)`, etc.

The first usage of this will be moving the search addon into the new addon API https://github.com/xtermjs/xterm.js/pull/2065, built 100% on stable API! The buffer API is currently marked experimental to make sure it meets the needs of the search addon, we can remove experimental at or before v4 https://github.com/xtermjs/xterm.js/issues/2075

Fixes #1994


src/public/Terminal.ts
```diff
@@ -169,3 +171,30 @@ export class Terminal implements ITerminalApi {
...
+
+class BufferApiView implements IBufferApi {
+  constructor(private _buffer: IBuffer) {}
+
+  public get cursorY(): number { return this._buffer.y; }
+  public get cursorX(): number { return this._buffer.x; }
+  public get viewportY(): number { return this._buffer.ydisp; }
+  public get baseY(): number { return this._buffer.ybase; }
+  public get length(): number { return this._buffer.lines.length; }
+  public getLine(y: number): IBufferLineApi { return new BufferLineApiView(this._buffer.lines.get(y)); }
```
jerch: Does this need a sanity check to avoid a lazy error due to a non existing line?

src/public/Terminal.ts
```diff
@@ -169,3 +171,30 @@ export class Terminal implements ITerminalApi {
...
+}
+
+class BufferLineApiView implements IBufferLineApi {
+  constructor(private _line: IBufferLine) {}
+
+  public get isWrapped(): boolean { return this._line.isWrapped; }
+  public getCell(x: number): IBufferCellApi { return new BufferCellApiView(this._line, x); }
```
jerch: Same here - how about x overflowing the actual length?

jerch:
@Tyriar
The way you currently create the BufferView by referencing is not "stable" for buffer lines between data updates of the terminal due to recycling and line movements. Not sure yet if thats a drawback/issue at all, still it means that every action from an addon will have to re-request the view to see the real terminal buffer state with the correct buffer line order.

I think this is not an issue if we clearly document that ppl should avoid holding refs to members on nonlocal scope (rule of thumb - no event loop in between), instead always refer to them by the full descent like `terminal.buffer().getLine(y).getCell(x)....`. If they really have to hold stuff, they will have to copy the parts over to some own container. Local refs are not affected by this.

Tyriar:
@jerch I think this is fine if it's called out in the documentation as you mention, I'll add some more details to the jsdoc

src/public/Terminal.ts
Tyriar: This doesn't but it now correctly returns `IBufferLineApi | undefined`

PR #2080, opened 2019-05-12
apply attach changes to demo

jerch:
Apply attach addon changes to the demo.


Tyriar:
Merged the attach addon and I updated the version here after testing to ensure it worked with both 👍

PR #2083, opened 2019-05-14
READY option in attach addon v3 and demo

jerch:
Backport of the READY message of the attach addon from here https://github.com/xtermjs/xterm-addon-attach/pull/10.

Changes:
- `sendReady` flag for `attach` method
- respect READY message in demo

Note: Since the demo already switch to v4 addons, https://github.com/xtermjs/xterm-addon-attach/pull/10 should be applied first.


jerch:
Not an issue anymore, see https://github.com/xtermjs/xterm-addon-attach/pull/10 for further details.

PR #2085, opened 2019-05-15
Add sanity check in AccessibilityManager

### LINKED ISSUE #2082: Using accessiblity mode a property 'textContent' can be undefined
Using accessibility mode can sometime end up with trace as:
```
TypeError: Cannot set property 'textContent' of undefined
at AccessibilityManager../node_modules/xterm/lib/AccessibilityManager.js.AccessibilityManager._renderRows (...)
at RenderDebouncer../node_modules/xterm/lib/ui/RenderDebouncer.js.RenderDebouncer._innerRefresh (...)
```

This is new regression seen only in version `3.13.0`, version `3.12.2` was not affected.
In general this is hard to reproduce, it happens in our CI, or I can only reproduce it with very high load on my machine and running tests in loop.

Possible place of happening https://github.com/xtermjs/xterm.js/blob/master/src/AccessibilityManager.ts#L252

I know this is no reproducer, possibly just conditionally check that these elements are present? Seems related to https://github.com/xtermjs/xterm.js/commit/4c3082c924ac or more general to https://github.com/xtermjs/xterm.js/pull/2026

Tyriar:
Fixes #2082

---

I couldn't reproduce or see why this could happen, so let's just add this sanity check.

### DIFF
--- a/src/AccessibilityManager.ts
+++ b/src/AccessibilityManager.ts
@@ -249,9 +249,11 @@ export class AccessibilityManager extends Disposable {
       const lineData = buffer.translateBufferLineToString(buffer.ydisp + i, true);
       const posInSet = (buffer.ydisp + i + 1).toString();
       const element = this._rowElements[i];
-      element.textContent = lineData.length === 0 ? Strings.blankLine : lineData;
-      element.setAttribute('aria-posinset', posInSet);
-      element.setAttribute('aria-setsize', setSize);
+      if (element) {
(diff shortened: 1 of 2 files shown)


Tyriar:
@jerch it's not just t0001 that fails, it's only the first test as I tried disabling it and the second test failed.

Tyriar:
Maybe the container disk access sucks, moving it out passed (for the first time)

PR #2088, opened 2019-05-16
Fix what branch coverage is sent with

Tyriar:
This should fix the coveralls badge as it appears to be working now, just it's reporting against `refs/head/master` instead of `master`


Tyriar:
Merging, hopefully it'll work for master :)

Tyriar:
It worked, and we went down 6% since it last worked 😉

PR #2094, opened 2019-05-19
Move ColorManager to ui

Tyriar:
Part of #2005
Part of #1507


Tyriar:
Merging to unblock update in https://github.com/xtermjs/xterm.js/pull/1790

PR #2113, opened 2019-05-21
Run release against builds on release/* branches

### LINKED ISSUE #2111: Streamline patch releases
Right now patch releases are manual, it would be good if the Release pipelines job also supported branches `release/*`

Tyriar:
Fixes #2111

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -77,7 +77,7 @@ jobs:
   - Linux
   - macOS
   - Windows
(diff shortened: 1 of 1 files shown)


Tyriar:
This will be verified the next time we need a patch release 😛, not 100% sure if * is the right thing here

Tyriar:
Updated https://github.com/xtermjs/xterm.js/wiki/Releasing#patch-releases

PR #2120, opened 2019-05-22
trim processed chunks from write buffer

### LINKED ISSUE #2108: Input memory exhaustion
Forked off from https://github.com/xtermjs/xterm.js/issues/2077

> The time-based input limiting contains a buffer, that gets drained only once all chunks were handled. 
This way fast producers can easily exhaust memory and will crash the terminal / browser engine.
>
> Partial fix: Trim/slice already handled chunks from the buffer.

https://github.com/xtermjs/xterm.js/blob/af81acc50e336d3fd39b10fa4322f6949a5ba926/src/Terminal.ts#L1441-L1447

jerch:
Partial fix for a possible OOM in `Terminal.write`.

Fixes #2108.

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -72,6 +72,7 @@ const WRITE_BUFFER_PAUSE_THRESHOLD = 5;
  * depends on the time it takes for the renderer to draw the frame.
  */
 const WRITE_TIMEOUT_MS = 12;
+const WRITE_BUFFER_LENGTH_THRESHOLD = 50;
 
 const MINIMUM_COLS = 2; // Less than 2 can mess with wide chars
 const MINIMUM_ROWS = 1;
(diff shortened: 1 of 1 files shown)


mofux:
I'm not too much into this topic, but for my understanding: does this mean that we potentially trash data and never have it processed by our parser?

jerch:
@mofux: This fix just removes already handled chunks from the writebuffer, so no data will be lost (yet). Thats the reason why it can only partially fix the issue, a really fast producer will still run the terminal into OOM without a flow control back pressure mechanism. This is left for an upcoming PR and will prolly have a discard option to stop malicious backends from flooding the terminal.

jerch:
Some numbers from the demo for `ls -lR /usr`:

image

and for `yes`:

image

Pending shows the unhandled amount of data and chunks, more than obvious that this patch cannot achieve much. Even `ls -lR` stores up chunks, for `yes` its disastrous.

Difference to before - no trimming at all during busy input, so just add all trimmed data onto the pending side.

src/Terminal.ts
```diff
@@ -72,6 +72,7 @@ const WRITE_BUFFER_PAUSE_THRESHOLD = 5;
  * depends on the time it takes for the renderer to draw the frame.
  */
 const WRITE_TIMEOUT_MS = 12;
+const WRITE_BUFFER_LENGTH_THRESHOLD = 50;
```
Tyriar: Should this be larger than 50? It's ok to use a bit of RAM, we just don't want to use it all?

jerch: Well I dont see any difference with 20 or 100 either (beside the memory usage), so what number do you suggest?

Tyriar: If you don't see any difference this is fine 🙂

Issue #1918, opened 2019-01-31
Add a flow control mechanism to the attach addon


jerch:
The old flowcontrol approach with XOFF/XON does not work in envs that disable flow control on the pty. Since many shell startup scripts disable it, we can remove it from `Terminal.write`.

For local apps like electron based flow control can be achieved by avoiding async buffers in between, for remote apps we might want to add some flowcontrol mechanism to the attach addon.

PerBothner:
DomTerm keeps a running count of the number of bytes received and the number of bytes confirmed, which are tracked both in the browser and the (WebSockets) server. (Both of these counts are modulo 0x10000000 to deal with wrap-around.)  When the difference between the number of received and confirmed bytes is larger than some threshhold (currently 500), then the browser sends a `RECEIVED` message with the current received count, and updates the confirmed count to match.  This is done _after_ processing the received bytes.

Similarly, the server keeps track of the number of sent and confirmed bytes.  (The confirmed count is the most recent value of a `RECEIVED` message.) When the difference between these values gets above a certain threshold (currently 8000 bytes) then the server will disable polling of the file descriptor for the output from the pty.  The user process will pause when the kernel tty driver buffers are full.  When the browser has caught up, it will send a new `RECEIVED` message, and polling and reading of the pty will resume.

DomTerm uses a ["report-event" escape sequence](http://domterm.org/Wire-byte-protocol.html#Special-sequences-sent-by-DomTerm-to-back-end) for messages sent from the browser to the server. The byte-stream from the browser to the server contains primarily user key-strokes (encoded as in xterm), as well as these "event" messages.  A domterm server needs to scan the stream for event messages and handle them appropriately.

The xterm.js project concentrates on the browser part of the terminal, leaving the server part to embedders.  I suggest adding an escape sequence that the server can send to notify xterm.js that the server understands these event messages, including the `RECEIVED` message.

jerch:
Yeah we dont want to focus more on the server side as this is left to integrators (could be build in any language/framework/whatsoever). Thus we might need some general way to indicate the server part "hey - slow down abit".
An easy approach would be a high/low watermark buffer observer, that would just signal a "stop" at the high mark and a "go" at the low mark. Since we do not control the server part it would be free to ignore these signals and just keep on sending data. Not sure if we need to go that far and have to implement lossy data handling to deal with "malicious" servers (this certainly would save the xterm.js process from crashing).

[...]

PR #2122, opened 2019-05-24
custom flow control and discard limit

### LINKED ISSUE #2077: Flow control/back pressure
The time-based input limiting contains a buffer, that gets drained only once all chunks were handled. 
This way fast producers can easily exhaust memory and will crash the terminal / browser engine.

Relevant code: https://github.com/xtermjs/xterm.js/blob/af81acc50e336d3fd39b10fa4322f6949a5ba926/src/Terminal.ts#L1441-L1447

Partial fix: Trim/slice already handled chunks from the buffer.

Even with this partial fix a long running producer that is faster than the terminal can handle incoming chunks will grow the buffer up to a possible crash. For a full fix we have to implement a reliable back pressure mechanism.

jerch:
This PR introduces a new flow control mechanism based on recent changes in `node-pty`. The flow control works similar to the current XON/XOFF way with writing special messages to the backend to indicate whether data streaming should be paused/resumed. Main differences are customizable PAUSE/RESUME messages. Those will be filtered in `node-pty` to pause/resume the pty slave program by buffer back pressure.

Changes:
- track write buffer state with a watermark
- sanity check for fast incoming data with discard option (fixing the possible OOM)
- high and low watermarks to pause/resume data flow

TODO:
- determine default strings for PAUSE/RESUME (should work for all common platforms/envs)
- API options, which should go there?
- documentation / example usage / special notes for non `node-pty` backends
- tests

Fixes #2077, #1918.


vincentwoo:
Very interesting stuff!

src/Terminal.ts
```diff
@@ -59,11 +59,29 @@ import { RenderCoordinator } from './renderer/RenderCoordinator';
...
+ */
+const LOW_WATERMARK = 100000;
+const HIGH_WATERMARK = 300000;
+
+/**
+ * Flow control PAUSE/RESUME messages.
+ *
+ * TODO: make this configurable
+ */
+const FLOW_CONTROL_PAUSE  = '\x1b^p\x1b\\'; // PM p ST
```
Tyriar: How about we do just use `PM pause ST` and `PM resume ST` and not make it configurable (until requested)?

src/Terminal.ts
```diff
@@ -59,11 +59,29 @@ import { RenderCoordinator } from './renderer/RenderCoordinator';
...
  */
-const WRITE_BUFFER_PAUSE_THRESHOLD = 5;
+const DISCARD_WATERMARK = 10000000; // FIXME: should this be bigger?
+
+/**
+ * Flow control watermarks for the write buffer.
+ * low: send resume to pty
+ * high: send pause to pty
+ *
+ * TODO: make this configurable
```
Tyriar: Not sure we need to make this configurable either?

src/Terminal.ts
```diff
@@ -1371,18 +1390,23 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
       return;
     }
 
-    this.writeBufferUtf8.push(data);
+    // safety measure: dont allow the backend to crash
+    // the terminal by writing to much data to fast.
+    if (this._watermark > DISCARD_WATERMARK) {
```
Tyriar: I don't think we need this check here, jsdoc in API (and more docs when we improve that) should be enough

src/Terminal.ts
```diff
@@ -187,6 +205,7 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
   public writeBuffer: string[];
   public writeBufferUtf8: Uint8Array[];
   private _writeInProgress: boolean;
+  private _watermark: number = 0;
```
Tyriar: `_writeBufferLength`?

src/Terminal.ts
jerch: Sounds good to me. If we document it properly ppl will not miss it (hopefully).

src/Terminal.ts
jerch: Dont think so either. But we would need some good defaults here (prolly lower ones than mine).

src/Terminal.ts
jerch: You mean like no DISCARD check at all? It is meant to stop malicious or faulty backends (thise that ignore the PAUSE) from crashing ppls browsers. Sure we dont want that?

src/Terminal.ts
jerch: Ah this would not quite reflect what it is - it sums up the length of pending chunks in both write buffers (string and utf8). Maybe `_writeBuffersPendingSize`?

jerch:
Did some fine tuning of the limits in the demo with the last commits (still contains debug logs).

There is a tradeoff between raw throughput speed and response to interrupts like Ctrl-C. Furthermore fast producers behave differently than slower ones due to different buffer fill states.

The numbers I have found are the compromise between no/low negative impact on frequently used semi fast producers like `ls` and still acceptable response time to Ctrl-C with fast ones (`yes` and [./fast_producer](https://github.com/xtermjs/xterm.js/blob/ca3e47da975165f1134a35c5c41cf6da0b2b80b0/fast_producer.c))

Currently I see the following:
- `ls` string throughput slightly decreased (~ -7%)
- `ls` utf8 throughput slightly increased (~ +10%)
- Ctrl-C on `ls` is immediately
- Ctrl-C on `yes` shows a tiny bit of latency
- Ctrl-C on `./fast_producer` shows a bigger but still acceptable latency

It get it there I aslo had to introduce a buffer cap in the server.js pty --> xterm buffer, otherwise chunks from running `./fast_producer` get so big within 5ms that the terminal shows a lag.

@Tyriar Could you test if these numbers will do on MacOS? With its bigger pty buffer `yes` and `./fast_producer` might show a big latency again for Ctrl-C.

src/Terminal.ts
Tyriar: This could discard some important things which would corrupt the terminal, for example `ls -lR / && vim` may not enter the alt buffer. I guess the main reason I don't think it's needed is I haven't seen the demo crash for a long time and we could avoid this critical path check if people implement correctly.

Tyriar:
Ctrl+C on `yes`/macOS I see the server pause immediately (logs stop) and then about 2 seconds before the renderer catches up.

src/Terminal.ts
jerch: What I see with `./fast_producer` in the demo on master:
- FF: +2.5 GB / min, kernel eventually kills it for being this memory hungry (happens for me around minute 3 to 5 in several runs)
- Chrome: +1 GB / min in the beginning, after 2-3 min this drops back to ~250 MB and is stable there, devtools die after 30s

No clue what going on with Chrome, I think they are cheating once a process hit some memory limit (killing the Websocket? - idk, did not investigate further). Firefox shows the expected linear memory exhaustion.

I am abit uneasy about not having this security setting, since it might kill the browser (under FF at least the kernel killed the whole browser with all tabs). Maybe we should set `DISCARD_WATERMARK` higher even higher to avoid losing data in your example?

jerch:
@Tyriar Hmm 2s is quite long, but I cannot lower that much further without sacraficing throughput performance. Can we live with it taking this long? Imho `yes` is already an extrem example of a fast producer, things that typically have longer run times with quite noisy output are buildtools. Those should work pretty well with these settings.

Tyriar:
@jerch I wasn't actually testing it right, it works great when I toggle `useFlowControl` on after 
34c3a77

Tyriar:
I don't think this is a WIP anymore right?

demo/server.js
```diff
@@ -45,15 +45,18 @@ function startServer() {
...
-          encoding: USE_BINARY_UTF8 ? null : 'utf8'
+          encoding: USE_BINARY_UTF8 ? null : 'utf8',
+          handleFlowControl: true,
+          flowControlPause: '\x1b^pause\x1b\\',
+          flowControlResume: '\x1b^resume\x1b\\'
         });
 
     console.log('Created terminal with PID: ' + term.pid);
     terminals[term.pid] = term;
     logs[term.pid] = '';
```
Tyriar: Should this be commented out too?

demo/server.js
```diff
@@ -107,10 +121,11 @@ function startServer() {
         }
       };
     }
-    const send = USE_BINARY_UTF8 ? bufferUtf8(ws, 5) : buffer(ws, 5);
+    const send = USE_BINARY_UTF8 ? bufferUtf8(ws, 5, 16384) : buffer(ws, 5, 16384);
```
Tyriar:
```ts
const MAX_SEND_INTERVAL = 5;
const MAX_CHUNK_SIZE = 16384;
const send = (USE_BINARY_UTF8 ? bufferUtf8 : buffer)(ws, MAX_SEND_INTERVAL, MAX_CHUNK_SIZE);
```

demo/server.js
```diff
@@ -107,10 +121,11 @@ function startServer() {
...
 
     term.on('data', function(data) {
       try {
+        console.log(data.length);
```
Tyriar: Remove

src/Terminal.ts
```diff
@@ -1371,18 +1386,23 @@ export class Terminal extends EventEmitter implements ITerminal, IDisposable, II
       return;
     }
 
-    this.writeBufferUtf8.push(data);
+    // safety measure: dont allow the backend to crash
+    // the terminal by writing to much data to fast.
+    if (this._watermark > DISCARD_WATERMARK) {
+      // FIXME: do something more useful
```
Tyriar: `throw new Error`?

fast_producer.c
```diff
@@ -0,0 +1,8 @@
+#include <stdio.h>
+#include <unistd.h>
+
+int main(int argc, char **argv) {
+  while (1) {
+    putchar('#');
```
Tyriar: `yes` seems to work fine, and I see the pausing and catching up in the log of `yarn start`. fast_producer however really starts to chug things and it's hard to tell if the pausing is working at all as all the logs say 1024.

demo/server.js
jerch: Yeah, I think we can remove the whole `logs` stuff? Dont see where this is used.

demo/server.js
jerch: :+1:

src/Terminal.ts
jerch: Yeah throwing an error might be the only valid option, as the terminal stream got out of hand already and the terminal is prolly lost. I think we can set the discard limit to a really high number (50MB?) before this parachute opens?

fast_producer.c
jerch: Additionally I placed a `console.log(this._watermark)` in write to test the flowing with fast_producer.
I was expecting this to run less responsive on MacOS, MacOS has a dynamic pty buffer size depending on incoming data pressure (typically ranges from 16-64kB). Thus a single message will be like 64kB, alot to chew for xterm.js in one step.
fast_producer is an extreme example an IMHO not worth optimizing for. Since `yes` runs ok for you I think the numbers will do as they are?

src/Terminal.ts
jerch: Changed it to 50MB.

demo/server.js
Tyriar: Sure, I don't think I've ever used it 👍

jerch:
@Tyriar Added a better fast producer snippet. This is fastest I was able to find with some more structured output (almost double as fast as `yes`).

src/Terminal.ts
```diff
@@ -59,11 +59,34 @@ import { RenderCoordinator } from './renderer/RenderCoordinator';
...
+ * Since this limit is meant as a safety parachute to prevent browser crashs,
+ * it is set to a very high number. Typically xterm.js gets unresponsive with
+ * a much lower number (>500 kB).
  */
-const WRITE_BUFFER_PAUSE_THRESHOLD = 5;
+const DISCARD_WATERMARK = 50000000;  // ~50 MB
+/**
+ * send ACK every ACK_WATERMARK-th byte
+ */
+const ACK_WATERMARK = 131072;//524288; // 2^19
```
Tyriar: I think we may want to instead send some request ack sequence from the server and then once this is hit in the parser we can send back. Extra awesome super-duper benefit of this is that it could all be done via the xterm.js API via a custom handler 😮 

So server asks `'\x1b^reqack;1\x1b\\'` (not sure if an ID, the 1,  is needed or not yet), then our custom ESC handler sends back `'\x1b^ack;1\x1b\\'`.

jerch:
Closing this PR for now. We still have to work out the details for the offscreen core lib (some downgraded Terminal.ts thingy), prolly with some input service like thingy as well. Any attempts towards better flow control should go there once we have that.

PR #2128, opened 2019-05-25
Only update a11y announcements on animation frame

### LINKED ISSUE #1965: Empty lines in vim take too long to show
**This is definitely related to the fact we are using accessibility mode (screenreader support). If I turn off this mode it works just fine**

Opening vim in terminal results in a very long loading time. During this time browser (tested on Chrome and Firefox) becomes unresponsive. Afterwards everything works fine, unless the browser window gets resized which again results in unresponsiveness.

## Details
- Browser and browser version: Firefox 65
- OS version: Fedora 29
- xterm.js version: 3.10.1

### Steps to reproduce
Opening nonexistent or empty file is much slower than opening existing file. Opening empty file takes ~4s, opening file with 1 line takes ~3.8s, opening file with 2 lines takes ~3.5s, 10 lines takes ~2.3s, opening file with more than 20 lines is immediate (out of 40 visible lines). On smaller screen it works proportionally. If I fit on screen only 20 lines and open empty file it is immediate, if I fit 30 lines on screen and open empty file it takes around those 2 seconds.

It seems that every empty line in vim sends something that xterm.js parses for accessibility mode.

Tyriar:
Fixes #1965

### DIFF
--- a/src/AccessibilityManager.ts
+++ b/src/AccessibilityManager.ts
@@ -43,6 +43,8 @@ export class AccessibilityManager extends Disposable {
    */
   private _charsToConsume: string[] = [];
 
+  private _charsToAnnounce: string = '';
+
   constructor(
     private _terminal: ITerminal,
     private _dimensions: IRenderDimensions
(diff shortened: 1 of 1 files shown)


src/AccessibilityManager.ts
```diff
@@ -283,13 +286,11 @@ export class AccessibilityManager extends Disposable {
     element.style.height = `${this._dimensions.actualCellHeight}px`;
   }
 
-  private _announceCharacter(char: string): void {
-    if (char === ' ') {
-      // Always use nbsp for spaces in order to preserve the space between characters in
-      // voiceover's caption window
-      this._liveRegion.innerHTML += '&nbsp;';
```
Tyriar: I don't think this was actually that important, and if we wanted to keep it the setting would more complicated.

PR #2132, opened 2019-05-26
Enable absolute imports and minify distributable

### LINKED ISSUE #1254: Is there a need for a build/ and a dist/ folder?
Follow up from https://github.com/xtermjs/xterm.js/pull/1244#discussion_r164167635

We change the build folder to `dist/` in the publish script, why not just always use `dist/`?

### LINKED ISSUE #1314: Use tsconfig baseUrl and absolute imports

Tyriar:
Fixes #1314
Fixes #1254

---

This PR seems like a good place to start to move to v4 as it solves a bunch of problems and will let us remove the browserify build step as it works in `lib/` now. If you want to review alt+click the collapse button in GitHub to collapse all files and then ignore all files inside `src/` except for `tsconfig.json`.

It makes the following high-level changes:

- Enables absolute imports via `baseUrl` and `paths` in `tsconfig.json` for all paths referencing `common`, `core` and `ui` modules:
   ```ts
   // before
   import { DEFAULT_COLOR } from '../../common/Types';
   import { IColorSet } from '../../ui/Types';
   // after
   import { DEFAULT_COLOR } from 'common/Types';
   import { IColorSet } from 'ui/Types';
   ```
- Moved the tsc build output to `out/`, this is no longer suitable to directly import because the absolute paths need to be packaged.
- Webpacks `out/` as a production umd module in `lib/`, this is the new `main` file in `package.json
- The demo imports from `out/` by default but can import from `lib/` by just uncommenting a line to verify the packaged portion (it requires you to run `yarn package` though):
   ```ts
   // Use tsc version (yarn watch)
   import { Terminal } from '../out/public/Terminal';
   // Use webpacked version (yarn package)
   // import { Terminal } from '../lib/xterm';
   ```
- Removed gulp and lots of dependencies
- Replaced gulp tests with npm scripts, removed build gulp tasks completely
- Removed coverage reporting as it wasn't helping and can be misleading
- Moved xterm.css into `css/`, removed tasks for move it around

Follow up changes after this look good:

- [ ] Sort imports
- [ ] Remove `src/addons` and all build steps completely
- [ ] Update documentation:
  - [ ] dist/ -> lib/
  - [ ] New css path

Notes:

- While ts-loader is capable of doing the build in a single step, we lose great tsc features like strict null checking and composite projects. Splitting it up like it has always been also makes it much easier to understand imo.


Tyriar:
Need to figure out how to resolve the paths for the tests.

Tyriar:
And so it begins.

PR #2149, opened 2019-05-31
Update cells after an ECH

### LINKED ISSUE #2148: Cells are not marked for update after an ECH
This causes an issue _sometimes_ in conpty where cls would not redraw the screen. https://github.com/microsoft/vscode/issues/72708

Tyriar:
Fixes #2148

---

Need to test on Windows before I merge


Tyriar:
Verified by bumping node-pty locally and forcing conpty to be on

PR #2164, opened 2019-06-01
Move new format addons into repo, remove old addons

### LINKED ISSUE #2142: Remove applyAddon and old addons

### LINKED ISSUE #2161: Should we use a monorepo structure for addons?
I'm beginning to think it will be much easier to manage the addons if we just include them as part of the xterm repo (in `<root>/addons/<addon>`?) but still upload them separately to npm.  That way we could test them all as part of our CI to ensure breakages don't happen across versions, like what might have happened to ligatures recently. The main downsides I see are that issues will all remain in the 1 repo (maybe that's good?) and they won't be as good examples of what an external addon should be.

Tyriar:
Fixes #2142
Fixes #2161

---

Notes

- The addons are built like they were before using tsconfig.all.json, they have been moved out of src however to further disassociate them with the core code
- The demo uses the addons in a very similar way to how they would when using the module from npm.
- The addon folders don't get their dependencies installed, instead they rely on the main node_modules and only have a single script for publish which calls `../../../node_modules/.bin/tsc` directly
- The addons are scoped to @xtermjs as this is what it will be when we use the GH package registry
- I archived the repos xterm-addon-attach, xterm-addon-search and xterm-addon-web-links
- Removed the addon label and added separate area labels for each addon


mofux:
Works great and LGTM 👍. Does it make sense to add some words about how the addons work and how they share their dependencies with the root project to our wiki?

Tyriar:
@mofux good idea, I was also thinking it might be good to detail the https://github.com/xtermjs/xterm.js/issues/1507 wip as things will probably be in flux for quite a while still.

Tyriar:
Updated the wiki

Issue #1823, opened 2018-12-10
cleanup/enhancement of parser


jerch:
Things to do (PR #2177):
- [x] remove Array fallback for the transition table
This is not needed anymore since all supported engines have typed array support.
- [x] simplify transition table creation
With the transition table being a typed array the table can be prefilled with the default transition by a `.fill` call.
- [x] dont recreate arrays in `r()` everytime, instead slice from a blueprint
- [x] simplify unicode handling in `parse`
Use the new `NON_ASCII_PRINTABLE` where appropriate and restore the error state as error state.

Prolly done with next PRs:
- [x] eval read ahead in switch case (PR #2187)
The shortcuts for PRINT and CSI_PARAM are currently done before entering the switch statement, which makes them eval'd for every char that goes through the outer loop. Might be better repositioned to their own case in the switch statement.
- [x] colon notation in params state (see #2200, PR #2241)
With the true color support the parser should also get support for element notation, with is a rather big change to param state handling.
- [x] cleanup `dcsHandler` in `parse` - remove `null` type to lower branching (part of #2346)
- [x] rework fallback handler logic for parser actions
- [x] extend all parser actions with addHandler capability (#2332, PR #2346)
- [x] move all parser actions to typed array propagation (part of #2346)
- [ ] ~~optimize DCS unhook (#1909)~~ wont fix

PR #2168, opened 2019-06-01
Move EscapeSequenceParser to core

Tyriar:
Part of #1507

@jerch can you check out the EscapeSequenceParser.ts diff? I needed to tweak a bunch of stuff and want to make sure it's right 🙂


src/core/parser/EscapeSequenceParser.ts
```diff
@@ -50,8 +50,8 @@ export class TransitionTable {
    * @param action parser action to be done
    * @param next next parser state
    */
-  add(code: number, state: number, action: number | null, next: number | null): void {
-    this.table[state << 8 | code] = ((action | 0) << 4) | ((next === undefined) ? state : next);
+  add(code: number, state: number, action: number | null, next: number | undefined): void {
+    this.table[state << 8 | code] = ((action || 0) << 4) | ((next === undefined) ? state : next);
```
Tyriar: I changed `action | 0` to `action || 0`, was this right?

jerch: Hmm the old JS version relied on ` | 0` (anything cast to a number), but I think its not needed anymore (I think all add calls now have the full args list). Lemme check.

jerch: Changed arg types to `ParserAction` and `ParserState` and removed the undefined and null types.

src/core/parser/EscapeSequenceParser.ts
```diff
@@ -279,21 +279,10 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
   }
 
   public dispose(): void {
-    this._printHandlerFb = null;
```
jerch: I think we cannot (yet) remove these as they might hold a reference from the input handler with a terminal object?

src/core/parser/EscapeSequenceParser.ts
```diff
@@ -453,7 +442,9 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
             this._printHandler(data, print, i);
             print = -1;
           } else if (~dcs) {
-            dcsHandler.put(data, dcs, i);
+            if (dcsHandler) {
```
jerch: Was there an error regarding this? In theory `dcsHandler` should never be `null` here (if so the parser has a serious logic flaw). I am abit unhappy about `dcsHandler` being `null` in between, guess this needs some refactoring (to reduce branching).
Well its good for now with this addtional check.

jerch:
@Tyriar Added the need to rework the `dcsHandler` logic in https://github.com/xtermjs/xterm.js/issues/1823 (my parser TODO ticket). Guess I should this get this done rather soon.

src/core/parser/EscapeSequenceParser.ts
Tyriar: I think in that case it will still get disposed correctly and release the reference to input handler and terminal when the consumers reference to terminal is removed

src/core/parser/EscapeSequenceParser.ts
Tyriar: dcsHandler could be null according to the type system, if we're certain and want to avoid the if we could compel the type system with `!`, that's unsafe though.

PR #2176, opened 2019-06-02
ability to inject particular testfile to yarn test

jerch:


bin/test.js
```diff
@@ -18,8 +18,16 @@ const args = [
...
 ];
 
-cp.spawnSync(path.resolve(__dirname, '../node_modules/.bin/mocha'), args, {
-  cwd: path.resolve(__dirname, '..'),
-  env,
-  stdio: 'inherit'
-});
+// ability to inject particular test files via
+// yarn test [testFileA testFileB ...]
+const testFilesFromArgs = process.argv.slice(2);
```
Tyriar: Let's rename args to `testFiles` and then set args to a variable to avoid the ternary inside the spawn call

jerch: Changed.

PR #2177, opened 2019-06-02
parser cleanup

### LINKED ISSUE #1823: cleanup/enhancement of parser

jerch:
The parser has accumulated needed changes over time. Part of #1823.

Things to do:
- [x] remove Array fallback for the transition table
This is not needed anymore since all supported engines have typed array support.
- [x] simplify transition table creation
With the transition table being a typed array the table can be prefilled with the default transition by a `.fill` call.
- [x] dont recreate arrays in `r()` everytime, instead slice from a blueprint
- [x] simplify unicode handling in `parse`
Use the new `NON_ASCII_PRINTABLE` where appropriate and restore the error state as error state.

### DIFF
--- a/src/core/parser/EscapeSequenceParser.ts
+++ b/src/core/parser/EscapeSequenceParser.ts
@@ -189,7 +189,12 @@ export const VT500_TRANSITION_TABLE = (function (): TransitionTable {
   table.addMany(PRINTABLES, ParserState.DCS_PASSTHROUGH, ParserAction.DCS_PUT, ParserState.DCS_PASSTHROUGH);
   table.add(0x7f, ParserState.DCS_PASSTHROUGH, ParserAction.IGNORE, ParserState.DCS_PASSTHROUGH);
   table.addMany([0x1b, 0x9c], ParserState.DCS_PASSTHROUGH, ParserAction.DCS_UNHOOK, ParserState.GROUND);
+  // special handling of unicode chars
+  table.add(NON_ASCII_PRINTABLE, ParserState.GROUND, ParserAction.PRINT, ParserState.GROUND);
   table.add(NON_ASCII_PRINTABLE, ParserState.OSC_STRING, ParserAction.OSC_PUT, ParserState.OSC_STRING);
+  table.add(NON_ASCII_PRINTABLE, ParserState.CSI_IGNORE, ParserAction.IGNORE, ParserState.CSI_IGNORE);
+  table.add(NON_ASCII_PRINTABLE, ParserState.DCS_IGNORE, ParserAction.IGNORE, ParserState.DCS_IGNORE);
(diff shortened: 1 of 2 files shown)


src/core/parser/EscapeSequenceParser.ts
```diff
@@ -50,8 +45,8 @@ export class TransitionTable {
    * @param action parser action to be done
    * @param next next parser state
    */
-  add(code: number, state: ParserState, action: ParserAction, next: ParserState): void {
-    this.table[state << 8 | code] = (action << 4) | next;
+  public add(code: number, state: ParserState, action: ParserAction, next: ParserState): void {
+    this.table[state << 8 | code] = action << 4 | next;
```
Tyriar: What's 4 and 8?

src/core/parser/EscapeSequenceParser.ts
```diff
@@ -61,41 +56,40 @@ export class TransitionTable {
...
+
 /**
  * VT500 compatible transition table.
  * Taken from https://vt100.net/emu/dec_ansi_parser.
  */
 export const VT500_TRANSITION_TABLE = (function (): TransitionTable {
   const table: TransitionTable = new TransitionTable(4095);
 
+  // range macro
+  const blueprint = Array.apply(null, Array(256)).map((unused: any, i: number) => i);
```
Tyriar: What's 256?

jerch: byte range (0-255). The blueprint is used to slice value ranges from (faster version of `r`).

src/core/parser/EscapeSequenceParser.ts
jerch: Those values come from packing the values into bytes (explained in the class doc).

PR #2187, opened 2019-06-05
relocate parser look ahead loops

jerch:
Part of #1823.

This PR relocates the read ahead loops into the proper switch cases. It further reduces code complexity (less branching) with a slight speedup for the optimized actions.


jerch:
Speed comparison (tested with [this script](https://gist.github.com/jerch/827f18ba4af0b214ff45079403f5a1a3)):

- before:
```
   Context "examples/parser_throughput.js"
      Context "Parser performance - 50MB data"
         Context "print - a"
            Case "throughput" : 5 runs - average throughput: 432.44 MB/s
         Context "execute - \n"
            Case "throughput" : 5 runs - average throughput: 95.12 MB/s
         Context "escape - ESC E"
            Case "throughput" : 5 runs - average throughput: 53.92 MB/s
         Context "escape with collect - ESC % G"
            Case "throughput" : 5 runs - average throughput: 22.37 MB/s
         Context "simple csi - CSI A"
            Case "throughput" : 5 runs - average throughput: 84.30 MB/s
         Context "csi with collect - CSI ? p"
            Case "throughput" : 5 runs - average throughput: 55.03 MB/s
         Context "csi with params - CSI 1;2 m"
            Case "throughput" : 5 runs - average throughput: 104.91 MB/s
         Context "osc (small payload) - OSC 0;hi ST"
            Case "throughput" : 5 runs - average throughput: 50.15 MB/s
         Context "osc (big payload) - OSC 0;<text> ST"
            Case "throughput" : 5 runs - average throughput: 48.93 MB/s
         Context "DCS (small payload)"
            Case "throughput" : 5 runs - average throughput: 49.25 MB/s
         Context "DCS (big payload)"
            Case "throughput" : 5 runs - average throughput: 94.54 MB/s
```
- after:
```
   Context "examples/parser_throughput.js"
      Context "Parser performance - 50MB data"
         Context "print - a"
            Case "throughput" : 5 runs - average throughput: 632.97 MB/s
         Context "execute - \n"
            Case "throughput" : 5 runs - average throughput: 89.43 MB/s
         Context "escape - ESC E"
            Case "throughput" : 5 runs - average throughput: 58.36 MB/s
         Context "escape with collect - ESC % G"
            Case "throughput" : 5 runs - average throughput: 22.37 MB/s
         Context "simple csi - CSI A"
            Case "throughput" : 5 runs - average throughput: 89.36 MB/s
         Context "csi with collect - CSI ? p"
            Case "throughput" : 5 runs - average throughput: 94.44 MB/s
         Context "csi with params - CSI 1;2 m"
            Case "throughput" : 5 runs - average throughput: 124.12 MB/s
         Context "osc (small payload) - OSC 0;hi ST"
            Case "throughput" : 5 runs - average throughput: 50.08 MB/s
         Context "osc (big payload) - OSC 0;<text> ST"
            Case "throughput" : 5 runs - average throughput: 48.31 MB/s
         Context "DCS (small payload)"
            Case "throughput" : 5 runs - average throughput: 60.97 MB/s
         Context "DCS (big payload)"
            Case "throughput" : 5 runs - average throughput: 256.35 MB/s
```

src/core/parser/EscapeSequenceParser.ts
```diff
@@ -413,56 +423,48 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
     for (let i = 0; i < length; ++i) {
       code = data[i];
 
-      // shortcut for most chars (print action)
```
Tyriar: Did this not actually speed up mosts character printing like the comment indicates? Can you explain in a couple of sentences why this approach was so much slower for the `print - a` benchmark?

> Case "throughput" : 5 runs - average throughput: 432.44 MB/s

vs

> Case "throughput" : 5 runs - average throughput: 632.97 MB/s

src/core/parser/EscapeSequenceParser.ts
```diff
@@ -413,56 +423,48 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
...
-
       // normal transition & action lookup
       transition = table[currentState << TableAccess.INDEX_STATE_SHIFT | (code < 0xa0 ? code : NON_ASCII_PRINTABLE)];
       switch (transition >> TableAccess.TRANSITION_ACTION_SHIFT) {
         case ParserAction.PRINT:
-          print = (~print) ? print : i;
+          // read ahead with loop unrolling
+          // Note: 0x20 (SP) is included, 0x7F (DEL) is excluded
+          for (let j = i + 1; ; ++j) {
+            if (j >= length || (code = data[j]) < 0x20 || (code > 0x7e && code < NON_ASCII_PRINTABLE)) {
```
Tyriar: Is this section constructing the UTF-8 char? We can probably do this in another for look iterating on `j` since all the condition internals are the same.

src/core/parser/EscapeSequenceParser.ts
```diff
@@ -498,47 +504,46 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
...
+              }
+              i = j - 1;
+              break;
+            }
+          }
           break;
         case ParserAction.DCS_UNHOOK:
           if (dcsHandler) {
-            if (~dcs) dcsHandler.put(data, dcs, i);
+            // if (~dcs) dcsHandler.put(data, dcs, i);
```
Tyriar: ?

src/core/parser/EscapeSequenceParser.ts
jerch: Originally the parser was at ~120 MB/s. That was the reason for the shortcut in the first place, which boosted it up to >400 MB/s. This is still suboptimal, since the shortcut was placed before entering the switch statement, means the condition had to be test for every single char going through the outer loop.
So I moved the shortcut where it belongs to - the PRINT action in the switch statement. This cuts off the tests for all other chars (tiny boost to >500 MB/s). The last bit of the boost over 600 is the result of the loop unrolling (therefore the tripled lines of code).

src/core/parser/EscapeSequenceParser.ts
jerch: Yeah thats not beautiful, but the only way to get the benefit from loop unrolling. I resisted to it likewise for the other loops, as they dont show that much of a further speedup (the loop workload itself is to high to see the unrolling effect).

src/core/parser/EscapeSequenceParser.ts
jerch: Maybe the cat hitting the UNDO button? Should have been deleted :blush:

Tyriar: 😼

PR #2190, opened 2019-06-06
inital integration of xterm-benchmark

jerch:
PR to integrate benchmark tests to xterm.js repo.

TODO:
- [x] link against npm package
- [x] finetuning of config file:
  Skipping all tests that we are not interested in. Also skipping stddev/cv tests, since xterm-benchmark currently lacks the ability to test against absolute values (https://github.com/xtermjs/xterm-benchmark/issues/10).
- [x] add tests from xterm-benchmark repo
  Basic tests for raw parser speed and write/writeUtf8 throughput are added.
- [x] write some docs
  Basic usage description added below.

@Tyriar Still a few things to do before this really gets usable :cat: 

To run the benchmark tests, either type:
     `yarn benchmark single`
     `yarn benchmark baseline`
     `yarn benchmark eval`
    `yarn benchmark --help` (if you want the cmdline behind the script)


jerch:
Some hints about usage:
- `yarn benchmark baseline`
  Runs all benchmark tests 5 times and creates a file under `benchmark/baseline.log` with the baseline stats.
- `yarn benchmark eval`
  Runs all benchmark tests 5 times and creates a file under `benchmark/eval.log` with the test results. The results of the tests are printed to stdout to directly inspect them. The eval run will return with zero if all tests passed (useful for command chaining).

By default every run creates a log file under `benchmark/<epoch>.log`. The log files are line separated json (NDJSON), thus the results can easily be read for further processing. Baseline and eval log files contain another line with the aggregates test values. This line also contains the test results for eval runs.

How to use the benchmark tool during development:
- before changing the code do a baseline run
- change the code, either directly or by checking out another branch
- do an eval run
- results are printed to the console

Since most changes at hand affect only a small part of the code base it is often not useful to do full baseline and eval runs (which will trigger all benchmarks). To select a particular benchmark that covers your changes, run the benchmark like this:
`yarn benchmark <path/to/benchmark/file> -r 5` (5 runs without stats)
`yarn benchmark <path/to/benchmark/file> --baseline -r 5` (5 runs with baseline stats)
`yarn benchmark <path/to/benchmark/file> --eval -r 5` (5 runs with eval stats)

Its even possible to run a single `PerfCase` within a benchmark file, see `yarn benchmark --help`.

src/Terminal.benchmark.ts
```diff
@@ -0,0 +1,75 @@
+import { perfContext, before, ThroughputRuntimeCase } from 'xterm-benchmark';
```
Tyriar: Need a copyright

src/Terminal.benchmark.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import { Terminal } from 'Terminal';
+import { spawn } from 'node-pty';
+import { Utf8ToUtf32, stringFromCodePoint } from '../out/core/input/TextDecoder';
```
Tyriar: This points at `../out`

src/Terminal.benchmark.ts
```diff
@@ -0,0 +1,75 @@
+import { perfContext, before, ThroughputRuntimeCase } from 'xterm-benchmark';
+
+import { Terminal } from 'Terminal';
+import { spawn } from 'node-pty';
```
Tyriar: This line might be a problem when we move this to a separate tsconfig.json, I at that point think we might need to add an `electron` sub-project and move this there (as well as `Terminal.integration.ts`) since it touches node.

src/Terminal.benchmark.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+perfContext('Terminal: ls -lR /usr', () => {
+  let content = '';
+  let contentUtf8: Uint8Array;
+
+  before(async () => {
+    // grab output from "ls -lR /usr"
+    const p = spawn('ls', ['--color=auto', '-lR', '/usr'], {
+      name: 'xterm-color',
```
Tyriar: 256color?

src/core/parser/EscapeSequenceParser.benchmark.ts
```diff
@@ -0,0 +1,262 @@
+import { perfContext, before, beforeEach, ThroughputRuntimeCase } from 'xterm-benchmark';
```
Tyriar: Copyright

bin/benchmark.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * Copyright (c) 2019 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+const cp = require('child_process');
+const path = require('path');
+const glob = require('glob');
```
Tyriar: Do we need to depend on this?

src/Terminal.benchmark.ts
jerch: So keep it for now or change it right away?

src/Terminal.benchmark.ts
jerch: You have eagle eyes :smile:, happened from c&p.

bin/benchmark.js
jerch: This is needed to get wildcard shell expansion. I was not able to get it working with child_process with `{shell: true}`, although its stated that this should work. Maybe you have better idea?

src/Terminal.benchmark.ts
Tyriar: Yeah keep it for now, this change would need to be made towards the end of the project stuff

Tyriar:
Let's move the tests and benchmark.json into `./test` and fix up any strict problems.

jerch:
@Tyriar Its kinda impossible to separate the benchmark tests from `./src` without running into tons of troubles. Seems it relies way to much on internals that are not exposed. The last 2 commits therefore show a way to keep them in the source tree but in a separated folder. Maybe you have a better idea.

Also removed `bin/benchmark.js` as it was cumbersome from the begin. Now the benchmark commands are directly created in package.json as `benchmark`, `benchmark-baseline`, `benchmark-eval`. Hope this ill also work on Windows (as it has to set an env variable before running xterm-benchmark).

jerch:
@Tyriar
Done with it, guess this way it works best.

benchmarks/tsconfig.json
```diff
@@ -0,0 +1,32 @@
+{
+  "compilerOptions": {
```
Tyriar: I think we should be able to include the benchmarks into the `./test/tsconfig.json` project as well?

benchmarks/tsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+    "lib": [
+      "dom",
+      "es6",
+    ],
+    "outDir": "../out/test",
+    "types": [
+      "../../node_modules/@types/node"
```
Tyriar: Shouldn't this be `../node_modules/...`? I'm not sure you're using anything from node?

src/tsconfig.json
```diff
@@ -11,6 +11,7 @@
     ],
     "rootDir": ".",
     "outDir": "../out",
+    "composite": true,
```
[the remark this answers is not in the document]
jerch: This does not work, it keeps telling me that it cannot find module xy (from node_modules folder). Do I miss something?

benchmarks/tsconfig.json
jerch: Yes we can merge them and the test files once we managed to get them covered by just one tsconfig file. I have troubles to do so (see above).

benchmarks/tsconfig.json
jerch: I am using `Buffer` and `process` in one test file. For some reason both `../../` and `../` work. Yep, it should be `../`.

jerch:
@Tyriar 
Was able to get it working with these settings:
```json
{
  "compilerOptions": {
    "lib": [
      "dom",
      "es6",
    ],
    "outDir": "../out/test",
    "types": [
      "../node_modules/@types/node"
    ],
    "moduleResolution": "node",
    "strict": false,
    "target": "es2015",
    "module": "commonjs",
    "baseUrl": "./",
    "paths": {
      "common/*": [ "../src/common/*" ],
      "browser/*": [ "../src/browser/*" ],
      "Terminal": [ "../src/Terminal" ]
    },
  },
  "include": [
    "./**/*",
    "../typings/xterm.d.ts",
    "../src/**/*"
  ],
  "exclude": [
    "../**/*test.ts"
  ],
  "references": [
    { "path": "../src/common" },
    { "path": "../src/browser" },
  ]
}
```
Since this includes all source files from `src` it pulls in the whole xterm.js compilation under `out/test/src` and moves the compiled test files to `out/test/test` (to keep the folder lookup structure intact?). This way it is basically a second project compilation with tests included, which is fine for me. We could avoid this by including `out/**/*.d.ts` files instead of the raw TS source files, but those are only created for the new sub modules, thus I cannot import/type `Terminal.ts`.

If that is what you want for api tests as well, we can merge them with the benchmark tests.

package.json
```diff
@@ -20,7 +20,10 @@
     "prepare": "npm run build",
     "prepublishOnly": "npm run package",
     "watch": "tsc -b -w ./tsconfig.all.json --preserveWatchOutput",
-    "clean": "rm -rf lib out addons/*/lib"
+    "clean": "rm -rf lib out addons/*/lib",
+    "benchmark": "NODE_PATH=./out xterm-benchmark -r 5 -c benchmarks/benchmark.json",
```
Tyriar: What does benchmark do? I tried using it and it did nothing?

benchmarks/tsconfig.json
```diff
@@ -0,0 +1,32 @@
+{
+  "compilerOptions": {
+    "lib": [
+      "dom",
+      "es6",
+    ],
+    "outDir": "../out/test",
```
Tyriar: I was thinking we can put it into `out/benchmark` but that got me thinking why not just stick the test projects into src/ as well so that layout matches. So we would have a src/benchmark and src/test, thoughts?

package.json
jerch: Since I removed the `bin/benchmark.js` because of the ugly globbing I inserted `yarn benchmark` as an alias to `xterm-benchmark` with the project settings applied. Called without any benchmark files `xterm-benchmark` does nothing, but this way you can directly call into `xterm-benchmark`, e.g.:
```
yarn benchmark --help
yarn benchmark <some benchmark file>
```

benchmarks/tsconfig.json
jerch: I think it depends on the question whether we want to treat the benchmark tests as a separate project using the xterm sources in a slightly different way (thats what the settings in my last comment above do to get a hold of the not exported internals) or integrate them right into one project using the same tsconfig derivations as the xterm build.
Imho both is possible, from packaging/bundling viewpoint separate projects might be easier to handle later on, as we can separate them on folder level and just skip any test/benchmark related folder (test sources and builds) for bundling?

Tyriar: This structure looks good to me with a few notes:

- Let's use out-test instead of test-out so they're sorted next to each other.
- Let's reconsider the mocha changes after src/tsconfig.json is gone, I kind of like how the tests are right next to the source files as it makes it obvious which files have and don't have tessts.

jerch:
Changed the folder structure as you suggested. Currently the benchmark project will also pull half of the xterm output files (due to my hack with referencing `Terminal` under paths). Once the new absolute imports are all in place I can remove it and the folder will only contain the test files.

azure-pipelines.yml
```diff
@@ -83,6 +83,32 @@ jobs:
...
+  - task: geeklearningio.gl-vsts-tasks-yarn.yarn-installer-task.YarnInstaller@2
+    inputs:
+      versionSpec: "1.9.4"
+    displayName: 'Install Yarn'
+  - script: |
+      yarn
+    displayName: 'Install dependencies and build'
+  - script: |
+      yarn benchmark-baseline
+    displayName: 'Baseline data'
```
Tyriar: Should this be `git checkout <some tag>` before the baseline is done? Also I'm thinking we should disable this for now, maybe we want to take the bot approach eg. `@xterm-bot benchmark` would trigger it?

package.json
```diff
@@ -14,13 +14,16 @@
     "lint": "tslint 'src/**/*.ts' './demo/**/*.ts' './addons/**/*.ts'",
     "test": "npm run test-unit",
     "posttest": "npm run lint",
-    "test-api": "mocha \"**/*.api.js\"",
+    "test-api": "mocha \"./out-test/api/*.api.js\"",
```
Tyriar: `test-api` actually pulls in the addon API tests as well, I think we're safe to put it back how it was?

azure-pipelines.yml
jerch: Yes, I also think we should it disable here and go with the bot approach to not delay tests any further.

package.json
jerch: Woops, ok changed it back.

PR #2197, opened 2019-06-06
Properly clone default attr data on DECSTR

### LINKED ISSUE #2196: Soft reset (DECSTR) breaks the default attribute
Causing input text to become bold:

https://github.com/microsoft/vscode/issues/75021

I was seeing strange things as well where it would become bold, colored and dimmed.

These changes seem to carry over to new terminal as well which indicates to me the char atlas gets corrupted after a reset.

Screen Shot 2019-06-06 at 2 12 58 PM

Tyriar:
Fixes #2196


Tyriar:
Wrapped DEFAULT_ATTR_DATA's value in Object.freeze, so the old code would hit this error:

```
Uncaught TypeError: Cannot assign to read only property 'fg' of object '#<AttributeData>'
```

PR #2199, opened 2019-06-07
fix repeat preceding character sequence (REP)

jerch:
Due to the change to `TERM=xterm-256color` ncurses based programs are likely to emit REP sequences to repeat the previous character in the data stream. We already had REP in `InputHandler` but it was broken in more than one way.
The changes are backed up with API tests and one integration test (test number 33).

### DIFF
--- a/src/InputHandler.api.ts
+++ b/src/InputHandler.api.ts
@@ -276,14 +276,14 @@ describe('InputHandler Integration Tests', function(): void {
       assert.deepEqual(await getLinesAsArray(2), ['##########', '######']);
       await page.evaluate(`
         window.term.reset();
-        window.term._core.wraparoundMode = false;
+        window.term.write('\x1b[?7l');  // disable wrap around
         window.term.write('#\x1b[15b');
       `);
       assert.deepEqual(await getLinesAsArray(2), ['##########', '']);
       // any successful sequence should reset REP
       await page.evaluate(`
         window.term.reset();
-        window.term._core.wraparoundMode = true;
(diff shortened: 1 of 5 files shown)


src/InputHandler.api.ts
```diff
@@ -274,3 +324,11 @@ async function simulatePaste(text: string): Promise<string> {
   `);
   return await page.evaluate(`window.result_${id}`);
 }
+
+async function getCursor(): Promise<{col: number, row: number}> {
+  return page.evaluate(`
+  (function() {
+    return {col: term._core.buffer.x, row: term._core.buffer.y};
```
Tyriar: This should be `term.buffer.cursorX`/`Y`

src/InputHandler.api.ts
```diff
@@ -241,6 +241,56 @@ describe('InputHandler Integration Tests', function(): void {
...
+      assert.deepEqual(await getLinesAsArray(1), ['e\u0301eeeee']);
+      // should wrap correctly
+      await page.evaluate(`
+        window.term.reset();
+        window.term.write('#\x1b[15b');
+      `);
+      assert.deepEqual(await getLinesAsArray(2), ['##########', '######']);
+      await page.evaluate(`
+        window.term.reset();
+        window.term._core.wraparoundMode = false;
```
Tyriar: We should use the sequence to set this as _core could change

src/InputHandler.api.ts
jerch: :blush: API tests with private access, lol... fixed.

PR #2201, opened 2019-06-08
Speedup DOM renderer

jerch:
Just a playground PR atm, but the early results look promising. The DOM renderer can run much faster if we manage to lower the span creation. Currently every cell with content from the left gets its own span element, which creates massive workload for the layouter.

Idea:
The idea is not new, in fact an earlier version of the DOM renderer worked like this - since we are in a monospace world, we can omit span creation for cells that share their attributes. The catch is that some chars might not fit the monospace grid of the default font for several reasons (a broken glyph in the font itself, bold variant with wider characters, the font renderer stepping in returning a glyph from a different font). To circumvent this one has to measure every single char's running width and compare it with the cell width. Thats what this PR tries to achieve. It measures the width once (lazy, on first occurence) and stores whether the char meets the cell width.

Results so far:
- penalty of the char measuring/lookup is neglectible (dont see it in timeline reports)
- with a monospace font `ls -lR /usr` runtime is only 20% longer than for the canvas renderer
- with Arial `ls -lR /usr` runtime is about the same as for the current DOM renderer

### DIFF
--- a/demo/client.ts
+++ b/demo/client.ts
@@ -93,7 +93,9 @@ function createTerminal(): void {
 
   const isWindows = ['Windows', 'Win16', 'Win32', 'WinCE'].indexOf(navigator.platform) >= 0;
   term = new Terminal({
-    windowsMode: isWindows
+    windowsMode: isWindows,
+    rendererType: 'dom',
+    // fontFamily: "Arial"
   } as ITerminalOptions);
 
   // Load addons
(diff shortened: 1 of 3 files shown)


jerch:
- **canvas renderer:**
grafik

- **current DOM renderer:**
grafik

- **PR with monospace font:**
grafik

- **PR with Arial:**
grafik

jerch:
@Tyriar: Yes the correct alignment is tricky, and its not yet clear if this approach will do here. The behavior of `getBoundingClientRect` in chrome is weird. prolly the current `CharMeasure` is slightly off because of this, too.

Closing this PR as it was just meant as an eval playground. A real implementation would need to cover further things like:
- prolly also measure and store italics and bolds
- full integration with codebase (respect events and such)
- prolly a better approach for the measuring itself (canvas based with `measureText` is about >5 times faster and saves additonal DOM roundtrips)
- grid alignment tests

PR #2203, opened 2019-06-08
Move CharWidth into common and move API tests into ./test

Tyriar:
Part of #1507


jerch:
@Tyriar Should we move the upcoming benchmark tests also in a separate folder?

Tyriar:
@jerch yes I think they should live in this same folder, this is the problem I mentioned in one of the comments over there.

Tyriar:
@jerch benchmark.json could also live in test/

PR #2216, opened 2019-06-09
Partially support IE11

### LINKED ISSUE #1889: Remove IE hacks
I think it's about time to drop IE11 support, let's get rid of it when we ship v4

Tyriar:
Fixes #1889


Tyriar:
The clipboard stuff works fine in IE11, so i changed to partial support as everything should stuff work fine.

PR #2217, opened 2019-06-10
Fix saved cursor position on resize of alt screen

thekondr:
Hi everyone,

Just wanted to let you know that there is an issue with keeping the saved cursor position in a correct state when resize happens while the alternative screen is active.

Steps to reproduce:

1. Produce a long output (a few screens is enough)
1. Make a window with xterm small
1. Open any shell application to enter an alternate screen
1. Enlarge xterm window
1. Close the shell application

Expected: cursor is on the last line of shell
Actual: cursor is somewhere in the middle of a screen and all shell output goes above the old output

Basically, this PR fixes the issue by storing the absolute value of cursor's Y position instead of relative, but it breaks the interface of `IBuffer` because the meaning of the `savedY` variable has changed.

Do you think it would be good enough to create a property getter/setter for `savedY` and implement them in terms of `savedAbsoluteY`?


thekondr:
@Tyriar does the overall problem make sense?

Tyriar:
Are you sure this is how the saved cursor works in other terminals?

thekondr:
> Are you sure this is how the saved cursor works in other terminals?

Yes, I am and this makes sense because the old output "under the alt screen" is resized accordingly while the saved cursor is not.

When reproduced, the current behavior looks like a rendering bug because the new output goes over the old output.

image

Tyriar:
@thekondr I'm trying to repro the issue, am I doing something wrong?

Run `ll` and resize to 30 cols

image

Open vim

image

Resize to 80 cols

image

Exit vim

image

I wouldn't have expected the cursor to be saved when entering/exiting the alt buffer?

thekondr:
@Tyriar I wasn't able to reproduce it with zsh. If you are using zsh it is better to try with bash. And instead of increasing the number of cols, please try to increase the number of rows.

Tyriar:
@thekondr ah makes sense, I can repro now 👍

thekondr:
@Tyriar that wasn't actually ready for a merge, I planned to discuss further changes once you reproduce the issue.

These changes break the interface of `IBuffer` because the meaning of the `savedY` variable has changed. It contains an absolute Y value instead of relative.

Do you think it would be good enough to create a property getter/setter for `savedY` and implement them in terms of `savedAbsoluteY`? Or maybe since the `IBuffer` is still experimental just break the behavior?

Tyriar:
It ended up breaking the tests locally but not in pipelines for some reason, I merged in https://github.com/xtermjs/xterm.js/pull/2236 to fix that.

> These changes break the interface of IBuffer because the meaning of the savedY variable has changed. It contains an absolute Y value instead of relative.

Do we need to differentiate the 2? Would we use one and not the other in some cases?

> Or maybe since the IBuffer is still experimental just break the behavior?

There are 2 `IBuffer`'s the one in the API (experimental) and the internal class which is not, `savedY` isn't included in the API part so I think that's fine? https://github.com/xtermjs/xterm.js/blob/795ca2af2e44029a02b80e071ca1dd18706df339/typings/xterm.d.ts#L836

thekondr:
I was thinking that the `savedY` is a part of the public API. Since it is not, then there is no reason to differentiate the 2 and everything seems fine. Thanks.

Tyriar:
@thekondr 👌 thanks again for finding and fixing the bug, if any other issues do some out of it we can fix and write tests for them.

PR #2222, opened 2019-06-13
Fix NPE in _setTheme

### LINKED ISSUE #2220: Cannot read property 'setTheme' of undefined
Hi, we've had such code:
```
this.terminal.setOption('theme', ....);
this.terminal.open(....);
```

However, with the latest changes, the internal `_colorManager` gets initialized on `.open()`, thus the first line fails with:
```
TypeError: Cannot read property 'setTheme' of undefined
    at e._setTheme (Terminal.js.pre-build-optimizer.js:587)
    at e.setOption (Terminal.js.pre-build-optimizer.js:307)
    at t.setOption (Terminal.js.pre-build-optimizer.js:193)
```

It's an easy fix on our side (just swap two lines), however, it makes sense to decouple visual customization from connectivity responsibilities. Was it an unintentional change?

Tyriar:
Fixes #2220

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -700,7 +700,9 @@ export class Terminal extends Disposable implements ITerminal, IDisposable, IInp
    */
   private _setTheme(theme: ITheme): void {
     this._theme = theme;
-    this._colorManager.setTheme(theme);
+    if (this._colorManager) {
+      this._colorManager.setTheme(theme);
+    }
     if (this._renderService) {
       this._renderService.setColors(this._colorManager.colors);
     }


Tyriar:
@jerch are you sure about that? `Terminal.open` should call `_setTheme` with `options.theme` which is still valid I think?

Tyriar:
Also this is one of the class of bugs that won't happen anymore when we move this part of strict mode

jerch:
@Well the OP calls `setTheme` before `open`, thus this fix will silently ignore the `setTheme` call.

Tyriar:
Nah I just verified that it works, `setOption('theme'` sets `this.optionsService.options.theme`, this is used in `open` to recall `_setTheme`.

Tyriar:
And yes this will get much more polished and less error prone when the theme stuff is all hooked up with the new architecture.

JounQin:
https://unpkg.com/xterm@3.14.4/lib/Terminal.js
image
v3.14.4 on npm is still broken...

Tyriar:
@JounQin reopened https://github.com/xtermjs/xterm.js/issues/2220 and set milestone

PR #2241, opened 2019-06-17
colon notation for the parser

### LINKED ISSUE #2200: parser colon support
Since more and more escape sequences with the official colon notation for subparameters appear in other emulators, we should change the parser to support it as well.

General notation: 
 > CSI ... ; Px : S1 :  S2 : ... ; ... 

where S1 and S2 can be seen as subparameters to Px.

For subparameters the following rules apply:
- can only be digits
- link to the parameter right before first colon
- missing subparameters (empty colons) are not 0 by default, instead some custom default value defined by target action
- skipped colons are treated as missing subparameters and fill up with default values to the right

In particular we are interested in supporting the colon notation for color SGRs like this:
```
#sequences                     #params to SGR
// truecolor
38 : 2 : : r : g : b ; ... ==> [38, [2, , r, g, b], ...] (new syntax)
38 ; 2 : : r : g : b ; ... ==> [38, 2, [ , r, g, b], ...] (mixed syntax)
38 ; 2 ; r ; g ; b ; ...   ==> [38, 2, r, g, b] (old syntax)
// 256 palette
38 : 5 : v ; ...           ==> [38, [5, v], ...] (new syntax)
38 ; 5 : v ; ...           ==> [38, 5, [v], ...] (mixed syntax)
38 ; 5 ; v ; ...           ==> [38, 5, v] (old syntax) 
```

These variants are in line with recent SGR colon support in xterm (tested with v346).

Edit:
Also add env var `COLORTERM=truecolor` as suggested by vte.

jerch:
PR to get support for colon parameter notation in the parser. Should fix #2200.

Several things todo:
- [x] new `Params` type to hold params and subparams in a speedy manner
- [x] apply ':' to the transition table and extend parse loop accordingly
- [x] apply new `Params` type to parser API and inputhandler methods
- [x] fix SGR handling for sub params
- [x] parser tests for new colon sequences
- [x] integration tests for stacked SGR params/subparams
- [x] add `COLORTERM=truecolor` to env


jerch:
@Tyriar:
The last commit introduces the new params type to the rest of the codebase. Due to the `addCsiHandler` this type gets pulled right up to `xterm.d.ts` - we need to find a solution for that.

jerch:
Some notes on the subparams implementation:

- parser support
  So far there is only one known CSI sequence with subparams - SGR. Due to syntactical limitations of sequences where the last byte decides which final action has to be triggered there is no elegant way to limit the subparams handling to SGR only at parser level. Thus the parser applies the subparams logic to all escape sequences with digit params (all `CSI`, all `DCS`).
- new `Params` type
  Any action handler that received `params: number[]` before now receives `params: IParams`. Compared to the old number array this new data type has several constraints:
  - all params and subparams are borrowed and only valid for the current handler invocation (clone the data if you need a persistant version)
  - params limits: int16_t, default: 0 (ZDM - zero default mode)
  - subparams limits: int16_t, default: -1
  - 32 slots for each params / subparams (configurable via ctor), >32 will be silently ignored
  - params can be accessed by `.params[idx]`
  - reading multiple params requires a `.length` check, reading beyond length might return arbitrary data
  - existence of `.params[0]` is guaranteed, most sequences that only support one param can omit the length check
  - existence of subparams can be checked for every valid param
- SGR
  This PR implements the subparams handling for foreground and background colors in SGR in an xterm compatible fashion (allowing some mixed semicolon/colon notations additionally to both standards). Support for underline color (SGR 58) and style (SGR 4) is not implemented yet.

src/InputHandler.ts
```diff
@@ -618,11 +613,8 @@ export class InputHandler extends Disposable implements IInputHandler {
    * CSI Ps C
    * Cursor Forward Ps Times (default = 1) (CUF).
    */
-  public cursorForward(params: number[]): void {
-    let param = params[0];
-    if (param < 1) {
-      param = 1;
-    }
+  public cursorForward(params: IParams): void {
+    const param = params.params[0] || 1;
```
Tyriar: I think this will let negative numbers though now unless `IParams` handles this, is that a concern? `Math.max(params.params[0], 1)`?

jerch: Well, currently there is no way to set this to a negative value from a sequence. The parser currently applies zero default mode to params (as it always did), maybe just change the params typed array to uint16_t to be on the save side?

src/common/parser/Params.ts
```diff
@@ -0,0 +1,230 @@
...
+    newParams._subParamsLength = this._subParamsLength;
+    newParams._subParamsIdx.set(this._subParamsIdx);
+    return newParams;
+  }
+
+  /**
+   * Get a JS array representation of the current parameters and sub parameters.
+   * The array is structured as follows:
+   *    sequence: "1;2:3:4;5::6"
+   *    array   : [1, 2, [3, 4], 5, [-1, 6]]
```
Tyriar: `::` => `-1`? Is that the default value for sub parameters?

demo/server.js
```diff
@@ -37,14 +37,16 @@ function startServer() {
   });
 
   app.post('/terminals', function (req, res) {
+    const env = Object.assign({}, process.env);
+    env['COLORTERM'] = 'truecolor';
```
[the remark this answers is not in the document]
jerch: Yes, once the colon notation is in place we are compatible with whats implied by that env var.

src/Types.d.ts
```diff
@@ -112,42 +113,42 @@ export interface IInputHandler {
...
-  /** CSI h */ setMode(params?: number[], collect?: string): void;
-  /** CSI l */ resetMode(params?: number[], collect?: string): void;
-  /** CSI m */ charAttributes(params?: number[]): void;
-  /** CSI n */ deviceStatus(params?: number[], collect?: string): void;
-  /** CSI p */ softReset(params?: number[], collect?: string): void;
-  /** CSI q */ setCursorStyle(params?: number[], collect?: string): void;
-  /** CSI r */ setScrollRegion(params?: number[], collect?: string): void;
-  /** CSI s */ saveCursor(params?: number[]): void;
-  /** CSI u */ restoreCursor(params?: number[]): void;
+  /** CSI @ */ insertChars(params?: IParams): void;
```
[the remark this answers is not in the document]
jerch: Eww good finding. Def. have to check all handler again, if they check for `params` before getting into business. Alternatively we could make params non optional for all handlers, but I am not sure if this would break on direct invocations here and there. So Id have to check this first.
Any preference whether to go with optional / non optional here?

src/common/parser/Params.ts
jerch: Kinda.

In the past I had a longer discussion with @egmontkob about ZDM (zero default mode) for params (vte does not do this for params either). We do currently to be in line with the parser from vt100.net, and I would not change that for params unless we run into troubles for a certain sequence.
For subparams there is no ZDM thingy, which means it might be anything. I took `-1` a placeholder for default since negative values cannot be reached by normal sequence means. This way the final action can place its real default value whenever it sees -1.

src/InputHandler.ts
jerch: Thought again about this - imho it can kepp it this way as long as we stick to ZDM. If we ever change this, we have to change default handling in all handlers anyway.

src/Types.d.ts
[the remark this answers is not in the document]
jerch: Removed the question mark from `params` thus making it mandatory. They were not optional on `Inputhandler` anyway thus this chance is just getting the interface in line with the class.

jerch:
Last changes:
- allow greater values (int32 now), no need to stick to smallish int16
- clamp max value to 0x7FFFFFFF (fixes possible negative overflow)
- reject values below `-1`

typings/xterm.d.ts
```diff
@@ -499,7 +499,7 @@ declare module 'xterm' {
      * The most recently-added handler is tried first.
      * @return An IDisposable you can call to remove this handler.
      */
-    addCsiHandler(flag: string, callback: (params: number[], collect: string) => boolean): IDisposable;
+    addCsiHandler(flag: string, callback: (params: IParams, collect: string) => boolean): IDisposable;
```
Tyriar: Maybe we should just use `(number | number[])[]` here to hide the complexity?

jerch: Fixed. Note that the translation to `(number | number[])[]` by calling `Params.toArray()` is ~5 times slower than directly using the `Param` type. Imho not a problem, since most handlers will be not invoked often.

src/InputHandler.ts
```diff
@@ -1286,17 +1242,15 @@ export class InputHandler extends Disposable implements IInputHandler {
    * Modes:
    *   http: *vt100.net/docs/vt220-rm/chapter4.html
    */
-  public setMode(params: number[], collect?: string): void {
-    if (params.length > 1) {
```
Tyriar: Much cleaner 👍

src/common/parser/EscapeSequenceParser.test.ts
```diff
@@ -26,11 +29,14 @@ class TestEscapeSequenceParser extends EscapeSequenceParser {
   public set osc(value: string) {
     this._osc = value;
   }
-  public get params(): number[] {
-    return this._params;
+  public get params(): ParamsArray {
+    return this._params.toArray() as ParamsArray;
```
Tyriar: We shouldn't need to cast this.

src/common/parser/EscapeSequenceParser.test.ts
```diff
@@ -3,11 +3,14 @@
...
 
-import { IDcsHandler, IParsingState } from 'common/parser/Types';
+import { IDcsHandler, IParsingState, IParams } from 'common/parser/Types';
 import { EscapeSequenceParser, TransitionTable, VT500_TRANSITION_TABLE } from 'common/parser/EscapeSequenceParser';
 import * as chai from 'chai';
 import { StringToUtf32, stringFromCodePoint } from 'common/input/TextDecoder';
 import { ParserState } from 'common/parser/Constants';
+import { Params } from 'common/parser/Params';
+
+type ParamsArray = (number | number[])[];
```
Tyriar: Should this type be in parser/Types?

src/common/parser/EscapeSequenceParser.test.ts
```diff
@@ -562,16 +568,17 @@ describe('EscapeSequenceParser', function (): void {
...
+    it('trans CSI_ENTRY --> CSI_PARAM for ":" (0x3a)', function (): void {
       parser.reset();
       parser.currentState = ParserState.CSI_ENTRY;
       parse(parser, '\x3a');
-      chai.expect(parser.currentState).equal(ParserState.CSI_IGNORE);
+      chai.expect(parser.currentState).equal(ParserState.CSI_PARAM);
       parser.reset();
     });
+    /*
     it('trans CSI_PARAM --> CSI_IGNORE', function (): void {
```
Tyriar: Commented out?

src/common/parser/EscapeSequenceParser.test.ts
```diff
@@ -580,6 +587,19 @@ describe('EscapeSequenceParser', function (): void {
...
+    */
+    it('trans CSI_PARAM --> CSI_IGNORE', function (): void {
+      parser.reset();
+      const chars = ['\x3c', '\x3d', '\x3e', '\x3f'];
+      for (let i = 0; i < chars.length; ++i) {
+        chai.expect(parser.params).eql([0]);
+        parser.currentState = ParserState.CSI_PARAM;
+        parse(parser, '\x3b' + chars[i]);
+        chai.expect(parser.currentState).equal(ParserState.CSI_IGNORE);
+        // chai.expect(parser.params).eql([0, 0]);
```
Tyriar: Comment?

src/common/parser/EscapeSequenceParser.ts
```diff
@@ -173,19 +171,16 @@ export const VT500_TRANSITION_TABLE = (function (): TransitionTable {
   table.add(0x7f, ParserState.DCS_ENTRY, ParserAction.IGNORE, ParserState.DCS_ENTRY);
   table.addMany(r(0x1c, 0x20), ParserState.DCS_ENTRY, ParserAction.IGNORE, ParserState.DCS_ENTRY);
   table.addMany(r(0x20, 0x30), ParserState.DCS_ENTRY, ParserAction.COLLECT, ParserState.DCS_INTERMEDIATE);
-  table.add(0x3a, ParserState.DCS_ENTRY, ParserAction.IGNORE, ParserState.DCS_IGNORE);
-  table.addMany(r(0x30, 0x3a), ParserState.DCS_ENTRY, ParserAction.PARAM, ParserState.DCS_PARAM);
-  table.add(0x3b, ParserState.DCS_ENTRY, ParserAction.PARAM, ParserState.DCS_PARAM);
```
Tyriar: I'm not sure what all these changes are about?

src/common/parser/EscapeSequenceParser.ts
```diff
@@ -472,7 +469,7 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
               currentState,
               osc,
               collect,
-              params,
+              params: params,
```
Tyriar: This can be undone

src/common/parser/Params.ts
```diff
@@ -0,0 +1,245 @@
...
+const MAX_SUBPARAMS = 256;
+
+/**
+ * Params storage class.
+ * This type is used by the parser to accumulate sequence parameters and sub parameters
+ * and transmit them to the input handler actions.
+ *
+ * Usage in action handler:
+ *   ```typescript
+ *   function handler(params: IParams): void {
```
Tyriar: Probably best to keep samples in the core code as they'll be subject to bitrot

src/common/parser/EscapeSequenceParser.test.ts
jerch: Yeah, its remnant from a transit state, where `ParamsArray` pointed to `number[]`. Removed.

src/common/parser/EscapeSequenceParser.test.ts
jerch: Yepp, good idea.

src/common/parser/EscapeSequenceParser.test.ts
jerch: Eww forgot to re-enable this.

src/common/parser/EscapeSequenceParser.test.ts
jerch: Same as above, fixed.

src/common/parser/EscapeSequenceParser.ts
jerch: All the table changes move ":" (0x3a) from `*_IGNORE` state to `*_PARAM` state. Previously a ':' would mark a sequence as faulty (thus ignoring), now it will be accounted under `ParserAction.PARAM`. The differentiation into params/subparams itself happens in the inner loop of `ParserAction.PARAM` in `parse`.

I had also to apply the subparams rules to DCS since we have only one `ParserAction.PARAM` action which deals with digit params in any sequence. Imho thats not a big deal as subparams can always be ignored on final action handler level later on. Doing otherwise (separating this early in the parser) would have introduced a new parser action with all the bookkeeping along the way. I try to avoid adding states/actions to not break to much with the parser from vt100.net.

src/common/parser/EscapeSequenceParser.ts
jerch: Done.

src/common/parser/Params.ts
jerch: Imho thats a common problem for docs in general (always have to remind myself to double check docs after changing something lol). Removed the snippet, the class API is self explaining.

jerch:
@Tyriar Further documented ZDM. If you feel uneasy about it (we are not in line with the latest specs here), we could lift that restriction. For this to work we only have to fix all action handlers in `Inputhandler` to do a proper check against the param default of -1.

Up to you, I currently see no need to switch.

Issue #2091, opened 2019-05-17
Make word border configurable


Eugeny:
Currently `SelectionManager._getWordAt` is hardwired to expand selection until the nearest whitespace. Could we have an option to configure what characters are considered to be a part of word?

Tyriar:
Sure we could have a `wordSeparator` option that replaces this:

https://github.com/xtermjs/xterm.js/blob/610c4c9726402591fa783cdece0f1166fa3437a5/src/SelectionManager.ts#L43

PR #2244, opened 2019-06-21
Define wordSeparator option

MaxRis:
Fixes https://github.com/xtermjs/xterm.js/issues/2091


MaxRis:
@Tyriar thank you for fast response on that!

PR #2281, opened 2019-07-02
Centers search scroll in viewport

lramos15:
This causes search to be centered inside the viewport rather than defaulting to the top of the terminal. This is a fix for https://github.com/microsoft/vscode/issues/50623.

### DIFF
--- a/addons/xterm-addon-search/src/SearchAddon.ts
+++ b/addons/xterm-addon-search/src/SearchAddon.ts
@@ -4,7 +4,6 @@
  */
 
 import { Terminal, IDisposable, ITerminalAddon } from 'xterm';
-import { floor } from 'mathjs';
 
 export interface ISearchOptions {
   regex?: boolean;
@@ -375,7 +374,7 @@ export class SearchAddon implements ITerminalAddon {
     }
     terminal.select(result.col, result.row, result.term.length);
     let scroll = result.row - terminal.buffer.viewportY;
-    scroll = scroll - floor(terminal.rows / 2);
+    scroll = scroll - Math.floor(terminal.rows / 2);
     terminal.scrollLines(scroll);
     return true;
   }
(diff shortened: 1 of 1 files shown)


addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -4,6 +4,7 @@
  */
 
 import { Terminal, IDisposable, ITerminalAddon } from 'xterm';
+import { floor } from 'mathjs';
```
Tyriar: Accidental import? Use `Math.floor`

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -373,7 +374,9 @@ export class SearchAddon implements ITerminalAddon {
       return false;
     }
     terminal.select(result.col, result.row, result.term.length);
-    terminal.scrollLines(result.row - terminal.buffer.viewportY);
+    let scroll = result.row - terminal.buffer.viewportY;
+    scroll = scroll - floor(terminal.rows / 2);
```
Tyriar: I think you need to wrap this in a `Math.max(..., 0)` to avoid negatives?

lramos15: I will test it out, I wasn't having any issues after testing this code. I will see if Math.max changes any functionality.

addons/xterm-addon-search/src/SearchAddon.ts
lramos15: Ah yes. I did `math.floor` and then that didn't work so I did just `floor` and it told me to import that module. I forgot that it is capital `Math`.

addons/xterm-addon-search/src/SearchAddon.ts
lramos15: This doesn't work in the case of when the word I want is above my viewport because I need a negative scroll to scroll up. Therefore it defaults to 0 in the Math.max and it looks like nothing happens.

Tyriar: :+1: I was thinking it was the absolute row

PR #2282, opened 2019-07-03
Recent changes from master

MaxRis:


Tyriar:
Normally we only do patch releases for very bad bugs, I'm not sure the isWrapped change makes the cut? We expect v4 to go out in about 3-4 weeks but you can use the beta build published to npm if you want to keep up to date.

MaxRis:
`isWrapped` issue can be pretty annoying for users https://github.com/felixse/FluentTerminal/issues/420

Thanks, I afraid that Windows 10 UWP WebView control based on Microsoft Edge rendering engine might will not work good with v4.

Tyriar:
@felixse https://github.com/felixse/FluentTerminal/issues/420 isn't a regression from 3.13 is it?

> Thanks, I afraid that Windows 10 UWP WebView control based on Microsoft Edge rendering engine might will not work good with v4.

@MaxRis oh? This is news to me? 🤔

MaxRis:
@Tyriar we had discussion with @felixse previously regarding FluentTerminal incompatibility with upcoming changes in v4:

UWP app uses WebView control to render Xterm.js; WebView control embeds "Microsoft Edge" engine; And Edge still doesn't support webgl2 https://wpdev.uservoice.com/forums/257854-microsoft-edge-developer/suggestions/6517037-webgl-2-0 ; it seems you did the switch to webgl2 based rendering https://github.com/xtermjs/xterm.js/pull/1790 .

I'm curious if it's possible to develop some plan of support for UWP platform in upcoming v4 release?
Felix figured out that Microsoft pretty close to release WebView2 control based on Chromium ( https://docs.microsoft.com/en-us/microsoft-edge/hosting/webview2 ), but before it happened we need some solution.

Tyriar:
@MaxRis the WebGL renderer is completely opt-in right now, v5 is probably when we'll remove the canvas renderer, but there will always be the DOM one that will work in all browsers. v5 is probably another year or so off given our history. FluentTerminal should work just fine as it always has, just don't use the WebGL renderer on Windows.

I'm going to close this as I don't think these changes meet the bar for a patch release.

MaxRis:
@Tyriar makes sense, thank you!

Tyriar:
@MaxRis I want to fix https://github.com/xtermjs/xterm.js/issues/2220 in 3.14.5 so I'll include https://github.com/xtermjs/xterm.js/issues/2296, the word separator change is a new feature though so we'll leave that out.

MaxRis:
@Tyriar great! Should I create PR for #2296 ?

Tyriar:
@MaxRis I'll do it tomorrow if I get the time, should be just cherry picking the commits.

PR #2288, opened 2019-07-04
fix edge cases in inputhandler methods

### LINKED ISSUE #1434: cursor at wrong column position behind last cell
I tried to repair some of the escape sequence test files and find the same failure pattern for many of them - the cursor points to `terminal.cols` instead of `terminal.cols - 1`. Note a real VT cant move the cursor outside the screen to the right.

Not sure yet if this can be fixed at individual method level (some already have a fix in place like `HPositionRelative`, maybe from earlier bug reports) or if it has to be global change (guess the latter since I am not able to fix the ICH test with only local changes while a global change does the trick). Affected by this are at least the following terminal primitives:
- BS (backspace)
- ED (erase display)
- EL (erase line)
- ICH (insert character)

Also `top` output is affected by this, xterm places the cursor at the end of the screen (invisible) while xterm.js has the cursor at different places after a refresh in the first column (cursor is not hidden, `CSI ? 25 l` is currently not implemented). The white underlined table header is one cell short in xterm while xterm.js fills all cells to the right. With the global change top output looks identical. Other typical apps seem not to be affected by this (quick tests with ls, vim, emacs, mc, htop).

Whatever the fix will be it has to be tested extensively due to a fundamental change in the cursor handling :disappointed: 

By global I mean that a method of `InputHandler` adjusts the cursor column position to `cols - 1` before returning, thus the global accessible cursor position never contains a value bigger than `cols - 1` between method calls. A local change would allow this faulty intermediate state and fix the position on entering the method (partly implemented now). Not sure if any other code parts beside the autowrap in `addChar` are relying on the `buffer.x >= terminal.cols` condition, those would need to be changed as well for a global fix.

Edit:
Sorry for the messed up bug description, tried to clean it up a bit. :blush:

### LINKED ISSUE #2020: Probably incorrect horizontal cursor positioning with certain CSI sequences
It appears to be possible to put the `buffer.x` position into an odd state with certain CSI sequences.

### Sample

```
\E[4000G\E[B
```

Or:

```
\E[4000G\E[e
```

### Analysis

```js
    InputHandler.prototype.cursorCharAbsolute = function (params) {
        var param = params[0];
        if (param < 1) {
            param = 1;
        }
        this._terminal.buffer.x = param - 1;
    };

    InputHandler.prototype.cursorDown = function (params) {
        var param = params[0];
        if (param < 1) {
            param = 1;
        }
        this._terminal.buffer.y += param;
        if (this._terminal.buffer.y >= this._terminal.rows) {
            this._terminal.buffer.y = this._terminal.rows - 1;
        }
        if (this._terminal.buffer.x >= this._terminal.cols) {
            this._terminal.buffer.x--;
        }
    };

    InputHandler.prototype.vPositionRelative = function (params) {
        var param = params[0];
        if (param < 1) {
            param = 1;
        }
        this._terminal.buffer.y += param;
        if (this._terminal.buffer.y >= this._terminal.rows) {
            this._terminal.buffer.y = this._terminal.rows - 1;
        }
        if (this._terminal.buffer.x >= this._terminal.cols) {
            this._terminal.buffer.x--;
        }
    };
```

Not quite sure what the purpose of the `x--` is.  If the goal is to always cap it within the max columns, then that won't necessarily work properly here.

On an unrelated note, `cursorDown` and `vPositionRelative` contain identical code.  `cursorForward` and `hPositionRelative` also contain identical code.  Also, `cursorPosition` and `hVPosition` appear to have logically identical code but `hVPosition` is shorter and more readable, IMO.

jerch:
This PR addresses edge case handling for `InputHandler` methods. Currently many methods are still slightly off compared to xterm's behavior.

This PR shall enable passing most of the escape sequence files. Shall also fix #1434 and fix #2020. Shall give xterm.js much better results in `vttest`.

Things to rework:
- [x] cursor positioning (see #1434 and #2020)
      Other than stated in #1434 I hope to get away with fixing cursor position issues on individual methods. Basic cursor positioning is done now with range checks. Closes #2020.
- [x] Fixed REP: REP itself should also reset previous char
- [x] tab positioning (~~all tab sequence files are failing~~)
      All except CBT working.
- [ ] CBT still not working: This is also related to margin/wrap handling and prolly fixable by a later PR. (Background: cursor restrictions are not correctly applied after RI + IND)
- [x] cursor position sanity checks for:
    - [x] EL
    - [x] ED
    - [x] DCH, ECH, ICH
- [x] wrapping after escape code
- [ ] rework DECSTBM (~~all tests failing~~)
      We are slightly off in the margin handling for `print`, `lineFeed` and `Terminal.scroll`. Changes to those are more involved and better done with isolated PRs.


jerch:
Gonna stop on this PR as it is already quite big.

Remaining issues:
- DECSTBM: We have margin handling issues in `InputHandler.print`, `InputHandler.lineFeed` and `Terminal.scroll`. Any changes to those are more involved, thus should be addressed in separate PRs.
- CBT: Only found one issue with it - cursor restrictions are not applied correctly after RI + IND. This particular issue is not fixable before dealing with the margin/wrap issues seen in DECSTBM. All other conditions seem to work correctly.
- DECOM: We did not respect origin mode at all, this PR does and passes vttest in this regard. Still the edgecases are not worked out yet. To be done with a separate PR.
- cursor handling: As pointed out by #2290, we should unify the cursor stuff into better maintainable code.

Up for review.

fixtures/escape_sequence_files/run_tests.py
```diff
@@ -0,0 +1,73 @@
...
+                with open(os.path.join(os.path.dirname(testfile), 'output',
+                                       os.path.basename(testfile)), 'w') as t_out:
+                    t_out.write(terminal_output)
+            else:
+                passed += 1
+                log(True, '\x1b[32mpass\x1b[0m')
+    return count, passed
+
+
+if __name__ == '__main__':
```
Tyriar: New language 😱, I'm guessing termios is really good for this stuff? Could this replace Terminal2.test.ts as that file is the only thing blocking us from updating node-pty.

fixtures/escape_sequence_files/t0074-DECSTBM_SU_SD.text
```diff
@@ -1,24 +1,25 @@
 a
```
Tyriar: How come there were diffs in these files? Did we have them wrong?

src/Terminal.test.ts
```diff
@@ -750,10 +751,14 @@ describe('Terminal', () => {
       for (let i = 0xDC00; i <= 0xDCFF; ++i) {
         term.buffer.x = term.cols - 1;
         term.wraparoundMode = false;
+        const width = wcwidth((0xD800 - 0xD800) * 0x400 + i - 0xDC00 + 0x10000);
+        if (width !== 1) {
+          continue;
```
Tyriar: Why continue?

src/InputHandler.ts
```diff
@@ -2086,4 +2149,32 @@ export class InputHandler extends Disposable implements IInputHandler {
...
     this._terminal.setgLevel(level);  // TODO: save to move from terminal?
   }
+
+  /**
+   * ESC # 8
+   *   DEC mnemonic: DECALN (https://vt100.net/docs/vt510-rm/DECALN.html)
+   *   This control function fills the complete screen area with
+   *   a test pattern (E) used for adjusting screen alignment.
+   *
+   * TODO: move DECALN into compat addon
```
Tyriar: Should we add this at all if you're going to remove it?

src/InputHandler.ts
```diff
@@ -576,149 +581,228 @@ export class InputHandler extends Disposable implements IInputHandler {
...
-      this._terminal.buffer.getNullCell(this._terminal.eraseAttrData())
-    );
-    this._terminal.updateRange(this._terminal.buffer.y);
+
+  /**
+   * Restrict cursor to viewport size / scroll margin (origin mode).
+   */
+  private _restrictCursor(): void {
+    // cols
+    if (this._terminal.buffer.x < 0) {
```
Tyriar: FYI I've found `Math.min(Math.max(..)..)` to be both faster perf-wise and simpler for these sorts of things

src/InputHandler.ts
```diff
@@ -576,149 +581,228 @@ export class InputHandler extends Disposable implements IInputHandler {
   }
 
   /**
-   * CSI Ps @
-   * Insert Ps (Blank) Character(s) (default = 1) (ICH).
+   * Cursor movements.
```
Tyriar: We have an issue so we can remove this

fixtures/escape_sequence_files/run_tests.py
jerch: Ah I just copied the python file over to regenerate the .text output from xterm. To run it in xterm just do `python run_tests.py` and skip through the tests with Ctrl-D.
The test coverage stuff in the file comes from my old terminal, which also used this to run the sequnces files (with a custom sequence that reports back the terminal state to the python process). I was simply to lazy to clean it up.

In theory we could run the sequence files like that and remove Terminal2.test.ts, but not sure if we want to introduce a custom sequence to run it like that.

fixtures/escape_sequence_files/t0074-DECSTBM_SU_SD.text
jerch: Yes some were slightly off because xterm.js resets to `wrapAround=true` and we feed the files through a pty, which replaces every '\n' with a '\r\n' (thus we dont see the staggered char lines in the outputs, but left aligned chars).

src/Terminal.test.ts
jerch: `print` still skips wide chars at the end of a line. Still need to figure out, what to do with those at the last cell in line. I had to rrewrite the test, because it was testing the wrong thing - before it assumed no content can be written to a line at EOL, but thats not true, instead the char should go into the last cell.

src/InputHandler.ts
jerch: Well I dont mind, this way its a bit simpler by just moving it :smile_cat:

src/InputHandler.ts
jerch: Oh really? I'd expect a function call being magnitudes slower (>5 times) :thinking:

src/InputHandler.ts
jerch: Ok.

src/InputHandler.ts
jerch: Changed it to max/min. Regarding speed I found two different micro benchmarks with opposite results. Since micro benchmarks are not reliable in JS I go with the better readable version.

fixtures/escape_sequence_files/run_tests.py
jerch: Removed the native access into node-pty from the test file. Cannot get a puppeteer version reliably working without doing weird master/slave waiting conditions, thus I went with `pty.open`.

src/InputHandler.ts
jerch: I'd say lets keep it for now and move it once we are settled with the new service structure.

src/Terminal2.test.ts
```diff
@@ -1,45 +1,127 @@
...
 
-class TestTerminal extends Terminal {
-  innerWrite(): void { this._innerWrite(); }
+// all test files expect terminal in 80x25
+const COLS = 80;
+const ROWS = 25;
+
+const TESTFILES = glob.sync('**/escape_sequence_files/*.in', { cwd: path.join(__dirname, '..')});
+const SKIP_FILES = [
+  // 't0008-BS.in',
```
Tyriar: Can we remove all these commented ones?

src/Terminal2.test.ts
```diff
@@ -1,45 +1,127 @@
...
-    const b = new Buffer(64000);
-    const bytes = fs.readSync(primitivePty.master, b, 0, 64000, null);
-    cb(b.toString('utf8', 0, bytes));
+describe('Escape Sequence Files', function(): void {
+  this.timeout(20000);
+
+  let ptyTerm: any = null;
+  let slaveEnd: any = null;
+  let term: Terminal;
+  let customHandler: any = null;
```
Tyriar: `: IDisposable | undefined`?

src/Terminal2.test.ts
```diff
@@ -1,45 +1,127 @@
...
-function ptyWriteRead(data: string, cb: (result: string) => void): void {
-  fs.writeSync(primitivePty.slave, data);
-  setTimeout(() => {
-    const b = new Buffer(64000);
-    const bytes = fs.readSync(primitivePty.master, b, 0, 64000, null);
-    cb(b.toString('utf8', 0, bytes));
+describe('Escape Sequence Files', function(): void {
+  this.timeout(20000);
+
+  let ptyTerm: any = null;
```
Tyriar: Prefer undefined over null

src/Terminal2.test.ts
```diff
@@ -1,45 +1,127 @@
...
+        customHandler.dispose();
+      }
+      slaveEnd.write('\r\n');
+      term.reset();
+      slaveEnd.write('\x1bc\x1b[H');
+
+      // register handler to trigger viewport scraping, wait for it to finish
+      let content = '';
+      await new Promise(resolve => {
+        customHandler = term.addOscHandler(12345, () => {
```
Tyriar: Let's put 12345 into a named const

src/InputHandler.ts
```diff
@@ -1467,13 +1467,18 @@ export class InputHandler extends Disposable implements IInputHandler {
           this._coreService.decPrivateModes.applicationCursorKeys = false;
           break;
         case 3:
+          // TODO: move DECCOLM into compat addon
```
Tyriar: We need an issue for this?

src/Terminal2.test.ts
jerch: Done.

src/Terminal2.test.ts
jerch: Done.

src/InputHandler.ts
jerch: Yeah.

jerch: Yeah gonna add issues for the things listed above.

PR #2294, opened 2019-07-08
Don't handle keyup that is handled by custom handler

### LINKED ISSUE #2293: keyup is handled when the custom key event handler returns false
This causes any `keyup` event to trigger focus which may not be desirable.

Tyriar:
Fixes #2293

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -1613,7 +1613,7 @@ export class Terminal extends Disposable implements ITerminal, IDisposable, IInp
     if (!wasModifierKeyOnlyEvent(ev)) {
       this.focus();
     }
-    
+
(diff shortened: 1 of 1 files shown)


src/Terminal.ts
```diff
@@ -512,29 +512,7 @@ export class Terminal extends Disposable implements ITerminal, IDisposable, IInp
    * Apply key handling to the terminal
    */
   private _bindKeys(): void {
-    const self = this;
-    this.register(addDisposableDomListener(this.element, 'keydown', function (ev: KeyboardEvent): void {
-      if (document.activeElement !== this) {
```
Tyriar: This will always return so we can safety remove this.

Issue #1962, opened 2019-03-07
X10 mouse protocol is capped at 127


jerch:
Currently the X10 mouse coord handling is capped at 127, I guess this is due to UTF8 problems with higher values. Still the protocol officially encodes up to position 223 (charCode 255) in a binary fashion. 

To support those non UTF8 values we need to find a way to get the data to the backend (either by letting non valid UTF8 sequences passing through or by wrapping the binary data into some other encoding, e.g. base64).

jocutajar:
Trying X10 mouse tracking with UTF-8 extended support in VS Code, I get a very confusing result. 

Reproduce:
1. Switch on SET_EXT_MODE_MOUSE (1005): \x1b[?1005h.
2. Press and hold ctrl and left button and drag to the top left corner.
3. Capture the mouse tracking sequence.

Actual result:
CSI M (195) (130) (194) (128) ! !

Expected result (as in xterm):
CSI M P ! !

Why is it a problem?
Mouse tracking is broken. Basic mouse events are OK but anything more advanced is an oops... Try also Ctrl + wheel up/down.

Insights:
1. The terminal does not need to switch to UTF-8 as the mouse event parameters can be expressed in plain ASCII
2. Even if UTF-8 switch was in order, the functions expects 3 UTF-8 characters, 4 were given. It makes no sense to me. What are the two utf-8 characters supposed to represent?
3. There are other cases where mouse events are way off the mark. A complete mouse tracking code review would be in order.

jerch:
@jocutajar We currently have no tests for the different mouse protocols, its likely that the other protocol extensions are broken to some degree as well. I think this went unnoticed due to the fact that only a few apps use the mouse at all and some protocol versions are very uncommon.
Writing some tests based on other terminals would be a good starting point to get a better coverage rolling. Also the current code is hard to read and prolly hard to fix, feel free to come up with a better solution.

egmontkob:
@jocutajar SET_EXT_MODE_MOUSE (1005) is a badly designed extension, every application should be using 1006 instead.

jerch:
@jocutajar Actually I cannot repro the weird additional bytes in UTF8 mode in the demo (the encoding itself works fine). Did you test this with winpty or conpty under windows by any chance? This pretty much looks like some layer in between layer did additional faulty encoding on top.
We still have alot to fix in mouse field, see #2316.

[...]

PR #2295, opened 2019-07-09
flow control and discard watermark

jerch:
Second attempt to solve the flow control issues. Other than the first attempt this relies on ENQ --> ACK cycles:
- server sends request (ENQ) every n-th bytes
- xterm.js responds with answerbackString

xterm.js handles the ENQ request in-band with the escape sequence parser, thus a response guarantees that previous data already got processed. With this the server side can implement strategies to measure data throughput and can pause the pty if needed.

The demo server implements a simple pending ACK strategy: Every n-th data bytes it sends an ENQ and peeks for ACK responses. If the pending requests hit a limit the pty gets paused.

Note: The request message is hardcoded to ENQ in xterm.js, the response message can be cumstomized by the option `answerbackString` (as suggested by xterm docs).

Again this PR comes with a discard parachute to avoid crashing the browser. It is hardcoded to ~50MB.


src/Terminal.ts
```diff
@@ -1211,17 +1217,16 @@ export class Terminal extends Disposable implements ITerminal, IDisposable, IInp
...
-      // XOFF - stop pty pipe
-      // XON will be triggered by emulator before processing data chunk
-      this._coreService.triggerDataEvent(C0.DC3);
-      this._xoffSentToCatchUp = true;
+    // safety measure: dont allow the backend to crash
+    // the terminal by writing to much data to fast.
+    // If we hit this, the terminal cant keep up with data written
+    // and will start to degenerate.
+    if (this._writeBuffersPendingSize > DISCARD_WATERMARK) {
+      throw new Error('write data discarded, use flow control to avoid losing data');
```
mofux: Logging a warning and then circulating (shift & push) the write buffer might be a more sensible solution here? This may still lead to corruption, but it's also very likely that only the last couple of bytes of the write buffer will end up in terminal & scrollback anyway. TBH I'm not sure how a customer should handle this situation gracefully otherwise (except implementing the flowControl mechanism)

jerch: Not sure if I understand what you want to shift & push here. Can you elaborate on that?

Basically this code ensures that the writebuffers never will grow beyond ~50MB. Note that if the terminal reaches this state its already lost long before. A sane limit for responsiveness is around ~ 500kB - 1MB, here we are already 50 times higher. Thus this limit is a last security resort not to crash the browser.

A customer cannot deal with "to much data per time" at all. Its the implementor's responsibility to limit the throughput to sane values, since he is the only one controlling the server side.

mofux: I agree, it's just - if we throw an error here it is very likely that the consumer of our lib does not catch it (because he does not expect that `terminal.write` and `terminal.writeUtf8` can throw). And we don't really expect him to catch that error either (because - what is he supposed to do?). 

Therefore my idea was to instead of throwing an error, we could log a warning to the console, and then to perform the least destructive action - which would be to append the new data to our write buffer (push), but at the same moment to remove the oldest writes from the write buffer (shift). So we can still keep the buffer at a constant 50MB. 

Let's say somebody is crazy enough to run the `yes` command, it will fill up the 50MB buffer with `y`s. Once he cancels that action with `Ctrl+C`, the next data chunk that arrives is likely the prompt line -- but since we are discarding any data while our buffer is in the limit, the prompt line will go lost. It would be better if we'd discard the `y` sequences at the beginning of the write buffer, and still allow the most recent writes to be written out the terminal.

On the other hand I agree that this is an extreme scenario and engineering such a solution might just be too much effort 😅

jerch: Yes that def to much to ask for :smiling_imp:, just kidding - we can do that.

Still I see not much value in it, the terminal state is likely to break whenever stream data gets discarded without proper state handling, we simply dont know what we discard, let it be an sequence introducer or finalizer - hell might break loose. We could try to discard more clever by degrading the sequence actions (like removing print handler and such, would basically skip screen updates) - but for what purpose, the output is still broken (even worse, it might be incomplete). No matter what we do, we can never compete with really fast producers which can pump data at rates of GBs/s while we sit at 35MB/s (with screen updates by the DOM renderer we even drop back to ~5MB/s). Flow control with the ability to pause the producer is the only way to fix this without losing data. Imho flow control is a must-have (thats why I am insisting on it lol).

Back to your idea: Yes the chances are higher, that the stream "self-healed" the state with your idea and we still can present the user some valid late content. But that comes with a big MAYBE. I kinda think we should not try to do this, maybe we should increase the limit further (as RAM is no big deal these days) and do a hard stop with a blocking alert once that limit was hit (terminal is not working anymore)?

mofux: Yeah let's keep it like you've implemented it for now, we'll see how the limits fit once it's out in the wild 👍

demo/client.ts
```diff
@@ -236,7 +236,8 @@ function initOptions(term: TerminalType): void {
     fontWeight: ['normal', 'bold', '100', '200', '300', '400', '500', '600', '700', '800', '900'],
     fontWeightBold: ['normal', 'bold', '100', '200', '300', '400', '500', '600', '700', '800', '900'],
     rendererType: ['dom', 'canvas'],
-    wordSeparator: null
+    wordSeparator: null,
+    answerbackString: null
```
Tyriar: This should be `''`?

src/InputHandler.ts
```diff
@@ -495,6 +498,14 @@ export class InputHandler extends Disposable implements IInputHandler {
     return this._parser.addOscHandler(ident, callback);
   }
 
+  /**
+   * ENQ
+   * Enquiry (Ctrl-E).
+   */
+  public enquiry(): void {
+    this._coreService.triggerDataEvent(this._terminal.options.answerbackString);
```
Tyriar: Can we enable this by default and wrap this line in an `if (this._terminal.options.answerbackString.length > 0)`?

demo/server.js
```diff
@@ -74,59 +88,97 @@ function startServer() {
...
+          bytesSent = 0;
+          if (ackPending > MAX_PENDING_ACK) {
+            term.pause();
+          }
+        }
       }
     });
     ws.on('message', function(msg) {
+      if (USE_FLOW_CONTROL && msg === FLOW_CONTROL_ACK_RESPONSE) {
+        ackPending--;
```
Tyriar: I think we need to `Math.max(..., 0)` this just in case programs send `ENQ`

demo/client.ts
Tyriar: Oh this is the option UI stuff, ignore this.

demo/client.ts
```diff
@@ -224,7 +224,7 @@ function initOptions(term: TerminalType): void {
     'handler',
     'screenKeys',
     'termName',
-    'useFlowControl',
+    // 'useFlowControl',
```
Tyriar: Let's remove useFlowControl

src/common/services/OptionsService.ts
```diff
@@ -47,7 +47,8 @@ export const DEFAULT_OPTIONS: ITerminalOptions = Object.freeze({
   debug: false,
   cancelEvents: false,
   useFlowControl: false,
-  wordSeparator: ' ()[]{}\'"'
+  wordSeparator: ' ()[]{}\'"',
+  answerbackString: '\x06\x06\x06\x06'
```
Tyriar: I think this needs to be added to the internal `ITerminalOptions` and API `ITerminalOptions`

demo/client.ts
jerch: Done.

src/common/services/OptionsService.ts
jerch: Done.

demo/client.ts
Tyriar: Missed this line

jerch: Grepped only over `./src` lol.

demo/server.js
```diff
@@ -9,13 +9,31 @@ var pty = require('node-pty');
  */
 const USE_BINARY_UTF8 = false;
 
+/**
+ * Whether to use flow control.
+ * This must be in sync with answerbackString in xterm.js.
```
Tyriar: We should add a `USE_FLOW_CONTROL` const to client.ts and init `Terminal` with:

```
answerbackString: USE_FLOW_CONTROL ? '\x06\x06\x06\x06' : '',
```

jerch: Good idea.

jerch:
Did the state tests for ENQ and well, we cannot rely on a C0 byte for this as they are typically ignored / valid payload for the subparser states OSC, DCS, APC, PM and SOS.

To keep going with a simple one byte in-band variant we would have to go with an unused C1 byte. Hottest candidates for this are PU1 or PU2 (meant for private usage).

Alternatively we could go with a totally custom sequence (more bytes allowed), but would have to patch the parser to carry forward the previous state to be re-applied afterwards.

jerch:
@Tyriar Done with the basic rewrite. What we have now:
- optional callback on `write`, `writeUtf8` and `writeln`
- callback get triggered right after the chunk was parsed, implications: chunk is guaranteed to be applied to offscreen parts of the terminal (parsing is sync code down to the buffer), still this does not apply to screen state, which gets not updated before the next event loop invocation

With these changes the inner write states can be tracked. Now its easy to decide whether the terminal is under high data pressure and falling behind, thus flow control can be implemented.

The demo implements flow control as following:
- attach addon:
  - tracks incoming data size
  - places every n-th bytes a callback into the write chain (customizable, 0 for disabled)
  - callback sends ACK reply back to the server
- server,js:
  - tracks outgoing data size
  - expects an ACK reply for every n-th byte (must be in sync with attach addon setting!)
  - does high/low watermark on pending ACKs (pause pty at high, resume pty at low watermark)

To avoid problems on implementors' side the flow control setting in the attach addon is disabled by default.

New Protocol:
ACKs are sent in-band over the websocket to avoid latency problems with a second (TCP) connection. To achieve this a tiny protocol was built on top to allow difference message types (this is needed since the data stream can contain arbitrary bytes). I think this protocol will also come handy for other stuff later on, like in-band resize updates and raw mouse reports (we have problems with the mousey for non utf8 reports, see #1962).

About the ENQ:
Not sure what to do with this, imho its ancient feature we shortly revived for the wrong purpose. Still we get one more point in vttest rating with it - maybe move to the yet to come compat addon?

Well slowly getting there, up for a review.

jerch:
@Tyriar Looked further into the UTF8 poblem with abitrary data and come to the conclusion that we should give up on string based transport to avoid the utf8 conversions. Thus I created a binary version of the protocol.
I think we need to further discuss how to layout this, removing string transport from the attach addon might break with many installations. Still I see more reasons to drop it.

jerch:
@Tyriar:
Should we close this PR in favor of #2362 and a separate attach addon PR later on?

Tyriar:
Yes let's close this, we also need to remove some stuff in node-pty since our thinking changed right?

PR #2298, opened 2019-07-10
add SerializeAddon

JavaCS3:
This PR is refer to #2213 
Color and Wrap mode are *NOT* supported in `serialize` function. Will continue improve.

Part of #595


addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -0,0 +1,42 @@
...
+      if (!last) {
+        data += '\r\n';
+      }
+    }
+
+    return data;
+  }
+
+  public dispose(): void {
+    if (this._terminal !== undefined) { }
```
Tyriar: You can remove this line

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  constructor() { }
+
+  public activate(terminal: Terminal): void {
+    this._terminal = terminal;
+  }
+
+  public serialize(rows?: number): string {
+    if (!this._terminal) {
+      return '';
```
Tyriar: Should this throw instead of silently failing?

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+    const buffer = this._terminal.buffer;
+    const length = Math.max(0, Math.min((rows === undefined ? buffer.length : rows), buffer.length));
+    let data = '';
```
Tyriar: Constructing a `string[]` and calling `join('')` at the end should be faster here, important considering we could be dealing with thousands of lines.

addons/xterm-addon-serialize/typings/xterm-addon-serialize.d.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    constructor();
+
+    /**
+     * Activates the addon
+     * @param terminal The terminal the addon is being loaded in.
+     */
+    public activate(terminal: Terminal): void;
+
+    public serialize(rows?: number): string;
```
Tyriar: We need some docs here

package.json
```diff
@@ -15,6 +15,7 @@
     "test": "npm run test-unit",
     "posttest": "npm run lint",
     "test-api": "mocha \"**/*.api.js\"",
+    "test-addons": "mocha \"**/*.test.js\"",
```
Tyriar: I think you left this in accidentally

addons/xterm-addon-serialize/src/SerializeAddon.ts
JavaCS3: ok

addons/xterm-addon-serialize/src/SerializeAddon.ts
JavaCS3: We can do a benchmark to see which one has better performance.

addons/xterm-addon-serialize/typings/xterm-addon-serialize.d.ts
JavaCS3: ok

package.json
JavaCS3: It is accidentally left here.
 but I found there's no such a command to run unit test for addons folder. https://github.com/xtermjs/xterm.js/blob/master/bin/test.js will only run test under `out` folder. 
So how about this
```suggestion
    "test-addons": "mocha \"addons/**/*.test.js\"",
```

Tyriar: If addons ever add unit tests (vs api tests) we can add the glob to the array in https://github.com/xtermjs/xterm.js/blob/master/bin/test.js, that way they would get run as part of `yarn test`

addons/xterm-addon-serialize/src/SerializeAddon.ts
Tyriar: I actually did the other day, took about 33% of the time using the join approach based on micro benchmarks I did for https://github.com/Tyriar/hangul-romanization/pull/5

package.json
JavaCS3: Agree

addons/xterm-addon-serialize/src/SerializeAddon.ts
JavaCS3: That's unbelievable

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -0,0 +1,35 @@
...
+    this._terminal = terminal;
+  }
+
+  public serialize(rows?: number): string {
+    // TODO: Add frontground/background color support later
+    if (!this._terminal) {
+      throw new Error('No terminal found!');
+    }
+    const buffer = this._terminal.buffer;
+    const length = Math.max(0, Math.min((rows === undefined ? buffer.length : rows), buffer.length));
```
Tyriar: I think the rows should actually start from the _bottom_ of the terminal, so passing in `Terminal.rows` here (which could be the default) would only serialize the viewport.

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -0,0 +1,35 @@
...
+  constructor() { }
+
+  public activate(terminal: Terminal): void {
+    this._terminal = terminal;
+  }
+
+  public serialize(rows?: number): string {
+    // TODO: Add frontground/background color support later
+    if (!this._terminal) {
+      throw new Error('No terminal found!');
```
Tyriar: This is the error we use in other addons:

```
throw new Error('Cannot use addon until it has been loaded');
```

addons/xterm-addon-serialize/typings/xterm-addon-serialize.d.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Activates the addon
+     * @param terminal The terminal the addon is being loaded in.
+     */
+    public activate(terminal: Terminal): void;
+
+    /**
+     * Serializes terminal rows into a string that can be written back to the terminal
+     * to restore the state. The cursor will also be positioned to the correct cell.
```
Tyriar: I think we should add an explicit cursor move sequence (`\x1b[y;xH`) if it's not where we expect it, otherwise this statement is false.

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+    const buffer = this._terminal.buffer;
+    const length = Math.max(0, Math.min((rows === undefined ? buffer.length : rows), buffer.length));
+    const lines: string[] = new Array<string>(length);
+
+    for (let i = 0; i < length; i++) {
+      const line = buffer.getLine(i);
+      lines[i] = line ? line.translateToString() : '';
```
jerch: If the goal of the addon is to get the closest possible match to be replayed on other terminals it should also serialize/restore gaps in lines (empty cells up to trim border). Those can happen if the cursor jumped over cells (e.g. some cursor command or TAB). Note that `translateToString` fills those with whitespaces to get a close string representation, that does not rely on terminal output (cannot use cursor commands for that).

To also grab those gaps imho its better to walk over all cells and check explicitly for content. If a real gap was found (watch out for empty cells after wide chars, those dont count here), the cursor should be moved forward. Doing this will ensure that the content is really the same after serialize/restore (beside the attrs that are not exported yet).

But as always, the devil is in the details. For this to work reliably, serialize would also have to export the viewport grid size, and a later restore action would have to take this into account.

addons/xterm-addon-serialize/typings/xterm-addon-serialize.d.ts
jerch: Same as with my note above - absolute positioning of the cursor would have to be corrected later during restore, if the target terminal has other rows X rows width, thus serialize needs to export the viewport size to not lose this important bit of information.

addons/xterm-addon-serialize/src/SerializeAddon.ts
JavaCS3: My goal is to write a terminal session player which is the same as [asciinema](https://asciinema.org) but with audio support. examples: https://asciinema.org/a/138296

But asciinema is written in ClosureScript that I believe rare people is familiar with. So I have to implement it in my own way. 

To answer your question, Yes, I need to get the closest possible match to the original terminal session

jerch: Ah ok, then I think you cannot go with `translateToString` directly and need a more advanced buffer scraping method instead. Also there is no simple replacement character to keep the null cells in the output (you cannot use NUL as it will be skipped by emulators as stream padding byte).

Hmm, do you want the data to be re-applicable to other targets (other emulators beside xterm.js)? If so, then the serializer would have to know about the features of that target. Note that there are dumb terminals, that cannot deal with cursor movements at all and would map CUF to whitespace (basically `translateToString` assumes this dumb control char capabilities). Same with text attrs once you expose them - not all targets understand them.

So if you want to support those "other targets", imho serialize needs options to be eval'ed during export:
```typescript
public serialize(rows?: number, opts?: ExportOptions): string {
  ...
  if (!cell.hasContent) {  // empty cell
    for (let i = 0; i < cell.width; ++i) {
      output += opts.CUF;  // ' ' for dumb output, '\x1b[C' for terminals supporting it
  }
  ...
}
```
Not sure if we expose enough buffer state yet in API to let you do it this fine grained.

jerch: Edit: A different approach would be to serialize as much as possible information into a not directly applicable format, and do the features switch during restore. This has one advantage: you still have all the needed bits of information can apply the same data to HTML output, dumb output, xterm.js output (to be fed back). The major disadvantage is the fact, that this will need a restore component, while doing it during serialize would not.

JavaCS3: IMO xterm.js is enough for me and I don't need to support some special terminal emulator currently. All I want is to "copy" all the features that asciinema already have by Typescript.

Another thing is asciinema is based on VT terminal (see https://github.com/asciinema/vt), I believe it's the same as xterm.js

And I count this commit as a beginning, it doesn't need to cover all that cases for now. Continuously add features steps by steps is a good choice for me.

This commit based from the discussion of https://github.com/xtermjs/xterm.js/pull/2213#issuecomment-500233758

jerch: Yeah lets not over complicate things. If you only want to go with xterm.js compatibility, you can take any escape sequence in the final output thats listed in InputHandler. (We dont have a list of supported sequences yet.)

JavaCS3:
Can I merge to master now?

mofux:
Hi @JavaCS3, thanks for your contribution, we think your code quality is really good and we appreciate your effort! 

We've just had an internal discussion within the xterm.js maintainer team, and we came to the conclusion that this addon would be better off living in a separate repository outside the responsibility of our core maintainer team.

We feel that at the current state, this addon is very specific to a certain use-case (extracting plain text from the buffer), and it won't benefit many users.

I've got some ideas on how you can achieve #2213 with a different technique, so let's continue the discussion over there.

Tyriar:
@JavaCS3 we're still discussing, I think we do want this. Hold tight 😃 

I suggest checking out @mofux's suggestion though in https://github.com/xtermjs/xterm.js/pull/2213#issuecomment-511902629 as that's probably a better solution to your problem in the end.

JavaCS3:
Any update?

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -0,0 +1,44 @@
...
+  public activate(terminal: Terminal): void {
+    this._terminal = terminal;
+  }
+
+  public serialize(rows?: number): string {
+    // TODO: Add frontground/background color support later
+    if (!this._terminal) {
+      throw new Error('Cannot use addon until it has been loaded');
+    }
+    const terminalRows = this._terminal.rows;
```
MaxRis: Does it make sens to serialize whole buffer, but not only visible lines? ( `const terminalRows = this._terminal.buffer.length;` )

JavaCS3: Yes, It used to be like that.
But after I took @Tyriar 's [suggestion](https://github.com/xtermjs/xterm.js/pull/2298#discussion_r303037652) into account, I changed to this.

Tyriar: Doing the whole buffer could be pretty expensive so opting into more than the minimum is the thinking here.

Tyriar:
@mofux @jerch where do you guys stand on this now? Some form of serialization is needed when we want to retain processes across reloads in VS Code (https://github.com/microsoft/vscode/issues/20013). I think it's a fine compromise to only serialize the active single buffer.

For serializing terminal modes we could add APIs down the track to do that, for example `Terminal.modes` or something and then this extension could read that is toggle all the relevant modes

jerch:
@Tyriar Well up to you, I still think that we dont need an addon in the repo base that does not yet add any real value. The PR still feels like a stub to me with much more things to come.

The serializing idea is promising, if we could get more relevant data serialized. Thus I was hoping the discussion would go into this direction with API extensions and implementation here.

Well we can do that in the future and add real value to this code later on. So again, up to you.

Tyriar:
Ok I've merged this into `feat/serialize-addon` for the time being, @JavaCS3 if you need this to unblock you I suggest to publish the package yourself under an ID until it's more fleshed out.

JavaCS3:
@Tyriar Thanks for your understanding. I almost want to give up. Is there any existing APIs can I use to capture buffers color?

Tyriar:
@JavaCS3 not yet, we just need to agree on the shape of them and add them though.

MaxRis:
Hey, thank you @JavaCS3 for your work on serialization plugin and Xterm.JS's team for addons feature. [Fluent](https://github.com/felixse/fluentterminal) already started usage of buffer serialization to allow tabs drag and drop (new xterm.js instance should be created and restored to the state before user started tab drag).
What is missed now for the best users experience, it's possibility to serialize and restore colors for each cell.

JavaCS3:
@MaxRis We can progressively improve it!

ZedYeung:
how to deserialize? `term.write('what is serialize')`?

Tyriar:
@ZedYeung yes deserializing is just writing the serialized result to a new terminal. The idea was to leverage the existing data format and our optimized parser (`write`) instead of coming up with a new format and a bunch of new APIs

PR #2304, opened 2019-07-13
Use IBufferService for buffer access in InputHandler

### LINKED ISSUE #1560: Make "debug mode" public API

Tyriar:
Also adds the new `logLevel` option.

Part of #1507 
Fixes #1560

@jerch a quick once over of the non-test files would be good 😃

### DIFF
--- a/addons/xterm-addon-webgl/src/WebglUtils.ts
+++ b/addons/xterm-addon-webgl/src/WebglUtils.ts
@@ -24,7 +24,7 @@ export function createProgram(gl: WebGLRenderingContext, vertexSource: string, f
     return program;
   }
 
-  console.log(gl.getProgramInfoLog(program));
+  console.error(gl.getProgramInfoLog(program));
   gl.deleteProgram(program);
 }
 
@@ -37,7 +37,7 @@ export function createShader(gl: WebGLRenderingContext, type: number, source: st
     return shader;
   }
 
-  console.log(gl.getShaderInfoLog(shader));
+  console.error(gl.getShaderInfoLog(shader));
   gl.deleteShader(shader);
 }
 
--- a/src/InputHandler.ts
+++ b/src/InputHandler.ts
@@ -19,7 +19,7 @@ import { NULL_CELL_CODE, NULL_CELL_WIDTH, Attributes, FgFlags, BgFlags, Content 
 import { CellData } from 'common/buffer/CellData';
 import { AttributeData } from 'common/buffer/AttributeData';
 import { IAttributeData, IDisposable } from 'common/Types';
-import { ICoreService } from 'common/services/Services';
+import { ICoreService, IBufferService } from 'common/services/Services';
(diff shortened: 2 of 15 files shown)


src/InputHandler.ts
```diff
@@ -41,7 +41,10 @@ const GLEVEL: {[key: string]: number} = {'(': 0, ')': 1, '*': 2, '+': 3, '-': 1,
 class DECRQSS implements IDcsHandler {
   private _data: Uint32Array = new Uint32Array(0);
 
-  constructor(private _terminal: any) { }
+  constructor(
+    private _terminal: any,
```
jerch: Saw that this comes from me, wonder why I typed it as `any` in the first place. :thinking: 

Imho DECRQSS can be fixed and (terminal removed) in a separate issue later on when we have all the needed services. Also the SGR sub command does not yet do the right thing.

Tyriar:
updateRange should probably live in render service. I'll include cols/rows in this as well as they live on buffer service too.

src/InputHandler.ts
Tyriar: It was actually broken as it was using the now removed handler 😱

Tyriar: I removed terminal usage from it completely, pretty easy

jerch: Lol missing test cases, tsss. Good you spotted it :+1:

Tyriar:
Also added a log service and added the `logLevel ` option to fix https://github.com/xtermjs/xterm.js/issues/1560

addons/xterm-addon-webgl/src/WebglUtils.ts
```diff
@@ -24,7 +24,7 @@ export function createProgram(gl: WebGLRenderingContext, vertexSource: string, f
     return program;
   }
 
-  console.log(gl.getProgramInfoLog(program));
+  console.error(gl.getProgramInfoLog(program));
```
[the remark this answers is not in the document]
Tyriar: Well the WebGL addon can use the internal one but doesn't seem worth the effort. This might be better to throw anyone

src/common/services/LogService.ts
```diff
@@ -0,0 +1,80 @@
...
+        this._updateLogLevel();
+      }
+    });
+  }
+
+  private _updateLogLevel(): void {
+    this._logLevel = optionsKeyToLogLevel[this._optionsService.options.logLevel];
+  }
+
+  debug(message: string, ...optionalParams: any[]): void {
```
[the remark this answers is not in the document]
Tyriar: Since we have `ILogService` now, we could always go there pretty easily if we really wanted.

PR #2315, opened 2019-07-14
Move renderer into browser

Tyriar:
Part of #1507


Tyriar:
I think the reason I was holding off on merging was there are still 3 files left in `src/renderer`, merging this now to avoid conflicts and will handle the other 3 in a later PR.

PR #2321, opened 2019-07-17
Mouse tracking tests

jerch:
PR to have a solid mouse tracking test base.

The tests are a bit degraded atm to test and pass our current implementation. A new implementation should pass at least these tests to not introduce regressions. All tests have extensive comments about bugs and the desired behavior. A new implementation can work with these towards more standard conform behavior.


jerch:
Added test stubs for modifier keys. None of these is working so all is commented out.

test/api/MouseTracking.api.ts
```diff
@@ -0,0 +1,1191 @@
...
+}
+async function mouseDown(button: "left" | "right" | "middle" | undefined): Promise<void> {
+  return await page.mouse.down({button});
+}
+async function mouseUp(button: "left" | "right" | "middle" | undefined): Promise<void> {
+  return await page.mouse.up({button});
+}
+async function wheelUp(): Promise<void> {
+  const self = (page.mouse as any);
+  return await self._client.send('Input.dispatchMouseEvent', {
```
Tyriar: `_client`? You might be better off dispatching an event inside a `page.evaluate` than this.

jerch: I am not sure how to do this, is there a way to fake wheel event from within the browser JS engine?

Tyriar: I think something like this should work:

```ts
await page.evaluate(`
  const e = new WheelEvent(...);
  element.dispatchEvent(e);
`);
```

jerch: Yepp, works.

test/api/MouseTracking.api.ts
```diff
@@ -0,0 +1,1181 @@
...
+    window.term.element.dispatchEvent(new WheelEvent('wheel', {clientX: ${self._x}, clientY: ${self._y}, deltaX: 0, deltaY: 10, modifiers: ${self._keyboard._modifiers}}));
+  `);
+}
+
+// extract button data from buttonCode
+function evalButtonCode(code: number) {
+  // 2 bits: 0 - left, 1 - middle, 2 - right, 3 - release
+  // higher bits: 4 - shift, 8 - meta, 16 - control
+  const modifier = {shift: !!(code & 4), meta: !!(code & 8), control: !!(code & 16)};
+  const wheel = code & 64;
```
textshell: This is not actually a bit for wheel events. Xterm reports wheel events basically using the usual X11 button codes. But X11 only uses 4 buttons as wheel events and further traditional buttons use the same code space.
See https://terminalguide.netlify.com/mouse/ and [the implementation](https://github.com/ThomasDickey/xterm-snapshots/blob/f9c66ec493c2a434d0fb22bcc34bc8a9ea077f0a/button.c#L4750) for how xterm encodes the buttons.
X11 uses the buttons with the following mapping {1: left, 2: middle, 3: right, 4/5: scroll wheel, 6/7 scroll wheel left/right, 8-x: further buttons}
The xterm button encoding starts at 0, but reserves 3 as special "some button released" marker. So the buttons get shuffled to {0: left, 1: middle, 2: right, 4/5: scroll wheel, 6/7 scroll wheel left/right, 8-x: further buttons}
As the encoding of the button started with only 2 bits needed those 2 bits are used normally. The later addition of button numbers now up to 15 while the modifiers already got bits in between resulted in the bits with values 4 and 8 to be transposed up to the bits with values 64 and 128.

So while (code & 64) == 1 && (code & 128) == 0 end up encoding wheel events, (code & 64) == 1 && (code & 128) == 1 encodes button events again.

test/api/MouseTracking.api.ts
```diff
@@ -0,0 +1,1181 @@
...
+    `);
+  });
+
+  describe('DECSET 9 (X10)', async () => {
+    /**
+     * X10 protocol:
+     *  - only press events
+     *  - no wheel? (TBD)
+     *  - no move
+     *  - no modifiers? (TDB)
```
textshell: Yes mode 9 does not report anything else than button presses for left, middle and right buttons.

Basically that mode was frozen in time and does not get updates.

See [xterm source 1](https://github.com/ThomasDickey/xterm-snapshots/blob/f9c66ec493c2a434d0fb22bcc34bc8a9ea077f0a/button.c#L4774), [xterm source 2](https://github.com/ThomasDickey/xterm-snapshots/blob/f9c66ec493c2a434d0fb22bcc34bc8a9ea077f0a/button.c#L4840)

test/api/MouseTracking.api.ts
```diff
@@ -0,0 +1,1181 @@
...
+      // wheel
+      await mouseMove(43, 24);
+      await getReports(encoding); // clear reports
+      await wheelUp();
+      assert.deepEqual(await getReports(encoding), []);
+      await wheelDown();
+      assert.deepEqual(await getReports(encoding), []);
+
+      // modifiers
+      // TODO
```
textshell: As modifiers should not be reported that should also be implemented currently in xterm.js? So this test could already be included.

test/api/MouseTracking.api.ts
```diff
@@ -0,0 +1,1181 @@
...
+    });
+  });
+  describe('DECSET 1000 (VT200 mouse)', () => {
+    /**
+     * VT200 protocol:
+     *  - press and release events
+     *  - wheel up/down
+     *  - no move
+     *  - modifiers? (TBD)
```
textshell: Yes, modifiers should be reported in mode 1000 and up. Modifiers actually supported vary by terminal implementation though. (actually there is no common set, shift is usually not supported (used for regaining access to the selection/clipboard functions)

test/api/MouseTracking.api.ts
```diff
@@ -0,0 +1,1181 @@
...
+      // await page.keyboard.up('Alt');
+      // bug: not working - reports wheel instead
+      // assert.deepEqual(await getReports(encoding), [
+      //   {col: 44, row: 25, state: {action: 'press', button: 'left', modifier: {control: true, shift: false, meta: false}}},
+      //   {col: 44, row: 25, state: {action: 'release', button: '<none>', modifier: {control: true, shift: false, meta: false}}},
+      // ]);
+      // await getReports(encoding); // clear reports
+      // shift
+      // seems shift pressed alone would supress any mouse event - limitation of chrome?
+
```
textshell: possibly a todo for behavior with multiple mouse buttons pressed would be useful.

test/api/MouseTracking.api.ts
```diff
@@ -0,0 +1,1181 @@
...
+  let col: number;
+  // unpack msg
+  const report = String.fromCharCode.apply(null, msg);
+  // console.log([report]);
+  // skip non mouse reports
+  if (!report || report[0] !== '\x1b') {
+    return report;
+  }
+  switch (encoding) {
+    case 'DEFAULT':
```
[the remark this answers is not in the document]
jerch: Yepp I know. Since we cannot create the needed UTF8 violating bytes yet, this is currently not implemented. Gonna add a comment to that.

test/api/MouseTracking.api.ts
jerch: Oh thats kinda different from whats stated in the xterm docs itself. Seems I have to work through your links first. Also we are currently limited to 3 button mouse + wheel (cannot test higher buttons myself, not even sure if browsers agree for those on the buttons reported).

test/api/MouseTracking.api.ts
jerch: Ah ok.

test/api/MouseTracking.api.ts
jerch: Well I did not do in the first place, because I couldnt test the positive case. With the new implementation of #2323 now I can.

test/api/MouseTracking.api.ts
jerch: So from 1000+ we should always report all 3 modifiers? Asking this because our old messy code reported only CTRL for DECSET 1000 by purpose.

test/api/MouseTracking.api.ts
jerch: Yes thats yet to come. With our old code we always reported the last pressed one, the new code does not save that bit of information, thus we do as most others do - report the lowest one pressed. (Emulators vary here much though)

test/api/MouseTracking.api.ts
jerch: For my understanding - what would (code & 64) == 1 && (code & 128) == 1 encode here? Maybe I have a wrong assumption here, for me doing +64 is equivalent to |64, thus the opposite &64 can test for it. Ofc this only works if the lower range does not overflow (create that bit by other means), which seems to be the case?

jerch: Hmm ok seems past button 11 overflow actually happens, thus my |64 == +64 does not hold here.

jerch: @textshell: Changed the buttonCode eval to the reverse of the xterm code. Plz have a look at it again, hope I did not miss anything.

test/api/MouseTracking.api.ts
```diff
@@ -0,0 +1,2021 @@
...
+    if (code >= 128) {
+      button |= 8;
+      code -= 128;
+    }
+    if (code >= 64) {
+      button |= 4;
+      code -= 64;
+    }
+    button += code & 3;
+  }
```
textshell: The special case for `code === 3` is not needed.
I think you could also use something like this:
```
let button = code & 3;
if (code & 128) {
    button |= 8;
}
if (code & 64) {
   button |= 4
}
```
This would ignore bits allocated in the future, but otherwise be very safe. But in a unit test it might be better to reject the sequence if `code > 255` (like using `actionS = 'invalid'`)

I think your code would get wrong results when confronted with code that has further bits set.

test/api/MouseTracking.api.ts
```diff
@@ -0,0 +1,2021 @@
...
+  let actionS = 'press';
+  let buttonS = reverseButtons[button];
+  if (move) {
+    actionS = 'move';
+  } else if (button === 3) {
+    buttonS = '<none>';
+    actionS = 'release';
+  } else if ((button & 4) && !(button & 8)) {
```
textshell: I think directly using the button numbers would be clearer:
```suggestion
  } else if (4 <= button && button <= 7) {
```

For `actionS` the bit-ops could be replaced by simple `button == 4` etc or even simply indexing into `buttons`

test/api/MouseTracking.api.ts
textshell: I've added a few comments to the new commit. For the actually defined range your commit should work, but see my comments for ideas on simplifying it and making it a bit more meaningful with further bits set.

The encoding is a bit historical, but i don't think it could be a lot easier given the history and the strong preference for terminal sequence to use ultra compact encoding (which is still strong even today).

If you are using linux (or a linux VM) you can test with other buttons using the trick in 
https://gitlab.gnome.org/GNOME/vte/issues/93#note_433817

test/api/MouseTracking.api.ts
jerch: @textshell Yeah much cleaner.

Is there a valid way to get here `code > 255`? To me it seems this cannot be reached by normal "button addressing" (no +256 rule for 4th bit)? However put the `code > 255` in just to be safe against malformed stuff.

test/api/MouseTracking.api.ts
jerch: Yepp, simplified it abit.

jerch:
@Tyriar
Phew, MacOS tests work now. Problem was the Shift modifier, for some reason that gets not caught by the selection manager. I think we should unify the escape strategies for default actions across the platforms and maybe make it customizable so vendors can do as they wish. Already scratched that topic here https://github.com/xtermjs/xterm.js/pull/2323#issuecomment-513567188.

test/api/MouseTracking.api.ts
```diff
@@ -332,6 +335,8 @@ describe('Mouse Tracking Tests', function(): void {
 
       // SHIFT
       // note: caught by selection manager
+      // bug? Why not caught by selection manger on macos?
```
Tyriar: `TODO: `?

test/api/MouseTracking.api.ts
```diff
@@ -20,6 +20,9 @@ const fontSize = 6;
 const cols = 260;
 const rows = 50;
 
+// for some reason shift gets not caught by selection manager on macos
```
Tyriar: Hmm, what's shift meant to do? `macOptionClickForcesSelection` is the main mac different thing around mouse that I can think of.

test/api/MouseTracking.api.ts
jerch: Not sure, in the selection manager there is an explicit test against the platform. Seems it was done that way by purpose?

test/api/MouseTracking.api.ts
jerch: Culprit is this code: https://github.com/xtermjs/xterm.js/blob/7db22c24087dede2506d9be6cb5aede3ae3d8288/src/browser/services/SelectionService.ts#L395-L401

For some reason Mac is treated differently.

Tyriar: Oh OK, I think I used another terminal as a basis for that (user expectations for platform).

test/api/MouseTracking.api.ts
textshell: These things are always evolving. So it's always good to think about what should happen if someone happens to allocate a new bit (which is not that hard if you have a good use case). Or in a unit test, if the code under test does something unexpected.

The button code decoding looks good now.

PR #2331, opened 2019-07-23
Setting fontFamily and fontSize priority as important to avoid css overriding issue

hemkaran:
I am having the issue in one of my project, where `xterm` inserts `'W'` character to measure the width and height of character after adding specified (specified in xtermOptions) `font-family`. But my application is overriding that css by adding `font-family` as `!important` in its parent selector.

### DIFF
--- a/src/CharMeasure.ts
+++ b/src/CharMeasure.ts
@@ -40,8 +40,8 @@ export class CharMeasure implements ICharMeasure {
   }
 
   public measure(options: ITerminalOptions): void {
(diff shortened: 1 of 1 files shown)


Tyriar:
Normally in my experience using `!important` is a hack that should be fixed by some other means. If we do this we would need it everywhere in the DOM renderer as well. A better fix it probably to avoid targeting xterm in your CSS using `:not(.xterm)` or something.

PR #2346, opened 2019-07-25
Parser hooks

### LINKED ISSUE #1823: cleanup/enhancement of parser

### LINKED ISSUE #2332: Rework/sanitize parser hooks
Currently we expose custom hooks to CSI and OSC commands. Idea is to extend this to ESC and DCS commands as well. Furthermore the current interface needs some overhaul to land as stable in v4.

TODO:
- implement `addEscHandler` for public usage
- implement `addDcsHandler` for public usage
- review `addCsiHandler`, remove "experimental" badge
- review `addOscHndler`, remove "experimental" badge

### LINKED ISSUE #2342: error in OSC subparsing
We have an error in OSC identifier parsing, the following sequence is sent by `reset`:
```
OSC 104 BEL
```

Currently our identifier parsing searches for the first ';' as delimiter of the identifier and the payload, thus the sequence above leads to an error message of an unknown OSC command with identifier -1.

jerch:
PR to get the parser actions hookable from outside. Shall fix #2332.

TODO:
- [x] review OSC handling and hook
       Also introduces `OscParser` which simplifies `EscapeSequenceParser` and fixes #2342. OSC handlers now can be attached with full state introspection (START, PUT, END) and UTF32 codepoints. The simplified attaching of a single function as handler is now done with a proxy class `OscHandlerFactory`. Payload has a hard limit of 10MB.
- [x] review CSI handling and hook
  Experimental badge removed. Tests were already in place.
- [x] review ESC handling and create public interface
  Interface and tests added.
- [x] review DSC handling and create public interface
  Interface added similar to OSC with callback registration (again via a proxy class to catch the substates). Payload has a hard limit of 10MB.

Also closes #1823.


src/InputHandler.ts
```diff
@@ -152,9 +153,10 @@ export class InputHandler extends Disposable implements IInputHandler {
     this._parser.setExecuteHandlerFallback((code: number) => {
       this._logService.debug('Unknown EXECUTE code: ', { code });
     });
-    this._parser.setOscHandlerFallback((identifier: number, data: string) => {
-      this._logService.debug('Unknown OSC code: ', { identifier, data });
-    });
+    this._parser.setOscHandlerFallback((identifier, action, data) => {
+      this._logService.debug('Unknown OSC code: ', { identifier, action, data });
+    }
```
Tyriar: Bad new line

jerch: The cat again, tsss :smile_cat:

jerch:
DCS is halfway done, still missing:
- dcs parser tests
- multiple handler tests
- public interface

jerch:
@Tyriar Imho there is no formal payload limitation spec'ed for OSC and DCS commands, instead they are meant to work streamlined with chunks. I tried to deal with that by the new `OscParser` / `DcsParser` classes, which provide an interface to the chunk loading to `IDcsHandler` resp. `IOscHandler`.
At this point we clash with our one shot callback handler registration, that gives the callback the payload as one final big chunk, thus we have to store things in between until the sequences gets finished. To not run into OOM I have hardcoded a payload limit for those handlers of 50 MB (done on the proxy classes `OscHandlerFactory` and `DcsHandlerFactory`). Note that most emulators would abort these sequences at much lower limits, maybe @textshell has some numbers here?

jerch:
Up for review.

typings/xterm.d.ts
```diff
@@ -501,7 +501,7 @@ declare module 'xterm' {
     attachCustomKeyEventHandler(customKeyEventHandler: (event: KeyboardEvent) => boolean): void;
 
     /**
-     * (EXPERIMENTAL) Adds a handler for CSI escape sequences.
+     * Adds a handler for CSI escape sequences.
      * @param flag The flag should be one-character string, which specifies the
      * final character (e.g "m" for SGR) of the CSI sequence.
```
textshell: This does not sound right as stable API. The prefixes ('<', '=', '>', '?') and the intermediates are really a fundamental part of the sequence, not something that is optional to check in the parameters. Often these sequences have totally different meaning and they are an essential part of the code space reserve for new sequences.

Just a small sample of sequences which could be mismatched when organizing primarily by final instead of the full sequence identification.
```
ESC [ Ⓝ ␣ @  	Scroll Left
ESC [ Ⓝ ␣ A  	Scroll Right
ESC [ > Ⓝ T  	Unset Title Mode
ESC [ > Ⓝ c  	Secondary Device Attributes
ESC [ = Ⓝ c  	Tertiary Device Attributes 
ESC [ Ⓝ ␣ d  	Unset Tab Stop
ESC [ > Ⓝ ; Ⓝ m  	xterm keyboard mode configuration
ESC [ ? Ⓝ n  	private Device Status Report (different expected response)
ESC [ Ⓝ q  	Load LEDs (DECLL)
ESC [ Ⓝ * q  	DEC secure reset
ESC [ Ⓝ " q  	Select Character Protection Attribute (DECSCA)
ESC [ ? Ⓝ s  	Save Mode
ESC [ Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ ; Ⓝ $ r  	Change attributes in rectangular area 
ESC [ Ⓝ ␣ u  	Set Margin Bell Volume
```

jerch: Oops, thx for reminding on that particular issue, this def. needs to be resolved before going public.

Background on this: the parser originates to single callbacks for ESC/CSI/etc. which did the resolving of target actions by switch based jumps tables in C. Those switch statements perform pretty bad in JS thus I hacked the callback jumps on top of it. I did not take collects and intermediates into account for CSI for a simple reason - the intermediates are still string based which still performs pretty bad in JS and need a better representation than string based. (Other actions that account those show very low throughput performance  - ESC/DCS). So yes, the collect/intermediates have be on the registering part, not the arguments of the callback.

textshell: If the main concern here is working with smallish numbers instead of strings you might get away with a fast path for situations where each part is only one character long. finals are just 6bits, the prefix part is 3 bit and the intermediates are 4 bits. So all in all 13 bit which should still have most properties of smallish numbers. And depending on how general you want this to be you could either ignore sequences with more than one of any or have a slow path just based on string based maps.

jerch: Yeah gonna try to get away with a compound number. I also think I gonna hard limit this to up to 2 intermediates for now, although the spec does not limit this.

Regarding prefix chars - ecma48 does not limit those either, they simply account them under parameter bytes: `CSI P..P I..I F`, thus a sequence could technically look like that: `CSI ?45>1;5 ...`. This is not allowed for our VT500 based transition table, thus I gonna limit them as well to one optional occurence as the first parameter byte. Is this safe to sail with this simplification?

The final number might be accumulated like this (bytes shifted):
```
<prefix> | <intermediate one> | <intermediate two> | <final>
```
Unless we have to support more prefix/intermediates one day I will not yet strip it to the meaningful bytes as JS guarantees to handle 32 bit integers (4 bytes) natively.

textshell: I believe that limiting to one char prefix and one intermediate is safe already. I haven't yet seen anything that requires more. The one important thing is to properly ignore the sequences that are not dispatchable.
If you keep the public interface generic to support more you can always extend the limits. And it might be a good idea anyway to discourage usage of multiple characters of the same type without agreement that more terminals are willing to support that.

jerch:
Done with the rework of function registration. To see the impact here are some benchmark numbers:
```
"PRINT - a"
  master: 619.79 MB/s
  PR    : 626.24 MB/s
"EXECUTE - \n"
  master: 98.41 MB/s
  PR    : 110.16 MB/s
"ESCAPE - ESC E"
  master: 57.01 MB/s
  PR    : 49.53 MB/s
"ESCAPE with collect - ESC % G"
  master: 22.33 MB/s
  PR    : 62.60 MB/s
"CSI - CSI A"
  master: 92.12 MB/s
  PR    : 86.78 MB/s
"CSI with collect - CSI ? p"
  master: 58.06 MB/s
  PR    : 74.98 MB/s
"CSI with params (short) - CSI 1;2 m"
  master: 101.13 MB/s
  PR    : 97.11 MB/s
"CSI with params (long) - CSI 1;2;3;4;5;6;7;8;9;0 m"
  master: 146.17 MB/s
  PR    : 141.96 MB/s
"OSC (short) - OSC 0;hi ST"
  master: 52.22 MB/s
  PR    : 55.69 MB/s
"OSC (long) - OSC 0;<text> ST"
  master: 48.43 MB/s
  PR    : 65.18 MB/s
"DCS (short)"
  master: 61.88 MB/s
  PR    : 45.79 MB/s
"DCS (long)"
  master: 253.15 MB/s
  PR    : 214.90 MB/s
```
Across all functions its pretty equal. Still interesting that removing the intermediate string concat for "ESCAPE with collect - ESC % G" shows almost 3x boost while "CSI with collect - CSI ? p" is only slightly faster. OSC seems to benefit slightly from the OscParser while DCS decreased (prolly due looping into registered handlers, before it was a single handler).

jerch:
Introduced at new interface to the API for the callback registration:
```typescript
  export interface IFunctionIdentifier {
    prefix?: string;
    intermediates?: string;
    final: string;
  }
```
This data type is used during callback registration of CSI, DCS and ESC sequences on the parser to create numerical ids as the jump table keys. As shown above this greatly speeds up sequences with intermediates.

Dispatching rules:
- other than before handlers register now on one `IFunctionIdentifier` endpoint, thus if a handler should be reused for a slightly different endpoint, it has to be registered explicitly for that too, example: https://github.com/xtermjs/xterm.js/blob/7252a5921d05ba87de97a36e3d012de6853f0e3c/src/InputHandler.ts#L189-L190
- if no handler was attached a fallback handler will be triggered as before (we currently install those with debug notifications like "unknown sequence XY ...", inspectable with `{logLevel: debug}` in the demo)

Again up for review.

src/common/parser/OscParser.ts
```diff
@@ -0,0 +1,204 @@
...
+    this._id = -1;
+    this._state = OscState.START;
+  }
+}
+
+/**
+ * Convenient class to allow attaching string based handler functions
+ * as OSC handlers.
+ */
+export class OscHandlerFactory implements IOscHandler {
```
Tyriar: I think this is just an OscHandler as it doesn't create something, it is the thing.

typings/xterm.d.ts
```diff
@@ -501,27 +501,65 @@ declare module 'xterm' {
...
+     * @param callback The function to handle the sequence. The callback is
+     * called with the numerical params. If the sequence has subparams the
+     * array will contain subarrays with their numercial values.
+     * Return true if the sequence was handled; false if we should try
+     * a previous handler (set by addCsiHandler or setCsiHandler).
      * The most recently-added handler is tried first.
      * @return An IDisposable you can call to remove this handler.
      */
-    addCsiHandler(flag: string, callback: (params: (number | number[])[], collect: string) => boolean): IDisposable;
+    addCsiHandler(id: IFunctionIdentifier, callback: (params: (number | number[])[]) => boolean): IDisposable;
```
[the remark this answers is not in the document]
jerch: Yeah thats a really good idea. Should I create a `ParserApi` thingy?

src/common/parser/OscParser.ts
jerch: True, it just translates the chunk interface into the callback interface with an instance of its own. Same with DcsHandlerFactory. How about `OscCallbackHandler` or just `OscCallback`?

jerch:
> Any parts in particular that need me to review?

@Tyriar Well the payload limit of 10MB bugs me abit, but I think this can be increased with a string concat patch for chrome once we have a use case for a DCS/OSC command using that much payload.
Also could you have a look on the sub parsers dispose code? Just in case I overlooked something.

typings/xterm.d.ts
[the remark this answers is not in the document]
jerch: Created a ParserApi and namespaced the methods and the function identifier.

src/common/parser/OscParser.ts
jerch: Removed the *Factory on both names.

src/public/Terminal.ts
```diff
@@ -33,6 +35,7 @@ export class Terminal implements ITerminalApi {
   public get onResize(): IEvent<{ cols: number, rows: number }> { return this._core.onResize; }
 
   public get element(): HTMLElement { return this._core.element; }
+  public get parser(): Parser.IParser { return new ParserApi((this._core as any)._inputHandler._parser); }
```
Tyriar: Access to `_inputHandler` isn't very nice, we could leave the pass through methods on core so `ParserApi` just calls into `this._core.add*`? Also should we cache the `ParserApi` object?

jerch: `ParserApi` is cached now, also removed the namespace, thus its flat again (+ `IParser`).

src/common/parser/DcsParser.ts
```diff
@@ -0,0 +1,146 @@
...
+import { IDisposable } from 'common/Types';
+import { IDcsHandler, IParams, IHandlerCollection, IDcsParser, DcsFallbackHandler } from 'common/parser/Types';
+import { utf32ToString } from 'common/input/TextDecoder';
+import { Params } from 'common/parser/Params';
+import { PAYLOAD_LIMIT } from 'common/parser/Constants';
+
+
+export class DcsParser implements IDcsParser {
+  private _handlers: IHandlerCollection<IDcsHandler> = Object.create(null);
+  private _empty: IDcsHandler[] = [];
```
Tyriar: Should this be a `ReadonlyArray` so it cannot be modified? It seems like it's being used as a constant so maybe it should live as a const outside of `DcsParser`?

src/common/parser/DcsParser.ts
```diff
@@ -0,0 +1,146 @@
...
+
+import { IDisposable } from 'common/Types';
+import { IDcsHandler, IParams, IHandlerCollection, IDcsParser, DcsFallbackHandler } from 'common/parser/Types';
+import { utf32ToString } from 'common/input/TextDecoder';
+import { Params } from 'common/parser/Params';
+import { PAYLOAD_LIMIT } from 'common/parser/Constants';
+
+
+export class DcsParser implements IDcsParser {
+  private _handlers: IHandlerCollection<IDcsHandler> = Object.create(null);
```
Tyriar: `{}`?

src/common/parser/DcsParser.ts
```diff
@@ -0,0 +1,146 @@
...
+  private _empty: IDcsHandler[] = [];
+  private _active: IDcsHandler[] = this._empty;
+  private _ident: number = 0;
+  private _handlerFb: DcsFallbackHandler = () => {};
+
+  public dispose(): void {
+    this._handlers = Object.create(null);
```
Tyriar: `{}`?

src/common/parser/OscParser.ts
```diff
@@ -0,0 +1,204 @@
...
+import { OscState, PAYLOAD_LIMIT } from 'common/parser/Constants';
+import { Disposable } from 'common/Lifecycle';
+import { utf32ToString } from 'common/input/TextDecoder';
+import { IDisposable } from 'common/Types';
+
+
+export class OscParser extends Disposable {
+  private _state = OscState.START;
+  private _id = -1;
+  private _handlers: IHandlerCollection<IOscHandler> = Object.create(null);
```
Tyriar: Some comments on DscParser apply here too

src/common/parser/Types.d.ts
```diff
@@ -53,16 +53,21 @@ export interface IParsingState {
...
   // collect buffer with intermediate characters
-  collect: string;
+  collect: number;
   // params buffer
   params: IParams;
   // should abort (default: false)
   abort: boolean;
 }
 
+export interface IHandlerCollection<T> {
```
Tyriar: Probably not worth having this as it's just a simple map?

src/common/parser/OscParser.ts
Tyriar: Maybe we should be using `Map` instead here? IE11 has partial support apparently, not sure what that means exactly.

src/common/parser/Types.d.ts
```diff
@@ -112,31 +143,66 @@ export interface IEscapeSequenceParser extends IDisposable {
...
+  setDcsHandler(id: IFunctionIdentifier, handler: IDcsHandler): void;
+  clearDcsHandler(id: IFunctionIdentifier): void;
+  setDcsHandlerFallback(handler: DcsFallbackHandler): void;
+  addDcsHandler(id: IFunctionIdentifier, handler: IDcsHandler): IDisposable;
 
   setErrorHandler(callback: (state: IParsingState) => IParsingState): void;
   clearErrorHandler(): void;
 }
+
+export interface IOscParser extends IDisposable {
```
Tyriar: `interface IBaseSomethingParser<T>` could share a bunch of the interface between this and `IDscParser`?

src/public/Terminal.ts
```diff
@@ -33,6 +35,7 @@ export class Terminal implements ITerminalApi {
   public get onResize(): IEvent<{ cols: number, rows: number }> { return this._core.onResize; }
 
   public get element(): HTMLElement { return this._core.element; }
+  public get parser(): IParser { return this._parser; }
```
Tyriar: Should we lazy load?

```ts
if (!this._parser) {
  this._parser = new ...;
}
return this._parser;
```

typings/xterm.d.ts
```diff
@@ -951,4 +925,116 @@ declare module 'xterm' {
...
+     * Final byte, must be in range \x40 .. \x7e for CSI and DCS,
+     * \x30 .. \x7e for ESC.
+     */
+    final: string;
+  }
+
+  /**
+   * Parser interface.
+   */
+  export interface IParser {
```
Tyriar: I think we should keep this as experimental considering all the changes

src/common/parser/EscapeSequenceParser.ts
```diff
@@ -370,43 +393,62 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
...
     this.currentState = this.initialState;
-    this._osc = '';
+    this._oscParser.reset();
+    this._dcsParser.reset();
     this._params.reset();
     this._params.addParam(0); // ZDM
-    this._collect = '';
-    this._activeDcsHandler = null;
+    this._collect = 0;
+    // this._activeDcsHandler = this._dcsHandlerFb;
```
Tyriar: Remove?

src/common/parser/Constants.ts
```diff
@@ -43,3 +43,16 @@ export const enum ParserAction {
...
+ */
+export const enum OscState {
+  START = 0,
+  ID = 1,
+  PAYLOAD = 2,
+  ABORT = 3
+}
+
+// payload limit for OSC and DCS
+export const PAYLOAD_LIMIT = 10000000;
```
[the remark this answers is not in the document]
jerch: No biggie, just that it might be to low rather soon and the fact that we size limited by some weird platform internals (again the string concat problem on chrome).

src/common/parser/DcsParser.ts
jerch: Ah good catch yeah, guess the const on file level is more suitable here.

src/common/parser/OscParser.ts
jerch: Oh we support Map as type now? Imho Map has the nicer interface as it does not have to convert the key type to string, still this is not really an issue with integers as all engines I am aware of heavily optimize object key access.
Gonna benchmark with Map and switch if thats on par or better.

src/common/parser/Types.d.ts
jerch: Yes can simplify the handler registration interface with that :+1:

src/public/Terminal.ts
jerch: Right, prolly used in <5% of all integrations thus worth to make it lazy (would not save much though).

src/common/parser/EscapeSequenceParser.ts
jerch: :sweat_smile:

src/common/parser/Types.d.ts
jerch: Yeah, see comment above, will change if it passes the speed tests.

src/InputHandler.ts
```diff
@@ -1549,6 +1576,7 @@ export class InputHandler extends Disposable implements IInputHandler {
     }
   }
 
+
```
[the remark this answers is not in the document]
jerch: OMG, fixed *phew

src/Types.d.ts
```diff
@@ -113,7 +113,8 @@ export interface IInputHandler {
   /** CSI ` */ charPosAbsolute(params: IParams): void;
   /** CSI a */ hPositionRelative(params: IParams): void;
   /** CSI b */ repeatPrecedingCharacter(params: IParams): void;
-  /** CSI c */ sendDeviceAttributes(params: IParams, collect?: string): void;
+  /** CSI c */ sendDeviceAttributesPrimary(params: IParams): void;
+               sendDeviceAttributesSecondary(params: IParams): void;
```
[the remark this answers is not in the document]
jerch: Yep.

src/common/parser/DcsParser.ts
jerch: This is to avoid prototype lookups, which can happen if not excluded with additional `hasOwnProperty` check, which has bad performance. To cut both issues, `Object.create(null)` is the way to go.

src/common/parser/OscParser.ts
jerch: Well seems the key access into a Map is worse in performance in nodejs - I get ~100MB/s with object vs. 45 MB/s with `Map` for my "EXECUTE - \n" test case.

Gonna stick with `{[key: number]: HandlerXY[]}` for now. Still gonna remove the `IHandlerCollection` type as its only a difference in the `[]`.

typings/xterm.d.ts
```diff
@@ -500,32 +500,6 @@ declare module 'xterm' {
...
-    /**
-     * (EXPERIMENTAL) Adds a handler for OSC escape sequences.
-     * @param ident The number (first parameter) of the sequence.
-     * @param callback The function to handle the escape sequence. The callback
-     * is called with OSC data string. Return true if the sequence was handled;
-     * false if we should try a previous handler (set by addOscHandler or
-     * setOscHandler). The most recently-added handler is tried first.
-     * @return An IDisposable you can call to remove this handler.
-     */
-    addOscHandler(ident: number, callback: (data: string) => boolean): IDisposable;
```
Tyriar: I don't think `parser: IParser` is declared in `Terminal`? Let's add experimental to it as well.

jerch: Good catch, fixed.

Issue #2324, opened 2019-07-19
Replace chai with built in assert node module


Tyriar:
I don't think we need it, we should convert expect-style assertions to assert and replace chai with https://nodejs.org/api/assert.html

Tyriar:
Closing as designed, see https://github.com/xtermjs/xterm.js/pull/2347#issuecomment-516041920

PR #2347, opened 2019-07-25
Removes Chai Assertion Library

lramos15:
Lowers test expectations by removing all expects (programmers can be punny too). Removes chai assertion library in favor of Node.JS one. Closes https://github.com/xtermjs/xterm.js/issues/2324

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "clean": "rm -rf lib out addons/*/lib addons/*/out"
   },
   "devDependencies": {
-    "@types/chai": "^3.4.34",
+    "@types/assert": "^1.4.2",
     "@types/glob": "^5.0.35",
     "@types/jsdom": "11.0.1",
     "@types/mocha": "^2.2.33",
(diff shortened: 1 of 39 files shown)


package.json
```diff
@@ -26,7 +26,7 @@
     "clean": "rm -rf lib out addons/*/lib addons/*/out"
   },
   "devDependencies": {
-    "@types/chai": "^3.4.34",
+    "@types/assert": "^1.4.2",
```
Tyriar: I would expect it to be included in @types/node, is it not?

Tyriar: Yeah, I think you need to remove this:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e244bc5f6b22e0a89c631b986edf1d183a5bfc95/types/assert/index.d.ts#L2

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/assert.d.ts

lramos15:
Because src/common does not allow dependencies on any node APIs we would not be able to convert all of the code over and the work that would be needed to get it to work would not be worth the effort to remove Xterm's dependency on Chai.

PR #2360, opened 2019-08-04
column sequences

jerch:
Basic SL / SR / DECIC / DECDC support.

TODO:
- [x] test files against xterm behavior
- [ ] ~implement DECLRMM~
- [ ] ~test interaction with DECLRMM~

Update: DECLRMM should go into a separate PR as its more involved (also influences other VT primitives).


jerch:
@Tyriar To not get stale I removed DECLRMM from this PR as its more involved and better done separately.

Up for review.

jerch:
Geez not yet done with this PR - DECIC and DECDC are margin aware, thus should only shift content between margins, its a NOOP outside. See http://www.bitsavers.org/pdf/dec/standards/EL-SM070-00_DEC_STD_070_Video_Systems_Reference_Manual_Dec91.pdf#page=406&zoom=auto,-38,762
Also for SL/SR xterm seems to apply margins...

jerch:
@Tyriar With DECLRMM it can be used by apps to do partial viewport updates (like splitting in a multiplexer and such). Since this is supported for long in xterm and other emulators apps might shift to it, not sure if any does atm. Maybe ncurses already uses this, idk.

Already fixed in with the last commit, tests still to come.

Tyriar:
@jerch k, ping me when ready to review 😃

jerch:
@Tyriar Ping :smile:

src/InputHandler.ts
```diff
@@ -1033,18 +1037,108 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+   */
+  public scrollLeft(params: IParams): void {
+    const buffer = this._bufferService.buffer;
+    if (buffer.y > buffer.scrollBottom || buffer.y < buffer.scrollTop) {
+      return;
+    }
+    const param = params.params[0] || 1;
+    for (let y = buffer.scrollTop; y <= buffer.scrollBottom; ++y) {
+      const line = buffer.lines.get(buffer.ybase + y);
+      line.deleteCells(0, param, buffer.getNullCell(this._terminal.eraseAttrData()));
```
Tyriar: lgtm, the only thing I'm not sure about is whether eraseAttrData is right here?

jerch: Ah good point, have to see whether xterm applies BCE here...

jerch: @Tyriar Yepp its with the erase attrs, thus can go as it is.

PR #2364, opened 2019-08-06
Support "alt+enter" sending `\e\r`

NHDaly:
This is a PR to add support for "alt+enter" key binding to send `\e\r` (used in some REPLs to insert a newline).

EDIT:
It adds a test that "alt+return" key combination sends `\e\r`.
And I've added my guess at what the fix would be.


Tyriar:
Does xterm or other terminals do this?

pfitzseb:
At least Konsole and alacritty do -- this is <kbd>Enter</kbd> followed by <kbd>Alt-Enter</lbd>:
```
λ showkey -a

Press any keys - Ctrl-D will terminate this program

^M       13 0015 0x0d
^[^M     27 0033 0x1b
         13 0015 0x0d
^D        4 0004 0x04
```
[Apparently](https://github.com/xtermjs/xterm.js/blob/a89771a8de58fbc7a82323fb7d6d9c421ef63535/src/InputHandler.ts#L1192-L1196) xterm specifies modes for this as well.

NHDaly:
Oops! Sorry, i didn't realize i'd submitted the PR! I thought i was still editing it. ... haha it must have been that darned "shift+enter", which is very funny considering the situation. 😅

Sorry about that! :) I've just edited the first comment with a bit more information, including that I've just pushed up what I _think_ is the right place for the fix.


> Does xterm or other terminals do this?
I'm not exactly sure where to find documentation on this, but I think so. iTerm seems to do this on macOS, and a coworker who uses linux says his terminal does this as well.

But now that I'm digging into it more, this might just be because iTerm defaults to setting the option-keys to "Esc+"?

Tyriar:
> Apparently xterm specifies modes for this as well.

Looks like we need another DECSET mode? These are currently maintained here https://github.com/xtermjs/xterm.js/blob/a89771a8de58fbc7a82323fb7d6d9c421ef63535/src/common/services/CoreService.ts#L18

jerch:
Yes I think we should do as xterm does here and make it usable from application side with DECSET 1036 and 1039. I think we need to support both as we have an option to declare <kbd>Option</kbd> on Mac as Meta key.

Next question is whether this should be enabled by default. Seems most emulators enable this (quickly tested gnome-terminal, konsole, rxvt) only xterm sends by default the 8bit chars on Alt+Key, thus is needs to be explicitly set via sequence or startup options. I'd say we should enable this by default too and make the default customizable via ctor option.

NHDaly:
Thanks for thinking seriously about the right way to handle this! 🎉 :)

NHDaly:
Friendly ping: have there been any further thoughts on this? :) Thanks!

jerch:
@NHDaly No updates on this. Since you opened this PR question at hand is whether you want to implement the DECSET and the tests yourself here (if thats to much "internal rubbish" for you we can start over with a new PR).

Tyriar:
@NHDaly friendly ping, we wouldn't be able to merge this without only conditionally doing it in that mode.

NHDaly:
> No updates on this. Since you opened this PR question at hand is whether you want to implement the DECSET and the tests yourself here (if thats to much "internal rubbish" for you we can start over with a new PR).

Hi, thanks for the ping! ❤️

I'm sorry i dropped this PR. Yes, unfortunately I don't have as many free cycles now as I had when I opened this PR, so if you'd be able to take over this work -- especially since it's going to be more involved and internal -- I would _super_ appreciate it.

Feel free to either close this PR or to take it over. I currently have it set to "Allow updates from maintainers," so you should be able to do whatever is easiest for you! :)

Thanks again for your quick responses!!

Tyriar:
@NHDaly no worries, thanks for the effort. I'll close this for now and call out this WIP is available for reference.

Issue #2075, opened 2019-05-11
Stabilize buffer API


Tyriar:
Introduced in https://github.com/xtermjs/xterm.js/pull/2074, we should remove experimental after we know confirm no changes are needed for the search addon to function.

jerch:
@Tyriar Made some progress in that field in #2369 together with @JavaCS3, see this comment https://github.com/xtermjs/xterm.js/pull/2369#issuecomment-526765632 and my changes here: https://github.com/xtermjs/xterm.js/compare/master...jerch:serialize_with_private#diff-3cf93a85b230e197db3cdf45b888018f

I am sure you not gonna like the changes much, basically it seems we should still avoid getters.

Tyriar:
@jerch this blocks the v4 release as we should stabilize such that we don't run into issues with the search addon having broken API in a later v4 release. How about instead converting all the getters into a function we preemptively fetch all the line's values and expose them as regular properties. It's readonly on the API and doesn't matter much if people try to workaround it anyway.

So instead of this:

```ts
return new BufferCellApiView(this._line, x);

...

class BufferCellApiView implements IBufferCellApi {
  constructor(private _line: IBufferLine, private _x: number) {}
  public get char(): string { return this._line.getString(this._x); }
  public get width(): number { return this._line.getWidth(this._x); }
}
```

We do this instead:

```ts
return {
    char: this._line.getString(x),
    width: this._line.getWidth(x)
}
```

Or this:

```ts
return new BufferCellApiView(this._line, x);

...


class BufferCellApiView implements IBufferCellApi {
  public char: string;
  public width: number;
  constructor(line: IBufferLine, x: number) {
    this.char = line.getString(x);
    this.width = line.getWidth(x);
  }
}
```

Reasons I don't think we should do all functions:

- We would be breaking existing (experimental) API, that would result in the search addon for example breaking for all future versions. While we can do this, it is a bit of a pain, plus VS Code uses the buffer API.
- Functions are uglier to use than properties, it would be best if we would be consistent across the API ( `IBufferApi.cursorY`, `IMarker.line`, `Terminal.element`, etc.).
- While there will always be faster ways, they aren't as ergonomic and there are ways around it (don't serialize as much, support incremental serialization, etc.).

/cc @JavaCS3

jerch:
@Tyriar 

[...]

PR #2369, opened 2019-08-08
Add foreground/background color support for SerializeAddon

### LINKED ISSUE #2383: Expose buffer cell's color data via API
It's related to https://github.com/xtermjs/xterm.js/issues/595#issuecomment-500233934

Currently plugin can access only "text" data in the Buffer.
To have better support of XTerm.JS serialization it's highly desirable to have possibility to access colors as well.

Is it the right time to extend the API?

JavaCS3:
Fixes #2383

@Tyriar Here's the new update

Currently there're some un-implemented features: re-position cursor, wrap mode, combinedData
Will continue improve it.

xterm js record


addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -4,9 +4,223 @@
  */
 
 import { Terminal, ITerminalAddon } from 'xterm';
+import { ICellData } from 'common/Types';
+import { CellData } from 'common/buffer/CellData';
```
Tyriar: Usage of anything in `common/` is pulling in private API which will cause versioning issues for the addon down the track. Also note that this includes the contents of CellData.ts and Constants.ts inside the addon as well.

I think the right fix here is to expose the necessary attributes on `IBufferCell` in xterm.d.ts and commit to that as API.

src/public/Terminal.ts
```diff
@@ -204,19 +205,43 @@ class BufferLineApiView implements IBufferLineApi {
...
   }
 }
 
 class BufferCellApiView implements IBufferCellApi {
-  constructor(private _line: IBufferLine, private _x: number) {}
-  public get char(): string { return this._line.getString(this._x); }
-  public get width(): number { return this._line.getWidth(this._x); }
+  constructor(private _cell: ICellData) {}
+
+  public get fg(): number { return this._cell.fg; }
```
Tyriar: Sorry I didn't elaborate on the exact direction here but I think we need a slimmer API here for simplicity. ICellData is quite complex and just exposes a bunch of numbers and booleans so that the internals can be as fast as possible, but for the buffer API we assume the usage isn't going need that sort of level of optimization so we can put a nice simple facade on top. Here is what I'm thinking but I'm curious what @jerch thinks:

```ts
interface IBufferCell {
    readonly char: string;
    readonly width: number;
    readonly foregroundColor: Color;
    readonly backgroundColor: Color;
    readonly style: Style;
}

type Color = IDefaultColor | IPalette16Color | IPalette256Color | IRgbColor;

type CellStyle = 'inverse' | 'bold' | 'underline' | 'blink' | 'invisible' | 'italic' | 'dim';

interface IDefaultColor {
    type: 'default';
}

interface IRgbColor {
    type: 'rgb';

    // 0-255
    red: number;
    green: number;
    blue: number;
}

interface IPalette16Color {
    type: 'palette16';

    // 0-15
    id: number;
}

interface IPalette256Color {
    type: 'palette256';
    // 0-255
    id: number;
}
```

Note that we do use colors as `string`s elsewhere in the API but those are provided by the consumer which `#rrggbb` is a convenient format for, here we're exposing to consumers and `Color` which saves people from parsing that out. It does get a little complex with all the color modes but I'm not sure we can get simpler than that.

JavaCS3: Thanks for quick response. I have some concerns about the `CellStyle`
1. There's a case a cell will have not only one style. For example, `\x1b[1;4mBoldAndUnderlined\x1b[0m`. 
2. Another thing is I need to make serialize result as small as possible. For example 
If previous cell style is Bold and Underline the next cell is Bold. I only need to unset Underline flag by `\x1b[24m` instead of always reset to `\x1b[0m` for each cell. That's why I have XOR caculation which bit of Style changed. 
So I think bitwise based style is more efficient. Maybe you have better way to solve this problem.

jerch: @Tyriar, @JavaCS3:
I second what @JavaCS3 said, maybe we can keep the bitflags for CellStyle, but merged from `BgFlags` and `FgFlags` into one:
```typescript
enum CellStyle {
  // first byte FgFlags
  bold = ... (must match bit pos in FgFlags + conversion)
  // second byte BgFlags
  ...
};

// conversion in API getter:
public get style(): CellStyle {
  // maybe cache the result
  return (this._cell.bg >> 16) | (this._cell.fg >> 24);
}

// boolean eval in addon
if (style & CellStyle.BOLD) ...
```

About IColor:
The `type` member on the color types is important to avoid costly `instanceof` checks, yeah.

Offtopic but relevant in the bigger API picture:
Problem with the color types and number exposed here I see is that ppl might not be interested in this pre-render stage data but want to grab the color thats going to be shown. Note that this is basically the same issue as I pointed out in the previous PR about empty cells vs. whitespace. I think we have to make clear what we want to expose in `BufferApi` - shall this be a buffer data representation layer or more in line with what the renderer would make of it?

Proposal to solve this ambiguity: Make `BufferApi` what the name suggests, expose the data from the buffer, thus with different color types (that have no meaning without a theme) and WITH empty cells by default.
Later on in a second API expose things that would be applied on the way to the screen, like right trimming / whitespace replacements and colors from the active theme.

Tyriar: :+1: to flags for cell style.

> Later on in a second API expose things that would be applied on the way to the screen, like right trimming / whitespace replacements and colors from the active theme.

Not sure we want to go in that direction, the current method of exposing the different color types works for @JavaCS3 and my use case as its purely for serializing and restoring.

jerch: @Tyriar Well to me it seems kinda arbitrary why we expose buffer content with different rules applied here:
- char content: Gets the trimming and the whitespace replacements as the renderer would do, thus we actually dont expose buffer content here
- colors: Exposes buffer content with on renderer stuff applied

Imho thats confusing for a "buffer" API.

JavaCS3: Should I ahead with @jerch 's proposal or wait until the BufferAPI to be finialized?

Tyriar: > Gets the trimming and the whitespace replacements as the renderer would do, thus we actually dont expose buffer content here

Not sure I understand, we're returning the character or `''` for `char`:

https://github.com/xtermjs/xterm.js/blob/95ff15457b51d4fd9738b9949457ab3f0e25a002/src/public/Terminal.ts#L217

If you request a column that is "null" whitespace on the right of a column wouldn't it would return as `''`?

jerch: Ah sorry thought you are working with `translateToString` :sweat_smile: , `getString` correctly returns an empty string for empty cells. So on individual cell level you get the right value, all good.

JavaCS3: I didn't meet any trouble using the exisiting `public get char(): string`.

My latest change `NullBufferCell` may mislead u guys. The main reason why I do that is I want to make changes small. There's a case (`_bgChanged(...)` function) I need to tell the previous cell of a cell. But when it's the first cell, the previous cell is not even exisit. So I just use a "NullObject Strategy" to solve it in a quick way. It doesn't mean I need to treat null cell differently.

Maybe later on, I will meet some kind of null cell problems. We will see. For now it's ok to me.

typings/xterm.d.ts
```diff
@@ -955,27 +952,58 @@ declare module 'xterm' {
...
+  }
+
+  interface ICellColor {
+    type: string;
+    hash: number;
+  }
+
+  interface IDefaultColor extends ICellColor {
+    type: 'default';
+    hash: 0;
```
JavaCS3: `hash` is designed to quickly tell if two color is different.

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -3,8 +3,30 @@
  * @license MIT
  */
 
-import { Terminal, ITerminalAddon, IBuffer, IBufferCell } from 'xterm';
-import { Attributes, FgFlags, BgFlags } from 'common/buffer/Constants';
+import { Terminal, ITerminalAddon, IBuffer, IBufferCell, Color } from 'xterm';
+
+// TODO: Workaround here, will remove this later
+// If I use `import { CellStyle } from 'xterm'` instead, demo page will raise bellow error
```
JavaCS3: I don't know how to solve it in a quick way. I just wanna give u guys a glance of using the new API proposal.

JavaCS3:
Please check this new update

src/public/Terminal.ts
```diff
@@ -220,28 +221,46 @@ class BufferLineApiView implements IBufferLineApi {
...
+    const hash = cell.fg & COLOR_MASK;
+    if (cell.isFgDefault()) {
+      return { type: 'default', hash: 0 };
+    } else if (cell.isFgPalette()) {
+      switch (cell.getFgColorMode()) {
+        case Attributes.CM_P16: return { type: 'palette16', hash, id: cell.getFgColor() };
+        case Attributes.CM_P256: return { type: 'palette256', hash, id: cell.getFgColor() };
+      }
+    } else if (cell.isFgRGB()) {
+      const [red, green, blue] = CellData.toColorRGB(cell.fg);
```
jerch: Just curious - how does this perform if you serialize tons of lines (>10000)? Note that the interim creation of objects is quite expensive, basically for every cell 2 color objects have to be created and for RGB even 4 (due to the interim [r,g,b] array).
You basically hit the same problem here I had when I designed the buffer cells access. I went with the "ref style" approach with `loadCell` to avoid these costs in renderers. In my benchmarks the interim objects were about 3-4 times slower.
Not sure if we could resemble this to some degree here (not even sure if its worth to try, thus the question about the performance).

JavaCS3: I don't have benchmark yet. I think we could refactor it later if needed. There's a saying "premature optimization is the root of all evil". So I think we can optimize performance later after this feature is working.

jerch: Well if you cite Donald Knuth you should also grasp the context where he explains this, which might not be true here: Its about hunting for small efficiency gains, which tend to eat tons of time and distract programmers. I suspect this to run at least 5 times slower, thats not what I'd call "small". Furthermore it might have direct impact on the API design which cannot be undone easily afterwards.

JavaCS3: I this case, you might be right. I have no idea how to improve it currently other than hex based algorithm do u have any design suggestions?

jerch: No not yet.
Maybe try to benchmark your current version and another one which uses the private `loadCell` semantic to get a first impression. If thats within a factor of 2 I would'nt care much (its expected to lose some performance due to a stable API layer). If its bigger we might have to borrow some ideas from the internal way.

To only test the impact of the API layer with interim object creation its important that you dont hide the difference behind much more expensive code down the road (basically make `.serialize` a NOOP). You can either use our benchmark tool for it (see other benchmarks under /test/benchmark) or simply do some `Date.now()` debugging (less accurate but enough to grasp the difference if the amount of cells processed is high enough).

JavaCS3: @jerch You're right the new design performance is worser than the original one.

Notes: because I need to serialize color, so I use `lolcat` to help me generate a colorful output  

Highlights:
1. Private API: average throughput: 8.03 MB/s
2. The New API: average throughput: 2.51 MB/s

I have pushed my comparison code into my personal repo. you can check it by this [link](https://github.com/JavaCS3/xterm.js/compare/benchmark/serialize-addon-private-api...JavaCS3:benchmark/serialize-addon-public-api)

I think the key difference is the `Color` schema is dynamic and need to be created for each cell. But the original private api still need to `new CellData()` for each cell as well. Below is the details.
 
```shell
# Private API
➜  xterm.js yarn benchmark-baseline
yarn run v1.17.3
$ NODE_PATH=./out:./out-test/benchmark xterm-benchmark -r 5 -c test/benchmark/benchmark.json --baseline out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js
   Context "out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js"
      Context "Terminal: sh -c "ls -lR /usr/lib | lolcat -f""
         Context "serialize"
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
            Case "#1" : 5 runs - average throughput: 8.03 MB/s

### Baseline data ###
"out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js|Terminal: sh -c "ls -lR /usr/lib | lolcat -f"|serialize|#1"
   #averageRuntime
      STAT             BASE  TOLERANCE
      mean            18.77  0.75-1.50
      median          18.71    skipped
      dev             12.16    skipped
      cv               0.65    skipped
      runs             5.00    skipped
   #averageThroughput
      STAT             BASE  TOLERANCE
      mean             8.03  0.75-1.50
      median           5.22    skipped
      dev              6.01    skipped
      cv               0.75    skipped
      runs             5.00    skipped
✨  Done in 4.57s.

# The New API
➜  xterm.js yarn benchmark-eval                              
yarn run v1.17.3
$ NODE_PATH=./out:./out-test/benchmark xterm-benchmark -r 5 -c test/benchmark/benchmark.json --eval out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js
   Context "out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js"
      Context "Terminal: sh -c "ls -lR /usr/lib | lolcat -f""
         Context "serialize"
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
            Case "#1" : 5 runs - average throughput: 2.51 MB/s
"out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js|Terminal: sh -c "ls -lR /usr/lib | lolcat -f"|serialize|#1"
   #averageRuntime
      STAT             BASE  TOLERANCE      VALUE  CHANGE(%)       EVAL
      mean            18.77  0.75-1.50      40.53     115.96     FAILED
      median          18.71    skipped      39.43                  SKIP
      dev             12.16    skipped       8.87                  SKIP
      cv               0.65    skipped       0.22                  SKIP
      runs             5.00    skipped       5.00                  SKIP
   #averageThroughput
      STAT             BASE  TOLERANCE      VALUE  CHANGE(%)       EVAL
      mean             8.03  0.75-1.50       2.51     -68.77     FAILED
      median           5.22    skipped       2.48                  SKIP
      dev              6.01    skipped       0.54                  SKIP
      cv               0.75    skipped       0.22                  SKIP
      runs             5.00    skipped       5.00                  SKIP


 Success: 0
 Missing: 0
 Skipped: 8
 Failed: 2
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

addons/xterm-addon-serialize/src/SerializeAddon.ts
Tyriar: I think the issue you're seeing is you're mixing types and implementation. xterm.d.ts should not contain the implementation of the enum, only the contract. Try moving all the = values out into Terminal.ts. I don't think we've donee this with enums yet but I expect it to work.

typings/xterm.d.ts
```diff
@@ -955,27 +952,58 @@ declare module 'xterm' {
...
+    underline = 0x10000000 >>> 24,
+    blink = 0x20000000 >>> 24,
+    invisible = 0x40000000 >>> 24,
+    // background style
+    italic = 0x4000000 >>> 16,
+    dim = 0x8000000 >>> 16
+  }
+
+  interface ICellColor {
+    type: string;
```
Tyriar: You might need to remove type here? Ideally we want it to error when you try to do something like this:

```
if (color.type === 'fake')
```

If you have it declared as string it might become this:

```
type: 'default' | 'rgb' | 'palette16' | 'palette256' | string;
```

We want it to be this:

```
type: 'default' | 'rgb' | 'palette16' | 'palette256';
```

typings/xterm.d.ts
Tyriar: I think you can safely remove `hash` from all the extended interfaces, declaring it as `0`/static here probably doesn't give us anything?

src/public/Terminal.ts
jerch: Wow, thanks for benchmarking it this extensively :+1:. The numbers are within what I expected, hmm. About the `new CellData()` - yeah the renderers even skip that by reusing the same object over and over and just load the primitive numbers with `loadCell`.

Geez, seems this will be a tradeoff between a nice and easy to use API and performance. I dont want to be to picky here about performance, so maybe we can find some middle ground to shape the API.

Still have to go through your benchmark tests, will see if I can come up with some alternative API suggestions. Feel free to do so as well.

Tyriar: It doesn't seem that bad considering this will happen so infrequently, making the API easy to use is pretty important.

jerch: Worst case estimation:
Given a fullscreen terminal with 400 cols x 10k scrollback `serialize` would have to process 4M cells. If every cell contains different RGB foreground and RGB background a cell string would end up as a string of length ~40, thus total runtime is about:
```
4M * 40 / 2.5 = 64s
4M * 40 / 8 = 20s
```
The values are extreme values by purpose, a typical terminal instance would be more in the range of 120 cols x 1k scrollback with most cells containing colors of their neighbors (the string contains only ~2 chars on average), thus the typical runtime might be more like:
```
120k * 2 / 2.5 = 96ms
120k * 2 / 8 = 30ms
```
With 8MB/s we are right at the perceivable delay threshold, with 2.5 MB/s we have a small hiccup. Note that this will freeze everything else within a bigger application/page for that amount of time, so I am not sure if thats bearable. I see two possible solutions for that:
- stick with the slower API design, but move `.serialize` and buffer data to be processed to a webworker
- extend the API with an optional ref style way, maybe we just need to add a `getCell(cell?: BufferCell)` method, that can optionally copy stuff into a given cell instead of creating new objects and keeping the GC busy.

JavaCS3: webworker is overkill to me. I even want to run xterm.js on server side. I'm ok even if the API doesn't looks pretty, I think 2nd looks better. 

How about combine palette16/256/RGB into one single data structure?
```
interface ICellColor {
  type: 'default' | 'rgb' | 'palette16' | 'palette256'';
  hash: number;
  id: number;
  red: number;
  green: number;
  yellow: number;
}
```
or 
```
interface ICellColor {
  type: 'default' | 'rgb' | 'palette16' | 'palette256'';
  equals(c: ICellColor): boolean;
  toPaletteId(): number;
  toRGBColor(): [number, number, number]; 
}
```

JavaCS3:
The latest change still have trouble compiling demo page with the following error, I need some help. 
```
ERROR in ./addons/xterm-addon-serialize/out/SerializeAddon.js
Module not found: Error: Can't resolve 'xterm' in '/Users/javacs3/Lab/Playground/xterm.js/addons/xterm-addon-serialize/out'
 @ ./addons/xterm-addon-serialize/out/SerializeAddon.js 16:14-30
 @ ./demo/client.ts
```

jerch:
@JavaCS3 I encounter similar errors after switching branches.  For same reason the watcher gets often stuck by this. My solution then is to remove all builds (all `out` folders) and to restart the compile watcher. After that it works again for me.

Sorry for not answering on the topic above, hope to get back on track by the weekend.

JavaCS3:
> @JavaCS3 I encounter similar errors after switching branches. For some reason the watcher gets often stuck by this. My solution then is to remove all builds (all `out` folders) and to restart the compile watcher. After that it works again for me.
> 
> Sorry for not answering on the topic above, hope to get back on track by the weekend.

@jerch My error is not related to `yarn watch`. It happens when `yarn start`. I found the root cause but I don't know how to solve it because I'm not a webpack expert. I checked the compilation output of `SerializeAddon.ts`. It will require `xterm` in the addon, but it cannot find it.
```javascript
// ...
Object.defineProperty(exports, "__esModule", { value: true });
var xterm_1 = require("xterm");
var CellStyle;
// ...
```

JavaCS3:
@jerch I accidently fixed that problem after read webpack docs and google. Hope this is the right way to solve it.

JavaCS3:
@jerch  How about this now? I run another benchmark and workaround some imporing issues to get this new benchmark. Looks like a little improve that the previous one. 
```
➜  xterm.js yarn benchmark-baseline
yarn run v1.17.3
$ NODE_PATH=./out:./out-test/benchmark xterm-benchmark -r 5 -c test/benchmark/benchmark.json --baseline out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js
   Context "out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js"
      Context "Terminal: sh -c "ls -lR /usr/lib | lolcat -f""
         Context "serialize"
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
serialize addon: startRow= 0 endRow= 706
            Case "#1" : 5 runs - average throughput: 3.73 MB/s

### Baseline data ###
"out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js|Terminal: sh -c "ls -lR /usr/lib | lolcat -f"|serialize|#1"
   #averageRuntime
      STAT             BASE  TOLERANCE
      mean            32.54  0.75-1.50
      median          25.68    skipped
      dev             17.10    skipped
      cv               0.53    skipped
      runs             5.00    skipped
   #averageThroughput
      STAT             BASE  TOLERANCE
      mean             3.73  0.75-1.50
      median           3.81    skipped
      dev              1.78    skipped
      cv               0.48    skipped
      runs             5.00    skipped
✨  Done in 3.75s.
```

jerch:
@JavaCS3 Yeah thats much better. Sadly I cannot try myself - cannot get it working prolly due to the same error as you had earlier (missing 'xterm' stuff). If you have some pointers how to set it up correctly I could have a look as well.

I still think trying a ref style cell thingy on API level might be worth a shot, I expect this to run at +15 MB/s rates (I still see 35 MB/s for the input benchmark with the lolcat command). But at those rates serialize might suffer from a string concat performance bottleneck, to level this out you might have to play with different string creation semantics (e.g. `s += chunk` vs. `chunks.join('')`).

JavaCS3:
@jerch If you need to run benchmark on the latest code, I did some tricks
1. cherry-pick the last commit of my `serialize-addon-public-api` branch from my forked repo.  i.e `git cherry-pick 5fdf196ad75fe06913cc5ede9cde322b9c54cf69`
2. Make some change like this and yarn build it.
```diff
diff --git a/test/benchmark/tsconfig.json b/test/benchmark/tsconfig.json
index 51abeba5..60c6bcc1 100644
--- a/test/benchmark/tsconfig.json
+++ b/test/benchmark/tsconfig.json
@@ -18,7 +18,8 @@
       "browser/*": [ "../../src/browser/*" ],
       "addons/xterm-addon-serialize/src/*": ["../../addons/xterm-addon-serialize/src/*"],
       "public/*": ["../../src/public/*"],
-      "Terminal": ["../../src/Terminal"]
+      "Terminal": ["../../src/Terminal"],
+      "xterm": ["../../src/public/Terminal"]
     },
   },
   "include": [
```
3. Directly change the compilation output under `out-test/benchmark/addons/xterm-addon-serialize/src/SerializeAddon.js`. Change the line `require("xterm");` to `require("public/Terminal");`

The key point of "missing 'xterm' stuff" is the compilation output of `SerializeAddon.ts` has a line `require("xterm");`. However the xterm source code doesn't have that kind of file (xterm.ts/xterm.js). The entry point of xterm package is `src/public/Terminal`, So you have to find a way to redirect all `require("xterm")` to `require("src/public/Terminal")`. I'm wondering if we could rename `src/public/Terminal.ts` to `src/public/xterm.ts`

jerch:
@JavaCS3 Thx, got it working (with a few more changes though). These are the results for my machine (serializing 80x1000 lines of lolcat):
```
   Context "out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js"
      Context "Terminal: sh -c "ls -lR /usr/lib | lolcat -f""
         Context "serialize"
            Case "#1" : 1 - runtime: 51.04 ms
            Case "#1" : 1 - throughput: 2.71 MB/s
            Case "#1" : 2 - runtime: 57.70 ms
            Case "#1" : 2 - throughput: 2.39 MB/s
            Case "#1" : 3 - runtime: 33.46 ms
            Case "#1" : 3 - throughput: 4.13 MB/s
            Case "#1" : 4 - runtime: 30.12 ms
            Case "#1" : 4 - throughput: 4.59 MB/s
            Case "#1" : 5 - runtime: 26.34 ms
            Case "#1" : 5 - throughput: 5.25 MB/s
            Case "#1" : 5 runs - average runtime: 39.73 ms
            Case "#1" : 5 runs - average throughput: 3.81 MB/s
```
I switched on the reports for individual runs - as I already suspected by your mean/median and cv values from above they differ alot and are badly skewed. Since it shows much better runtime in the last runs its most likely due to some late optimizations done by v8.
To really see what going on further profiling would be needed (can be switched on for xterm-benchmark), but meh I am to lazy for it. My guess is that the cold runs are closer to the truth and v8 finds later some shortcuts that would not apply during normal (non repetitive) usage.

jerch:
@JavaCS3 Hacked quickly a version using private APIs with ref style, see https://gist.github.com/jerch/2dfaea9b52f8573456894dad70fa0f1c.

Benchmark results with same test file:
```
   Context "out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js"
      Context "Terminal: sh -c "ls -lR /usr/lib | lolcat -f""
         Context "serialize"
            Case "#1" : 1 - runtime: 22.00 ms
            Case "#1" : 1 - throughput: 6.14 MB/s
            Case "#1" : 2 - runtime: 11.83 ms
            Case "#1" : 2 - throughput: 11.41 MB/s
            Case "#1" : 3 - runtime: 9.69 ms
            Case "#1" : 3 - throughput: 13.92 MB/s
            Case "#1" : 4 - runtime: 6.16 ms
            Case "#1" : 4 - throughput: 21.92 MB/s
            Case "#1" : 5 - runtime: 5.71 ms
            Case "#1" : 5 - throughput: 23.64 MB/s
            Case "#1" : 6 - runtime: 7.37 ms
            Case "#1" : 6 - throughput: 18.30 MB/s
            Case "#1" : 7 - runtime: 5.90 ms
            Case "#1" : 7 - throughput: 22.86 MB/s
            Case "#1" : 8 - runtime: 9.62 ms
            Case "#1" : 8 - throughput: 14.03 MB/s
            Case "#1" : 9 - runtime: 5.73 ms
            Case "#1" : 9 - throughput: 23.54 MB/s
            Case "#1" : 10 - runtime: 5.71 ms
            Case "#1" : 10 - throughput: 23.62 MB/s
            Case "#1" : 10 runs - average runtime: 8.97 ms
            Case "#1" : 10 runs - average throughput: 17.94 MB/s
```
It still shows that weird speedup, but much earlier this time. I did not optimize anything in the code, it is mostly c&p from core stuff. I think we should rethink the public API from here, not the way around. 17 MB/s vs. 4 MB/s is quite alot.

jerch:
Proposal for public API:
```diff
diff --git a/src/public/Terminal.ts b/src/public/Terminal.ts
index 6d6cedca..f63b926b 100644
--- a/src/public/Terminal.ts
+++ b/src/public/Terminal.ts
@@ -3,7 +3,7 @@
  * @license MIT
  */
 
-import { Terminal as ITerminalApi, ITerminalOptions, IMarker, IDisposable, ILinkMatcherOptions, ITheme, ILocalizableStrings, ITerminalAddon, ISelectionPosition, IBuffer as IBufferApi, IBufferLine as IBufferLineApi, IBufferCell as IBufferCellApi, IParser, IFunctionIdentifier } from 'xterm';
+import { Terminal as ITerminalApi, ITerminalOptions, IMarker, IDisposable, ILinkMatcherOptions, ITheme, ILocalizableStrings, ITerminalAddon, ISelectionPosition, IBuffer as IBufferApi, IBufferLine as IBufferLineApi, IBufferCell as IBufferCellApi, IParser, IFunctionIdentifier, IBufferCellColor as IBufferCellColorApi } from 'xterm';
 import { ITerminal } from '../Types';
 import { IBufferLine } from 'common/Types';
 import { IBuffer } from 'common/buffer/Types';
@@ -12,6 +12,9 @@ import * as Strings from '../browser/LocalizableStrings';
 import { IEvent } from 'common/EventEmitter';
 import { AddonManager } from './AddonManager';
 import { IParams } from 'common/parser/Types';
+import { CellData } from '../../out/common/buffer/CellData';
+import { Attributes } from '../../out/common/buffer/Constants';
+import { AttributeData } from '../../out/common/buffer/AttributeData';
 
 export class Terminal implements ITerminalApi {
   private _core: ITerminal;
@@ -199,17 +202,24 @@ class BufferApiView implements IBufferApi {
     }
     return new BufferLineApiView(line);
   }
+  public getNullCell(): IBufferCellApi {
+    return new BufferCellApiView(new CellData());
+  }
 }
 
 class BufferLineApiView implements IBufferLineApi {
   constructor(private _line: IBufferLine) {}
 
   public get isWrapped(): boolean { return this._line.isWrapped; }
-  public getCell(x: number): IBufferCellApi | undefined {
+  public getCell(x: number, cell?: BufferCellApiView): IBufferCellApi | undefined {
     if (x < 0 || x >= this._line.length) {
       return undefined;
     }
-    return new BufferCellApiView(this._line, x);
+    if (cell) {
+      this._line.loadCell(x, cell.cell);
+      return cell;
+    }
+    return new BufferCellApiView(this._line.loadCell(x, new CellData()));
   }
   public translateToString(trimRight?: boolean, startColumn?: number, endColumn?: number): string {
     return this._line.translateToString(trimRight, startColumn, endColumn);
@@ -217,9 +227,42 @@ class BufferLineApiView implements IBufferLineApi {
 }
 
 class BufferCellApiView implements IBufferCellApi {
-  constructor(private _line: IBufferLine, private _x: number) {}
-  public get char(): string { return this._line.getString(this._x); }
-  public get width(): number { return this._line.getWidth(this._x); }
+  public flags: {[flag: string]: boolean};
+  public fg: IBufferCellColorApi;
+  public bg: IBufferCellColorApi;
+  constructor(public cell: CellData) {
+    this.flags = {
+      get bold(): boolean { return !!cell.isBold(); },
+      get underline(): boolean { return !!cell.isUnderline(); },
+      get blink(): boolean { return !!cell.isBlink(); },
+      get inverse(): boolean { return !!cell.isInverse(); },
+      get invisible(): boolean { return !!cell.isInvisible(); },
+      get italic(): boolean { return !!cell.isItalic(); },
+      get dim(): boolean { return !!cell.isDim(); }
+    };
+    this.fg = {
+      get colorMode(): "RGB" | "P256" | "P16" | "DEFAULT" {
+        return cell.getFgColorMode() === Attributes.CM_RGB ? 'RGB'
+            : cell.getFgColorMode() === Attributes.CM_P256 ? 'P256'
+            : cell.getFgColorMode() === Attributes.CM_P16 ? 'P16'
+            : 'DEFAULT'
+      },
+      get color(): number { return cell.getFgColor(); },
+      get rgb(): number[] { return AttributeData.toColorRGB(cell.getFgColor()); }
+    };
+    this.bg = {
+      get colorMode(): "RGB" | "P256" | "P16" | "DEFAULT" {
+        return cell.getBgColorMode() === Attributes.CM_RGB ? 'RGB'
+            : cell.getBgColorMode() === Attributes.CM_P256 ? 'P256'
+            : cell.getBgColorMode() === Attributes.CM_P16 ? 'P16'
+            : 'DEFAULT'
+      },
+      get color(): number { return cell.getBgColor(); },
+      get rgb(): number[] { return AttributeData.toColorRGB(cell.getBgColor()); }
+    };
+  }
+  public get char(): string { return this.cell.getChars(); }
+  public get width(): number { return this.cell.getWidth(); }
 }
 
 class ParserApi implements IParser {
diff --git a/typings/xterm.d.ts b/typings/xterm.d.ts
index cae54dc1..43c8fc0b 100644
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -878,6 +878,13 @@ declare module 'xterm' {
      * @param y The line index to get.
      */
     getLine(y: number): IBufferLine | undefined;
+
+    /**
+     * Creates an empty cell object suitable as a cell reference in
+     * `line.getCell(x, cell)`. Use this to avoid costly recreation of
+     * cell objects when dealing with tons of cells.
+     */
+    getNullCell(): IBufferCell;
   }
 
   /**
@@ -892,13 +899,10 @@ declare module 'xterm' {
     /**
      * Gets a cell from the line, or undefined if the line index does not exist.
      *
-     * Note that the result of this function should be used immediately after
-     * calling as when the terminal updates it could lead to unexpected
-     * behavior.
-     *
      * @param x The character index to get.
+     * @param cell Optional cell object to load data into.
      */
-    getCell(x: number): IBufferCell | undefined;
+    getCell(x: number, cell?: IBufferCell): IBufferCell | undefined;
 
     /**
      * Gets the line as a string. Note that this is gets only the string for the
@@ -911,6 +915,36 @@ declare module 'xterm' {
     translateToString(trimRight?: boolean, startColumn?: number, endColumn?: number): string;
   }
 
+  /**
+   * Represents foreground and background color settings of a cell.
+   */
+  interface IBufferCellColor {
+    /**
+     * Color mode of the color setting.
+     * RGB        Color is an RGB color, use `.rgb` to grab the different channels.
+     * P256       Color is an indexed value of the 256 color palette.
+     * P16        Color is an indexed value of the 8 color palette (+8 for AIX bright colors).
+     * DEFAULT    No color set, thus default color should be used.
+     */
+    colorMode: 'RGB' | 'P256' | 'P16' | 'DEFAULT';
+
+    /**
+     * Color value set in the current color mode.
+     * Note that the color value can only be interpreted in conjunction
+     * with the color mode:
+     * RGB      color contains 8 bit channels in RGB32 bitorder, e.g. red << 16 | green << 8 | blue
+     * P256     color contains indexed value 0..255
+     * P16      color contains indexed value 0..15
+     * DEFAULT  color always contains -1
+     */
+    color: number;
+
+    /**
+     * Helper to get RGB channels from color mode RGB. Reports channels as [red, green, blue].
+     */
+    rgb: number[];
+  }
+
   /**
    * Represents a single cell in the terminal's buffer.
    */
@@ -928,6 +962,21 @@ declare module 'xterm' {
      * - This is `0` for cells immediately following cells with a width of `2`.
      */
     readonly width: number;
+
+    /**
+     * Text attribute flags like bold, underline etc.
+     */
+    readonly flags: {[flag: string]: boolean};
+
+    /**
+     * Foreground color.
+     */
+    readonly fg: IBufferCellColor;
+
+    /**
+     * Background color.
+     */
+    readonly bg: IBufferCellColor;
   }
 
   /**
```

Changes:
- uses `CellData` internally to decouple cell view from bufferline and cell position
- optional argument `cell` for `line.getCell(x, cell)` to make ref style possible
- create an empty cell container with `term.buffer.getNullCell()`
- flags and colors are lazy eval'ed with getters, thus changing the underlying data in `CellData` will change these values as well (needed to make ref style possible here)
- `IBufferCellColor` does not change interface, instead it exposes color as a number for all color modes; to get the RGB channels separately it has an additional property `,rgb`

This API approach tries to be easy to use while still resembling some parts of the C-ish internals for performance reasons. Main pro here is the possibility to reuse a previous cell object, which will greatly lower the allocation + GC pressure. Not yet benchmarked though...

jerch:
The new API works, but the needed lazy evaluation of the different attributes makes it still slow (~5MB/s). Thus I added a few convenient compare methods that lift some burden from the long if cascades in serialize:
```typescript
// public API in xterm.d.ts
    /**
     * Whether cells have the same text attributes (flags and colors).
     * @param other Other cell.
     */
    equalAttibutes(other: IBufferCell): boolean;

    /**
     * Whether cells have the same text attribute flags.
     * @param other Other cell.
     */
    equalFlags(other: IBufferCell): boolean;

    /**
     * Whether cells have the same foreground color.
     * @param other Other cell.
     */
    equalFg(other: IBufferCell): boolean;

    /**
     * Whether cells have the same background color.
     * @param other Other cell.
     */
    equalBg(other: IBufferCell): boolean;

// impl in src/public/Terminal.ts
  public equalAttibutes(other: BufferCellApiView): boolean {
    return this.cell.fg === other.cell.fg
      && this.cell.bg === other.cell.bg;
  }
  public equalFlags(other: BufferCellApiView): boolean {
    return (this.cell.fg & fgFlagMask) === (other.cell.fg & fgFlagMask)
      && (this.cell.bg & bgFlagMask) === (other.cell.bg & bgFlagMask);
  }
  public equalFg(other: BufferCellApiView): boolean {
    return this.cell.getFgColorMode() === other.cell.getFgColorMode()
      && this.cell.getFgColor() === other.cell.getFgColor();
  }
  public equalBg(other: BufferCellApiView): boolean {
    return this.cell.getBgColorMode() === other.cell.getBgColorMode()
      && this.cell.getBgColor() === other.cell.getBgColor();
  }
```

With these additions we are back on the speedy track:
```
   Context "out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js"
      Context "Terminal: sh -c "ls -lR /usr/lib | lolcat -f""
         Context "serialize"
            Case "#1" : 1 - runtime: 136.77 ms
            Case "#1" : 1 - throughput: 12.36 MB/s
            Case "#1" : 2 - runtime: 105.10 ms
            Case "#1" : 2 - throughput: 16.09 MB/s
            Case "#1" : 3 - runtime: 103.72 ms
            Case "#1" : 3 - throughput: 16.30 MB/s
            Case "#1" : 4 - runtime: 104.93 ms
            Case "#1" : 4 - throughput: 16.11 MB/s
            Case "#1" : 5 - runtime: 105.25 ms
            Case "#1" : 5 - throughput: 16.06 MB/s
            Case "#1" : 6 - runtime: 104.70 ms
            Case "#1" : 6 - throughput: 16.15 MB/s
            Case "#1" : 7 - runtime: 104.43 ms
            Case "#1" : 7 - throughput: 16.19 MB/s
            Case "#1" : 8 - runtime: 103.45 ms
            Case "#1" : 8 - throughput: 16.34 MB/s
            Case "#1" : 9 - runtime: 103.90 ms
            Case "#1" : 9 - throughput: 16.27 MB/s
            Case "#1" : 10 - runtime: 103.89 ms
            Case "#1" : 10 - throughput: 16.27 MB/s
            Case "#1" : 10 runs - average runtime: 107.62 ms
            Case "#1" : 10 runs - average throughput: 15.82 MB/s
```

Source of my serialize implementation can be found here: https://gist.github.com/jerch/cfe4360e7f23df8bc7189c3761bd13f5 (Note that this is still not optiomized nor does it handle all weird circumstances.)

jerch:
Made a direct comparison of private access vs. public API:
image

The test was done with more typical `ls -lR /usr/` output (without lolcat) with 100k lines scrollback. The output is equal for `serialize` and `serializePrivate` and about 9.2 MB, thus throughput is 16.8 MB/s for `serialize` vs. 20.9 MB/s for `serializePrivate`. The same difference shows in the benchmark results with shilghty higher values (19.9 vs. 25 MB/s, prolly due to better optimization with multiple runs).

The public API sacrifice is 20 - 25% in performance overall, which is imho good for for the gains in simplicity (the API is much more JS-ish compared to the low level C-ish private stuff). Still the overhead for the getters is significant, `_extractAttributes` is 5 times slower than `_extractAttributesPrivate` (but only a small part overall).

The code can be viewed in my playground branch https://github.com/jerch/xterm.js/tree/serialize_with_private.

JavaCS3:
@jerch I'm trying to integrate your proposal. IMO this addon can not only for serialize terminal session, it can also be used to create a screenshot of a terminal (i.e serialize to SVG/PNG/HTML) for people to share. That should be cool

jerch:
> IMO this addon can not only for serialize terminal session, it can also be used to create a screenshot of a terminal (i.e serialize to SVG/PNG/HTML) for people to share.

Yes that would be nice. Imho the addon could have several "renderers" later on to select from like `.serialize(rows: number, renderer: 'HTML' | 'ESCString' | 'PNG' | ... | 'fancyOutputFormatXY' = 'ESCString')`. But I think we should get one use case straight first.

JavaCS3:
@jerch How to get that perf picture?

jerch:
@JavaCS3 Did that with devtools in the demo:
- register SerializeAddon in client.ts
- yarn start + open demo in chrome
- set scrollback to 100000
- run `ls -lR /usr` several times until the scrollbuffer is full (`term.buffer.length` should report 100026)
- place a longish setTimeout in console, e.g. `setTimeout(() => term._addonManager._addons[3].instance.serializePrivate(), 3000)`
- switch to timeline and start profiling within the timeout
- grab the timeline results

jerch:
@JavaCS3 Oh btw, forgot to test your flag copying with my API changes. If you care enough you can test copying flags over as well, it might squeeze a few cycles out of the flag access (mainly the `_extractAtttribues` part in my code).

JavaCS3:
@jerch I just found we can use `v8-profiler` with slightly code change

jerch:
@JavaCS3  Have never used this package, if you get it working that would be really nice.

With xterm-benchmark you can switch on a debugger break and attach devtools by doing this:
- run a single perfCase in forked mode and append the cmdline break option, e.g. `{fork: true, forkOptions: {execArgv: ['--inspect-brk']}}`
- run xterm-benchmark with -s switch (runs a single perfCase), it will halt execution at that perfCase waiting for a debugger to connect
- open URL `chrome://inspect` in chrome, there the waiting node process should appear in a list
- attach to that one and do whatever is needed like starting a profiling
- resume execution
- when it is finished, stop profiling and inspect the data (flame chart + tables)

If you really want to dive into this stuff I recommend reading through this article: https://nodesource.com/blog/why-the-new-v8-is-so-damn-fast/. Esp. the described `deoptigate` is very helpful to find problems in hot code (like objects with shifting shapes/interfaces). `serialize` will almost always qualify as "hot code" as it always runs the cell attributes logic in a tight loop, thus it might be worth to go this deep (after fixing more obvious higher level stuff). But keep in mind that hot code optimizations often leave idiomatic JS as it follows much stricter rules (welcome back to C-ish JS). This is the ground where you can gain alot but have to know where to stop (dont over optimize, some things wont apply to other engines at all). If in doubt go with Mr. Knuth here - idiomatic code is better in the long run than squeezing the last 10% performance out of it.

jerch:
@JavaCS3 Just found out that I had installed an old version of lolcat not capable to output truecolor. After installing as gem and switching truecolor on (-t switch), I get worse results with the current API. Seems the weird fg/bg objects with closured `cell` in getters is the culprit. Seems the API still needs some care...

jerch:
Pushed a new API version. Main changes:
- replaced property getters with `getXY()` methods
- removed `.flag`, `.fg` and `.bg` properties with its closured getters, instead all lives now flat on the cell object itself

The changes were needed due to my tests with truecolor cells created by lolcat. The old API with its closured getters got really slow after one or two rounds for no obvious reason - while the first invocation finishs after ~1200ms any later run takes 3000 - 4000 ms with the same data :(

`deoptigate` shows that the first run is optimized while any later one runs deoptimized, no clue why. Both xterm-benchmark (nodejs) and the demo in chrome showed that behavior. After removing the getters the runs are speedy and stable again.

@JavaCS3 Sorry for the constant API changes, but I think this is now in a pretty good shape and working more reliable.

@Tyriar It seems the getters are still not a good idea when it comes to hot code. I think we should deprecate the `char` and `width` getters on the cell as well and go with the method counterparts instead.

jerch:
@JavaCS3 Btw the benchmark runs with your code and the new API with 14.5 MB/s now which is almost 4 times faster (3.8 MB/s before).
Imho with cleaning up your flow control you can get it up to 18 - 20 MB/s. Your code contains a few antipattern like doubled condition checks, intermediate array creations and unnecessary class wide states. Also the base class approach seems overly complicated atm, I suggest to develop a concise single escapestring output variant first and extend alternative output handlers from that (bottom up).

JavaCS3:
> @JavaCS3 Btw the benchmark runs with your code and the new API with 14.5 MB/s now which is almost 4 times faster (3.8 MB/s before).
> Imho with cleaning up your flow control you can get it up to 18 - 20 MB/s. Your code contains a few antipattern like doubled condition checks, intermediate array creations and unnecessary class wide states. Also the base class approach seems overly complicated atm, I suggest to develop a concise single escapestring output variant first and extend alternative output handlers from that (bottom up).

@jerch 
1. For "the doubled condition checks": because `\x1b[0m` will reset everything(both fg/bg color and flags). So I need to make sure If I need reset a single flags/colors or reset everything. I think there's a better solution.
2. For "Intermediate array creations and unnecessary class wide states": it's not too hard to remove that.
3. For "base class styled approach": I think this serialize can not only be used for `ESCString` but also HTML, SVG and so on. So, I want to extract some base process template to deal with serializing stuffs. I think my current design is larged influnced by Java SAX XML parser. Anyway I'm open minded to change.

JavaCS3:
@jerch I've pushed a new update including a "v8-profiler-node" profileing tool ("v8-profiler" doesn't work for me due to compilinng error). With this you can directly run benchmark and get a profiling a file naming `serialize-profile.cpuprofile`. 
You then need to load it from Chrome's Javascript Profiler panel and you will get this.

profiler

I don't know why some functions call are missing (maybe samples is very few? You could increase the `dd` cmd `count` argument a little bit to give it more samples) and I'm not familiar with reading this profiling data currenlty. Maybe you could help.

jerch:
> For "the doubled condition checks": because \x1b[0m will reset everything(both fg/bg color and flags). So I need to make sure If I need reset a single flags/colors or reset everything.

Yeah right - my code does not respect this as my goal was the API rather than serialize itself.

> I think this serialize can not only be used for ESCString but also HTML, SVG and so on. So, I want to extract some base process template to deal with serializing stuffs. I think my current design is larged influnced by Java SAX XML parser.

I am not sure if a SAX inspired interface is a good choice here as it has very different goals/constraints. I have not thought about this in detail, just think that by starting with a working concrete impl it might be easier to develop the abstract parts from instead of starting with a slightly off interface. But you prolly have good reasons for that interface , so all good.


About the profiling data:
> I don't know why some functions call are missing (maybe samples is very few? You could increase the dd cmd count argument a little bit to give it more samples)

There are two reason why a call might not appear:
- sampling - The profiler is a sampling profiler, it peeks into JS callstack every x-th time and grabs the current active function, the stack path (function descent) and a few more stats. Finally this gets summed up and presented in these tables. If a function runs very short it is very likely to never get caught by this snapshotting. This can be avoided by running the profiling longer (which increases the chance that it got caught at least once). But note that very low values should not be trusted, they are to close to the measurement uncertainty.
- inlining - Short invocations might get inlined, thus will not show in the call stack at all. This is really good for hot code, as it even saves the function call overhead.

src/public/Terminal.ts
```diff
@@ -198,25 +201,89 @@ class BufferApiView implements IBufferApi {
...
+          case Attributes.CM_P16: return 'P16';
+          default: return 'DEFAULT';
+        }
+      },
+      get color(): number { return cell.getFgColor(); },
+      get rgb(): [number, number, number] { return AttributeData.toColorRGB(cell.getFgColor()); }
+    };
+    this.bg = {
+      get colorMode(): 'RGB' | 'P256' | 'P16' | 'DEFAULT' {
+        switch (cell.getFgColorMode()) {
```
Tyriar: Fg -> Bg?

Tyriar:
> @Tyriar It seems the getters are still not a good idea when it comes to hot code. I think we should deprecate the char and width getters on the cell as well and go with the method counterparts instead.

I see some throughput numbers above but how did that translate into actual usage of this addon for say the default scrollback buffer?

jerch:
~1200ms any later run takes 3000 - 4000 ms

> I see some throughput numbers above but how did that translate into actual usage of this addon for say the default scrollback buffer?

The number above (~1200ms first run, any later run 3000 - 4000 ms) are taken from demo with 100k scrollback. They scale linear thus translate to 12ms vs. 30-40 ms for default scrollback.

By removing the getters the weird deopt vanished for later runs, also the runtime got a little faster (~900ms or 9ms for default scrollback, not stated above).

Tyriar:
@jerch but these numbers come from your normal test cases right, not a real world scenario like actually serializing a real buffer? Is that was this was? https://github.com/xtermjs/xterm.js/pull/2369#issuecomment-526160555 ~400ms to serialize 100k scrollback? That seems totally reasonable to me since the typical case would be serializing just the viewport to 1k-10k lines which would result in 4-40ms serialize time.

jerch:
@Tyriar The numbers are the results of `ls -lR /usr | lolcat -t` in the demo with 100k scrollback. Thus it was a real scenario, but artificial in the sense that all cells with content had an RGB fg color (which hardly will be the case).
Compared to the numbers in https://github.com/xtermjs/xterm.js/pull/2369#issuecomment-526160555 it basically means, that
- first run with RGB cells takes 1200 ms (instead of 400-600 ms)
- any later run with RGB cells takes 3000 - 4000 ms (instead of 400-600 ms)

And without getters:
- any run with RGB cells takes 900 ms (instead of 400-600 ms)

Since all numbers scale almost linear the runtime for a smaller scrollbuffer is a fraction of those (`* actul_scrollbuffer_size / 100000`)

The problem I see here is the first run vs. later runs difference. With getters the addon is likely to run into some deopt, that results in 10 times longer runtimes (4000 vs. 400, or with a default scrollback size of 1k - 40 vs. 4 ms).

src/public/Terminal.ts
JavaCS3: Your right

Tyriar:
See response on getters vs functions: https://github.com/xtermjs/xterm.js/issues/2075#issuecomment-529615367

vincentwoo:
Definitely in favor of CellData coming with more data than char/width!

JavaCS3:
Hey Guys, I'm back. Please help me to review this! @jerch @Tyriar 
Do I need to squash all these commits into one single commit?
Looks like I rebased my branch some commits from master. What kind of merge strategy do u guys follow?

jerch:
@JavaCS3 We have no explicit merging/squashing strategy. I mostly try to deal with longish PRs this way:
- keep any commits that introduce real new value (hopefully with a meaningful commit message, not always the case for my commits :blush:)
- revert testing stuff or do partial squashs for "side experiments", that do not add up anything in final code

I would not worry to much here, unless you committed every single typo-fix separately.

JavaCS3:
Any comments?

Tyriar:
I updated the feature branch and fixed conflicts

Tyriar:
Looks like spaces (via \t?) are going missing:

Screen Shot 2019-11-24 at 10 04 59 AM

Screen Shot 2019-11-24 at 10 05 04 AM

Tyriar:
Made the API tests nice and fast:

```
  SerializeAddon
    ✓ empty content
    ✓ trim last empty lines
    ✓ digits content
    ✓ serialize half rows of content
    ✓ serialize 0 rows of content
    ✓ serialize all rows of content with color16
    ✓ serialize all rows of content with fg/bg flags
    ✓ serialize all rows of content with color256
    ✓ serialize all rows of content with color16 and style separately
    ✓ serialize all rows of content with color16 and style together
    ✓ serialize all rows of content with color256 and style separately
    ✓ serialize all rows of content with color256 and style together
    ✓ serialize all rows of content with colorRGB and style separately
    ✓ serialize all rows of content with colorRGB and style together


  14 passing (1s)
```

Tyriar:
Added a failing test for the tabs, the expected serialize value would probably ideally be moving the cursor over since `\t` depends on the tab size which can change.

demo/start.js
```diff
@@ -48,7 +48,8 @@ const clientConfig = {
     extensions: [ '.tsx', '.ts', '.js' ],
     alias: {
       common: path.resolve('./out/common'),
-      browser: path.resolve('./out/browser')
+      browser: path.resolve('./out/browser'),
+      xterm$: path.resolve('./out/public/Terminal.js')
```
Tyriar: What's this for?

demo/index.html
```diff
@@ -20,6 +20,12 @@ <h3>Actions</h3>
         <label>Case sensitive<input type="checkbox" id="case-sensitive"/></label>
         <label>Whole word<input type="checkbox" id="whole-word"/></label>
       </p>
+      <div>
+        <h3>SerializeAddon</h3>
+        <button id="serialize">Serialize the content of terminal</button>
+        <label><input type="checkbox" id="write-to-terminal">Write back to terminal</label>
+        <div><pre id="serialize-output"></pre></div>
+      </div>
```
Tyriar: I think this is a little too much for having so high in the demo. How about we move this and the find control down to inside the new "Addon" section below?

demo/style.css
```diff
@@ -30,3 +30,15 @@ p {
...
 }
+
+pre {
+    display: block;
+    padding: 9.5px;
+    font-size: 13px;
+    color: #c7254e;
+    background-color: #f9f2f4;
+    word-break: break-all;
+    word-wrap: break-word;
```
Tyriar: `word-wrap` is used twice

package.json
```diff
@@ -23,9 +23,9 @@
     "presetup": "node ./bin/install-addons.js",
     "prepublishOnly": "npm run package",
     "watch": "tsc -b -w ./tsconfig.all.json --preserveWatchOutput",
-    "benchmark": "NODE_PATH=./out xterm-benchmark -r 5 -c test/benchmark/benchmark.json",
-    "benchmark-baseline": "NODE_PATH=./out xterm-benchmark -r 5 -c test/benchmark/benchmark.json --baseline out-test/benchmark/test/benchmark/*benchmark.js",
-    "benchmark-eval": "NODE_PATH=./out xterm-benchmark -r 5 -c test/benchmark/benchmark.json --eval out-test/benchmark/test/benchmark/*benchmark.js",
+    "benchmark": "NODE_PATH=./out:./out-test/benchmark xterm-benchmark -r 5 -c test/benchmark/benchmark.json",
+    "benchmark-baseline": "NODE_PATH=./out:./out-test/benchmark xterm-benchmark -r 5 -c test/benchmark/benchmark.json --baseline out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js",
+    "benchmark-eval": "NODE_PATH=./out:./out-test/benchmark xterm-benchmark -r 5 -c test/benchmark/benchmark.json --eval out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js",
```
Tyriar: These need to change back

test/benchmark/SerializeAddon.benchmark.ts
```diff
@@ -0,0 +1,84 @@
+/**
+ * Copyright (c) 2019 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { perfContext, before, after, ThroughputRuntimeCase } from 'xterm-benchmark';
+
+import * as fs from 'fs'
+import { spawn } from 'node-pty';
+import * as profiler from 'v8-profiler-node8';
```
Tyriar: Can we remove this?

This is what happens when I try run baseline:

```
❯ yarn benchmark-baseline
yarn run v1.13.0
$ NODE_PATH=./out:./out-test/benchmark xterm-benchmark -r 5 -c test/benchmark/benchmark.json --baseline out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js
   Context "out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js"
      Context "Terminal: sh -c "dd if=/dev/random count=40 bs=1k | hexdump | lolcat -f""
         Context "serialize"
error Command failed with signal "SIGSEGV".
```

test/benchmark/SerializeAddon.benchmark.ts
```diff
@@ -0,0 +1,84 @@
+/**
+ * Copyright (c) 2019 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { perfContext, before, after, ThroughputRuntimeCase } from 'xterm-benchmark';
+
```
Tyriar: This file has a bunch of lint errors

src/public/Terminal.ts
```diff
@@ -201,27 +203,81 @@ class BufferApiView implements IBufferApi {
...
+const COLOR_MASK = Attributes.CM_MASK | Attributes.RGB_MASK;
+
 class BufferCellApiView implements IBufferCellApi {
-  constructor(private _line: IBufferLine, private _x: number) {}
-  public get char(): string { return this._line.getString(this._x); }
-  public get width(): number { return this._line.getWidth(this._x); }
+  constructor(public cell: CellData) {}
+
+  public get char(): string { return this.cell.getChars(); }
+  public get width(): number { return this.cell.getWidth(); }
```
Tyriar: Can we just remove this class all together now and return the CellData directly?

demo/client.ts
```diff
@@ -380,3 +387,16 @@ function updateTerminalSize(): void {
   terminalContainer.style.height = height;
   addons.fit.instance.fit();
 }
+
+function serializeButtonHandler(): void {
+  const output = addons.serialize.instance.serialize();
+  const outputString = JSON.stringify(output);
+  console.log('serialize output', outputString);
```
Tyriar: Can remove this since we add it to the DOM

package.json
JavaCS3: Ooops, It must be a mistake

test/benchmark/SerializeAddon.benchmark.ts
JavaCS3: OKAY

demo/client.ts
JavaCS3: Sure

src/public/Terminal.ts
JavaCS3: Let me try it out

demo/start.js
JavaCS3: It used to be fixing some importing error since the demo page cannot find out what `xterm` package is. Let me try to remove it.

JavaCS3:
@Tyriar A silly question: Looks like you appended some commits onto this PR. How can I continue update it based on yours? I guess if I push updates to my forked repo will remove your changes.

JavaCS3:
> Looks like spaces (via \t?) are going missing:
> 
> Screen Shot 2019-11-24 at 10 04 59 AM
> 
> Screen Shot 2019-11-24 at 10 05 04 AM

I logged each cell's char and I found some continuous empty char. Looks like I need some other ways to deal with this case

image

Tyriar:
> A silly question: Looks like you appended some commits onto this PR. How can I continue update it based on yours? I guess if I push updates to my forked repo will remove your changes.

Yep, I just added some commits to your fork's branch, just pull and add new commits on top like normal.

> I logged each cell's char and I found some continuous empty char. Looks like I need some other ways to deal with this case

This in `_nextCell` or something similar is probably the problem:

```ts
this._currentRow += cell.char;
```

There's a difference between whitespace and "null" cells, when `\t` is send to the terminal it moves the cursor instead of printing `' '` a bunch of times. I think we should be counting the number of spaces and serialize this using this sequence:

> CSI Ps C  Cursor Forward Ps Times (default = 1) (CUF).

So count the number of spaces and send `\u001b[4C` for example for 4 moves.

addons/xterm-addon-serialize/src/SerializeAddon.api.ts
```diff
@@ -17,42 +16,30 @@ const height = 600;
 
 describe('SerializeAddon', () => {
   before(async function (): Promise<any> {
-    this.timeout(20000);
```
JavaCS3: @Tyriar Looks like my machine is slow, I still need to keep this timeout but maybe around 8 secs.

package.json
JavaCS3: Oh I think this is necessary because this is the first time we do benchmark on Addons. Otherwise it can't find that addon. Maybe u have better idea.
```
$ yarn benchmark-baseline
yarn run v1.17.3
$ NODE_PATH=./out xterm-benchmark -r 5 -c test/benchmark/benchmark.json --baseline out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js
{ Error: Cannot find module 'addons/xterm-addon-serialize/src/SerializeAddon'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/Users/javacs3/Lab/Playground/xterm.js/out-test/benchmark/test/benchmark/SerializeAddon.benchmark.js:20:26)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3) code: 'MODULE_NOT_FOUND' }
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

src/public/Terminal.ts
JavaCS3: @Tyriar I found that I extended CellData's original API like. If I directly return CellData, I then need to add those extended API to CellData. Is that OK?
```
    equalFg(cell: IBufferCell): boolean;
    equalBg(cell: IBufferCell): boolean;
    equalFlags(cell: IBufferCell): boolean;
    isFgPalette16(): boolean;
    isBgPalette16(): boolean;
    isFgPalette256(): boolean;
    isBgPalette256(): boolean;
    isAttributeDefault(): boolean;
```

addons/xterm-addon-serialize/src/SerializeAddon.api.ts
```diff
@@ -17,89 +16,254 @@ const height = 600;
...
+    assert.equal(await page.evaluate(`serializeAddon.serialize();`), lines.join('\r\n'));
+  });
+
+  it('serialize tabs correctly', async () =>  {
+    const lines = [
+      'a\tb',
+      'foo\tbar\tbaz'
+    ];
+    await writeSync(page, lines.join('\\r\\n'));
+    assert.equal(await page.evaluate(`serializeAddon.serialize();`), lines.join('\r\n'));
```
JavaCS3: I think this case might be hard to cover because each `\t`'s width is different. Do u have some better ideas? @Tyriar

jerch: Yeah TABs are a nightmare to deal with, its not clear whether they should be treated as cursor moving control code only or should find its way back into string representation.
Btw the canonical line editor of linux's line discipline solves this issue by translating TABs into modulo 8 whitespaces right from the beginning (thus the terminal would not see a TAB). But thats not true for any ~ICANON input.

@Tyriar Should we address this explicitly? (we already had requests for this) VTE tracks TABs somehow, I think the impl will be rather ugly, maybe something like a TAB cell flag where a TAB will clear cells covered by it. Overwritten cells should clear the TAB flag range again. During serialization (and c&p in general) this flag can be eval'ed and place a '\t' instead of whitespaces.

@JavaCS3 In the meantime it makes not much sense to test TAB serialization beside the modulo whitespace approach. (Ofc this will fail, if the tabwidth was changed in between.)

addons/xterm-addon-serialize/src/SerializeAddon.api.ts
Tyriar: That doesn't seem right, all tests run in under a second for me
Screen Shot 2019-12-07 at 10 03 55 AM

addons/xterm-addon-serialize/src/SerializeAddon.api.ts
Tyriar: I pushed a change to support serializing tabs, just needed to count the number of null cells between non-null cells and move the cursor

src/public/Terminal.ts
```diff
@@ -201,27 +203,81 @@ class BufferApiView implements IBufferApi {
...
+  public getFgColorMode(): number { return this.cell.getFgColorMode(); }
+  public getBgColorMode(): number { return this.cell.getBgColorMode(); }
+  public isFgRGB(): boolean { return this.cell.isFgRGB(); }
+  public isBgRGB(): boolean { return this.cell.isBgRGB(); }
+  public isFgPalette(): boolean { return this.cell.isFgPalette(); }
+  public isBgPalette(): boolean { return this.cell.isBgPalette(); }
+  public isFgPalette16(): boolean { return this.cell.getFgColorMode() === Attributes.CM_P16; }
+  public isBgPalette16(): boolean { return this.cell.getBgColorMode() === Attributes.CM_P16; }
+  public isFgPalette256(): boolean { return this.cell.getFgColorMode() === Attributes.CM_P256; }
+  public isBgPalette256(): boolean { return this.cell.getBgColorMode() === Attributes.CM_P256; }
```
Tyriar: I don't think we need this differentiation, can't we just check if `color` is < 16 in Serialize Addon here?

```ts
else if (cell.isFgPalette256()) { sgrSeq.push(38, 5, color); }
else if (cell.isFgPalette16()) { sgrSeq.push(color & 8 ? 90 + (color & 7) : 30 + (color & 7)); }
```

to:

```ts
else if (cell.isFgPalette()) {
  if (color >= 16) {
    sgrSeq.push(38, 5, color); }
  } else {
    sgrSeq.push(color & 8 ? 90 + (color & 7) : 30 + (color & 7));
  }
}
```

src/public/Terminal.ts
```diff
@@ -201,27 +203,81 @@ class BufferApiView implements IBufferApi {
...
+  public equalFlags(other: BufferCellApiView): boolean {
+    return (this.cell.fg & FG_FLAG_MASK) === (other.cell.fg & FG_FLAG_MASK)
+      && (this.cell.bg & BG_FLAG_MASK) === (other.cell.bg & BG_FLAG_MASK);
+  }
+  public equalFg(other: BufferCellApiView): boolean {
+    return (this.cell.fg & COLOR_MASK) === (other.cell.fg & COLOR_MASK);
+  }
+  public equalBg(other: BufferCellApiView): boolean {
+    return (this.cell.bg & COLOR_MASK) === (other.cell.bg & COLOR_MASK);
+  }
```
Tyriar: Instead of adding these can we just compare color mode/color and compare flags using `isUnderline`, etc.? I wouldn't this that would impact performance that much.

src/public/Terminal.ts
Tyriar: Responded in new review

package.json
Tyriar: But doesn't hardcoding SerializeAddon make it not work on the normal benchmarks anymore?

src/public/Terminal.ts
JavaCS3: u r right

addons/xterm-addon-serialize/src/SerializeAddon.api.ts
JavaCS3: I think I know what's going on here. I usually directly run SerializeAddon.api.test and it would take 5 secs bringing Chromium up. I think I can remove that timeout here.

package.json
JavaCS3: Sorry, I don't get it. I didn't hardcode anything in my opinion.

src/public/Terminal.ts
JavaCS3: @Tyriar When I was trying to follow your suggestion I found a problem. Basically, Color palette256 contains palette16.
So in my test case "serialize all rows of content with color256", your input  `\x1b[38;5;0m` will be serialized into `\x1b[30m` instead which is a little bit different from the original color.
I don't know if the color code 0-16 in palette256 is equal to the color in palette16 all the time.
My test case is designed to make serialize output as closer as the original input.
What's your opinion?

jerch: @JavaCS3 Imho you cannot assume the lower 16 colors of P256 being the same as the P16 colors. ECMA-48 does not state any linkage between those palettes, guess it should be treated as implementation/emulator dependent (I think some emulators short-circuit those colors internally).
This becomes interesting for theming - we currently only support theming of the P16 colors, here it seems arguable whether to apply those to P256 as well (we currently dont and I think it should stay that way). Also xterm supports custom colors to be set via an OSC command (we dont support this yet), I think those would not change the P8/P16 likewise (have not tested it).

JavaCS3:
@Tyriar Here's the new update

package.json
Tyriar: @jerch see https://github.com/xtermjs/xterm.js/pull/2369#discussion_r356092514, I'm a little uneasy having the addon name in the package.json here. Any thoughts on how this should be structured?

jerch: Hmm yeah, that should not be here. Maybe move it under `addon-xy/test/...` in a similar fashion as done on main repo level? Getting tsconfig to correctly build things there might be complicated thought (until we have everything modularized)...

Tyriar: :+1: that way you can do `cd addon/ && yarn benchmark` or `yarn --cwd addon benchmark`

JavaCS3: @Tyriar mission complete 😆

JavaCS3:
mission complete

src/common/buffer/Constants.ts
```diff
@@ -116,13 +116,23 @@ export const enum FgFlags {
   BOLD = 0x8000000,
   UNDERLINE = 0x10000000,
   BLINK = 0x20000000,
-  INVISIBLE = 0x40000000
+  INVISIBLE = 0x40000000,
+
+  /**
+   * bit 27..31 (32th bit unused)
+   */
+  FM_MASK = 0x7C000000
```
Tyriar: These FM_MASK values aren't referenced anywhere anymore

typings/xterm.d.ts
```diff
@@ -1011,6 +1020,35 @@ declare module 'xterm' {
      * - This is `0` for cells immediately following cells with a width of `2`.
      */
     readonly width: number;
```
Tyriar: Let's add these 2 messages for char and width:

```
* @deprecated use `getChars` instead

* @deprecated use `getWidth instead
```

src/public/Terminal.ts
```diff
@@ -201,27 +203,65 @@ class BufferApiView implements IBufferApi {
...
     if (x < 0 || x >= this._line.length) {
       return undefined;
     }
-    return new BufferCellApiView(this._line, x);
+
+    if (cell) {
+      this._line.loadCell(x, cell.cell);
+      return cell;
+    }
+    return new BufferCellApiView(this._line.loadCell(x, new CellData()));
```
Tyriar: This needs to change to be just:

```ts
return this._line.loadCell(x, new CellData());
```

In order to get all the perf savings we're after, no point wrapping the object if they're not essentially equivalent. From what @jerch said earlier I think we can just check the color value when the mode is palette inside the addon instead of adding a new method to the API:

> Thus its not possible to address the lower 16 colors in P256 (as it always maps to P16).

typings/xterm.d.ts
JavaCS3: @Tyriar I need to either remove these interface in `xterm.d.ts` because `ICellData` doesn't have `width`, `char` interface if you want to remove `BufferCellApiView` OR add `width`, `char` interface in `ICellData` and fix all the implementations of `ICellData`.

Tyriar: Good point, let's remove them then since Terminal.buffer is marked experimental

src/public/Terminal.ts
JavaCS3: @Tyriar See the latest update

JavaCS3:
@Tyriar Please check the latest update

Issue #2363, opened 2019-08-06
Support fast scrolling using a modifier key


Tyriar:
Downstream: https://github.com/microsoft/vscode/issues/78431

VS Code has a setting `editor.fastScrollSensitivity` which multiplies the scroll speed when a particular modifier is down. This should be pretty easy to implement, we need to think about how the API should look. We probably need a number for the multiplier and an enum for which modifier (see https://github.com/microsoft/vscode/pull/78555).

Tyriar:
Proposal:

```
fastScrollSensitivity: number
fastScrollModifier: 'alt' | 'ctrl' | 'shift'
```

csurquin:
Hello, 

I am looking for a good first issue to get started in open source projects and this one seems doable. May I try to tackle this? Do you have any hint about where to start?

Thank you!

Tyriar:
@csurquin sure, this is a good one as it's pretty well defined. You should be able to find all the places you need to add a setting for by following one of the existing settings, eg.

https://github.com/xtermjs/xterm.js/blob/a89771a8de58fbc7a82323fb7d6d9c421ef63535/src/common/services/OptionsService.ts#L25

I think these are only the places where the scroll would need to be multiplied:

https://github.com/xtermjs/xterm.js/blob/a89771a8de58fbc7a82323fb7d6d9c421ef63535/src/browser/Viewport.ts#L171-L209

csurquin:
I made a [pull request](https://github.com/xtermjs/xterm.js/pull/2372) for this feature. I hope I did it the right way.

Tyriar:
Fixed in https://github.com/xtermjs/xterm.js/pull/2375

PR #2372, opened 2019-08-09
Support fast scrolling using a modifier key #2363

csurquin:
Here is my attempt to implement this [feature](https://github.com/xtermjs/xterm.js/issues/2363). The sensitivity is adjustable and the modifier key can be chosen from 'alt', 'ctrl', 'shift' or undefined to disable the feature.

I did not add automated tests because I was not sure how to do it but I may work on it later following your feedback.

### DIFF
--- a/src/renderer/dom/DomRenderer.ts
+++ b/src/renderer/dom/DomRenderer.ts
@@ -165,7 +165,13 @@ export class DomRenderer extends Disposable implements IRenderer {
         `}`;
     // Blink animation
     styles +=
-        `@keyframes blink {` +
+        `@keyframes blink_box_shadow {` +
+        ` 50% {` +
+        `  box-shadow: none;` +
+        ` }` +
+        `}`;
+    styles +=
+        `@keyframes blink_block {` +
         ` 0% {` +
         `  background-color: ${this._colors.cursor.css};` +
         `  color: ${this._colors.cursorAccent.css};` +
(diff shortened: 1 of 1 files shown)


demo/client.ts
```diff
@@ -228,6 +228,7 @@ function initOptions(term: TerminalType): void {
     bellSound: null,
     bellStyle: ['none', 'sound'],
     cursorStyle: ['block', 'underline', 'bar'],
+    fastScrollModifier: ['alt', 'ctrl', 'shift', undefined],
```
Tyriar: Not sure undefined will work here? This drives the dropdown in the demo

csurquin: I work on the development container and it seems to behave as expected. Here is what is displayed in the demo.

demo
If I select undefined, then none of the modifier keys activates the fast scrolling.

csurquin:
I will try to implement a test as soon as possible. Since it is my first pull request I have a question, I read that it is possible to update a pull request by pushing additional commits on the branch you submitted. However, I naively pushed my commits on `master`. Will it work the same way?

demo/client.ts
Tyriar: Perfect 👌

Tyriar:
@csurquin you can just push to master, you just won't be able to do another PR without extra hoops. I'd suggest finishing this one off first, then in the future checkout a branch and submit any more against that (`git checkout -b myfeature`)

PR #2375, opened 2019-08-10
Support fast scrolling using a modifier key

csurquin:
This is a follow up on the closed pull request #2372 as an attempt to resolve issue #2363. This time I pushed the changes to a new branch and rebased it onto `master`. 

As discussed in the closed pull request, I am currently working on a test for this feature. However, I am not sure to understand the structure of the test suite. As mentioned [here](https://github.com/xtermjs/xterm.js/blob/master/CONTRIBUTING.md#contributing-code), tests go into the `test` directory but there exist test files outside of this directory. Should I edit an existing file or create a new one?


Tyriar:
Put the test in here: https://github.com/xtermjs/xterm.js/blob/master/test/api/Terminal.api.ts

csurquin:
Here are some news about the ongoing work. 

First of all, I have some trouble launching the test suite on the development container. Here are the [logs](https://pastebin.com/cgwc4sTr), it seems that libraries are missing. If needed, I can open an issue about this.

I have some ideas for the test but I encountered the problem above. If there is no immediate solution to it, I will try to run the demo in a different way. 

By reviewing the code, I thought of two potential improvements:
1) at the moment, the fastScrollSensitivity parameter is unbound. This implies three unwanted behaviours:
    - a zero value will prevent the terminal from scrolling causing the page to scroll instead;
    - a negative value will reverse the scrolling;
    - the value 1 will act as default scrolling;

    Furthermore, it may not be pertinent that a positive value goes beyond a certain threshold. I think that limiting the range could be more user-friendly, maybe between 2 and 10.

2) I am not sure about this but the method `Viewport.getLinesScrolled` might need to be modified to reflect the new feature.

Tyriar:
> at the moment, the fastScrollSensitivity parameter is unbound. This implies three unwanted behaviours:

Let's make 1 the minimum value for the setting to avoid these issues.

> I am not sure about this but the method Viewport.getLinesScrolled might need to be modified to reflect the new feature.

`getLinesScrolled` is used for sending mouse events so I think we should use it here as well. To test this you can enable mouse mode in tmux and scroll with the wheel.

Tyriar:
Dev container issue: https://github.com/xtermjs/xterm.js/issues/2379

PR #2384, opened 2019-08-16
PR: Update fit-addon for supporting 3.14.5 xterm version

### LINKED ISSUE #2330: xterm 3.14.5 is not compatible with xterm-addon-fit 0.1.0 version
I am using the first xterm-addon-fit version which 0.1.0 in which `renderService` is used instead of `renderCoordinate`. But xterm 3.14.5 (latest release) still contains old code. 
Commit: https://github.com/xtermjs/xterm.js/commit/cbfbe232d00fd3ff6acb622b3b5469d7f882f731

You need to release a minor version of xterm to accommodate these changes.

## Details
- Browser and browser version: NA
- OS version: NA
- xterm.js version: 3.14.5

steff456:
Minor fix for making the fit addon compatible with the new version of xterm and the new addon API.

Fixes #2330

### DIFF
--- a/addons/xterm-addon-fit/src/FitAddon.ts
+++ b/addons/xterm-addon-fit/src/FitAddon.ts
@@ -71,8 +71,8 @@ export class FitAddon implements ITerminalAddon {
     const availableHeight = parentElementHeight - elementPaddingVer;
     const availableWidth = parentElementWidth - elementPaddingHor - core.viewport.scrollBarWidth;
     const geometry = {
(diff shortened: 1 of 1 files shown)


Tyriar:
This would break compat with master. The release of the new addons wasn't handled the best but we plan on making it better for future releases. In the meantime you should try out `xterm-addon-fit@0.1.0-beta2` (or beta1 if that doesn't work) which uses coordinator.

PR #2392, opened 2019-08-23
fix bugs in Params

### LINKED ISSUE #2389: Params.addSubParam not cancelled correctly
Repro - input in demo JS console:
- `term.parser.addCsiHandler({final: 'z'}, console.log)`
- `term.write('\x1b[;;;;;;;;;10;;;;;;;;;;20;;;;;;;;;;30;31;32;33;34;35::::::::z')`

This incorrectly appends the subparams of the 35-th param to the 32-th param. Basically if the param space is exhausted no subparams should be appendable. Prolly only a minor fix is needed, also check if an attempt to add another params beyond maxLength was made in https://github.com/xtermjs/xterm.js/blob/5a9e6fe94b17f8a27938a3c26ece0056cd8f6976/src/common/parser/Params.ts#L156-L159

jerch:
Fixes #2389.

Behavior affected:
- cancel digits for params and subparams if space is exhausted (part 1 of #2389)
- save isSub state so split input goes in to the right slot (part 2 of #2389)
- correctly handle -1 as initial value in digit input (was only respected for subparams before)
- do not try to add digit if no slot was taken before (was only respected for subparams before)

### DIFF
--- a/src/common/parser/Params.ts
+++ b/src/common/parser/Params.ts
@@ -212,7 +212,6 @@ export class Params implements IParams {
   /**
    * Add a single digit value to current parameter.
    * This is used by the parser to account digits on a char by char basis.
-   * Do not use this method directly, consider using `addParam` or `addSubParam` instead.
    */
   public addDigit(value: number): void {
     let length;
--- a/src/common/parser/Params.ts
+++ b/src/common/parser/Params.ts
@@ -41,6 +41,7 @@ export class Params implements IParams {
   private _subParamsIdx: Uint16Array;
   private _rejectDigits: boolean;
   private _rejectSubDigits: boolean;
+  private _digitIsSub: boolean;
 
   /**
    * Create a `Params` type from JS array representation.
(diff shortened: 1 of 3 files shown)


src/common/parser/Params.ts
```diff
@@ -204,30 +212,19 @@ export class Params implements IParams {
   /**
    * Add a single digit value to current parameter.
    * This is used by the parser to account digits on a char by char basis.
-   * Do not use this method directly, consider using `addParam` instead.
+   * Do not use this method directly, consider using `addParam` or `addSubParam` instead.
```
Tyriar: This comment seems to be saying you shouldn't use this and it should be private, but it's used in EscapeSequenceParser.ts?

jerch: Yes thats meant as `friend` in C++. Not sure how to express this in TS.

The method is not part of the interface by purpose as it skips boundary check due to being hotcode. The parser uses it but ensures by other means, that it stay within bounds.

Tyriar: Can this be something that is public on `Params` but not on `IParams` and only `EscapeSequenceParser` uses the concrete `Params`?

jerch: Thats how it is atm?

Tyriar: Cool 🙂 we don't really need the comment then but up to you. That's the closest to friend we can get afaik

PR #2393, opened 2019-08-24
Support for window options

### LINKED ISSUE #1762: Support for Extended Window Manager Hints
When using xterm.js for Windows SAC sessions, the input line stays fixed in place without adjusting to the size of the terminal (screenshot below). This seems to be because xtermjs doesn't seem to support a `Esc  + [18t` control sequence, which is emitted by SAC to determine the screen size. Order of events is as follows:
1. SAC emits `ESC + [18t` (`\x1b[18t`)
1. The terminal emulator is expected to reply with `ESC + [8;RRR;CCCt` (`\x1b[8;RRR;CCCt`)
    * RRR is the number of rows, and CCC is the number of columns
    * This must be sent by the terminal emulator within 3 seconds

Looks like in order for xtermjs to support Windows SAC resizing, there will need to be support added for the `ESC + [18t` control sequence. Can this be added?

image

jerch:
PR for support of window option reports and manipulations (CSI ... t).

To get this working we have several things to fiddle out:
- reporting: Its not yet clear which size should be reported where. My current guess is that text area is meant for exact cell * row (pixel) size while our closest match as "window" would be the outer terminal container. Needs investigation.
- separation of concerns: Some parts of CSI ... t address the offscreen parts, other are meant as "window" thingy thus need calls into browser related code. This might need some internal helpers to avoid tons of `_private._private` access. Resizing itself would need something like the fit addon in the codebase.
- features to support: Fullscreen stuff seems questionable to me, some commands dont apply at all as xterm.js typically doesnt run in a windowed env. Still standalone apps like hyper might benefit from interfaces exposing those things. Needs to be discussed. 
- security: Many of the commands are seen as instrusive as they can leak/manipulate things on the terminal's host machine from application side. To not earn a CVE here, we need to decide whether to:
  - support feature xy at all
  - hide feature behind ctor options like `options.allowTitleOps`, `options.allowGeometryOps` and such so ppl/integrators are still in control over those features

The early stub only contains a hacked version of grid size report and resize working with the demo.

Closes #1762.


jerch:
@Tyriar This turns out to be a bigger construction site than initially thought, parts of the features xterm offers under this sequence are deeply linked to X11 stuff (thus for xterm.js use cases prolly platform window manager dependent). I think we should not try to implement things here on our own, maybe just forward it via some API so integrators can fill the gaps they want to support. Will see if I can come up with an easy to intercept interface.

src/InputHandler.ts
```diff
@@ -1919,6 +1920,102 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+          const term = Function('return this.term;')();
+          // move fit into browser codebase?
+          for (const addon of (term as any)._addonManager._addons) {
+            if (`${addon.instance.constructor}`.indexOf('FitAddon') !== -1) {
+              addon.instance.fit();
+            }
+          }
+        }
+        break;
+      case 18:
```
Tyriar: Maybe we should just ship with 18t to close the issue off? Seems like a pretty easy win

src/InputHandler.ts
```diff
@@ -1919,6 +1920,102 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+          const rows = Math.max(2, params.params[1]);
+          // make it work with demo thus shamelessly taken from demo's client.ts
+          const width = (cols * (this._terminal as any)._renderService.dimensions.actualCellWidth + this._terminal.viewport.scrollBarWidth).toString() + 'px';
+          const height = (rows * (this._terminal as any)._renderService.dimensions.actualCellHeight).toString() + 'px';
+          (this._terminal as any)._parent.style.width = width;
+          (this._terminal as any)._parent.style.height = height;
+          const term = Function('return this.term;')();
+          // move fit into browser codebase?
+          for (const addon of (term as any)._addonManager._addons) {
+            if (`${addon.instance.constructor}`.indexOf('FitAddon') !== -1) {
```
Tyriar: 😱

src/InputHandler.ts
```diff
@@ -1919,6 +1920,102 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+   *    Ps = 23 ; 0  -> Restore xterm icon and window title from stack.   TBD
+   *    Ps = 23 ; 1  -> Restore xterm icon title from stack.              TBD
+   *    Ps = 23 ; 2  -> Restore xterm window title from stack.            TBD
+   *    Ps >= 24                                                          TBD
+   *    Resize to Ps lines (DECSLPP), VT340 and VT420. xterm adapts this by resizing its window.
+   */
+  public manipulateWindowOptions(params?: IParams): void {
+    console.log(params);
+    switch (params.params[0]) {
+      case 8:
```
Tyriar: I think introducing `8t` doesn't make as much sense in a web context and could cause issues if the application doesn't expect it. I suggest we don't bother thinking about this until we need to, and an addon might be the way to do it if someone every wants it?

src/InputHandler.ts
jerch: Yes, my current thinking is - implement the low hanging fruits like the reports we actually have all bits and leave the others. If an integrator wants some of the other functionality they can add it with `addCsiHandler` in their platform dependent way.

src/InputHandler.ts
jerch: As legit as placing a long jump in a precompiled executable :smile:

src/InputHandler.ts
jerch: Well resizing seems to  be an issue for azure integration as #1762 indicates. But I think we better dont try to mess with this at xterm.js level directly, instead integrators can use `addCsiHandler` and fill in missing bits respecting their platform needs.

jerch:
@Tyriar Reviewed the sequence again and indeed - it makes not much sense to build a sub API for all those platform dependent things, instead we should only deal with reports we can directly handle and leave all other things to a custom `addCsiHandler` implementation.

Things we can directly support/handle from within xterm.js:
- report textarea/window size in pixels (Ps 14)
- report cell pixel size (Ps 16)
- report textarea size in cols/rows (Ps 18)
- report window icon label (Ps 20)
- report window title (Ps 21)
- title and label stack handling (Ps 22/23)

Things left to integrators:
- iconify (Ps 1 / 2)
- window move (Ps 3)
- window resize (Ps 4)
- window stack handling (Ps 5 / 6)
- window refresh (Ps 7)
- textarea resize (Ps 8)
- maximize / fullscreen handling (Ps 9 / 10)
- report window state (Ps 11)
- report window/textarea position (Ps 13)
- report screen size in pixels (Ps 15)
- report screen size in cols/rows (Ps 19)
- line resize (Ps >= 24)

As you can see there is not much left to implement for us atm. To help integrators to get the other things done, we might need to implement some helpers on public API (like directly exposing some container metrics and such), but we can do that as it gets requested.
Gonna cleanup the PR accordingly.

Sidenote: I added icon label and window title handling on our side. Imho we can provide these features with a tiny implementation as we always know the propagated values from the OSC calls. There is a minor downside to this - if an implementation does not adopt the `titleChanged` event in a certain way the visual representation might be different to what gets reported here. Imho thats not a biggie as an emulator should care for title changes if it puts the title somewhere. xterm.js integrations that dont deal with the title will just keep working as expected.

jerch:
@Tyriar Updated the PR with implementations for params we actually can handle without further knowledge about the embedding env.

About security: Put all handling behind an option `allowedWindowOps` derived from xterm names. It is currently a comma separated string with either option names or param numbers, that gets sanitized during setting (`setOption('allowedWindowOps', 'RestoreWin, 16, SetWinLines')` ==> `[1, 16, 24]`). This option gate is also active for custom `CSI Ps ; Ps ; Ps t` handlers registered later on, thus a handler for a missing window option in that list will never be called.

About DECCOLM: I also put DECCOLM handling behind `'SetWinLines'` (24). That option is meant for DECSLPP which allows a more fine-grained column setting. We dont support that yet, still DECCOLM is just an older version of that with 80 vs. 132 column support.

Need some help with that line: https://github.com/xtermjs/xterm.js/blob/da3ae79c37932271e322f3c7c4b125b07cb06ec9/src/InputHandler.ts#L2084

What is the supposed way to get a hold of the pixel dimensions? Also the values I return under 14/16 might not be right ones. Please have a look at this.

src/InputHandler.ts
```diff
@@ -2017,6 +2046,97 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+    if (!~this._optionsService.options.allowedWindowOps.indexOf(params.params[0])) {
+      return;
+    }
+    const second = (params.length > 1) ? params.params[1] : 0;
+    const rs = (this._terminal as any)._renderService;  // FIXME: import renderService to get rid of any type here
+    switch (params.params[0]) {
+      case 14:  // GetWinSizePixels, returns CSI 4 ; height ; width t
+        if (rs && second !== 2) {
+          const w = rs.dimensions.canvasWidth.toFixed(0);
+          const h = rs.dimensions.canvasHeight.toFixed(0);
```
Tyriar: I think we want to use scaledCanvas* here as that will result in real pixels, not CSS pixels.

src/InputHandler.ts
```diff
@@ -2017,6 +2046,97 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+          this._coreService.triggerDataEvent(`${C0.ESC}[4;${h};${w}t`);
+        }
+        break;
+      case 16:  // GetCellSizePixels, returns CSI 6 ; height ; width t
+        if (rs) {
+          const w = rs.dimensions.actualCellWidth.toFixed(0);
```
Tyriar: Same here, I think `scaledCell*` is right

src/common/services/OptionsService.ts
```diff
@@ -128,6 +157,20 @@ export class OptionsService implements IOptionsService {
           throw new Error(`${key} cannot be less than or equal to 0, value: ${value}`);
         }
         break;
+      case 'allowedWindowOps':
+        const values = (value as string).split(',');
```
Tyriar: How about something more structured, similar to how theme works?

```ts
interface ITerminalOptions {
	// Better name?
	parserCapabilities: IParserCapabilities
}

interface IParserCapabilies {
	// Split by sequence type, eg. CSI Ps ; Ps ; Ps t below
	readonly windowManipulation?: {
		readonly restoreWin?: boolean;
		readonly minimizeWin?: boolean;
		readonly setWinPosition?: boolean;
		readonly setWinSizePixels?: boolean;
		readonly raiseWin?: boolean;
		readonly lowerWin?: boolean;
		readonly refreshWin?: boolean;
		readonly setWinSizeChars?: boolean;
		readonly maximizeWin?: boolean;
		readonly fullscreenWin?: boolean;
		readonly getWinState?: boolean;
		readonly getWinPosition?: boolean;
		readonly getWinSizePixels?: boolean;
		readonly getScreenSizePixels?: boolean;
		readonly getCellSizePixels?: boolean;
		readonly getWinSizeChars?: boolean;
		readonly getScreenSizeChars?: boolean;
		readonly getIconTitle?: boolean;
		readonly getWinTitle?: boolean;
		readonly pushTitle?: boolean;
		readonly popTitle?: boolean;
		readonly setWinLines?: boolean;
	}
}
```

This also got me thinking, I haven't liked the stringly-typed setOption and getOption for a while, we should just expose a readonly `options` object whose properties have getters and setters. https://github.com/xtermjs/xterm.js/issues/2522

jerch: Changed the option type to a const enum internally, that gets expanded for the outer API.

jerch:
@Tyriar Changed the windowOptions into a const enum for internal usage (number). The external API turns it into a `{[key: string]: boolean}` mapping. Sadly we dont support different types for internal vs. external options in `OptionService`, thus I put the internal representation on the service object for now. Maybe you have a better idea how to simplify this.

Imho a new settings impl should support external type --> internal type conversions (my stub does not yet).

jerch:
@Tyriar The const enum indirection is quite ugly and we dont need the speed benefit, thus I removed it.

src/Terminal.ts
```diff
@@ -1432,7 +1430,9 @@ export class Terminal extends Disposable implements ITerminal, IDisposable, IInp
 
     // Sync the scroll area to make sure scroll events don't fire and scroll the viewport to an
     // invalid location
-    this.viewport.syncScrollArea(true);
+    if (this.viewport) {
+      this.viewport.syncScrollArea(true);
+    }
```
Tyriar: `this.viewport?.syncScrollArea(true)`

src/Terminal.ts
```diff
@@ -128,8 +128,6 @@ export class Terminal extends Disposable implements ITerminal, IDisposable, IInp
   public sendFocus: boolean;
 
   // misc
```
Tyriar: Let's remove this comment too

src/common/Types.d.ts
```diff
@@ -254,3 +254,31 @@ export interface ICoreMouseProtocol {
  * (DEFAULT only supports up to 223 1-based as coord value).
  */
 export type CoreMouseEncoding = (event: ICoreMouseEvent) => string;
+
+/**
+ * WindowOption settings.
```
Tyriar: `windowOptions`

test/api/InputHandler.api.ts
```diff
@@ -403,3 +441,17 @@ async function getCursor(): Promise<{col: number, row: number}> {
   })();
   `);
 }
+
+async function getDimensions(): Promise<any> {
+  const dim = await page.evaluate(`
+    (function() {
+      return term._core._renderService.dimensions;
+    })();
+  `);
```
Tyriar: You should be able to just do this:

```ts
const dim = await page.evaluate(`term._core._renderService.dimensions`);
```

typings/xterm.d.ts
```diff
@@ -374,6 +380,133 @@ declare module 'xterm' {
...
   }
 
+  /**
+   * Enable various window manipulation and report features (CSI Ps ; Ps ; Ps t).
+   * 
+   * Note that most settings have no default implementation, as they heavily
+   * rely on the embedding environment.
+   * 
+   * To implement a features, create a custom CSI hook like this:
+   * ```Typescript
```
Tyriar: \`\`\`ts

typings/xterm.d.ts
```diff
@@ -374,6 +380,133 @@ declare module 'xterm' {
     tooMuchOutput: string;
   }
 
+  /**
+   * Enable various window manipulation and report features (CSI Ps ; Ps ; Ps t).
+   * 
+   * Note that most settings have no default implementation, as they heavily
+   * rely on the embedding environment.
```
Tyriar: They probably shouldn't be represented in the setting if they don't have an implementation? The below example is showing how to override the t handler to intercept the setting change?

src/InputHandler.ts
```diff
@@ -2056,14 +2210,22 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+   *   Proxy to set window title.
    */
   public setTitle(data: string): void {
+    this._windowTitle = data;
     this._terminal.handleTitle(data);
   }
 
+  /**
+   * OSC 1; <data> ST
+   * Note: Icon name is not exposed.
```
Tyriar: Is this meant to set the icon of the terminal tab? Should we expose it? Is it base64 encoded, a path?

src/InputHandler.ts
```diff
@@ -2022,6 +2084,98 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+        }
+        break;
+      case 20:  // GetIconTitle, returns OSC L label ST
+        this._coreService.triggerDataEvent(`${C0.ESC}]L${this._iconName}${C0.ESC}\\`);
+        break;
+      case 21:  // GetWinTitle, returns OSC l label ST
+        this._coreService.triggerDataEvent(`${C0.ESC}]l${this._windowTitle}${C0.ESC}\\`);
+        break;
+      case 22:  // PushTitle
+        if (!second || second === 2) {
```
Tyriar: Probably more readable for these to explicitly use `second === 0`?

src/InputHandler.ts
```diff
@@ -2022,6 +2084,98 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+      case 18:  // GetWinSizeChars, returns CSI 8 ; height ; width t
+        if (this._bufferService) {
+          this._coreService.triggerDataEvent(`${C0.ESC}[8;${this._bufferService.rows};${this._bufferService.cols}t`);
+        }
+        break;
+      case 20:  // GetIconTitle, returns OSC L label ST
+        this._coreService.triggerDataEvent(`${C0.ESC}]L${this._iconName}${C0.ESC}\\`);
+        break;
+      case 21:  // GetWinTitle, returns OSC l label ST
+        this._coreService.triggerDataEvent(`${C0.ESC}]l${this._windowTitle}${C0.ESC}\\`);
```
Tyriar: Danger 😱 

The API needs a big scary warning that it is very dangerous, we probably shouldn't even have this implemented at all unless we can come up with some way to sanitize the string reliably even if it is very aggressive sanitization (and let bug reports fix the edge cases we missed).

src/InputHandler.ts
```diff
@@ -2022,6 +2084,98 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+   *    Ps = 23 ; 1  -> Restore xterm icon title from stack.              supported
+   *    Ps = 23 ; 2  -> Restore xterm window title from stack.            supported
+   *    Ps >= 24                                                          not implemented
+   */
+  public windowOptions(params: IParams): void {
+    if (!paramToWindowOption(params.params[0], this._optionsService.options.windowOptions)) {
+      return;
+    }
+    const second = (params.length > 1) ? params.params[1] : 0;
+    const rs = (this._terminal as any)._renderService;  // FIXME: import renderService to get rid of any type here
```
Tyriar: I guess we need an `attachToDom` function to set this for the time being? Can think more about the right fix when Terminal.ts is moved over.

src/InputHandler.ts
```diff
@@ -1395,11 +1448,16 @@ export class InputHandler extends Disposable implements IInputHandler {
           this._terminal.setgCharset(3, DEFAULT_CHARSET);
           // set VT100 mode here
           break;
-        case 3: // 132 col mode
-          // TODO: move DECCOLM into compat addon
-          this._terminal.savedCols = this._bufferService.cols;
```
Tyriar: savedCols was wrong?

src/Terminal.ts
jerch: Ah good we have that now :)

typings/xterm.d.ts
jerch: > They probably shouldn't be represented in the setting if they don't have an implementation?

Well they have to be in the settings, otherwise I cannot provide the security guard for them. The guard is always on, thus will filter all `CSI t` calls (no matter if custom handler or default one).

> The below example is showing how to override the t handler to intercept the setting change?

Yepp. Would it work in the docs page later on? I can remove it here and put up a page to describe the window options further (incl. example snippet).

src/InputHandler.ts
jerch: This is really ancient stuff from those early X window managers, that had those big icons on the desktop for minimized windows (like CDE, fvwm and such). For xterm.js I cannot think of any use case here beside someone wants to fake a desktop with the exact same behavior (or hyper running on fvwm2).

I had to implement this to get a hold of the icon name for the stack shifts. Its basically the same as the window title, but not exposed (as not being useful). Still some multiplexer like tmux and screen use the stack stuff to save window and icon name before running and to restore them on exit.

The content is plain text similar to window title, which raises security concerns for the stack thingy (basically writing it back to the terminal).

src/InputHandler.ts
jerch: Yes thats fishy and basically allows the same attack vector as we had with our first (and hopefully only) CVE. Difference here is the security settings - its always disabled, if a users enables it, he does it on his own risk. All window options have that restriction, they are kinda all subject of CVEs if enabled by default.

Background on this:
As far as I know all common linux distros disable those options for xterm by default. Some of the other emulators implement only a subset, though I have not checked which one understands which option. I think some emulators have the supported options always on, which is a matter of time and the size of user base to earn a CVE.

We have everything disabled by default here, so no harm in the first place. The more I think about it - guess we need a guide about windows options in the docs that covers:
- security implications with a big warning (e.g. always go disabled by default)
- default implementations by xterm.js
- guide to implement custom other options

src/InputHandler.ts
jerch: Well kinda. I checked against xterm - it always applies 80 vs. 132, and does not fall back to some custom size that might have been there before.

Again, this setting depends on `CSI 24 t` which is disabled by default. So no screwed up embedding page anymore, if delivered with defaults. Those terminal window resizing commands need alot of extra impl on integrator side to correctly deal with it (prolly worth to be mentioned/explained in the guide).

typings/xterm.d.ts
Tyriar: > Well they have to be in the settings, otherwise I cannot provide the security guard for them. The guard is always on, thus will filter all CSI t calls (no matter if custom handler or default one).

Hmm, I don't understand. We don't implement `restoreWin` so why would we have a `restoreWin?: boolean;`  setting that does nothing?

src/InputHandler.ts
Tyriar: Well enabling it will enable an arbitrary code execution exploit so we shouldn't include it at all if that's the case. Can't we just sanitize _windowTitle and _iconName here and strip almost everything out though, like only include alphanumeric and `-`, `_` or something?

src/Terminal.ts
jerch: Also removed `params` and `currentParam` 2 lines below, that were not used anywhere.

test/api/InputHandler.api.ts
jerch: Yes works.

typings/xterm.d.ts
jerch: Its handled as security guard for an option - no matter if or what kind of implementation there is for a particular options (a default one or 10 custom overloads from integrators) the guard is always in place. I made it this way to ensure integrators will not introduce security issues by implementing it but not checking whether the option is enabled. Furthermore for this to work the option has to be there in the first place (otherwise integrators would have to implement their own option in OptionService, which has no APi for that).

Bottom line of the defensive handling of the security setting:  Many of the windowOptions allow grabbing infos from the host machine where the terminal runs on, thus the final decision whether an option is enabled/disabled should be on users side.

src/InputHandler.ts
jerch: Ok removed these two. They are pretty simple to implement, so lets forward the security concerns to integrators who are keen to deal with them.

jerch:
Ready for the next review :smile:

src/InputHandler.ts
```diff
@@ -2037,6 +2096,92 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+   *    Ps = 23 ; 1  -> Restore xterm icon title from stack.              supported
+   *    Ps = 23 ; 2  -> Restore xterm window title from stack.            supported
+   *    Ps >= 24                                                          not implemented
+   */
+  public windowOptions(params: IParams): void {
+    if (!paramToWindowOption(params.params[0], this._optionsService.options.windowOptions)) {
+      return;
+    }
+    const second = (params.length > 1) ? params.params[1] : 0;
+    const rs = (this._terminal as any)._renderService;  // FIXME: import renderService to get rid of any type here
```
Tyriar: Let's just expose a `InputHandler.setRenderService` to work around this for now.

typings/xterm.d.ts
```diff
@@ -379,6 +385,162 @@ declare module 'xterm' {
...
+   *   return false;      // any Ps that was not handled
+   * });
+   * ```
+   * 
+   * Note on security:  
+   * Most features are meant to deal with some information of the host machine
+   * where the terminal runs on. This is seen as a security risk possibly leaking
+   * sensitive data of the host to the program in the terminal. Therefore all options
+   * (even those without a default implementation) are guarded by the boolean flag
+   * and disabled by default.
```
Tyriar: Don't custom handlers get executed before the default implementation so this setting doesn't actually guard anything?

jerch: Yes, thats why I have to do the security check one level higher during registering (kinda currying all CSI t handlers, that might show up): https://github.com/xtermjs/xterm.js/blob/479fddfeec049d1527accd4f86acc26951427e3f/src/InputHandler.ts#L576

Tyriar: Oh ok, I missed that part.

src/common/services/OptionsService.ts
```diff
@@ -50,6 +50,7 @@ export const DEFAULT_OPTIONS: ITerminalOptions = Object.freeze({
   convertEol: false,
   termName: 'xterm',
   cancelEvents: false,
+  windowOptions: {},
```
Tyriar: windowOptions should move above (with wordSeparator), the bottom part is for old/private options

src/common/services/Services.ts
```diff
@@ -279,6 +280,7 @@ export interface ITerminalOptions {
   cancelEvents: boolean;
   convertEol: boolean;
   termName: string;
+  windowOptions: IWindowOptions;
```
Tyriar: windowOptions should move above, the bottom part is for old/private options

src/InputHandler.ts
Tyriar: Fixed by adding a `IInstantiationService.getService` which is better than the `setRenderService` approach (which has caused bugs in the past).

PR #2448, opened 2019-10-07
Don't handle cmd+arrow

### LINKED ISSUE #597: Should cmd+left/right be printing escape sequences?
I don't think xterm considers the meta key, cmd+left and cmd+right however send 9D and 9C. Regardless we should align with or allow consumers to align with how standard terminal emulators work in this regard on macOS (Terminal.app and iTerm2).

@parisk interested in your input as I don't normally work on a Mac.

Code: https://github.com/sourcelair/xterm.js/blob/ff29fd69a316c59f55dbec1cc86f00c262b73275/src/xterm.js#L1430
VS Code discussion: https://github.com/Microsoft/vscode/pull/22005#issuecomment-286524481

Tyriar:
Fixes #597

### DIFF
--- a/src/common/input/Keyboard.ts
+++ b/src/common/input/Keyboard.ts
@@ -113,6 +113,9 @@ export function evaluateKeyboardEvent(
       break;
     case 37:
       // left-arrow
+      if (ev.metaKey) {
+        break;
+      }
       if (modifiers) {
         result.key = C0.ESC + '[1;' + (modifiers + 1) + 'D';
         // HACK: Make Alt + left-arrow behave like Ctrl + left-arrow: move one word backwards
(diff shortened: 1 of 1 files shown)


koreno:
I'm not clear on this - it appears my MacOS users are expecting this combo to trigger the "Home"/"End" keys. Is there some so solution for them? or is it a bug?

mig8447:
In macOS the Fn + ArrowLeft or Cmd + ArrowLeft make the same function as the Home and the same shortcuts with ArrowRight act as the End key. I'm not able to use this in the current demo

Tyriar:
@mig8447 it's a non-standard hotkey, if you want it in your terminal you need to set it up yourself via `addCustomKeyEventHandler`

mig8447:
@Tyriar , I attempted to do so, and tried several ansi escapes, but most take me to the beginning of the line, without accounting for the prompt. I even tries to look on the VSCode and Hyper repos as they can handle those shortcuts but I got lost in an ocean of code and didn't find anything useful on the matter.  Would you have any advice or pointers that could help me implement this?

Tyriar:
Here's what we do in vscode for cmd 🙂 

https://github.com/microsoft/vscode/blob/041313bea584743ae251e79ff0cc2bce34ec1c7e/src/vs/workbench/contrib/terminal/browser/terminal.contribution.ts#L256-L267

PR #2455, opened 2019-10-08
Update unfocused cursor to match current cursor style

### LINKED ISSUE #2356: When not focused the cursor should match the style of the terminal
This needs to get fixed individually for [canvas](https://github.com/xtermjs/xterm.js/blob/master/src/renderer/CursorRenderLayer.ts), [dom](https://github.com/xtermjs/xterm.js/blob/master/src/renderer/dom/DomRenderer.ts) and [webgl](https://github.com/xtermjs/xterm.js/blob/master/addons/xterm-addon-webgl/src/renderLayer/CursorRenderLayer.ts) renderers so they are consistent:

JeffreyCA:
Fixes #2356.

### Block cursor (unchanged)
When terminal is unfocused, it shows a outline of the block cursor.

### Underline and bar cursors
When terminal is unfocused, it shows the underline or bar cursor. It always showed the outlined block cursor prior to this change.

### Additional Change
This change revealed an unintended behaviour where if blinking is enabled and terminal is not focused, changing the cursor style did not update screen's cursor. A second change was made to force a refresh after options have changed regardless of whether blinking is enabled.

### Tests
I tested unfocusing the terminal with all three styles and all three renderers (DOM, Canvas, WebGL) and verified proper cursor behaviour.

### DIFF
--- a/addons/xterm-addon-webgl/src/renderLayer/CursorRenderLayer.ts
+++ b/addons/xterm-addon-webgl/src/renderLayer/CursorRenderLayer.ts
@@ -139,8 +139,8 @@ export class CursorRenderLayer extends BaseRenderLayer {
       this._clearCursor();
       this._ctx.save();
       this._ctx.fillStyle = this._colors.cursor.css;
-      const cursorStyle = terminal.getOption('cursorStyle')
-      if (cursorStyle && cursorStyle != 'block') {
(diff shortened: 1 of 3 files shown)


lolimay:
Thanks for your hardworking.

PR #2467, opened 2019-10-11
Don't use ctrl+up/down hack on macOS

### LINKED ISSUE #2387: Option+up/down sends wrong thing on macOS
This is used by fish shell to navigate completions.

VS Code issue: https://github.com/microsoft/vscode/issues/79237

Tyriar:
Fixes #2387


lgarron:
Cheers, thanks so much for the fix! :-D

Tyriar:
@lgarron I'll update vscode so the insiders build will get it tomorrow, would be great if you could verify the fix works right as I'm not a fish user.

lgarron:
> @lgarron I'll update vscode so the insiders build will get it tomorrow, would be great if you could verify the fix works right as I'm not a fish user.

I'm a bit late here, but this works perfectly now, thanks! :-D

(There's another bug with oversized text sometimes, but I think it's not related, and it's only visual.)

PR #2470, opened 2019-10-14
Support link position in hover tooltip callback

### LINKED ISSUE #2468: Support link matcher start position
The registerLinkMatcher() should support returning the row and column position of the link so it can be used for position the hover message. Related to https://github.com/microsoft/vscode/issues/77964. 

For VSCode, I believe this is needed in order to match the terminal link experience (the popup starts where the link starts) to the rest of the editor. Currently, there is no way to get the position of the link, and the hover position is just the mouse position, which is inconsistent.

jmbockhorst:
Added support for link position in the hover tooltip callback and tested that it works in the demo. 
I defined `ILinkLocation` in both the public API and the browser types. I know that this probably isn't correct, but I wasn't sure how to get around it. Closes #2468.

### DIFF
--- a/src/browser/Linkifier.ts
+++ b/src/browser/Linkifier.ts
@@ -306,7 +306,7 @@ export class Linkifier implements ILinkifier {
       e => {
         this._onLinkTooltip.fire(this._createLinkHoverEvent(x1, y1, x2, y2, fg));
         if (matcher.hoverTooltipCallback) {
-          matcher.hoverTooltipCallback(e, uri);
+          matcher.hoverTooltipCallback(e, uri, { startRow: y1, startColumn: x1, endRow: y2, endColumn: x2 });
         }
       },
       () => {
--- a/src/browser/Linkifier.ts
+++ b/src/browser/Linkifier.ts
@@ -306,7 +306,7 @@ export class Linkifier implements ILinkifier {
       e => {
         this._onLinkTooltip.fire(this._createLinkHoverEvent(x1, y1, x2, y2, fg));
         if (matcher.hoverTooltipCallback) {
-          matcher.hoverTooltipCallback(e, uri, { startRow: y1, startColumn: x1, endRow: y2, endColumn: x2 });
+          matcher.hoverTooltipCallback(e, uri, { start: { row: y1, col: x1 }, end: { row: y2, col: x2 } });
         }
       },
       () => {
--- a/src/browser/Linkifier.ts
+++ b/src/browser/Linkifier.ts
@@ -306,7 +306,9 @@ export class Linkifier implements ILinkifier {
       e => {
         this._onLinkTooltip.fire(this._createLinkHoverEvent(x1, y1, x2, y2, fg));
         if (matcher.hoverTooltipCallback) {
-          matcher.hoverTooltipCallback(e, uri, { start: { row: y1, col: x1 }, end: { row: y2, col: x2 } });
+          // Note that IViewportRange use 1-based coordinates to align with escape sequences such
+          // as CUP which use 1,1 as the default for row/col
+          matcher.hoverTooltipCallback(e, uri, { start: { row: y1 + 1, col: x1 + 1 }, end: { row: y2 + 1, col: x2 } });
         }
       },
       () => {
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -862,12 +862,12 @@ declare module 'xterm' {
    */
   interface IViewportCellPosition {
     /**
-     * The column of the cell.
+     * The column of the cell. Note that this is 1-based; the first column is column 1.
      */
     col: number;
 
     /**
-     * The row of the cell.
+     * The row of the cell. Note that this is 1-based; the first row is row 1.
      */
     row: number;
   }
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -269,7 +269,7 @@ declare module 'xterm' {
     /**
      * A callback that fires when the mouse hovers over a link for a moment.
      */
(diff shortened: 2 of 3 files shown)


typings/xterm.d.ts
```diff
@@ -842,6 +842,31 @@ declare module 'xterm' {
...
+    /**
+     * The end column of the link.
+     */
+    endColumn: number;
+
+    /**
+     * The end row of the link.
+     */
+    endRow: number;
+  }
```
Tyriar: @jerch any thoughts on the naming here? It's basically identical to `ISelectionPosition` but `ILinkPosition` is within the viewport and `ISelectionPosition` is within the whole buffer.

Should we normalize these to `IBufferRange` and `IViewportRange`? row vs line vs y? 🤔

jerch: I am not quite sure about our naming convention here, I think we use "column"/"cols" and "row" for stuff in viewport? And maybe use "line" everywhere something has whole buffer access as distinction? `IBufferRange` and `IViewportRange` seems to be easier to comprehend the access range/limits.

More general thoughts on this - we maintain already several coord systems with different metrics: text grid level with viewport / whole buffer access with different meanings of the index numbers and with different point of origin (API is 0 based, sequences 1 based), pixel level stuff in browser parts and so on. Maybe its time for 2d vector types to distinguish those better, but I am unsure about this, as it again introduces custom types that will increase runtime for construction/gc objects where in fact only two numbers are needed. Still custom vector types would help to clarify the domain and even could be made translateable (e.g. a `ViewportCoord` can always translate into a `BufferCoord` but not the way around).

Tyriar: > I am not quite sure about our naming convention here, I think we use "column"/"cols" and "row" for stuff in viewport (cursor being the only exception with x/y)? And maybe use "line" everywhere something has whole buffer access as distinction?

This has always bothered me how many names there are for similar things but no clear distinction.

> Maybe its time for 2d vector types to distinguish those better, but I am unsure about this, as it again introduces custom types that will increase runtime for construction/gc objects where in fact only two numbers are needed. Still custom vector types would help to clarify the domain and even could be made translateable (e.g. a ViewportCoord can always translate into a BufferCoord but not the way around).

We shouldn't worry about perf too much with these types of APIs are they're rarely called (both this and `getSelectionPosition`). How about this?

```ts
interface IViewportRange {
  start: IViewportCell;
  end: IViewportCell;
}

// We would transition getSelectionPosition to this at v5?
interface IBufferRange {
  start: IBufferCell;
  end: IBufferCell;
}
```

> Btw, would be good to state anywhere in the docs, whether the range end is meant inclusive or exclusive.

:+1:, they should say inclusive.

mofux: @Tyriar How would the interfaces for `IViewportCell` and `IBufferCell` look like? Shouldn't they much rather be an `IViewportCellPosition` and `IBufferCellPosition`? Or maybe a unified `IPosition` with `x` and `y` properties?

Tyriar: @mofux good point 👍, `IBufferCell` is already a thing

```ts
interface IViewportRange {
  start: IViewportCellPosition;
  end: IViewportCellPosition;
}

// We would transition getSelectionPosition to this at v5?
interface IBufferRange {
  start: IBufferCellPosition;
  end: IBufferCellPosition;
}
```

Tyriar: I'm fine with col/row for viewport (to match how escape sequences are named) and x/y for viewport, we could surface this in the API too so it's clear that the interfaces are different things:

```ts
interface IViewportRange {
  start: IViewportCellPosition;
  end: IViewportCellPosition;
}
interface IViewportCellPosition {
  col: number;
  row: number;
}

// We would transition getSelectionPosition to this at v5?
interface IBufferRange {
  start: IBufferCellPosition;
  end: IBufferCellPosition;
}
interface IBufferPosition {
  x: number;
  y: number;
}
```

typings/xterm.d.ts
```diff
@@ -843,28 +843,33 @@ declare module 'xterm' {
...
     /**
-     * The start row of the link.
+     * The end cell of the range.
      */
-    startRow: number;
+    end: IViewportCellPosition;
+  }
 
+  /**
+   * An object representing a cell within the viewport of the terminal.
```
mofux: I think this should say:
An object representing a cell **position** within the viewport of the terminal.

src/browser/Linkifier.ts
```diff
@@ -306,7 +306,9 @@ export class Linkifier implements ILinkifier {
       e => {
         this._onLinkTooltip.fire(this._createLinkHoverEvent(x1, y1, x2, y2, fg));
         if (matcher.hoverTooltipCallback) {
-          matcher.hoverTooltipCallback(e, uri);
+          // Note that IViewportRange use 1-based coordinates to align with escape sequences such
+          // as CUP which use 1,1 as the default for row/col
+          matcher.hoverTooltipCallback(e, uri, { start: { row: y1 + 1, col: x1 + 1 }, end: { row: y2 + 1, col: x2 } });
```
Tyriar: I made a change to expose this as a 1-based number to align with https://invisible-island.net/xterm/ctlseqs/ctlseqs.html. It's looks a little weird how the end col doesn't add 1 but I guess that's how the mouse zones work right now, they include everything up to but excluding the end cell

mofux: Did you forget `col: x2` here? Shouldn't it be `col: x2 + 1` as like the others?

Tyriar:
@mofux the right numbers are coming through like that, I think mouse zone uses coords a little strangely (I wrote it 😅) and the end isn't included in the range

Tyriar:
image
image
---
image
image
---
image
image

Issue #2463, opened 2019-10-10
Change default word separator to include comma


Tyriar:
Some links would be negatively affected by this but in general I think it would make a lot more sense to exclude make comma a word separator for double clicking.

VS Code issue: https://github.com/microsoft/vscode/issues/82173

glneto:
I had put up #2474 to address this. I added just comma for now but I saw there was some discussion regarding also adding colon, so let me know if it also needs to be added as a default separator.

Tyriar:
Fixed in https://github.com/xtermjs/xterm.js/pull/2474

PR #2474, opened 2019-10-16
#2463 Add comma as a default word separator

glneto:
This is a simple change just to make sure when we double click a word in xterm the selection stops when it finds a comma.

### DIFF
--- a/src/common/services/OptionsService.ts
+++ b/src/common/services/OptionsService.ts
@@ -49,7 +49,7 @@ export const DEFAULT_OPTIONS: ITerminalOptions = Object.freeze({
   screenKeys: false,
   cancelEvents: false,
   useFlowControl: false,
-  wordSeparator: ' ()[]{}\'"'
+  wordSeparator: ' ()[]{}\',"'
 });
 
 /**


glneto:
> @glneto yes I think we should do `;` and `:` as well.

Ok I just added `;` and `;` to the PR as well.

cu:
> @glneto yes I think we should do `;` and `:` as well.

Can I ask why this was done? What is the rationale or benefit? Can it be undone?

I used xtermjs by way of vscode and one action that I perform multiple times every day is double-clicking to select URLs in the terminal for copy+paste into specific form fields. This change breaks that rather badly.

Tyriar:
Yes it can be undone, I undid the default in vscode very recently for similar reasons https://github.com/xtermjs/xterm.js/issues/2665

jerch:
@Tyriar Is this worth to be exposed as terminal setting in vscode, like we have in our demo?

Tyriar:
@jerch maybe, was hoping vscode wouldn't need yet another word separators list though.

jerch:
@Tyriar Yeah understandable, also the more low level and ancient a setting gets, the more likely to trigger follow-up issues.

PR #2476, opened 2019-10-16
Fix scrollback going missing when increase size on Windows

### LINKED ISSUE #2459: Scrollback goes missing when resizing a conpty-based terminal
VS Code issue: https://github.com/microsoft/vscode/issues/78634

I think it's related to how we try to pull scrollback into the viewport when resizing upwards, but then conpty still draws its current state of the work. Maybe we should disable that in windows mode?

Tyriar:
This also forces the scroll bar to sync after a resize early, this may fix
other bugs where the renderer and scroll bar get out of sync.

Fixes #2459

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -1469,6 +1469,10 @@ export class Terminal extends Disposable implements ITerminal, IDisposable, IInp
       this._charSizeService.measure();
     }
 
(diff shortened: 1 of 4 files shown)


Tyriar:
Still need to test this without Windows mode

PR #2483, opened 2019-10-18
Introduce FitAddon tests

### LINKED ISSUE #2167: Create puppeteer tests for FitAddon
Follow up from #2164

leomoty:
Fixes #2167

I have a few questions:

- Can I rely on the default #terminal-container size?
- Is changing width good enough? should cover height as well?
- I noticed that if I try something stupid, like having a 1x1 pixel #terminal-container the fit will propose a negative cols. Is this expected behavior?

I want to remove the duplicated code to a helper function before merging.

### DIFF
--- a/addons/xterm-addon-fit/src/FitAddon.api.ts
+++ b/addons/xterm-addon-fit/src/FitAddon.api.ts
@@ -14,7 +14,7 @@ let page: puppeteer.Page;
 const width = 1024;
 const height = 768;
 
-describe('FitAddon', () => {
+describe.only('FitAddon', () => {
   before(async function(): Promise<any> {
     this.timeout(20000);
     browser = await puppeteer.launch({
@@ -24,25 +24,25 @@ describe('FitAddon', () => {
     });
     page = (await browser.pages())[0];
     await page.setViewport({ width, height });
+    await page.goto(APP);
+    await openTerminal();
   });
 
   after(async () => {
     await browser.close();
   });
 
-  beforeEach(async function(): Promise<any> {
-    this.timeout(20000);
-    await page.goto(APP);
+  it('no terminal', async function(): Promise<any> {
+    await page.evaluate(`window.fit = new FitAddon();`);
+    assert.equal(await page.evaluate(`window.fit.proposeDimensions()`), undefined);
   });
 
   describe('proposeDimensions', () => {
-    it('no terminal', async function(): Promise<any> {
-      await page.evaluate(`window.fit = new FitAddon();`);
-      assert.equal(await page.evaluate(`window.fit.proposeDimensions()`), undefined);
+    afterEach(async () => {
+      return unloadFit();
     });
 
     it('default', async function(): Promise<any> {
-      await openTerminal();
       await loadFit();
       assert.deepEqual(await page.evaluate(`window.fit.proposeDimensions()`), {
         cols: 87,
@@ -51,7 +51,6 @@ describe('FitAddon', () => {
     });
 
     it('width', async function(): Promise<any> {
-      await openTerminal();
       await loadFit(1008);
       assert.deepEqual(await page.evaluate(`window.fit.proposeDimensions()`), {
         cols: 110,
@@ -60,7 +59,6 @@ describe('FitAddon', () => {
     });
 
     it('small', async function(): Promise<any> {
-      await openTerminal();
       await loadFit(1, 1);
       assert.deepEqual(await page.evaluate(`window.fit.proposeDimensions()`), {
         cols: 2,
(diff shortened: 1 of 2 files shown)


addons/xterm-addon-fit/src/FitAddon.api.ts
```diff
@@ -0,0 +1,100 @@
...
+      rows: 26
+    });
+  });
+
+  it('width propose fit', async function(): Promise<any> {
+    await openTerminal();
+    await page.evaluate(`
+      window.fit = new FitAddon();
+      window.term.loadAddon(window.fit);
+      document.querySelector('#terminal-container').style.width='100%';
```
Tyriar: This might be better as an absolute `px` value in case `#terminal-container`'s container changes its size?

addons/xterm-addon-fit/src/FitAddon.api.ts
```diff
@@ -0,0 +1,100 @@
...
+    this.timeout(20000);
+    await page.goto(APP);
+  });
+
+  it('no terminal', async function(): Promise<any> {
+    await page.evaluate(`window.fit = new FitAddon();`);
+    assert.equal(await page.evaluate(`window.fit.proposeDimensions()`), undefined);
+  });
+
+  it('default propose fit', async function(): Promise<any> {
```
Tyriar: Can you move all tests relating to just proposeDimensions into `describe('proposeDimensions', () => {`, and all fit ones into the same but `'fit'`?

addons/xterm-addon-fit/src/FitAddon.api.ts
```diff
@@ -0,0 +1,100 @@
...
+    assert.equal(await page.evaluate(`window.term.cols`), 87);
+    assert.equal(await page.evaluate(`window.term.rows`), 26);
+  });
+
+  it('width fit', async function(): Promise<any> {
+    await openTerminal();
+    await page.evaluate(`
+      window.fit = new FitAddon();
+      window.term.loadAddon(window.fit);
+      document.querySelector('#terminal-container').style.width='100%';
```
Tyriar: Again it might be safer to use `px`

addons/xterm-addon-fit/src/FitAddon.api.ts
leomoty: Sounds good

addons/xterm-addon-fit/src/FitAddon.api.ts
leomoty: Leave the `no terminal` test outside?

Tyriar: The no terminal test just uses proposeDimensions so it can live in there

leomoty:
How about this instead @Tyriar? it removes a bit of duplicated code, but we don't rely directly on the default width? I can revert otherwise.

I still need to add one test for the 1x1 case, will do after work.

leomoty:
I don't understand why the MacOS integration tests failed, I kept the same timeout that exists elsewhere?

Also, regarding the FitAddon, it doesn't need to be smaller than 1x1 to break:

```javascript
    it('small width', async function(): Promise<any> {
      await openTerminal();
      await loadFit(1);
      console.log(await page.evaluate(`window.fit.proposeDimensions()`));
    });
```
returns:

```javascript
{ cols: -2, rows: 26 }
```

What is the desired result? 

```javascript
    it('small width', async function(): Promise<any> {
      await openTerminal();
      await loadFit(1);
      await page.evaluate(`window.fit.fit()`);
      console.log(await page.evaluate(`window.term.cols`));
      console.log(await page.evaluate(`window.term.rows`));
    });
```
returns:
```
2
26
```

leomoty:
Checking the `BufferService.ts`, I would expect FitAddon to match MINIMUM_COLS and MINIMUM_ROWS defined there. Since that is what Terminal seems to do nonetheless.

Tyriar:
> I don't understand why the MacOS integration tests failed, I kept the same timeout that exists elsewhere?

Not sure, try narrow down which await is not returning? I'll requeue the build in case it's flakiness

leomoty:
I have updated the `loadFit` to allow both width/height (defaults are based from demo's style.css).

Also updated FitAddon to respect MINIMUM_COLS and MINIMUM_ROWS.

Mac OS is still flaky, but since I only have linux to test, I am not sure why.

Tyriar:
Looks good, I'll look into the mac test now

Tyriar:
I couldn't repro the flakiness on my mac, I pushed a change to avoid loading the page every time though which makes the tests take about 25% of the time. Hopefully the flakiness is fixed magically 🤞

PR #2485, opened 2019-10-18
Adds scrollSensitivity option for scrolling speed

### LINKED ISSUE #1805: Terminal is mouse wheel is scrolling too much
Original issue: https://github.com/Microsoft/vscode/issues/63669

Using these mouse settings on Windows 10:

image

The terminal seems to scroll 7 lines instead of 3.

Code pointer:

https://github.com/xtermjs/xterm.js/blob/290c95ad71f8b1264139a7ed6f070c4741a8987e/src/Viewport.ts#L160-L190

eamodio:
Adds a new `scrollSensitivity` terminal option to allow control over the normal scrolling speed.
Closes #1805

### DIFF
--- a/src/browser/Viewport.ts
+++ b/src/browser/Viewport.ts
@@ -198,7 +198,7 @@ export class Viewport extends Disposable implements IViewport {
     }
 
     // Fallback to WheelEvent.DOM_DELTA_PIXEL
-    let amount = this._applyFastScrollModifier(ev.deltaY, ev);
+    let amount = this._applyScrollModifier(ev.deltaY, ev);
     if (ev.deltaMode === WheelEvent.DOM_DELTA_LINE) {
       amount *= this._currentRowHeight;
     } else if (ev.deltaMode === WheelEvent.DOM_DELTA_PAGE) {
@@ -219,7 +219,7 @@ export class Viewport extends Disposable implements IViewport {
     }
 
     // Fallback to WheelEvent.DOM_DELTA_LINE
-    let amount = this._applyFastScrollModifier(ev.deltaY, ev);
+    let amount = this._applyScrollModifier(ev.deltaY, ev);
     if (ev.deltaMode === WheelEvent.DOM_DELTA_PIXEL) {
       amount /= this._currentRowHeight + 0.0; // Prevent integer division
       this._wheelPartialScroll += amount;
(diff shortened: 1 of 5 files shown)


src/common/services/OptionsService.ts
```diff
@@ -33,6 +33,7 @@ export const DEFAULT_OPTIONS: ITerminalOptions = Object.freeze({
   letterSpacing: 0,
   logLevel: 'info',
   scrollback: 1000,
+  scrollSensitivity: 0,
```
Tyriar: Let's make this 1 by default

src/browser/Viewport.ts
```diff
@@ -231,15 +231,18 @@ export class Viewport extends Disposable implements IViewport {
...
     const modifier = this._optionsService.options.fastScrollModifier;
     // Multiply the scroll speed when the modifier is down
     if ((modifier === 'alt' && ev.altKey) ||
         (modifier === 'ctrl' && ev.ctrlKey) ||
         (modifier === 'shift' && ev.shiftKey)) {
       return amount * Math.max(1, this._optionsService.options.fastScrollSensitivity);
     }
-    return amount;
+
+    return this._optionsService.options.scrollSensitivity > 0
```
Tyriar: We should be able to simplify this to just `return amount * this._optionsService.options.scrollSensitivity;`

To handle invalid values via `setOption` add this as a special case in `OptionsService._sanitizeAndValidateOption` (we should be doing the same for `fastScrollSensitivity` too instead of using the `Math.max` above).

src/common/services/OptionsService.ts
eamodio: Doh -- that was supposed to be 1 😄

src/browser/Viewport.ts
eamodio: OK, will do

src/common/services/OptionsService.ts
```diff
@@ -121,6 +122,16 @@ export class OptionsService implements IOptionsService {
           throw new Error(`${key} cannot be less than 0, value: ${value}`);
         }
         break;
+      case 'fastScrollSensitivity':
+          if (value < 1) {
```
Tyriar: Let's fall through to the `scrollSensitivity` case, a 0 < fastScrollSensitivity < 1 could make sense on some setups after our research.

PR #2491, opened 2019-10-21
Prevent search result from being deselected if it is the only result

### LINKED ISSUE #2456: Search should not deselect search term if it's the only result
- Run ls
- Search for OSS in the terminal
- Press enter and see one result for OSSREADME.json correctly highlighted
- Now search multiple times

VS Code: https://github.com/microsoft/vscode/issues/62348

miggs125:
fixes #2456 

Prevents single results from being deselected after pressing enter multiple times on a successful search. I couldn't find tests for addons so I wasn't sure if some needed to be made. I will be happy to include them if they are needed.

### DIFF
--- a/addons/xterm-addon-search/src/SearchAddon.ts
+++ b/addons/xterm-addon-search/src/SearchAddon.ts
@@ -59,12 +59,12 @@ export class SearchAddon implements ITerminalAddon {
 
     let startCol = 0;
     let startRow = 0;
-
+    let currentSelection = null;
     if (this._terminal.hasSelection()) {
       const incremental = searchOptions ? searchOptions.incremental : false;
       // Start from the selection end if there is a selection
       // For incremental search, use existing row
-      const currentSelection = this._terminal.getSelectionPosition()!;
+      currentSelection = this._terminal.getSelectionPosition()!;
       startRow = incremental ? currentSelection.startRow : currentSelection.endRow;
       startCol = incremental ? currentSelection.startColumn : currentSelection.endColumn;
     }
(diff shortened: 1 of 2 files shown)


Tyriar:
@miggs125 sorry I just overwrote your commit. I think the test I added should be good now, what was that last commit you added on?

miggs125:
@Tyriar I saw that some tests had not passed after you added the test suite so I commited the fix just recently. The last commit on my branch should be 0e6fde54

miggs125:
@Tyriar for some reason the tests you added are passing on my local copy of the repo but not for the CI tests, not sure what is going on here.

image

miggs125:
I modified the search tests to output the result of the functions being tested just to confirm nothing unexpected was going on and got this: 

image

Based on this, the fix appears to be working as expected so I am unsure as to why the CI tests are failing.

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -97,6 +97,9 @@ export class SearchAddon implements ITerminalAddon {
       }
     }
 
+    // If there is only one result, return false.
+    if (!result && currentSelection) return false;
```
Tyriar: The test fails for me, it's because it returns false if there's no selection. I think what we actually want to do is return true instead, this make sense since we succeeded but just didn't touch the selection as it would be redundant to do so.

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -59,12 +59,12 @@ export class SearchAddon implements ITerminalAddon {
 
     let startCol = 0;
     let startRow = 0;
-
+    let currentSelection = null;
```
Tyriar: Let's make this `let currentSelect: ISelectionPosition | undefined`, using undefined saves us from initializing it.

PR #2493, opened 2019-10-22
Forbid only in mocha

leomoty:
I saw #2489 and was thinking, how about forbidding `.only` in Mocha for CI only? Should be way easier to catch if someone mistakenly commits it.

```
 mocha "**/*.api.js" --headless --forbid-only
C:\Users\leomo\Documents\projects\xterm.js\node_modules\yargs\yargs.js:1163
      else throw err
           ^

Error: `.only` forbidden
```

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,7 +20,7 @@ jobs:
       yarn
     displayName: 'Install dependencies and build'
   - script: |
-      yarn test-unit
+      yarn test-unit --forbid-only
     displayName: 'Unit tests'
   - script: |
       yarn lint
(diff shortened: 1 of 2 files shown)


leomoty:
How do you wanna handle that? It seems that `argv.length > 2` is considered as test files, is this feature used anywhere?

Tyriar:
Yeah @jerch added that, how about using the same flag name and doing the test file check by filtering out flag args: `nonFlagArgs = argv.filter(e => e.startsWith('--'))`?

leomoty:
I hope I understood you correctly, I think you meant `flagArgs` there. This should work, by the way.

PR #2512, opened 2019-10-25
Refine IViewportRange API to use 'cursor positions' for x value

### LINKED ISSUE #2484: Close on decision regarding 0-based and 1-based coordinates in API
https://github.com/xtermjs/xterm.js/pull/2470 added the 1-based interface `IViewportCellPosition` (uses col/row, aligns with [how escape sequences work](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html)), https://github.com/xtermjs/xterm.js/issues/2480 will add `IBufferCellPosition` (x/y) which may be 0-based? Is that confusing? What should we do?

@mofux tells me monaco uses 1-based and language servers use 0-based. Maybe we should just use 0-based everywhere (should we avoid col/row in the API)?

Tyriar:
Fixes #2484

### DIFF
--- a/src/browser/Linkifier.ts
+++ b/src/browser/Linkifier.ts
@@ -308,7 +308,7 @@ export class Linkifier implements ILinkifier {
         if (matcher.hoverTooltipCallback) {
           // Note that IViewportRange use 1-based coordinates to align with escape sequences such
           // as CUP which use 1,1 as the default for row/col
-          matcher.hoverTooltipCallback(e, uri, { start: { row: y1 + 1, col: x1 + 1 }, end: { row: y2 + 1, col: x2 } });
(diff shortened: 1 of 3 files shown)


src/browser/Linkifier.ts
```diff
@@ -308,7 +308,7 @@ export class Linkifier implements ILinkifier {
         if (matcher.hoverTooltipCallback) {
           // Note that IViewportRange use 1-based coordinates to align with escape sequences such
           // as CUP which use 1,1 as the default for row/col
-          matcher.hoverTooltipCallback(e, uri, { start: { row: y1 + 1, col: x1 + 1 }, end: { row: y2 + 1, col: x2 } });
+          matcher.hoverTooltipCallback(e, uri, { start: { x: x1, y: y1 }, end: { x: x2, y: y2 } });
```
jerch: Thats a win compared to the left side :+1:.

PR #2516, opened 2019-10-25
conditional build for addons

jerch:
Coming from https://github.com/xtermjs/xterm.js/pull/2503#discussion_r338691866.

This uses a custom addon-install script, that is only present for repo forks (not contained in the npm package). The conditional execution is simply done at package.json level as a `try { require(...) } ...` call.

@Tyriar:
This works for me, the conditional stuff is merely a hack but should work on all platforms. Did not like the idea to provide different shell scripts here.


package.json
```diff
@@ -8,10 +8,11 @@
   "repository": "https://github.com/xtermjs/xterm.js",
   "license": "MIT",
   "scripts": {
+    "postinstall": "node -e \"try { require('./bin/install-addons'); } catch(e) {}\"",
```
Tyriar: You could stick the catch inside the script, it probably isn't needed though?

bin/install-addons.js
```diff
@@ -0,0 +1,49 @@
+const path = require('path');
```
Tyriar: Copyright

bin/install-addons.js
```diff
@@ -0,0 +1,49 @@
...
+    files.forEach(folder => {
+      const addonPath = path.join(addonsPath, folder);
+
+      // install only if there are dependencies listed
+      // also skip addon if it does not contain any package.json
+      // (might happen after branch switches)
+      let packageJson;
+      try {
+        packageJson = require(path.join(addonPath, 'package.json'));
+      } catch (e) {}
```
Tyriar: Is the try catch needed?

package.json
jerch: Well this wont work for normal npm package, since the script cannot be required at all - thus `npm install` errors out.

bin/install-addons.js
jerch: Well thats there to not annoy to much during dev cycles - if you switch from an (not yet merged) addon branch to another branch, the folder is likely to stay in place, but w'o package.json (git will remove any source related stuff due to being version controlled).

bin/install-addons.js
jerch: Fixed.

package.json
Tyriar: `prepare` is what we're after https://stackoverflow.com/a/49629544/1156119

Tyriar: Try `postprepare` as we build there currently

bin/install-addons.js
```diff
@@ -0,0 +1,56 @@
...
```
Tyriar:
```suggestion
      } catch (e) {
        // swallow as changing branches can leave folders around
      }
```

leomoty:
Just for reference, an alternative would be lerna:

https://github.com/lerna/lerna/tree/master/commands/bootstrap

Not sure if it is worth the trouble.

package.json
jerch: See new commit - with `prebuild` it works as intended.

jerch:
@leomoty Thx for the suggestion. I think we should keep it as simple as possible, to me lerna seems to be overkill here.

leomoty:
@jerch, yeah agreed.

package.json
```diff
@@ -11,11 +11,12 @@
     "prepackage": "npm run build",
     "package": "webpack",
     "start": "node demo/start",
-    "lint": "tslint 'src/**/*.ts' 'addons/**/*.ts'",
+    "lint": "tslint 'src/**/*.ts' 'addons/*/src/**/*.ts'",
     "test": "npm run test-unit",
     "posttest": "npm run lint",
     "test-api": "mocha \"**/*.api.js\"",
     "test-unit": "node ./bin/test.js",
+    "prebuild": "node ./bin/install-addons.js",
```
Tyriar: I think we should keep this strictly as a prepare step (is postprepare a thing?), I'd like to keep `build` strictly as an equivalent to `watch` but just a single compile.

jerch: pre/postprepare does not work - it gets never called. How about adding a prewatch? This way build/watch behave the same regarding the addons.

jerch: Fixed with the `setup` indirection. Now it gets only called during `yarn` or `npm install`for full repo folder.

PR #2517, opened 2019-10-26
Support disposing of EventEmitters

### LINKED ISSUE #2509: Support disposing of EventEmitters
VS Code disposes Events which frees references to all its listeners, we should do the same:

https://github.com/microsoft/vscode/blob/4d1e08fec954853878c692a4c6629d7f0f884fd0/src/vs/base/common/event.ts#L591-L602

revanthmahesh:
dispose function to clear all the listener references at once.

Fixes #2509

### DIFF
--- a/src/common/EventEmitter.ts
+++ b/src/common/EventEmitter.ts
@@ -16,11 +16,13 @@ export interface IEvent<T> {
 export interface IEventEmitter<T> {
   event: IEvent<T>;
   fire(data: T): void;
+  dispose(): void;
 }
 
 export class EventEmitter<T> implements IEventEmitter<T> {
   private _listeners: IListener<T>[] = [];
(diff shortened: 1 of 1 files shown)


src/common/EventEmitter.ts
```diff
@@ -7,6 +7,7 @@ import { IDisposable } from 'common/Types';
 
 interface IListener<T> {
   (e: T): void;
+  (): void;
```
Tyriar: Why was this added?

Tyriar: This is probably the cause of the build failure

revanthmahesh: IListener is allowing only a function with at least one param. It is not allowing Array.prototype.clear. In vs code it is  handled by type LinkedList.

PR #2520, opened 2019-10-27
remove conditional preventing search in wrapped line

### LINKED ISSUE #2325: Search addon does not search start of wrapped line when ydisp is in the middle of it
Repro:

Write one really long wrapped line that spans multiple viewports
Search for a string above the viewport, no results will be found

miggs125:
fixes #2325.  I removed a conditional that appears to prevent searching in wrapped lines.


addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -228,11 +228,6 @@ export class SearchAddon implements ITerminalAddon {
   protected _findInLine(term: string, row: number, col: number, searchOptions: ISearchOptions = {}, isReverseSearch: boolean = false): ISearchResult | undefined {
     const terminal = this._terminal!;
 
-    // Ignore wrapped lines, only consider on unwrapped line (first row of command string).
```
Tyriar: Hmm, I'm not sure we can remove this. The idea with this early exit was to only allow searching from the first line since we need to find terms that could span multiple lines. The call to `_translateBufferLineToStringWithWrap` just below this is meant to fetch the entire string for the _wrapped_ line that could include multiple individual terminal rows.

Tyriar:
Closing as it's stale.

PR #2526, opened 2019-10-31
Progress toward #1686: some desirable selection behaviour

Silvyre:
Some progress was made towards #1686:
- Error-prone calls to `String.length` were replaced with calls to `wcwidth`.
   - (Note: I made a copy of `common/CharWidth.ts` within `addons/xterm-addon-search/`, as helper functions in `src/common`, e.g. `wcwidth`, cannot currently be imported into addons; I will open an issue to address this.)
- Selections of full-width and diacritical characters are better now, but still require more work.
   - Selection of "full words" of diacritical and fullwidth characters now appear to fully function.
      - e.g. searching for `ééé` or `￥￥￥` after inserting `echo -en 'combining: ééé\nfullwidth: ￥￥￥\n'` into the demo now works well.
   - Selection of "partial words" of such characters is improved, but requires further tweaking. See:  https://github.com/xtermjs/xterm.js/issues/1686#issuecomment-547811304

### DIFF
--- a/addons/xterm-addon-search/src/CharWidth.ts
+++ b/addons/xterm-addon-search/src/CharWidth.ts
@@ -1,0 +1,171 @@
+/**
+ * Copyright (c) 2016 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { fill } from './TypedArrayUtils';
+
+export const wcwidth = (function(opts: {nul: number, control: number}): (ucs: number) => number {
+  // extracted from https://www.cl.cam.ac.uk/%7Emgk25/ucs/wcwidth.c
+  // combining characters
+  const COMBINING_BMP = [
+    [0x0300, 0x036F], [0x0483, 0x0486], [0x0488, 0x0489],
+    [0x0591, 0x05BD], [0x05BF, 0x05BF], [0x05C1, 0x05C2],
+    [0x05C4, 0x05C5], [0x05C7, 0x05C7], [0x0600, 0x0603],
+    [0x0610, 0x0615], [0x064B, 0x065E], [0x0670, 0x0670],
+    [0x06D6, 0x06E4], [0x06E7, 0x06E8], [0x06EA, 0x06ED],
+    [0x070F, 0x070F], [0x0711, 0x0711], [0x0730, 0x074A],
+    [0x07A6, 0x07B0], [0x07EB, 0x07F3], [0x0901, 0x0902],
+    [0x093C, 0x093C], [0x0941, 0x0948], [0x094D, 0x094D],
+    [0x0951, 0x0954], [0x0962, 0x0963], [0x0981, 0x0981],
+    [0x09BC, 0x09BC], [0x09C1, 0x09C4], [0x09CD, 0x09CD],
+    [0x09E2, 0x09E3], [0x0A01, 0x0A02], [0x0A3C, 0x0A3C],
+    [0x0A41, 0x0A42], [0x0A47, 0x0A48], [0x0A4B, 0x0A4D],
+    [0x0A70, 0x0A71], [0x0A81, 0x0A82], [0x0ABC, 0x0ABC],
+    [0x0AC1, 0x0AC5], [0x0AC7, 0x0AC8], [0x0ACD, 0x0ACD],
+    [0x0AE2, 0x0AE3], [0x0B01, 0x0B01], [0x0B3C, 0x0B3C],
+    [0x0B3F, 0x0B3F], [0x0B41, 0x0B43], [0x0B4D, 0x0B4D],
+    [0x0B56, 0x0B56], [0x0B82, 0x0B82], [0x0BC0, 0x0BC0],
+    [0x0BCD, 0x0BCD], [0x0C3E, 0x0C40], [0x0C46, 0x0C48],
+    [0x0C4A, 0x0C4D], [0x0C55, 0x0C56], [0x0CBC, 0x0CBC],
+    [0x0CBF, 0x0CBF], [0x0CC6, 0x0CC6], [0x0CCC, 0x0CCD],
+    [0x0CE2, 0x0CE3], [0x0D41, 0x0D43], [0x0D4D, 0x0D4D],
+    [0x0DCA, 0x0DCA], [0x0DD2, 0x0DD4], [0x0DD6, 0x0DD6],
+    [0x0E31, 0x0E31], [0x0E34, 0x0E3A], [0x0E47, 0x0E4E],
+    [0x0EB1, 0x0EB1], [0x0EB4, 0x0EB9], [0x0EBB, 0x0EBC],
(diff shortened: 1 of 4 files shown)


addons/xterm-addon-search/src/CharWidth.ts
```diff
@@ -0,0 +1,171 @@
+/**
+ * Copyright (c) 2016 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { fill } from './TypedArrayUtils';
+
+export const wcwidth = (function(opts: {nul: number, control: number}): (ucs: number) => number {
```
Tyriar: If we're going to need internal code in an addon we can actually reference it directly using the same trick as the webgl addon:

https://github.com/xtermjs/xterm.js/blob/16409da5178731dbc7a309385bcb89efba4ad8ed/addons/xterm-addon-webgl/src/tsconfig.json#L14-L27

https://github.com/xtermjs/xterm.js/blob/16409da5178731dbc7a309385bcb89efba4ad8ed/addons/xterm-addon-webgl/src/WebglAddon.ts#L8-L9

Care needs to be used here as when webpack does its thing it will pull in all code from the file (and linked from that file), but CharWidth is a pretty standalone thing so it should be good.

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -211,7 +212,7 @@ export class SearchAddon implements ITerminalAddon {
    */
   private _isWholeWord(searchIndex: number, line: string, term: string): boolean {
     return (((searchIndex === 0) || (NON_WORD_CHARACTERS.indexOf(line[searchIndex - 1]) !== -1)) &&
-      (((searchIndex + term.length) === line.length) || (NON_WORD_CHARACTERS.indexOf(line[searchIndex + term.length]) !== -1)));
+      (((searchIndex + getStringCellWidth(term)) === line.length) || (NON_WORD_CHARACTERS.indexOf(line[searchIndex + getStringCellWidth(term)]) !== -1)));
```
jerch: I think `term.length` was right here since it evals data in the `line` string (thus the string index should be used). The wcwidth correction is only needed for buffer offset/index calculation.

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -251,21 +252,21 @@ export class SearchAddon implements ITerminalAddon {
       if (isReverseSearch) {
         // This loop will get the resultIndex of the _last_ regex match in the range 0..col
         while (foundTerm = searchRegex.exec(searchStringLine.slice(0, col))) {
-          resultIndex = searchRegex.lastIndex - foundTerm[0].length;
+          resultIndex = searchRegex.lastIndex - getStringCellWidth(foundTerm[0]);
           term = foundTerm[0];
-          searchRegex.lastIndex -= (term.length - 1);
+          searchRegex.lastIndex -= (getStringCellWidth(term) - 1);
```
jerch: Same problem here - the regex operates on string data, thus go with the string index/length.

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -251,21 +252,21 @@ export class SearchAddon implements ITerminalAddon {
...
         }
       } else {
         foundTerm = searchRegex.exec(searchStringLine.slice(col));
-        if (foundTerm && foundTerm[0].length > 0) {
-          resultIndex = col + (searchRegex.lastIndex - foundTerm[0].length);
+        if (foundTerm && getStringCellWidth(foundTerm[0]) > 0) {
+          resultIndex = col + (searchRegex.lastIndex - getStringCellWidth(foundTerm[0]));
```
jerch: Imho this does not work at all this way - it mixes cell offsets (`col`) with string offsets into `resultIndex`. Note that cell offsets must contain the wcwidth correction, while string offsets should never.

Possible fix:
- do the regexp index handling/matching beforehand
- if a match was found, take the match index and do a `getStringCellWidth(stringLine.slice(0, matchIndex))` - this gives the correct cell offset of the start character
- add `getStringCellWidth(term)` to that - cell offset of the end character + 1
- do a divmod correction for both values to translate them into actual col/row values (`col = startOffset % terminal.cols`, `row += startOffset / terminal.cols`)

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -251,21 +252,21 @@ export class SearchAddon implements ITerminalAddon {
...
           term = foundTerm[0];
         }
       }
     } else {
       if (isReverseSearch) {
-        if (col - searchTerm.length >= 0) {
-          resultIndex = searchStringLine.lastIndexOf(searchTerm, col - searchTerm.length);
+        if (col - getStringCellWidth(searchTerm) >= 0) {
```
jerch: Same here but in opposite direction.

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -292,7 +293,7 @@ export class SearchAddon implements ITerminalAddon {
           }
           // Adjust the searchIndex to normalize emoji into single chars
           const char = cell.char;
-          if (char.length > 1) {
+          if (char.length == 2 && getStringCellWidth(char) == 2) {
             resultIndex -= char.length - 1;
```
jerch: This code does not make any sense to me. It definitely does not (only) do what the comment says. Again `resultIndex` is involved in this, I think the whole block from line 288 on needs a rewrite introduced by the needed rewrites above.

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -349,7 +350,7 @@ export class SearchAddon implements ITerminalAddon {
       terminal.clearSelection();
       return false;
     }
-    terminal.select(result.col, result.row, result.term.length);
+    terminal.select(result.col, result.row, getStringCellWidth(result.term));
```
jerch: Yepp here `getStringCellWidth` is needed. Still `result.col` and `result.row` may contain wrong value due to the cell and string offset mixing above.
Also note, that you might have to do the divmod correction on the end position here (if the term happens to wrap at the rightmost column).

Silvyre:
Thank you so much for the review, @Tyriar @jerch!

I plan to find some time to address your comments in the following week!

jerch:
@Silvyre Any progress on your side? Plz drop me a note if you some help.

Silvyre:
> 
> 
> @Silvyre Any progress on your side? Plz drop me a note if you need some help.

Really appreciately the offer for help, and will definitely will take you up on it. I am a bit apprehensive of making further changes, as I am still not very confident in my understanding of cell and string offsetting.

Tyriar:
Thanks for the PR but closing off since it's a bit stale 🙂

jerch:
@Silvyre Thx for dealing with this issue. Sorry from my side, it went somewhat under my radar.

Silvyre:
@jerch No problem, thanks for all of your help!

Issue #2115, opened 2019-05-21
Clickable multi-line links have regressed


Tyriar:
Upstream issue: https://github.com/microsoft/vscode/issues/61708
Old PR that solved the issue (things have changed a lot in the area though): https://github.com/xtermjs/xterm.js/pull/1769

We should verify the state of the bug before tackling this.

kamalesh0406:
I would like to look into this

Tyriar:
@kamalesh0406 sounds good 👍

kamalesh0406:
<img width="997" alt="Screen Shot 2019-06-24 at 3 40 55 PM" src="https://user-images.githubusercontent.com/29024165/60011023-8f015200-9696-11e9-8374-0922d5d1a352.png">

I think the bug has been resolved.

jerch:
@kamalesh0406 Multiline support in general works yeah, but there are issues with special conditions of chars:
- combining chars/wide chars in the line prior the URL (or any content to be linkified) might offset the underline (both start and end)
- combing chars/wide chars within the linkified content might offset end of underline

lramos15:
![](http://g.recordit.co/bdd1uO4vlD.gif)
Repro steps:
1. Fill terminal
2. clear
3. Execute a program which prints a link longer than the current size of the terminal.

Tyriar:
Seems to be fixed, probably with https://github.com/xtermjs/xterm.js/pull/2530

Issue #2458, opened 2019-10-09
Links don't work when terminal is hidden and shown


Tyriar:
I think this has been happening in VS Code since it was added.

VS Code issue: https://github.com/microsoft/vscode/issues/73129

PR #2530, opened 2019-11-01
Basic link parser for better link detection

jmbockhorst:
Here is my first attempt at a basic link parser for #583. This is a very rough draft and I realize that it may be in the wrong direction, but I wanted feedback on how this compares to what you guys had in mind. 

This adds a new `Linkifier2` class with that handles registering a link parser. I also updated the Web Links Addon to use the new parser system. I mainly used the `LinkDetector` from VSCode and modified it as needed for this application. It seems to work well and it fixes most of the cases that were broken with the old system.

I didn't do anything with markers because I'm not sure that how that fits in, since a marker only holds the line number, and we need to store the link range. 

Still things to do:
- Support for multiline links
- Handle when the buffer changes above a cached link

EDIT: Sorry, I meant to make this a draft pull request


addons/xterm-addon-web-links/src/WebLinkProvider.ts
```diff
@@ -0,0 +1,329 @@
...
+
+    const CANNOT_END_WITH_CHARACTERS = '.,;';
+    for (let i = 0; i < CANNOT_END_WITH_CHARACTERS.length; i++) {
+      classifier.set(CANNOT_END_WITH_CHARACTERS.charCodeAt(i), CharacterClass.CANNOT_END_IN);
+    }
+  }
+  return classifier;
+}
+
+export class LinkComputer {
```
Tyriar: Instead of pulling in the VS Code code here I think we should adapt the previous web links regex-based approach to work on hover, so we could use all the same regex's but just evaluate it on hover instead.

addons/xterm-addon-web-links/src/WebLinksAddon.ts
```diff
@@ -42,7 +44,9 @@ export class WebLinksAddon implements ITerminalAddon {
 
   public activate(terminal: Terminal): void {
     this._terminal = terminal;
-    this._linkMatcherId = this._terminal.registerLinkMatcher(strictUrlRegex, this._handler, this._options);
+    this._linkProvider = new WebLinkProvider(this._terminal, this._handler);
+    // this._linkMatcherId = this._terminal.registerLinkMatcher(strictUrlRegex, this._handler, this._options);
+    this._terminal.registerLinkProvider(this._linkProvider);
```
Tyriar: Maybe we could do this?

```ts
if ('registerLinkProvider' in this._terminal) {
  // Use new approach
} else {
  // Use old approach
}
```

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,150 @@
...
+ */
+
+
+/**
+ */
+export class Linkifier2 implements ILinkifier2 {
+  private _element: HTMLElement | undefined;
+  private _linkProviders: ILinkProvider[] = [];
+  private _mouseService: IMouseService | undefined;
+  private _linkCache: ILinkCache[] = [];
```
Tyriar: I think we need to invalidate the entire cache when a scroll event happens as line indexes change when scrolling if scrollback is full (ie. `CircularList.onTrim`).

Additionally we need to invalidate on:

- A line changes, example 'a' is printed
- Lines are removed (`CircularList.onDelete`)
- Lines are added (`CircularList.onInsert`)

We could look into optimizing this later if we wanted but for now we can just take the safe route or alternatively only cache the single link that is under the cursor for the MVP?

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,150 @@
...
+    this._linkCache.forEach((cachedLink, i) => {
+      if (this._linkAtPosition(cachedLink.link, position)) {
+        cachedLink.link.handle(event, cachedLink.link.url);
+      }
+    });
+  }
+
+  private _handleNewLink(link: ILink | undefined): void {
+    if (link && !this._linkCache.find(cachedLink => cachedLink.link = link)) {
+      this._linkCache.push({ link: link, mouseOver: false });
```
Tyriar: Shorthand: `{ link, mouseOver: false }`

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  /**
+   * Check if the buffer position is within the link
+   * @param link
+   * @param position
+   */
+  private _linkAtPosition(link: ILink, position: IBufferCellPosition): boolean {
+    return link.range.start.x <= position.x
+      && link.range.start.y <= position.y
```
Tyriar: We normally stick the && on the end of the line

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,150 @@
...
+      && link.range.end.x >= position.x
+      && link.range.end.y >= position.y;
+  }
+
+  /**
+   * Get the buffer position from a mouse event
+   * @param event
+   */
+  private _positionFromMouseEvent(event: MouseEvent): IBufferCellPosition | undefined {
+    if (!this._element) {
```
Tyriar: Include a check for `_mouseService` here to avoid using `!`

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,150 @@
...
+  public attachToDom(element: HTMLElement, mouseService: IMouseService): void {
+    this._element = element;
+    this._mouseService = mouseService;
+
+    this._element.addEventListener('mousemove', this._onMouseMove.bind(this));
+    this._element.addEventListener('click', this._onMouseDown.bind(this));
+  }
+
+  private _onMouseMove(event: MouseEvent): void {
+    const position = this._positionFromMouseEvent(event);
```
Tyriar: We could avoid the `!` on `this._element` below by checking both `_element` and `_mouseService` in this function and passing them into `this._positionFromMouseEvent`. Then changing `this._linkCache` to a regular `for` loop. `!` is bad generally as it tells TS not to check something, it definitely looks safe now but future refactors could cause issues that TS would catch.

src/browser/Linkifier2.ts
jmbockhorst: How can we determine when a line changes? I might be misunderstanding how `CircularList` works, but the `onDelete` or `onInsert` events are not called when I try to add or delete lines.

Tyriar: It looks like `onInsert`/`onDelete` only fire right now when a reflow occurs (after resizing the col count):

https://github.com/xtermjs/xterm.js/blob/91d1f97754edef5c7f9fcdd6d3c3ea4c92e65f5b/src/common/buffer/Buffer.ts#L455

Created https://github.com/xtermjs/xterm.js/issues/2533 to follow that up.

---

> How can we determine when a line changes?

I'm beginning to think this should hook into the rendering dirty row marking code here:

https://github.com/xtermjs/xterm.js/blob/bc9649bfbcb4c7a3ef8b53dceae7bffa96e7f4e8/src/common/services/DirtyRowService.ts#L14-L15

Also noticed what might be another issue in https://github.com/xtermjs/xterm.js/issues/2534, the move to DirtyRowService might have only been half done.

Tyriar: The latter part was bad advice, that's the hope of `DirtyRowService` eventually but you can't do it yet (see https://github.com/xtermjs/xterm.js/issues/2535). I'd recommend avoiding caching outside of the current range for now.

jmbockhorst: Alright, sounds good. Also, the tests are failing because `page.hover()` doesn't seem to be hovering over the links. I'm not sure why this is happening, since it uses the same mouse move events as the current `MouseZoneManager`.

jmbockhorst:
@Tyriar I've implemented a new caching system that only uses the cache for the lines before the current one. (Is there any issues doing it like this) When there is any data input, if the mouse is currently over a link, it hides the tooltip and deletes it from the cache. This also happens on scroll. This seems to work well from my testing. 

Still to do:
- Handle when the scrollback is full (Invalidate the entire cache)
- Multi line link detection in the `WebLinksAddon`

src/Terminal.ts
```diff
@@ -1166,6 +1170,13 @@ export class Terminal extends Disposable implements ITerminal, IDisposable, IInp
     }
   }
 
+  public registerLinkProvider(linkProvider: ILinkProvider): IDisposable {
+    if (!this.linkifier2) {
+      return;
+    }
```
Tyriar: This check shouldn't be needed.

Tyriar:
> Handle when the scrollback is full (Invalidate the entire cache. Using onTrim?)

There's some discussion on caching strategy in https://github.com/xtermjs/xterm.js/issues/583#issuecomment-548741361, I still recommend for the MVP just going with invalidating the cache when the cursor leaves the link's range and improving upon it after this PR is merged.

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,250 @@
...
+
+/**
+ * Copyright (c) 2017 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+
+/**
+ */
+export class Linkifier2 implements ILinkifier2 {
```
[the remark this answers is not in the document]
jmbockhorst: I'm having trouble with getting the type definitions in the right places. `LinkService` in common needs an `ILinkProvider`, but that can't be defined in common Types.d.ts because it needs an `ILink` which needs a `MouseEvent` (From my understanding, it can't be used in common). I don't know how or where to define it since common can't rely on browser Types (Where it is currently defined).

jmbockhorst:
This PR might fix #2115 and #2458, but it'll need to be tested thoroughly.

Tyriar:
Yeah we should be getting https://github.com/xtermjs/xterm.js/issues/2458 for free, multi-line maybe not as I think it was to do with line wrapped status breaking after a resize.

addons/xterm-addon-web-links/src/WebLinkProvider.ts
```diff
@@ -0,0 +1,114 @@
...
+  constructor(
+    private readonly _terminal: Terminal,
+    private readonly _regex: RegExp,
+    private readonly _handler: (event: MouseEvent, uri: string) => void
+  ) {
+
+  }
+
+  provideLink(position: IBufferCellPosition, callback: (link: ILink | undefined) => void): void {
+    callback(LinkComputer.computeLink(position, this._regex, this._terminal, this._handler));
```
Tyriar: Currently any `mousemove` will trigger `provideLink`, it should only fire as a minimum when the mouse is moved off of the current cell:

Screen Shot 2019-11-07 at 11 23 43 AM

Ideally we would cache just the current link and only fire provideLink when the cursor moved off of the current link's range. We could clear out the current link by listening to `onData` and `onScroll`:

```ts
// Register listeners and dispose them immediately after either of them are fired
this._linkCacheDisposables.push(this._coreService.onData(() => this._clearCurrentLink()));
this._linkCacheDisposables.push(this._onScroll(() => this._clearCurrentLink()));

...

private _clearCurrentLink() {
  this._currentLink = undefined;
  this._linkCacheDisposables.forEach(l => l.dispose());
  this._linkCacheDisposables = [];
}));
```

`onScroll` is currently owned by `src/Terminal.ts` but it will eventually move into `common/` and be available on a service, in the meantime I suggest just passing it into `Linkifier2` via the constructor much like we're doing for this:

https://github.com/xtermjs/xterm.js/blob/16409da5178731dbc7a309385bcb89efba4ad8ed/src/common/services/CoreService.ts#L26-L27

src/browser/Linkifier2.ts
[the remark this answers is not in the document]
Tyriar: Hmm good call, maybe we should think more about this when it comes time to remove `src/Terminal.ts`. One idea is that we could just register the callback with `any` instead of `MouseEvent` only in `common/` and cast it if needed when accessing, I wonder if there's a better way though...

addons/xterm-addon-web-links/src/WebLinksAddon.ts
```diff
@@ -42,12 +44,21 @@ export class WebLinksAddon implements ITerminalAddon {
 
   public activate(terminal: Terminal): void {
     this._terminal = terminal;
-    this._linkMatcherId = this._terminal.registerLinkMatcher(strictUrlRegex, this._handler, this._options);
+
+    if ('registerLinkProvider' in this._terminal as any) {
+      this._linkProvider = this._terminal.registerLinkProvider(new WebLinkProvider(this._terminal, strictUrlRegex, this._handler));
+    } else {
+      this._linkMatcherId = this._terminal.registerLinkMatcher(strictUrlRegex, this._handler, this._options);
+    }
```
Tyriar: This seems a little safer:

```ts
    if ('registerLinkProvider' in this._terminal) {
      this._linkProvider = this._terminal.registerLinkProvider(new WebLinkProvider(this._terminal, strictUrlRegex, this._handler));
    } else {
      // HACK: This is an older version of xterm.js, use registerLinkMatcher
      this._linkMatcherId = (<Terminal>this._terminal).registerLinkMatcher(strictUrlRegex, this._handler, this._options);
    }
```

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,246 @@
...
+  ) {
+
+  }
+
+  public registerLinkProvider(linkProvider: ILinkProvider): IDisposable {
+    this._linkProviders.push(linkProvider);
+    return {
+      dispose: () => {
+        // Remove the link provider from the list
+        this._linkProviders.splice(this._linkProviders.indexOf(linkProvider), 1);
```
Tyriar: We should probably verify indexOf results in >= 0 so the dispose call is idempotent.

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,246 @@
...
+  public attachToDom(element: HTMLElement, viewportElement: HTMLElement, mouseService: IMouseService): void {
+    this._element = element;
+    this._viewportElement = viewportElement;
+    this._mouseService = mouseService;
+
+    this._element.addEventListener('mousemove', this._onMouseMove.bind(this));
+    this._element.addEventListener('click', this._onMouseDown.bind(this));
+    this._viewportElement.addEventListener('scroll', this._onScroll.bind(this));
+
+    this._coreService.onData(this._onData.bind(this));
```
Tyriar: I talked about onData and onScroll in another comment, we should take care about disposing these when we don't want them as onData and onScroll can fire _a lot_.

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    if (!mouseOver) {
+      this._mouseOverLink = false;
+      this._linkCache = [];
+    }
+
+    // The is no link in the cache, so ask for one
+    this._linkProviders.forEach(linkProvider => {
+      linkProvider.provideLink(position, this._handleNewLink.bind(this));
+    });
```
Tyriar: I think that providers are registered is important here, it looks like currently it's non-deterministic if 2 link providers that call `callback` async are used? How about we use a strategy something like asking all link providers to provide links and wait for them to all callback before actually showing the link?

Improving upon this a little, consider what VS Code will probably do which is register a web link provider first (top priority), then a FS link provider second. We could fire off both link providers and if the web link provider comes back first action that immediately as it's the highest priority, if the FS link provider comes back first then wait for the web link provider to return no results first before calling that. This would mean that we get deterministic results based on the order that link providers are registered and also that the web link provider would not need to wait on the FS call before creating the link (the FS call can be expensive for example when using Remote-SSH in vscode).

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,246 @@
...
+    this._linkCache = [];
+  }
+
+  private _onData(e: string): void {
+    if (this._lastMouseEvent && this._mouseOverLink) {
+      this._hideAllTooltips();
+    }
+
+    this._linkCache = [];
+  }
```
Tyriar: These can be merged together

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    return;
+  }
+
+  /**
+   * Check if the buffer position is within the link
+   * @param link
+   * @param position
+   */
+  private _linkAtPosition(link: ILink, position: IBufferCellPosition): boolean {
```
Tyriar: I see an issue where the cell immediately following a link will count as part of the link's range as well, something related to end.x will probably need to change in here to fix that.

addons/xterm-addon-web-links/src/WebLinkProvider.ts
```diff
@@ -0,0 +1,66 @@
+import { ILinkProvider, IBufferCellPosition, ILink, Terminal, IBuffer } from 'xterm';
```
Tyriar: Needs the license header

addons/xterm-addon-web-links/src/WebLinksAddon.ts
```diff
@@ -42,12 +44,21 @@ export class WebLinksAddon implements ITerminalAddon {
...
   }
 
   public dispose(): void {
     if (this._linkMatcherId !== undefined && this._terminal !== undefined) {
       this._terminal.deregisterLinkMatcher(this._linkMatcherId);
     }
+
+    if (this._linkProvider) {
+      this._linkProvider.dispose();
+    }
```
Tyriar: New syntax for this as of a couple of days ago (TS 3.7):

```ts
this._linkProvider?.dispose();
```

addons/xterm-addon-web-links/src/WebLinkProvider.ts
jmbockhorst: Looks like scroll events don't get fired when the viewport is scrolled, only when a new line is added. Ref: https://github.com/xtermjs/xterm.js/blob/16409da5178731dbc7a309385bcb89efba4ad8ed/src/browser/Viewport.ts#L159
Changing the second parameter to false gives what we what, but I'm sure it breaks something elsewhere.

jerch:
I dont want to disrupt the review process with @Tyriar thus I dont go into coding details. Just a few general remarks from my side:
- `Terminal.onScroll` This is mainly used for ringbuffer rotations (terminal buffer). Imho in earlier releases viewport scrolls were also coupled to it to some degree if it happened together with line inserts/removals. We might need here a better separation of concerns and better wording.
- about linkprovider and caching later on: Currently you call providers with `IBufferCellPosition` as argument. Imho thats an unlucky choice, since the position can and will mutate by many buffer actions. This makes caching almost undoable or invalidates entries often. Furthermore all indiviual providers will have to pull the buffer content with `translateToString` since their workdata is not the position but the line's content.

  Suggestion: Do the `translateToString` on `Linkifier2` level once for the current line under the mouse (with wrapping in mind) and call the individual providers with this string data. Now they can directly do their link pattern parsing and return string offsets. The string offsets can be turned on `Linkifier2` level back to buffer positions.
Caching can be achieved quite simple with this - use the line content as cache keys and save the provider results as entries. This has the big advantage that any buffer content change will not match anymore and trigger a fresh eval. The cache itself can be made as a short FIFO ringbuffer (maybe 10 entries only).

- debouncing the mouse events to grid metric changes: I had the same problem in the `CoreMouseService` and went with a simple row/col check. This has a disadvantage - if a user steals the focus, the event will no re-trigger when the focus comes backand the mouse didnt move out of the cell. @Tyriar Maybe we should move mouse-to-grid normalization into some "MouseService" later on to avoid doing the same over and over (well there is still my early stub to get some unification for mouse stuff ;) ).

jmbockhorst:
I made all of the requested changes and switched everything to using a single cached link. Everything should work the same as before. (Except for scrolling because of the onScroll issue)

Tyriar:
Fixed conflicts caused by https://github.com/xtermjs/xterm.js/pull/2547

addons/xterm-addon-web-links/src/WebLinkProvider.ts
Tyriar: Changing that to false will break a bunch of other stuff... Hmm I thought there was an event for when scroll happens but I guess not. I think what we're actually after is this one:

https://github.com/xtermjs/xterm.js/blob/b68a4d27ab6cea738cd7b4a69fd5f0e819fa1847/src/browser/services/Services.ts#L46

This fires immediately after lines in the range `start` to `end` have been re-rendered. We could cache the current mouse position from the `mousemove` event and then provide links for it in the `onRender` listener?

Tyriar:
> Terminal.onScroll This is mainly used for ringbuffer rotations (terminal buffer).

This is right, we should be using `IRenderService.onRender` instead of `onData`/`onScroll` as I suggested.

> about linkprovider and caching later on

Again we should avoid caching outside of just the current line for the first iteration of this which I think is what's happening now. The reason I push for this so much as because cache invalidation is complicated and the sooner this gets in (after we've sorted out the fundamentals), the easier it is to iterate on.

This does bring up the point that links with unicode probably won't work at the moment. I guess the question is how much information should be exposed to `provideLink`, resolving buffer vs string indexes gets awfully complicated and one of the elegant things about how we expose it right now is that it's up to the addon to fetch the content of each cell in which case they already know what the buffer range is. Personally I'd say we can go ahead as it is and see if my idea of having the link provider addon fetch the buffer cells manually works out. We can always change it later since it's experimental (and should do if it makes sense).

> debouncing the mouse events to grid changes

Some debouncing would definitely be useful, consider when moving the mouse quickly from one side of the terminal to the other for example. This can easily get deferred to a future PR though imo.

jmbockhorst:
I changed it to use onRender instead of onData and onScroll and it works well. 
Not sure what you meant by this:
> We could cache the current mouse position from the mousemove event and then provide links for it in the onRender listener?

Tyriar:
@jmbockhorst it should be fine how you're doing it, nvm 😃

src/browser/Linkifier2.ts
```diff
@@ -162,9 +162,10 @@ export class Linkifier2 implements ILinkifier2 {
       this._currentLink = link;
       this._showTooltip(this._element, link, this._lastMouseEvent);
 
-      // Add listeners for onData and onScroll
-      this._linkCacheDisposables.push(this._coreService.onData(() => this._clearCurrentLink()));
-      this._linkCacheDisposables.push(this._onScroll(() => this._clearCurrentLink()));
+      // Add listener for rerendering
+      if (this._renderService) {
+        this._linkCacheDisposables.push(this._renderService.onRender(() => this._clearCurrentLink()));
```
Tyriar: We only need to clear the link if the link is on one of the lines between `event.start` or `event.end`

Tyriar:
I might have broken the tests with my merge, not sure if they were broken before, bit busy to fix them atm though

jmbockhorst:
It was failing actually due to a logic error when links are right next to each other. When one link is cleared (Mouse leaves), it needs to ask for a new link immediately.

typings/xterm.d.ts
```diff
@@ -1073,6 +1081,85 @@ declare module 'xterm' {
...
+  interface ILink {
+    /**
+     * The buffer range of the link.
+     */
+    range: IBufferRange;
+
+    /**
+     * The url of the link.
+     */
+    url: string;
```
Tyriar: I'm thinking this should change to just "text" as this may not necessarily be a url, it could be simply "foo" which opens a file called "foo" for example.

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,242 @@
+/**
```
Tyriar: I'll try write up some tests for this.

src/browser/Linkifier2.ts
```diff
@@ -0,0 +1,242 @@
...
+  private _showTooltip(element: HTMLElement, link: ILink, event: MouseEvent): void {
+    const range = link.range;
+    const scrollOffset = this._bufferService.buffer.ydisp;
+
+    this._onShowTooltip.fire(this._createLinkHoverEvent(range.start.x - 1, range.start.y - scrollOffset - 1, range.end.x - 1, range.end.y - scrollOffset - 1, undefined));
+    element.classList.add('xterm-cursor-pointer');
+
+    if (link.showTooltip) {
+      link.showTooltip(event, link.text);
+    }
```
Tyriar: This is now being fired immediately and seems like hover was merged into this event (as renderers get notified about this for hover underlines). Maybe this is fine and providers should handle the timeout themselves, but this is a change.

Regardless, the API names are inaccurate now and should be changes to hover and leave.

Tyriar:
Found a couple of bugs writing some tests: https://github.com/xtermjs/xterm.js/pull/2530/commits/0103c689f7966c4909f51855b90053bdfe15d4cb

Tyriar:
I made a change to use link matcher for `WebLinksAddon` unless an argument is provided, that was my main concern with merging this as link matcher and link providers will probably cause problems when used together. This lets people slowly adopt it with less risk.

jmbockhorst:
Thank you for all of the guidance along the way and for finishing it up. I’ll be looking forward to contributing more in the future.

Tyriar:
@jmbockhorst if you're interested in pulling this into VS Code and hooking it up to `LinkComputer` as well you can give https://github.com/microsoft/vscode/issues/90298 a shot. Basically we'll want an opt-in setting to use the new system, can defer supporting local links initially as getting that right is hard.

I'll probably be updating xterm in vscode in the coming days.

PR #2531, opened 2019-11-01
new settings

jerch:
This is just an early stub for now to test several TS capabilities to propagate type information.

Rough idea:
- collect terminal settings from several `IOptionCollection` definitions across the code
- an option definition (`IOption<T>`) contains all the needed bits for proper initialization like a default value and getter/setter for additional work (optional)
- final settings are contained in a `Settings` object that can handle property updates transparently with these constrains:
  - nested option collections are nested `Settings` objects
  - `Settings` objects never change identity, only last level primitives are mutable
  - properties can be updated by simply doing an assignment like `settings.xy = {a: true, z: {...}}`
- make the final settings object public API replacing current `ITerminalOptions`


jerch:
@Tyriar Made some progress with the settings abstraction. What works so far:

- create settings definition with collections of `IOption<basetype, readonly>`
- create initial values derived from a definition as `InitialOptions<definitionType>`
- create settings object with `createSettings(definition, initialValues)`, type of the returned object is `ISettings<T>` meant a **read interface** (contains all option members)
- settings object contains a special **write interface** under `asWriteable` meant for updates by simple assignments. The separation of read/write interface is needed since Typescript does not allow getter and setter to point to different interfaces. Other than the read interface the write interface contains all non readonly members as optional (for partial updates).
- comments on definition members are correctly forwarded to all derived types
- settings object is stable, assigments never change identity, only the option level is mutable (lowest level), thus the object can be carried around easily

A note on `defineProperty` - I have no clue how to achieve the generic creation of the settings type without it, thus `Settings` relies on that. For nested `IOptionCollections` it simply does a recursive descent. Maybe you have a better idea here.

Left TODO: The final getter/setter/toSettings/fromSettings housekeeping of options is not implemented yet. This is intended to be used later on to trigger additional tasks when a setting gets assigned and such. But before digging deeper there, I'd like get a review of the current stage.

jerch:
> Do we need the readonly/writeable distinction at compile time?

Well if we can go without readonly options, it would simplify alot for sure (the whole write interface stuff would not be needed).

> Where would the option's current sanitize/validate step go?

This could be done in `setter` or `toSettings`, did not further style this yet (would be the next step).

jerch:
@Tyriar:

Slowly getting there, tested the `Settings` class with our current terminal settings. Still far from usable as a new contruction site has opened: many option assignments trigger post actions, thus the class/options need either some kind of post action handler or, to be more versatile, an `onChange` event. Not sure yet how to address it in the end, have to revisit all action conditions first.

Tyriar:
In VS Code the settings change events let you ask whether the change affected a particular scope, eg. if anything under the `terminal.integrated` scope changed then I might action something, or if a particular key changed. I always found this to be handy to avoid extra work.

jerch:
@Tyriar Ah good to know. Do you have some code pointers for me? Maybe I can borrow some ideas :smile_cat:

Tyriar:
Good starting point: https://github.com/microsoft/vscode/blob/434662c6cb11b4cad41bd4bf80c23812aeb319d6/src/vs/platform/configuration/common/configuration.ts#L65

jerch:
Gonna close this PR and keep it under references. Might get back it one day :smiley_cat:

Issue #2444, opened 2019-10-02
Search (upwards) is missing words in wrapped lines


Tyriar:
See https://github.com/microsoft/vscode/issues/81510 for repro

dmitryrn:
I’ll probably check this out

Tyriar:
@dmitryrn you'll want to get the demo up and running, after that the source for the search addon is in addons/xterm-addon-search

dmitryrn:
Yeah, I’ll find out, thanks

Tyriar:
Fixed in https://github.com/xtermjs/xterm.js/pull/2532

PR #2532, opened 2019-11-02
#2444: Fix wrapped lines search

glneto:
### Issue
Wrapped lines are not being considered correctly in search. Because of that, findPrevious and findNext are skipping entries that match the searched pattern.

### Proposed solution
I changed a bit how the findInLine works to use a SearchPosition object instead of param values.
That way we can update the startRow/startCol in the function and make sure the caller function will also consider them.

**For reverse search:**
The searchStart was being calculated incorrectly. Instead of using terminal.cols as the max value, we should consider that wrapped lines can have a startCol higher than this. (Ln. 163)

**For regular search:**
In case we have multiple entries of the search term in a single wrapped line, findNext was finding only the first entry. In case we find a wrapped line, we need to iterate back until we find the line start, while increasing the startCol. This will make us start the search in the correct line/col, and make sure we go through the whole wrapped line before going back to the start.


addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -220,18 +239,28 @@ export class SearchAddon implements ITerminalAddon {
    * started on an earlier line then it is skipped since it will be properly searched when the terminal line that the
    * text starts on is searched.
    * @param term The search term.
-   * @param row The line to  start the search from.
-   * @param col The column to start the search from.
+   * @param {ISearchPosition} position The position to start the search.
```
Tyriar: Can skip the type annotations since we're in TS

```suggestion
   * @param position The position to start the search.
```

glneto:
> Thanks @glneto, sorry about the delay though. I wrote up some thorough regression tests so this shouldn't happen again 🎉

that's amazing! thanks for going through all that :)

PR #2538, opened 2019-11-04
Experimenting with GH Actions

Tyriar:


Tyriar:
Ended up polishing azure pipelines instead of this

PR #2540, opened 2019-11-05
Update TS to 3.7 and adopt optional chaining

Tyriar:
@jerch fyi new syntax https://devblogs.microsoft.com/typescript/announcing-typescript-3-7/


jerch:
@Tyriar Oh nice, I see here some gems :smile_cat:

PR #2541, opened 2019-11-05
Skip puppeteer download on unit test runs, fix windows tests, polish pipeline

### LINKED ISSUE #2539: Split out puppeteer's install to be acquired separately to the rest of the modules
It downloads Chromium which means is busts the GH Actions cache if we try to cache node_modules and also takes a long time to install locally and in CI. We should move it out and only install it when we need the API tests (`cd test/api && yarn`? `"pretest-api": "npm install puppeteer"`?).

### LINKED ISSUE #2542: Replace build quality CI check with nyc --check-coverage

### LINKED ISSUE #2543: Unit tests are not running on Windows
https://dev.azure.com/xtermjs/xterm.js/_build/results?buildId=2643

Probably the slashes or not using .cmd in ./bin/test.js

Tyriar:
Fixes #2539
Fixes #2542
Fixes #2543

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -98,6 +98,11 @@ jobs:
     inputs:
       versionSpec: '8.x'
     displayName: 'Install Node.js'
+  - task: CacheBeta@1
+    inputs:
+      key: yarn_puppeteer | $(Agent.OS) | yarn.lock
+      path: node_modules
+    displayName: Cache Yarn packages
   - task: YarnInstaller@3
     inputs:
       versionSpec: '1.x'
@@ -120,6 +125,11 @@ jobs:
     inputs:
       versionSpec: '8.x'
     displayName: 'Install Node.js'
+  - task: CacheBeta@1
+    inputs:
+      key: yarn_puppeteer | $(Agent.OS) | yarn.lock
+      path: node_modules
+    displayName: Cache Yarn packages
   - script: yarn --frozen-lockfile
     displayName: 'Install dependencies and build'
   - script: yarn add puppeteer
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -22,7 +22,7 @@ jobs:
     inputs:
       key: yarn2 | $(Agent.OS) | yarn.lock
       path: node_modules
-    displayName: Cache Yarn packages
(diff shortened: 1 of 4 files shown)


Tyriar:
Skipping puppeteer download seems to save about 15 seconds on Linux, let's see if caching yarn cache works

Tyriar:
- Baseline (2m58s): https://dev.azure.com/xtermjs/xterm.js/_build/results?buildId=2643
- Cache yarn cache, no puppeteer
  - Miss (3m22s, cache took 28s): https://dev.azure.com/xtermjs/xterm.js/_build/results?buildId=2647
  - Hit: (2m29s): https://dev.azure.com/xtermjs/xterm.js/_build/results?buildId=2648 
- Cache node_modules (~300mb), no puppeteer
  - Miss (3m12s)
  - Hit (2m15s)

jerch:
@Tyriar So the difference is basically a minute for a single pipeline?

Tyriar:
@jerch yeah, integration tests are mostly unchanged at but I'll see if I can reduce that a bit too. So right now we'll get faster compile/unit test feedback.

Tyriar:
Caching node_modules for integration tests didn't seem to work, also it ended up as 1gb so it was going to be slower anyway if it did.

Tyriar:
Final results gives faster compile and unit test builds to let us iterate faster on PRs:

| &nbsp; | [Before](https://dev.azure.com/xtermjs/xterm.js/_build/results?buildId=2643) | [After](https://dev.azure.com/xtermjs/xterm.js/_build/results?buildId=2669)
|---|---|---
| **Linux** | 2m 58s | 1m 42s
| **macOS** | 1m 58s | 1m 50s\*
| **Windows** | 3m 23s\** | 2m 3s

\* The actual run took 2m 23s but that's because cache restore took an abnormally long time just this run, number was adjusted for typical runtime
\*\* This is the approx value, Windows wasn't running tests before so it's the actual number 2m 55s plus 28s for unit tests

Integration tests are basically the same but I'll look into using multi-stage pipelines instead of having 2 separate ones for Linux and macOS.

Tyriar:
This passed fully on a rebuild

Issue #2252, opened 2019-06-23
Webgl: Support allowTransparency option


Tyriar:
Webgl addon added in https://github.com/xtermjs/xterm.js/pull/1790

I'm not sure what the state of this is currently.

Eugeny:
Any updates on this one? This and character joiner support are the only things keeping me from using the WebGL renderer as default in Terminus.

Tyriar:
No updates, needs someone to look into it until I get time to invest into bringing it to vscode.

Tyriar:
Transparency seems to mostly work in from https://github.com/xtermjs/xterm.js/pull/2560/commits/ef05a6a05354901eef52177d03858a4e05111697 in https://github.com/xtermjs/xterm.js/pull/2560:

<img width="818" alt="Screen Shot 2019-11-25 at 2 33 26 PM" src="https://user-images.githubusercontent.com/2193314/69583854-a8356400-0f90-11ea-9910-4980411caab1.png">

The scrollbar seems a bit off though and haven't tested very thoroughly.

Tyriar:
I tried to not draw a background rectangle at all as some colors ended up being off. There are two options for whoever wants too look into this in the future:

- Explore not using a background color for the viewport (would be inconsistent with other renderers)
- Explore not using a completely transparent background for webgl renderer (little worried about regressions here)

Eugeny:
Works perfectly for me, including the scrollbar!

Tyriar:
@Eugeny oh? The screenshot above of the demo is showing that the alpha color isn't correct as it's drawn twice (once in `.xterm-viewport`, once in `canvas`).

Eugeny:
@Tyriar my primary concern was being able to have a completely transparent background - even in the current state `transparent` being drawn twice results in a usable transparent webgl xterm :)

Tyriar:
@Eugeny ah ok, good 😃 I'll keep this open until the background color is correct though.

Tyriar:
I believe the background color problem is fixed, notice how the right of the row is the same color:

![image](https://user-images.githubusercontent.com/2193314/181916880-b4c77f30-b1d5-4f44-9054-930364acc3c1.png)

There have also been several fixes that have gone in for v5. Let's close this off and track individual issues as they come up as this one's a little ambiguous what's left to do.

PR #2560, opened 2019-11-11
force alpha to 1 when using background color as inverted foreground color

### LINKED ISSUE #1898: Handling transparent BG color when used as FG
Nano displays its shortcuts with FG/BG colors reversed:
image

However, when theme BG is `transparent`, following happens:
image

The solution could be to have a separate flag that will render characters with no background as if background were transparent instead of `theme.backgroundColor`, but will still use the actual color for inverted elements.

ivanwonder:
This problem has been described there #1898.

Fixes #1898


src/browser/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -253,7 +254,13 @@ export class DynamicCharAtlas extends BaseCharAtlas {
       `${fontStyle} ${fontWeight} ${this._config.fontSize * this._config.devicePixelRatio}px ${this._config.fontFamily}`;
     this._tmpCtx.textBaseline = 'middle';
 
-    this._tmpCtx.fillStyle = this._getForegroundColor(glyph).css;
+    const fgColor = this._getForegroundColor(glyph);
+    this._tmpCtx.fillStyle = fgColor.css;
+
+    if (glyph.fg === INVERTED_DEFAULT_COLOR) {
+      const rgba = fromRgba(fgColor.rgba);
+      this._tmpCtx.fillStyle = toCss(rgba[0], rgba[1], rgba[2]);
```
Tyriar: These transformations are a little more expensive that needed, we could just have a helper that converts an `IColor` (or css or rgba) to be opaque, for example for rgba all that needs to happen is `rgba | 0xFF`.

src/browser/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -253,7 +254,13 @@ export class DynamicCharAtlas extends BaseCharAtlas {
       `${fontStyle} ${fontWeight} ${this._config.fontSize * this._config.devicePixelRatio}px ${this._config.fontFamily}`;
     this._tmpCtx.textBaseline = 'middle';
 
-    this._tmpCtx.fillStyle = this._getForegroundColor(glyph).css;
+    const fgColor = this._getForegroundColor(glyph);
```
Tyriar: For the WebGL renderer it should be a very similar here:

https://github.com/xtermjs/xterm.js/blob/339d4d6a7316d1c1791b7a5859ba9c8175ce3cc6/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts#L209-L211

src/browser/renderer/atlas/DynamicCharAtlas.ts
ivanwonder: now I add new function `opaque` in Color.ts

Eugeny:
Works with the canvas renderer, but not WebGL. I've poked around a bit but could find why right away

Tyriar:
@Eugeny do you mean transparency in general or this specific issue (`\x1b[7m█\x1b[0m` should be `#ff0000` when the background color is `#ff000080` and allowTransparency is true?

Eugeny:
I meant #1898

Tyriar:
@Eugeny is my test case correct? That was working on all 3 renderers when I tested.

Eugeny:
With a white-on-transparent theme, 

Canvas:
image

WebGL:
image

Tyriar:
@Eugeny what's the background set to exactly?

Eugeny:
Literally `"transparent"`

Tyriar:
@Eugeny oh, that's the problem. I was assuming you were using `#rgba` or `#rrggbbaa`, it should work if you change that to `#00000000`. I hadn't even considered using `transparent` here, that's also why you don't get the scrollbar problem in https://github.com/xtermjs/xterm.js/issues/2252#issuecomment-558371374

Tyriar:
@Eugeny what's your proposal to fix that? It feels like that's sort of undefined behavior and we shouldn't allow using `transparent` for this reason, using 0 for the alpha channel instead lets you specify the opaque inverted version.

Eugeny:
What do you use to convert named colors to RGBA? `transparent` should be equivalent to `rgba(0, 0, 0, 0)` per CSS spec

Tyriar:
@Eugeny 👍 opening issue again to add this case:

https://github.com/xtermjs/xterm.js/blob/8b5c48fe878f830914248e4c1ed78e621deb16c4/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts#L231-L236

ivanwonder:
@Tyriar Because of the `transparent`, so I remove the bg alpha channel by using the RGBA value and then convert it to CSS string. 
https://github.com/ivanwonder/xterm.js/blob/1ff1809ab75be9a9aeee134e6dd5c040f68efda0/src/browser/Color.ts#L59

Tyriar:
@ivanwonder the fix I did for webgl didn't cover when background=transparent though, I'll fix that before we do a release.

ivanwonder:
@Tyriar https://html.spec.whatwg.org/multipage/canvas.html#serialisation-of-a-color
`context.fillStyle` will convert `transparent` to `rgba(0, 0, 0, 0)`.
`if it has alpha equal to 1.0, then the string is a lowercase six-digit hex value, prefixed with a "#" character, Otherwise, the color value has alpha less than 1.0, and the string is the color value in the CSS rgba()`
maybe it will help

Tyriar:
@ivanwonder thanks, good info

ivanwonder:
I just try the idea that using `context.fillStyle`. Looks work fine.
https://github.com/ivanwonder/xterm.js/tree/reverse-transparency

PR #2566, opened 2019-11-13
onBinary event

### LINKED ISSUE #1962: X10 mouse protocol is capped at 127

### LINKED ISSUE #2545: implement onBinary data event
Analogous to `onData` we need a `onBinary` event to support sending raw byte data to the backend. This is needed to solve issues with mouse reports.

Also add this to the encoding article later on:
> Beside string data `xterm.js` might send raw byte data with the `onBinary` event. For easier consumption the payload is a JS string, but should be treated as 8-bit data (`'binary'` encoding in nodejs). Currently this is only used for mouse reports that cannot be encoded legally in UTF-8.

jerch:
Implements an `onBinary` event to be used to send non UTF-8 compatible data to the backend. Currently only used for DEFAULT mouse reports.

The event data is a string, but its important to treat it as a binary string, example with `node-pty`:
```TS
term.onBinary(data => pty.write(Buffer.from(data, 'binary')));
```

in difference to the normal UTF8 data handling:
```TS
term.onData(data => pty.write(data));
// equivalent to
term.onData(data => pty.write(Buffer.from(data, 'utf-8')));
```

Closes #2545, fixes #1962.

Reminder: Add note to encoding article when done.

### DIFF
--- a/src/common/services/CoreMouseService.test.ts
+++ b/src/common/services/CoreMouseService.test.ts
@@ -6,7 +6,7 @@ import { CoreMouseService } from 'common/services/CoreMouseService';
 import { MockCoreService, MockBufferService } from 'common/TestUtils.test';
 import { assert } from 'chai';
 import { ICoreMouseEvent, CoreMouseEventType, CoreMouseButton, CoreMouseAction } from 'common/Types';
-
+declare const console: any;
 // needed mock services
 const bufferService = new MockBufferService(300, 100);
 const coreService = new MockCoreService();
@@ -79,6 +79,7 @@ describe('CoreMouseService', () => {
       cms = new CoreMouseService(bufferService, coreService);
       reports = [];
       coreService.triggerDataEvent = (data: string, userInput?: boolean) => reports.push(data);
+      coreService.triggerBinaryEvent = (data: string) => reports.push(data);
     });
     it('NONE', () => {
       assert.equal(cms.triggerMouseEvent({ col: 0, row: 0, button: CoreMouseButton.LEFT, action: CoreMouseAction.DOWN }), false);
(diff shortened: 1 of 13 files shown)


src/common/services/CoreMouseService.test.ts
```diff
@@ -143,11 +144,11 @@ describe('CoreMouseService', () => {
         cms.activeProtocol = 'ANY';
         for (let i = 0; i < bufferService.cols; ++i) {
           assert.equal(cms.triggerMouseEvent({ col: i, row: 0, button: CoreMouseButton.LEFT, action: CoreMouseAction.DOWN }), true);
-          // capped at 95
-          if (i < 95) {
-            assert.deepEqual(toBytes(reports.pop()), [0x1b, 0x5b, 0x4d, 0x20, i + 33, 0x21]);
+          if (i > 222) {
+            // supress mouse reports if we are out of addressible range (max. 222)
```
Tyriar: Also note that the suppress only happens there because they go through `onData` until that point is reached?

test/api/MouseTracking.api.ts
```diff
@@ -220,6 +220,7 @@ describe('Mouse Tracking Tests', () => {
     await page.evaluate(`
       window.calls = [];
       window.term.onData(e => calls.push( Array.from(e).map(el => el.charCodeAt(0)) ));
+      window.term.onBinary(e => calls.push( Array.from(e).map(el => el.charCodeAt(0)) ));
```
Tyriar: Should you differentiate the 2 different call types or do you think it's good enough like this?

src/common/services/CoreMouseService.ts
```diff
@@ -121,15 +121,13 @@ const DEFAULT_ENCODINGS: {[key: string]: CoreMouseEncoding} = {
...
-   * to 95 in xterm.js).
+   * Can encode values up to 223 (1-based).
    */
   DEFAULT: (e: ICoreMouseEvent) => {
-    let params = [eventCode(e, false) + 32, e.col + 32, e.row + 32];
-    // FIXME: we are currently limited to ASCII range
-    params = params.map(v => (v > 127) ? 127 : v);
-    // FIXED: params = params.map(v => (v > 255) ? 0 : value);
+    const params = [eventCode(e, false) + 32, e.col + 32, e.row + 32];
+    if (params[0] > 255 || params[1] > 255 || params[2] > 255) {
```
Tyriar: This means something went wrong?

src/common/services/CoreMouseService.test.ts
jerch: Ah well thats the hard cap here, the protocol does not support any higher. Can rephrase it.

test/api/MouseTracking.api.ts
jerch: Hmm, I think thats ok here, I kinda treat the calls array as utf-16 codepoints, should correctly encode all needed bits here (while utf8 cannot).

src/common/services/CoreMouseService.ts
jerch: Thats the max addressible range for the protocol. Gonna add a note.

src/common/services/CoreMouseService.test.ts
jerch: Rephrased it in the encoding itself.

src/common/services/CoreMouseService.ts
jerch: Added further notes to this. The empty string now will skip any report to be sent.

PR #2568, opened 2019-11-15
wcwidth rules for unicode 11

jerch:
A new attempt to solve the emoji issues. It can already be tested with this PR, but is still in a very early state.

~Decided not to go the full unicode provider path, gonna start just with wcwidth table replacements. This way we can still deliver the old and the new versions in core, that can be selected by some option later on. Still the unicode provider approach of earlier attempts might get relevant again, if we go for other unicode features in the future (e.g. bidi, grapheme).~ Edit: UnicodeService added, V11 comes as addon implementation.

This PR gonna fix most emoji issues on newer systems. There are some newer unicode compose rules, that cannot be covered by simple wcwidth lookups. If we want those, we would have to implement true grapheme support. I already investigated in this direction some months ago with the result that this is not respected by any console program either. Thus not worth the impl headache atm.

TODO:
- ~shrink data tables (create merge-diff to old table?)~ not maintainable in the long run
- cleanup

@Tyriar, @mofux Could you test under macOS and Windows if the emojis get the right width? (Still have no clue what Unicode versions these use for a certain OS version)


jerch:
Differences to before:
- +65KB lookup table at runtime
- +10KB source code

Imho the +10KB in source code hurt not much, since the V10 tables are somewhat repetitive to V6, I could prolly lower that to +5KB  with a diff-merge. Not sure if its worth the trouble (much harder to maintain). Ideas?

src/browser/Linkifier.ts
```diff
@@ -277,8 +278,7 @@ export class Linkifier implements ILinkifier {
     if (!this._mouseZoneManager || !this._element) {
       return;
     }
-
-    const width = getStringCellWidth(uri);
+    const width = this._optionsService.options.unicodeVersion === '11' ? getStringCellWidthV11(uri) : getStringCellWidthV6(uri);
```
mofux: I'm afraid having to do something like this will get quite unmaintainable in the long run. I'd much rather see this abstracted away in a `UnicodeService` or similar, so the linkifier (or addons) wouldn't have to deal with such an implementation detail:

```typescript
const width = this._unicodeService.getStringCellWidth(uri);
```

instead of
```typescript
const width = this._optionsService.options.unicodeVersion === '11' ? getStringCellWidthV11(uri) : getStringCellWidthV6(uri);
```

mofux:
Tried on OSX and regular emojis won't break my prompt anymore 🎉 Combined emojis (👨‍👩‍👦‍👦) don't work, but they still look better than in `Terminal.app` and `iTerm` 👍

jerch:
@mofux Thx for testing. Yeah those compound emojis will still behave differently, its not doable with wcwidth calculations alone. This would need grapheme parsing rules + output metrics returned by the font renderer. Both is hard to achieve and cannot be made failsafe across diff. systems (even fonts), well it should not be part of this PR.

Youre right about with your concerns regarding addons, have not taken those into account. Guess I gonna revive the provider idea and put it into a service.

src/browser/Linkifier.ts
jerch: Answered below.

Tyriar:
@jerch I forgot, did the other terminals have settings for the unicode version or just try to the best fit one based on popular OS's?

jerch:
No most I looked into have one static wcwidth version, that evtl. gets updated with new releases and OS updates along the way. Only one I remember with configurable unicode settings is iTerm2.

Tyriar:
> Youre right with your concerns regarding addons, have not taken those into account. Guess I gonna revive the provider idea and put it into a service.

I think the model with the renderer works pretty well; shipping a (small) standard implementation (v6) and allowing to plug in a different one (v11). Using addons also enables the use of code splitting, while it might not make much of a difference by itself I fear death by 1000 papercuts if we keep including more and more into the core. Any opportunity for code splitting is great imo.

I remember tossing the idea around of having the unicode ones special so v6 isn't loaded at all if we're not using it, that might be over complicating things and we could defer improving upon that at a later date.

jerch:
Refactored the PR to contain an UnicodeService, which maintains different Unicode versions. By default its linked with V6 (hardcoded), any other version can be added by an addon later on. Created an example addon for V11.

mofux:
@jerch @Tyriar What are your thoughts on making `v11` the default, and shipping `v6` as an addon? In my eyes making `v11` the default does more good than harm. Maybe a change like that should at least be considered for the next major release?

jerch:
@mofux Sounds good to me, I think most common systems are already on v10+. Only systems that will suffer are older server setups, like CentOS 7 or lower, Ubuntu prior 18 LTS, not sure when Debian switched to v9+. We should at least keep V6 around as long as those are still in actively maintained (at least a couple more years).

Tyriar:
> What are your thoughts on making v11 the default, and shipping v6 as an addon?

If the experience is better on Windows 10 and macOS which I suspect it will be then :+1:

Tyriar:
Pasting in `😀😁😂🤣😃😄😅😆😉😊😋😎😍😘🥰😗😙😚☺️🙂🤗🤩🤔🤨😐😑😶🙄😏😣😥😮🤐😯😪😫😴😌😛😜😝🤤😒😓😔😕🙃🤑😲☹️🙁😖😞😟😤😢😭😦😧😨😩🤯😬😰😱🥵🥶😳🤪😵😡😠🤬😷🤒🤕🤢🤮🤧😇🤠🤡🥳🥴🥺🤥🤫🤭🧐🤓😈👿👹👺💀👻👽🤖💩😺😸😹😻😼😽🙀😿😾` into the demo breaks and it looks like they're all single width?

Screen Shot 2019-12-07 at 1 35 41 PM

The cursor is the end of the line in above, pressing left jumps 2 characters though 😕

demo/client.ts
```diff
@@ -69,7 +74,8 @@ const addons: { [T in AddonType]: IDemoAddon<T>} = {
   fit: { name: 'fit', ctor: FitAddon, canChange: false },
   search: { name: 'search', ctor: SearchAddon, canChange: true },
   'web-links': { name: 'web-links', ctor: WebLinksAddon, canChange: true },
-  webgl: { name: 'webgl', ctor: WebglAddon, canChange: true }
+  webgl: { name: 'webgl', ctor: WebglAddon, canChange: true },
+  unicode11: { name: 'unicode11', ctor: Unicode11Addon, canChange: false }
```
Tyriar: Cannot change?

typings/xterm.d.ts
```diff
@@ -215,6 +215,11 @@ declare module 'xterm' {
      */
     theme?: ITheme;
 
+    /**
+     * Unicode version to be used for wcwidth. Default is version 6.
+     */
+    unicodeVersion?: string;
```
Tyriar: We're using the addon approach right so a setting wouldn't be able to load it?

tsconfig.all.json
```diff
@@ -9,6 +9,7 @@
     { "path": "./addons/xterm-addon-fit/src" },
     { "path": "./addons/xterm-addon-search/src" },
     { "path": "./addons/xterm-addon-web-links/src" },
-    { "path": "./addons/xterm-addon-webgl/src" }
+    { "path": "./addons/xterm-addon-webgl/src" },
+    { "path": "./addons/xterm-addon-unicode11/src" }
```
Tyriar: 2 lines up for alphabetical 😃

addons/xterm-addon-unicode11/src/Unicode11Addon.ts
```diff
@@ -0,0 +1,265 @@
...
+ * polyfill for TypedArray.fill
+ * This is needed to support .fill in all safari versions and IE 11.
+ */
+function fill<T extends TypedArray>(array: T, value: number, start?: number, end?: number): T {
+  // all modern engines that support .fill
+  if (array.fill) {
+    return array.fill(value, start, end) as T;
+  }
+  return fillFallback(array, value, start, end);
+}
```
Tyriar: I think we can link directly to TypedArrayUtils like in the WebGL addon? We can probably set up tree shaking eventually to avoid pulling in things we're not using but it's a pretty small file already.

addons/xterm-addon-unicode11/src/Unicode11Addon.ts
```diff
@@ -0,0 +1,265 @@
...
+      for (let r = 0; r < UnicodeV11.bmpCombining.length; ++r) {
+        fill(UnicodeV11.bmpTable, 0, UnicodeV11.bmpCombining[r][0], UnicodeV11.bmpCombining[r][1] + 1);
+      }
+      for (let r = 0; r < UnicodeV11.bmpWide.length; ++r) {
+        fill(UnicodeV11.bmpTable, 2, UnicodeV11.bmpWide[r][0], UnicodeV11.bmpWide[r][1] + 1);
+      }
+    }
+  }
+
+  public wcwidth(num: number): 0 | 1 | 2 {
```
Tyriar: `type CharWidth = 0 | 1 | 2` since it's used in multiple places?

addons/xterm-addon-unicode11/src/Unicode11Addon.ts
```diff
@@ -0,0 +1,265 @@
...
+    end = (array.length + end) % array.length;
+  }
+  for (let i = start; i < end; ++i) {
+    array[i] = value;
+  }
+  return array;
+}
+
+class UnicodeV11 implements IUnicodeVersionProvider {
+  static bmpCombining = [
```
Tyriar: Normally instead of static I prefer to just put them outside fo the class, I think this one can be const/readonly too?

src/browser/Linkifier.ts
```diff
@@ -277,8 +278,8 @@ export class Linkifier implements ILinkifier {
     if (!this._mouseZoneManager || !this._element) {
       return;
     }
-
-    const width = getStringCellWidth(uri);
+    // FIXME: get cell length from buffer to avoid mismatch after Unicode version change
+    const width = this._unicodeService.getStringCellWidth(uri);
```
Tyriar: This might be tricky?

demo/client.ts
jerch: Ah thought not to much about this one lol. Ofc it can be loaded lazy, but cannot be unloaded once in use (or the terminal must revert to something else).

typings/xterm.d.ts
jerch: Yes, basically an unicode addon will feed a version string to `UnicodeService.versions` upon registering, and unicodeVersion must be one of these version strings. Due to lazy loading we dont know them prehand.

addons/xterm-addon-unicode11/src/Unicode11Addon.ts
jerch: Oops yes ofc. :+1: 

Well that design had a rather big evolution where I had to fight bad runtime penalty due to switching `wcwidth` under the hood (v8 kept running into deopts all the time), so had to fiddle around until v8 was happy about it. Kinda forgot the do the final cleanup (like make it module global and static lol).

src/browser/Linkifier.ts
jerch: Well its possible, but needs some work, thus I opted for the easy way to get this up. Also its a rather rare edge case - this will only trigger links wrongly if:
- the link contains wide chars at all (very unlikely)
- the unicode version got changed between `Terminal.write` and the linkifying (even more unlikely)

Thus somewhat debatable whether this becomes  a real issue or is just an academic problem.

jerch:
> Pasting in `😀😁😂🤣😃😄😅😆😉😊😋😎😍😘🥰😗😙😚☺️🙂🤗🤩🤔🤨😐😑😶🙄😏😣😥😮🤐😯😪😫😴😌😛😜😝🤤😒😓😔😕🙃🤑😲☹️🙁😖😞😟😤😢😭😦😧😨😩🤯😬😰😱🥵🥶😳🤪😵😡😠🤬😷🤒🤕🤢🤮🤧😇🤠🤡🥳🥴🥺🤥🤫🤭🧐🤓😈👿👹👺💀👻👽🤖💩😺😸😹😻😼😽🙀😿😾` into the demo breaks and it looks like they're all single width on macOS Catalina 10.15.1?
> 
> Screen Shot 2019-12-07 at 1 35 41 PM
> 
> The cursor is the end of the line in above, pressing left jumps 2 characters though confused

Tested your string of emojis - works for me (Ubuntu 18). Did you set `unicodeVersion` to `'11'`?

jerch:
@Tyriar Tried to test this under windows with these results:
- winpty does not work with these emojis above, it never returns the unicode char if the codepoint is > 0xFFFF, always some replacement
- with conpty I get 2 times � (\ufffd) + more garbage chars from node-pty, if I press backspace the actual emoji gets through with more garbage

I am not sure whether the issues come from node-pty or winpty/conpty, its def. nothing we can fix at xterm.js level (output of node-pty is already broken). So I cannot address it with this PR.

About v6 vs. v11 as default:
Thinking again about this, maybe a more slow paced change will result in less possible frictions - I suggest to stick with v6 as default and deliver v11 as addon, and swap this with the next major release. This way users can test v11 and report back major issues before we roll it out on a much bigger userbase as default. I dont see any other way how to get some testing/fixes done for this, the "correctness" of the width of a particular char is even more questionable since the Unicode consortium clearly states, that the asian width flag is only a weak indicator and not meant for terminal wcwidth business at all.

demo/client.ts
jerch: Changed it to true, but note that unloading has no effect.

src/browser/Linkifier.ts
jerch: I think we should postpone this because of the new linkiifer, which changes alot of the internal business.

jerch:
Thought about the question whether loading an Unicode version addon should activate that version right away. Personally I am not a fan of this, as it kinda breaks with the "loading a plugin should add functionality but have no surprising side effects" rule. Also unloading of a unicode addon is not supported (as its almost not needed and would introduce nonsense bookkeeping code), where coupling activation on loading raises the question whether disposing an addon should deactivate and remove that version.
Furthermore binding the version switch on loading/unloading makes multiple version support kinda awkward, as the active version depends on addon loading order. I'd call this a side effect leading to a weird constraint to be respected during coding.

Imho we should stick with only registering a version through loading. The version can be activated right after that.

Up for a new review.

thefallentree:
the multi code-point emoji are always chained with zwj-s , right? couldn't we have an special rule to handle that? At least that way dom renderer will work. Right now it is broken into multiple <span>'s

jerch:
> the multi code-point emoji are always chained with zwj-s , right? couldn't we have an special rule to handle that? At least that way dom renderer will work. Right now it is broken into multiple 's

The problem here is the unicode spec itself about those - its up to the renderer to decide whether to print them as compound emojis (if it is capable to join glyphs or a font provides compound glyphs), it also states that if that not the case - print them as a sequence of single emojis. Now we a problem - the actual "run width" depends on renderer caps, this NEVER can work reliable for app side in a terminal, since app side always operates on an abstract screen idea. The terminal interface does not allow us to promote output/renderer caps to app side, thus we have to find a least denominator, thats likely to work under all circumstances. Thus print them as single emojis.

thefallentree:
well terminal is just an medium to display things, just like the browser, old browser don't support emojis and eventually everyone upgraded and it did.   The browser will *not* stop upgrading and eventually it will correctly render everything according to the rules. it's up to the application to decide if it can or can not rely on the terminal to display things correctly, just like some website tell you you need firefox to work, not IE.  

Also, I don't think there is a strict requirement on making sure dom render and canvas render are exactly the same, there is no reason to break dom renderer explicitly to match what canvas can do

jerch:
> Also, I don't think there is a strict requirement on making sure dom render and canvas render are exactly the same, there is no reason to break dom renderer explicitly to match what canvas can do

Not sure what you mean here? There no explicit code telling the DOM renderer to follow the canvas' way to render things. This split up is done before at terminal buffer level without any renderer involvement. We cannot pull renderer stuff into there, since those parts of xterm.js have to keep working without any real output being attached (as in nodejs). Thats the reason why we cannot rely on renderer caps at this stage.

Tyriar:
When I go onto some emoji and press space it seems to mess a lot of stuff up, notice some of the spaces have width of 2 and pressing space duplicates a lot of symbols:

Kapture 2020-01-03 at 7 31 54

addons/xterm-addon-unicode11/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Usage
+
+```ts
+import { Terminal } from 'xterm';
+import { Unicode11Addon } from 'xterm-addon-unicode11';
+
+const terminal = new Terminal();
+const unicode11Addon = new Unicode11Addon();
+terminal.loadAddon(unicode11Addon);
```
Tyriar: Add `terminal.setOption` example?

thefallentree:
Btw, I have this script to generate stuff from Unicode table

https://github.com/fluffos/fluffos/tree/master/src/thirdparty/widecharwidth

I am currently fixed it to 12.1 but it seems if xtermjs would only support unicode11 I would need to downgrade backend as well.

I think an decision has to be made on what version to support, and I propose it to be same as the hosting browser( not that we have we to detect it right now) because in an application , xtermjs need to have some consistentcy with surrounding HTML elements, if an emoji works in an input box , but not when write to the terminal it is very confusing to the user.

src/public/Terminal.ts
```diff
@@ -255,3 +258,23 @@ class ParserApi implements IParser {
     return this.registerOscHandler(ident, callback);
   }
 }
+
+class UnicodeApi implements IUnicodeHandling {
+  constructor(private _core: ITerminal) {}
+
+  public register(provider: IUnicodeVersionProvider): void {
+    (this._core as any)._unicodeService.register(provider);
```
Tyriar: How about making `public readonly unicodeService` on src/Terminal.ts and adding it to the `ITerminal` interface to avoid this cast?

typings/xterm.d.ts
```diff
@@ -215,6 +215,11 @@ declare module 'xterm' {
      */
     theme?: ITheme;
 
+    /**
+     * Unicode version to be used for wcwidth. Default is version 6.
```
Tyriar: Some more info?

```
... New unicode versions can be registered through the `Terminal.unicode.register` API.
```

typings/xterm.d.ts
```diff
@@ -1155,4 +1166,29 @@ declare module 'xterm' {
...
+   */
+  export interface IUnicodeHandling {
+    /**
+     * Register a custom Unicode version provider.
+     */
+    register(provider: IUnicodeVersionProvider): void;
+    /**
+     * Registered Unicode versions.
+     */
+    versions: string[];
```
```suggestion
    readonly versions: ReadonlyArray<string>;
```

typings/xterm.d.ts
Tyriar: What do you think about not making this a traditional setting but instead making `Terminal.unicode.activeVersion` have a getter and setter? It feels weird to have settings influenced by addons like this.

I still haven't thought this through, the idea just popped up though.

typings/xterm.d.ts
```diff
@@ -1155,4 +1166,29 @@ declare module 'xterm' {
...
     addOscHandler(ident: number, callback: (data: string) => boolean): IDisposable;
   }
+
+  /**
+   * (EXPERIMENTAL) Unicode version provider.
+   * Used to register custom Unicode versions with `Terminal.unicode.register`.
+   */
+  export interface IUnicodeVersionProvider {
+    /** String indicating the Unicode version provided. */
+    version: string;
```
Tyriar: readonly

thefallentree:
Also, I think we need to distinguish an display issue vs functionality issue. For example, that 7 code point emoji should always be drawn into an 2 width column. However maybe canvas renderer will mess it up and drawn broken chacters, or even overflow, but that is an display issue, even display broke. but to the user / cursor it should still occupy 2 column width. Yes it is broken graphically, but it will be automatically fixed when renderer is fixed.

However like in dom renderer right now, it is broken in two 4x2cols , it then become an functionality issue, and it won’t be fixed even if chrome already supports displaying it correctly.

So, in conclusion , I am saying functionality wise everyone should always stick to rules of latest Unicode standard, and display issue should be solved at display layer. Maybe renderer could use some hack to render unsupported character with placeholder png files instead of relying on font.

But, display issue should just be display issue.

Tyriar:
@thefallentree all renderers are meant to display essentially all the same thing (AA and exact pixel positioning differs) and are pretty separate discussion from this PR which is purely focused on how each cell that comes in from the parser is allocated a width (0, 1 or 2).

jerch:
The discussion gets somewhat out of bound here. @thefallentree Yes we would have to support multiple Unicode versions at once to give best experience to the users. Currently v11 tries to be a compromise to get the most out of the newer versions while not hurting ppl still on v9-v10 too much. Later on we prolly will have multiple addons with different versions.

Ideally the Unicode version is in line between between browser host system, xterm.js and app side. Then things go smoothly. Now imagine someone accessing an older remote host with an older Unicode version. Here the user should be able to switch xterm.js' version, otherwise libreadline will go bonkers on the terminal lines. The least interesting version in that chain is what the browser system actually supports.

@Tyriar Can you track down this behavior to single emojis? I suspect a version mismatch between libreadline of the shell and our v11 implementation (much like I found above for Ubuntu 18 LTS).

Tyriar:
@jerch well ☺️ definitely has some problems as you can see it's the first overlapping one. Going back to it and pressing space messes things up. After pasting it and pressing left, space:

Screen Shot 2020-01-03 at 8 25 50 AM

thefallentree:
Hey guys, I appreciate and enjoyed the discussion here .  I am still trying to wrap around my head on how xtermjs deal with characters . It seems we can circle back to the PTY idea that it by introducing an PTY abstractIon and make xtermjs an true character base interface it would be better. Of course we still need to have different Unicode version of PTY that matches remote pty, but that seems could be done/detected in the forker level.

Regardless, I appericate the new interface, I would be happy to help in anyway, since I plan to use unicdoe12.1 in my non pty application anyway so I would need to hook in Unicode 12.1 support to xtermjs using the new interface

jerch:
@Tyriar Could you test ☺️ with Terminal.app and iTerm2? Under Mojave I see the same weird behavior for both emulators.

Looking closer on your emoji reveals, that this emoji has a representation variation selector appended: `\u263a\ufef0`. The U+FEF0 would force the renderer to use the pictogram representation of the emoji (thats needed for U+263A since it defaults to text representation). And now it gets weird: This emoji belongs to the oldest set of emojis (from 1992 or so) and had a text representation for long as half width. With Unicode 9 they changed all pictogram emojis into full width symbols, and stated that this should be applied to all upcoming emojis as well. But for those old emojis the old text repr was kept, which leads to this:
- old emoji in text repr: width 1
- old emoji in pictogram repr: width 2

Now looking at Unicode v10 up to v12 documents and the Emoji annexes, they dont mention a different width handling anymore and just refer to east asian width (beside country flags being at >150%). Since all other emulators and shells (readline/editline) show the same "broken" behavior, I wonder whether there is another document stating that the text repr for old emojis should not change at all. I think currently we cannot do much about this.

typings/xterm.d.ts
jerch: @Tyriar Internally it already is a getter/setter, I didnt want to deviate from the general options handling to much in the public API. I can make this a getter/setter too, but isnt it abit weird that while all other options are orchestrated by `get/setOption` this will behave differently?

Tyriar: I'm leaning towards using `unicode.activeVersion` because it can't be set in the ctor like a regular option (similar to cols/rows with `resize` which probably shouldn't be options either) and it feels wrong to have a nicely organized `unicode` namespace but then you set it somewhere else.

jerch: Ok, then lets just remove it from options.

jerch: Its gone.

jerch:
> Btw, I have this script to generate stuff from Unicode table
https://github.com/fluffos/fluffos/tree/master/src/thirdparty/widecharwidth

@thefallentree Thats pretty awesome, could we refactor this to a general unicode version extractor? I have something similar written JS, but its just a bunch of scripts that still needs handwork to get the final tables extracted. I plan to add this with the next unicode PR for more versions, maybe I can use yours instead?

thefallentree:
> > Btw, I have this script to generate stuff from Unicode table
> > https://github.com/fluffos/fluffos/tree/master/src/thirdparty/widecharwidth
> 
> @thefallentree Thats pretty awesome, could we refactor this to a general unicode version extractor? I have something similar written JS, but its just a bunch of scripts that still needs handwork to get the final tables extracted. I plan to add this with the next unicode PR for more versions, maybe I can use yours instead?

Let's do it! Better still we could make current python script parse and generate per unicode version of both JS and C headers  that way people can freely choose to use whatever version they want to support.

jerch:
> Let's do it! Better still we could make current python script parse and generate per unicode version of both JS and C headers that way people can freely choose to use whatever version they want to support.

Yes, ideally we manage to make it working for any or at least the newer versions to create the wcwidth tables, maybe for different language targets. This way it could act as a solid base for any wcwidth needs. Later on it could also deal with grapheme clusters and other important Unicode stuff (once we need / support those).

thefallentree:
@jerch  could you also add an option here to treat ambiguous chars as wide? For example Roman Numeral Ⅱ  is not rendered in full-width right now

and base on the EAW table, it is classified as ambiguous , almost all tty implementations including mintty, putty has this option to treat ambiguous as double width.

U+2161 | Ⅱ | ROMAN NUMERAL TWO 

image

jerch:
@thefallentree 
Ah yes good find. We already had reports regarding this, but it kinda never made it into the code for various reasons, mainly:
The treatment of ambiguous chars as wide is seen as an ancient feature that was commonly used with legacy CJK encodings, but is now somewhat discouraged with Unicode.

I think exposing such a setting would help in circumstances, where the app side's `wcwidth` implementation still uses the encoding dependent switch. So yes, we should have that, downside though is ~3KB bigger source files for the ambiguous definitions. I think we should not add it with this PR, as it would introduce many more chances across the code (also to the old tables). --> Tracking issue: #2668

thefallentree:
will this be merged in soon?

thefallentree:
I looked at this code, but I couldn't understand how it would solve the char-breaking issue i mentioned previously in  https://github.com/xtermjs/xterm.js/issues/2660

The issue here is that DOM renderer is breaking this 7 codepoint emoji into 4 <span>s  each contains an codepoint and and ZWJ,  This PR just changed wcwidth, I don't think it will change that

jerch:
@thefallentree Correct, this PR does not solve that issue. As I wrote above it is even worse - it is questionable if we ever can find a proper solution for that part of unicode rules for the terminal. While grapheme support itself is doable, it has to be in line with the what the other side thinks of the output representation. Currently none of the common terminal libs have proper support for that and would simply account those emojis as single consecutive chars. So this has to be fixed on both ends together to keep working.

The compound emojis have another problem (as already noted here https://github.com/xtermjs/xterm.js/issues/2664#issuecomment-572024333) - they have multiple valid output representations - depending on the renderer capabilities, which makes the runwidth undetermined for app side - a clear no-go. This needs some specification of a default behavior so appside has something to work with. Again thats a problem of the terminal interface in general being incompatible with what the apple guys specced in the newer unicode versions (they dont care much even for their own system, Terminal.app will also fail badly here).

TL;DR: Dont use those compound emojis in terminals, they will screw up output in any emulator atm.

thefallentree:
@jerch Clearly there is no cookie cutter solution here, I'm saying the only solution is to refactor xtermjs to provide an interface so people can set their version of desired unicode version to support.  (not just wcwdith, but also character joining rules) .

 That way , xtermjs can be tuned accordingly matching what the application desires. (For my websocket terminal application, I want to use unicode 12.1 and I've tuned everything with it except xtermjs)

jerch:
@thefallentree Yes, I think the unicode provider in this PR is a good starting point to enhance unicode rules later on in a more version aware fashion.

jerch:
@Tyriar I think the provider stuff is quite solid, not so the new table, it might be subject of several changes for many entries. There is no reasonalbe source of truth anymore, thus I mostly went with xterm compliance (v349).
Windows with conpty is another sad story, I was not able to get far with it yet. For some reason conpty keeps to return weird things for many emojis (does weird screen handling with wrong back erasing and such), I even started to wonder whether there is some issue in node-pty. Thus I started to dig into the ms terminal code, but cannot get it build either (from repo code). This def. needs some more attention, my tests showed very poor emoji (I think newer unicode chars in general) support.

Tyriar:
@jerch do we need a tracking issue for conpty emoji support?

plastikfan:
Hi, just for my 2 pence worth; I have built some PowerShell functionality that heavily uses emojis and seeing undesirable results in Fluent Terminal which depends on xtermjs as opossed to the correct alignment of emojis seen in Windows Terminal. Here is as example of the same code, running in Fluent Terminal vs Windows Terminal:

* Fluent Terminal:

fluent-terminal emojis-bad

* WIndows Terminal:

windows-terminal emojis-ok

The terminal in Visual Studio Code also uses xterm.js and the emoji display is similarly mis-aligned there too.

In these examples, you will notice that in Fluent Terminal, the values inside the tables are not correcly aligned within their columns, but in Windows Terminal they are. And this mis-alignment is not just left to right. Take a look at the green dot preceding 'Parameter Sets', you can see that it is slightly truncated at the top in FT but not in WT.

So I'll be looking forward to seeing this get fixed, because other than this slight issue, I love using Fluent Terminal which is a great console app.

Actually, I have a better example to illustrate this issue. Take a look at this output which comes from another command:

* Fluent Terminal:
fluent-terminal emojis-bad signals

* WIndows Terminal:
windows-terminal emojis-ok signals

jerch:
@plastikfan This is related to getting a proper hold of the used unicode version from windows and subsequently loading the right unicode addon into xterm.js. Currently we only have only support for v6 and v11, a more versatile approach to generate unicode version addons is needed (see #2668).
My guess is that those emojis are from Unicode v12+, if you want to get involved - we'd need the current Unicode version your windows (terminal) uses to address that.

plastikfan:
Thanks @jerch. To be honest, I wasnt really that aware of different versions of unicode, so your comment is enlightening to me. I just used emojis with impunity and as they were displayed ok in WT I thought nothing more of it. Its only when I started to test in FT that I began to become aware of problems with displaying emojis in different environments. Currently, I am not aware of version of Unicode supported in WT, but I can do some research and post again.

plastikfan:
Its proving to be quite difficult to get definitive info about unicode version support in WIndows Terminal. However, I did find this [pull request](https://github.com/microsoft/terminal/pull/8035) which does give us a hint. This one concerns Unicode 13, but I don't know if this is in production or not, I assume not as the pull request is still not closed, but is merged. I would assume but dont know how to confirm, that Unicode 12 in WT is actually in production.

plastikfan:
Hi @jerch, I can confirm (from that pull request previously referenced) that Windows Terminal currently supports Unicode 13.

jerch:
Thx for investigating. Will see if we can get a v13 addon rolling. This should fix most single emoji issues. Note that more complex things still wont work, as we currently dont support grapheme clusters (needed for compound emojis).

plastikfan:
Fyi, @DHowett responded with [this comment](https://github.com/microsoft/terminal/pull/8035#issuecomment-816817745), which will mean more to you than it does to me.

jerch:
> Fyi, @DHowett responded with this comment, which will mean more to you than it does to me.

What he points out is clear and in fact the reason, why we have to support multiple runtime changable versions (down to v6 for older systems accessed via ssh and such).

PR #2575, opened 2019-11-23
Prevent atlas missing exception when opened before attach

Tyriar:
See microsoft/vscode#85048

This should hopefully prevent that issue by preventing drawing and remembering to defer atlas creation if there are issues with dimensions on render.


Tyriar:
This didn't end up fixing the problem after bringing it to VS Code 🙁, hard to repro this case in the demo

PR #2590, opened 2019-11-30
Allow the thickness of the bar cursor to be configured

npezza93:
This setting will allow consumers to change the thickness of the bar cursor style and defaults to the current behavior. 

Before (cursorBarWidth = 1):
image
After (cursorBarWidth = 2):
image

### DIFF
--- a/addons/xterm-addon-webgl/src/renderLayer/BaseRenderLayer.ts
+++ b/addons/xterm-addon-webgl/src/renderLayer/BaseRenderLayer.ts
@@ -153,7 +153,7 @@ export abstract class BaseRenderLayer implements IRenderLayer {
    * @param x The column to fill.
    * @param y The row to fill.
    */
-  protected _fillLeftLineAtCell(x: number, y: number, width: number = 1): void {
+  protected _fillLeftLineAtCell(x: number, y: number, width: number): void {
     this._ctx.fillRect(
         x * this._scaledCellWidth,
         y * this._scaledCellHeight,
--- a/addons/xterm-addon-webgl/src/renderLayer/BaseRenderLayer.ts
+++ b/addons/xterm-addon-webgl/src/renderLayer/BaseRenderLayer.ts
@@ -153,11 +153,11 @@ export abstract class BaseRenderLayer implements IRenderLayer {
    * @param x The column to fill.
    * @param y The row to fill.
    */
-  protected _fillLeftLineAtCell(x: number, y: number): void {
+  protected _fillLeftLineAtCell(x: number, y: number, width: number = 1): void {
     this._ctx.fillRect(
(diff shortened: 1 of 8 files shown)


typings/xterm.d.ts
```diff
@@ -72,6 +72,11 @@ declare module 'xterm' {
      */
     cursorStyle?: 'block' | 'underline' | 'bar';
 
+    /**
+     * The width of the bar cursor.
+     */
+    cursorBarWidth?: number;
```
Tyriar: I guess we should align with VS Code here and just go with `cursorWidth`

typings/xterm.d.ts
```diff
@@ -72,6 +72,11 @@ declare module 'xterm' {
      */
     cursorStyle?: 'block' | 'underline' | 'bar';
 
+    /**
+     * The width of the bar cursor.
```
```suggestion
     * The width of the cursor in CSS pixels when `cursorStyle` is set to 'bar'.
```

addons/xterm-addon-webgl/src/renderLayer/BaseRenderLayer.ts
```diff
@@ -153,11 +153,11 @@ export abstract class BaseRenderLayer implements IRenderLayer {
    * @param x The column to fill.
    * @param y The row to fill.
    */
-  protected _fillLeftLineAtCell(x: number, y: number): void {
+  protected _fillLeftLineAtCell(x: number, y: number, width: number = 1): void {
```
Tyriar: Since this is only used by the cursor layer let's remove the default value

src/common/services/OptionsService.ts
```diff
@@ -20,6 +20,7 @@ export const DEFAULT_OPTIONS: ITerminalOptions = Object.freeze({
   rows: 24,
   cursorBlink: false,
   cursorStyle: 'block',
+  cursorBarWidth: 1,
```
Tyriar: We should sanitize the value and prevent < 1 here:

https://github.com/xtermjs/xterm.js/blob/ab66cb5d263012af54161f29d61fcc115ba520ca/src/common/services/OptionsService.ts#L114-L119

src/browser/renderer/BaseRenderLayer.ts
```diff
@@ -171,11 +171,11 @@ export abstract class BaseRenderLayer implements IRenderLayer {
    * @param x The column to fill.
    * @param y The row to fill.
    */
-  protected _fillLeftLineAtCell(x: number, y: number): void {
+  protected _fillLeftLineAtCell(x: number, y: number, width: number = 1): void {
```
Tyriar: Same, remove the default value

npezza93:
Should be good now! Thanks for the review @Tyriar!

PR #2618, opened 2019-12-07
Merge serialize addon into master

Tyriar:
We should merge after https://github.com/xtermjs/xterm.js/pull/2369 gets merged


JavaCS3:
Please merge #2687 to fix the pipeline failure

typings/xterm.d.ts
```diff
@@ -955,6 +955,13 @@ declare module 'xterm' {
      * @param y The line index to get.
      */
     getLine(y: number): IBufferLine | undefined;
+
+    /**
+     * Creates an empty cell object suitable as a cell reference in
+     * `line.getCell(x, cell)`. Use this to avoid costly recreation of
+     * cell objects when dealing with tons of cells.
+     */
+    getNullCell(): IBufferCell;
```
Tyriar: I'm not convinced we need to expose this, for the usage in `serialize`, can't we just store the first and second calls to `line.getCell`? The point of the optional param is so you can reuse `IBufferCell`s that have already been created.

typings/xterm.d.ts
```diff
@@ -965,6 +972,7 @@ declare module 'xterm' {
      * Whether the line is wrapped from the previous line.
      */
     readonly isWrapped: boolean;
+    readonly length: number;
```
Tyriar: Needs jsdoc

typings/xterm.d.ts
```diff
@@ -974,8 +982,9 @@ declare module 'xterm' {
      * behavior.
      *
      * @param x The character index to get.
+     * @param cell Optional cell object to load data into.
```
Tyriar: Could do with some elaboration as to why this is good.

typings/xterm.d.ts
```diff
@@ -992,19 +1001,30 @@ declare module 'xterm' {
...
-    /**
-     * The width of the character. Some examples:
-     *
-     * - This is `1` for most cells.
-     * - This is `2` for wide character like CJK glyphs.
-     * - This is `0` for cells immediately following cells with a width of `2`.
-     */
-    readonly width: number;
+    getWidth(): number;
+    getChars(): string;
```
Tyriar: Missing jsdoc

typings/xterm.d.ts
JavaCS3: @Tyriar "Null Object Pattern" makes the code looks simpler. Otherwise there will have a lot of null check. Currently, I cannot find any other way to make it simpler than using this technic  .

typings/xterm.d.ts
JavaCS3: @Tyriar This code is originally from @jerch 's sample. We could make it better to understand for sure.

typings/xterm.d.ts
Tyriar: Yeah I tried to change it and ended up reverting it as it was complicated

PR #2627, opened 2019-12-11
Expose texture atlas as API and use in demo (release/4.3)

### LINKED ISSUE #2623: WebGL: Canvas added to body
With the new WebGL renderer I'm seeing a spurious problem where (what I think is) the char atlas is sometimes being added into `<body>` when the terminal loses focus:
image
Also see https://github.com/JunoLab/atom-ink/pull/242#issuecomment-562863787.

It's entirely possible that we're doing something stupid when initializing the terminal/renderer, but I'm still opening this issue on the off chance that someone has seen this before.

## Details
- Browser and browser version: Electron 4.2.12
- OS version: Manjaro
- xterm.js version: 4.3.0

### Steps to reproduce
Not sure how to reproduce this, sadly.

### LINKED ISSUE #2634: Change cursor to hand/click for links
Small nit, but it would be super helpful if the cursor changed to hand/click when hovering over links or "link mode" is activated like here in VS Code.

Screen Shot 2019-12-17 at 10 45 04 AM

No sure if this is core in xterm or code?

Tyriar:
Fixes #2623

### DIFF
--- a/addons/xterm-addon-webgl/src/WebglAddon.ts
+++ b/addons/xterm-addon-webgl/src/WebglAddon.ts
@@ -10,6 +10,7 @@ import { IColorSet } from 'browser/Types';
 
 export class WebglAddon implements ITerminalAddon {
   private _terminal?: Terminal;
+  private _renderer?: WebglRenderer;
 
   constructor(
     private _preserveDrawingBuffer?: boolean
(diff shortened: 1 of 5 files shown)


Tyriar:
Actually we can fix this with just a webgl addon publish

PR #2632, opened 2019-12-15
Keep textarea in sync with cursor

### LINKED ISSUE #2598: Keep textarea in sync with cursor
We only sync the textarea when composition and right click events happen due to an old IE bug where the textarea gripper would always appear. We no longer care around IE support so we should be able to keep the textarea in sync with the cursor which would mean accessibility features like the magnifier app on Windows will work.

VS Code: https://github.com/microsoft/vscode/issues/86107

BuraChuhadar:
Fixes #2598

### DIFF
--- a/src/Terminal.ts
+++ b/src/Terminal.ts
@@ -1302,6 +1302,7 @@ export class Terminal extends Disposable implements ITerminal, IDisposable, IInp
       return thirdLevelKey;
     }
 
+    this._compositionHelper.moveTextAreaUnderMouseCursor();
     // Don't invoke for arrows, pageDown, home, backspace, etc. (on non-keypress events)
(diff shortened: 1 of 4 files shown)


src/Terminal.ts
```diff
@@ -1264,6 +1264,7 @@ export class Terminal extends Disposable implements ITerminal, IDisposable, IInp
       return thirdLevelKey;
     }
 
+    this._compositionHelper.moveTextAreaUnderMouseCursor();
```
Tyriar: It's unexpected for this function to have side effects since it's name starts with "is"

Tyriar:
Closing due to radio silence and pressing space scrolls the whole page down. Still open to another PR though.

PR #2637, opened 2019-12-19
Remove a large portion of InputHandler's dependency on Terminal

Tyriar:
Part of #1507 

Behavior changes:

- `parser.reset()` is now called when `Terminal.reset` is called (before it was only on `RIS`)
- Calling `Terminal.open(undefined|null)` (compile error in strict TS) will no longer work, previously if it was opened on an element that parent would be a fallback

EventEmitter now support 2 arguments, allowing to pass in 2 primitives (start and end for request refresh). This can be extended to support more in the future as required.

### DIFF
--- a/src/InputHandler.test.ts
+++ b/src/InputHandler.test.ts
@@ -7,7 +7,7 @@ import { assert, expect } from 'chai';
 import { InputHandler } from './InputHandler';
 import { MockInputHandlingTerminal, TestTerminal } from './TestUtils.test';
 import { Terminal } from './Terminal';
-import { IBufferLine } from 'common/Types';
+import { IBufferLine, IAttributeData } from 'common/Types';
 import { DEFAULT_ATTR_DATA } from 'common/buffer/BufferLine';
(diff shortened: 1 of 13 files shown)


Tyriar:
@jerch a quick once over might be a good idea if you have time, I think it's pretty solid and tests already caught a few problems during refactoring.

jerch:
@Tyriar Guess its to late now :smile: 

Few remarks: Calling parser reset during `Terminal.reset` is a problem, if there is still data pending in the write buffers (which gets not cleared by `Terminal.reset` due to a previous issue with that). Basically we run into band synchronous vs. "do it right away" (JS calling synchronous) issues here. Since its a rather uncommon edge case I dont think we have to fix that right away, but should keep that in mind. A proper solution would need to specify whether an "intrusive" action like `Terminal.reset` should always be band- or JS-call sychronous (and all applied to all subactions as well).

Tyriar:
@jerch 👉 https://github.com/xtermjs/xterm.js/pull/2640

Issue #2614, opened 2019-12-06
Look into removing cursor render layer from webgl renderer


Tyriar:
The cursor in the webgl renderer is drawn on a transparent 2d context canvas that sits on top of the webgl canvas, this was done purely to get an MVP out and speed up development. Now that true color is done we should be able to roll this into the single canvas fairly easily.

There's one for links as well but we should wait on that until underline is supported https://github.com/xtermjs/xterm.js/issues/2251

Tyriar:
I looked into this (https://github.com/xtermjs/xterm.js/pull/2641) and I'm not sure if it's a good idea to do this. While rendering would be guaranteed to be the same which would avoid problems like https://github.com/xtermjs/xterm.js/issues/2613 and some drawing code would be consolidated, it would mean every 600ms a full frame gets rendered which seems overkill when it's just a blinking cursor. Closing for now.

PR #2641, opened 2019-12-19
Experiment with moving webgl cursor drawing into rectangle/glyph renderers

Tyriar:
#2614

Opening as a reference so I can delete the upstream branch.


Tyriar:
See https://github.com/xtermjs/xterm.js/issues/2614#issuecomment-567627178

PR #2644, opened 2019-12-21
Fullwidth handling in buffer writes

### LINKED ISSUE #1779: Better handling of fullwidth chars
Currently non of the cell content insert/delete/replace methods handle previously added fullwidth chars at the position where the action takes place. All of those can break "fullwidth char cell pairs" by either cutting the left or the right part or even move the cells apart (insert).

Possible solution: Whenever a cell gets modified by any of the actions, we would have to check whether the cell is part of a fullwidth char and clear the other cell.

Affected by this:
- `InputHandler.eraseInLine`
- `InputHandler.eraseInDisplay`
- `InputHandler.deleteChars`
- `InputHandler.insertChars`
- `InputHandler.eraseChars`
- `InputHandler.print` (only after cursor jumps, normal input flow already handles this)

Might be easier to fix on buffer line level, since all above fall back to 3 methods - `insertCells`, `deleteCells` and `replaceCells`.

To avoid fixing this several times - blocked until we are settled with the buffer line implementation.

### LINKED ISSUE #2592: Chinese character displaying wrong in 'less' command
WX20191203-150755

jerch:
Adds fullwidth checks for buffer primitives (replaceCells, deleteCells, insertCells) and `InputHandler.print`. Fixes #1779, fixes #2592.

### DIFF
--- a/src/InputHandler.test.ts
+++ b/src/InputHandler.test.ts
@@ -1270,4 +1270,63 @@ describe('InputHandler', () => {
       [131072, 131072], [131072, 131072], [131072, 300000 - 131072 - 131072]
     ]);
   });
+  describe('should correctly reset cells taken by wide chars', () => {
+    let term: TestTerminal;
+    beforeEach(() => {
+      term = new TestTerminal({cols: 10, rows: 5, scrollback: 1});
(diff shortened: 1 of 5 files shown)


jerch:
> lgtm, love the tests

Ikr, the tests are abit cumbersome with all those '￥' chars, well I tried to cover all the weird insert here and there edge cases :smile_cat:

PR #2670, opened 2020-01-09
Fixes Alt + click behaviour when rows are wrapped

### LINKED ISSUE #2658: Take isWrapped cell status into account when alt+clicking to navigate the prompt in the normal buffer
VS Code: https://github.com/microsoft/vscode/issues/86482

kumaran-14:
Fixes #2658.

### DIFF
--- a/src/browser/input/MoveToCell.test.ts
+++ b/src/browser/input/MoveToCell.test.ts
@@ -23,14 +23,14 @@ describe('MoveToCell', () => {
       assert.equal(moveToCellSequence(4, 3, bufferService, false), '\x1b[C');
     });
     it('should ignore the Y value', () => {
-      assert.equal(moveToCellSequence(1, 1, bufferService, false), '\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D');
+      assert.equal(moveToCellSequence(1, 1, bufferService, false), '\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D\u001b[D');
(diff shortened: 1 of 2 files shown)


kumaran-14:
@Tyriar How to lint files in place? ```npm run lint``` only gives out err, but not fixes it.

kumaran-14:
@Tyriar Some MoveToCell tests are failing. Should they be updated or, my implementation is wrong?

Tyriar:
@kumaran-14 well this test was enforcing the old behavior which you've now broken:

```
 MoveToCell
       normal buffer
         should ignore the Y value
```

You'll need to update the test to respect the new behavior. Make sure there's a test for moving across a single line and multiple lines as well.

kumaran-14:
@Tyriar This looks fine?

kumaran-14:
@Tyriar This is fine?

kumaran-14:
@Tyriar can you look into this?

Tyriar:
@kumaran-14 I'll probably free up next week to power through xterm.js stuff and release 4.4

kumaran-14:
@Tyriar Sure. That sounds awesome. I'd like to finish this and contribute more.

Tyriar:
Another thing I noticed is that the current algorithm doesn't look at whether the row is wrapped, it should only move along the current wrapped row. I'm looking into fixing these issues now so we can get this merged in.

kumaran-14:
@Tyriar How do I get the positions of buffer end?

localhost_3000_

kumaran-14:
CI fails on windows and mac, passes on Linux.
What system did you test on?
I get this behaviour on Linux
ezgif com-video-to-gif

kumaran-14:
@Tyriar Sure. Glad to help.

PR #2697, opened 2020-02-06
v4.4.1: add UTF8 and URXVT encodings

olitvin:


jerch:
@olitvin Is there a specific reason why you re-added those encodings? Just curious, because we had a longer discussion about that between several terminal developers that kinda all agreed about removing those. Apps should not use any of these anymore, always go with SGR reports. Also avoid the X10 encoding, which is very limited in the coords and breaks UTF8 conformance.

PR #2710, opened 2020-02-10
Fix Link Provider end positions

### LINKED ISSUE #2708: Link Provider end positions are incorrect
The current `WebLinksProvider` incorrectly calculates the end position of the link. They are 1 character longer than it should be for both `hover()` and `leave()`. 

I also noticed a situation when implementing VS Code's link provider, that only the `leave()` is being called on that extra character. So there may be bugs in both the `LinkProvder` API and the `WebLinksProvider`, but I am not sure.

jmbockhorst:
Fixes #2708.

### DIFF
--- a/addons/xterm-addon-web-links/src/WebLinkProvider.ts
+++ b/addons/xterm-addon-web-links/src/WebLinkProvider.ts
@@ -49,7 +49,7 @@ export class LinkComputer {
         break;
       }
 
-      let endX = stringIndex + text.length + 1;
+      let endX = stringIndex + text.length;
       let endY = startLineIndex + 1;
 
       while (endX > terminal.cols) {
--- a/src/browser/Linkifier2.test.ts
+++ b/src/browser/Linkifier2.test.ts
@@ -1,0 +1,76 @@
+/**
+ * Copyright (c) 2020 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { assert } from 'chai';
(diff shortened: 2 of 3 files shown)


jmbockhorst:
I'm not sure how to simulate the mouse events needed to fire the onLinkHover/onLinkLeave events from the test file.

From my understanding as `LinkifierEvent` is used here:
https://github.com/xtermjs/xterm.js/blob/1bb2fc912ba1a1312d7e13f1506d4f1732a88f0d/src/browser/renderer/LinkRenderLayer.ts#L68

A link from `col: 5 - 7` (1 based) should create a link event where `x: 4 - 7` (0-based).

Tyriar:
Yeah that's a bit tricky... Well an easy, slightly hacky way to do it would be to:

- Make `_linkHover` protected
- Extend `Linkifier2` with a `TestLinkifier2`
- Expose `TestLinkifier2.linkHover` that calls into `_linkHover`

Then make a test that passes in some `any` objects, while not good because of the `any`, the bad type shouldn't matter so much as the tests will fail if when something goes wrong (or when something is refactored):

```ts
it('...', done => {
  linkifier.onLinkHover(e => {
    // assert on e based on the link used
    done();
  });
  linkifier.linkHover({ classList: { add: () => {} } } as any, { /*fill in link*/ }, {} as any);
});
```

jmbockhorst:
Thanks, that works.

PR #2712, opened 2020-02-10
Use playwright instead of puppeteer

### LINKED ISSUE #2700: Consider replacing puppeteer with playwright
Playwright supports Chromium, Firefox and Safari.

jmbockhorst:
Use playwright instead of puppeteer and adds support for `yarn test-api --browser [chromium|firefox|webkit]`. Closes #2700. 

Requires Node `>=10.15.0`.


jerch:
@jmbockhorst Very nice :+1: 

About the failing mouse tests - my guess is that my hack on the puppeteer protocol for the wheel events fail here. I had to do this, because puppeteer does not expose those with enough settings in the official API (like setting the modifier keys).

package.json
```diff
@@ -50,7 +51,7 @@
     "mustache": "^3.0.1",
     "node-pty": "^0.9.0",
     "nyc": "13",
-    "puppeteer": "^1.15.0",
+    "playwright": "^0.10.0",
```
Tyriar: We need to remote `@types/puppeteer`, are the types for playwright included in the package?

test/api/TestUtils.ts
```diff
@@ -3,10 +3,11 @@
  * @license MIT
  */
 
-import * as puppeteer from 'puppeteer';
+// import * as playwright from 'playwright';
```
Tyriar: Remove

test/api/TestUtils.ts
```diff
@@ -29,3 +30,18 @@ export async function writeSync(page: puppeteer.Page, data: string): Promise<voi
 export async function timeout(ms: number): Promise<void> {
   return new Promise<void>(r => setTimeout(r, ms));
 }
+
+export function getBrowserType(): playwright.BrowserType {
```
Tyriar: This could be one of the shared functions after https://github.com/xtermjs/xterm.js/pull/2711 gets merged, we can always create a `./bin/test-api.js` if that's needed too.

test/api/InputHandler.api.ts
```diff
@@ -260,8 +260,8 @@ describe('InputHandler Integration Tests', function(): void {
 
     describe('SM: Set Mode', () => {
       describe('CSI ? Pm h', () => {
-        it('Pm = 1003, Set Use All Motion (any event) Mouse Tracking', async() => {
-          const coords = await page.evaluate(`
+        it('Pm = 1003, Set Use All Motion (any event) Mouse Tracking', async () => {
+          const coords: any = await page.evaluate(`
```
Tyriar: `{ left: number, top: number, bottom: number, right: number }`

test/api/InputHandler.api.ts
```diff
@@ -443,7 +443,7 @@ async function getCursor(): Promise<{col: number, row: number}> {
 }
 
 async function getDimensions(): Promise<any> {
-  const dim = await page.evaluate(`term._core._renderService.dimensions`);
+  const dim: any = await page.evaluate(`term._core._renderService.dimensions`);
```
Tyriar: Can we reference `IRenderDimensions` from here?

addons/xterm-addon-attach/src/AttachAddon.api.ts
```diff
@@ -3,39 +3,39 @@
...
+    page = (await browser.defaultContext().pages())[0];
     await page.setViewport({ width, height });
   });
 
   after(async () => {
     await browser.close();
   });
 
-  beforeEach(async function(): Promise<any> {
+  beforeEach(async function (): Promise<any> {
```
Tyriar: I've noticed in some PRs the space goes away and some it gets added. Is this prettier acting up or something? I prefer: `async function() {` and `async () => {`

addons/xterm-addon-web-links/src/WebLinksAddon.api.ts
```diff
@@ -3,25 +3,25 @@
  * @license MIT
  */
 
-import * as puppeteer from 'puppeteer';
+import * as playwright from 'playwright';
```
Tyriar: It's probably best to change this to `import { Page, Browser }` on each file to be consistent

package.json
```diff
@@ -31,6 +31,7 @@
   },
   "devDependencies": {
     "@types/chai": "^3.4.34",
+    "@types/debug": "^4.1.5",
```
Tyriar: What's this for?

test/api/InputHandler.api.ts
```diff
@@ -271,7 +271,7 @@ describe('InputHandler Integration Tests', function(): void {
           await page.mouse.click((coords.left + coords.right) / 2, (coords.top + coords.bottom) / 2);
           await page.mouse.down();
           await page.mouse.move((coords.left + coords.right) / 2, (coords.top + coords.bottom) / 4);
-          assert.ok(await page.evaluate(`window.term.getSelection().length`) > 0, 'mouse events are off so there should be a selection');
+          assert.ok(await page.evaluate(`window.term.getSelection().length`) as number > 0, 'mouse events are off so there should be a selection');
```
Tyriar: I'm curious, was this erroring out for you? Do the types of `Page.evaluate` differ in playwright?

addons/xterm-addon-attach/src/AttachAddon.api.ts
jerch: Hehe, this is quite a sensible topic for some ppl. There are discussions about this in linter blueprint / style guide projects, some argue, that the whitespace should stay, since the official form is `function SP <name>(` and if name is empty the SP would still be there. Well JS allows both, I personally favour the short form w'o the space as well. (But also couldnt care less here, I would be more concerned if someone starts to remove all `;` in the code :imp: )

Tyriar: I normally don't mention it but in https://github.com/xtermjs/xterm.js/pull/2711 the oppposite change happened:

Screen Shot 2020-02-12 at 7 20 58 AM

jmbockhorst: I was going to ask about this. I’m just using VS Code’s default TypeScript formatting settings. Then I would see my formatter would change any entire file just with formatting so I would revert it and try to match the settings. There is some inconsistencies throughout the project so it becomes a mess sometimes and it’s easy to miss things. Maybe we can set some standardized workplace format settings or something?

package.json
jmbockhorst: Yes, the types for playwright are included.

test/api/InputHandler.api.ts
jmbockhorst: Yes, the return type of `Page.evaluate` in playwright is `Promise<unknown>` vs `Promise<Serializable>` in puppeteer.

addons/xterm-addon-web-links/src/WebLinksAddon.api.ts
jmbockhorst: Yeah, I can do that after #2711 once I can remove the `getBrowserTypes()` function from the addon files (which needs the namespace import).

package.json
jmbockhorst: Playwright needs it for some reason. See: https://github.com/microsoft/vscode/commit/8daf3f2270a126d90ab5c07b9c8c7763c00104f2.

addons/xterm-addon-attach/src/AttachAddon.api.ts
Tyriar: Sounds good to me https://github.com/xtermjs/xterm.js/issues/2717

package.json
Tyriar: Weird, hopefully it'll get fixed

test/api/InputHandler.api.ts
Tyriar: Ah nice, it's probably better to force the explicit cast 👍

jmbockhorst:
A new playwright version came out, so upping that might help stuff too.

Tyriar:
@jmbockhorst cool, just been working through all the failures on the other browsers 😃

Tyriar:
Tests take longer due to download again, I think we need to use `playwright-core` and download the browser on first run instead of `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` https://github.com/microsoft/playwright/issues/823#issuecomment-582134501

Tyriar:
Doesn't look like Windows tests works yet, maybe they'll work after https://github.com/xtermjs/xterm.js/issues/2714

Issue #2694, opened 2020-02-05
API to check if alternate screen is active


Eugeny:
Could we have an API to check whether xterm is currently showing the primary or the alternate screen?

Tyriar:
@Eugeny what do you need it for out of curiosity?

Eugeny:
I've got automatic "progress detection" in Terminus that scans for percentage-like strings in terminal output. However if a text editor like vi or nano is open (usually in alternate screen mode) and the text contains "12%" or similar, it's getting redrawn by the application again and again, triggering progress detection code again.

I'm looking to just disable it as long as alternate screen is active

Tyriar:
Perhaps we should also take this opportunity to namespace buffer like we did with parser, something like this?

```ts
interface buffer {
  normal: IBuffer
  alt: IBuffer
  active: IBuffer
  onDidChangeBuffer: IEvent<IBuffer>
}
```

mofux:
@Tyriar Yes that sounds good. So checking if the normal buffer is active would be an expression like `terminal.buffer.active === terminal.buffer.normal`? Or should we also introduce a convenience method for it either on the `buffer` or `IBuffer` interface?

Tyriar:
I guess having a `IBuffer.type: 'normal' | 'alt'` would be the best of both worlds?

jerch:
Just a minor headsup - we should take care with offering too much synchronous state introspection for a simple reason - ppl will end up trying to do things based on that but mostly ignore the current input buffer state, which might change things at any time again. (@Tyriar  Remember that issue about someone trying to write a local lisp interface? Very hard to spot de-sync issues will result from that.)

Or maybe we should be more clear about that in the API doc, which part is sync/async...

Tyriar:
@jerch I guess a note on `write` would help those problems. Trying to address that problem with docs would be better than not offering something like this.

Open to PRs, we might want to tweak the naming a bit but I think the basic shape of the API is there.

JavaCS3:
@Tyriar serialize-addon also need a way to read alternative buffer. Otherwise it's very hard to implement terminal session playback player even terminal session sharing tool that allow anyone can join at anytime(I found vscode's live share ext already supported terminal sharing, I don't know how it works instead of sending all the terminal output from the beginning)

jerch:
[...]

PR #2713, opened 2020-02-12
expose IBufferSet API

JavaCS3:
See #2694


JavaCS3:
@Tyriar But there're a lot of test cases using `buffer` API.
image
I would suggest adding a new `buffers` API and deprecate the old API, that will make the transition smoother

Tyriar:
@JavaCS3 it's primarily just internal things though, most of those would probably get fixed by doing a project-wide find and replaces, for example "term.buffer.x" -> "term.buffers.active.x"

JavaCS3:
@Tyriar updated

src/public/Terminal.ts
```diff
@@ -210,6 +215,22 @@ class BufferApiView implements IBufferApi {
   public getNullCell(): IBufferCellApi { return new CellData(); }
 }
 
+class BufferNamespaceApi implements IBufferNamespaceApi {
+  constructor(private _buffers: IBufferSet) { }
+
+  public get alternate(): IBufferApi { return new BufferApiView(this._buffers.alt, this._buffers); }
+  public get active(): IBufferApi { return new BufferApiView(this._buffers.active, this._buffers); }
+  public get normal(): IBufferApi { return new BufferApiView(this._buffers.normal, this._buffers); }
```
Tyriar: I think it would be best to only initialize 2 BufferApiViews in the constructor and change `active` based on when `onBufferChange` fires. Otherwise users who try this would get unexpected results:

```ts
const normalBuffer = term.buffer.normal;
term.buffer.onBufferChange(buffer => {
  if (buffer === normalBuffer) ...  // this is always false
}); 
```

We should be able to make a test which verifies this equality check works for normal and alternate buffers.

JavaCS3: @Tyriar  See the latest update

JavaCS3:
Updated

JavaCS3:
@Tyriar Updated, Looks like there're a lot of pending PRs. Do we have some schedules to clean them up?

Tyriar:
@JavaCS3 I've been a bit busy with other things in VS Code, this week has been catching up on this stuff and hopefully getting a release out.

typings/xterm.d.ts
```diff
@@ -1168,6 +1168,10 @@ declare module 'xterm' {
    * Represents a terminal buffer.
    */
   interface IBuffer {
+    /**
+     * The type of the buffer.
+     */
+    readonly type: 'normal' | 'alternate';
```
Tyriar, line 1174: Missing trailing new line

src/public/Terminal.ts
```diff
@@ -193,8 +193,19 @@ export class Terminal implements ITerminalApi {
...
 
+  init(buffer: IBuffer, buffers: IBufferSet): BufferApiView {
+    this._buffer = buffer;
+    this._buffers = buffers;
+    return this;
+  }
+
+  public get type(): 'normal' | 'alternate' {
+    if (this._buffers.normal === this._buffer) { return 'normal'; }
+    if (this._buffers.alt === this._buffer) { return 'alternate'; }
```
Tyriar: This is the only place it's used, we could just pass the type into the ctor.

src/public/Terminal.ts
```diff
@@ -210,6 +221,30 @@ class BufferApiView implements IBufferApi {
...
+    throw new Error('Active buffer is neither normal nor alternate');
+  }
+  public get normal(): IBufferApi { return this._normal.init(this._buffers.normal, this._buffers); }
+  public get alternate(): IBufferApi { return this._alternate.init(this._buffers.alt, this._buffers); }
+
+  public get onBufferChange(): IEvent<IBufferApi> {
+    return (listener: (buffer: IBufferApi, unused: void) => any): IDisposable => {
+      return this._buffers.onBufferActivate(() => listener(this.active));
+    };
+  }
```
Tyriar: It would be better to use a separate emitting on `BufferNamespaceApi`.

Issue #2661, opened 2020-01-02
Composition with tilde broken on Firefox on macOS with Finnish keyboard


ssh-peppe:
## Details
- Browser and browser version: Firefox 71.0
- OS version: macOS 10.15.2 - Finnish keyboard
- xterm.js version: 4.3.0

### Steps to reproduce

1. Type tilde + space or tilde + n
2. Observe that instead of getting ~ and ñ you get ~~ and ~ññ

### Description

For me on macOS, Fnnish keyboard, with Safari and Chrome a tilde + space results in ~ and tilde + n results in ñ as it should. But on Firefox a tilde + space results in ~~ and tilde + n results in ~ññ.

Interestingly this behavior difference can be seen (but a bit differently for some reason) on the https://xtermjs.org/ home page example terminal. If you enter tilde + space there on Safari you get nothing at all. Same for tilde + n. But with Firefox while tilde + space gets you nothing a tilde + n gets you ñ. Not sure why it's working differently there.

Debug logging shows the following when typing tilde + space on Firefox:

```
xterm.js: sending data "~" Array [ 126 ]
xterm.js: sending data "~" Array [ 126 ]
xterm.js: parsing data ~~
```

For tilde + n it shows:

```
xterm.js: sending data "~" Array [ 126 ]
xterm.js: sending data "ñ" Array [ 241 ]
xterm.js: sending data "ñ" Array [ 241 ]
xterm.js: parsing data ~
xterm.js: parsing data ññ
```

The "sending data" log lines reported by xterm.js also match what I'm seeing being sent to the onData() callback.

Note that there are differences in browsers and platforms with regards to what is reported when entering composed characters from e.g. compositionend vs. input vs. keydown vs. keypress events.

I should add that nothing is being sent until you hit space or n. That is, the initial tilde that starts the character compose does not result in any callbacks firing, whch is correct.

The sequence of keyboard events recording by attaching a custom keyboard handler using the Terminal API, for tilde + n is on Windows Firefox (71) with Finnish keyboard is the following:

```
keydown Control { target: textarea.xterm-helper-textarea, key: "Control", charCode: 0, keyCode: 17 }
keydown { target: textarea.xterm-helper-textarea, key: "AltGraph", charCode: 0, keyCode: 18 }
keydown { target: textarea.xterm-helper-textarea, key: "Dead", charCode: 0, keyCode: 160 }
[...]

Issue #2479, opened 2019-10-17
altgr+num prints wrong character on French azerty keyboard


Tyriar:
VS Code issue: https://github.com/microsoft/vscode/issues/82489

Using French (France) AZERTY keyboard

Expected:

- 2: é
- 7: è
- altgr+2: (nothing)
- altgr+7: (nothing)
- altgr+2, space: ~
- altgr+7, space: `

Actual:

- 2: é
- 7: è
- altgr+2: é
- altgr+7: è
- altgr+2, space: é~
- altgr+7, space: è`

This does not happen in monaco or other apps.

Tyriar:
Might be related to https://github.com/xtermjs/xterm.js/issues/2151

Tyriar:
All keyboard event handling code is in this file:

https://github.com/xtermjs/xterm.js/blob/29e53c9c1171e4e3fce7c7d172cd8f182b13aece/src/common/input/Keyboard.ts

willemavjc:
@Tyriar What kind of help are you looking for? (In case I would be capable of.)

Tyriar:
@willemavjc someone needs to investigate why the extra character gets printed (in Keyboard.ts) as since the monaco text editor doesn't we should be able to avoid the problem too. There may be something else on KeyboardEvent that we need to be looking at?

utybo:
I've been able to reproduce this on all dead keys that require using the AltGr key, including on the [BÉPO](https://bepo.fr) keymap, which is a French DVORAK keyboard. There is probably a missing `return` or `break` somewhere that would tell xterm.js to stop trying to identify the key if it's a dead key. It should be noted that dead keys that do not use the AltGr key still work perfectly fine (both for AZERTY and BÉPO).

I'm unfamiliar with the TypeScript language but the exact cause may be found by debugging the Keyboard.ts file

utybo:
Also, I just found out that typing Ctrl+Alt and then the key works fine, as AltGr is some sort of alias for Ctrl+Alt on Windows.

To recap:

| Key sequence | Expected | Actual |
| --- | --- | --- |
| Ctrl+Alt+2 then Space | ~ | ~ |
| AltGr + 2 then Space | ~ | é~ (é appears on AltGr + 2, ~ appears on Space) |
| Ctrl+Alt+7 then Space | \` | \` |
| AltGr + 7 then Space | \` | è\` (è appears on AltGr + 2, \` appears on Space) |

[...]

PR #2718, opened 2020-02-14
Support alt+enter key signal by default

### LINKED ISSUE #2498: Support "alt+enter" sending `\e\r`
Reminder issue coming from #2364.

kumaran-14:
Fixes #2498.


kumaran-14:
@jerch If this implementation is correct, we need add the `macOptionIsMeta` case too.

src/common/input/Keyboard.ts
```diff
@@ -103,7 +104,12 @@ export function evaluateKeyboardEvent(
       break;
     case 13:
       // return/enter
-      result.key = C0.CR;
+      if (altEnterMode && modifiers === 2) {
+        result.key = C0.ESC + C0.CR;
+      }
+      else {
+        result.key = C0.CR;
+      }
```
jerch: I have not tested xterm's behavior, but I think DECSET 1039 is meant to switch the leading ESC on/off for any key? Thus we would have to extend the flag to other keys as well.

Tyriar:
alt = option on macs, I guess to answer the question we'd be best to check other terminals that support this.

> extend the flag evaluation to other keys as well

This feature has always confused me a little, we could really do with a solid use case of where this is actually used so we can study the behavior of other terminals. https://github.com/xtermjs/xterm.js/pull/2364 mentions it's used in "some REPLs".

jerch:
@Tyriar @kumaran-14  Yes we should try to get a hold of the uses, maybe start with a simple script to check what kes are affected by that setting in xterm.

@NHDaly, @pfitzseb Do you have some insights for us, how `\e\ŗ` is used by REPLs? Also are there other keys involved in Alt+key usage? Note that we already support that for some other keys, but the support is kinda spread over `Keyboard.ts` and not really backed up by real world cases.

PerBothner:
See [note](https://github.com/xtermjs/xterm.js/issues/2498#issuecomment-587117721) about how Alt+Enter is used in [Fish shell](https://fishshell.com/docs/current/index.html#multiline-editing).

pfitzseb:
[Julia's](https://julialang.org/) REPL works the same way -- Alt-Enter inserts a newline regardless of whether a block is closed or not.

jerch:
@PerBothner, @pfitzseb Thx for the pointers. Yes xterm has by default this 8bit mode active for Alt+key and for Alt+Enter the fullscreen switch, at least in Ubuntu/Debian. Those key setup settings are somewhat complicated in xterm due to its additional 8bit key mode.

@Tyriar, @kumaran-14 
I just tested gnome-terminal, konsole and rxvt - they all support Alt+ for various keys but most dont expose any special DECSET mode for it. I wonder if we could just hardcode Alt+Enter as most do and forget about DECSET 1039/1036? xterm seems to be kinda special here, but since we are not interested in the 8bit key mode, we probably dont need to switch Alt+key mode on/off at all.

So my current suggestion is to go with a slimmer implementation:
- hardcode the Alt+Enter behavior in `Keyboard.ts`
- no new DECSET mode
- map option key if `macOptionIsMeta` as well (needs to be checked in other  mac terminals, still unclear whether it applies here)

Any other ideas?

kumaran-14:
@jerch I just saw the recent comment.
Can I go ahead with removing the DECSET mode and prepend `\e` for other keys?

jerch:
@kumaran-14 Yes I think we currently dont need the additional DECSET mode. It just complicates things. Sorry for the the forth and back, the needed bits is abit unclear and differs alot across emulators.

Tyriar:
> map option key if macOptionIsMeta is set (needs to be checked in other mac terminals, still unclear whether it applies here)

We should check how iTerm in particular handles this case.

kumaran-14:
@Tyriar @jerch  I tried to find out how few terminals behave.
`alacritty`
```
command executed : showkey -a

KeyPress        Ascii output

Enter            ^M      13 0015 0x0d
Ctrl + Enter     ^M      13 0015 0x0d
Shift + Enter    ^M      13 0015 0x0d
Windows + Enter   ^M     13 0015 0x0d
Alt + Enter      ^[^M    27 0033 0x1b
                         13 0015 0x0d
                         
command executed : from executable built from http://catb.org/esr/showkey/

KeyPress        output

Enter            <CTL-J=LF>
Ctrl + Enter     <CTL-J=LF>
Shift + Enter    <CTL-J=LF>
Windows + Enter  <CTL-J=LF>
Alt + Enter      <ESC><CTL-J=LF>


command executed : ctrl + v and keypress

KeyPress        output

Enter            ^M 
Ctrl + Enter     ^M 
Shift + Enter    ^M 
Windows + Enter   ^M
Alt + Enter      ^[^M

```

`iterm2`
```

command executed : from executable built from http://catb.org/esr/showkey/

KeyPress        output

Enter            <CTL-J=LF>
Ctrl + Enter     <CTL-J=LF>
Shift + Enter    <CTL-J=LF>
// Option key in Normal mode
Option + Enter      <CTL-J=LF>
// Option key in Meta mode
Option + Enter      <ALT-CTL-M=CR>
// Option key in Esc mode
Option + Enter      <ESC><CTL-J=LF>




command executed : ctrl + v and keypress

KeyPress        output

Enter            ^M 
Ctrl + Enter     ^M 
Shift + Enter    ^M 
// Option key in Normal mode
Option + Enter      ^M
// Option key in Meta mode
Option + Enter      <ffffffff>
// Option key in Esc mode
Option + Enter      ^[^M

```

jerch:
@kumaran-14 Thx for testing. Well, this is quite a naming mess. I think our `macOptionIsMeta` comes from Terminal.app, which indeed puts `ESC` in front of the pressed key char:
- `macOptionIsMeta` is unset: 'a' --> 3rd level shift char of key 'a' (layout dependent)
-  `macOptionIsMeta` is set: 'a' --> 'ESC a'

iTerm2 calls this "Esc+" mode, I have no clue what iTerm's "Meta" mode is about, so the name is abit misleading here. We are after "Esc+" in iTerm's terms.

Also this seems to be in line with emacs docs and their recommendations regarding keyboard setup (https://www.emacswiki.org/emacs/MetaKeyProblems#toc15).

But now the global `macOptionIsMeta` settings raises another question - how is a user supposed to access 3rd level shifts if this option is set? iTerm has an elegant solution for it - it separates the left and the right option key (btw this is pretty similar to PC keyboards, where Alt and AltGr are treated separately). My gut feeling tells me that several of our issues with other keyboard layouts on Macs could be solved, if we would allow this separation as well (in particular #2661, #2479 ff). This is nothing to be solved in this PR though, but should be investigated in one of those issues.

Basically we should follow here iTerm's behavior:
- `macOptionIsMeta` is set: output the same as option+Enter in "Esc+" mode
- `macOptionIsMeta`is unset: output the same as option+Enter in default mode

What is still unclear to me - is there any other key active on mac keyboards as Meta key, if `macOptionIsMeta`is unset?

@Tyriar Plz correct my assumptions/ideas above, the whole `macOptionIsMeta` stuff is pretty opaque to me.

Tyriar:
Sorry about super slow response.

@jerch making `macOptionIsMeta` act like it does in iTerm sounds good to me. So we don't need the mode we were talking about here, just alt+enter always add the esc?

kumaran-14:
@Tyriar if it is so, i think the changes in the pr now should be right.

src/common/services/CoreService.ts
```diff
@@ -9,6 +9,7 @@ import { IDecPrivateModes, ICharset } from 'common/Types';
 import { clone } from 'common/Clone';
 
 const DEFAULT_DEC_PRIVATE_MODES: IDecPrivateModes = Object.freeze({
+  altEscMode: true,
```
Tyriar, line 12: @kumaran-14 yep, except for this line

kumaran-14: @Tyriar updated.

Tyriar:
Good to merge when @jerch signs off.

Issue #1882, opened 2019-01-05
isWrapped does not get lifted for longish bash commands


jerch:
Repro with reflow PR in the demo:

- input some longish multiline command, e.g. `echo "Lorem ipsum....."`, press enter
- use arrow up to see the command again, dont press enter
- use arrow down to get back to clean prompt
- input `ls`, press enter
- shrink/enlarge viewport width

![grafik](https://user-images.githubusercontent.com/6193135/50718885-ad49a600-1094-11e9-9bd4-f28ddd38fde2.png)

Above the first 3 lines of the `ls` output wrongly reflow as they belong to one wrapped line.

Note this is not an issue of the reflow PR, the PR just makes the problem obvious. It seems we currently do not lift the `isWrapped` flag for `eraseChar` which happens to be send by libreadline when navigating through bash comands with arrow keys.

egmontkob:
Sounds similar to (perhaps the same as) former [VTE 780011](https://bugzilla.gnome.org/show_bug.cgi?id=780011) :-)

Tyriar:
Just tested, this still happens.

jerch:
@Tyriar Yeah, the buggy code is still there...

PR #2724, opened 2020-02-16
Reverse wraparound

### LINKED ISSUE #2716: Support reverse-wraparound mode?
I’m implementing a highly-restricted terminal with Xterm.js to allow users to edit their desired remote execution command before they connect to a true terminal. The editable portion of the line is only at the end; the rest represents the full command that one would use to perform the same action in a non-web terminal.

It’s a bit hackish to be interpreting keyboard events to edit a string, but it’s been working well enough. However, I’ve now found that if the editable part of the command wraps to another line, backspacing doesn’t wrap to the previous line. I found [this Super User answer](https://superuser.com/a/1474412) that suggests I could use “reverse-wraparound mode” in a true xterm, but sending that to the Xterm.js terminal didn’t change the behaviour.

a GIF of a terminal where a wrapped command is being backspaced but backspacing doesn’t wrap to the previous line

Reverse-wraparound mode [is mentioned in some comments](https://github.com/xtermjs/xterm.js/blob/7a8e2c67b0ea6c3db67773e0b21ab69a2d503f53/src/InputHandler.ts#L1634) in Xterm.js’s input handler, but it seems like it’s not handled.

Is there any chance that this mode will be supported? If not I can surely detect whether backspacing has reached the beginning of the line and move to the previous one if necessary, but I thought I’d ask in case there’s hope that this will be added or if there’s something I’m missing, because handling it manually seems like it’ll add more brittleness. If there’s any desire for it to be incorporated into Xterm.js, I could try making a PR!

Thanks for all the work on this, it works perfectly once the remote session has begun, it’s impressively powerful 💞

jerch:
There are a few things to consider, see @egmontkob's listing here: https://gitlab.gnome.org/GNOME/vte/issues/62

Basic implementation is done. ~I had to disable the `_restrictCursor()` call, which is abit unfortunate. This needs further investigation and is prolly related to our weird way of handling pending wraps.~ Turns out that xterm allows the last cell to be deleted if reverse wrap-around is set, otherwise not. Fixed.

 Furthermore this PR revealed another bug - the last char before EOL is not shown under the "sticky" cursor, prolly the renderer does not take the char into account under the cursor in that position (data is correct in buffer).

Another bug: we currently dont clear cells to the right when early wraparound happens. Fixed.

TODO:
- [x] create DECSET 45 (default = false)
- [x] decide whether to operate on all lines or only softwraps (latter seems preferrable)
- [x] skip empty cells at EOL after wide chars
- [x] write backspace tests
- [x] ~write reverse wrap-around xterm test case files?~ No comparison possible due to different behavior.
- [x] investigate `_restrictCursor()` behavior / scroll margins
- [x] clear cells to the right on early wraparound in print, write tests
- [x] followup issue for renderer bug of last cell --> #2729
- [x] followup issue to investigate correct lifting of `isWrapped` flag (related #1882) --> #2728

Closes #2716.


jerch:
Update: Eww, reverse wrap-around does some weird things on xterm - any cursor movement sequence does not respect grid borders anymore, instead the cursor wraps as well. Not sure why it was made like this, it leads to a totally different behavior for all of those sequences. DECAWM does not do this, from a "rule of least surprise" i'd call this behavior flawed. Ofc this is a bold statement since it seems to be an xterm feature originally. Still it changes the way the terminal content gets addressed, I think we should not follow that path.

Suggestion, up for discussion:
- limit reverse wrap-around to soft NLs (can only happen in a line, that was wrapped beforehand)
- always stop at hard NLs
- within a line BS should act as it would normally w'o reverse wrap-around set
- cursor movement sequences do not change anything:
  - they are bound to top/bottom/left/right and cannot pass these borders
  - a following BS would be eval'ed within the new position and apply the rules above
  - cursor jumps into an empty line "realize" that line up to the new cursor position, line itself is treated as not wrapped (cursor cannot pass into prev line)
- scrollborders should work/respected as with DECAWM

This is more in line how DECAWM works and makes reverse wrap-around kinda the opposite/undo action.

FYI: @Tyriar, @egmontkob

egmontkob:
> limit reverse wrap-around to soft NLs

This is what tmux does, and is probably the most reasonable behavior.

jerch:
@egmontkob Thx for your feedback. Yes I think so too, seems to be much more reasonable than altering the behavior of 50% of the common terminal sequences. Though I wonder why xterm did it this way in the first place (pros/cons compared to a stricter "DECAWM reverse") and whether there is some significant usage in the wild with that odd behavior. Not sure if it is worth to be brought up in terminal-wg...

FYI: @textshell

jerch:
@Tyriar Done with the implementation and tests. Up for review.

Tyriar:
> Btw, you can test the difference with the following:

@jerch I don't see any difference with this branch and master when doing this? 😕

jerch:
> @jerch I don't see any difference with this branch and master when doing this?

Oh thats weird. Maybe you are "holding it wrong" :smile_cat: Here are the exact steps to repro the difference:
- switch on reverse-wrap with `echo -e "\x1b[?45h"`
- open `cat` in demo
- hold down some character key until it fills more than one line (thus wrapped at least once)
- dont press <kbd>Enter</kbd>
- hold down <kbd><--</kbd>:
  - before: deleting chars would always stop at the beginning of the current cursor line
  - this PR: deleting wraps back into prev line up to last '\n' or top-left of the viewport

Tyriar:
@jerch oh I ran `cat README.md` 😅, I see it now

src/InputHandler.ts
```diff
@@ -715,12 +719,63 @@ export class InputHandler extends Disposable implements IInputHandler {
...
-    this._restrictCursor();
-    if (this._bufferService.buffer.x > 0) {
-      this._bufferService.buffer.x--;
+    const buffer = this._bufferService.buffer;
+
+    // by default (resverse wrap-around not set) we restrict the cursor to 0 .. rows/cols - 1
+    // thus the last cell of a row cannot be accessed by BS
+    if (!this._coreService.decPrivateModes.reverseWraparound) {
+      this._restrictCursor();
+    } else {
```
Tyriar, line 735: It might be better to return early so it's more clear how simple the !reverseWraparound case is:
```ts
// Reverse wraparound is disabled
if (!this._coreService.decPrivateModes.reverseWraparound) {
  this._restrictCursor();
  if (buffer.x > 0) {
      buffer.x--;
  }
  return;
}

// Reverse wraparound is enabled
...
```

src/InputHandler.ts
```diff
@@ -715,12 +719,63 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+      // copy of _restrictCursor(), but with x=cols as allowed max cursor position
+      // to make last cell accessible by BS
```
Tyriar, line 737: How about this with jsdoc for `maxCol`?

```ts
private _restrictCursor(maxCol?: number): void {
```

src/InputHandler.ts
```diff
@@ -715,12 +719,63 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+    }
+
+    if (buffer.x > 0) {
+      buffer.x--;
+    } else {
+      /**
+       * reverse wrap-around:
+       * Our implementation deviates from xterm on purpose. Details:
+       * - only previous soft NLs can be reversed (isWrapped=true)
+       * - only works within scrollborders (top/bottom, left/right not yet supported)
```
Tyriar, line 752: Not yet supported? Do we need to note this in vt features?

src/InputHandler.ts
```diff
@@ -715,12 +719,63 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+          && buffer.lines.get(buffer.y + buffer.ybase).isWrapped)
+      {
+        buffer.lines.get(buffer.y + buffer.ybase).isWrapped = false;
+        buffer.y--;
+        buffer.x = this._bufferService.cols - 1;
+        // find last taken cell - last cell can have 3 different states:
+        // - hasContent(true) + hasWidth(1): narrow char - we are done
+        // - hasWidth(0): second part of wide char - we are done
+        // - hasContent(false) + hasWidth(1): empty cell due to early wrapping wide char, go one cell further back
+        const line = buffer.lines.get(buffer.y + buffer.ybase);
```
Tyriar, line 769: Tiny thing, but I always found it easier to parse when you order it `ybase + y` (scroll to viewport, then apply cursor y

src/InputHandler.ts
```diff
@@ -715,12 +719,63 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+          && buffer.y <= buffer.scrollBottom
+          && buffer.lines.get(buffer.y + buffer.ybase).isWrapped)
+      {
+        buffer.lines.get(buffer.y + buffer.ybase).isWrapped = false;
+        buffer.y--;
+        buffer.x = this._bufferService.cols - 1;
+        // find last taken cell - last cell can have 3 different states:
+        // - hasContent(true) + hasWidth(1): narrow char - we are done
+        // - hasWidth(0): second part of wide char - we are done
+        // - hasContent(false) + hasWidth(1): empty cell due to early wrapping wide char, go one cell further back
```
Tyriar, line 768: I didn't verify but there are tests for these wide cases?

src/InputHandler.ts
jerch: I dont think so, we dont support left/right scroll margins at all atm, it is still on the TODO list of missing vt features. The comment is more a reminder to cover it here as well once we implement it (which is abit more involved, as it needs to be respected in several vt commands, nothing to deal with here).

src/InputHandler.ts
jerch: Fixed.

src/InputHandler.ts
jerch: Yep, all new tests with the Yen sign are related to the correct wide char handling.

src/common/InputHandler.test.ts
```diff
@@ -1375,6 +1386,127 @@ describe('InputHandler', () => {
...
+        assert.equal(bufferService.buffer.x, 2);
+        inputHandler.parse(ttyBS);
+        assert.deepEqual(getLines(bufferService, 2), ['    ', '  ']);
+        assert.equal(bufferService.buffer.x, 1);
+        inputHandler.parse(ttyBS);
+        assert.deepEqual(getLines(bufferService, 2), ['    ', '  ']);
+        assert.equal(bufferService.buffer.x, 0);
+      });
+      /*
+      it.only('should not reverse outside of scroll margins', () => {
```
jerch, line 1469: @Tyriar I cannot get this test working anymore without `TestTerminal`, seems the scrollback is not respected at all. Any clue why?

Tyriar: Yep, move it over here:

https://github.com/xtermjs/xterm.js/blob/45077d710627fb2aec61ac2dfc41879d8c32371e/src/browser/Terminal.test.ts#L1421-L1430

`Terminal` still owns `scroll`, `scroll` should probably live into `BufferService` and then `Terminal` reacts to it via an event. Created https://github.com/xtermjs/xterm.js/issues/2881 for this

Tyriar:
I'll give this another review and hopefully merge next week 😃

Issue #1145, opened 2017-12-12
Curly and colored underlines


egmontkob:
This is originally a [feature of Kitty](https://github.com/kovidgoyal/kitty/blob/master/protocol-extensions.asciidoc), now also adopted by VTE (GNOME Terminal and friends). Technically two separate features, but they mostly make sense together, e.g. for spell checking.

Apparently vim and neovim have already / are about to support these, see e.g. [vim undercurl](https://github.com/vim/vim/issues/1306), [vim color](https://github.com/vim/vim/pull/2405), [neovim](https://github.com/neovim/neovim/issues/7479).

---

The new `SGR 4:3` (`\e[4:3m`) attribute, strictly with a colon as separator, was introduced to start a curly underline.

In the mean time, `4:0`, `4:1` and `4:2` were also added as aliases for the standard `24` (turn off all kinds of underlining), `4` (single underline) and `21` (double underline), respectively.

At some point in the future, probably `4:4` and `4:5` could also stand for dotted and dashed underlines in some order (these are the five types of underlining supported by HTML/CSS).

---

The new `SGR 58` and `59` sequences specify the color of the underline, following the pattern of `38` and `39`. That is, `58;5;idx` for an entry of the 256-color palette, or `58;2;r;g;b` for direct RGB. There's no shortcut notation for the first 16 entries (corresponding to SGR 30-37 and 90-97), use the 256-color mode with indices of 0-15 instead.

`59` reverts to the default, that is, the underline's color auto-following the text color.

In case you're short of bits, I believe it's okay to drop some precision, e.g. store only 4 bits per color channel. We were also considering this in [the VTE bug](https://bugzilla.gnome.org/show_bug.cgi?id=721761).

Update: In VTE we ended up approximating the truecolor underline to 4+5+4 bits (R, G, B respectively). This means: 8+8+8+1 = 25 bits for foreground (in addition to truecolor, we have to be able to denote palette and default values too). Same for background. 4+5+4+1 = 14 bits for underline (again, an extra bit to denote palette and default values). That's a total of 64 bits for the colors, stored in an int64 (bold, italic etc. are additional bits are next to this in another variable). With multiplication instead of bitpacking, it could have been 5+5+5 bits for the underline. Of couse it's not a necessity to put all the color and no other attributes in the same integer, we just decided it's nice this way (and results in some performance improvement).

egmontkob:
[...]

Issue #1134, opened 2017-11-28
Support hyperlink ansi escapes


sindresorhus:
Terminal emulators are [starting to support hyperlinks](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda). While many terminals have long detected URL's and linkified them, allowing you to Command-Click or Control-Click them to open a browser, you were forced to print the long unsightly URL's on the screen. As of spring 2017 [a few terminals](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) began supporting HTML like links, where the link text and destination could be specified separately.

Example from iTerm2 3.1:
<img width="1004" alt="screen shot 2017-11-28 at 12 08 20" src="https://user-images.githubusercontent.com/170270/33316814-8b35c2d4-d435-11e7-8524-da4fa076e6fb.png">

mofux:
Good idea, PR welcome!

chabou:
It could become a standard and make parsing easier.
Related issue: https://github.com/xtermjs/xterm.js/issues/583

sindresorhus:
The biggest benefit is that we can linkify text without having to spam the terminal with a long URL. For example, one could link to the documentation in error messages without making it super verbose.

albinekb:
This should be implemented with care, you need to be able to hover the link to see where it leads, otherwise you could link users unknowingly to malicious sites. This is already a huge problem in email phishing.

sindresorhus:
iTerm2 shows the URL like this:

<img width="582" alt="screen shot 2017-11-28 at 14 50 20" src="https://user-images.githubusercontent.com/170270/33323037-84a0dee8-d44b-11e7-8314-346ef039b890.png">

jamestalmage:
If / when this is implemented, please submit an issue or PR to [`supports-hyperlinks`](https://github.com/jamestalmage/supports-hyperlinks), which is intended to detect support for this capability.

Tyriar:
@jamestalmage this will probably need to be left up to xterm.js embedders, right now the environment is owned by the terminal emulators that use xterm.js. For example Hyper and VS Code each individually set `TERM_PROGRAM`.

jamestalmage:
@Tyriar - In that case, perhaps just include a reference to `supports-hyperlinks` in the commit that lands this, and in the release notes when it goes wide. I'm not sure if there are `supports-hyperlinks` equivalents for other languages, but if so, probably worth mentioning them as well.

[...]

PR #2751, opened 2020-03-08
Underline styles and colors - core part

jerch:
PR to support additional underline styles and colors in `AttributeData`. Implements core parts of #1145.

Note that this PR extends `AttributeData` in a generic way, which can be use to store other data later on (like #1134).

Changes to the internal interface:
```TS
interface IAttributeData {
  ...
  // whether the attr object has any extended attrs defined
  hasExtendedAttrs(): number;
  // update internal HAS_EXTENDED flag
  updateExtended(): void;
  // transparently either returns FG color or underline color if set
  getUnderlineColor(): number;
  // color mode accessors
  getUnderlineColorMode(): number;
  isUnderlineColorRGB(): boolean;
  isUnderlineColorPalette(): boolean;
  isUnderlineColorDefault(): boolean;
  // actual underline
  getUnderlineStyle(): UnderlineStyle;
}
```

The color and the color mode accessors transparently map to the FG variants, if no independent color was set or underline is unset. Thus its important to test beforehand for `isUnderline`.

TODO:
- [x] handle extended SGR 4 sequences
- [x] handle SGR 58 and SGR 59 sequences
- [x] extend `AttributeData` to expose extended attributes
- [x] extend `BufferLine` to hold and export extended attributes
- [x] test cases for all


src/common/Types.d.ts
```diff
@@ -98,6 +106,16 @@ export interface IAttributeData {
...
+
+  // extended attrs
+  hasExtendedAttrs(): number;
+  updateExtended(): void;
+  getUnderlineColor(): number;
+  getUnderlineColorMode(): number;
+  isUnderlineColorRGB(): boolean;
+  isUnderlineColorPalette(): boolean;
+  isUnderlineColorDefault(): boolean;
+  getUnderlineStyle(): number;
```
Tyriar: Shouldn't these be encapsulated within `IExtendedAttrs`?

src/common/Types.d.ts
```diff
@@ -68,10 +68,18 @@ export interface ICharset {
...
+  underlineColor: number;
+  clone(): IExtendedAttrs;
+  isEmpty(): boolean;
+}
+
 /** Attribute data */
 export interface IAttributeData {
   fg: number;
   bg: number;
+  extended: IExtendedAttrs;
```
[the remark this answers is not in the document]
jerch: I had it as optional before, which lead to toxic deopts in the input chain, guess due to shape shifting of the attr object (tried with both `undefined` and `null`). So now this is a 1:1 object relation. Do you issues in that?

Offtopic: Btw I found the reduced timeline resolution issue by playing with this.

src/common/Types.d.ts
jerch: They refer to data on `AttributeData`, which `IExtendedAttrs` is not aware of. (Currently `IExtendedAttrs` extends `AttributeData`, not the way around).

src/InputHandler.ts
```diff
@@ -2026,6 +2026,21 @@ export class InputHandler extends Disposable implements IInputHandler {
     }
   }
 
+  /**
+   * Helper to write color information packed with color mode.
+   */
+  private _updateColor(color: number, mode: number, c1: number, c2: number, c3: number): number {
```
Tyriar, line 2032: Little more specific?

```suggestion
  private _updateAttrColor(color: number, mode: number, c1: number, c2: number, c3: number): number {
```

src/InputHandler.ts
```diff
@@ -2115,13 +2152,13 @@ export class InputHandler extends Disposable implements IInputHandler {
...
    * | 3         | Italic.                                                  | #Y      |
-   * | 4         | Underlined. (no support for newer underline styles)      | #Y      |
+   * | 4         | Underlined (see below for style support).                | #Y      |
    * | 5         | Slowly blinking.                                         | #N      |
    * | 6         | Rapidly blinking.                                        | #N      |
    * | 7         | Inverse. Flips foreground and background color.          | #Y      |
    * | 8         | Invisible (hidden).                                      | #Y      |
    * | 9         | Crossed-out characters.                                  | #N      |
-   * | 21        | Doubly  underlined.                                      | #N      |
+   * | 21        | Doubly  underlined.                                      | #P[Currently outputs a single underline.] |
```
Tyriar, line 2161: 😱 

```suggestion
   * | 21        | Doubly underlined.                                       | #P[Currently outputs a single underline.] |
```

src/common/Types.d.ts
Tyriar, line 118: Just to clarify, all underline styles will still result in the `isUnderline` public API being non-zero?

src/common/buffer/CellData.ts
```diff
@@ -19,10 +19,11 @@ export class CellData extends AttributeData implements ICellData {
...
   }
   /** Primitives from terminal buffer. */
-  public content: number = 0;
-  public fg: number = 0;
-  public bg: number = 0;
-  public combinedData: string = '';
+  public content = 0;
+  public fg = 0;
+  public bg = 0;
+  public extended: IExtendedAttrs = new ExtendedAttrs();
```
Tyriar, line 25: So the approximate memory increase per cell will be 1 pointer for this, and then the standard `ExtendedAttrs` object will get shared between all cells?

jerch:
> Looks good after the minor issues are resolved +1, next PR would be adding rendering?

Yepp. But thats non trivial, as you already might have guessed from the comments in #1145. Gonna fix the issues above in the next couple of days, but prolly wont make it to the renderer parts :disappointed:

Tyriar:
@jerch no worries, we should still merge to avoid conflicts piling up

src/common/Types.d.ts
jerch: The idea here for renderer side is to first test for `isUnderline`, if that evals to `true`, you'd have to request the style and the color to finally draw it. This way it is compatible to our current code, which only relies on a `isUnderline` test and always draws a single line.

The color methods cannot be used to determine whether underline is in place, as they would always report a color (fg color). The style method can be used for that, yes.

src/common/buffer/CellData.ts
jerch: But only for cells, that actually have extended attribs set. All other (thus most cells) dont use more memory in the buffer. To get this working I used another bit in `bg` as indicator whether a cell holds extended attrs (`HAS_EXTENDED` flag).

PR #2754, opened 2020-03-10
vtfeatures v2

jerch:
Jekyll is really picky and does not like the leading empty line before the YAML block.

@Tyriar: Not being able to devel and test this locally before adding to the repo is really annoying. Even worse here, since it spans two repos.

### DIFF
--- a/bin/extract_vtfeatures.js
+++ b/bin/extract_vtfeatures.js
@@ -32,8 +32,7 @@ const TYPES = [
   'SOS'
 ];
 
-const MARKDOWN_TMPL = `
----
+const MARKDOWN_TMPL = `---
 title: Supported Terminal Sequences
 category: API
 ---


Tyriar:
> Not being able to devel and test this locally before adding to the repo is really annoying. Even worse here, since it spans two repos.

Would it be better to move this to xtermjs.org so it's a similar model to https://github.com/xtermjs/xtermjs.org/blob/master/bin/update-docs? The problem then is you could make a change in xterm.js@master that causes a breakage when we run this during release?

Ideally we'd have a pipeline to handle this and all updates to the website on release for us. Something you we could as a quick fix would be to move this over to xtermjs.org and [set up a new pipeline job](https://github.com/xtermjs/xterm.js/blob/eaadb75d3904dd55057ffad8a370d38974d4352b/azure-pipelines.yml#L7) that clones it and verifies it runs against master without erroring out?

jerch:
Tracked in #2755.

PR #2761, opened 2020-03-16
Add CoCalc to "Real-world uses" in README

slel:


slel:
Added at the end now.

PR #2767, opened 2020-03-22
fix: addDisposableDomListener dispose() leaking

JavaCS3:


JavaCS3:
@Tyriar I'm wondering if xterm.js can migrate from mocha to jest. Because jest provided almost all in one test tool(chai's like assert/expect, jsdom, mock...), which is quite convenient! 

See how easy it is to test this function with jest
https://github.com/JavaCS3/xterm-player/blob/0c847ee1ae3218f36b692a0a206cbe3da39e2e5b/src/ui/DomHelper.test.ts#L34-L52
```typescript
test(title(addDisposableDomListener, 'dispose'), () => {
  const div = document.createElement('div')
  const mock = jest.fn()
  const disposable = addDisposableDomListener(div, 'test', mock)

  div.dispatchEvent(new Event('test'))
  expect(mock).toBeCalledTimes(1)

  disposable.dispose()

  div.dispatchEvent(new Event('test'))
  expect(mock).toBeCalledTimes(1)

  // double dispose
  disposable.dispose()

  div.dispatchEvent(new Event('test'))
  expect(mock).toBeCalledTimes(1)
})```

Tyriar:
@JavaCS3 that seems like a lot of effort for little payoff, I'd also like to align with VS Code which remains on mocha for now.

PR #2800, opened 2020-03-30
Upgrade mac agent in pipelines

Tyriar:
Looks like capacity has run out for old macOS, this should get builds
green again.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -39,7 +39,7 @@ jobs:
 
 - job: macOS
   pool:
-    vmImage: 'xcode9-macos10.13'
+    vmImage: 'macOS-10.15'
   steps:
   - task: NodeTool@0
     inputs:


JavaCS3:
Looks like there's some trouble in pipeline. Is it related to azure running out of capacity? Maybe Github action would help

Tyriar:
@JavaCS3 yes I think it's related to capacity, I initially tried out actions for xterm.js and they weren't as reliable as actually this sort of thing was happening (jobs not starting) when there was definitely no capacity issue. May consider switching it up again in the future but for now pipelines does the job.

JavaCS3:
@Tyriar You could refer to my repo's Github action as an example
https://github.com/JavaCS3/xterm-player/blob/master/.github/workflows/ci.yml
https://github.com/JavaCS3/xterm-player/blob/master/.github/workflows/release.yml

PR #2830, opened 2020-04-11
Add api-test script to log poll result on passing --debug flag

### LINKED ISSUE #2702: Add debug option to API tests
I find it very useful when debugging API tests to log the result of polling:

https://github.com/xtermjs/xterm.js/blob/56990f2a2904f1d8b5f868be9ce14895b2156c8f/test/api/TestUtils.ts#L13

Would be handle to run `yarn test-api --debug` to log this instead of changing it locally.

kumaran-14:
Fixes #2702 

Example usage : 
```
> yarn test-api ./out-test/api/CharWidth.api.js --debug
yarn run v1.21.1
$ node ./bin/test_api.js ./out-test/api/CharWidth.api.js --debug
(node:23379) Warning: "--debug" is not available in Node.js v13.6.0; use "--inspect" instead.
Debugger listening on ws://127.0.0.1:9229/a66d1a72-150a-4ef1-9185-974d4680fe42
For help, see: https://nodejs.org/en/docs/inspector


  CharWidth Integration Tests
    getStringCellWidth
pollFor result:  25
      ✓ ASCII chars
pollFor result:  30
pollFor result:  10
      ✓ combining chars
pollFor result:  30
pollFor result:  28
      ✓ surrogate chars
pollFor result:  30
pollFor result:  12
      ✓ surrogate combining chars
pollFor result:  30
pollFor result:  21
      ✓ fullwidth chars (49ms)
pollFor result:  30
pollFor result:  22
      ✓ fullwidth chars offset 1


  6 passing (640ms)

Done in 0.99s.
```


bin/test_api.js
```diff
@@ -0,0 +1,49 @@
...
+if (process.argv.length > 2) {
+  const args = process.argv.slice(2);
+  flagArgs = args.filter(e => e.startsWith('--'));
+  // ability to inject particular test files via
+  // yarn test [testFileA testFileB ...]
+  files = args.filter(e => !e.startsWith('--'));
+  if (files.length) {
+    testFiles = files;
+  }
+}
```
Tyriar, line 31: Did this end up working too?

kumaran-14: Yes. The example usage at the top illustrates this.

kumaran-14:
@Tyriar I've been working with my fork which is not updated.
I cannot pull from upstream because I do not have access.
Only then I can check in my local.

Tyriar:
@kumaran-14 try using this button:

image

It's fine if you do this wrong and then fix up the merge in another commit on your branch.

Alternatively with the command line:

```sh
git remote add upstream https://github.com/xtermjs/xterm.js
git fetch upstream
git merge upstream/master
# Then:
# 1. Fix commit
# 2. git add .
# 3. git commit
```

kumaran-14:
@Tyriar I made some changes to the argument parser script.

These are the features I wanted to implement
1. allow `--debug` flag to log poll result.
2. `--browser <browser_name>` to be parsed normally as expected
3. Any other other flags of type `--<flag_name> <flag_val>` to be parsed normally 
for any other puporse
3. allow `<filename>` args to be passed without `--` prefix to run test on particular files easily

As the flags start with `--` prefix and flag values and test filenames do not have the said prefix, I could not differentiatle them.

So I added this format `--<flag>=<flag_value>` for flags with value and kept the `<filename>` format for testfiles.
Eg: `--browser=chromium`

A complete example command.
`yarn test-api --debug ./out-test/api/CharWidth.api.js` 

I felt this would be easier to type and remember. We could change this according to our wish.

The build is breaking because `flatMap` function I used is an ES2019 feature. Older versions of chrome/firefox/webkit may not support it.

kumaran-14:
There is a workaround for `flatMap`. We can use the workaround if we go ahead with this way of doing things.

Tyriar:
It's not the browsers we need to worry about but node.js. I need to use node 10 for VS Code development so we should support at least that which doesn't include flatMap, we should use a different method because of this.

kumaran-14:
@Tyriar I have updated it.

PR #2847, opened 2020-04-15
Merge ligatures addon into core repo

### LINKED ISSUE #2726: Merge ligatures addon into core repo
The thinking changes from having separate addon repos to just including them all in one to simplify development and releasing. The ligatures addon also keeps having regressions, we should pull https://github.com/xtermjs/xterm-addon-ligatures into `addons/xterm-addon-ligatures`.

It should be relatively easy to pull the code over by following existing patterns established by the other addons.

LabhanshAgrawal:
Fixes #2726 
@Tyriar see if this seems ok (first time contributing to xterm)


Tyriar:
Looks pretty good, tests even run and pass 😃 

I'll give it a proper review as soon as I can

LabhanshAgrawal:
I was trying to add a webpack config similar to other addons, but getting errors for `fs` module which is used by `font-finder`
Is this addon currently supposed to run only on electron/node?
If yes, would it be ok to use target node or electron-renderer in webpack config?

Tyriar:
Yes it can only be used in node/electron as it pulls fonts from disk.

LabhanshAgrawal:
added webpack config similar to other addons

LabhanshAgrawal:
@Tyriar the ci on windows is failing because of `../../node_modules/.bin/tsc` type of commands. cmd is not able to use this path, other addons have similar commands though. Can you suggest any workaround for this.
Also, webpack is including the `font-finder` and `font-ligatures` dependencies in the bundle. Should I add them as `external`?

Tyriar:
@LabhanshAgrawal that's odd, pretty sure the same thing is run for the other addons so I don't know what the problem could be. Unless the addons don't actually get built on other platform in which case we should fix that generally before merging. If that is the case could you create an issue?

LabhanshAgrawal:
@Tyriar While doing `npm install` at root, the other addons are skipped as they don't have any dependencies. Since the ligature addon depends on few things, it's build step is triggered and that's failing on windows because of the `../../.....tsc` thing.
Same thing might happen with the webpack, but as the Release step in the ci runs on ubuntu, it works.
All the addons get built in the `tsc -b ./tsconfig.all.json` step though.
Directly doing `tsc -p` was working (feab7c2d18e9b2c80c7a5de79b5a0bdcd3882938), I changed it to be same as other addons.
Also, this issue is not there if you use powershell instead of cmd on windows.

LabhanshAgrawal:
@Tyriar please review.

LabhanshAgrawal:
rebase and pushed

Tyriar:
This week I'm focusing on testing VS Code, next week I'll be merging in xterm.js PRs and releasing 4.6. I expect this should be a pretty easy merge then 🙂

Tyriar:
I notice the addon doesn't yet depend on the core codebase so breakages could still happen. We should write integration test(s) so breakages are detected early https://github.com/xtermjs/xterm.js/issues/2896

Tyriar:
Had an idea which could maybe do away with a lot of the complexity and native access in this addon https://github.com/xtermjs/xterm.js/issues/2897

LabhanshAgrawal:
Will you be moving the issues from the old repo? if that's possible.

Tyriar:
Done thanks, and archived the old one.

regisbsb:
https://github.com/vercel/hyper/issues/3607

PR #2867, opened 2020-04-25
Different behavior when search two same words in only one row

### LINKED ISSUE #2840: Different behavior when search two same words in only one row
## Details
- Browser and browser version: Chrome 81.0.4044.92
- OS version: Mac OS 10.13.6 
- xterm.js version: 4.4.0

### Steps to reproduce
I find `findNext` in `SearchAddon` behaves differently when search two same words in only one row.

1. As the picture below, when I am searching the word `bash` and press the `Enter` button, the highlight mark will start from the first(1) position to the next(2) position, then stop at the position 2 and never go back to position 1.

image

2. But when I am searching in multi rows(pressing the `Enter` button), the highlight mark will go 1->2->3->4->5->1->2... as a circle.

image

Although users seldom search in only one row but the different behavior makes it a little bit strange.

rishab-pdx:
Fixes #2840. We must wrap back search to 0th column if we have a ```currentSelection```. @Tyriar let me know if any further changes are required.


rishab-pdx:
@Tyriar do you mean if we hit the beginning(column 0) with find previous we must wrap again to the end of the column or last search selection?

rishab-pdx:
Let me know if any other changes are required.

Issue #2749, opened 2020-03-04
Create a node.js target for xterm.js


JavaCS3:
## Details
- OS version: node
- xterm.js version: 4.4.0

```
> require('xterm')
Thrown:
ReferenceError: window is not defined
    at Object.<anonymous> (/private/tmp/test/node_modules/xterm/lib/xterm.js:1:224)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
```

### Steps to reproduce
https://runkit.com/embed/q7w97d0b8wj2

Looks like webpack4 bug,
See: https://medium.com/@JakeXiao/window-is-undefined-in-umd-library-output-for-webpack4-858af1b881df

Tyriar:
xterm.js won't work in node because it relies on window and the DOM, it needs a browser.

JavaCS3:
@Tyriar Though xterm.js doesn't support node env, it should be working at importing level.
I wish xterm.js will support headless mode some day.

Tyriar:
@JavaCS3 if https://github.com/JavaCS3/xterm.js/commit/54ec873c037d87a535afab3e710f58dbda4fc95d is all that's needed to unblock you for this then we can take that, there may be other runtime problems though? I'm also not totally sure how a headless node-only version is useful. It looks like xterm-player is a browser app?

JavaCS3:
@Tyriar It's not that urgent for https://github.com/JavaCS3/xterm.js/commit/54ec873c037d87a535afab3e710f58dbda4fc95d
I can wait until xterm.js core parts get separated or workaround in someway.

xterm-player is a browser app.

The idea of running xterm.js in headless mode is, I want to create a terminal based live sharing stuff that allow anyone to join that sharing session at anytime. Just like a video sharing, but it's a terminal based version. I want xterm.js can generate some key frames on the backend server. If anyone joins this session, the server will just send that keyframe and the following term sequence to him. That's why I want a headless mode.

the21st:
[...]

PR #2876, opened 2020-04-26
Move more parts of Terminal into CoreTerminal and move everything else into browser/

### LINKED ISSUE #1507: Improved Layering and enabling TS strict

Tyriar:
Fixes #1507 
Part of #2749 

To move Terminal and Terminal.tests in there are whole lot of `!` being added. This is still a problem and is being deferred to https://github.com/xtermjs/xterm.js/issues/2877. At least from now on we will be using strict mode for all new code!


src/Terminal.ts
```diff
@@ -1334,18 +1237,13 @@ export class Terminal extends CoreTerminal implements ITerminal {
...
-    this._bufferService.reset();
-    this._charsetService.reset();
-    this._coreService.reset();
-    this._coreMouseService.reset();
+    super.reset();
     this._selectionService?.reset();
 
     // reattach
     this._customKeyEventHandler = customKeyEventHandler;
-    this._userScrolling = userScrolling;
```
Tyriar, line 1348: @jerch I changed this to no longer retain user scrolling on RIS. It seems like this is correct behavior since the buffers are reset as well, just want to make sure. (originally added in https://github.com/xtermjs/xterm.js/commit/b748eca980c8fe30af10a995cc99938e70bf0d73)

jerch: Technically RIS should behave like a total new terminal instance, as a hardware device would do a reboot with flushing everything and even resetting the line (pretty much like switching it off/on).

We kinda never treated RIS that "cold boot" way, not sure why. The issue #2127 was about loosing data from the write buffer, but after reading the spec again I solved it the wrong way. It clearly should lose write buffer data, as well as any other terminal state and apply initial settings.

So yes - I'm fine with treating RIS more like it was intented, but I am not sure if this will break on other ends.

Tyriar: @jerch for userScrolling in particular, that essentially tracks whether the user has scrolled up (and therefore lock ydisp). If we clear the normal buffer, there is no scrollback so my thinking was it could cause bugs.

Issue #2818, opened 2020-04-06
DOM renderer selection isn't re-drawn after resizing the terminal


Tyriar:
VS Code issue: https://github.com/microsoft/vscode/issues/90903

![image](https://user-images.githubusercontent.com/2193314/78584490-203c5200-77ed-11ea-94dd-154728166c33.png)

rishab-pdx:
@Tyriar I would like to take this. Can you give a pointer as to where I should start tackling this? I understand that the sections of the canvas must be redrawn.

Tyriar:
Sounds good!

onResize in WebGL: https://github.com/xtermjs/xterm.js/blob/master/addons/xterm-addon-webgl/src/WebglRenderer.ts#L133
onResize in DOM: https://github.com/xtermjs/xterm.js/blob/cdcbb1d6daa2b972de9ccb510e67f80ec8904319/src/browser/renderer/dom/DomRenderer.ts#L259-L262

This is how the dom selection gets cleared, we should look at related parts and clear during resize in a similar way as the webgl renderer:

https://github.com/xtermjs/xterm.js/blob/cdcbb1d6daa2b972de9ccb510e67f80ec8904319/src/browser/renderer/dom/DomRenderer.ts#L276-L280

rishab-pdx:
@Tyriar My findings so far:
onRezise with WebGL rendered selection elements are not reset. However, selection model is reset. 

I made changes so that selection elements are reset on resize and drawn back as in the case of canvas rendererType.
As this is a vscode related issue is there any document I can follow to test by integrating these changes with vscode?
Let me know if any other changes are required.

Tyriar:
@rishab-pdx I think it's an xterm.js issue, if needed for testing you can access `term` in the devtools console in the xterm demo.

If you want to test inside VS Code you would set up your environment so you can launch "Code - OSS" and then do this:

```sh
cd node_modules
rm -rf xterm
git clone https://github.com/rishab-pdx/xterm.js xterm
cd xterm
yarn
yarn package # call this whenever you make changes
```

Not very smooth as I rarely do this, you can use symlinks too to make it nicer.

tabaddor:
I can take a further look at this issue, as it is directly related to microsoft/vscode#90903.

Tyriar:
[...]

PR #2889, opened 2020-04-30
DOM renderer selection isn't re-drawn after resizing the terminal

rishab-pdx:
#2818 
Made behavior of terminal for selection elements consistent across all ```rendererTypes```  wrt ```onResize``` event.


Tyriar:
Looking at this behavior I'm thinking maybe we should try keep the selection by clamping the start/end indexes to be within the viewport and then redrawing the screen.

It would be cool is the terminal worked just like the editor:

recording (3)

There's nothing wrong with that terminal selection, it just needs to be redrawn to cover the right portion of the terminal. For the clamp example imagine this:

image

Resized down should look like this:

image

Or maybe with the last cell of the last row selected 🤔, we probably want to keep the `\n` at the very least.

rishab-pdx:
Makes sense with keeping up the last ```\n``` if last cell of the last row is selected.

tabaddor:
> There's nothing wrong with that terminal selection, it just needs to be redrawn to cover the right portion of the terminal. 

All renderer types need to be re-drawn to cover the right portion when resized, or just the DOM type, or just keep the new line character for the DOM type?

Tyriar:
Closing for now as it's a bit stale, additional changes welcome though.

PR #2892, opened 2020-05-02
fixed ScreenDprListener never being fired

### LINKED ISSUE #2137: Display density changes lead to corrupted characters
When switching between a built-in Retina display and an external normal-density display, some characters will be drawn with incorrect size by the canvas renderer.

Now, I can't confirm whether it's possibly Electron's fault, but I suspect that Chrome would behave the same.

xterm 3.13.2

image

Eugeny:
Fixes #2137

### DIFF
--- a/src/browser/ScreenDprMonitor.ts
+++ b/src/browser/ScreenDprMonitor.ts
@@ -49,9 +49,7 @@ export class ScreenDprMonitor extends Disposable {
     }
 
     // Clear listeners for old DPR
-    if (this._resolutionMediaMatchList) {
-      this._resolutionMediaMatchList.removeListener(this._outerListener);
-    }
+    this._resolutionMediaMatchList?.removeListener(this._outerListener);
 
     // Add listeners for new DPR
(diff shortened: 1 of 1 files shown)


src/browser/ScreenDprMonitor.ts
```diff
@@ -44,12 +44,14 @@ export class ScreenDprMonitor extends Disposable {
...
-    if (!this._resolutionMediaMatchList || !this._outerListener) {
+    if (!this._outerListener) {
       return;
     }
 
     // Clear listeners for old DPR
-    this._resolutionMediaMatchList.removeListener(this._outerListener);
+    if (this._resolutionMediaMatchList) {
+      this._resolutionMediaMatchList.removeListener(this._outerListener);
+    }
```
Tyriar, line 54:
```suggestion
    this._resolutionMediaMatchList?.removeListener(this._outerListener);
```

Eugeny:
Added the change.

Tyriar:
👍 thanks!

Issue #2848, opened 2020-04-15
Link provider probably needs to change to provide links for entire line


Tyriar:
Consider the following:

![image](https://user-images.githubusercontent.com/2193314/79397999-66538d00-7f34-11ea-82bb-e75d19c329ca.png)

Currently in VS Code, depending on which cell you focus first gives a different link.

Good (validated local link provider):

![image](https://user-images.githubusercontent.com/2193314/79398041-81be9800-7f34-11ea-86b4-3071da64b0a9.png)

Bad (word link provider):

![image](https://user-images.githubusercontent.com/2193314/79398027-7a978a00-7f34-11ea-8026-229fde62feb0.png)

fyi @jmbockhorst

jmbockhorst:
So it would look something like this?
``` typescript
provideLink(bufferLineNumber: number, callback: (link: ILink | undefined) => void): void;
```

Tyriar:
Yes but multiple returned:

```ts
provideLinks(bufferLineNumber: number, callback: (links: ILink[] | undefined) => void): void;
```

Also when I was looking into https://github.com/xtermjs/xterm.js/issues/2849 I noticed that right now in vscode we're using `leave` like a `dispose`, which is good as xterm drops the link reference when leaving the cell range. But if we did this with the `provideLinks` idea then we would want to differentiate `leave?(event: MouseEvent, text: string): void;` and `dispose?(): void` so xterm owns the decision of when to get rid of it (leave = it may get reused, dispose = embedder should drop the ref).

Tyriar:
I don't really see another way of fixing this problem of competing providers creating different links based on what is hovered, which leads to an inconsistent experience.

Tyriar:
This isn't actually fixed yet, we need to delete links that are "below" higher priority links still.

PR #2904, opened 2020-05-04
hyperlinks addon

jerch:
Early WIP.

### DIFF
--- a/addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
+++ b/addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
@@ -5,8 +5,7 @@
  * UnicodeVersionProvider for V11.
  */
 
-import { Terminal, ITerminalAddon, IDisposable, IBufferRange } from 'xterm';
-import { ILinkProvider, IBufferCellPosition, ILink } from 'xterm';
+import { Terminal, ITerminalAddon, IDisposable, IBufferRange, ILinkProvider, IBufferCellPosition, ILink } from 'xterm';
 
 /**
  * TODO:
@@ -71,7 +70,7 @@ class HyperlinkProvider implements ILinkProvider {
     }
 
     // fix ranges to 1-based, right inclusive
-    for (let r of ranges) {
+    for (const r of ranges) {
       r.start.x++;
       r.start.y++;
       r.end.y++;
--- a/addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
+++ b/addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
@@ -1,0 +1,191 @@
+/**
+ * Copyright (c) 2019 The xterm.js authors. All rights reserved.
+ * @license MIT
+ *
+ * UnicodeVersionProvider for V11.
+ */
+
+import { Terminal, ITerminalAddon, IDisposable, IBufferRange } from 'xterm';
+import { ILinkProvider, IBufferCellPosition, ILink } from 'xterm';
+
+/**
+ * TODO:
+ * Need the following changes in xterm.js:
+ * - allow LinkProvider callback to contain multiple ranges (currently hacked)
+ * - make extended attributes contributable by outer code
+ * - extend public API by extended attributes
+ */
+
+class HyperlinkProvider implements ILinkProvider {
+
+  constructor(
+    private readonly _terminal: Terminal,
+    private readonly _urlMap: {[key: number]: string}
+  ) {}
+
+  public provideLink(position: IBufferCellPosition, callback: (link: ILink | undefined) => void): void {
+    const pos = {x: position.x - 1, y: position.y - 1};
+
+    // test whether we are above a cell with an urlId
+    // TODO: need API extension
+    const urlId = (this._terminal.buffer.active.getLine(pos.y)?.getCell(pos.x) as any).extended.urlId;
+    if (!urlId) {
+      callback(undefined);
+      return;
+    }
+
+    // we got an url cell, thus we need to fetch whole viewport and
+    // mark all cells with that particular urlId
+    const yOffset = this._terminal.buffer.active.viewportY;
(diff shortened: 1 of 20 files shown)


addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
```diff
@@ -0,0 +1,302 @@
+/**
+ * Copyright (c) 2019 The xterm.js authors. All rights reserved.
+ * @license MIT
+ *
+ * UnicodeVersionProvider for V11.
```
Tyriar, line 5: Remove

addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
```diff
@@ -0,0 +1,302 @@
...
+ */
+
+import { Terminal, ITerminalAddon, IDisposable, IBufferRange, ILinkProvider, IBufferCellPosition, ILink } from 'xterm';
+
+/**
+ * TODO:
+ * Need the following changes in xterm.js:
+ * - allow LinkProvider callback to contain multiple ranges (currently sloppy hacked into Linkifier2)
```
Tyriar, line 13: I'll be working on https://github.com/xtermjs/xterm.js/issues/2848 soon which will change the API to return all links for the current line.

addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
```diff
@@ -0,0 +1,302 @@
...
+ * - make extended attributes contributable by outer code
```
Tyriar, line 14: How about an `ICellMarker` API instead? Or tweaking the current one to optionally track a cell in a wrapped line?

addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
```diff
@@ -0,0 +1,302 @@
...
+
+  private _limitCache(): void {
+    if (this._internalId - this._lowestId > this.cacheLimit * 2) {
+      this._lowestId = this._internalId - this.cacheLimit;
+      [...this._urlMap.keys()]
+        .filter(key => key < this._lowestId)
+        .forEach(key => this._urlMap.delete(key));
+      [...this._idMap.entries()]
+        .filter(([unused, value]) => value < this._lowestId)
+        .forEach(([key, unused]) => this._idMap.delete(key));
```
Tyriar, line 145: You probably know but `[...arr].filter()` is probably a little slow to put into a parser hook?

Would `LRUMap` be useful here? We could pull into the addon?

addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    // TODO: make this better customizable from outside
+    callback({
+      ranges,
+      text: this._urlMap.get(urlId)!.url,
+      decorations: {pointerCursor: true, underline: true},
+      activate: this._urlMap.get(urlId)!.schemeHandler.opener,
+      hover: (event: MouseEvent, text: string) => {
+        console.log('tooltip to show:', text);
+      }
```
Tyriar, line 299: `activate`, `hover` and `leave` need to be settable when creating the addon.

addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
jerch: Victim of c&p :smile:

addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
jerch: Well for the purpose of this addon all it needs are multiple buffer ranges for an URL to still recognize `http://example.com` here as one link:
```
╔═ file1 ════╗
║          ╔═ file2 ═══╗
║http://exa║Lorem ipsum║
║le.com    ║ dolor sit ║
║          ║amet, conse║
╚══════════║ctetur adip║
           ╚═══════════╝
```

addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
jerch: This needs some more design work, yes I think this can be turned into something like cell markers (that stick to a cell, thus would move with it).

addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
jerch: Yeah, well it gets only executed once in a while (with default limits every 1000 urls). For some reason this declarative way was the only one I could find with proper type inference. Every loop attempt (which tend to be faster in general) was countered with type issues by TS lol.

LRUMap:
I think a linked list is not needed here, since the identifiers are just auto increasing integers (always to be cut from left side). Gonna rework that to a loop with early exit condition.

addons/xterm-addon-hyperlinks/src/HyperlinksAddon.ts
jerch: As talked about earlier, this prolly will end up with the LinkProvider being exposed on addon API level for individual customization.

jerch:
Playing around with this addon and a custom `xtermjs:` scheme:

recorded

The `openBox` function replaces buffer parts with the box. A `closeBox` function triggered on **x** restores the old state. This would also work for lines in the scrollbuffer. Hurray, terminal getting into browser realms now. Who said a `LinkProvider` should only handle http urls? :smile_cat:

Tyriar:
That gif 🤯

Tyriar:
@jerch link providers have been working great in vscode, 👍 for removing experimental from it

Tyriar:
https://github.com/xtermjs/xterm.js/issues/3493

jerch:
This is far off, thus closing it.

Issue #2826, opened 2020-04-10
Add an option to enable experimental APIs (and throw if it's off)


Tyriar:
Right now we have our experimental APIs in the same dts as the regular APIs and there is no guard to prevent people from using them. I'm proposing we add an option `enableProposedApi` (aligns with vscode's naming) that would throw an exception when any proposed API is used if that is not explicitly set to true. So anyone trying to use these APIs will get an exception unless they want to opt into proposed APIs which should involve reading the warnings and/or understanding the implications of doing so.

While technically not a breaking change as it would affect experimental APIs only, it would probably cause a lot of issues and it's worth a version bump. I'd feel a lot better about breaking experimental APIs if we had something like this.

PR #2921, opened 2020-05-08
Add a proposed API check option

Tyriar:
Part of #2826

### DIFF
--- a/src/browser/public/Terminal.ts
+++ b/src/browser/public/Terminal.ts
@@ -24,6 +24,12 @@ export class Terminal implements ITerminalApi {
     this._addonManager = new AddonManager();
   }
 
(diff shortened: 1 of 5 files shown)


Tyriar:
We could have both, there's an issue for that too. I really like a clear opt in this has where the apis just won't work until you flip the switch.

Tyriar:
Also the still reachable under experimental thing puts extra burden on us, needing to keep 2 implementations for a bit. Less burden is what I'm going for 🙂

jerch:
@Tyriar Yepp, the extra `experimental` stuff does not help much I guess, also if we'd go only with that, embedders have a hard time to fix the shifted symbols later on. So lets keep it simple.

Issue #2737, opened 2020-02-25
non-transparent selection covers text


UziTech:
Setting selection color to non-transparent color (e.g. `#0000ff`) covers text in canvas but not when using webgl addon.

## Details
- Browser and browser version: Chrome 80.0.3987.122
- OS version: Windows 10
- xterm.js version: 4.4.0

### Steps to reproduce

1. Go To https://codepen.io/UziTech/pen/gOpmPGo
2. Select text in each terminal

![image](https://user-images.githubusercontent.com/97994/75286256-d455c080-57dd-11ea-95c3-ebedab87ed83.png)

Tyriar:
This is kind of by design. For the normal renderers this was always a limitation as there was no efficient way of doing it, monaco has a similar issue where text background is not supported at all for performance reasons. So for historic reasons the selection color is meant to be transparent.

The WebGL renderer however was finally able to do it efficiently so we can draw the selection behind the text (and even change selection text color for a11y reasons if we want) so opaque colors work best.

The problem is though then there's an inconsistency in how the color is presented. My thinking to solve this was to blend the color with the background manually (0.3 opacity or something?) when an opaque color is provided such that the text is still visible. This is a problem in VS Code that I haven't gotten around to solving as well; using webgl and switching off it often necessitates switching my theme override color which currently looks like this:

```
  "workbench.colorCustomizations": {
    "[Default Light+]": {
      "terminal.selectionBackground": "#add6ff"
    },
    "[Visual Studio Light]": {
      "terminal.selectionBackground": "#add6ff"
    },
    "[Sapphire]": {
      "terminal.selectionBackground": "#153958"
    },
    "[Topaz (Dim)]": {
      "terminal.selectionBackground": "#42400f"
    }
  }
```

TL;DR: Let's change the canvas and dom renderers to blend the color with the background color at 0.3 opacity if the selection color is opaque.

Puneethnaik:
@Tyriar I would like to take this issue. Upon first inspection, I figured we need to make changes in the setTheme function of ColorManager.ts
https://github.com/xtermjs/xterm.js/blob/869d791f444f77024d97af0e2fe430996259a937/src/browser/ColorManager.ts#L113
And also, I am thinking there should be some changes in the color.blend function as well. 
https://github.com/xtermjs/xterm.js/blob/869d791f444f77024d97af0e2fe430996259a937/src/browser/Color.ts#L29
Please guide me further.

Tyriar:
[...]

PR #2952, opened 2020-05-31
non-transparent selection covers text

Puneethnaik:
#2737 issue fix. @Tyriar Please look into this. Would love to hear feedback.


Puneethnaik:
@Tyriar Please consider the changes. I observed some checks had failed.(Particularly the integration tests). I have committed few changes that passes all the checks. I moved logic from ColorManager setTheme function to a function in color namespace in Color.ts. Also, I invoke the function in DomRenderer and Renderer class constructor. Would love to hear feedback.

Tyriar:
@Puneethnaik I'll check this out next week, after a quick look at the code it looks pretty much like what I had in mind 👍

Puneethnaik:
@Tyriar thanks for considering the PR. Since this is my first PR in xterm.js, I am very excited. Will wait for the review and any feedback :+1:. Meanwhile I will keep studying the codebase, learnt a lot.

Tyriar:
@Puneethnaik good to hear 🙂, there's always more [help wanted issues](https://github.com/xtermjs/xterm.js/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) if you're keen.

src/browser/Color.ts
```diff
@@ -47,6 +47,17 @@ export namespace color {
...
+    const [selectionR, selectionG, selectionB, selectionA] = rgba.toChannels(colors.selection.rgba);
+    // The selection color is opaque. It needs to be blended with background color at 0.3 opacity Issue #2737
+    if (selectionA === 0xFF) {
+      const newAlpha: number = Math.round(opacity*255);
+      colors.selection = {
+        css: channels.toCss(selectionR, selectionG, selectionB, newAlpha),
+        rgba: channels.toRgba(selectionR, selectionG, selectionB, newAlpha)
+      };
+    }
+  }
```
Tyriar, line 60: This file is meant to just be a set of util functions around `IColor`, I'd prefer to keep the `IColorSet` concept out of here because of that.

Looking at the code, since `selectionOpaque` already exists for the WebGL renderer and `IColorSet.selection` is only used in the DOM and Canvas renderers, how about we move this logic into `ColorManager.setTheme` where the `IColorSet` is defined?

https://github.com/xtermjs/xterm.js/blob/8a3366b1eb5d10e86b3e7a49b9003ae49ff93a44/src/browser/ColorManager.ts#L118-L119

It's probably also worth renaming it to `selectionTransparent` to be more clear of its purpose. When you move it over, take some around `selectionOpaque` as it currently relies on `selection`.

src/browser/Color.ts
```diff
@@ -47,6 +47,17 @@ export namespace color {
     const rgba = channels.toRgba(r, g, b);
     return { css, rgba };
   }
+  export function blendSelectionWithBgWithOpacity(colors: IColorSet, opacity: number=0.3): void {
+    const [selectionR, selectionG, selectionB, selectionA] = rgba.toChannels(colors.selection.rgba);
+    // The selection color is opaque. It needs to be blended with background color at 0.3 opacity Issue #2737
+    if (selectionA === 0xFF) {
+      const newAlpha: number = Math.round(opacity*255);
```
Tyriar, line 54: We tend to use hex when working with color channels:

```suggestion
      const newAlpha: number = Math.round(opacity * oxFF);
```

src/browser/Color.ts
```diff
@@ -47,6 +47,17 @@ export namespace color {
     const rgba = channels.toRgba(r, g, b);
     return { css, rgba };
   }
+  export function blendSelectionWithBgWithOpacity(colors: IColorSet, opacity: number=0.3): void {
+    const [selectionR, selectionG, selectionB, selectionA] = rgba.toChannels(colors.selection.rgba);
+    // The selection color is opaque. It needs to be blended with background color at 0.3 opacity Issue #2737
+    if (selectionA === 0xFF) {
```
Tyriar, line 53: We could skip putting the channels until in the if statement by creating an `isOpaque` helper in the `color` namespace of `Color.ts` that does something like this `return color.rgba & 0xFF === 0xFF;`

src/browser/Color.ts
```diff
@@ -47,6 +47,17 @@ export namespace color {
...
+      const newAlpha: number = Math.round(opacity*255);
+      colors.selection = {
+        css: channels.toCss(selectionR, selectionG, selectionB, newAlpha),
+        rgba: channels.toRgba(selectionR, selectionG, selectionB, newAlpha)
+      };
```
Tyriar, line 58: We have a `colors.opaque` helper, we could move this into Color.ts as `colors.opacity(color: IColor, opacity: number): IColor`

Puneethnaik: I have incorporated the opacity as a parameter in the color.opacity util function in the color namespace. I have provided 1 as the default value of opacity. The function declaration is like this : export function opaque(color: IColor, opacity: number = 1): IColor. But, it is failing some tests when yarn test is run. 

>       + expected - actual
>        {
>           -  "css": "#000000ff"
>           +  "css": "#000000"
>               "rgba": 255
>        }
This is coming because i have made a change as in now the color.opaque function returns an alpha component in css as well
`
    return {
      css: channels.toCss(r, g, b, a),
      rgba: rgbaColor
    };
`
Please guide me. Maybe I am missing something. Or should the tests be modified? Since this might result in regression, I would need guidance from you on the same.

Tyriar: > export function opaque(color: IColor, opacity: number = 1): IColor

Let's make a separate function `opacity` with its own set of tests as making a color opaque is a bit easier than making it transparent using a 0-1 number. That particular test was failing because `opaque` removes the alpha channel all together from the "css" representation (`#000000` is the same as `#000000FF`).

Puneethnaik:
@Tyriar Thank you for the feedback. I fixed the code while consulting the feedback and pushed a commit. However, the checks are not passing. Upon some investigation, it seems a 404 Not found error occured, When I looked at the Azure devops [log](https://dev.azure.com/xtermjs/xterm.js/_build/results?buildId=3538&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=73c4289c-110b-5085-4a98-4009fa5cc9c8). I checked under linux tests section. There the message contained `url:
      'https://github.com/tonsky/FiraCode/raw/master/distr/otf/FiraCode-Regular.otf',
`
Upon following the link, it gave a 404. I checked the recent commit of FiraCode on the master branch(link [here](https://github.com/tonsky/FiraCode/commit/63c6434ab7786a4db18a31284e8f04cda87a0f87) it seems that FiraCode-Regular.otf is removed in this commit.
Please guide me on this.

Tyriar:
The font CI problem should get fixed with https://github.com/xtermjs/xterm.js/pull/2971

Puneethnaik:
Thank you @Tyriar. It is really an honor to start my open source track record by contributing to a great product like xterm.js. I will keep looking at more issues and try to contribute more.

PR #2953, opened 2020-05-31
Using putout linter on xterm codebase

coderaiser:
Just read about [switching eslint rules to warnings](https://github.com/xtermjs/xterm.js/pull/2895/files), and want to suggest usage results of [putout](https://github.com/coderaiser/putout) linter that fixes everything it can find.

Here is list of things in code of `xterm.js` that can be improved:
- [convert-for-each-to-for-of](https://github.com/coderaiser/putout/tree/master/packages/plugin-convert-for-each-to-for-of) (*enabled*)
- [convert-for-to-for-of](https://github.com/coderaiser/putout/tree/master/packages/plugin-convert-for-to-for-of)
- [remove-console](https://github.com/coderaiser/putout/tree/master/packages/plugin-remove-console)
- [simplify-ternary](https://github.com/coderaiser/putout/tree/master/packages/plugin-simplify-ternary) (*enabled*)
- [remove-empty](https://github.com/coderaiser/putout/tree/master/packages/plugin-remove-empty)
- [apply-shorthand-properties](https://github.com/coderaiser/putout/tree/master/packages/plugin-apply-shorthand-properties) (*enabled*) 
- [extract-sequence-expressions](https://github.com/coderaiser/putout/tree/master/packages/plugin-extract-sequence-expressions) (*enabled*)
- [convert-math-pow](https://github.com/coderaiser/putout/tree/master/packages/plugin-convert-math-pow)
- [convert-template-to-strings](https://github.com/coderaiser/putout/tree/master/packages/plugin-convert-template-to-string)

Here is used config`.putout.json`:

```json
{
    "rules": {
        "remove-console": "off",
        "apply-destructuring": "off",
        "merge-if-statements": "off",
        "convert-math-pow": "off",
        "remove-empty": "off"
    }
}
```

I'm willing to run `putout` again disabling rules you choose If it's appropriate , I think this can make codebase of `xterm.js` a little bit better :).

About an error:

```sh
1) CoreMouseService
       triggerMouseEvent
         eventCodes with modifiers (DEFAULT encoding):

      AssertionError: expected [ Array(8) ] to deeply equal [ Array(4) ]
      + expected - actual

       [
      -  "\u001b[M !!"
      -  "\u001b[M!!!"
      -  "\u001b[M\"!!"
      -  "\u001b[Ma!!"
         "\u001b[M#!!"
         "\u001b[M#!!"
         "\u001b[M#!!"
         "\u001b[M`!!"
      
      at Context.<anonymous> (out/common/services/CoreMouseService.test.js:166:27)
```

I have the same error on `master`. Have no idea what it is :).


src/common/services/CoreMouseService.test.ts
```diff
@@ -171,23 +171,20 @@ describe('CoreMouseService', () => {
       assert.equal(cms.triggerMouseEvent({ col: 0, row: 0, button: CoreMouseButton.RIGHT, action: CoreMouseAction.DOWN, ctrl: false, alt: false, shift: false }), true);
       assert.equal(cms.triggerMouseEvent({ col: 0, row: 0, button: CoreMouseButton.WHEEL, action: CoreMouseAction.DOWN, ctrl: false, alt: false, shift: false }), true);
       assert.deepEqual(reports, ['\x1b[M !!', '\x1b[M!!!', '\x1b[M"!!', '\x1b[Ma!!']);
-      while (reports.pop()) { }
```
jerch, line 174: This cannot be removed without replacement, as it clears the reports array (which gets reused right below again). For an optimized replacement, maybe do `reports.length = 0;` here.

src/common/services/CoreMouseService.ts
```diff
@@ -174,8 +174,10 @@ export class CoreMouseService implements ICoreMouseService {
     @ICoreService private readonly _coreService: ICoreService
   ) {
     // register default protocols and encodings
-    Object.keys(DEFAULT_PROTOCOLS).forEach(name => this.addProtocol(name, DEFAULT_PROTOCOLS[name]));
-    Object.keys(DEFAULT_ENCODINGS).forEach(name => this.addEncoding(name, DEFAULT_ENCODINGS[name]));
+    for (const name of Object.keys(DEFAULT_PROTOCOLS))
+    {this.addProtocol(name, DEFAULT_PROTOCOLS[name]);}
+    for (const name of Object.keys(DEFAULT_ENCODINGS))
+    {this.addEncoding(name, DEFAULT_ENCODINGS[name]);}
```
jerch, line 180: What with these replacements? Why are they done and formatted that way?

.putout.json
```diff
@@ -0,0 +1,10 @@
+{
+    "rules": {
+        "remove-console": "off",
+        "remove-unused-variables": "off",
+        "remove-useless-variables": "off",
+        "apply-destructuring": "off",
+        "merge-if-statements": "off",
+        "convert-apply-to-spread": "off"
+    }
+}
```
jerch, line 10: Whats this about?

src/browser/AccessibilityManager.ts
```diff
@@ -116,7 +116,7 @@ export class AccessibilityManager extends Disposable {
 
     // Don't scroll if the buffer top has reached the end in that direction
     const posInSet = boundaryElement.getAttribute('aria-posinset');
-    const lastRowPos = position === BoundaryPosition.TOP ? '1' : `${this._terminal.buffer.lines.length}`;
+    const lastRowPos = position === BoundaryPosition.TOP ? '1' : String(this._terminal.buffer.lines.length);
```
jerch, line 119: How is that better than the original? Doesn't `String` create a string object instead of a primitive string?

src/browser/Color.ts
```diff
@@ -111,9 +111,9 @@ export namespace rgb {
     const rs = r / 255;
     const gs = g / 255;
     const bs = b / 255;
-    const rr = rs <= 0.03928 ? rs / 12.92 : Math.pow((rs + 0.055) / 1.055, 2.4);
-    const rg = gs <= 0.03928 ? gs / 12.92 : Math.pow((gs + 0.055) / 1.055, 2.4);
-    const rb = bs <= 0.03928 ? bs / 12.92 : Math.pow((bs + 0.055) / 1.055, 2.4);
+    const rr = rs <= 0.03928 ? rs / 12.92 : ((rs + 0.055) / 1.055) ** 2.4;
+    const rg = gs <= 0.03928 ? gs / 12.92 : ((gs + 0.055) / 1.055) ** 2.4;
+    const rb = bs <= 0.03928 ? bs / 12.92 : ((bs + 0.055) / 1.055) ** 2.4;
```
jerch, line 116: Imho thats a dangerous change, as it is quite new still. To be on the safe side we should stick to Math.pow() for now. Although Math.pow is a function call, it creates not much overhead, I see no reason why switching at all here.

src/browser/Linkifier2.ts
```diff
@@ -98,7 +97,7 @@ export class Linkifier2 extends Disposable implements ILinkifier2 {
       if (target.classList.contains('xterm-hover')) {
         return;
       }
-    }
+    };
```
jerch, line 100: Why the semicolon here? Its superfluous.

src/browser/renderer/Renderer.ts
```diff
@@ -71,7 +71,8 @@ export class Renderer extends Disposable implements IRenderer {
   }
 
   public dispose(): void {
-    this._renderLayers.forEach(l => l.dispose());
+    for (const l of this._renderLayers)
+    {l.dispose();}
```
jerch, line 75: Formatting broken again.

src/browser/renderer/Renderer.ts
```diff
@@ -89,18 +90,19 @@ export class Renderer extends Disposable implements IRenderer {
...
   }
 
   public onResize(cols: number, rows: number): void {
     // Update character and canvas dimensions
     this._updateDimensions();
 
     // Resize all render layers
-    this._renderLayers.forEach(l => l.resize(this.dimensions));
+    for (const l of this._renderLayers)
+    {l.resize(this.dimensions);}
```
jerch, line 105: Formatting, happening several times more across the changes(not commenting on it).

src/browser/renderer/atlas/DynamicCharAtlas.ts
```diff
@@ -324,12 +324,11 @@ export class DynamicCharAtlas extends BaseCharAtlas {
 
       // Mark all new glyphs as in bitmap, excluding glyphs that came in after
       // the bitmap was requested
-      for (let i = 0; i < glyphsMovingToBitmap.length; i++) {
-        const value = glyphsMovingToBitmap[i];
+      for (const value of glyphsMovingToBitmap) {
```
jerch, line 327: This would have to be benchmarked as it is time critical code. Note that in past the fastest access to array members was good  old loop with index access. Might have changed though...

src/common/input/TextDecoder.ts
```diff
@@ -188,15 +188,11 @@ export class Utf8ToUtf32 {
             target[size++] = cp;
           }
         } else if (type === 3) {
-          if (cp < 0x0800 || (cp >= 0xD800 && cp <= 0xDFFF)) {
-            // illegal codepoint
-          } else {
+          if (!(cp < 0x0800 || cp >= 0xD800 && cp <= 0xDFFF)) {
```
jerch, line 191: Thats not a good replacement, as now all sub conditions have to be tested, to get the `!` running. Again time critical code.

src/common/services/CoreMouseService.ts
coderaiser: Formatting can be fixed adding eslint rule [brace-style](https://eslint.org/docs/rules/brace-style), but it will affect all code like:

```js
-const f = () => {return x};
+const f = () => {
    return x
};
```
I can fix this by hand to avoid big code changes if you agree with conversion :)

src/browser/renderer/atlas/DynamicCharAtlas.ts
coderaiser: I think it is now also the fastest, so we can just keep it. Much better to convert `for-each` to `for-of` because `for-of` should be faster :)

src/browser/Linkifier2.ts
coderaiser: I can add [semi](https://eslint.org/docs/rules/semi) eslint rule to prevent all such cases, is it OK?

src/browser/Color.ts
coderaiser: I see nothing dangerous but I’m OK with it, it can be disabled :)

src/browser/AccessibilityManager.ts
coderaiser: It create string in direct obvious and readable form :), rule can be disabled

.putout.json
coderaiser: This is current `putout` config, I can remove it, or disable rules you mentioned and fix all comments

coderaiser:
@jerch thank you for review :), I did look at replacement but want to receive feedback before proceeding. I’ll fix all comments when get to a computer.

coderaiser:
@jerch just force pushed updates, we have only 4 rules not disabled, from found list :):

- convert-for-each-to-for-of
- simplify-ternary
- apply-shorthand-properties
- extract-sequence-expressions

jerch:
@coderaiser 
Did a quick [microbenchmark](https://jsperf.com/classical-loop-vs-loop-of/1) comparing for-of with classical loop types known to be fast. Summary: for-of takes ~2 times longer on recent chrome, ~10 times longer on recent Firefox. Perfwise for-of is still smelly and does not look good compared to index loops.

I think for time critical parts we should resort to good old index-for-loops. For non critical paths imho for-of is good to go with it. `forEach` is way slower than all of them above, and has weird restrictions on the used type. for-of is clearly the better choice here.

@Tyriar Do you see any issues with for-of, like ECMA version supported? Imho it got official with ES6, I am not sure if TS will transpile it for lower targets correctly.

coderaiser:
@jerch 
> Did a quick microbenchmark comparing for-of with classical loop types known to be fast. Summary: for-of takes ~2 times longer on recent chrome, ~10 times longer on recent Firefox. Perfwise for-of is still smelly and does not look good compared to index loops.

> I think for time critical parts we should resort to good old index-for-loops.

Absolutely agree with you :). The rule `convert-fo-to-for-of` already disabled.

> I think for time critical parts we should resort to good old index-for-loops. For non critical paths imho for-of is good to go with it. forEach is way slower than all of them above, and has weird restrictions on the used type. for-of is clearly the better choice here.

Yep, `for-of` definitely better then `forEach` :).

> remove-empty is the only other rule that I think we may want to run. If that only catches the while (reports.pop()) { }, which would definitely be better written as reports = [] as suggested anyway.

Yes, I'll add it, it is have sense :).

PR #2962, opened 2020-06-03
Only removeChild valid children in DomRenderer dispose

### LINKED ISSUE #2960: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
https://github.com/xtermjs/xterm.js/issues/920#issuecomment-635014667

JoshuaKGoldberg:
Fixes #2960 

I did a scan for where I should add unit tests but it looks like the DOM renderer itself isn't unit tested - just `DomRendererRowFactory`.

### DIFF
--- a/src/browser/renderer/dom/DomRenderer.ts
+++ b/src/browser/renderer/dom/DomRenderer.ts
@@ -102,13 +102,13 @@ export class DomRenderer extends Disposable implements IRenderer {
       this._rowContainer,
       this._selectionContainer,
       this._themeStyleElement,
-      this._dimensionsStyleElement,
+      this._dimensionsStyleElement
     ]) {
       if (element.parentElement === this._screenElement) {
         this._screenElement.removeChild(element);
       }
     }
-  
+
     super.dispose();
   }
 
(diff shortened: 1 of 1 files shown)


src/browser/renderer/dom/DomRenderer.ts
```diff
@@ -95,10 +95,20 @@ export class DomRenderer extends Disposable implements IRenderer {
...
+    // https://github.com/xtermjs/xterm.js/issues/2960
+    for (const element of [
+      this._rowContainer,
+      this._selectionContainer,
+      this._themeStyleElement,
+      this._dimensionsStyleElement
+    ]) {
+      if (element.parentElement === this._screenElement) {
+        this._screenElement.removeChild(element);
+      }
```
Tyriar, line 109: Would this work instead to be more concise?

```
this.._rowContainer.parentNode?.removeChild(this._rowContainer);
```

We could also start a `browser/Dom.ts` for helpers like this: `removeFromParent(...elements: HTMLElement[])`

JoshuaKGoldberg: I think that works, yes! Updated.

For `browser/Dom.ts`, I'll let you do that 😄 I don't have a good feeling for this repository's style preferences.

Tyriar: The helper: https://github.com/xtermjs/xterm.js/pull/2966

PR #2976, opened 2020-06-12
keyboard fixes

### LINKED ISSUE #2151: Scandinavian keyboard layout can't handle input tilde (~)
To type `~` with Finnish/Swedish keyboard I have to press hotkey `alt+^`:
image

After that it shows something similar to `~` but right after when I press any other character (except Enter) it will erase "pre-tilde" symbol. Press `Enter` "converts" to normal ~ and nothing more (without real Enter).

I checked how this works with vanilla JS and it seems something wrong at browser-level:
image

Is it fixable? Wdyt?

## Related bugs:
https://github.com/xtermjs/xterm.js/issues/174

## Details
- Browser and browser version: `Chrome (latest)`
- OS version: `MacOS X (latest)`
- xterm.js version: `3.14.0`

### Steps to reproduce
1. Add  support of Finnish keyboard in your OS
2. Try to use ~ from keyboard

jerch:
Shall fix #2151.

TODO:
- [ ] collect key and event info from systems / browsers
- [ ] create `KeyboardService`, move handlers over
- [ ] pull printables from `input` event, others from `keydown`
- [ ] special deadkey state (passthrough to next input)
- [ ] IME compat?
- [ ] How to test automatically?

@Tyriar Did some further tests with https://w3c.github.io/uievents/tools/key-event-viewer.html and some spec reading to get an idea, what we can/should use. Specwise `beforeinput` is the best event to grab pending input data, but thats currently not supported by firefox (got fixed a few weeks ago, but not yet rolled out). `input` is the second best, but only gets fired AFTER the content change was made, which is abit cumbersome (we might have to clear the textarea afterwards, not sure yet, if this might cause problems with IMEs). Featurewise `keypress` is the best event, but it is deprecated, we should not rely on it anymore. For non printable chars we can basically stick with the `keydown` handler we already have, but need to extend it to pass printables along to be caught by the later `input` event.

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -336,11 +336,11 @@ export class Terminal extends CoreTerminal implements ITerminal {
     //this.register(addDisposableDomListener(this.textarea!, 'keyup', (ev: KeyboardEvent) => this._keyUp(ev), true));
     //this.register(addDisposableDomListener(this.textarea!, 'keydown', (ev: KeyboardEvent) => this._keyDown(ev), true));
     //this.register(addDisposableDomListener(this.textarea!, 'keypress', (ev: KeyboardEvent) => this._keyPress(ev), true));
-    const keyboardHelper = new KeyboardHelper();
-    this.register(addDisposableDomListener(this.textarea!, 'keyup', keyboardHelper.up, true));
-    this.register(addDisposableDomListener(this.textarea!, 'keydown', keyboardHelper.down, true));
-    this.register(addDisposableDomListener(this.textarea!, 'keypress', keyboardHelper.press, true));
-    this.register(addDisposableDomListener(this.textarea!, 'input', keyboardHelper.input, true));
+    const keyboardHelper = new KeyboardHelper(this._coreService);
+    //this.register(addDisposableDomListener(this.textarea!, 'keyup', keyboardHelper.up.bind(keyboardHelper), true));
+    this.register(addDisposableDomListener(this.textarea!, 'keydown', keyboardHelper.down.bind(keyboardHelper), true));
+    //this.register(addDisposableDomListener(this.textarea!, 'keypress', keyboardHelper.press.bind(keyboardHelper), true));
+    this.register(addDisposableDomListener(this.textarea!, 'input', keyboardHelper.input.bind(keyboardHelper), true));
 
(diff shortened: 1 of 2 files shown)


mofux:
I can also recommend https://github.com/Microsoft/vscode/wiki/Keybinding-Issues, especially the section [e.code, e.keyCode and e.key](https://github.com/Microsoft/vscode/wiki/Keybinding-Issues#ecode-ekeycode-and-ekey) that provides some very nice observations about the differences with evt.keyCode and evt.key. It also gives some insights how vscode deals with different keyboard layouts / mappings.

Tyriar:
Also note that we can't use `native-keymap`.

Tyriar:
Can we organize things a little more so events are tracked all the way through, like we can gather the `input`, `keydown` events (and any others needed) and then consider what to do together all in one spot?

jerch:
@mofux Yeah - I gonna try to rely mostly on `.key` and `.code` for non printables, `.keyCode` is not officially specced in a platform independent way.

@Tyriar Thats prolly a good idea. Back at the mouse events I came up with the idea to do all browser related event registering stuff in one place for mouse events (not happened yet). Not sure yet, if that idea would come handy here as well, I am still in the "collecting phase" to understand the platform/browser awkwardities.

jerch:
@Tyriar, @mofux Found a working set of rules to catch all keys with deadkeys under linux (even 3rd shifted deadkeys work correctly). Tested with Firefox and Chromium.

Imho these rules should work mostly on OSX and windows as well. Only saw minor event differences, that might need special adjustments.

jerch:
I have currently no time to further deal with this, if anyone wants to take it, plz go for it.

@Tyriar The rewrite of keyboard input will prolly need a custom event to be spawned, while the original key events are swallowed and stored up until the custom handler can make sense of the input (final char(s) appears from the sequence of pressed keys). The terminal then would only proceed with those final chars. This way makes it possible to deal with complex key combinations like third+ shifts. Normally the browser would already do the correct translation (fiddle out the final char), if the original key events are allowed to be processed normally (not captured isolated, not cancelled, bubbling etc).

Tyriar:
@jerch thanks for the update, monaco wraps dom events in custom events too that normalize things and makes working with them easier. I'll close this off as the diff will stick around and the PR is pointed to from the issue.

PR #2996, opened 2020-07-09
Sync textarea with cursor

### LINKED ISSUE #2598: Keep textarea in sync with cursor

jeanp413:
Fixes #2598

### DIFF
--- a/css/xterm.css
+++ b/css/xterm.css
@@ -59,10 +59,10 @@
 }
 
 .xterm .xterm-helper-textarea {
-    /*
-     * HACK: to fix IE's blinking cursor
-     * Move textarea out of the screen to the far left, so that the cursor is not visible.
-     */
+    padding: 0;
+    border: 0;
+    margin: 0;
+    /* Move textarea out of the screen to the far left, so that the cursor is not visible */
     position: absolute;
     opacity: 0;
     left: -9999em;
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -282,7 +282,7 @@ export class Terminal extends CoreTerminal implements ITerminal {
   }
 
   private _syncTextArea(): void {
-    if (!this.buffer.isCursorInViewport || this._compositionHelper!.isComposing) {
+    if (!this.textarea || !this.buffer.isCursorInViewport || this._compositionHelper!.isComposing) {
       return;
     }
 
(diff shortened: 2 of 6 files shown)


src/browser/Terminal.ts
```diff
@@ -281,6 +281,26 @@ export class Terminal extends CoreTerminal implements ITerminal {
...
+      return;
+    }
+
+    const cellHeight = Math.ceil(this._charSizeService!.height * this.optionsService.options.lineHeight);
+    const cursorTop = this._bufferService.buffer.y * cellHeight;
+    const cursorLeft = this._bufferService.buffer.x * this._charSizeService!.width;
+
+    // Sync the textarea to the exact position of the composition view so the IME knows where the
+    // text is.
+    this.textarea!.style.position = 'absolute';
```
Tyriar, line 295: We shouldn't need to set position here anymore as it's always absolute afaict?

src/browser/Terminal.ts
```diff
@@ -281,6 +281,26 @@ export class Terminal extends CoreTerminal implements ITerminal {
...
+    this.textarea!.style.left = cursorLeft + 'px';
+    this.textarea!.style.top = cursorTop + 'px';
+    this.textarea!.style.width = this._charSizeService!.width + 'px';
+    this.textarea!.style.height = cellHeight + 'px';
+    this.textarea!.style.lineHeight = cellHeight + 'px';
+    this.textarea!.style.zIndex = '-5';
```
Tyriar, line 301: Use of `!` is dangerous, this would throw an error if `onCursorMove`s fires before `Terminal.open` is called.

css/xterm.css
```diff
@@ -59,10 +59,10 @@
 }
 
 .xterm .xterm-helper-textarea {
-    /*
-     * HACK: to fix IE's blinking cursor
-     * Move textarea out of the screen to the far left, so that the cursor is not visible.
-     */
+    padding: 0;
+    border: 0;
+    margin: 0;
```
Tyriar, line 64: Are these needed when `opacity: 0` is used?

jeanp413: Not really but I added them just to get the textarea exactly match the cursor :sweat_smile: 
Before:
Screenshot from 2020-07-09 20-18-35
After:
Screenshot from 2020-07-09 20-15-25

Let me know if that's too much so I can revert that change

src/browser/Terminal.ts
jeanp413: done

jeanp413:
Pushed some changes addressing the feedback

css/xterm.css
Tyriar: Looks good 👌

struanb:
I believe this may have created a regression. Between v4.7.0 (good) and v4.8.1 (bad), I am seeing something that looks like a second cursor displayed next to the real cursor. Inspecting the xterm in DOM rendering mode shows it to be the the textarea.xterm-helper-textarea element.

According to the CSS class for this element, it should be hidden, except that the element's attributes override its class and cause it to be displayed.

Please see the attached screenshot for what this looks like.

image

jeanp413:
@struanb Is that on a mac? I tested it on linux and for me the textarea is not visible. What rule overrides the opacity?

struanb:
@jeanp413 Yes it is on MacOS Chrome. Good question what rule overrides opacity. Not sure. Does this screenshot help?

image

jeanp413:
@struanb oh I see, that's a theia specific issue related to the css rule `:focus`, there is an issue about it on theia https://github.com/eclipse-theia/theia/issues/8158

struanb:
Ah thanks @jeanp413. Sorry this turned out to be not your issue, but glad we've got to the bottom of it. Then I think you can safely close this issue again.

PR #2999, opened 2020-07-09
fix renderer pausing inadvertently in some cases

### LINKED ISSUE #2998: Terminal.Open during expand animation can cause indefinitely paused renderer
I've run into what i think is an edge case that causes my terminal to be stuck in a paused state where no output is displayed until the terminal element is scrolled out and back into the viewport, or the user switches tabs and then switches back.

Continue reading for further details, or skip to the bottom for my proposed cause and solution.

I am running xterm in an Angular app in an element inside of a material expansion panel. The terminal component is only rendered if the expansion panel is open. Terminal.Open is called when the component is initialized.

[Terminal.Open](https://xtermjs.org/docs/api/terminal/classes/terminal/#open) states "This element must be visible (have dimensions) when open is called as several DOM- based measurements need to be performed when this function is called." so I initially thought this was the problem, but I have tried checking to ensure that the target element is in the DOM and visible in every way I could think of (parentOffset, display style, width, height, etc) and it always shows as visible before the call to open. However, even though the element is visible, the expand animation has not yet completed, which it seems is causing the terminal renderer to be paused, and it is not unpaused when the animation completes.

## Details
- Browser and browser version: Chrome 83.0.4103.97
- OS version: Windows 10 1903
- xterm.js version: 4.7.0

### Steps to reproduce

I created a reproduction of the issue [here](https://stackblitz.com/edit/angular-ivy-hb17bb?file=src%2Fapp%2Fterminal%2Fterminal.component.ts). This simple example shows 2 expansion panels that should both output a line to the terminal every 2 seconds when expanded. The "Broken" one calls Terminal.Open as soon as the component is initialized and shows a blank terminal that is paused. The "Working" one waits 2 seconds before calling Terminal.Open and shows the terminal output correctly, even though both panels check that the target element's offsetParent is not null. If you switch tabs and switch back while the panels are open, the "Broken" one will start to work, without re-opening it.

I know that there are workarounds to this. For example, if I switch the example to use the afterExpand and afterCollapse outputs of the expansion-panel instead of open and close, but this isn't always practical and it still shows an underlying issue that was difficult to troubleshoot.

To manually reproduce:

1. Call Terminal.Open after the target element is visible, but before a material-expansion-panel expand is complete
2. Call Terminal.Write
3. Observe that no output can be seen
4. Scroll the terminal out of the viewport and back in
5. Observe output is now visible

### Proposed Cause & Solution
I believe this is caused by [RenderService.ts:89](https://github.com/xtermjs/xterm.js/blob/master/src/browser/services/RenderService.ts#L89). On [RenderService.ts:82](https://github.com/xtermjs/xterm.js/blob/master/src/browser/services/RenderService.ts#L82) the threshhold being passed in is explicitly 0, "meaning as soon as even one pixel is visible, the callback will be run" (https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API). But then you are pausing if the intersectionRatio is 0, which is a valid ratio that you asked for by setting the threshold to 0. I believe the appropriate check would be `isIntersecting` instead. From the docs: " Check the value of the isIntersecting property to see if the entry represents an element that currently intersects with the root."

I've submitted a PR to use `isIntersecting` for this check instead that I believe fixes this issue and still maintains the desired functionality of pausing the renderer when the terminal is not in view. 

Thank you.

andrew-schlackman:
Fixes #2998

### DIFF
--- a/src/browser/services/RenderService.ts
+++ b/src/browser/services/RenderService.ts
@@ -86,7 +86,7 @@ export class RenderService extends Disposable implements IRenderService {
   }
 
   private _onIntersectionChange(entry: IntersectionObserverEntry): void {
-    this._isPaused = entry.intersectionRatio === 0;
+    this._isPaused = !entry.isIntersecting;
     if (!this._isPaused && this._needsFullRefresh) {
       this.refreshRows(0, this._rowCount - 1);
       this._needsFullRefresh = false;


andrew-schlackman:
Thanks for the accept!

I did verify that it fixes the issue in my case.

Tyriar:
Great! I just released 4.8.0 so you can pick this up 🙂

Issue #2522, opened 2019-10-29
Move away from string-based options


Tyriar:
How about we expose `options: ITerminalOptions` that replaces `getOption` and `setOption`?

```ts
class Terminal {
  options: ITerminalOptions;
}
```

This would also enable setting single settings (via getters/setters on the internal `ITerminalOptions` implementation) or changing all settings at once (via getter/setter of `Terminal.options`).

```ts
// before
term.setOption('rendererType', 'dom');

// after
term.options.rendererType = 'dom';
// or if it's the only setting
term.options = {
  rendererType: 'dom'
};
```

jerch:
We have several such gems in options - how about enums vs. union types? Though enums are not that useful and cannot be `const` at public API level as long as we still target JS users.

Tyriar:
Union types are typically preferred by JS devs as they're more concise and easier to use (no importing needed), luckily typescript can treat them just as strictly as an enum so doesn't matter from my perspective.

PR #3062, opened 2020-08-31
Add number type to fontWeight & fontWeightBold options

### LINKED ISSUE #3051: Support number in fontWeight option
It's valid to have numbers ranging 1-1000: https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#Values

Ask: https://github.com/microsoft/vscode/pull/101468#pullrequestreview-468102004

IllusionMH:
Fixes #3051 

`_sanitizeAndValidateOption` now implements custom normalization for `fontWeight` & `fontWeightBold` options that matches normalization from [VSCode PR](https://github.com/microsoft/vscode/pull/101468/files#diff-bf729abb47c6d125996e13230061c850R348-R353).
This is needed because if invalid value (`7000` or `normal123`) is passes to `fontWeight` when canvas or webgl rendenderer is used it will lead to really small characters.
image

In browsers that don't support numeric values for `font-weight` property when values that doesn't match existing enum would be reset to `normal`/`400` - this is current browser behavior for `font-weight` and no special logic added to handle this case. I assume that users of these browsers shouldn't won't use numeric values e.g. 350 because they don't work anyway for everything else.

### DIFF
--- a/src/browser/public/Terminal.ts
+++ b/src/browser/public/Terminal.ts
@@ -169,15 +169,16 @@ export class Terminal implements ITerminalApi {
   public paste(data: string): void {
     this._core.paste(data);
   }
-  public getOption(key: 'bellSound' | 'bellStyle' | 'cursorStyle' | 'fontFamily' | 'fontWeight' | 'fontWeightBold' | 'logLevel' | 'rendererType' | 'termName' | 'wordSeparator'): string;
+  public getOption(key: 'bellSound' | 'bellStyle' | 'cursorStyle' | 'fontFamily' | 'logLevel' | 'rendererType' | 'termName' | 'wordSeparator'): string;
   public getOption(key: 'allowTransparency' | 'cancelEvents' | 'convertEol' | 'cursorBlink' | 'disableStdin' | 'macOptionIsMeta' | 'rightClickSelectsWord' | 'popOnBell' | 'visualBell'): boolean;
   public getOption(key: 'cols' | 'fontSize' | 'letterSpacing' | 'lineHeight' | 'rows' | 'tabStopWidth' | 'scrollback'): number;
+  public getOption(key: 'fontWeight' | 'fontWeightBold'): string | number;
   public getOption(key: string): any;
   public getOption(key: any): any {
     return this._core.optionsService.getOption(key);
   }
   public setOption(key: 'bellSound' | 'fontFamily' | 'termName' | 'wordSeparator', value: string): void;
-  public setOption(key: 'fontWeight' | 'fontWeightBold', value: 'normal' | 'bold' | '100' | '200' | '300' | '400' | '500' | '600' | '700' | '800' | '900'): void;
+  public setOption(key: 'fontWeight' | 'fontWeightBold', value: 'normal' | 'bold' | '100' | '200' | '300' | '400' | '500' | '600' | '700' | '800' | '900' | number): void;
   public setOption(key: 'logLevel', value: 'debug' | 'info' | 'warn' | 'error' | 'off'): void;
   public setOption(key: 'bellStyle', value: 'none' | 'visual' | 'sound' | 'both'): void;
   public setOption(key: 'cursorStyle', value: 'block' | 'underline' | 'bar'): void;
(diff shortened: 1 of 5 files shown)


src/browser/public/Terminal.ts
```diff
@@ -169,15 +169,16 @@ export class Terminal implements ITerminalApi {
   public paste(data: string): void {
     this._core.paste(data);
   }
-  public getOption(key: 'bellSound' | 'bellStyle' | 'cursorStyle' | 'fontFamily' | 'fontWeight' | 'fontWeightBold' | 'logLevel' | 'rendererType' | 'termName' | 'wordSeparator'): string;
+  public getOption(key: 'bellSound' | 'bellStyle' | 'cursorStyle' | 'fontFamily' | 'logLevel' | 'rendererType' | 'termName' | 'wordSeparator'): string;
   public getOption(key: 'allowTransparency' | 'cancelEvents' | 'convertEol' | 'cursorBlink' | 'disableStdin' | 'macOptionIsMeta' | 'rightClickSelectsWord' | 'popOnBell' | 'visualBell'): boolean;
   public getOption(key: 'cols' | 'fontSize' | 'letterSpacing' | 'lineHeight' | 'rows' | 'tabStopWidth' | 'scrollback'): number;
+  public getOption(key: 'fontWeight' | 'fontWeightBold'): string | number;
```
IllusionMH, line 175: I've used wide type `string | number` to match previous wide `string`, however now `fontWeight` should always match `FontWeight` type because of custom normalization on `OptionsService` level.
Should I use narrow `FontWeight` type here and in `xterm.d.ts`?

src/browser/public/Terminal.ts
```diff
@@ -169,15 +169,16 @@ export class Terminal implements ITerminalApi {
...
   public getOption(key: string): any;
   public getOption(key: any): any {
     return this._core.optionsService.getOption(key);
   }
   public setOption(key: 'bellSound' | 'fontFamily' | 'termName' | 'wordSeparator', value: string): void;
-  public setOption(key: 'fontWeight' | 'fontWeightBold', value: 'normal' | 'bold' | '100' | '200' | '300' | '400' | '500' | '600' | '700' | '800' | '900'): void;
+  public setOption(key: 'fontWeight' | 'fontWeightBold', value: 'normal' | 'bold' | '100' | '200' | '300' | '400' | '500' | '600' | '700' | '800' | '900' | number): void;
```
IllusionMH, line 181: Is there a reason why `value` uses union here instead of `FontWeight` type? Is it to improve intellisense/hovers or I should use narrow `FontWeight` type here and in `xterm.d.ts`?

IllusionMH (commented):
Couple of questions

src/browser/public/Terminal.ts
Tyriar: Yeah we can narrow to `FontWeight`, I think tsc ends up allowing anything right now anyway because of `getOption(key: any): any`. I wanted to move to something stricter later on which would fix this: https://github.com/xtermjs/xterm.js/issues/2522

src/browser/public/Terminal.ts
Tyriar: Yes, I think the reason was to make it more obvious using intellisense as API users wouldn't know what a `FontWeight` is within drilling into it. Let's stick to listing them out for now, again tsc is permissive because `setOption(key: any, value: any): void` is in the public API.

Tyriar (changes requested):

IllusionMH:
As per first feedback I've changed return type of `getOption` to `FontWeight` alias.
As for second comment about argument - I think there was no call to action so not changed.
Also added tests for this property behavior.

P.S. If function previously had return type `string` but now it also has `number` in return type - can this change be counted as minor/feature change or it's major/breaking change?
In fact previously it would return anything except falsy values, and now it's subset of actual previous return values, but superset of previously declared type.

Tyriar (approved):
> If function previously had return type string but now it also has number in return type - can this change be counted as minor/feature change or it's major/breaking change?

Yes it is technically a breaking change but I'm fine with letting this one go into a minor patch as it's such a small thing that maybe no one would eve. I don't know when v5 will happen and we don't want to block the VS Code change until then. I checked a few apps that use xterm.js and saw Hyper uses `getOption` only once for `bellSound`.

PR #3064, opened 2020-08-31
Validate constructor options

### LINKED ISSUE #3063: OptionsService should validate and sanitize options passed to constructor
In scope of #3062 I've noticed that there is call to [`_sanitizeAndValidateOption`](https://github.com/xtermjs/xterm.js/blob/60419a7cc6edf46852394119401927f45c46a515/src/common/services/OptionsService.ts#L104) during `setOption` call
https://github.com/xtermjs/xterm.js/blob/60419a7cc6edf46852394119401927f45c46a515/src/common/services/OptionsService.ts#L82-L93

But no calls in `constructor`.

Therefore it's possible to call `new Terminal({fontWeight: 'normal123'})` and receive same problem as shown in #3062
Xterm.js renders extremely small characters when invalid fontWeight value passed

I think it make sense to implement validation in `constructor` as well to remove difference in behavior between `constructor` and `setOption` and ensure that only valid values are used during construction.

IllusionMH:
Fixes #3063

Added high level tests for Terminal, but can move it to separate test just for `OptionsService`.

### DIFF
--- a/src/common/services/OptionsService.ts
+++ b/src/common/services/OptionsService.ts
@@ -74,7 +74,7 @@ export class OptionsService implements IOptionsService {
     for (const k of Object.keys(options)) {
       if (k in this.options) {
         const newValue = options[k as keyof IPartialTerminalOptions] as any;
-        this.options[k] = newValue;
+        this.options[k] = this._sanitizeAndValidateOption(k, newValue);
       }
     }
   }
--- a/src/common/services/OptionsService.ts
+++ b/src/common/services/OptionsService.ts
@@ -73,8 +73,12 @@ export class OptionsService implements IOptionsService {
     this.options = clone(DEFAULT_OPTIONS);
     for (const k of Object.keys(options)) {
       if (k in this.options) {
-        const newValue = options[k as keyof IPartialTerminalOptions] as any;
-        this.options[k] = this._sanitizeAndValidateOption(k, newValue);
+        try {
+          const newValue = options[k as keyof IPartialTerminalOptions] as any;
+          this.options[k] = this._sanitizeAndValidateOption(k, newValue);
+        } catch (e) {
+          console.error(e);
+        }
       }
     }
   }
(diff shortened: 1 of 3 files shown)


src/common/services/OptionsService.ts
```diff
@@ -74,7 +74,7 @@ export class OptionsService implements IOptionsService {
     for (const k of Object.keys(options)) {
       if (k in this.options) {
         const newValue = options[k as keyof IPartialTerminalOptions] as any;
-        this.options[k] = newValue;
+        this.options[k] = this._sanitizeAndValidateOption(k, newValue);
```
Tyriar, line 77: I like the change but it would be best if this logged instead of threw here, otherwise it's a breaking change.

Since log service depends on options service we can't really log property so maybe we should just wrap it in a try/catch with a `console.error` here?

IllusionMH: > just wrap it in a try/catch with a console.error here

Yes, that would be easier to avoid breaking change here, however couple of questions about prefered approach for test rises:
Do I need to keep test and stub `console.error` or just remove test?
If I need to keep test - can I install `sinon` (currently [used only for xterm-addon-ligatures](https://github.com/xtermjs/xterm.js/search?q=sinon&unscoped_q=sinon)) or just manually replace/restore function?

Tyriar: Instead of an extra dep or patching console.error, how about checking the setting after the ctor has run?

```
notEqual(new TestTerminal({tabStopWidth: 0}).getOption('tabStopWidth'), 0);
```

IllusionMH: I've checked that it doesn't throw, that there is expected message in console, and that value is same as default.
It's even easier to check if it just wasn't applied (will also keep that it doesn't throw)! 👍

IllusionMH:
Updated implementation to log errors instead of throwing.

Created separate test for `OptionsService` because checking options values in `TestTerminal` requires accessing private property `_core` and I think it make sense to add more test for `getOption`/`setOption`(+`_sanitizeAndValidateOption`).

In the end I decided to stub `console.error` for this subset of tests because otherwise it actually logs error in the middle of test runner output.

PR #3067, opened 2020-09-03
Fix cjk handling in SerializeAddon

### LINKED ISSUE #3066: xterm-addon-serialize incorrectly handle double width charaters (a.k.a. CJK)
## Details
- Browser and browser version: electron 9.2.1
- OS version: mac
- xterm.js version: 4.8.1
- xterm-addon-serialize version: 0.3.0

### Steps to reproduce

1.  type/copy some Chinese characters into console, for example `中文`
2. serialize it
3. write it into console

### Expect

It looks like what it is originaly saved

### Actually

There was extra `"\x1b1C"` after chinese charater

### Probably caused by

The xterm-addon-serialize incorrectly assume charaters are always 1 slot width, however it is an incorrect assumption for chinese charaters.

mmis1000:
Use `cell.getWidth()` to get correct cell width instead of assume 1 width of every empty cell.

https://github.com/xtermjs/xterm.js/blob/106ea62b4c5f64d7d9a9221f5fbb6270554b4212/addons/xterm-addon-serialize/src/SerializeAddon.ts#L140-L142

Fix #3066

### DIFF
--- a/addons/xterm-addon-serialize/src/SerializeAddon.api.ts
+++ b/addons/xterm-addon-serialize/src/SerializeAddon.api.ts
@@ -270,6 +270,35 @@ describe('SerializeAddon', () => {
     await writeSync(page, lines.join('\\r\\n'));
     assert.equal(await page.evaluate(`serializeAddon.serialize();`), expected.join('\r\n'));
   });
+
+  it('serialize CJK correctly', async () => {
+    const lines = [
+      '中文中文',
+      '12中文',
+      '中文12',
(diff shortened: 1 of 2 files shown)


jerch (commented):
Nice find, LGTM :+1: 

On the background of this: We actually have 2 types of null cells - the default one has a width of 1, which still accounts towards cursor offsets and has a default output representation of one SP. The second one appears after full width chars with a width of 0, which does not account anywhere and has no own output repr. Thus it is important to grab to true cell width for nullish cells to correctly calculate cursor offsets and/or filling spaces.

Tyriar (approved):
Thanks 👍

PR #3075, opened 2020-09-06
Expose API to clear texture atlas

Tyriar:
Using canvases on Electron (and plain Chromium I believe), when resuming
from sleep the graphics card may release the texture which results in
the terminal display and char atlas being corrupt. This API allows the
consumer to clear the texture atlas right after resuming, atlas warm up
is not done to simplify this process because the fact that atlases are
shared between multiple terminals complicates this.

Related microsoft/vscode#69665

### DIFF
--- a/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
+++ b/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
@@ -101,7 +101,7 @@ export class WebglCharAtlas implements IDisposable {
     }
   }
 
-  protected _doWarmUp(): void {
+  private _doWarmUp(): void {
     // Pre-fill with ASCII 33-126
     for (let i = 33; i < 126; i++) {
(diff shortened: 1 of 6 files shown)


addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
```diff
@@ -115,17 +115,26 @@ export class WebglCharAtlas implements IDisposable {
...
     return false;
   }
 
+  public clearTexture(): void {
+    if (this._currentRowX === 0 && this._currentRowY === 0) {
+      return;
+    }
+    this._cacheCtx.clearRect(0, 0, TEXTURE_WIDTH, TEXTURE_HEIGHT);
+    this._cacheMap = {};
+    this._cacheMapCombined = {};
```
Tyriar, line 131: This PR also clears the combined cache map on texture clear which wasn't happening before, probably resulting in weird graphical artifacts.

Issue #3085, opened 2020-09-09
`entry.intersectionRatio` is not exactly 1.0 


aminya:
## Details
- Browser and browser version: Electron 6
- OS version: Win 10
- xterm.js version: 4.9.0

### Steps to reproduce

We hit a strange bug that was blocking us for a couple of days. Until we found the `intersectionRatio` is sometimes near to 1.0 (like 0.99) but not exactly 1.0. 

https://github.com/atom-ide-community/terminal/issues/5#issuecomment-689720693
https://github.com/atom-ide-community/terminal/issues/5

I just wanted to let you know about this behavior.

We had a condition to create xterm Element only if `intersectionRatio === 1.0` which was not being triggered (this condition was to avoid https://github.com/bus-stop/x-terminal/issues/57).

### Suggestion
Introduce a boolean value that shows if the `intersectionRatio` is almost 1.

UziTech:
This is probably a bug in chrome not xterm.

aminya:
Where can we report this bug? Should it be chromium? https://chromium.googlesource.com/chromium/src/

aminya:
There seems to exist a `isIntersecting` property:
https://github.com/xtermjs/xterm.js/blob/b89d4b102e13ad0521097e8f0fb5d3b89910a494/src/browser/services/RenderService.ts#L89

Tyriar:
Looks like you've already fixed it, no?

https://github.com/atom-ide-community/terminal/commit/7a0545149fc5d5917a322bd06f9f9af7a5308717

We only care if it's partially visible (`entry.intersectionRatio === 0`):

https://github.com/xtermjs/xterm.js/blob/b89d4b102e13ad0521097e8f0fb5d3b89910a494/src/browser/services/RenderService.ts#L89

aminya:
We solved it by a workaround, but this is a xterm issue. xterm should not allow the following error:
https://github.com/bus-stop/x-terminal/issues/57

UziTech:
That error comes from the fit addon. Running fit when the canvas that has a 0 width or height will throw the error `Uncaught Error: This API only accepts integers`

Tyriar:
We could handle that better but in the end the right behavior is probably still throw, just with a better message about NaN terminal dimensions.

UziTech:
I feel like fit being called on a hidden element or one with 0 width or height shouldn't throw an error. It should just check for that situation and return early without doing anything. I can't see a situation where that would be a problem.

[...]

PR #3088, opened 2020-09-10
fix: fit addon proposing infinite dimensions

UziTech:
fixes https://github.com/xtermjs/xterm.js/issues/3085#issuecomment-690318136

### DIFF
--- a/addons/xterm-addon-fit/test/FitAddon.api.ts
+++ b/addons/xterm-addon-fit/test/FitAddon.api.ts
@@ -23,11 +23,19 @@ describe('FitAddon', () => {
     page = await (await browser.newContext()).newPage();
     await page.setViewportSize({ width, height });
     await page.goto(APP);
+  });
+
+  beforeEach(async function(): Promise<any> {
+    await page.evaluate(`document.querySelector('#terminal-container').style.display=''`);
     await openTerminal(page);
   });
 
   after(async () => {
     await browser.close();
+  });
+
+  afterEach(async function(): Promise<any> {
+    await page.evaluate(`window.term.dispose()`);
   });
 
   it('no terminal', async function(): Promise<any> {
@@ -62,6 +70,15 @@ describe('FitAddon', () => {
         cols: 2,
         rows: 1
       });
+    });
+
+    it('hidden', async function(): Promise<any> {
+      await page.evaluate(`window.term.dispose()`);
+      await page.evaluate(`document.querySelector('#terminal-container').style.display='none'`);
+      await page.evaluate(`window.term = new Terminal()`);
+      await page.evaluate(`window.term.open(document.querySelector('#terminal-container'))`);
+      await loadFit();
+      assert.equal(await page.evaluate(`window.fit.proposeDimensions()`), undefined);
     });
   });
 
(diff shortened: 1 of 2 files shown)


addons/xterm-addon-fit/test/FitAddon.api.ts
```diff
@@ -63,6 +63,11 @@ describe('FitAddon', () => {
         rows: 1
       });
     });
+
+    it('hidden', async function(): Promise<any> {
+      await loadFit(0, 0, false);
```
Tyriar, line 68: Should there also be a test for `loadFit(0, 0, true)`?

Tyriar (changes requested):
Looks like the new test is failing in CI?

addons/xterm-addon-fit/test/FitAddon.api.ts
UziTech: I'm not sure how to test this. Changing the terminal div to `display: none;` makes `fit()` fail in x-terminal but not in the tests. I'll have to dig deeper on x-terminal and see what is making the dimensions equal to 0.

UziTech: I found the issue. It seems that this issue only happens when the termina container div is hidden when it is attached to the terminal so the renderer doesn't have a chance to create a width before calling fit.

Now to figure out how to hide the terminal in the tests before attaching it to the terminal.

UziTech:
It looks like this solves the issue of fit erroring when the terminal is attached to a hidden div but there are still other things that go wrong with the renderer. I will create another PR to address those issues.

addons/xterm-addon-fit/test/FitAddon.api.ts
```diff
@@ -111,3 +123,8 @@ async function loadFit(width: number = 800, height: number = 450): Promise<void>
 async function unloadFit(): Promise<void> {
   await page.evaluate(`window.fit.dispose();`);
 }
+
+async function resetTerminal(): Promise<void> {
+  await page.evaluate(`window.term.dispose()`);
+  await page.evaluate(`window.term.open(document.querySelector('#terminal-container'))`);
+}
```
Tyriar, line 130: We can instead move the `openTerminal` call into a `beforeEach` so the standard setup happens for each test, it's a pretty cheap action.

addons/xterm-addon-fit/test/FitAddon.api.ts
```diff
@@ -63,6 +63,18 @@ describe('FitAddon', () => {
...
     });
+
+    it('hidden', async function(): Promise<any> {
+      try {
+        await page.evaluate(`document.querySelector('#terminal-container').style.display='none'`);
+        await resetTerminal();
+        await loadFit();
+        assert.equal(await page.evaluate(`window.fit.proposeDimensions()`), undefined);
+      } finally {
+        await page.evaluate(`document.querySelector('#terminal-container').style.display=''`);
```
Tyriar, line 74: This display change could be moved into the `beforeEach` instead, that's always seemed safer than relying on try/finally to me to handle resetting things.

UziTech:
I created #3089 to handle the other issues created when the terminal is attached to a hidden div.

Tyriar (approved):

PR #3089, opened 2020-09-11
Render initially hidden terminal when visible

UziTech:
Render the terminal after it is visible when the parent was hidden when `open()` was called.

This PR uses the Intersection Observer created inside the RenderService to recalculate the dimensions.

Alternatively, we could create an Intersection Observer in the Terminal class to do this.


UziTech:
Thanks for updating it. :+1:

UziTech:
@Tyriar is this waiting on something?

UziTech:
@Tyriar can I help with anything to get this merged?

aminya (approved):
I think this fixes a lot of undebuggable strange bugs that prevent xterm to start. I look forward to seeing this merged.

src/browser/services/RenderService.ts
```diff
@@ -91,6 +91,11 @@ export class RenderService extends Disposable implements IRenderService {
       this.refreshRows(0, this._rowCount - 1);
       this._needsFullRefresh = false;
     }
+
+    // Terminal was hidden on open
+    if (!this._isPaused && !this._charSizeService.hasValidSize) {
```
jerch, line 96: Shouldn't this be done before the other condition above? Otherwise imho in case the renderer has no valid char measures, the refresh would fail/draw nonsense as well?

jerch (commented):
LGTM - but I have one question about the code order (see below).

src/browser/services/RenderService.ts
UziTech: Theoretically they should never be run at the same time since `this._needsFullRefresh === true` means that it has already been visible and `this._charSizeService.hasValidSize === false` means it has never been visible.

But I think you are correct if they ever were to run through both if statements we would want them to measure charsize first.

jerch: But isnt that the case if you do the following:
- spawn and show new terminal
- hide terminal
- change font settings
- show it again - now renders with wrong metrics

I did not write the renderer code so plz bear with me if that scenario is handled otherwise in the control flow, thus cannot happen at all.

UziTech: No because charsize is only not valid if measure has never been called. The first step will measure the charsize so even when the terminal is hidden the charsize will still be valid.

jerch: So I guess we just lost the changed font size in my steps above and would always render wrong glyphs?

UziTech: Measure is called in other places. This call is just for when initializing xterm on a hidden div.

jerch: Just tested it: font size changes during display='none' do not work, visibility='hidden' are correctly applied. Hmm.

But as you said, this has nothing to do with the code here, it gives up at a different place. I still think that the code order should be reversed (and maybe cleaned up/merged).

UziTech:
I update the order of the calls. This should be merged then the font size change on `display: none` should be fixed in another PR.

Tyriar (approved):
This looks good after the order switch. @UziTech can you create another issue capturing the bug with font size/option changes?

Issue #3014, opened 2020-07-16
Support event when the bell goes off


Tyriar:
Electron has an API to play the shell bell https://www.electronjs.org/docs/api/shell#shellbeep

To support this with xterm.js we would need to allow `bellSound` to take a `() => void`:

https://github.com/xtermjs/xterm.js/blob/ab63b5168161d5a3a93708f4dc274fadee176801/typings/xterm.d.ts#L47-L50

bolariin:
I would like to work on this issue

Tyriar:
@bolariin sounds good 👍

PR #3091, opened 2020-09-13
Allow bellSound to take a lambda

QuiiBz:
This PR adds a lambda callback for the `bellSound` option, as seen in #3014.

I've added a `bellSound` type which allows a string, or a callback, so we can still use the old way in addition to the new one:
```typescript
export type BellSound = string | (() => void);
```


typings/xterm.d.ts
```diff
@@ -25,6 +25,12 @@ declare module 'xterm' {
    */
   export type RendererType = 'dom' | 'canvas';
 
+  /**
+   * A string representing a data uri or a callback to play electron's
+   * shell beep for example.
+   */
+  export type BellSound = string | (() => void);
```
Tyriar, line 32: Let's remove this and inline the `string | () => void` type instead, this will lead to easier to use intellisense.

typings/xterm.d.ts
```diff
@@ -25,6 +25,12 @@ declare module 'xterm' {
    */
   export type RendererType = 'dom' | 'canvas';
 
+  /**
+   * A string representing a data uri or a callback to play electron's
+   * shell beep for example.
```
Tyriar, line 30: This isn't Electron specific, it just lets the embedding application play the bell however it wants.

src/browser/services/SoundService.ts
```diff
@@ -29,18 +29,32 @@ export class SoundService implements ISoundService {
...
       bellAudioSource.connect(ctx.destination);
       bellAudioSource.start(0);
     });
   }
 
+  public playBellSoundFromLambda(): void {
+    const callback = this._optionsService.options.bellSound as () => void;
+    callback();
+  }
+
```
Tyriar, line 57: Let's merge these back into a single function and assign `typeof this._optionsService.options.bellSound === 'string'` to a variable so we can take advantage of type narrowing, instead of using unsafe casting.

Tyriar (changes requested):

typings/xterm.d.ts
QuiiBz: Yes, this is why I specified "for example".

QuiiBz: Yes, this is why I specified "for example".

QuiiBz:
I pushed new changes as asked. Tell me if that good (it's my 2nd PR ever)

jerch (commented):
Tbh I find that API abit awkward, it is lit. the only place, where we would allow a function pointer to be attached directly to a terminal setting.

I'd propose a different handling here - since ringing the bell is something that can happen anytime from within the terminal, how about a `onBell` event, anyone could register on? The default could be our standard sound service being attached with its API to load different sound formats. Would be much more in line with our codebase with tons of `onEventXY` and kinda more JS-ish.

jerch:
@QuiiBz Do you want to change it in a real event on your own? If so, a look at `onTitleChanged` and how that gets through from `InputHandler` to the API might be worth a look. It would be basically the same for the bell, beside a switch to either route it through the standard sound service or fire an outgoing event. You can prolly do the distinction here with an else-clause firing the final event:

https://github.com/xtermjs/xterm.js/blob/8e151ee6ef8782f59828ca379a5ad5a0e641abd6/src/browser/Terminal.ts#L1114-L1117

mofux:
I like the idea of an `onBell` event handler. One additional thought: I think it should always fire, no matter how the `bellStyle` option is configured (which can be `none` or `sound`).

jerch:
@mofux Yepp same here. Do you have a good idea how to do that without breaking the API (otherwise it would have to wait until v5). Maybe ship with an interim state supporting both and always firing onBell, and remove the direct settings with v5?

Tyriar:
The event handler is way better than my idea 😄 

@jerch it's new API so we don't need to worry about breakages, right? We would still handle bell internally based on `bellStyle`, but fire the event regardless of `bellStyle`'s value.

jerch:
@Tyriar Well about an event here vs. a simple function pointer - thats a two sided thing for me and I kinda argue against myself:

In the past I mostly went with direct function pointers myself (see those `setHandler` calls in the parser). This was mainly for perf reasons, as a simple null check + call is way faster than the additional loop handling any event abstraction needs. But I was teached otherwise, when @PerBothner introduced the `addHandler` abstraction on the parser (which is a downstripped event system). Ofc this adds runtime on top, but it is really neglectable. And the advantages of events outweigh by far any slightly faster trickery.
 
For the bell I dont see perf as a criterion at all, which seems to make it a perfect fit for a custom event. Which fits much nicer into our API with its other events.

Tyriar:
Thanks for the PR for starting the discussion, we'll end up going with an event for this so closing this off.

PR #3101, opened 2020-09-16
Feature/xterm addon serialize restore layout and cursor

### LINKED ISSUE #3090: xterm-addon-serialize only serialize the alt screen if the terminal is in alt mode
## Details
- Browser and browser version: any (not a layout issue)
- OS version: windows/linux
- xterm.js version: https://github.com/xtermjs/xterm.js/tree/56bfdb64d47bee71b192566a5c1c2f4e7403a6a3
- xterm-addon-serialize version: https://github.com/xtermjs/xterm.js/tree/56bfdb64d47bee71b192566a5c1c2f4e7403a6a3
### Steps to reproduce

1. create a empty terminal
2. install the xterm-addon-serialize
3. write the following string into it
```js
    const SMCUP = '\u001b[?1049h'; // enable alternative screen
    const CUP = '\u001b[H'; // move  cursor to top left
    terminal.write(`1${SMCUP}${CUP}2`)
```
4. read the result using `serializeAddon.serialize()`

### Expect

You get the normal screen + alt screen in the result.
(Not sure which sequence is used to move the cursor, because there are too many ways to do it)

### Actually

You get `2`
The addon did not realize there is another screen exist.

### Note

I am not sure whether is it designed to not do it or not.
But the normal screen is required to rebuild the terminal from empty (or exit the htop will no longer bring you back to original prompt), so I guess it is missing?

I play with it a bit. [/src/SerializeAddon.ts#L192-L201k](https://github.com/mmis1000/xterm.js/blob/feature/xterm-addon-serialize-alt-screen/addons/xterm-addon-serialize/src/SerializeAddon.ts#L192-L201)

And if it should do it
1. whether it is always enabled or expose as a option?
2. do alt screen affected by the `rows` option?

### LINKED ISSUE #3098: xterm-addon-serialize does not preserve empty cell background
## Details
- Browser and browser version: any(not browser issue)
- OS version: mac
- xterm.js version: https://github.com/xtermjs/xterm.js/commit/003cf969104229e2b1ffbf92c1b6c018d36f8f3c
- xterm-addon-serialize version: https://github.com/xtermjs/xterm.js/commit/003cf969104229e2b1ffbf92c1b6c018d36f8f3c

### Steps to reproduce

1. open the demo
2. run `echo -e '\x1B[0;37;44mHI2\x1b[10X\x1b[m'`
3. check the `Write back to terminal`option of `SerializeAddon `
4. click `Serialize the content of terminal`
5. Notice the background of empty cell is gone

### Note

Also observed when do this on `htop`

mmis1000:
This pull request.

- Fix #3098
- Fix #3090
- Implement cursor position restoration
- Handles isWrapped property
- changes the `rows` option to `scrollback` to make it work with cursor restoration.
- Add a test util to `xterm-addon-serialize` that export the screen buffer state as plain object for testing.

Tested on a few full screen console programs to make sure the layout is identical to original state.
- htop
- vim
- nano
- sl


mmis1000:
WTF, how do the same byte computation result in -0 instead of 0 in firefox?

mofux:
I didn't even know this was possible. `0` and even `Infinity` can be signed!
And guess what, while `-0 === 0`, `-Infinity !== Infinity`:

Screenshot 2020-09-16 at 15 44 05

mmis1000:
It looks like either xtermjs did something undefined behavior during computation (but i think js seldom do that?), or the firefox implement js incorrectly somewhere.

jerch:
@mmis1000 
For equality check of buffer content it might be better to do the JSON comparison on the private `Bufferline` objects, as they are the source of truth for that (it is ok to access privates in test cases). This has the benefit of revealing potential bugs in API mapping code as well.

mmis1000:
@jerch 
That probably won't work well currently, I guess?
Because I haven't handle the `isWrapped` yet (thus this is currently a draft)
And there is indeed some data in BufferLine that are not available to public api (the IExtendedAttrs).

BTW, it seems there is no way to copy a IBufferCell over another. But to handle the BCE, it seems I need to keep the cursor style as background style to decide whether ECH or not. Should I just implement one fake class with it?

jerch:
Ah ok, yeah true. Sorry thought that is handled already.

mmis1000:
This should do a complete layout restore now (include line wrap).
The code probably still need some clear up for the left over during try and error.

About the performance, it handles about 5000 rows in 200ms on chrome.
(I am not sure whether it is fast enough. But given this is not designed to run on every tick, it is probably fast enough?)

jerch:
@mmis1000 Imho speed is an issue for a serializer (as it has to be blocking to obtain a correct state). We already had quite the discussion about it when reshaping the buffer API (see #2369). The fastest serializer version I was able to find for basic serialization (doing only colors and flags) is at ~25MB/s for final string output (doing 100k scrollbuffer in roughly ~1s for single unoptimized run). Much higher than that is almost impossible with JS, as pure string concatenation power of JS gets the limiting factor. See https://github.com/xtermjs/xterm.js/pull/2369#issuecomment-526160555, there is also a private where you can look at the code. But note that I used very different code than the final serializer.

Roughly estimated your current version should be somewhat around ~2MB/s, prolly 4MB/s for optimized runs.

mmis1000:
圖片
I actually test for it a bit early today.

But the result it weird.

Profiler shows that most time was used on the `_nextCell` itself (the self is 81.3ms).
Even the complex check at line end use less time than it. (the 'self' is 1.3ms)
But the `_nextCell` is just a bunch of if else used to update current cursor style/position or `_currentRow`.

The only object it create is a array to hold sgr sequence which isn't different from the original code.
Do access `this` itself enough to slow down the v8?
Or due to string concatenation?

jerch:
@mmis1000 Yes profiling JS is abit cumbersome, as all profilers I know do only sampling (snapshotting into callstack from time to time). Furthermore the JS engine might decide to inline hot functions, thus you would never see them being called, instead they add on top of the caller's time. The latter gets really annoying if you try to optimize hot paths, as they simply wont show up in the profiler records anymore.

Hand-waving rules I found to be working with V8:
- avoid object re-creation at any cost, instead resort to re-using previous ones (otherwise GC will kill perf)
- never change object shape by adding or removing properties (it will de-opt the whole function forever)
- deep nested branching is bad (similar as in C), this actually applies here, as a good JS engine will try to JIT hot functions
- Slicing logic into several sub functions is good, this gives the engine the freedom to decide, whether that particular logic is hot and should be inlined.

Inlining on its own is a tough topic and might be handled very different by various engines/optimizers. I only tested this with V8/crankshaft years back, but my guess is, that most ideas still apply to newer optimizers as well. For stable inlining it is important, that the arguments and the return values never change the types (e.g. if you expect a string as argument it is better to deal with an empty string instead of `undefined`, `null`, whatsoever). Ideally those functions are real functions in a functional programming sense (re-entrant), otherwise the optimizer might bail out later on for non obvious reasons. This can be achieved by a simple rule of thumb - calculate your function result only from arguments to avoid side effects while keeping the return type stable. Furthermore the function to be inlined should not be overly long in code size, otherwise the optimizer would not try to inline it at all (was about ~600 chars back then).

About string creation/concat:
Yes thats hard to do for JS. What looks nice and easy on code level, is a tough thing to achieve for the engine underneath in a speedy manner. Esp. there is a big difference between engines in loop-concat like `for (c in arr) res += c;` and array string joining as `arr.join('')`. The latter is much faster in V8, while loop-concat sucks big time. For Firefox it is the opposite where the loop is twice as fast compared to array join suggesting they have a special optimization for that.
I did not dig much further into this, as I mostly tested on V8, thus went with array join most of the time. Overall the array join is nearly the same on both, thus gives the most reliable results without penalizing one engine too much. It might be possible to speed this up further by collecting codepoints first and postponing the real string creation (idk, did not test that, also I'd see this as a premature optimization, as I would not expect much of a speed gain here).

jerch:
@mmis1000 Is this in a review state yet or do you need some more bits/info to be addressed?

mmis1000:
@jerch This is already in a review state. The part I mostly concern about is performance, but that part of code (create a array and join) is already existed current code base. So I think it is better to left it for future pr instead of this.

Tyriar:
Hoping to do a release next week, I'll try review before then.

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -167,20 +373,53 @@ export class SerializeAddon implements ITerminalAddon {
...
+
+      lines.push(JSON.stringify(bufferLine));
+    }
 
-    return handler.serialize(maxRows - rows, maxRows);
+    return {
+      x: buffer.cursorX,
+      y: buffer.cursorY,
+      data: lines
+    };
```
Tyriar, line 422: Can this util be moved into the test file?

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -71,35 +72,146 @@ function equalFlags(cell1: IBufferCell, cell2: IBufferCell): boolean {
...
+    }
+
+    this._allRows[this._rowIndex] = this._currentRow;
+    this._allRowSeparators[this._rowIndex++] = rowSeparator;
     this._currentRow = '';
     this._nullCellCount = 0;
   }
 
-  protected _nextCell(cell: IBufferCell, oldCell: IBufferCell, row: number, col: number): void {
+  private _diffStyle (cell: IBufferCell, oldCell: IBufferCell): number[] {
```
Tyriar, line 204:
```suggestion
  private _diffStyle(cell: IBufferCell, oldCell: IBufferCell): number[] {
```

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -131,30 +243,124 @@ class StringSerializeHandler extends BaseSerializeHandler {
...
+      if (i + 1 < rowEnd) {
+        content += this._allRowSeparators[i];
+      }
+    }
+
+    // restore the cursor
+    const realCursorRow = this._buffer1.baseY + this._buffer1.cursorY;
+    const realCursorCol = this._buffer1.cursorX;
+
+    const cursorMoved =  (realCursorRow !== this._lastCursorRow || realCursorCol !== this._lastCursorCol);
```
Tyriar, line 340:
```suggestion
    const cursorMoved = (realCursorRow !== this._lastCursorRow || realCursorCol !== this._lastCursorCol);
```

addons/xterm-addon-serialize/src/MyBufferCell.ts
```diff
@@ -0,0 +1,96 @@
+import { IBufferCell } from 'xterm';
+
+/**
+ * This is a dummy buffer cell to hold data from real buffer cell
+ */
+export class MyBufferCell implements IBufferCell {
+  constructor (private _cell: IBufferCell) {}
```
Tyriar, line 7:
```suggestion
  constructor(private _cell: IBufferCell) {}
```

addons/xterm-addon-serialize/src/MyBufferCell.ts
```diff
@@ -0,0 +1,96 @@
+import { IBufferCell } from 'xterm';
+
+/**
+ * This is a dummy buffer cell to hold data from real buffer cell
+ */
+export class MyBufferCell implements IBufferCell {
```
Tyriar, line 6: It's not clear what the purpose of this class is? It kind of looks like an adapter but it just exposes the same interface `IBufferCell`, any reason the addon doesn't just use that?

addons/xterm-addon-serialize/src/MyBufferCell.ts
```diff
@@ -0,0 +1,96 @@
+import { IBufferCell } from 'xterm';
```
Tyriar, line 1: Let's add the copyright notice

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -71,35 +72,146 @@ function equalFlags(cell1: IBufferCell, cell2: IBufferCell): boolean {
...
+  // this is a null cell for reference for checking whether background is empty or not
+  private _backgroundCell: MyBufferCell = MyBufferCell.from(this._cursorStyle);
+
+  private _firstRow: number = 0;
+  private _lastCursorRow: number = 0;
+  private _lastCursorCol: number = 0;
+  private _lastContentCursorRow: number = 0;
+  private _lastContentCursorCol: number = 0;
+
+  constructor(private _buffer1: IBuffer,private _terminal: Terminal) {
```
Tyriar, line 98:
```suggestion
  constructor(private _buffer1: IBuffer, private _terminal: Terminal) {
```

Tyriar (changes requested):
Looks good and seems to work, a few comments. Sorry about the delay, I was out of work for a bit.

addons/xterm-addon-serialize/src/MyBufferCell.ts
mmis1000: The only reason for a separate class is because in line it directly will be messy, there are too many fields and methods.

It can be also refactored to a function that returns IBufferCell objects.

addons/xterm-addon-serialize/src/SerializeAddon.ts
mmis1000: I am not really sure. 
Because this function need to be called in the browser context.
Is there a way to side loading script into browser context of tests?

addons/xterm-addon-serialize/src/MyBufferCell.ts
mmis1000: Remove it in favor of just remember where was the _cursorStyle comes from

addons/xterm-addon-serialize/src/SerializeAddon.ts
mmis1000: Figured out how to move it into separate file in test

addons/xterm-addon-serialize/src/MyBufferCell.ts
mmis1000: File no longer exist

mmis1000:
@Tyriar 
I removed the separate dummy buffer cell in favor of just remember where did the cursor style comes from.
And also, the test utility is moved into separate file. (Is it proper to put it in the test directory?)

I am not sure whether there is some edge case that I didn't handle correctly.
Probably it needs a fuzz test someday to check whether it will choke on some specified pattern.

Tyriar:
@mmis1000 busy with VS Code endgame, I plan on getting to this next week.

Tyriar (approved):
Thanks @mmis1000, this works great. For the test change I moved it into `page.evaluate`, I mainly just wanted it inside the test/ folder

cyrus-and:
Hi, sorry for bumping an old thread but, how does the cursor restoration work? I mean, it *just* works or do I have to enable it somehow? I don't see any difference in my use case (except for the background restoration part, which works as expected).

cyrus-and:
Never mind, it was an issue on my side. Sorry for the noise.

PR #3114, opened 2020-09-27
marker fixes

### LINKED ISSUE #3108: Clarification about markers
Just tried to use markers for the image addon to get informed, when it is safe to free resources again. I am not quite sure how markers are supposed to be used via API. Is there a reason why `onDispose` is not exported? Is it unsafe to rely on it?

Due to this I did a quick hack with a marker collector, which checks for `isDisposed` and `line` and found that the line is still decrementing after the marker is meant to be disposed:

grafik

(The 'got disposed!' message is from my type overloading with a `onDispose` callback.)

This somewhat looks like it was not really disposed, as some code from xterm.js internals updates the now illegal line index (thus still holds a ref somehow)? Imho related code would be here: https://github.com/xtermjs/xterm.js/blob/a5e269a3f9ecdbaa240946e8724e1da9661a696d/src/common/buffer/Buffer.ts#L582-L614

@Tyriar Does this work as intended (bear with me, there is a high chance I simply dont get it :smile_cat:)

jerch:
Shall fix #3108.

Changes:
- calls `super.dispose()` in overloaded method
- `onDispose` added to public API


jerch:
@Tyriar I think this should get a few unit cases before it can be merged. Currently I am not sure, what action will trigger `onDispose` (only by scrolling out? line removal from buffer? before resize? during resize? after resize as batch?) Maybe some test cases will reveal the details and we might have to see whether it fits the bill.

Tyriar:
@jerch sounds good, I think the only cases are:

- Scroll off top (trim)
- Delete line (in viewport via sequence)
- Embedder disposing

Tyriar (approved):
Good to merge now or when tests are added

jerch:
@Tyriar Have implemented basic marker test cases for these actions:
- `onDispose` get called - in `Buffer.test.ts` (working)
- dispose on trim off the buffer top in `Terminal.test.js` (working)
- dispose on DL `Terminal.test.js` (skipped, not working)
- dispose on IL `Terminal.test.js` (skipped, not working)

There are multiple issues with the tests - first I tried to get those test cases running on `common/CoreTerminal`, but this does not work for some reason. Seems there is still logic bound to `browser/Terminal`, which better would reside in `CoreTerminal`. In regards of buffer insert/update logic `CoreTerminal`is currently dysfunctional (does not do any circular buffer rotation).

`DL` and `IL` do not trigger any marker updates (no line updates, no dispose). This looks like a bug to me, as it does not replay the lines changes in markers, thus after doing those actions markers simply point to the wrong line. Guess markers need to be abit more clever than just listen to `onTrim`?

Imho we can still add the fixes here, and maybe address the issues above with later PRs?

jerch:
@Tyriar Got `DL` and `IL` working by adding the insert and delete events in `CircularList.splice`. Plz have a look on those additions (hope they are free of any side effects).

jerch:
@Tyriar Thinking about the interface - could we make `onDispose` to return the last active line before it turned to -1? This would help to track area updates (needed to know whether a line was removed from top or within the cursor active area).

Somthing like this:
```typescript
interface IMarker {
  ...
  onDispose: IEvent<number>;
}
...
marker.onDispose(lastLine => {
  // lastLine holds idx of last line 
});
```
Would only need a tiny change here: https://github.com/xtermjs/xterm.js/blob/faa60be6fbffab39914254ea94b04bb484095964/src/common/buffer/Marker.ts#L32-L34

Tyriar:
Thanks for merging and sorry about the disappearance 🙂

jerch:
@Tyriar Ah no worries, I think the added events in `CircularList` do the right thing (as the tests work now). Since vscode uses markers to some extend I just think you might want to check if their logic still sees the right line numbers after a DL/IL.

PR #3133, opened 2020-10-19
Using putout linter on xterm codebase part 2

coderaiser:
Latest PR (https://github.com/xtermjs/xterm.js/pull/2953) was successfully merged, so I decide to do more checks with updated version of [putout](https://github.com/coderaiser/putout) and discuss improving codebase with rules approved earlier and I can disable any of rules that not fits to current code style :).

So here is config:

```json
{
    "rules": {
        "strict-mode": "off",
        "remove-useless-spread/object": "off",
        "remove-iife": "off",
        "convert-generic-to-shorthand": "off",
        "remove-console": "off",
        "apply-destructuring": "off",
        "merge-if-statements": "off",
        "convert-math-pow": "off",
        "remove-empty": "on",
        "convert-index-of-to-includes": "on",
        "convert-for-each-to-for-of": "on",
        "apply-shorthand-properties": "on",
        "apply-numeric-separators": "off"
    },
}

### DIFF
--- a/src/common/Platform.ts
+++ b/src/common/Platform.ts
@@ -17,7 +17,7 @@ const isNode = (typeof navigator === 'undefined') ? true : false;
 const userAgent = (isNode) ? 'node' : navigator.userAgent;
 const platform = (isNode) ? 'node' : navigator.platform;
 
-export const isFirefox = !!~userAgent.indexOf('Firefox');
+export const isFirefox = userAgent.includes('Firefox');
 export const isSafari = /^((?!chrome|android).)*safari/i.test(userAgent);
 
 // Find the users platform. We use this to interpret the meta key
--- a/src/common/tsconfig.json
+++ b/src/common/tsconfig.json
@@ -2,7 +2,8 @@
   "extends": "../tsconfig-library-base",
   "compilerOptions": {
     "lib": [
-      "es2017"
+      "es2015",
+      "es2016.Array.Include"
     ],
     "outDir": "../../out",
     "types": [
--- a/src/common/tsconfig.json
+++ b/src/common/tsconfig.json
@@ -2,7 +2,7 @@
   "extends": "../tsconfig-library-base",
   "compilerOptions": {
     "lib": [
-      "es2015"
+      "es2017"
     ],
     "outDir": "../../out",
     "types": [
--- a/bin/extract_vtfeatures.js
+++ b/bin/extract_vtfeatures.js
@@ -359,7 +359,7 @@ function empty(ar) {
 }
 
 function* parseMultiLineGen(filename, s) {
-  if (!~s.indexOf('@vt:')) {
+  if (!s.includes('@vt:')) {
     return;
   }
   const lines = s.split('\n').map(el => el.trim().replace(/[*]/, '').replace(/\s/, ''));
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

File: src/browser/Terminal.test.ts [hunk 1]
```diff
@@ -717,7 +717,7 @@ describe('Terminal', () => {
 
   describe('unicode - surrogates', () => {
     it('2 characters per cell', function (): void {
-      this.timeout(10000);  // This is needed because istanbul patches code and slows it down
+      this.timeout(10_000);  // This is needed because istanbul patches code and slows it down
```
jerch, line 720: I think this notation is still way too new and we should not yet rely on it. Same goes for the other occurrences.

src/common/Platform.ts
```diff
@@ -17,7 +17,7 @@ const isNode = (typeof navigator === 'undefined') ? true : false;
 const userAgent = (isNode) ? 'node' : navigator.userAgent;
 const platform = (isNode) ? 'node' : navigator.platform;
 
-export const isFirefox = !!~userAgent.indexOf('Firefox');
+export const isFirefox = !!userAgent.includes('Firefox');
```
jerch, line 20: Since `.includes` already returns a boolean, the `!!` type conversion gets redundant.

src/common/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../tsconfig-library-base",
   "compilerOptions": {
     "lib": [
-      "es2015"
+      "es2017"
```
jerch, line 5: Whats the reason for ES2017? I have no strong obligations against ES2017, but would not like to change that for a non-obvious reason. Being more vanilla/old-school is not a bad thing in terms of supported targets while maintaining the same functionality.

jerch (changes requested):
Thx for looking into that - looks good to me for most changes, but a few remarks below.

src/common/tsconfig.json
coderaiser: The thing is `es2017` needs to be set to have ability to build files that contains `includes` with `typescript`.

jerch: @coderaiser :+1:  Ah ok, well I'm fine to move on to ES2017 then, but would like to hear @Tyriar and @mofux 's opinion in this regard first.

Tyriar: My only thought is if we commit to this, that we also update all the other tsconfig.json files so they're not inconsistent. Since we only target evergreen browsers, I think this is safe to do?

Tyriar (changes requested):

src/common/tsconfig.json
jerch: @Tyriar I think it will change certain output shims. Imho our codebase is not really affected by that (Was it around generators? Or promises with async/await? Cant remember...)

Tyriar: Well that's for the target, the lib is just which typings to include. I see there is a lib `"ES2016.Array.Include"`, that sounds like exactly what we want without needing to think about changing es2015.

coderaiser: @Tyriar, I updated `browser/tsconfig.json`. Should I update`tsconfig.json` in addons directories? They don't have lib option right now.

Tyriar: @coderaiser no need if they don't use lib, does `ES2016.Array.Include` work instead of `es2017`?

coderaiser: @Tyriar yes, it works, just pushed

Tyriar (approved):
Thanks for the clean up 🙂

PR #3134, opened 2020-10-19
feat(build): allow require from NodeJS

redallen:
Currently when requiring `lib/xterm.js` from NodeJS you'll see this:
```
UnhandledPromiseRejectionWarning: ReferenceError: window is not defined:
at xterm.js:1 (xterm.js:1:178)
```

This fixes that issue by defining the module on self instead of window for NodeJS environments.

### DIFF
--- a/addons/xterm-addon-attach/webpack.config.js
+++ b/addons/xterm-addon-attach/webpack.config.js
@@ -25,7 +25,8 @@ module.exports = {
     filename: mainFile,
     path: path.resolve('./lib'),
     library: addonName,
-    libraryTarget: 'umd'
+    libraryTarget: 'umd',
+    globalObject: 'typeof self !== \'undefined\' ? self : this'
   },
   mode: 'production'
 };
--- a/addons/xterm-addon-fit/webpack.config.js
+++ b/addons/xterm-addon-fit/webpack.config.js
@@ -25,7 +25,8 @@ module.exports = {
     filename: mainFile,
     path: path.resolve('./lib'),
     library: addonName,
-    libraryTarget: 'umd'
+    libraryTarget: 'umd',
+    globalObject: 'typeof self !== \'undefined\' ? self : this'
   },
   mode: 'production'
 };
--- a/addons/xterm-addon-ligatures/webpack.config.js
+++ b/addons/xterm-addon-ligatures/webpack.config.js
@@ -26,7 +26,8 @@ module.exports = {
     filename: mainFile,
     path: path.resolve('./lib'),
     library: addonName,
-    libraryTarget: 'umd'
+    libraryTarget: 'umd',
+    globalObject: 'typeof self !== \'undefined\' ? self : this'
   },
   mode: 'production',
   externals: {
--- a/addons/xterm-addon-search/webpack.config.js
+++ b/addons/xterm-addon-search/webpack.config.js
@@ -25,7 +25,8 @@ module.exports = {
     filename: mainFile,
     path: path.resolve('./lib'),
     library: addonName,
-    libraryTarget: 'umd'
+    libraryTarget: 'umd',
+    globalObject: 'typeof self !== \'undefined\' ? self : this'
   },
   mode: 'production'
 };
--- a/addons/xterm-addon-serialize/webpack.config.js
+++ b/addons/xterm-addon-serialize/webpack.config.js
@@ -25,7 +25,8 @@ module.exports = {
     filename: mainFile,
     path: path.resolve('./lib'),
     library: addonName,
-    libraryTarget: 'umd'
+    libraryTarget: 'umd',
+    globalObject: 'typeof self !== \'undefined\' ? self : this'
   },
   mode: 'production'
 };
--- a/addons/xterm-addon-unicode11/webpack.config.js
+++ b/addons/xterm-addon-unicode11/webpack.config.js
@@ -32,7 +32,8 @@ module.exports = {
     filename: mainFile,
     path: path.resolve('./lib'),
     library: addonName,
-    libraryTarget: 'umd'
+    libraryTarget: 'umd',
+    globalObject: 'typeof self !== \'undefined\' ? self : this'
   },
   mode: 'production'
 };
--- a/addons/xterm-addon-web-links/webpack.config.js
+++ b/addons/xterm-addon-web-links/webpack.config.js
@@ -25,7 +25,8 @@ module.exports = {
     filename: mainFile,
     path: path.resolve('./lib'),
     library: addonName,
-    libraryTarget: 'umd'
+    libraryTarget: 'umd',
+    globalObject: 'typeof self !== \'undefined\' ? self : this'
   },
   mode: 'production'
 };
--- a/addons/xterm-addon-webgl/webpack.config.js
+++ b/addons/xterm-addon-webgl/webpack.config.js
@@ -33,7 +33,8 @@ module.exports = {
     filename: mainFile,
     path: path.resolve('./lib'),
     library: addonName,
-    libraryTarget: 'umd'
+    libraryTarget: 'umd',
+    globalObject: 'typeof self !== \'undefined\' ? self : this'
   },
   mode: 'production'
 };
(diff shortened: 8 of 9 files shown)


Tyriar:
This feels a little hacky as xterm.js isn't meant to run in node.js currently, what's your use case?

Note also that there is an issue to do a proper node target here: https://github.com/xtermjs/xterm.js/issues/2749

redallen:
@Tyriar My use case is server side rendering a terminal which we've adopted into our component library PatternFly.

Tyriar:
@redallen how does this work in node though; window and the DOM is critical to the rendering part of xterm.js, no?

redallen:
DOM is not critical to SSR (`React.renderToString()` works just fine on React component), and neither is add event listeners on `window` or `document` which is done on hydrate rather than when prerendering some HTML.

Tyriar:
But wouldn't xterm.js throw if you're running it? Sorry if I'm a bit slow but I haven't used React. I'm not understanding what prerendering would do for xterm.js since the lib must run with access to the DOM and why prerendering would care about what is in a script at all?

redallen:
TLDR; We want this to work from a node context to render some HTML but not actually run anything:
```js
const { Terminal } = require('xterm');
const { FitAddon } = require('xterm-fit-addon');
const myTerminal = Terminal();
myTerminal.onData(someFn);
myTerminal.onTitleChange(someFn);
myTerminal.loadAddon(new FitAddon());
myTerminal.open(divRef);
```

The story behind SSR is that we wrote a React app for our docs site https://patternfly.org. However, since React is a SPA we had to ship each page with an empty `<div id="root />`. This is bad for SEO, web crawling, initial page loads, and older devices that don't load JS quickly.s o we decided to use SSR which puts an initial rendering of all our components. Using it looks something like this `<div id="root">{React.renderToString(<OurSPA />)}</div>`.

We want `<OurSPA />` to include a new component we're developing called `XTerm`. You can see @KKoukiou 's implementation of it [here.](https://github.com/patternfly/patternfly-react/blob/dce77b97c688d32c42d6e5c15d203f2d1146c15e/packages/react-console/src/components/SerialConsole/XTerm.tsx) That means that the above code snippet to render some HTML into it when we call `.open(divRef)` even if there isn't a `window` object. As far as I can tell, this PR allows us to do that.

Tyriar:
@redallen but the contents of xterm.js does nothing for SEO, isn't this all overkill? I would expect a correctly "server side rendered" version of xterm.js to just look like the below.

```
<div id="xterm-container"></div>
<script src="xterm.js"></script>
<link rel="stylesheet" href="xterm.css">
```

The behavior of creating any more elements inside `#xterm-container` and then running xterm.js is undefined and I would expect it to break things.

redallen:
@Tyriar that's completely fine if `<div id="xterm-container"></div>` is the expected server side rendering, and I don't think it's overkill to want to use SSR on our XTerm documentation page (we document more than just examples of the component). Our problem and the reason for this PR is we can't server side render the page at all because as soon as we `require('xterm.js')` we get `ReferenceError: window is not defined: at xterm.js:1 (xterm.js:1:178)` because the global object is `window` rather than `self`. This PR is meant to address that issue.

Tyriar:
But you wouldn't want to `require` it though, it should only ever run on the client not the server? Does React server side rendering really want to run all client side scripts within node, does that mean any JS library you use on the client must be able to run without throwing in a node context?

My impression of this PR is that it's a hack to get React SSR to work when really we don't ever want people to `require('xterm')` within Node.js. Even when we have a real node target (that will not include the renderer), I don't think that's what you would want either as you need the actual version in order to render on the client?

Could you do something on your end to prevent `require('xterm')` from running in node but allow it on the client? eg.

```ts
let xterm;
if (isWeb) {
  xterm = require('xterm');
}
```

redallen:
> But you wouldn't want to `require` it though, it should only ever run on the client not the server?

You'd want to `require` it to get an initial rendering for apps that use SSR.

> Does React server side rendering really want to run all client side scripts within node, does that mean any JS library you use on the client must be able to run without throwing in a node context?

Yes, it really wants to run all client side scripts within node. Node libraries can work around this via checking `typeof window !== 'undefined'` or via webpack's [DefinePlugin](https://webpack.js.org/plugins/define-plugin/).

> My impression of this PR is that it's a hack to get React SSR to work when really we don't ever want people to `require('xterm')` within Node.js. Even when we have a real node target (that will not include the renderer), I don't think that's what you would want either as you need the actual version in order to render on the client?

I don't know of another way of doing the initial render of `xterm`'s HTML in a node context other than `require('xterm')`. Again, I'm not looking for any functionality, just some skeleton HTML to later call `myTerminal.open(divRef)` on.

> Could you do something on your end to prevent `require('xterm')` from running in node but allow it on the client? eg.
> 
> ```ts
> let xterm;
> if (isWeb) {
>   xterm = require('xterm');
> }
> ```

Yes, although it's not as simple as this if statement because `webpack` will ignore the if statement and bundle `xterm` anyways. We could use a webpack alias for our SSR to replace `xterm` with our own version that works. We have not had to do this for any other 3rd party library we use and we'd like to avoid it if possible.

Tyriar:
> I don't know of another way of doing the initial render of xterm's HTML in a node context other than require('xterm'). Again, I'm not looking for any functionality, just some skeleton HTML to later call myTerminal.open(divRef) on *in a real browser.

But the skeleton HTML you should be sending is just whatever element you call open on that you must create before calling `open`? It's a client lib and `open` will only work if you call it on the browser. If you call `open` on node and SSR creates a bunch of elements on the server, these elements (various divs, canvas, textarea, etc.) *will not* be tracked in the client and duplicates will be created.

redallen:
Double renders are fine and we can easily fix them if that's a problem. React is usually smart about hydrating and will remove or add event listeners when the server DOM does not match the client DOM.

Again, I'd just like to be able to call these functions that don't do anything server side or might render something into `divRef` -- that part is totally up to you:
```js
const { Terminal } = require('xterm');
const { FitAddon } = require('xterm-fit-addon');
const myTerminal = Terminal();
myTerminal.onData(someFn);
myTerminal.onTitleChange(someFn);
myTerminal.loadAddon(new FitAddon());
myTerminal.open(divRef);
```

Otherwise I have to implement my own
```js
xterm.js

module.exports = {
  Terminal: () => ({
    onData: () => {},
    onTitleChange: () => {},
    loadAddon: () => {},
    open: () => {},
  })
}
```

```js
xterm-fit-addon.js

module.exports = {
  FitAddon: class FitAddon() {}
}
```

and keep my noop implementation up-to-date with your API.

Tyriar:
I appreciate the PR but it seems like the noop implementation is the right way to go here since calling open on the server side would result in weirdness on the client if those elements get sent over. Just only add calls on the public API and it should be trivial to update, just checking whenever you jump major versions which are typically > 1 year apart, you could also potentially auto generate it by iterating over `Object.keys(term)` or something)

kena0ki:
My understanding.(Sorry if I'm wrong.)

I don't think `new Terminal()` is not necessary for SSR, and this PR doesn't fix the problem if it is. I suppose `new Terminal()` should be in `componentDidMount()` instead of `constructor()`.

However it's true that `require('xterm')` is necessary for SSR unless using workaround like the above, or [code splitting](https://reactjs.org/docs/code-splitting.html) and it's inconvenient for React users using SSR. (Although the issue seems to be essentially [caused by Webpack](https://github.com/webpack/webpack/issues/6642).)

redallen:
Yes that is correct @kena0ki , this PR is the minimal to just get `xterm.js` to build in an SSR context. The Webpack 4 behavior is not interoperable but I believe this has been fixed in Webpack 5.

Would you merge a PR to upgrade to Webpack 5 @Tyriar ?

Tyriar:
I just merged it yesterday, you could try the beta build (`xterm@beta`) which should have been done with webpack 5.

t0ster:
Just FYI, the issue is still present. I am trying to use next.js with [rendition](https://github.com/balena-io-modules/rendition) which uses xterm.js. And I have to fork xterm.js and apply the same `globalObject: 'typeof self !== \'undefined\' ? self : this'` workaround as @redallen.

PR #3136, opened 2020-10-21
Fix "term.fit is not a function" error when change padding value in the demo

kena0ki:
When we change a padding value in the demo, a console error occurs and xterm is not correctly fitted.
termfit

I replaced function call from term.fit() to addons.fit.instance.fit().


Tyriar (approved):
Thanks!

PR #3141, opened 2020-10-23
make contributions easier for beginners via the online automated dev setup.

nisarhassan12:
Hi. 

I work for Gitpod and both Gitpod and Theia love `xterm.js` and use it extensively.

This PR adds gitpod config to the repo to make contributions easier for newcomers i.e with a single click it will launch a workspace and automatically:

- clone the `xterm.js` repo.
- install the dependencies.
- run `yarn build` and `yarn start`.

This can be helpful for beginners i.e they can start coding instantly without spending any time on setting anything up.

You can give it a try on my fork of the repo via the following link:

https://gitpod.io/#https://github.com/nisarhassan12/xterm.js

This is how it looks:

image

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,6 @@
+tasks:
+  - init: yarn install && yarn run build
+    command: yarn run start
+ports:
+  - port: 3000
+    onOpen: open-preview
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # [![xterm.js logo](logo-full.png)](https://xtermjs.org)
 
-[![Build Status](https://dev.azure.com/xtermjs/xterm.js/_apis/build/status/xtermjs.xterm.js)](https://dev.azure.com/xtermjs/xterm.js/_build/latest?definitionId=3)
+[![Build Status](https://dev.azure.com/xtermjs/xterm.js/_apis/build/status/xtermjs.xterm.js)](https://dev.azure.com/xtermjs/xterm.js/_build/latest?definitionId=3) [![Gitpod Ready-to-Code](https://img.shields.io/badge/Gitpod-Ready--to--Code-blue?logo=gitpod)](https://gitpod.io/from-referrer/)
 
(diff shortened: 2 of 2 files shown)


jerch:
@nisarhassan12 I like what your team has created with theia/gitpod, still I have concerns about adding this to the xterm.js repo. Reasons:

- Its not a genuine content thing, it is helpful infrastructure. If it should be integrated, than on github's side (or any other tooling ppl use to work with this repo).
- Too much of viral marketing for my taste, ask yourself - should we also add quickstart badges for those other 10+ cloud dev envs? 
- Yes we need devs, but not beginners/newcomers. We have tons of bugs and missing features, most of them are hard to comprehend for beginners. If you love xterm.js and use it extensively - how about contributing yourself? Low hanging fruits are good bug descriptions or input from a greater user base to get future directions sorted out. PRs are more than welcome.
- There is a cost trap: Only 50 hours a month are free (thats what your price plans tell). And here is the catch - xterm.js is a crucial component in your platform. So basically you want us to pay for making your product better? This really raises questions about the MIT license in general. And I see that you released gitpod under AGPL with downstripped functionality. Should we do the same with xterm.js?

Also would like to hear other opinions.

FYI: @Tyriar, @mofux

nisarhassan12:
Thanks @mofux.

Sorry for the pricing page not being clear enough. if you do get near the 50 hour monthly limit, we currently lift it for professional open source developers with the Professional Open Source plan (see https://www.gitpod.io/docs/professional-open-source/) i.e where anyone can work for any number of hours on any public repo and there is no cost associated to it at all.

Tyriar:
Hi @nisarhassan12, I see https://github.com/gitpod-io/definitely-gp exists, so that seems like an ideal frictionless place to put this. We've been burned in the past by checking things in that we cannot maintain and would prefer not to get into the same situation. I'd also like to avoid additional "clutter" in the root of the repo and the single badge in the README was very intentional 🙂

PR #3156, opened 2020-10-29
Update CoreBrowserService.ts

shixiaobao17145:
Hi, 
Thanks for making such an awesome tool to display terminal on web.
This change is to fix focus state issue for wrapping xterm in the web component.

When wrapping Xterm in shadow DOM, the document.activeElement will always be the root of the web component, which will make the isFocus method return the incorrect value. 
FYI:https://dev.to/open-wc/mind-the-document-activeelement-2o9a

### DIFF
--- a/src/browser/services/CoreBrowserService.ts
+++ b/src/browser/services/CoreBrowserService.ts
@@ -14,6 +14,6 @@ export class CoreBrowserService implements ICoreBrowserService {
   }
 
   public get isFocused(): boolean {
-    return document.activeElement === this._textarea && document.hasFocus();
+    return (document.activeElement===this._textarea || (document.activeElement.shadowRoot && document.activeElement.shadowRoot.activeElement===this._textarea)) && document.hasFocus();
   }
 }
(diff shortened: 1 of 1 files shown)


src/browser/services/CoreBrowserService.ts
```diff
@@ -14,6 +14,6 @@ export class CoreBrowserService implements ICoreBrowserService {
   }
 
   public get isFocused(): boolean {
-    return document.activeElement === this._textarea && document.hasFocus();
+    return (document.activeElement===this._textarea || (document.activeElement.shadowRoot && document.activeElement.shadowRoot.activeElement===this._textarea)) && document.hasFocus();
```
Tyriar, line 17: Seems reasonable. Here's an improvement which should be functionally equivalent:

```suggestion
    return (document.activeElement===this._textarea || document.activeElement.shadowRoot?.activeElement===this._textarea) && document.hasFocus();
```

Tyriar (changes requested):

Tyriar (approved):

Tyriar:
I think there was already a change that did this since this PR was made, correct me if I'm wrong 🙂

PR #3169, opened 2020-11-08
Use RenderService.dimensions instead of CharSizeService for textarea position

### LINKED ISSUE #3034: textarea position incorrect with custom font size
Hidden textarea `xterm-helper-textarea` position computed incorrectly from **js** code when custom font size is used `terminal.setOption("fontSize", 16)`

image

This makes **the whole xterm view scroll to the right** cutting left edge and exposing unnecessary space on the right, when cursor is close to the right edge. The terminal should also occupy the whole window, I guess, for this to happen.

This is even reproducible on [Hyper](https://hyper.is/):

image

The easiest way to reproduce is to keep pasting two character string e.g. `"ed"` into the terminal `Ctrl+Shift+V` up to the right border. In fact, the issue only happens when pasting, regular typing won't give the effect.

the `xterm.css` initializes `textarea` width with 0:

```
.xterm .xterm-helper-textarea {
    ...
    width: 0;
    height: 0;
    ...
}
```
but it gets recomputed from **js** later on.

On **vscode** this does not occur though, they probably override `textarea` style in custom way. 

## Details
- Browser and browser version: **Chrome**
- OS version:  **Windows 10**
- xterm.js version: **4.8.1**
- rendererType: **canvas**

### Steps to reproduce

1. Download [Hyper](https://hyper.is/)
2. Keep on pasting (`Ctrl+Shift+V`) "ed" into the terminal

kena0ki:
Fixes #3034

- Before
Kapture 2020-11-09 at 0 44 17

- After
Kapture 2020-11-09 at 0 31 11

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -297,19 +297,23 @@ export class Terminal extends CoreTerminal implements ITerminal {
   }
 
   private _syncTextArea(): void {
-    if (!this.textarea || !this.buffer.isCursorInViewport || this._compositionHelper!.isComposing) {
+    if (!this.textarea || !this.buffer.isCursorInViewport || this._compositionHelper!.isComposing || !this._renderService) {
       return;
     }
-
-    const cellHeight = Math.ceil(this._charSizeService!.height * this.optionsService.options.lineHeight);
-    const cursorTop = this._bufferService.buffer.y * cellHeight;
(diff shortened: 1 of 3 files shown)


kena0ki:
I thought CI failure is my fault, but it wasn't.

Tyriar (approved):
Thanks for looking into this, works great just made a few polish tweaks.

kena0ki:
Thank you for merging.

PR #3170, opened 2020-11-09
Incorporate shadowRoot in focus calculation.

blois:
This adds a call to getRootNode to get either the Document or the ShadowRoot to use when calculating whether the textarea has focus. This allows the cursor to render as focused when the terminal is rendered within a shadow DOM.

getRootNode is guarded for older browser compatibility (IE).


Tyriar (approved):
Thanks!

PR #3172, opened 2020-11-10
Fix a11y violation because of using role='listitem' without role='list' parent

KKoukiou:
This breaks our accessibility tests:
See:
image

### DIFF
--- a/src/browser/AccessibilityManager.ts
+++ b/src/browser/AccessibilityManager.ts
@@ -56,6 +56,7 @@ export class AccessibilityManager extends Disposable {
     this._accessibilityTreeRoot.classList.add('xterm-accessibility');
 
     this._rowContainer = document.createElement('div');
+    this._rowContainer.setAttribute('role', 'list');
     this._rowContainer.classList.add('xterm-accessibility-tree');
     this._rowElements = [];
     for (let i = 0; i < this._terminal.rows; i++) {


Tyriar:
See https://github.com/microsoft/vscode/issues/98918 for context on why it's like this.

These 2 changes were made to try and improve the situation:

- https://github.com/xtermjs/xterm.js/commit/03df0565bda60f4af381083882d3f8fde218e4da#diff-7904a8cc1e295a78532b41608480fddf931f80c485c53b78f99ae8cd5759dedc
- https://github.com/xtermjs/xterm.js/commit/a32776e2af437af8f8172188539bb1dce3a800c3#diff-7904a8cc1e295a78532b41608480fddf931f80c485c53b78f99ae8cd5759dedc

It's not clear to me what the right fix is.

KKoukiou:
https://www.w3.org/TR/wai-aria-1.1/#listitem here it's clear that one should use group or list as parent of listitem. 
Maybe we should reconsider using listitems all in all, and find a more appropriate role for the terminal lines.
I was thinking that https://www.w3.org/TR/wai-aria-1.1/#feed with https://www.w3.org/TR/wai-aria-1.1/#article might be better match here.

Tyriar (approved):
Since this is what it originally was and the changes didn't seem to improve let's move forward with this. I've pulled the feed idea into https://github.com/xtermjs/xterm.js/issues/3173 as it will need a bunch of verification.

KKoukiou:
> https://www.w3.org/TR/wai-aria-1.1/#listitem here it's clear that one should use group or list as parent of listitem.
> Maybe we should reconsider using listitems all in all, and find a more appropriate role for the terminal lines.
> I was thinking that https://www.w3.org/TR/wai-aria-1.1/#feed with https://www.w3.org/TR/wai-aria-1.1/#article might be better match here.

Ignore the above. The comment https://github.com/microsoft/vscode/issues/98918#issuecomment-637401418 encapsulated what is needed here. We need document role wrapper as you correctly used. And the list which contains listitems.
I think this PR generates exactly this output.

Tyriar:
@KKoukiou oh ok, so document > listitem wasn't appropriate but document > list > listitem should work?

KKoukiou:
> @KKoukiou oh ok, so document > listitem wasn't appropriate but document > list > listitem should work?

Exactly. This should be a11y valid and fix the accessibility issues which were mentioned in the above linked issue.

PR #3181, opened 2020-11-21
Make alt+click behavior configurable

### LINKED ISSUE #3145: Option to disable alt-click move cursor to position behavior
Hi all,

I've been running into an issue that I don't see reported yet in this repo, so I'd like to bring it up.

The issue is described here: https://github.com/microsoft/vscode/issues/101136. When alt-clicking in the vscode terminal, if the program doesn't support (something), the screen can fill up with "^[[D" or "^[[C" characters.

Here's a screenshot:
image

In the above vscode issue, @Tyriar linked this PR: https://github.com/xtermjs/xterm.js/pull/896 which seems to have added the alt-click feature. There's apparently no way to disable it, and there's no workaround.

Changing the `"editor.multiCursorModifier"` setting was suggested as a workaround, but it didn't work for me.

Thanks

awmottaz:
Adds a new Boolean option `altClickMovesCursor` which enables/disables the alt+click behavior that moves the prompt cursor to the location of the mouse.

Closes #3145

### DIFF
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -45,6 +45,11 @@ declare module 'xterm' {
     allowTransparency?: boolean;
 
     /**
+     * If enabled, alt + click will move the prompt cursor to position underneath the mouse.
+     */
+    altClickMovesCursor?: boolean;
+
+    /**
      * A data uri of the sound to use for the bell when `bellStyle = 'sound'`.
      */
     bellSound?: string;
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -45,7 +45,8 @@ declare module 'xterm' {
     allowTransparency?: boolean;
 
     /**
-     * If enabled, alt + click will move the prompt cursor to position underneath the mouse.
+     * If enabled, alt + click will move the prompt cursor to position
+     * underneath the mouse. The default is true.
      */
     altClickMovesCursor?: boolean;
 
--- a/src/browser/public/Terminal.ts
+++ b/src/browser/public/Terminal.ts
@@ -170,7 +170,7 @@ export class Terminal implements ITerminalApi {
     this._core.paste(data);
   }
   public getOption(key: 'bellSound' | 'bellStyle' | 'cursorStyle' | 'fontFamily' | 'logLevel' | 'rendererType' | 'termName' | 'wordSeparator'): string;
-  public getOption(key: 'allowTransparency' | 'cancelEvents' | 'convertEol' | 'cursorBlink' | 'disableStdin' | 'macOptionIsMeta' | 'rightClickSelectsWord' | 'popOnBell' | 'visualBell'): boolean;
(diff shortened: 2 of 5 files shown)


awmottaz:
Just a note: I am working on adding a couple of tests:

- alt+click moves the cursor when option is enabled
- alt+click does not move cursor when option is disabled
- option is enabled by default

But I thought I would push up the PR with the implementation right away to get feedback on that.

awmottaz:
I am struggling with writing tests for this. It appears that interaction with the DOM is explicitly disabled in the `TestSelectionService`. Adding it back in causes lots of issues. Any advice for how to write these tests? Or would you consider this contribution without tests?

matthias-ccri:
Just wondering if we want to make this a breaking change. Before, the alt-click behavior was enabled by default. With this change it's disabled by default and must be enabled.

It may be that we should disable it; I don't know, but I thought I'd ask.

awmottaz:
> Just wondering if we want to make this a breaking change. Before, the alt-click behavior was enabled by default. With this change it's disabled by default and must be enabled.

I meant for this not to be a breaking change. I set the default value to `true`. Is there more I need to do to ensure it's enabled by default?

https://github.com/xtermjs/xterm.js/pull/3181/files#diff-d156a9a2b47b63f41faf54f727fc23ffd0ee8fa672cbfe8b77ce20dffa846082R53

mofux (changes requested):
Thanks for the contribution! Looks good so far, IMO no additional tests for this feature are required. There are still some missing bits though:

Please add the new option here:
https://github.com/xtermjs/xterm.js/blob/84fa6be370f9a6b6ced5ece234d2ed1046f0229e/src/browser/public/Terminal.ts#L173

and here:
https://github.com/xtermjs/xterm.js/blob/84fa6be370f9a6b6ced5ece234d2ed1046f0229e/src/browser/public/Terminal.ts#L185

and here:
https://github.com/xtermjs/xterm.js/blob/84fa6be370f9a6b6ced5ece234d2ed1046f0229e/src/common/services/Services.ts#L186

awmottaz:
Thanks @mofux, I made all of your requested changes

matthias-ccri:
@awmottaz I see, my mistake! 
Thanks for the contribution.

typings/xterm.d.ts
```diff
@@ -44,6 +44,11 @@ declare module 'xterm' {
      */
     allowTransparency?: boolean;
 
+    /**
+     * If enabled, alt + click will move the prompt cursor to position underneath the mouse.
+     */
```
Tyriar, line 49:
```suggestion
    /**
     * If enabled, alt + click will move the prompt cursor to position
     * underneath the mouse. The default is true.
     */
```

Tyriar (approved):

Tyriar:
Thanks @awmottaz 🙂

Issue #3189, opened 2020-12-02
Make xterm.js trusted types compliant


jrieken:
We are making VS Code trusted types compliant (see https://github.com/microsoft/vscode/issues/103699). This is to reduce potential XSS attacks and in essence about avoiding certain, dangerous DOM APIs, like assigning strings to innerHTML. 

Searching for `innerHTML` makes me positive about this because not too many results show up. Low hanging/simple "fixes" are using innerText/textContent over innerHTML, see https://github.com/microsoft/vscode/issues/106395#issue-697640147 for more details

PR #3190, opened 2020-12-02
Avoid innerHTML usages

jrieken:
This PR tackles most (all?) innerHTML usages inside xterm.js so that it can be used in a trusted types compliant context. Fixes  https://github.com/xtermjs/xterm.js/issues/3189 - also for more details.

### DIFF
--- a/src/browser/AccessibilityManager.ts
+++ b/src/browser/AccessibilityManager.ts
@@ -263,7 +263,7 @@ export class AccessibilityManager extends Disposable {
       const element = this._rowElements[i];
       if (element) {
         if (lineData.length === 0) {
-          element.innerText = '\u00a0;';
+          element.innerText = '\u00a0';
         } else {
           element.textContent = lineData;
         }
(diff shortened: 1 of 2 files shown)


src/browser/AccessibilityManager.ts
```diff
@@ -263,7 +263,7 @@ export class AccessibilityManager extends Disposable {
       const element = this._rowElements[i];
       if (element) {
         if (lineData.length === 0) {
-          element.innerHTML = '&nbsp;';
+          element.innerText = '\u00a0;';
```
jerch, line 266: Shouldn't this be just `'\u00a0'` without the `;` (as `&nbsp;` is the HTML enitity for that byte value)?

jerch (changes requested):
Thx for looking into this. :+1: 

Imho there is one little bug hidden in the changes (see below).

src/browser/AccessibilityManager.ts
jrieken: Yes - you are right. That was a copy&paste mistake

jrieken:
fixed the bad whitespace replacement

jerch (approved):
LGTM :+1:

PR #3203, opened 2021-01-04
cache buffer to fix VS Code issue #111562

meganrogge:
See microsoft/vscode#111562

### DIFF
--- a/src/browser/public/Terminal.ts
+++ b/src/browser/public/Terminal.ts
@@ -18,6 +18,7 @@ export class Terminal implements ITerminalApi {
   private _core: ITerminal;
   private _addonManager: AddonManager;
   private _parser: IParser | undefined;
+  private _buffer: BufferNamespaceApi | undefined;
 
   constructor(options?: ITerminalOptions) {
     this._core = new TerminalCore(options);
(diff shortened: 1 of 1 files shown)


Tyriar (approved):
👍

src/browser/public/Terminal.ts
```diff
@@ -58,7 +59,10 @@ export class Terminal implements ITerminalApi {
   public get cols(): number { return this._core.cols; }
   public get buffer(): IBufferNamespaceApi {
     this._checkProposedApi();
-    return new BufferNamespaceApi(this._core.buffers);
+    if (!this._buffer) {
+      return new BufferNamespaceApi(this._core.buffers);
+    }
+    return this._buffer;
```
mofux, line 65: I think this fix doesn't work because `this._buffer` is never assigned. I think it should have been:

```typescript
if (!this._buffer) {
  this._buffer = new BufferNamespaceApi(this._core.buffers);
}
return this._buffer;
```

mofux (commented):
Looking over the merged PR I think I found this not to be working as expected. Please see the comment for details.

@meganrogge @Tyriar

src/browser/public/Terminal.ts
meganrogge: yes you're absolutely correct

PR #3204, opened 2021-01-04
set this._buffer to new buffer

meganrogge:


src/browser/public/Terminal.ts
```diff
@@ -60,7 +60,7 @@ export class Terminal implements ITerminalApi {
   public get buffer(): IBufferNamespaceApi {
     this._checkProposedApi();
     if (!this._buffer) {
-      return new BufferNamespaceApi(this._core.buffers);
+      this._buffer = new BufferNamespaceApi(this._core.buffers);
```
jerch, line 63: I think it will not work that - `this._core.buffers` mutates at several occasions in codebase, thus should not be persisted in `BufferNamespaceApi`.

A quickfix would be to change the ctor to init with `this._core`, which is guaranteed to be stable.

Tyriar: Good find, yeah if we could just clear buffers that would be an easy way to fix it.

meganrogge: I think I did what you're suggesting?

jerch: Ah sorry, my description was misleading. I meant to remove `.buffers` from the ctor and just go with `this._core` instead, and to do the property descent to `this._core.buffers` in `BufferNamespaceApi`.

Seems this needs abit rework in the API classes, they are just thin views on the internal buffers, but dont respect ref changes of the internal buffers yet. The problem comes from the static buffer bindings in `BufferNamespaceApi` and `BufferApiView`.

I see two possible solutions:
- ugly way: When accessing `buffers` on the view object, always test for equality of `_core.buffers` to the one stored from ctor time, if different, recreate the subviews and update stored ones. This is ugly as it would keep a ref on the old internal buffer instance (could be avoided with `WeakMap`).
- preferred way (imho): Update all buffer access in `BufferNamespaceApi` and `BufferApiView` to do the full `_core --> buffers --> ...` descent. This way a changing ref at `_core.buffers` needs no special treatment.

Tyriar: @meganrogge let's look at this together today

Tyriar (approved):
Manual test and integration tests passing 🎉 

@jerch we decided to go with the approach of not disposing `BufferSet` at all but calling into the new `BufferSet.reset`, that way we encapsulate the weirdness that could lead to future bugs.

PR #3205, opened 2021-01-05
emit onScroll events when user is scrolling

### LINKED ISSUE #3201: onScroll doesn't emit when user is scrolling
User scrolling is currently [not emitted](https://github.com/xtermjs/xterm.js/blob/master/src/browser/Viewport.ts#L159) as an onScroll event. What's the reason for this? Is this really intended?
I'd like to monitor the scroll state of the terminal in order to notice when it's scrolled all the way down, so I have to notice when the user is scrolling, which is not possible with the current behavior.

## Details
- Browser and browser version: Firefox 84.0.1
- xterm.js version: 4.9.0

### Steps to reproduce

1. attach event listener with `terminal.onScroll()`
2. scroll the terminal using the mouse

schrej:
This changes the onScroll handler of the Viewport to no longer suppress onScroll events when the Viewport is scrolled by the user. I've removed the argument in the Viewport as its no longer needed.

fixes #3201


src/browser/Viewport.ts
```diff
@@ -156,7 +156,7 @@ export class Viewport extends Disposable implements IViewport {
 
     const newRow = Math.round(this._lastScrollTop / this._currentRowHeight);
     const diff = newRow - this._bufferService.buffer.ydisp;
-    this._scrollLines(diff, true);
+    this._scrollLines(diff);
```
Tyriar, line 159: Without the suppress, this ends up calling `Terminal.onScroll` which calls `Viewport.syncScrollArea`. It's needed to avoid the cyclic refreshing there. A new event may be needed.

Tyriar (changes requested):

src/browser/Viewport.ts
Tyriar: To clarify; a new internal event (src/browser/Terminal.ts), not a new API event (src/browser/public/Terminal.ts)

schrej: I've taken another look, you're of course right. This cyclic refreshing then even causes some graphical glitches. A new internal event would be almost identical to the onScroll event though.

I see two options to solve this in a clean way:

- Extend the onScroll event emitter to contain a boolean as it's second value (which is unused right now) to indicate whether the event originated from the Viewport.
- Use a richer `ScrollEvent`s that include information about the source

Both cases will require a second EventEmitter for the public facing onScroll method that "forwards" the events to avoid an API change.

Tyriar: Could you put together the change for the second option? That's probably the best as it's more maintainable.

src/browser/Terminal.ts
```diff
@@ -1168,7 +1170,7 @@ export class Terminal extends CoreTerminal implements ITerminal {
       this.buffer.lines.push(this.buffer.getBlankLine(DEFAULT_ATTR_DATA));
     }
     this.refresh(0, this.rows - 1);
-    this._onScroll.fire(this.buffer.ydisp);
+    this._onScroll.fire({position: this.buffer.ydisp, source: ScrollSource.TERMINAL });
```
schrej, line 1173: Or would you prefer a proper class for this?

Tyriar: Interface is fine

AndreSilva1993:
Curious about this as I have the xterm v4.9.0 and scrolling the terminal does not fire the `onScroll` event. Do I need to do anything?

Tyriar:
@AndreSilva1993 that was a long time ago so I don't remember 😅. You should probably update as I don't think there are any breaking changes, if your build does break when updating it's normally just making TS types stricter.

AndreSilva1993:
I must've been sleeping when I wrote this, we're on the 4.18.0 already 🤦‍♂️

Tyriar:
You're right this doesn't happen anymore, it's only firing `onScroll` when the "base" changes. Can you create an issue if one doesn't exist since that doesn't match up with the docs https://github.com/Tyriar/xterm.js/blob/fb950fcbe54b392c6aef268b899c33bcc9c3122f/typings/xterm.d.ts#L862-L867

AndreSilva1993:
Here's the [issue](https://github.com/xtermjs/xterm.js/issues/3864).

PR #3209, opened 2021-01-08
freeze the default ANSI colours const

Nokel81:
I noticed this code comment while debugging an issue of mine. `Object.freeze(...)` actually enforces this so I think it would be a good idea to do so.


Tyriar (approved):
👍 if tests pass

PR #3211, opened 2021-01-09
0x7f is printable in ground state

carlopi:
Align with https://vt100.net/emu/dec_ansi_parser state machine ground state rules

Currently launching the test setup, opening the console and writing: `term.write(new Uint8Array([127]))` lead to an error "xterm.js: Parsing error: ...", while it seems to me that 0x7f has to be treated as printable whenever in the GROUND state (check https://vt100.net/emu/dec_ansi_parser, upper left corner).
With this fix, that basically add to the table the rule that 0x7f is to be treated as printable whenever in GROUND state (while remaining in GROUND state), it works as intended (= nothing prints, but no error).

This came up while doing some additional work on: [repl.leaningtech.com](url), that uses xterm.js as console.

This is my first contribution to this project, I agree to code of conduct, licence and all.

I hope I have followed the proper process, thanks for the great work
Best,
Carlo

### DIFF
--- a/src/common/parser/EscapeSequenceParser.ts
+++ b/src/common/parser/EscapeSequenceParser.ts
@@ -95,8 +95,9 @@ export const VT500_TRANSITION_TABLE = (function (): TransitionTable {
 
   // set default transition
   table.setDefault(ParserAction.ERROR, ParserState.GROUND);
-  // printables
+  // printables and 7f
   table.addMany(PRINTABLES, ParserState.GROUND, ParserAction.PRINT, ParserState.GROUND);
+  table.add(0x7f, ParserState.GROUND, ParserAction.PRINT, ParserState.GROUND);
   // global anywhere rules
   for (state in states) {
     table.addMany([0x18, 0x1a, 0x99, 0x9a], state, ParserAction.EXECUTE, ParserState.GROUND);


jerch (changes requested):
@carlopi May I ask why you have `\x7F` ending up in a terminal data stream? Normally this character does not appear at all, it gets replaced by the TTY layer with `BS SP BS` (in ICANON).  This is also the reason why it has no default meaning/handling. See the notes on VT100.net about its handling depending on the loaded character sets. As we dont implement the full ECMA-35 character sets logic, we simply have no usage for it during output stage. The correct way to deal with it would be to ignore it with a 94-charset (in line with vt100) while displaying something with a 96-charset (vt320 and up). Currently we always ignore it and mark it as parsing error, as we dont expect it to occur at all in the input data.

Simply treating it as PRINTABLE does not solve the ambiguity, instead introduces another problem - it now will end up in the terminal buffer as "silent" character (as JS has no visual represantation for it) possibly leading to side effects during render stage.

A more error-proof fix would be to mark it as EXECUTABLE and to treat it depending on some terminal setting (whether it should print something or gets ignored). Then embedders can decide on their own whether they want to stick to VT100 behavior (ignore) or to see it on the screen as with VT320+.

carlopi:
My use case is that I am executing [](https://www.ioccc.org)'s entries inside a VM in the browser, and especially this year winner emits 0x7F to the stream (among a lot of other garbage, before allowing to play a game of tic-tac-toe in the console). It is a quite contrived example with plenty of corner cases, and I might solve it otherwise.

In any case it would seem to me that the current situation (ignoring + error) seems worst that any of:
- always ignoring
- always printing something
- allowing the user to specify the behavior

Do you think it could be worth having it fixed? I would like to help if possible.
Do you have any recommendation on how to move forward?

jerch:
> My use case is that I am executing https://www.ioccc.org's entries inside a VM in the browser, and especially this year winner emits 0x7F to the stream (among a lot of other garbage, before allowing to play a game of tic-tac-toe in the console). It is a quite contrived example with plenty of corner cases, and I might solve it otherwise.

Here I suggest to check if this issue remains on a "normal" terminal run from the system (e.g. check with our demo in this repo if the data still contains those spurious DEL errors). If you still get the the parsing error, maybe contact the creator and ask him what DEL is supposed to mean here. If there are no DELs anymore (most likely), then your in-browser VM has a broken TTY implementation (emscripten used to be broken in this regard mostly missing a working POSIX terminal abstraction, not sure if this got fixed somehow).

> Do you think it could be worth having it fixed? I would like to help if possible.
Do you have any recommendation on how to move forward?

If you want to silence the parsing error in xterm.js I suggest to do the following:
- map DEL to EXECUTABLES in GROUND
- implement a handler for DEL in `InputHandler`:
  - respect some terminal wide setting whether to ignore or print something
  - another setting for the printable char (or maybe just defaulting to ␡ in the beginning)
  - in the long run: maybe try to derive behavior from active charset (imho not worth doing that, but would be the correct way in terms of VT devices)
- set up some test cases for the new behavior:
  - change transition tests in `EscapeSequenceParser.test.ts` to reflect DEL being treated as EXECUTABLE in GROUND
  - add tests in `InputHandler.test.ts` for DEL inserted at various parsers states and different print/ignore setting
 
So sure feel free to do that :smile_cat:.

carlopi:
Thanks for the detailed feedback.
I will get back when I have made progress, for now I think it's better to close the PR.

Bests,
Carlo

PR #3212, opened 2021-01-11
Create a node.js target for xterm.js

joyceerhl:
Hey @Tyriar, here's an attempt at what you described in https://github.com/xtermjs/xterm.js/issues/2749#issuecomment-756272750. This PR:
- Duplicates `src/browser/public/Terminal.ts` --> `src/common/public/Terminal.ts`.
- Duplicates `src/browser/Terminal.ts` --> `src/common/public/TerminalCore.ts` since `src/browser/public/Terminal.ts` wraps `src/browser/Terminal.ts`.
- Removes DOM/UI references and `src/browser` imports from `src/common/public/Terminal.ts` and `src/common/public/TerminalCore.ts`. I took a guess at what to remove based on what I thought would still be meaningful in the Node environment and on the discussion in https://github.com/xtermjs/xterm.js/issues/2749, but I'd appreciate feedback here.
- Adds a separate webpack config and tsconfig as well as a small 'testcase', `node-test/index.js`, which I used to verify that the changes described above can be compiled, built, and used in a Node project based on the included README.
- I only checked in xterm-core.d.ts in the first commit in this PR to get the project compiling, since I expect that I've either taken out too much or not enough from `src/common/public/Terminal.ts`. Changes to that file should be disregarded for now.

I'm expecting this to need lots of work—looking forward to your feedback!


core-webpack.config.js
```diff
@@ -0,0 +1,41 @@
+/**
```
Tyriar, line 1: I might be wrong, but I think naming this webpack.config.core.js will light up some smarts in VS Code (icons, auto typings).

src/common/public/tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "types": [
+      "../../../node_modules/@types/mocha",
+      "../../../node_modules/@types/node"
+    ],
+    "baseUrl": "../../",
+    "extendedDiagnostics": true
+  },
+  "include": [
+    "../**/*",
+    "../../../typings/xterm-core.d.ts",
```
Tyriar, line 18: After a quick look I'm not sure what the CI/yarn errors are about here:

```
src/common/public/Terminal.ts:6:281 - error TS2307: Cannot find module 'xterm-core' or its corresponding type declarations.

6 import { Terminal as ITerminalApi, ITerminalOptions, IMarker, IDisposable, IBuffer as IBufferApi, IBufferNamespace as IBufferNamespaceApi, IBufferLine as IBufferLineApi, IBufferCell as IBufferCellApi, IParser, IFunctionIdentifier, IUnicodeHandling, IUnicodeVersionProvider } from 'xterm-core';
                                                                                                                                                                                                                                                                                          ~~~~~~~~~~~~

error TS2688: Cannot find type definition file for '../../../out-test/api/TestUtils'.

error TS2688: Cannot find type definition file for '../../../out-test/api/TestUtils'.

error TS2688: Cannot find type definition file for '../../../out-test/api/TestUtils'.

error TS2688: Cannot find type definition file for '../../../out-test/api/TestUtils'.
```

src/common/public/tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "../../../typings/xterm.d.ts", // common/Types.d.ts imports from 'xterm'
```
Tyriar, line 19: This seems like a problem, we can't import from xterm if we want it to run for node because it touches DOM types which will throw for tsconfig targets without the dom lib.

src/common/public/Terminal.ts
```diff
@@ -0,0 +1,239 @@
...
+  private _verifyIntegers(...values: number[]): void {
+    for (const value of values) {
+      if (value === Infinity || isNaN(value) || value % 1 !== 0) {
+        throw new Error('This API only accepts integers');
+      }
+    }
+  }
+}
+
+class BufferApiView implements IBufferApi {
```
Tyriar, line 121: These API view objects seem like good candidates to pull into smaller PRs to get merged into master quickly. I'm thinking `BufferApiView`, `BufferLineApiView`, `BufferNamespaceApi`, `ParserApi`, `UnicodeApi` could all be moved to live in `src/common/public/...` in master and have the `src/browser/public/Terminal.ts` import them

typings/xterm-core.d.ts
```diff
@@ -0,0 +1,1207 @@
+/**
+ * @license MIT
+ *
+ * This contains the type declarations for the xterm.js library. Note that
+ * some interfaces differ between this file and the actual implementation in
+ * src/, that's because this file declares the *public* API which is intended
+ * to be stable and consumed by external programs.
+ */
+
+declare module 'xterm-core' {
```
Tyriar, line 10: Another problem I'd like for us to solve is the sharing of these typings, I think having `xterm.d.ts` import from `xterm-core.d.ts` will cause problems because of this module declaration. We may need separate .d.ts that just export the core types without the module wrapping them to be able to share.

package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "prepackage": "npm run build",
     "package": "webpack",
+    "compile": "tsc -b ./src/common/public/tsconfig.json",
```
Tyriar, line 13: This can be added to the project references list in `tsconfig.all.json` to then build on `yarn watch`

Tyriar (changes requested):
Nice work, some high level comments below 🙂

PR #3214, opened 2021-01-12
Move API view classes under common/public

joyceerhl:
For https://github.com/xtermjs/xterm.js/pull/3212#discussion_r555103406


src/common/public/UnicodeApi.ts
```diff
@@ -0,0 +1,22 @@
+import { ITerminal } from 'browser/Types';
+import { IUnicodeHandling, IUnicodeVersionProvider } from 'xterm';
+
+export class UnicodeApi implements IUnicodeHandling {
+  constructor(private _core: ITerminal) { }
```
Tyriar, line 5: I think passing in `CoreTerminal` here will help work around the dependence on `browser/` which is one of the compile errors.

src/common/public/ParserApi.ts
```diff
@@ -0,0 +1,32 @@
+import { IParams } from 'common/parser/Types';
+import { ITerminal } from 'browser/Types';
+import { IDisposable, IFunctionIdentifier, IParser } from 'xterm';
```
Tyriar, line 3: There may be compile error here because it's importing dom types into a "common environment", the types depended upon by common/public I think will need their types pulled into a separate types file. I'd recommend `typings/xterm-common.d.ts` which does _not_ export a module, then `xterm.d.ts` can import & export them via something like this:

```ts
export { IDisposable, IFunctionIdentifier, IParser } from '../../../typings/xterm-common';
```

This might have knock on effects for the [API documentation](https://xtermjs.org/docs/api/terminal/modules/xterm/), but I think we have to do this because I see a big mess of things getting out of sync if we duplicate the API into core.

Tyriar (changes requested):

src/common/public/ParserApi.ts
Tyriar: Guess this isn't a problem, we already import `xterm` into this project anyway. It may be an issue when we try to build a common-only target in https://github.com/xtermjs/xterm.js/pull/3212

PR #3215, opened 2021-01-14
fire onSelectionChange on empty selection

### LINKED ISSUE #3193: onSelectionChange() is not being invoked when the selection is cleared.
Awesome JS terminal emulator - thanks for your contribution. I did notice the following (minor) bug today:

When adjusting the selection in the terminal, I see onSelectionChange() being triggered as expected. However, if I left-click with the mouse on the terminal (with no drag), the selection is cleared, but onSelectionChange() is NOT triggered as would be expected.

This may have been the original intent of the following issue, but the description in the issue is sparse and may have been misinterpreted.
https://github.com/xtermjs/xterm.js/issues/2070

## Details
- Browser and browser version: Chrome  87.0.4280.88, but not expected to be browser specific
- OS version: Windows 10 Pro, version 1903, but not expected to be OS specific
- xterm.js version: 4.9.0

## Steps to reproduce:
1. Setup an development environment that can run an xterm.js terminal.
2. After opening the terminal, add the following code:
```
        terminal.onSelectionChange(() => {
            console.log('onSelectionChange', terminal.getSelection());
        });
```
3. Run the application and select text with the mouse. Observe onSelectionChange being called with the selected text logged.
4. Click on the terminal (with no drag). Observe the selection has now been cleared. Observe that onSelectionChange is NOT called.
5. I think the expected behavior would be to call onSelectionChange for consistency reasons, as the selection has changed, right?

marvinthepa:
Fire the onSelectionChange event when a selection is cleared by
single-clicking somewhere inside the terminal area.

fixes #3193.

This also slightly changes the behaviour when re-selecting the same area
in the terminal. Before, this fired onSelectionChange, but it does not
anymore, even when changing the direction of the selection.

### DIFF
--- a/src/browser/services/SelectionService.ts
+++ b/src/browser/services/SelectionService.ts
@@ -690,16 +690,26 @@ export class SelectionService extends Disposable implements ISelectionService {
   }
 
   private _fireIfSelectionChanged(): void {
+    // Fire if there is no selection
     const hasSelection = this.hasSelection;
-    if (!hasSelection && !this._oldHasSelection) {
+    if (!hasSelection) {
+      if (this._oldHasSelection) {
+        this._onSelectionChange.fire();
+      }
       return;
     }
 
     const start = this._model.finalSelectionStart;
     const end = this._model.finalSelectionEnd;
 
-    if (start?.[0] !== this._oldSelectionStart?.[0] || start?.[1] !== this._oldSelectionStart?.[1] ||
-      end?.[0] !== this._oldSelectionEnd?.[0] || end?.[1] !== this._oldSelectionEnd?.[1]) {
+    // Sanity check, these should not be undefined as there is a selection
+    if (!start || !end) {
+      return;
+    }
+
+    if (!this._oldSelectionStart || !this._oldSelectionEnd || (
+      start[0] !== this._oldSelectionStart[0] || start[1] !== this._oldSelectionStart[1] ||
+      end[0] !== this._oldSelectionEnd[0] || end[1] !== this._oldSelectionEnd[1])) {
 
       this._oldSelectionStart = start;
       this._oldSelectionEnd = end;
--- a/src/browser/services/SelectionService.ts
+++ b/src/browser/services/SelectionService.ts
@@ -105,6 +105,9 @@ export class SelectionService extends Disposable implements ISelectionService {
   private _workCell: CellData = new CellData();
 
   private _mouseDownTimeStamp: number = 0;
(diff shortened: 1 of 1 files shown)


Tyriar (approved):
Thanks for the fix and for finding the other bug 🙂

src/browser/services/SelectionService.ts
```diff
@@ -690,16 +690,26 @@ export class SelectionService extends Disposable implements ISelectionService {
   }
 
   private _fireIfSelectionChanged(): void {
+    // Fire if there is no selection
     const hasSelection = this.hasSelection;
-    if (!hasSelection && !this._oldHasSelection) {
+    if (!hasSelection) {
+      if (this._oldHasSelection) {
+        this._onSelectionChange.fire();
```
marvinthepa, line 697: it seems to me that this is missing

  this._oldHasSelection = false

src/browser/services/SelectionService.ts
```diff
@@ -690,16 +690,26 @@ export class SelectionService extends Disposable implements ISelectionService {
...
     const end = this._model.finalSelectionEnd;
 
-    if (start?.[0] !== this._oldSelectionStart?.[0] || start?.[1] !== this._oldSelectionStart?.[1] ||
-      end?.[0] !== this._oldSelectionEnd?.[0] || end?.[1] !== this._oldSelectionEnd?.[1]) {
+    // Sanity check, these should not be undefined as there is a selection
+    if (!start || !end) {
+      return;
+    }
+
+    if (!this._oldSelectionStart || !this._oldSelectionEnd || (
```
marvinthepa, line 710: I don't know if this `!this._oldSelectionEnd` is correct here. I vaguely remember that selectionEnd can be undefined, even when something is selected. Maybe I am wrong, I will check it again.

Tyriar: Ah ok, I was trying to get rid of the `?.[]` due to what that would transpile down to:

```ts
a.?[0]
// compiles to:
a === null || a === void 0 ? void 0 : a[0];
```

If there is a bug, definitely create an issue so we can get it fixed for this upcoming version.

src/browser/services/SelectionService.ts
Tyriar: Created https://github.com/xtermjs/xterm.js/issues/3225

src/browser/services/SelectionService.ts
marvinthepa: I got it wrong.

PR #3217, opened 2021-01-15
fix: change textarea to input

testrocket:
It changes HTMLTextAreaElement to HTMLInputElement element with type="password" to disable suggestions on android mobile devices.

### DIFF
--- a/src/browser/Clipboard.ts
+++ b/src/browser/Clipboard.ts
@@ -42,7 +42,7 @@ export function copyHandler(ev: ClipboardEvent, selectionService: ISelectionServ
  * @param ev The original paste event to be handled
  * @param term The terminal on which to apply the handled paste event
  */
-export function handlePasteEvent(ev: ClipboardEvent, textarea: HTMLTextAreaElement, coreService: ICoreService): void {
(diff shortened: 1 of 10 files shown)


jerch (changes requested):
Sorry this gets a downvote from my side at the current stage:

Changing a DOM element type, even internally, is a big deal, it might trigger faulty behavior on other ends.

_How to proceed from here?_
Care to give a bit more context? There no issue linked to it, what would be helpful to repro and comprehend the problem, and discuss possible solutions. Here it is unclear what you have tried already.

Note that the textarea thingy is a hack to get around certain limitations of key input events, those all have to work as before - from event propagation and input reading (composed keys?) up to the repositioning trick for IME popups. Several of those things are not covered by our tests, thus need manual testing.

Tyriar (changes requested):
Yeah I'm pretty sure this will cause a bunch of weird regressions, additionally we don't want it to be marked as a password field because that would cause problems in the screen reader support.

testrocket:
I appologize for this PR. This was a mistake - I wanted create a PR for a fork and for some reason it landed here. Again sorry for that.

Tyriar:
No worries 🙂

PR #3223, opened 2021-01-20
cleanup parser handler interface

jerch:
Cleanup of the somewhat messed up handler interface of the parser.

Changes:
- remove setHandler
- apply boolean return to ESC | CSI | OSC | DCS handlers
  (rewrite void returns to true)
- rename add\*Handler to register\*Handler


jerch:
@Tyriar Ready for review.

Note that I did not remove the `setHandler` for PRINT and EXECUTABLE, as they have no disposable interface. We could unify this as well, but I am abit reluctant, as it will hurt performance much more than for the other types (both account for >>90% of terminal data). Also we had no request to expose those, so imho we are good to keep them with the simpler `setHandler` interface for now.

Tyriar (approved):
Looks good!

I have great confidence in the parser tests as well so we can merge this for 4.10 (release in 1-2 weeks) 👍

PR #3227, opened 2021-01-21
Store old selection when firing onSelectionChange

### LINKED ISSUE #3225: We're not clearing the old selection in selection service anymore
https://github.com/xtermjs/xterm.js/pull/3215/commits/66c80d8b788dae552df63f3df7a32e198767ae8e#r561386373

marvinthepa:
Fixes #3225.

Also see #3215.

I inlined `this.hasSelection` to avoid calling `this._model.finalSelectionStart` and `this._model.finalSelectionEnd` twice. So a bit of duplicated code for more performance.

For the `fire` arrow function I went with the opposite approach 😄 .

### DIFF
--- a/src/browser/services/SelectionService.ts
+++ b/src/browser/services/SelectionService.ts
@@ -685,22 +685,21 @@ export class SelectionService extends Disposable implements ISelectionService {
         }
       }
     } else {
-      this._fireIfSelectionChanged();
+      this._fireEventIfSelectionChanged();
     }
   }
 
-  private _fireIfSelectionChanged(): void {
-    // Fire if there is no selection
-    const hasSelection = this.hasSelection;
+  private _fireEventIfSelectionChanged(): void {
+    const start = this._model.finalSelectionStart;
+    const end = this._model.finalSelectionEnd;
+    const hasSelection = !!start && !!end && (start[0] !== end[0] || start[1] !== end[1]);
+
     if (!hasSelection) {
       if (this._oldHasSelection) {
-        this._onSelectionChange.fire();
+        this._fireOnSelectionChange(start, end, hasSelection);
       }
       return;
     }
-
-    const start = this._model.finalSelectionStart;
-    const end = this._model.finalSelectionEnd;
 
     // Sanity check, these should not be undefined as there is a selection
     if (!start || !end) {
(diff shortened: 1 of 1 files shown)


Tyriar (changes requested):

src/browser/services/SelectionService.ts
```diff
@@ -690,18 +690,24 @@ export class SelectionService extends Disposable implements ISelectionService {
...
+    const start = this._model.finalSelectionStart;
+    const end = this._model.finalSelectionEnd;
+    const hasNoSelection = !start || !end || (start[0] === end[0] && start[1] === end[1]);
+
+    const fire = (): void => {
+      this._oldSelectionStart = start;
+      this._oldSelectionEnd = end;
+      this._oldHasSelection = !hasNoSelection;
+      this._onSelectionChange.fire();
+    };
```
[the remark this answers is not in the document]
marvinthepa: Ah, I had considered that but wanted to avoid passing all those parameters.

src/browser/services/SelectionService.ts
```diff
@@ -690,18 +690,24 @@ export class SelectionService extends Disposable implements ISelectionService {
...
-    const hasSelection = this.hasSelection;
-    if (!hasSelection) {
+    const start = this._model.finalSelectionStart;
+    const end = this._model.finalSelectionEnd;
+    const hasNoSelection = !start || !end || (start[0] === end[0] && start[1] === end[1]);
+
+    const fire = (): void => {
+      this._oldSelectionStart = start;
+      this._oldSelectionEnd = end;
+      this._oldHasSelection = !hasNoSelection;
```
Tyriar, line 700: Actually I'm not sure this is right, `_oldHasSelection` would be wrong if changing from some selection to a different selection, no?

marvinthepa: > if changing from some selection to a different selection

In this case hasSelection is true and _oldHasSelection is not evaluated at all:

```typescript
    if (!hasSelection) {
      if (this._oldHasSelection) {
        this._fireOnSelectionChange(start, end, hasSelection);
      }
      return;
    }
```

Then the selections are compared.

As far as I can see, the worst that can happen in

```
      this._oldHasSelection = hasSelection;
```

is a no-op.

(This is referring to the last amend I have pushed (554b4a822403), I got rid of hasNoSelection is the negation just added mental overhead when reading the code. It is also now closer to `this._hasSelection` which should make it easier to adapt should that one need to change)

src/browser/services/SelectionService.ts
```diff
@@ -685,23 +685,22 @@ export class SelectionService extends Disposable implements ISelectionService {
...
     } else {
-      this._fireIfSelectionChanged();
+      this._fireEventIfSelectionChanged();
     }
   }
 
-  private _fireIfSelectionChanged(): void {
-    // Fire if there is no selection
-    const hasSelection = this.hasSelection;
+  private _fireEventIfSelectionChanged(): void {
```
marvinthepa, line 692: renamed to be easier distinguishable from `_fireOnSelectionChange`

Tyriar (approved):
Works great 👍

PR #3229, opened 2021-01-23
replace expect testing with assert

jerch:
Only a cleanup PR, but since this bugged me for a while and I have to deal with testing for the async PR anyway, time to replace the uglier expect based testing code with some simple assertion testing.

@Tyriar The replacements only use `.equal`, `deepEqual` and `doesNotThrow` from `chai.assert`. Not sure if this could be used to get rid of `chai` someday and just use nodejs' assert lib. I remember we once had an issue regarding this, but dont recall why it was not possible in the end (different semantics?)


Tyriar (approved):
Nice, this bugged me too.

> Not sure if this could be used to get rid of chai someday and just use nodejs' assert lib. I remember we once had an issue regarding this, but dont recall why it was not possible in the end (different semantics?)

The problem was around our common/browser environments, node.js' assertions won't work in browser due to layer breakages. I settles on keeping chai because it's a common interface for both our common (node/browser) and browser layers.

PR #3230, opened 2021-01-24
Use linkifier2 to double click select links

### LINKED ISSUE #682: Consider adding smarter double click to select word logic
https://github.com/sourcelair/xterm.js/pull/670 brings in new double click to select word that allows you to select whole paths/URLs instead of just small fragments of them. This could be improved though for a number of cases, for example:

- Not selecting trailing `:` in `ls -lR` output
- Not selecting `(`, `)`, `"` surrounding paths, etc.

marvinthepa:
Also works with right click select.
Fixes #682.

This relies on the fact that before a user (double/right)-clicks a link, the cursor will hover the link, triggering the asynchronous link provider(s). By the time the click event fires, the link provider "should" have done its thing, and the link is used.
If not, this falls back to the old behavior.
This seemed good enough to me. If it is not, one probably would have to expose some inner state of the linkifier to provide some way to check if there are currently link providers resolving and to wait for them to finish before selecting.

I feel a bit bad about duplicating the magic "-1"s (on both x and y for start, only on y for end) from `Linkifier2._fireUnderlineEvent`, but I couldn't think of a good way to extract that logic.
It seems to me to produce the correct results.

### DIFF
--- a/src/browser/Linkifier2.ts
+++ b/src/browser/Linkifier2.ts
@@ -3,7 +3,7 @@
  * @license MIT
  */
 
-import { ILinkifier2, ILinkProvider, IBufferCellPosition, ILink, ILinkifierEvent, ILinkDecorations } from 'browser/Types';
+import { ILinkifier2, ILinkProvider, IBufferCellPosition, ILink, ILinkifierEvent, ILinkDecorations, ILinkWithState } from 'browser/Types';
 import { IDisposable } from 'common/Types';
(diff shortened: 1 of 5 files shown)


Tyriar:
Thanks for checking this out, I'm busy testing/stabilizing the next version of VS Code so should get to PRs within the week.

Tyriar:
This works really well, thanks a bunch!

PR #3232, opened 2021-01-25
Fix webgl renderer does not render when writing if a selection exists

### LINKED ISSUE #3221: Webgl renderer does not render when writing if a selection exists
Repro:

1. Open demo
2. Make a selection
3. Run in console: `term.write('foo')`

jeanp413:
Fixes #3221

Also fixes the following issue:
1. Make a selection
2. Resize terminal rows or columns
3. :bug: Selected text is rendered incorrectly

I chose to behave as the canvas renderer by removing the selection when resizing
xterm_selection_1


Tyriar (changes requested):
Seems to work nicely 👍 

> I chose to behave as the canvas renderer by removing the selection when resizing

I think we want to keep this behavior to align with the dom renderer as we will be removing the canvas one eventually. It's useful because resizing the terminal by dragging isn't the only cause for resize, it can also happen when maximizing the terminal panel in VS Code or zooming the window (ctrl++/-); we shouldn't be dropping the selection if it's still valid because the user might still want it.

jeanp413:
Updated the PR to keep the selection on resize and when updating colors

Tyriar (approved):
This works great, thanks! 👍

PR #3233, opened 2021-01-25
Add hover and leave callbacks for linkprovider

### LINKED ISSUE #3056: Add hover callback for LinkProviders
The LinkMatchers are eventually removed, right?
If you pass `_useLinkProvider` to the xterm-web-links-addon, there is currently no way to get a hover callback to show a tooltip.
I would like to switch to the LinkProvider but I also want to show a tooltip when hovering a link

marvinthepa:
Fixes #3056.

I guess this needs another breaking change when LinkMatchers are removed, to replace the `ILinkMatcherOptions` argument to the `WebLinksAddon` constructor with something not referring to LinkMatchers..

### DIFF
--- a/addons/xterm-addon-web-links/src/WebLinkProvider.ts
+++ b/addons/xterm-addon-web-links/src/WebLinkProvider.ts
@@ -3,7 +3,12 @@
  * @license MIT
  */
 
-import { ILinkProvider, ILink, Terminal, IViewportRange, ILinkProviderOptions } from 'xterm';
+import { ILinkProvider, ILink, Terminal, IViewportRange } from 'xterm';
+
+interface ILinkProviderOptions {
+  hover?(event: MouseEvent, text: string, location: IViewportRange): void;
+  leave?(event: MouseEvent, text: string): void;
+}
 
 export class WebLinkProvider implements ILinkProvider {
 
--- a/addons/xterm-addon-web-links/src/WebLinkProvider.ts
+++ b/addons/xterm-addon-web-links/src/WebLinkProvider.ts
@@ -3,20 +3,35 @@
  * @license MIT
  */
 
-import { ILinkProvider, IBufferCellPosition, ILink, Terminal } from 'xterm';
+import { ILinkProvider, ILink, Terminal, IViewportRange, ILinkProviderOptions } from 'xterm';
 
 export class WebLinkProvider implements ILinkProvider {
 
   constructor(
     private readonly _terminal: Terminal,
     private readonly _regex: RegExp,
-    private readonly _handler: (event: MouseEvent, uri: string) => void
+    private readonly _handler: (event: MouseEvent, uri: string) => void,
+    private readonly _options: ILinkProviderOptions = {}
   ) {
 
   }
 
   public provideLinks(y: number, callback: (links: ILink[] | undefined) => void): void {
-    callback(LinkComputer.computeLink(y, this._regex, this._terminal, this._handler));
+    const links = LinkComputer.computeLink(y, this._regex, this._terminal, this._handler);
+    callback(this._addCallbacks(links));
+  }
+
+  private _addCallbacks(links: ILink[]): ILink[] {
+    return links.map(link => {
+      link.leave = this._options.leave;
+      link.hover = (event: MouseEvent, uri: string): void => {
+        if (this._options.hover) {
+          const { range } = link;
+          this._options.hover(event, uri, range);
+        }
+      };
+      return link;
+    });
   }
 }
 
--- a/addons/xterm-addon-web-links/src/WebLinksAddon.ts
+++ b/addons/xterm-addon-web-links/src/WebLinksAddon.ts
@@ -3,7 +3,7 @@
  * @license MIT
  */
 
-import { Terminal, ILinkMatcherOptions, ITerminalAddon, IDisposable, ILinkProviderOptions } from 'xterm';
+import { Terminal, ILinkMatcherOptions, ITerminalAddon, IDisposable, IViewportRange } from 'xterm';
 import { WebLinkProvider } from './WebLinkProvider';
 
 const protocolClause = '(https?:\\/\\/)';
@@ -36,6 +36,11 @@ function handleLink(event: MouseEvent, uri: string): void {
   }
 }
 
+interface ILinkProviderOptions {
+  hover?(event: MouseEvent, text: string, location: IViewportRange): void;
+  leave?(event: MouseEvent, text: string): void;
+}
+
 export class WebLinksAddon implements ITerminalAddon {
   private _linkMatcherId: number | undefined;
   private _terminal: Terminal | undefined;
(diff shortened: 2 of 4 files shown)


addons/xterm-addon-web-links/src/WebLinksAddon.ts
```diff
@@ -53,7 +53,7 @@ export class WebLinksAddon implements ITerminalAddon {
     this._terminal = terminal;
 
     if (this._useLinkProvider && 'registerLinkProvider' in this._terminal) {
-      this._linkProvider = this._terminal.registerLinkProvider(new WebLinkProvider(this._terminal, strictUrlRegex, this._handler));
+      this._linkProvider = this._terminal.registerLinkProvider(new WebLinkProvider(this._terminal, strictUrlRegex, this._handler, this._options));
```
Tyriar, line 56: I think it would be better if we has link provider-specific options:

```ts
export interface ILinkProviderOptions {
  hover?(event: MouseEvent, text: string): void;
  leave?(event: MouseEvent, text: string): void;
}

...

private _options: ILinkMatcherOptions | ILinkProviderOptions = {}
```

addons/xterm-addon-web-links/src/WebLinkProvider.ts
```diff
@@ -3,20 +3,35 @@
...
 
   }
 
   public provideLinks(y: number, callback: (links: ILink[] | undefined) => void): void {
-    callback(LinkComputer.computeLink(y, this._regex, this._terminal, this._handler));
+    const links = LinkComputer.computeLink(y, this._regex, this._terminal, this._handler);
+    callback(this._addCallBacks(links, this._options.tooltipCallback, this._options.leaveCallback));
+  }
+
+  private _addCallBacks(links: ILink[], hover?: (event: MouseEvent, uri: string, location: IViewportRange) => void, leave?: (event: MouseEvent, uri: string) => void): ILink[] {
```
Tyriar, line 24:
```suggestion
    callback(this._addCallbacks(links, this._options.tooltipCallback, this._options.leaveCallback));
  }

  private _addCallbacks(links: ILink[], hover?: (event: MouseEvent, uri: string, location: IViewportRange) => void, leave?: (event: MouseEvent, uri: string) => void): ILink[] {
```

addons/xterm-addon-web-links/src/WebLinkProvider.ts
```diff
@@ -3,20 +3,35 @@
...
+    return links.map((link): ILink => {
```
Tyriar, line 25: I think we can avoid the type annotation?

```suggestion
    return links.map(link => {
```

addons/xterm-addon-web-links/src/WebLinkProvider.ts
```diff
@@ -3,20 +3,35 @@
...
+      let hoverCallback = (event: MouseEvent, uri: string): void => { };
+      if (hover) {
+        hoverCallback = (event, uri): void => {
+          const { range } = link;
+          hover(event, uri, range);
+        };
+      }
+      return { ...link, leave, hover: hoverCallback };
```
Tyriar, line 33: I think we're safe here to add the props directly to the same object, rather than creating a new one?

```suggestion
      link.leave = leave;
      link.hover = hoverCallback;
      return link;
```

Tyriar (changes requested):

addons/xterm-addon-web-links/src/WebLinksAddon.ts
marvinthepa: > ```ts
> private _options: ILinkMatcherOptions | ILinkProviderOptions = {}
> ```

Do you want to allow both types as arguments on `WebLinksAddon`?
I am afraid my knowledge of typescript is not sufficient to get that to compile.

Or do you want to allow both as argument to `WebLinksProvider`?

As a suggestion, I will push an alternative that I could get working:
Converting from `ILinkMatcherOptions` to `ILinkProviderOptions` in `WebLinksAddon.activate`.
If you do not like that I am afraid I will need a little help.

addons/xterm-addon-web-links/src/WebLinksAddon.ts
```diff
@@ -53,7 +53,11 @@ export class WebLinksAddon implements ITerminalAddon {
     this._terminal = terminal;
 
     if (this._useLinkProvider && 'registerLinkProvider' in this._terminal) {
-      this._linkProvider = this._terminal.registerLinkProvider(new WebLinkProvider(this._terminal, strictUrlRegex, this._handler));
+      const options = {
+        hover: this._options.tooltipCallback,
+        leave: this._options.leaveCallback
+      };
```
marvinthepa, line 59: When linkmatchers are removed, this can go and the ILinkMatcherOptions argument can be replaced by ILinkProviderOptions..

addons/xterm-addon-web-links/src/WebLinksAddon.ts
Tyriar: I don't think TS would fuss about this because all properties on both interfaces are optional, so essentially any object matches the interface, it should however provide nice intellisense when something is using the API.

Tyriar: Ah I see the problem you were hitting now, I'll push a fix using a cast as it should be safe to do that since the interfaces only have optional props.

Tyriar (approved):

addons/xterm-addon-web-links/src/WebLinksAddon.ts
marvinthepa: > I don't think TS would fuss about this because all properties on both interfaces are optional, so essentially any object matches the interface, it should however provide nice intellisense when something is using the API.

It did fuss about it in my experience 🙁.

E.g.:

```
  constructor(
    private _handler: (event: MouseEvent, uri: string) => void = handleLink,
    private _options: ILinkMatcherOptions | ILinkProviderOptions = {},
    private _useLinkProvider: boolean = false
  ) {
    this._options.matchIndex = 1; //  Property 'matchIndex' does not exist on type 'ILinkProviderOptions'.ts(2339)
  }
```

I will try to solve this in a different way.

Tyriar: @marvinthepa yeah that's what I ended up seeing, you can check the change I just pushed for an alright workaround

addons/xterm-addon-web-links/src/WebLinksAddon.ts
```diff
@@ -36,27 +36,34 @@ function handleLink(event: MouseEvent, uri: string): void {
...
   ) {
-    this._options.matchIndex = 1;
   }
 
   public activate(terminal: Terminal): void {
     this._terminal = terminal;
 
     if (this._useLinkProvider && 'registerLinkProvider' in this._terminal) {
-      this._linkProvider = this._terminal.registerLinkProvider(new WebLinkProvider(this._terminal, strictUrlRegex, this._handler));
+      const options = this._options as ILinkProviderOptions;
```
Tyriar, line 60: @marvinthepa normally using `as` is bad, but it doesn't matter much here since the objects are definitely compatible.

addons/xterm-addon-web-links/src/WebLinksAddon.ts
```diff
@@ -36,27 +36,34 @@ function handleLink(event: MouseEvent, uri: string): void {
...
+      this._linkProvider = this._terminal.registerLinkProvider(new WebLinkProvider(this._terminal, strictUrlRegex, this._handler, options));
     } else {
       // TODO: This should be removed eventually
-      this._linkMatcherId = (this._terminal as Terminal).registerLinkMatcher(strictUrlRegex, this._handler, this._options);
+      const options = this._options as ILinkMatcherOptions;
+      options.matchIndex = 1;
```
Tyriar, line 65: I moved this here to avoid setting it when the object gets used for a link provider

addons/xterm-addon-web-links/src/WebLinksAddon.ts
```diff
@@ -36,31 +36,34 @@ function handleLink(event: MouseEvent, uri: string): void {
...
 
   public activate(terminal: Terminal): void {
     this._terminal = terminal;
 
     if (this._useLinkProvider && 'registerLinkProvider' in this._terminal) {
-      const options = {
-        hover: this._options.tooltipCallback,
-        leave: this._options.leaveCallback
-      };
+      const options = this._options as ILinkProviderOptions;
```
marvinthepa: I knew there was something about TS that I didn't know.

However, this will break clients that pass their callbacks as `tooltipCallback` and `leaveCallback` (instead of `hover` and `leave`, and then switch `_useLinkProvider` to `true`. Shouldn't there be at least a warning in this case?

Tyriar: I think it's fine like this since link matchers should be going away soon.

TS wasn't as clever as I thought it was by not narrowing the options type when you enter in `hover`, which makes sense now that I think about it:

image

If people are confused about the API in the meantime, they can go into the types to view the definition and/or look at the object type in intellisense:

image

Passing the object `as ILinkProviderOptions` makes it act nicely:

image

I personally don't think it's worth adding warnings since link matchers are deprecated and it's such a simple addon API.

PR #3236, opened 2021-01-29
Search Addon: Fix length calculation of wide unicode chars

### LINKED ISSUE #1686: Selection with search and unicode

### LINKED ISSUE #3235: SearchAddon shows wrong selection when bounding with wide unicode chars
When the matched content starts right after a wide unicode char, the selection appears to start from the middle of the unicode char before it.

Screen Shot 2021-01-29 at 11 51 20

When the matched content ends with a wide unicode char, the selection appears to include only half of the char.

Screen Shot 2021-01-29 at 12 01 30

This can be reproduced in browsers and VSCode.

## Details
- Browser and browser version: Firefox 85 / Chrome 88 / VSCode 1.52.1
- OS version: macOS 10.15.7
- xterm.js version: 4.9.0

### Steps to reproduce

1. Input `中文xxx` in terminal
2. Search `xxx` for the first case
3. Search `中` for the second case

gera2ld:
Fix #1686


addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -324,34 +325,42 @@ export class SearchAddon implements ITerminalAddon {
...
-            resultIndex -= char.length - 1;
-          }
-          // Adjust the searchIndex for empty characters following wide unicode
-          // chars (eg. CJK)
-          const charWidth = cell.getWidth();
-          if (charWidth === 0) {
-            resultIndex++;
-          }
-        }
+        resultIndex = this._stringLengthToBufferSize(line, resultIndex);
```
rheh, line 331: Would that be better named something other than resultIndex?  As it is called col on the return that sound reasonable? e.g.

```
const col = this._stringLengthToBufferSize(line, resultIndex);
```

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -324,34 +325,42 @@ export class SearchAddon implements ITerminalAddon {
...
+        length = this._stringLengthToBufferSize(line, length, resultIndex);
```
rheh, line 332: Same here, is length the write variable to use maybe introduce a new one instead of reusing the one passed in as a parameter.

```
const bufferSize = this._stringLengthToBufferSize(line, length, resultIndex);
```

jerch (changes requested):
Thx for looking into this.

Some remarks form my side:
- Could you move the changes to a separate branch? That would make testing of the PR alot easier (it is currently on master :fearful:).
- Imho this still does not work correctly, repro:
  - input `￥￥￥hello ￥￥￥hello##########café éé ééécafcaféééééécafcaféééééécaf𝄞𝄞𝄞𝄞𝄞𝄞𝄞𝄞𝄞𝄞𝄞ééhello` into the terminal
  - search for `￥`, `é` and `hello`
  - result: not all occurances found, some selections are totally off

Imho this needs a more fundamental rewrite - one way to approach it would be to get rid of the regexp positioning at all and search only on buffer indices. This would eliminate the string to buffer position errors. It is abit more implementation work but should run fine perfwise as well.

Another way is to stick with initial regexp search and also to translate the beginning string offset into a buffer index. This is needed as only the very first cell of a new line is guaranteed to have the same index position in the string and the buffer (index 0). Note that this also does not apply to wrapped lines (you basically have to start the buffer index translation always from the very first char of a wrapped line, which might span several buffer rows).

gera2ld:
Sorry for the late response.

I don't think it's a good idea to get rid of the regexp positioning because in that case we can hardly search by user provided regexp.
I've added a new method to translate buffer index to string offset so that the searching process only depends on string indices and works well with regexp.

But I noticed that searching in wrapped line does not work if the line contains special chars like emojis, because the `.substring` here cuts the string at `terminal.cols`, while such emojis occupies only one column but the string length is larger than 1:

https://github.com/xtermjs/xterm.js/blob/187f072f4336bb595539a3653e07e2cc97b3d2bd/addons/xterm-addon-search/src/SearchAddon.ts#L375

I'm wondering if the `.substring` can be removed safely.

addons/xterm-addon-search/src/SearchAddon.ts
gera2ld: done

gera2ld:
>     * Could you move the changes to a separate branch? That would make testing of the PR alot easier

Sorry for that, but it seems not possible to change branch name now. I can create another PR if you want.

jerch:
@gera2ld Yes, removing regexp would be big loss in flexibility, its better find a solution that keeps working them.

> [...] I'm wondering if the .substring can be removed safely.

Hmm looking at `_translateBufferLineToStringWithWrap` as a whole imho it does not the right thing and needs a rewrite (and yes without substring) for several reasons:
1. A search term might be split over several wrapped lines.
2. A wrapped line might end earlier if the next char does not fit anymore (a wide char at `term.col - 1` would flow to the next line because of taking 2 cells leaving one cell empty).
3. Any position deduction of string index <---> buffer index has to be done via the full wcwidth translation.

Esp. the last point is tricky to get done right, as any non 1:1 mapping char will create offsets - combining chars on BMP chars will create +1 offsets on the string index, while BMP wide chars will create -1 offsets (relative to buffer index). The only fixpoint in both metrics we have are real new lines at index 0 (lines that have no following `isWrapped` flag set).

Idea to revamp `_translateBufferLineToStringWithWrap`:
- Whenever a new line is found, consume all follow-up wrapped lines into one single (unwrapped) string. Thats pretty much what it does atm, but without substring. This will also correctly deal with 1. above.
- Special case early wrapping wide char (2. above): If the first char on the next line is a wide char and the current line is only `term.cols - 1` in length, treat it as early wrapping - ignore the empty cell, otherwise count empty cell as whitespace. This leads to a faulty handling, if the empty cell was indeed meant to stay empty. We already had an issue report about that, but sadly we have currently no way to track that special edge case.
- Whenever some match was found in the full (unwrapped) string, translate all string positions (start and end index) back by walking buffer positions + wcwidth continuing from the last saved match buffer position. This "buffer replay" will ensure 3. above.

> Sorry for that, but it seems not possible to change branch name now. I can create another PR if you want.

Hmm I think changing a PRs source branch is not possible. Guess creating a new one is the only option.

gera2ld:
Update: 
- the string offsets of the first char in each buffer row as well as the string of the whole line is cached now, so the buffer column can be calculated easily from the beginning of the corresponding row, no matter whether it's early wrapped or for whatever reason the string length does no match the buffer columns.

jerch:
Please review someone else, as it seems I wont get down to it.

Tyriar (approved):
Thanks @gera2ld and sorry about how long this took to merge in. I made a few minor doc related changes but since the tests pass and changes seem reasonable lgtm.

JasinYip:
This is the best news for me for Christmas and the New year, thank you all you guys fix this!

PR #3243, opened 2021-02-05
Force char size measurement when devicePixelRatio changes

jeanp413:
Fixes https://github.com/microsoft/vscode/issues/102194

Note that `this._charSizeService.measure()` could trigger a call to `this._renderer.onCharSizeChanged()` which uses the same logic as `this._renderer.onDevicePixelRatioChange()` underneath


Tyriar (approved):
Looks good, seems safe as well since `ICharSizeService.onCharSizeChange` only fires if they do actually differ

PR #3253, opened 2021-03-18
Handle webgl context loss 

### LINKED ISSUE #2816: Handle general canvas context loss
Contexts can become lost when graphics card drivers are reinstalled for example. This should be handled generically.

VS Code issue: https://github.com/microsoft/vscode/issues/94513

Related: https://github.com/xtermjs/xterm.js/issues/2253

meganrogge:
fix #2816


meganrogge:
We discussed that instead we will fire an event when the context is lost that vscode will listen to. When that happens, vscode will fall back to the canvas renderer type.

Tyriar:
Handled in https://github.com/xtermjs/xterm.js/pull/3276

PR #3256, opened 2021-03-19
silence writeSync warning

### LINKED ISSUE #3255: writeSync error is a bit too loud
I think we should use the log service so this doesn't surface in devtools unless logs are enabled:

https://github.com/xtermjs/xterm.js/blob/145b4874cb217d7c23b46c731a2bb9a780b66c3f/src/common/CoreTerminal.ts#L138

This happens in vscode right now and I wasn't planning on removing it yet as it will be a bit of a pain to get rid of it in the reconnect logic:

Screen Shot 2021-03-18 at 5 23 11 AM

jerch:
Fixes #3255.

### DIFF
--- a/src/common/CoreTerminal.ts
+++ b/src/common/CoreTerminal.ts
@@ -22,7 +22,7 @@
  */
 
 import { Disposable } from 'common/Lifecycle';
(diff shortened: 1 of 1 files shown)


Tyriar (approved):
Thanks 🙂

PR #3257, opened 2021-03-19
Change warning to one per session not terminal

Tyriar:
Slight improvement, it still happens once per terminal


jerch (approved):
:ok_hand: :see_no_evil:

PR #3259, opened 2021-03-20
feat: Improve ligatures addon

QuiiBz:
This PR adds a new `ILigaturesOptions` to specify a custom file protocol, used to load the fonts on the computer.

The commit https://github.com/xtermjs/xterm.js/commit/e23153635d51f7fd704a9ed116617f90c289e5fe also fixes a bug in the `parse` function:
- If you use only one font with single or double-quotes, the font is only parsed but not pushed to the `families` array:
Example:
```js
parse('"FiraCode Nerd Font"')
```
returns an empty array, where
```js
parse('"FiraCode Nerd Font", monospace')
```
returns an array with both `FiraCode Nerd Font` and `monospace`.
- This issue is fixed here, we push a font if one was found but wasn't added to the array: https://github.com/xtermjs/xterm.js/blob/e23153635d51f7fd704a9ed116617f90c289e5fe/addons/xterm-addon-ligatures/src/parse.ts#L56-L59

### DIFF
--- a/addons/xterm-addon-ligatures/src/LigaturesAddon.ts
+++ b/addons/xterm-addon-ligatures/src/LigaturesAddon.ts
@@ -11,11 +11,15 @@ export interface ITerminalAddon {
   dispose(): void;
 }
 
+export interface ILigaturesOptions {
+  protocol?: string;
+}
+
 export class LigaturesAddon implements ITerminalAddon {
   constructor() {}
 
-  public activate(terminal: Terminal): void {
-    enableLigatures(terminal);
+  public activate(terminal: Terminal, options?: ILigaturesOptions): void {
+    enableLigatures(terminal, options);
   }
 
   public dispose(): void {}
--- a/addons/xterm-addon-ligatures/src/LigaturesAddon.ts
+++ b/addons/xterm-addon-ligatures/src/LigaturesAddon.ts
@@ -16,10 +16,13 @@ export interface ILigaturesOptions {
 }
 
 export class LigaturesAddon implements ITerminalAddon {
-  constructor() {}
+  constructor(
+    private _options?: ILigaturesOptions
+  ) {
+  }
 
-  public activate(terminal: Terminal, options?: ILigaturesOptions): void {
-    enableLigatures(terminal, options);
+  public activate(terminal: Terminal): void {
+    enableLigatures(terminal, this._options);
   }
 
   public dispose(): void {}
--- a/addons/xterm-addon-ligatures/src/font.ts
+++ b/addons/xterm-addon-ligatures/src/font.ts
@@ -6,6 +6,7 @@
 import * as fontFinder from 'font-finder';
 import * as fontLigatures from 'font-ligatures';
 
+import { ILigaturesOptions } from './LigaturesAddon';
 import parse from './parse';
 
 let fontsPromise: Promise<fontFinder.FontList> | undefined = undefined;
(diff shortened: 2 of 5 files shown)


QuiiBz:
The tests error are caused by timeouts, is it normal?

Tyriar (commented):
I'm not sure we want to go in this direction considering we're moving towards relying on the browser's font access API to get fonts https://github.com/xtermjs/xterm.js/pull/3264

QuiiBz:
I’m not sure if it’s possible to enable the chrome flag (`#font-access`) in electron. If yes, does the ligatures addon will be completely removed?

LabhanshAgrawal:
> I’m not sure if it’s possible to enable the chrome flag (`#font-access`) in electron. If yes, does the ligatures addon will be completely removed?

It can be done using `enableBlinkFeatures: 'FontAccess'` in BrowserWindow constructor options.

QuiiBz:
> I'm not sure we want to go in this direction considering we're moving towards relying on the browser's font access API to get fonts #3264

Then should we close this PR?

Tyriar:
Yes let's close this, thanks for the pr though.

LabhanshAgrawal:
> The commit [e231536](https://github.com/xtermjs/xterm.js/commit/e23153635d51f7fd704a9ed116617f90c289e5fe) also fixes a bug in the `parse` function:
> 
> * If you use only one font with single or double-quotes, the font is only parsed but not pushed to the `families` array:
>   Example:
> 
> ```js
> parse('"FiraCode Nerd Font"')
> ```
> 
> returns an empty array, where
> 
> ```js
> parse('"FiraCode Nerd Font", monospace')
> ```
> 
> returns an array with both `FiraCode Nerd Font` and `monospace`.


@Tyriar @QuiiBz What about the parse bug mentioned above? Didn't look into it much, but this should be included, right?

QuiiBz:
You are right, I forgot about this issue. The fix is pretty easy as you can see here:
https://github.com/xtermjs/xterm.js/commit/e23153635d51f7fd704a9ed116617f90c289e5fe#diff-af81cfaa712b74298aa772bcae4571c997ec176cb3ad6c1353a4e5ae3b3994b7R56-R60

jerch:
@QuiiBz The fix will still miss the last family, if it is a quoted one. The issue results from loop, where `fontFamily` gets only pushed on the separator `,`. Suggestion:
```typescript
export default function parse(stringFamily: string): string[] {
  if (typeof stringFamily !== 'string') {
    throw new Error('Font family must be a string');
  }

  const context: IParseContext = {
    input: stringFamily,
    offset: 0
  };

  const families: string[] = [];
  let familyFromString = '';

  // Work through the input character by character until there are none left.
  // This lexing and parsing in one pass.
  while (context.offset < context.input.length) {
    const char = context.input[context.offset++];
    switch (char) {
      // String
      case '\'':
      case '"':
        if (familyFromString) {
          throw new Error('invalid declaration');
        }
        familyFromString = parseString(context, char);
        break;
      case ',':
        if (familyFromString && !families.includes(familyFromString)) {
          families.push(familyFromString);
        }
        familyFromString = '';
        break;
      default:
        // Identifiers (whitespace between families is swallowed)
        if (!/\s/.test(char)) {
          if (familyFromString) {
            throw new Error('invalid declaration');
          }
          context.offset--;
          const family = parseIdentifier(context);
          if (!families.includes(family)) {
            families.push(family);
          }
        }
    }
  }

  if (familyFromString && !families.includes(familyFromString)) {
    families.push(familyFromString);
  }

  return families;
}
```

This version also identifies declarations like `"Lucida" Grande, serif` as invalid. Furthermore it filter duplicates (not sure if thats a good idea, just cant see any use case for a later redeclaration of an already known entry).

Tyriar:
If we can get a good failure case and add some tests for it we should fix the parsing. Ideally it would align with browser parsing https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#formal_syntax

PR #3264, opened 2021-03-24
Enable using ligatures addon outside electron

LabhanshAgrawal:
Based on the discussion from #958 
To test this in the demo make sure to enable `#font-access` in `chrome://flags`
And then set some font with ligatures
Creating as draft right now as I still need to look a bit into the webpack stuff.

### DIFF
--- a/addons/xterm-addon-ligatures/package.json
+++ b/addons/xterm-addon-ligatures/package.json
@@ -32,7 +32,7 @@
   "license": "MIT",
   "dependencies": {
     "font-finder": "^1.1.0",
-    "font-ligatures": "^1.3.3"
+    "font-ligatures": "^1.4.0"
   },
   "devDependencies": {
     "@types/sinon": "^5.0.1",
--- a/addons/xterm-addon-ligatures/src/font.ts
+++ b/addons/xterm-addon-ligatures/src/font.ts
@@ -3,8 +3,8 @@
  * @license MIT
  */
 
-import {FontList} from 'font-finder';
-import {Font, loadBuffer, loadFile} from 'font-ligatures';
+import { FontList } from 'font-finder';
+import { Font, loadBuffer, loadFile } from 'font-ligatures';
 
 import parse from './parse';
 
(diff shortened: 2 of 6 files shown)


Tyriar:
This didn't seem to work for me with Fira Code and any of the 3 renderers (I think only dom and maybe canvas works with ligatures atm?). Any hints to get it to work? I enabled the chrome flag.

LabhanshAgrawal:
After checking this out you'd have to do a `yarn` install at the root as there's an update to the dependencies. (In case you haven't done this already)

Also, I guess I need to look into the tests.

LabhanshAgrawal:
Also, you'd need to relaunch chrome after changing the flag. (I don't remember if it asks for relaunch itself)
Are getting any error on the console?

LabhanshAgrawal:
@Tyriar did it work for you?

Tyriar:
There was an error in console:

> SecurityError User activation is required.

No prompt though.

I tried disabling ligatures by default so I can check the box and then I get this:

Screen Shot 2021-04-02 at 6 49 33 AM

Clicking allow didn't work though, even after reloading and trying again.

Tyriar:
Screen Shot 2021-04-02 at 6 50 57 AM

Tyriar:
Oh actually, it does work after that and setting the font. The prompt is a bit of a pain though that it needs user action 🤔

Tyriar:
😍 

Screen Shot 2021-04-02 at 6 53 27 AM

Tyriar (changes requested):

LabhanshAgrawal:
Will add the comments as you mentioned.
Expect a bit slow progress on the tests and webpack stuff though, as I'm caught up in a lot of work recently.
Great to see that I didn't have to use `It works on my machine though` 😜

LabhanshAgrawal:
The prompt doesn't come after the first time AFAIK so that should be ok.

Tyriar:
> The prompt doesn't come after the first time AFAIK so that should be ok.

Yep, we'd want to keep it off by default and call this out in the readme though that it may cause a permissions prompt and will require to be hooked up to a "user event".

Tyriar:
The tests hang because loading the font fails

Screen Shot 2021-04-02 at 8 08 19 AM

Tyriar (approved):
Thanks for the hard word 🙂

LabhanshAgrawal:
Thanks for resolving the tests and merging it 😊.
I hope this will help speed up development on the ligatures addon.
Cheers

PR #3266, opened 2021-03-26
Prevent stack overflow by firing writeSync while writing

### LINKED ISSUE #3265: Stack Overflow exception when writeSync is called in a data handler
We hit this in VS Code because our local echo feature manipulates process data and uses `writeSync('')` to flush the buffer. I know this is a hack in and of itself and plan to fix it eventually, but xterm.js should protect against this Stack Overflow regardless.

Tyriar:
Calling writeSync from the onData API event causes a stack overflow exception to
occur because _bufferOffset only gets incremented after the write callback is
fired. One approach to fix this is to just ensure _bufferOffset is incremented
beforehand, but I think it's still generally a bad idea to allow multiple writes
at the same time. So instead we just block them and return immediately if a
write is already in progress, regardless of whether it was sync or not.

Fixes #3265

@jerch please review 🙂

### DIFF
--- a/src/common/input/WriteBuffer.ts
+++ b/src/common/input/WriteBuffer.ts
@@ -42,6 +42,7 @@ export class WriteBuffer {
   private _callbacks: ((() => void) | undefined)[] = [];
   private _pendingData = 0;
   private _bufferOffset = 0;
+  private _isWriting = false;
 
   constructor(private _action: (data: string | Uint8Array, promiseResult?: boolean) => void | Promise<boolean>) { }
 
@@ -49,6 +50,12 @@ export class WriteBuffer {
    * @deprecated Unreliable, to be removed soon.
    */
   public writeSync(data: string | Uint8Array): void {
+    // Ensure a write is not already in progress
+    if (this._isWriting) {
+      return;
+    }
+    this._isWriting = true;
+
     // force sync processing on pending data chunks to avoid in-band data scrambling
     // does the same as innerWrite but without event loop
     if (this._writeBuffer.length) {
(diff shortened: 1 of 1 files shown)


jerch (changes requested):
Oh well yeah, `writeSync` was just a hack to be called from internal tests in sync to apply and flush the write buffer, not to drive frontend cycles. Calling it from `onData` over and over will show bad side effects:
1. endless call recursion as `onData` --> `writeSync` --> `onData` --> ...
2. loop inconveniences in `writeSync`
  - endless growing write buffer with possible OOM (chunk added before being removed after the loop)
  - unclear loop behavior due to altering loop subject (many languages forbid that), I think JS does not fail here, since the loop does a `.length` check and chunks being added at the end
3. no screen updates due to all being sync code

Your patch can successfully address 1. but will still fail for 2. To also address 2. I suggest to reshape the loop to a while-pop one. Avoiding 3. makes no sense at all, using `writeSync` was meant to skip screen updates in the first place, to avoid that, simply dont use `writeSync`.

There is another issue with your patch - it currently will skip continuation of paused async handlers (thats a real error condition for the parser) due to exiting early in `_innerWrite` at `_isWriting == true`. Imho that condition is not needed in `_innerWrite` at all, since the sync code of `writeSync` will bypass any `_innerWrite` calls. (Thats the reason for the failing parser API tests.)

I can make the needed changes if you want (just drop me a note). Plz keep in mind that using `writeSync` is dangerous as soon as you have the first async handler registered.

Tyriar:
@jerch if you could make the changes that would be great since you're most familiar with this.

> Plz keep in mind that using writeSync is dangerous as soon as you have the first async handler registered.

Yes we definitely don't use any async handlers yet and I want to get rid of writeSync, unfortunately local echo and persistent terminals are built on it and we certainly can't change in the version since we're stabilizing right now.

Tyriar:
Closing in favor of https://github.com/xtermjs/xterm.js/pull/3267

PR #3267, opened 2021-03-26
writeSync callstack fix

### LINKED ISSUE #3265: Stack Overflow exception when writeSync is called in a data handler

jerch:
Fixes #3265.

@Tyriar
The fix introduces another optional argument `maxSubsequentCalls` to limit further `writeSync` subcalls. Usage:
- undefined: Will loop forever, if subsequent `writeSync` calls add chunks. Imho thats the correct behavior, as your caller code asked for that. This time without any overflow. :smile_cat: 
- 0: Run once, any subsequent call will exit early discarding the actual chunk.
- 1: Allow one subcall. And so on...

Note that the last chunk is always lost if you have to use `maxSubsequentCalls` (should not contain important data).

### DIFF
--- a/src/common/CoreTerminal.ts
+++ b/src/common/CoreTerminal.ts
@@ -137,12 +137,12 @@ export abstract class CoreTerminal extends Disposable implements ICoreTerminal {
    *
    * @deprecated Unreliable, will be removed soon.
    */
-  public writeSync(data: string | Uint8Array): void {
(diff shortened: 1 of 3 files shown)


Tyriar (approved):
Thanks for the quick fix 🙂

Issue #1852, opened 2018-12-21
Allow embedders to decorate ranges in the buffer


Tyriar:
Creating this to centralize discussion on @mofux's idea to support monaco-like decorations as discussed in https://github.com/xtermjs/xterm.js/issues/1380#issuecomment-379848832 and https://github.com/xtermjs/xterm.js/issues/791#issuecomment-448372516.

Tyriar:
Was just thinking, once https://github.com/xtermjs/xterm.js/issues/1176 is done if we had some view part that could be inserted then the 1337 iterm sequence to insert an image could be completely implemented within an addon.

sdegutis:
@Tyriar I'm particularly interested in an xterm.js feature where plugins can get DOM elements added into the "flow" of the terminal grid and even between rows, which the plugin can then completely control by appending interactive elements into it. Thanks to your comment to the issue I just opened, I found and have been reading up on a lot of discussion between you, @mofux, and @PerBothner, and I see that xterm.js is very well positioned to be able to add this feature, but that it's not quite there yet. I may be able to allocate several days or perhaps even weeks of full time work towards helping to implement this feature within xterm.js, pending we can come up with a reasonable and concrete plan, and can come up with a way to coordinate effectively on it. Please let me know your thoughts. Thanks for your time :)

Tyriar:
@sdegutis glad to hear you're interested in contributing! It's been a while since we talked about this but here are some thoughts:

- We probably don't want to  get into the business of inserting elements in between rows as that would involve a lot of changes in how things are rendered and also mess with the interactive viewport at the bottom of the buffer
- I'm not too familiar with the Monaco decoration API but it's tried and true and worth taking inspiration from. Note that unlike in Monaco, we don't necessarily have DOM elements for the cells (depends on rendererType)
- I'm not sure what the right behavior would be when rows/cell with decorations get resized and get wrapped, I guess the embedder is responsible for listening to `Terminal.onResize` and reacting/recreating for that as appropriate

Given this, my current thinking is that we could maybe have a couple of flavors of decorations:

- Arbitrary DOM elements whose positioning is handled by xterm.js, for example you could use an `<img>` to decorate some cell, when the user scrolls the viewport then xterm.js will position that (with the help of `Marker`s). This type would probably need an extra container with `overflow:hidden` such that a decoration could span multiple rows. 
- Changing the background color of a cell, this would need to be handed over to the renderer internally to override the regular background color so it's drawn underneath the text.

[...]

PR #3274, opened 2021-03-31
Visual marker at end of the selection and selectionStyle option to Terminal

akariv:
We introduce a new configuration option - `selectionStyle` - used to style the text selection layer.
It has three 'modes': `plain` (the current default), `mark-end` (add a marker to the end) and `mark-start` (add a marker to the beginning of the selection range). The styling of the marker is taken from the theme itself.

The use case here is when integrating with applications that need to allow manipulating either end of the selection range via keyboard, and we need a way to show the user the currently 'selected' end.

mark-start:    Screen Shot 2021-03-31 at 18 43 09

mark-end:     Screen Shot 2021-03-31 at 18 43 17

### DIFF
--- a/src/browser/public/Terminal.ts
+++ b/src/browser/public/Terminal.ts
@@ -174,7 +174,7 @@ export class Terminal implements ITerminalApi {
   public paste(data: string): void {
     this._core.paste(data);
   }
-  public getOption(key: 'bellSound' | 'bellStyle' | 'cursorStyle' | 'fontFamily' | 'logLevel' | 'rendererType' | 'termName' | 'wordSeparator'): string;
+  public getOption(key: 'bellSound' | 'bellStyle' | 'cursorStyle' | 'fontFamily' | 'logLevel' | 'selectionStyle' | 'rendererType' | 'termName' | 'wordSeparator'): string;
   public getOption(key: 'allowTransparency' | 'altClickMovesCursor' | 'cancelEvents' | 'convertEol' | 'cursorBlink' | 'disableStdin' | 'macOptionIsMeta' | 'rightClickSelectsWord' | 'popOnBell' | 'visualBell'): boolean;
   public getOption(key: 'cols' | 'fontSize' | 'letterSpacing' | 'lineHeight' | 'rows' | 'tabStopWidth' | 'scrollback'): number;
   public getOption(key: 'fontWeight' | 'fontWeightBold'): FontWeight;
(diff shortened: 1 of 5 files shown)


jerch:
Thats an interesting feature. When you write 

> ...to allow manipulating either end of the selection range via keyboard, ...

Does that mean you have to move the cursor? Or how do you do that? If it involves cursor movements, how do you go about selections in the scrollback (not addressible by the cursor)? And how do you sync that with the application?

Sorry for so many questions, maybe you could explain it and the actual use cases abit more in detail?

daonb:
hi @jerch, this is a feature we need for Terminal 7. Adam is a friend who  loves TypeScript so his helping me with one of the key features we're missing for 1.0 - tmux style copy mode : tuzig/terminal7#66 .

In the past I tried and failed in developing this feature. The cursor is stuck on the bottom page and that's the way xterm.js is. Adam had a great idea: to let the cursor be and use the selection start|end instead. This PR is just a part of the feature, the other being a copy-mode plugin to use this new option. Here [is the code](https://github.com/akariv/terminal7/blob/master/src/copy-mode/copymode-addon.ts), it's still under works and part of terminal7 but not for long.

Tyriar:
I have a few concerns with this:

- Not a fan of introducing `selectionStyle` setting, it's very specific to Terminal 7's particular case and I doubt anyone else would end up using it
- It's only implemented in the canvas renderer, it would need to be added to both DOM and webgl as well if we did this

This is awfully similar to something I've wanted xterm.js to do for a while which is to just allow manipulating a cursor (not the actual cursor) within the buffer to enable:

- Navigating and selecting via the keyboard
- Reading of the character/word for screen readers

You would enter such a mode somehow and then exit it via escape. So I'd be all in for getting that or something that works towards that in.

---

I'd like to propose an alternative to this change, all you seem to need is to be able to mark the start and end of the selection, essentially allowing embedders/addons to add decorations to parts of the buffer which we have an open issue on: https://github.com/xtermjs/xterm.js/issues/1852

There's some discussion in that issue but my basic thinking would be to have an overlay element (`.xterm-decorations`) that uses `pointer-events: none` and allows addons to decorate certain ranges in the buffer. For your case this would involve decorating the range and then filling in the style to what ever you want, a very quick example:

```ts
const s = term.getSelectionPosition();
const decoration = term.decorateRange({ 
  x: s.startColumn,
  y: s.endColumn
});
decoration.element.style.border = '1px solid #fff';
```

Some thought would need to be given to the exact shape of the API, but the big benefits here is that is aligns with our goals and it would end up being used by several embedders (including VS Code, eg https://github.com/microsoft/vscode/issues/46561).

akariv:
Yes, implementing this as an add-on would have been my first preference - but modifying the style of the buffer seemed to be impossible with current API.

I'll take a stab at implementing the decorations API (or a first draft of it, at least)

Tyriar:
@akariv I'd recommend approaching it in a similar way to how `RenderService` works:

- Create a `DecorationService` in the style of other services inside the `browser/` layer
- Pass a `decorationElement` in when it's instantiated
- Rely on `IBufferService` so you can listen to `onScroll` and ask decorations to update

akariv:
Closing this as we've opted for another approach.

PR #3276, opened 2021-04-01
add onContextLoss to webgl addon

### LINKED ISSUE #2253: Webgl: Handle context loss when there are too many contexts
Webgl addon added in https://github.com/xtermjs/xterm.js/pull/1790

See https://github.com/Tyriar/xterm.js/pull/5 and the associated comments

meganrogge:
So that VS code can fall back to dom when the webgl renderer loses context

Fixes #2253


Tyriar (commented):
Should we call this _onContextLoss instead and pass the actual event over so it can be prevented if desired?

Tyriar (approved):
Good to go after adding the API to https://github.com/xtermjs/xterm.js/blob/master/addons/xterm-addon-webgl/typings/xterm-addon-webgl.d.ts

Tyriar:
Also I added a bit to the readme on handling context loss, improvements welcome 🙂

PR #3279, opened 2021-04-01
Don't pad powerline glyphs and use ideographic text baseline

### LINKED ISSUE #3278: Powerline glyphs don't render nicely in the webgl addon
In xterm.js demo

Dom:

dom

Canvas:

canvas

Webgl:

webgl

### LINKED ISSUE #3281: Webgl and canvas renderers are not correctly aligned
Looks like ideographic textBaseline might do the trick. https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textBaseline

Tyriar:
Fixes #3278
Fixes #3281

Webgl before:

image

Webgl after:

image

Canvas before:

image

Canvas after:

image


Tyriar:
Created https://github.com/xtermjs/xterm.js/wiki/Powerline-fonts for future reference on printing them out without changing your prompt

PR #3290, opened 2021-04-03
Fix #3014  - Add onBell event listener to allow embeders to hook into it

### LINKED ISSUE #3014: Support event when the bell goes off

nikonso:
WIP to fix #3014 based on the discussion in #3091

I'm not sure if the tests are okay but from my local testing it seems to work

recording (3)

### DIFF
--- a/src/browser/Terminal.test.ts
+++ b/src/browser/Terminal.test.ts
@@ -131,6 +131,12 @@ describe('Terminal', () => {
       });
       term.write('\x1b]2;title\x07');
     });
+    it('should fire the onBell event', (done) => {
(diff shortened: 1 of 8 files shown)


Tyriar (approved):
Thanks @nikonso, works great just needed some test tweaks 👍

PR #3298, opened 2021-04-06
Fix #2714 - Start demo server before running integration tests

### LINKED ISSUE #2714: Have integration tests run their own server on a port other than 3000
Right now the tests rely on the demo being up and running on port 3000, it would be nicer if instead the tests set up their own server on a different port and tore it down after running the tests.

bmf-ribeiro:
Fixes #2714 
- Start demo server before running integration tests
- Use port 3001 instead of  3000 to not conflict with the demo

### DIFF
--- a/addons/xterm-addon-attach/test/AttachAddon.api.ts
+++ b/addons/xterm-addon-attach/test/AttachAddon.api.ts
@@ -7,7 +7,7 @@ import WebSocket = require('ws');
 import { openTerminal, pollFor, getBrowserType } from '../../../out-test/api/TestUtils';
 import { Browser, Page } from 'playwright';
 
-const APP = 'http://127.0.0.1:3000/test';
+const APP = 'http://127.0.0.1:3001/test';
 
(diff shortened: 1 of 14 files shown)


Tyriar (changes requested):
This is great!

We can also remove the hacky server starts in the linux and mac integration tests jobs, reducing CI time by ~10 seconds:

https://github.com/xtermjs/xterm.js/blob/26cda84eabaa3b61f49d0a56f7aad72ebfc7aada/azure-pipelines.yml#L103-L106

nikonso:
Okay, that makes sense, should the integration tests start running on windows as well, then?

Tyriar:
It would be good to run windows too, I forgot why they were disabled though. Can try enable in a separate pr and see what happens?

nikonso:
This is ready to review.

Also I opened #3299 the tests ssem to be running okay on windows.

Tyriar (approved):
🎉

PR #3299, opened 2021-04-06
Feat/enable windows integration tests

bmf-ribeiro:
Try to enable windows integration tests


Tyriar (approved):
Thanks, if this passes we'll merge and then monitor how it goes.

PR #3317, opened 2021-05-05
Fix search cache when a trim occurs

### LINKED ISSUE #3262: Search cache seems to break when content is changing during a search
Notice how "test" is not being selected sometimes

recording (4)

Tyriar:
This introduced the dependency [`@tyriar/avl-tree`](https://www.npmjs.com/package/@tyriar/avl-tree) to search so that we don't need to implement our own custom keyed binary tree structure. It's a dev dep since we webpack before publish.

Performance is about the same since we're now looking up lines in O(log n) with the AVL tree instead of O(1) with the plain array.

Fixes #3262


Tyriar:
Closing as there were problems here.

PR #3330, opened 2021-05-10
update package dependencies

jerch:
Update build packages to newer versions. This was needed to get rid of a security warning.

@Tyriar I also updated `@types/node` to the 12.x branch, I think we should transit to v12 as base development node version as v10 reached EOL last month.


jerch:
Few more notes:
- Not sure why I had to change the eslint array rule, the rule was in place before but kinda ignored by older version? I went with `SomeType[]` for normal arrays, and `ReadonlyArray<SomeType>` for readonly, as it reflects our current code base.
- Holded `ts-loader@8.2.0` as it is the last working version with node v10. This makes transition to v12 easier (technically the whole repo still builds under v10, but this changes as soon as we use v12+ symbols).
- Not sure about the TS version, I simply went with the newest one.
- @Tyriar You might want to test if the packaging still work correctly, did not test that. Everything else kept working as before.

Tyriar (approved):
Looks good but let's hold off until we do the 4.12.0 release. I don't see why packaging wouldn't work but we can confirm via CD/beta builds

Tyriar:
Also we could move to node 14 too if it's not a big deal since it's LTS?

jerch:
@Tyriar Also runs smooth with v14, and we are on the newest package versions for all stuff (beside `@types/node` being tagged to 14.x) :smile_cat:

PR #3332, opened 2021-05-11
Make sure all rows are refreshed on input

### LINKED ISSUE #3323: DOM renderer doesn't refresh screen on input
Repro:

1. Enable dom renderer in demo
2. Full viewport + more
3. Scroll up with wheel
4. Press up, only the last line will rerender

Downstream: https://github.com/microsoft/vscode/issues/123291

Tyriar:
Fixes #3323


Tyriar:
I think the scroll event recursion is now handled by the scroll source, the event ignoring seems to be just causing issues now.

PR #3334, opened 2021-05-12
chore(types): use IEvent from xterm

UziTech:
Typescript complains when using `node-pty-prebuilt-multiarch` instead of `node-pty`.

```
node_modules/xterm-addon-webgl/typings/xterm-addon-w
ebgl.d.ts:6:24 - error TS2307: Cannot find module 'n
ode-pty' or its corresponding type declarations.

6 import { IEvent } from 'node-pty';
                         ~~~~~~~~~~


Found 1 error.
```

https://github.com/atom-community/terminal/pull/72#issuecomment-839361620


aminya (approved):

jerch (approved):
Whoops, thanks for the fix, `node-pty` definitely does not belong here...

PR #3336, opened 2021-05-12
fix 4.12 - chore(types): use IEvent from xterm

jerch:
(cherry picked from commit 7a0b462e7eedbf8824652409bce325afd19f16c2)

### DIFF
--- a/addons/xterm-addon-webgl/typings/xterm-addon-webgl.d.ts
+++ b/addons/xterm-addon-webgl/typings/xterm-addon-webgl.d.ts
@@ -3,8 +3,7 @@
  * @license MIT
  */
 
-import { IEvent } from 'node-pty';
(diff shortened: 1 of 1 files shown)


jerch:
@Tyriar Plz watchout, the target branch still points to `master`, as I am not allowed to directly create another branch. Not sure if this PR is enough to trigger the whole build.
Furthermore only the webgl addon package is affected, would still be good to have something like a `4.12.1` tag, which contains working code as a whole.

If this all does not work, we could also rewind all on master (2 PRs only) and add it sequentially.

jerch:
@Tyriar Ok got the target branch fixed by force pushing a new branch `hotfix_4_12` from tag `4.12.0`.

Tyriar (approved):

PR #3341, opened 2021-05-14
chore: lint using putout

coderaiser:
After successfully merged linting sessions [part1](https://github.com/xtermjs/xterm.js/pull/2953), [part2](https://github.com/xtermjs/xterm.js/pull/3133) and [part3](https://github.com/xtermjs/xterm.js/pull/3269), the time is come for part 4 :).

As usual, any rule can be disabled.
Command used:

```sh
putout {src,addons}/**/*.ts --fix
```

Applied rules:
- [apply-as-type-assertions](https://github.com/coderaiser/putout/tree/v17.4.0/packages/plugin-apply-as-type-assertions)
- [simplify-ternary](https://github.com/coderaiser/putout/tree/v17.4.0/packages/plugin-simplify-ternary)
- [remove-useless-async](https://github.com/coderaiser/putout/tree/v17.4.0/packages/plugin-remove-useless-async)
- [remove-useless-operand](https://github.com/coderaiser/putout/tree/v17.4.0/packages/plugin-remove-useless-operand)

Current config for [putout v17](https://github.com/coderaiser/putout):

```json
{
    "rules": {
        "apply-shorthand-properties": "on",
        "apply-destructuring": "off",
        "apply-numeric-separators": "off",
        "convert-assignment-to-comparison": "off",
        "convert-math-pow": "off",
        "strict-mode": "off",
        "remove-useless-spread/object": "off",
        "remove-useless-array-constructor": "off",
        "remove-boolean-from-assertions": "off",
        "remove-iife": "off",
        "remove-console": "off",
        "merge-if-statements": "off",
        "promises/add-missing-await": "off"
    },
    "match": {
        "SerializeAddon.benchmark.ts": {
            "remove-unused-expressions": "off"
        }
    },
    "plugins": [
        "apply-shorthand-properties"
    ]
}
```


src/common/Clone.ts
```diff
@@ -16,7 +16,7 @@ export function clone<T>(val: T, depth: number = 5): T {
 
   for (const key in val) {
     // Recursively clone eack item unless we're at the maximum depth
-    clonedObject[key] = depth <= 1 ? val[key] : (val[key] ? clone(val[key], depth - 1) : val[key]);
+    clonedObject[key] = val[key] || clone(val[key], depth - 1);
```
Tyriar, line 19: Looks like this change is a little too aggressive and changes the behavior

src/browser/Terminal.ts
```diff
@@ -1014,7 +1014,7 @@ export class Terminal extends CoreTerminal implements ITerminal {
   protected _keyDown(event: KeyboardEvent): boolean | undefined {
     this._keyDownHandled = false;
 
-    if (this._customKeyEventHandler && this._customKeyEventHandler(event) === false) {
+    if (this._customKeyEventHandler && !this._customKeyEventHandler(event)) {
```
Tyriar, line 1017: Not sure about this rule, typically I would explicitly use `false` when I think it's called for. This particular function is provided by embedders so it's a little more explicit when checking - perhaps internally it should return `unknown` 🤔

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -105,7 +105,7 @@ class StringSerializeHandler extends BaseSerializeHandler {
   }
 
   protected _beforeSerialize(rows: number, start: number, end: number): void {
-    this._allRows = new Array<string>(rows);
+    this._allRows = Array<string>(rows);
```
Tyriar, line 108: Not a fan of removing the `new`

Tyriar (changes requested):

src/browser/Terminal.ts
coderaiser: OK, I'll better disable the rule `remove-boolean-from-assertions`

addons/xterm-addon-serialize/src/SerializeAddon.ts
coderaiser: Not problem, at all, I'll disable `remove-useless-array-constructor` :).

src/common/Clone.ts
coderaiser: About this code, we can rewrite it like this:

```js
const clone = () => {
    if (depth <= 1) {
        return val[key];
    } else {
        if (val[key]) {
            clone(val[key], depth - 1);
       } else {
           return val[key];
      }
    }
}
```

And simplify like this:

```js
const clone = (val, key, depth) => {
    if (val[key] && depth > 1)
       return clone(val[key], depth - 1);
    
    return val[key];
}
```

Yes, you absolutely right!

Just fixed `simplify-ternary`, will update PR.

coderaiser:
Just disabled rules:
-  [remove-useless-array-constructor](https://github.com/coderaiser/putout/tree/v17.4.0/packages/plugin-remove-useless-array-constructor)
- [remove-boolean-from-assertions](https://github.com/coderaiser/putout/tree/v17.4.0/packages/plugin-remove-boolean-from-assertions)

And fixed:
- [simplify-ternary](https://github.com/coderaiser/putout/tree/v17.4.0/packages/plugin-simplify-ternary)(`@putout/plugin-simplify-ternary@v2.5.1`)

And updated `PR`, thanks a lot for quick response @Tyriar :)!

src/common/Clone.ts
```diff
@@ -16,7 +16,7 @@ export function clone<T>(val: T, depth: number = 5): T {
 
   for (const key in val) {
     // Recursively clone eack item unless we're at the maximum depth
-    clonedObject[key] = depth <= 1 ? val[key] : (val[key] ? clone(val[key], depth - 1) : val[key]);
+    clonedObject[key] = depth <= 1 ? val[key] : (val[key] || clone(val[key], depth - 1));
```
Tyriar, line 19: This still isn't quite right as it clones the key if it's there

src/browser/TestUtils.test.ts
```diff
@@ -415,13 +415,13 @@ export class MockRenderService implements IRenderService {
...
   public serviceBrand: undefined;
-  public register(handler: (text: string) => [number, number][]): number {
+  public register(handler: (text: string) => Array<[number, number]>): number {
     return 0;
   }
   public deregister(joinerId: number): boolean {
     return true;
   }
-  public getJoinedCharacters(row: number): [number, number][] {
+  public getJoinedCharacters(row: number): Array<[number, number]> {
```
Tyriar, line 424: We prefer [] to Array<>, I think this is an eslint rule but not sure if we use it

Tyriar (changes requested):

src/common/Clone.ts
coderaiser: You right! Fixed [simplify-ternary](
https://github.com/coderaiser/putout/tree/master/packages/plugin-simplify-ternary )([@putout/plugin-simplify-ternary@2.5.2](https://github.com/coderaiser/putout/commit/1ab404f33bf8ae89ab2933868f810326292aa622#diff-0ce4998d3f3a3d0ab4c56865edd5c0733ff26196f8c1f39bf14ca27e5e1386bd)).

src/browser/TestUtils.test.ts
coderaiser: Looks like it's related to https://github.com/xtermjs/xterm.js/commit/8c65fcacb37e729da5f4c0307065bbe3b89252f2, maybe I had old `master` 🤔

jerch: That was a weirdo before - we an array rule in place, but the older eslint version did not care much about it. After upgrading I got several linting warnings, and had to change the rule in 8c65fca to pass without warnings.

addons/xterm-addon-serialize/test/SerializeAddon.api.ts
```diff
@@ -487,9 +487,9 @@ function newArray<T>(initial: T | ((index: number) => T), count: number): T[] {
   const array: T[] = new Array<T>(count);
   for (let i = 0; i < array.length; i++) {
     if (typeof initial === 'function') {
-      array[i] = (<(index: number) => T>initial)(i);
+      array[i] = (initial as (index: number) => T)(i);
```
jerch, line 490: @Tyriar Do we have a preference here? Could/should this be an eslint rule?

I admit, that I like the `xy as Z` more than `<Z>xy`, its imho easier to grasp with less syntax overhead (makes me less thinking). But thats just me.

addons/xterm-addon-webgl/src/GlyphRenderer.ts
```diff
@@ -263,7 +263,7 @@ export class GlyphRenderer {
         // Get attributes from fg (excluding inverse) and resolve inverse by pullibng rgb colors
         // from bg. This is needed since the inverse fg color should be based on the original bg
         // color, not on the selection color
-        fg = (fg & ~(Attributes.CM_MASK | Attributes.RGB_MASK | FgFlags.INVERSE));
+        fg &= ~(Attributes.CM_MASK | Attributes.RGB_MASK | FgFlags.INVERSE);
```
jerch, line 266: :+1: for shorthand notation where possible.

Minor sidenote - in some older engines those were actually slower than the explicit long versions.

src/browser/Linkifier.test.ts
```diff
@@ -210,7 +210,7 @@ describe('Linkifier', () => {
         let count = 0;
         linkifier.registerLinkMatcher(/test/, () => assert.fail(), {
           validationCallback: (url, cb) => {
-            count += 1;
+            ++count;
```
jerch, line 213: @coderaiser Is there a reason for `++count` instead of `count++`? In C I kinda always use the latter, if I dont bind the result, which leads to same ASM code, but always use the first in loops (which used to make a difference in old compilers, not anymore though). Is there a difference in JS?

src/common/InputHandler.test.ts
```diff
@@ -77,7 +77,7 @@ describe('InputHandler', () => {
       optionsService.options.scrollback = 1;
       bufferService.reset();
     });
-    it('SL (scrollLeft)', async () => {
+    it('SL (scrollLeft)', () => {
```
jerch, line 80: Eww, guess I created a code smell here:

Your tool correctly found the asyncs to be stripped. The code smell happens at those `parseP` calls, those should have `await` in front to be consistent on code level. Created #3347 for that. (You can keep the rule as it is, gonna re-add the async keywords over there.)

jerch (commented):
Just some comments/Questions from my side.

src/browser/Linkifier.test.ts
Tyriar: I always prefer the postfix as well, with the only exception being if I actually need the updated value

addons/xterm-addon-serialize/test/SerializeAddon.api.ts
coderaiser: The thing is `Babel` can’t parse code that have both `react` and `typescript` enabled, for `xterm.js` it’s not a thing, but [new construction is preferable](https://basarat.gitbook.io/typescript/type-system/type-assertion#as-foo-vs-less-than-foo-greater-than).

src/browser/Linkifier.test.ts
coderaiser: There is only one reason: you see what’s going on the variable in the first place, so you can understand it faster:

`aVeryLongVariable++`

Anyways, it’s not a problem, I’ll add an option or change to postfix and disable a rule.

coderaiser: Just updated

addons/xterm-addon-serialize/test/SerializeAddon.api.ts
Tyriar: I prefer `as` more, more readable and a little easier to type. Looks like there is an eslint rule https://github.com/xtermjs/xterm.js/issues/3359

src/browser/Terminal2.test.ts
```diff
@@ -105,7 +105,7 @@ function formatError(input: string, output: string, expected: string): string {
   function addLineNumber(start: number, color: string): (s: string) => string {
     let counter = start || 0;
     return (s: string): string => {
-      counter += 1;
+      ++counter;
```
Tyriar, line 108: postfix?

Tyriar (approved):
Looks good, let's wait until https://github.com/xtermjs/xterm.js/pull/3360 is merged to merge this in since it fixes more `as` problems and also some typing bugs (like setTimeout instead of window.setTimeout)

addons/xterm-addon-serialize/test/SerializeAddon.api.ts
coderaiser: Yes, this is the same rule, with the only difference: [it cannot fix](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/rules/consistent-type-assertions.ts) :). Anyways  it will keep codebase in a clean state warning developers about assertions type used, so I think it is useful.

src/browser/Terminal2.test.ts
coderaiser: done ✅

addons/xterm-addon-serialize/test/SerializeAddon.api.ts
Tyriar: Yeah I noticed that when I did the PR ☹️

PR #3344, opened 2021-05-15
devcontainer: upgrade to node v14.x

silamon:
Upgrade the devcontainer setup from node v10 to node v14. Since some devdependencies require at least node v12, it was no longer possible to run `yarn` successfully with node v10.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10
+FROM node:14
 
 # Configure apt
 ENV DEBIAN_FRONTEND=noninteractive


jerch (commented):
@silamon 
Thx for finding this one. I cannot approve this one on my own, as I dont use the Docker build for anything. Do those manual lib installs in the Dockerfile still work? Not sure if the image changed the distro base, can you confirm, that `yarn test` and `yarn test-api` still works (not even sure how to test that)?

Tyriar (approved):
Works great 👍 

@jerch I think test-api doesn't work in dev containers yet

Issue #3353, opened 2021-05-28
Text shifted down with some characters truncated on Firefox with xtermjs 4.12.0


dstein64:
## Details
- Browser and browser version: Firefox 88.0.1
- OS version: Ubuntu 20.10
- xterm.js version: 4.12.0

### Steps to reproduce

Create a page with the following code:

```html
<!doctype html>
<html>
<head>
    <link rel="stylesheet" href="https://unpkg.com/xterm@4.12.0/css/xterm.css" />
    <script src="https://unpkg.com/xterm@4.12.0/lib/xterm.js"></script>
</head>
<body>
<div id="terminal"></div>
<script>
    const config = {
        rendererType: 'canvas',
        fontFamily: 'monospace',
    };
    const term = new Terminal(config);
    term.open(document.getElementById('terminal'));
    term.writeln('Hello world! {Line 1}.');
    term.write('Hello world! {Line 2} $ ');
</script>
</body>
</html>
```

The output has text shifted down (relative to version 4.11.0) with some characters truncated (e.g., see the "{" and "}" characters).

![4_12](https://user-images.githubusercontent.com/541289/120039080-4c5af980-bfd2-11eb-9919-3175ac14fa07.png)

Here's a screenshot using version 4.11.0, where the text is not shifted down nor truncated.

![4_11](https://user-images.githubusercontent.com/541289/120039100-5da40600-bfd2-11eb-95b2-01c633520af7.png)

This issue occurs on Firefox, but not on Chrome.

dstein64:
Running `git bisect` indicates that the issue started with commit a510ffb3c03c2db160579bea6fded2191ee59f0b.

dstein64:
Truncated text can be seen when using GitHub Codespaces on Firefox, but not Chrome.

This might be related to the issue reported here.

However, for Codespaces the truncated text only occurs at the bottom of the screen, as opposed to occurring generally, which suggests it might not be related to the issue here.

For example, here's a screenshot, where the "{}" is only truncated at the bottom.

![truncated_text_at_bottom](https://user-images.githubusercontent.com/541289/120087754-9dddb400-c0b8-11eb-86dc-d53f3ff07494.png)

nirui:
Not sure how this issue is going, but I think `bottom` is a better solution than `ideographic`.

From: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textBaseline

> "ideographic"
[...]

PR #3356, opened 2021-05-30
Change textBaseline from "ideographic" to "bottom"

dstein64:
This changes the `textBaseline` from `ideographic` to `bottom` to resolve Issue #3353. I made changes in all places where `middle` was changed to `ideographic` as part of PR #3279.

In my testing, the update worked as expected. However, I don't know whether there are assumptions elsewhere that this would break. Any feedback would be appreciated if there is further testing I should conduct and/or additional changes that should be made.

I tried using the code in https://github.com/xtermjs/xterm.js/wiki/Powerline-fonts to test whether powerline fonts continue to work as expected, but the fonts were not supported by my browser, either with or without the change in this PR. I tried on Chrome as well with no success. I also tried on Windows instead of Ubuntu, and encountered the same issue.

fonts_not_supported


dstein64:
I've installed Powerline-supported fonts. While this update helps with Firefox, it appears to break the the proper alignment in Chrome that was added in PR #3279.

Powerline before Firefox:

powerline_before_firefox

Powerline after Firefox:

powerline_after_firefox

Powerline before Chrome:

powerline_before_chrome

Powerline after Chrome:

powerline_after_chrome

PR #3370, opened 2021-06-16
Set text baseline conditionally by browser.

dstein64:
This changes the `textBaseline` from `ideographic` to `bottom` for Firefox, to resolve Issue #3353. I made changes in all places where `middle` was changed to `ideographic` as part of PR #3279. The change was not made for Chrome, to avoid Powerline fonts from becoming [slightly unaligned](https://github.com/xtermjs/xterm.js/pull/3356#issuecomment-850928179). On Firefox, it appears the Poweline text is slightly high when I set the font to `monospace` and slightly low when I set it to `DejaVu Sans Mono`, but in both cases an improvement versus the low text when using `ideographic` for `textBaseline` prior to this PR's update.

**Firefox `fontFamily=monospace` `textBaseline=ideographic`** (before this PR's update)

**Firefox `fontFamily=monospace` `textBaseline=bottom`** (after this PR's update)

**Firefox `fontFamily=DejaVu Sans Mono` `textBaseline=ideographic`** (before this PR's update)

**Firefox `fontFamily=DejaVu Sans Mono` `textBaseline=bottom`** (after this PR's update)

In my testing, the update worked as expected (other than the possible slight Powerline alignment issue mentioned above). However, I don't know whether there are assumptions elsewhere that this would break, and I wasn't sure where in the code would be an appropriate place for defining `TEXT_BASELINE`, which is currently in `browser/renderer/atlas/Constants.ts`.

Any feedback would be appreciated if this is a sufficient way to address Issue #3353 and there is further testing I should conduct and/or additional changes that should be made.

### DIFF
--- a/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
+++ b/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
@@ -4,7 +4,7 @@
  */
 
 import { ICharAtlasConfig } from './Types';
-import { DIM_OPACITY } from 'browser/renderer/atlas/Constants';
+import { DIM_OPACITY, TEXT_BASELINE } from 'browser/renderer/atlas/Constants';
 import { IRasterizedGlyph, IBoundingBox, IRasterizedGlyphSet } from '../Types';
 import { DEFAULT_COLOR, Attributes } from 'common/buffer/Constants';
 import { throwIfFalsy } from '../WebglUtils';
@@ -365,7 +365,7 @@ export class WebglCharAtlas implements IDisposable {
     const fontStyle = italic ? 'italic' : '';
     this._tmpCtx.font =
       `${fontStyle} ${fontWeight} ${this._config.fontSize * this._config.devicePixelRatio}px ${this._config.fontFamily}`;
-    this._tmpCtx.textBaseline = 'ideographic';
+    this._tmpCtx.textBaseline = TEXT_BASELINE;
 
     this._tmpCtx.fillStyle = this._getForegroundCss(bg, bgColorMode, bgColor, fg, fgColorMode, fgColor, inverse, bold);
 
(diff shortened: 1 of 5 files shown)


chkaos:
@dstein64 Great styling, how did you do that?

dstein64:
@chkaos, the prompt styling is from [Powerline](https://github.com/powerline/powerline), used here specifically to test that the update from PR #3279 would still apply as intended (that PR fixed an issue with the display of Powerline fonts).

Tyriar (changes requested):
I think we used to use `top` instead of `bottom` to avoid this cut off:

image

dstein64:
> "I think we used to use top instead of bottom to avoid this cut off:"

Was the cutoff in that image generated using this branch? I'm unable to reproduce that with the update here.

Changing `bottom` to `top` on this branch with no other changes results in no text showing. Are you suggesting 1) switching all browsers to `top` along with necessary corresponding updates (no more use of `ideographic`), 2) switching Firefox to `top` along with necessary updates only for Firefox (`ideographic` would continue to be used for other browsers), or 3) some other modification?

Tyriar:
Sorry about the delay, option 2; ideographic is fine but any uses of bottom need to switch to top.

nirui:
Based on the information on https://www.w3schools.com/Tags/canvas_textbaseline.asp, changing `bottom` to `top` requires some rather big changes to the text render (text alignment calculations etc), since `bottom` is basically the oppose of `top` in terms of vertical text positioning.

Also on the same page, it mentioned:

> **Note**: The textBaseline property works differently in different browsers, especially when using "hanging" or "ideographic".

dstein64:
@Tyriar, can you provide details on how you generated the screenshot in your earlier [comment](https://github.com/xtermjs/xterm.js/pull/3370#pullrequestreview-705566650)? I am unable to reproduce that behavior with the update on this branch.

Issue #3319, opened 2021-05-06
Scrollbar still visible after clearing screen and scrollback buffer with escape sequence


SamVerschueren:
## Details
- Browser and browser version: Google Chrome | 90.0.4430.93 (Official Build) (x86_64)
- OS version: macOS Big Sur
- xterm.js version: 4.9.0

### Steps to reproduce

1. Use a regular terminal and execute `ls -la` a couple of times
2. Run `printf '\u001B[2J\u001B[3J\u001B[H'`
2. The screen is cleared and the scrollbar is removed because the scrollback buffer got cleared

Run the previous steps in xterm.js (easiest is probably VS Code). The screen gets cleared, the scrollback buffer gets cleared as well, but the scrollbar is still visible.

Here's a screen capture of that behaviour

![Kapture 2021-05-06 at 08 01 17](https://user-images.githubusercontent.com/1913805/117295784-8e31cd80-ae74-11eb-9716-24f648809ec1.gif)

More context on this can be found https://github.com/xtermjs/xterm.js/issues/3315#issuecomment-833251913

Puneethnaik:
Greetings @Tyriar . I would like to work on this issue. On preliminary analysis, i am able to trace till when the websocket 'message' event makes a call to terminal.write(). I need your advice on how to process the default prompt string as it can change from one OS to other and even with different bashrcs (on linux). Please guide me through the next steps.

Puneethnaik:
@Tyriar my apologies :slightly_smiling_face: Thanks for directing me in the right direction. I will start studying Viewport.ts.

jerch:
@Puneethnaik `Viewport.ts` might be misleading, I think it is much simpler - for some reason `InputHandler.onScroll` does get respected in the renderer, while `BufferService.onScroll`, prolly some missing link down the event routing. Also see https://github.com/xtermjs/xterm.js/issues/3315#issuecomment-833393234 for some background.

Puneethnaik:
@jerch noted. Thanks for the hints. I will take all this into account and also check out the comment.

Puneethnaik:
[...]

PR #3372, opened 2021-06-20
Scrollbar still visible after clearing screen and scrollback buffer with escape sequence #3319

Puneethnaik:
#3319 issue fix. @Tyriar @jerch please look into it. Would love to hear feedback.


Tyriar (approved):
Works great and makes sense 👍

Puneethnaik:
@Tyriar @jerch Thanks for approving the changes. It is an honor for me to contribute to such a wonderful product as xterm.js :smile:

Issue #2102, opened 2019-05-19
Selection is drawn off the animation frame


Tyriar:
Right now selection is drawn when onSelectionChange happens in a renderer, ideally that would just update state and then RenderCoordinator would ensure the needed rows are updated.

PR #3373, opened 2021-06-20
Redraw selection on terminal resize

### LINKED ISSUE #1963: Selection doesn't always redraw when resizing the window (canvas renderer)

### LINKED ISSUE #2818: DOM renderer selection isn't re-drawn after resizing the terminal

silamon:
Redraw the selection after the terminal got resized for DOM renderer and Canvas renderer. 

Before:
For the DOM renderer: When a selection exists, a `xterm-selection` element is added. Upon resizing, the `xterm-selection` element remains.  
For the Canvas renderer: When a selection exists, the selection is kept in the `SelectionRenderLayer `and redrawn upon resizing. But the Fit Addon resets the canvas completely, so the selection in the `SelectionRenderLayer `is lost before the redrawn happens. 

After: 
For the DOM renderer: When a selection exists, a `xterm-selection` element is added. Upon resizing, the `xterm-selection` element remains. 
For the Canvas renderer:  When a selection exists, the selection is kept in the `SelectionRenderLayer `and redrawn upon resizing. The Fit Addon no longer resets the canvas completely, after testing, I didn't notice any changes, but I'm unsure. 

For both: If a resize alters the selection, a new `selectionChange `event is sent out. Both renderers can pick this up correctly now, allowing them to update their selection correctly. If the resize was attached to the render layers of DOM and Canvas renderer, it would be possible that the `SelectionService `send wrong selections out afterwards. So, attaching this to the `SelectionService `seems the right choice to me. 

Fixes #2818, Fixes #1963
Related #2102, https://github.com/microsoft/vscode/issues/90903


Tyriar (changes requested):
It doesn't look like the selection gets drawn correctly 
recording (40)

silamon:
Thanks a lot for testing. It doesn't keep track of unwrapped lines, I'll fix that soon. 👍

silamon:
This just doesn't work correctly yet if buffer reflow is enabled in all renderers. Somehow it wasn't enabled by default in the demo. I'm going to work on something, but it will tie in closely with whatever is calculated in the buffer reflow functions.

Tyriar:
Closing due to lack of activity to clear out our PR queue, we can always reopen or create a new one though.

Issue #3074, opened 2020-09-05
A way to turn off the scrollbar


NorfairKing:
I'm implementing https://smos.online.
It's essentially a terminal user interface in the browser. (Something like mutt, but in the browser.)
This type of app really needs no scrollbar so I would like to turn it off but I can't seem to figure out how to do that.
After reading through some of the source, I'm thinking there's no option for it.

Tyriar:
If it's just some custom app you can enter the alternate buffer which has no scroll bar `terminal.write('\x1b[?47h')` will do that I believe.

NorfairKing:
@Tyriar I'm using the websockets plugin, I'm doing no 'manual' `terminal.write`-ing. How would I go about this then?

Tyriar:
@NorfairKing you can try setting scrollback to 0, I think this should hide the scroll bar.

NorfairKing:
@Tyriar I tried that but it doesn't, as far as I can tell.

Tyriar:
Since there will never be a scroll bar if you use scrollback: 0, you can adjust styles to hide it by removing:

https://github.com/xtermjs/xterm.js/blob/master/css/xterm.css#L96

NorfairKing:
@Tyriar I managed to remove the scrollbar but the space where the scrollbar was is still empty then (and the terminal doesn't fill it up). There are also white flashes at the border that prevent me from using borders around the terminal. They appear where the scrollbar used to be.

magicnat:
I ran into the same issue when trying to remove the scroll bar; the scroll bar is gone, but the empty space was left behind.

It appears to be caused by the `scrollBarWidth` calculation here: https://github.com/xtermjs/xterm.js/blob/master/src/browser/Viewport.ts#L49

`this._viewportElement.offsetWidth - this._scrollArea.offsetWidth` returns 0 when there's no scrollbar, so the fallback value (15) was used.

I don't see any trivial ways to check if the scrollbar was hidden by the user, or if the user were on mac, so maybe make turning off scrollbar a configurable option?

Puneethnaik:
I would like to work on this issue. Please guide me on the next steps, @Tyriar @jerch

Tyriar:
@Puneethnaik I think the space caused by `scrollBarWidth` should be forced to 0 when the `scrollback` setting is also 0. Code pointer:

[...]

PR #3398, opened 2021-07-28
A way to turn off the scrollbar #3074

Puneethnaik:
issue #3074 fix. @Tyriar Please look into it. Would love to hear feedback.


src/browser/Viewport.ts
```diff
@@ -131,6 +136,9 @@ export class Viewport extends Disposable implements IViewport {
       this._refresh(immediate);
       return;
     }
+    // This is for refreshing the viewport if scrollBarWidth has to be updated
+    this._refresh(immediate);
+    return;
```
Tyriar, line 141: We should avoid calling `_refresh` unless we have to, could we track the last scroll bar width and only update if it's changed?

src/browser/Viewport.ts
```diff
@@ -93,7 +93,12 @@ export class Viewport extends Disposable implements IViewport {
       this._ignoreNextScrollEvent = true;
       this._viewportElement.scrollTop = scrollTop;
     }
-
+    if (this._optionsService.getOption('scrollback') === 0) {
```
Tyriar, line 96: `.options.scrollback` is better to use internally

Tyriar (approved):
Thanks this works pretty well, I'll just do a few clean ups before merging 👍

src/browser/Viewport.ts
Puneethnaik: Calling _refresh is costly. Therefore, we should call only when necessary; I will remember this next time. Thanks for the tip! :+1:

src/browser/Viewport.ts
Puneethnaik: Got it, Thanks for the tip! :+1:

Issue #3376, opened 2021-06-24
Throttle accessibility tree rendering


Tyriar:
`screenReaderMode` can cause layout thrashing when there is a lot of data. This much data doesn't help screen reader users anyway so we could easily throttle it to something like 500-1000ms to save many frames being rendered.

Context: https://github.com/xtermjs/xterm.js/issues/3374#issuecomment-867957990

pattch:
Just wanted to ping this issue and see if there's any update here - I have an open PR that fixes this and curious if that change is an acceptable fix?

PR #3399, opened 2021-07-29
Add a Time-Based Debouncer to update Screen Readers at most once per second

### LINKED ISSUE #3374: [Google Cloud Shell] Screen Reader causing full page crashes when terminal is updated frequently.
The current screen reader implementation waits for render frames in order to debounce refreshes in the accessibility tree. This can cause layout thrashing in pages that have enough elements with enough styles applied, etc. when terminals are updated frequently without newlines being entered. Even worse, when layout thrashing becomes a bottleneck for processing content to be updated in the accessibility tree, the internal character buffer used by xterm's screen reader will grow indefinitely, eventually causing out of memory errors.

This issue is not seen when debouncing in time domain independent of render frames (e.g. debouncing to update the screen reader every one second as opposed to every render frame), as well as when putting a limit on the maximum size of the character buffer. Since the screen reader doesn't need to be updated as frequently as animated output, this seems like a fair tradeoff.

Additionally, this issue is not seen when the screen reader is disabled altogether, using e.g. `screenReaderMode: false`.

Example:
In Google Cloud Shell, running long-running gsutil commands would cause page performance to degrade, due to the fast spinner being used to indicate progress over a long period of time in xterm Terminals. Eventually, this would result in an out of memory error, crashing the entire page.

## Details
- Browser and browser version: Chrome, Firefox verified; assume this issue effects all browsers.
- OS version: Verified across macOS, linux, and Windows.
- xterm.js version: Synced at commit 5cb048094e23da22ccfe46666b87d1ad427dce6d

Screen Shot 2021-06-09 at 11 06 28 AM
*Example profile of xterm terminal causing layout thrashing, not seen when the screen reader is disabled.*

pattch:
This change makes it so that screen readers are announced to based on the most recent update to Screen Readers, as opposed to every possible animation frame. Previously, the very frequent updates to screenreaders in conjunction with the unbounded internal character buffer for announcements caused full page crashes due to out of memory errors when screen reader refreshes weren't fast enough to keep up with fast updates to Terminals. This change ensures that Screen Readers are refreshed at most once per second, but preserves the behavior for visual Terminal refreshes by introducing a second debouncer based on the existing RenderDebouncer.

Closes #3374


pattch:
Note: this pull request DOESN'T trim any content from the Accessibility Manager's internal character buffer, as was previously intended. In practice, throttling the frequency of updates is sufficient in practice to prevent the screen reader updates from causing the crashes as before. As a precaution, I still think the internal buffer should still have a cap on the size it can grow to, but that's not addressed in this PR.

Closes #3376

pattch:
Just wanted to double check and see if there is any update here, and see whether this change is an acceptable fix for xterm.js?

src/browser/TimeBasedDebouncer.ts
```diff
@@ -0,0 +1,78 @@
...
+      // Enough time has lapsed since the last refresh; refresh immediately
+      this._lastRefreshMs = refreshRequestTime;
+      this._innerRefresh();
+    } else if (!this._additionalRefreshRequested) {
+      // This is the first additional request throttled; set up trailing refresh
+      const elapsed = refreshRequestTime - this._lastRefreshMs;
+      const waitPeriodBeforeTrailingRefresh = RENDER_DEBOUNCE_THRESHOLD_MS - elapsed;
+      this._additionalRefreshRequested = true;
+
+      setTimeout(() => {
```
Tyriar, line 52: We'll want to dispose of this timeout if it's set on dispose.

src/browser/TimeBasedDebouncer.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+const RENDER_DEBOUNCE_THRESHOLD_MS = 1000; // 1 Second
+
+import { IDisposable } from 'common/Types';
+
+/**
+ * Debounces calls to update screen readers to update at most once per second.
+ */
+export class TimeBasedDebouncer implements IDisposable {
```
Tyriar, line 13: Can you make an `IRenderDebouncer` interface in browser/Types.d.ts and use that as the prop type and where `RenderDebouncer` is used?

```ts
export interface IRenderDebouncer extends IDisposable {
  refresh(rowStart: number | undefined, rowEnd: number | undefined, rowCount: number): void;
}
```

src/browser/TimeBasedDebouncer.ts
```diff
@@ -0,0 +1,78 @@
+/**
+ * Copyright (c) 2018 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+const RENDER_DEBOUNCE_THRESHOLD_MS = 1000; // 1 Second
```
Tyriar, line 6: Passing this into the ctor will help make it a little more reusable

Tyriar (changes requested):
Sorry about the delay, approach looks good overall 👍

src/browser/TimeBasedDebouncer.ts
pattch: I've implemented clearing the timeout that gets scheduled on dispose if it's set here.

src/browser/TimeBasedDebouncer.ts
pattch: I've made this configurable as requested.

src/browser/TimeBasedDebouncer.ts
pattch: I've created an IRenderDebouncer interface that both RenderDebouncer and TimeBasedDebouncer now implement, and they're typed accordingly where they're used.

pattch:
No worries on the delay - thanks for the constructive feedback! I've done my best to address the issues you brought up, and so this PR should now conform to those expectations. Let me know if there's anything else missing from this fix, thanks

Tyriar (approved):
Perfect, thanks! CI test is a flake

PR #3401, opened 2021-07-30
Handle 'cmd + .' on macOS

### LINKED ISSUE #3400: Need to handle 'cmd + .' on macOS
It would be better if the shortcut is supported as in the macOS built-in Terminal app.

The key combination(cmd + .) is handled as 'ctrl + C'(break) on macOS. And it is more convenient because the ctrl key of Mac keyboards is not so easy to press.

So, the usability of xterm (and other applications like VS Code) will be improved by supporting this, in my opinion.

jihoon-ernesto:
- treat it as 'ctrl + C', as like in macOS Terminal
- checked the expected behavior manually:
  - in xterm demo page
  - in VS Code integrated terminal, using a customized local xterm lib
- fix #3400

### DIFF
--- a/src/common/input/Keyboard.ts
+++ b/src/common/input/Keyboard.ts
@@ -359,6 +359,8 @@ export function evaluateKeyboardEvent(
       } else if (isMac && !ev.altKey && !ev.ctrlKey && !ev.shiftKey && ev.metaKey) {
         if (ev.keyCode === 65) { // cmd + a
           result.type = KeyboardResultType.SELECT_ALL;
+        } else if (ev.keyCode === 190) { // cmd + . (same as ^C)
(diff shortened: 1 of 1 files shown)


Tyriar:
Since only the default macOS terminal does this but others don't seem this is probably something that should be left up to embedders to implement as we try keep xterm.js as small and non-opinionated as possible. Pulling this into VS Code as a custom keybinding for macOS would make sense though. I created https://github.com/microsoft/vscode/issues/130990 if you want to contribute there since it was your idea, otherwise I'll tackle it towards the end of the week 🙂

jihoon-ernesto:
Thank you for your comment. Then I will make a PR there soon.

PR #3404, opened 2021-08-02
Delete unnecessary string escape

nreid260:


addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -23,7 +23,7 @@ export interface ISearchResult {
   row: number;
 }
 
-const NON_WORD_CHARACTERS = ' ~!@#$%^&*()+`-=[]{}|\;:"\',./<>?';
+const NON_WORD_CHARACTERS = ' ~!@#$%^&*()+`-=[]{}|;:"\',./<>?';
```
alur, line 26: Judging by the name, I'm pretty sure this is meant to include \. Maybe add an extra \ instead?

Tyriar:
Yes this was intentional, merged https://github.com/xtermjs/xterm.js/pull/3405 instead. Thanks!

PR #3407, opened 2021-08-10
Add HashiCorp Nomad to the real world uses list

DingoEatingFuzz:
The equivalent of https://github.com/xtermjs/xtermjs.org/pull/149 but in the right place this time 🙂


Tyriar (approved):
Thanks 🙂

PR #3428, opened 2021-08-20
adjust for lineHeight and letterSpacing in webgl custom glyph rendering

### LINKED ISSUE #3427: Letter spacing and line height is off for custom glyphs under webgl
image
image

meganrogge:
fix #3427

### DIFF
--- a/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
+++ b/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
@@ -459,7 +459,7 @@ export class WebglCharAtlas implements IDisposable {
       return NULL_RASTERIZED_GLYPH;
     }
 
-    const rasterizedGlyph = this._findGlyphBoundingBox(imageData, this._workBoundingBox, allowedWidth, isPowerlineGlyph);
+    const rasterizedGlyph = this._findGlyphBoundingBox(imageData, this._workBoundingBox, allowedWidth, isPowerlineGlyph, drawSuccess);
     const clippedImageData = this._clipImageData(imageData, this._workBoundingBox);
 
     // Check if there is enough room in the current row and go to next if needed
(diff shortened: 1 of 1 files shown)


addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
```diff
@@ -567,8 +567,8 @@ export class WebglCharAtlas implements IDisposable {
         y: (boundingBox.bottom - boundingBox.top + 1) / TEXTURE_HEIGHT
       },
       offset: {
-        x: -boundingBox.left + (restrictedGlyph ? 0 : TMP_CANVAS_GLYPH_PADDING),
-        y: -boundingBox.top + (restrictedGlyph ? 0 : TMP_CANVAS_GLYPH_PADDING)
+        x: -boundingBox.left + (restrictedGlyph ? 0 : TMP_CANVAS_GLYPH_PADDING) + (customGlyph ? Math.floor(this._config.letterSpacing / 2) : 0),
+        y: -boundingBox.top + (restrictedGlyph ? 0 : TMP_CANVAS_GLYPH_PADDING) + (customGlyph ? this._config.lineHeight === 1 ? 0 : Math.round((this._config.scaledCellHeight - this._config.scaledCharHeight) / 2) : 0)
```
meganrogge, line 571: pretty ugly to put these inline, but thought creating consts might be costly?

jerch: If perf is a concern - wouldn't it be better to build the box glyphs in the atlas already with those spacing/scaledHeight deps precalculated? Would avoid any special casing here (always hits hard in hot paths), and I think changes to those values are rare, thus it should be no biggie to rebuild dependent glyphs in the atlas.

jerch: Yepp, was meant as a cache warm-up. But I have no clue of the bigger picture here, so I wasnt aware of the color issue :sweat_smile:

Tyriar (approved):
The optimization is probably overkill, can always come back and improve later

PR #3430, opened 2021-08-22
input: handle dead keys

### LINKED ISSUE #2151: Scandinavian keyboard layout can't handle input tilde (~)

Eugeny:
Fixes #2151
This adds extra handling to avoid canceling `keydown` events dead keys, since canceling a dead key's own `keydown` will cancel the future `keydown` for the combined character.

The typical event flow for e.g. `Swedish Tilde` + `Forward Slash` is: 

* While canceling keydown (current):
image
 
* Without canceling keydown (fixed):
image

For clarity, we want to let through the next `keydown` event that is responsible for a future `keypress`/`input` (so not just the "next" one since it might be a modifier key), hence the order of the checks.

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -94,6 +94,13 @@ export class Terminal extends CoreTerminal implements ITerminal {
    */
   private _keyDownHandled: boolean = false;
 
+  /**
+   * Records whether there has been a keydown event for a dead key without a corresponding keydown
+   * event for the composed/alternative character. If we cancel the keydown event for the dead key,
+   * no events will be emitted for the final character.
+   */
+  private _unprocessedDeadKey: boolean = false;
+
   public linkifier: ILinkifier;
   public linkifier2: ILinkifier2;
   public viewport: IViewport | undefined;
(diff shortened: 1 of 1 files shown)


Tyriar (approved):
Change looks fine, I'll assume you're better off testing this one than me 👍

Tyriar:
@Eugeny does this fix https://github.com/microsoft/vscode/issues/82489?

Eugeny:
I suppose it might, but I have only tested that specific tilde combination problem so far.

Eugeny:
Yep, that one's fixed too.

PR #3434, opened 2021-08-26
feat: copy tab characters

### LINKED ISSUE #734: Support rendering of tab characters in terminal
Downstream: https://github.com/Microsoft/vscode/issues/29523

`node -e "console.log('a\tb')"` should output an actual tab. This means it's considered a single character by the selection logic and that it's copied as a `\t` char not several spaces. This probably won't be too difficult to deal with by leveraging and improving upon the double width char code.

This is supported by gnome-terminal, Terminal.app and iTerm2.

daiyam:
This PR is enabling us the possibility to copy the tab characters from the terminal.

Screen Shot

My test script:
```
printf "\033[48;5;1mred\t - \tred\033[0mnormal\n"
```

It should successfully close #734


Tyriar (changes requested):
Sorry but I think we need to discuss this in the issue before a PR happens. One major thing that I think we would need to keep if we went with supporting copying `\t` is everything except for copying should remain is it was, adding `whitespace` to the already complicated `translateToString` is a breaking API change for example.

Additionally it's kind of hard to review this since it contains so many unrelated changes that would be better scoped to their own PR (eg. the launchBrowser refactor is nice, not sure why source-map-support changed).

daiyam:
First, I will try to make new PRs for:
- `launchBrowser`
- `.mocharc.yml` allows the extension `Mocha Test Explorer` to list the tests (`source-map-support` is used to find the the source in the typescript files. Without it, the errors would be linked to the compiled javascript)

Then I will see if I can do the copy without the filler.

Issue #3425, opened 2021-08-19
Move most terminal controls to the side of the demo


Tyriar:
Currently all demo controls are below the terminal, which means we need to scroll to use a lot of them, particular when on a smaller laptop screen:

![image](https://user-images.githubusercontent.com/2193314/130076587-371643ba-c2d8-477a-8e95-585a37d2a9f5.png)

It would be nice if we had a sidebar for most of these, maybe as a first step moving the options there and seeing how it looks? Something like this:

![image](https://user-images.githubusercontent.com/2193314/130076949-e8f78ddc-6c46-4b96-b69e-a494313dbbaa.png)

PR #3439, opened 2021-08-30
Issue #3425 (Terminal controls shifted to the right of the demo)

simran916:
Issue: #3425

All the terminal controls have been shifted to the side of the demo with a toggling tab bar for Options, AddOns, and Style. The screenshots of the modified UI have been attached for reference. Hope this would help to resolve issue #3425
Capture_Options
Capture_Add0ns
Capture_Style

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -15,42 +15,69 @@
   </head>
   <body>
     <h1 style="color: #2D2E2C">xterm.js: A terminal for the <em style="color: #5DA5D5">web</em></h1>
-    <div id="terminal-container"></div>
-    <div>
-      <h3>Options</h3>
-      <p>These options can be set in the <code>Terminal</code> constructor or by using the <code>Terminal.setOption</code> function.</p>
-      <div id="options-container"></div>
-    </div>
-    <div>
-      <h3>Addons</h3>
-      <p>Addons can be loaded and unloaded on a particular terminal to extend its functionality.</p>
-      <div id="addons-container"></div>
-      <h3>Addons Control</h3>
-      <h4>SearchAddon</h4>
-      <p>
-        <label>Find next <input id="find-next"/></label>
-        <label>Find previous <input id="find-previous"/></label>
-        <label>Use regex<input type="checkbox" id="regex"/></label>
-        <label>Case sensitive<input type="checkbox" id="case-sensitive"/></label>
-        <label>Whole word<input type="checkbox" id="whole-word"/></label>
-      </p>
-      <h4>SerializeAddon</h4>
-      <p>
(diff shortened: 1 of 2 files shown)


demo/index.html
```diff
@@ -15,42 +15,69 @@
...
       </div>
     </div>
-    <hr/>
-    <button id="dispose" title="This is used to testing memory leaks">Dispose terminal</button>
-    <button id="custom-glyph" title="Write custom box drawing and block element characters to the terminal">Test custom glyphs</button>
-    <script src="dist/client-bundle.js" defer ></script>
-  </body>
+      <hr/>
+      <button id="dispose" title="This is used to testing memory leaks">Dispose terminal</button>
+      <button id="custom-glyph" title="Write custom box drawing and block element characters to the terminal">Test custom glyphs</button>
```
Tyriar, line 64: Let's put these in a "Test" tab

demo/index.html
```diff
@@ -15,42 +15,69 @@
...
+          <div id="addons-container"></div>
+          <h3>Addons Control</h3>
+          <h4>SearchAddon</h4>
+          <p>
+            <label>Find next <input id="find-next"/></label>
+            <label>Find previous <input id="find-previous"/></label>
+            <label>Use regex<input type="checkbox" id="regex"/></label>
+            <label>Case sensitive<input type="checkbox" id="case-sensitive"/></label>
+            <label>Whole word<input type="checkbox" id="whole-word"/></label>
+          </p>
```
Tyriar, line 45: These would be better presented on new lines now (wrap each in `<p>`?)

demo/index.html
Tyriar: It would be really cool if we could save the last tab and open to it, no need to do in this PR unless easy though.

demo/style.css
```diff
@@ -41,3 +41,52 @@ pre {
     word-wrap: break-word;
     white-space: pre-wrap;
 }
+
+
+#container {
+    display: flex;
+    height: 75vh;
+}
```
Tyriar, line 49: How about setting up the options sidebar such that it fills the full height like this?

image

Tyriar (changes requested):
This is really nice 👍

demo/index.html
simran916: Yes, okay give me some time to give it a try!

simran916: Done!

demo/style.css
simran916: Yes, please check if it is okay!
Capture_Height

Tyriar (approved):
Very nice, thanks for the contribution 👍

PR #3444, opened 2021-08-31
refactor(test): use common function to launch the browser

daiyam:
This PR is adding a common function to launch the browser in the tests.

@Tyriar as discussed in #3434


Tyriar (approved):
Thanks!

PR #3445, opened 2021-08-31
refactor(test): enhance mocha usability

daiyam:
This PR is:
- adding the command `test:dev` to be able to run an unique test or set of tests (`yarn test:dev -g="Color channels"`)
- adding the support of [Mocha Test Explorer](https://marketplace.visualstudio.com/items?itemName=hbenl.vscode-mocha-test-adapter) in Visual Studio Code

The dependency `source-map-support` is used by `mocha` to reference back to the typescript source file (not the compiled javascript file) when there is an error.

@Tyriar as discussed in #3434


Tyriar:
Is there a trick to getting this to work? I checked out this branch, ran yarn, installed the extension and see this:

image

Refreshing loads for a bit and shows the same.

daiyam:
Here I'm getting:
Screen Shot 2021-08-31 at 2 51 53 PM

The updating `settings.json` is required. Without it, I'm don't get anything.
You might also need to click on the reload arrow.

Tyriar:
🤷 

image

daiyam:
Oh! Even `test:dev` doesn't work on Windows...

package.json
```diff
@@ -23,6 +23,7 @@
     "test-api-webkit": "node ./bin/test_api.js --browser=webkit --timeout=20000",
     "test-unit": "node ./bin/test.js",
     "test-unit-coverage": "node ./bin/test.js --coverage",
+    "test:dev": "NODE_PATH='./out' mocha",
```
Tyriar, line 26: Oh this would be the issue on Windows. You'll probably need to make a little wrapper like in test-unit to get it working on Windows

Tyriar (changes requested):

package.json
daiyam: yep. I've fixed that.

daiyam:
@Tyriar Can you retest the Mocha Explorer? You might need to reload VSCode...

Tyriar:
Still doesn't seem to work, even after a full clean and re-running yarn/restarting vscode

daiyam:
Not sure why.. I've tested on Windows with the Insiders version
Screen Shot 2021-08-31 at 4 32 07 PM

Do you have any global settings for `Mocha Test Explorer` or `Test Explorer`?

Tyriar:
Doh, had `"testExplorer.useNativeTesting": true` since it asked that was the issue. Guess it's not ready yet. Seems to be working now

package.json
```diff
@@ -23,6 +23,7 @@
     "test-api-webkit": "node ./bin/test_api.js --browser=webkit --timeout=20000",
     "test-unit": "node ./bin/test.js",
     "test-unit-coverage": "node ./bin/test.js --coverage",
+    "test:dev": "cross-env NODE_PATH='./out' mocha",
```
Tyriar, line 26: Is this `:dev` format required by the mocha test explorer? If there is a way to customize `test-unit-dev` would fit more with the existing style

.vscode/settings.json
```diff
@@ -1,4 +1,5 @@
 {
     "typescript.preferences.importModuleSpecifier": "non-relative",
-    "typescript.preferences.quoteStyle": "single"
+    "typescript.preferences.quoteStyle": "single",
+    "mochaExplorer.envPath": ".env.test"
```
Tyriar, line 4: Can this file name be anything? Maybe `.mocha.env` or similar so it's clear what the purpose of the file in the root is.

Tyriar (changes requested):

package.json
daiyam: It can be changed. It's just that `test:dev` is easier to type...

.vscode/settings.json
daiyam: Yep. Better name.

Tyriar (approved):
Looks good, thanks!

Issue #3369, opened 2021-06-13
windowOptions not typed in get/setOption API


jerch:
`windowOptions` currently gets handled by the `any` type fallthough in `getOption` and `setOption`, which is subpar.

Tyriar:
This actually happens for all options, because the `any` one exists none of the options get compile time checks. We should move to setters and getter imo:

```
terminal.options.windowOptions = ...
```

PR #3448, opened 2021-09-01
Expose typed options from OptionsService through terminal.options

silamon:
This PR makes it possible to use typed options in the terminal addons. Much or less the example from one of the related issues.

```
// before
term.setOption('rendererType', 'dom');

// after
term.options.rendererType = 'dom';
```

As far as I am aware, this doesn't introduce breaking changes (yet). 

There are some following up possibilities: 
* Align the ITerminalOptions in typings again with ITerminalOptions in code. That is, remove the undefined ? from the options, and make the constructor take a Partial<ITerminalOptions> instead of ITerminalOptions>. We can assure that all options are filled in with their default values if the options are passed through to OptionsService. This would be a breaking change whatsoever, but allows to remove a lot of undefined checks. 
* Start replacing getOption and setOption in the codebase to use the getter and setters of ITerminalOptions.
* Consider to deprecate and remove the getOption and setOption methods in favor of getters and setters.   

Related: #2522, #3369

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -152,7 +152,7 @@ export class Terminal extends CoreTerminal implements ITerminal {
    * @alias module:xterm/src/xterm
    */
   constructor(
-    options: ITerminalOptions = {}
+    options: Partial<ITerminalOptions> = {}
   ) {
     super(options);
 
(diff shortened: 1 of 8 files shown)


jerch:
This PR creates a significant perf decrease in common modules, which have to ask option service for options (e.g. `Inputhandler.print` alone decreases by ~20%).
Reason behind - https://github.com/xtermjs/xterm.js/blob/0bc02d6838a2a75048001003617bf172efcf622b/src/common/services/OptionsService.ts#L98-L107

Fix: dont use getter/setter in fast path code.

PR #3452, opened 2021-09-01
Throughput performance improvements

### LINKED ISSUE #3450: Investigate syncScrollArea taking up a lot of time even when the scrollback is full
When the scrollback is filled there should be next to no work happening in syncScrollArea.

Context: https://github.com/xtermjs/xtermjs.org/pull/156

Tyriar:
Fixes #3450

These changes boost the load test results from ~5MB/s -> ~16MB/s on my CPU (i5-4670K @3.4GHz)

### DIFF
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -20,6 +20,7 @@ import { AttributeData } from 'common/buffer/AttributeData';
 import { ICoreService, IBufferService, IOptionsService, ILogService, IDirtyRowService, ICoreMouseService, ICharsetService, IUnicodeService, LogLevelEnum } from 'common/services/Services';
 import { OscHandler } from 'common/parser/OscParser';
 import { DcsHandler } from 'common/parser/DcsParser';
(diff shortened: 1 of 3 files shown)


src/common/InputHandler.ts
```diff
@@ -569,8 +575,7 @@ export class InputHandler extends Disposable implements IInputHandler {
       }
     }
 
-    buffer = this._bufferService.buffer;
```
Tyriar, line 572: @jerch this was the main change that I'm not confident about, does this happen because an active buffer change could have occurred? The `onBufferActivate` event should handle this right?

jerch: Yepp, I kinda prefetch all properties, that are behind getters for speed reasons, but have to refetch it there in case it was changed. But since your new private property already reflects that, it can be removed.

Tyriar:
Looks like I broke something 😄

jerch:
Eww, maybe just a dangling old buffer ref somewhere (not updating at all needed places)?

PR #3453, opened 2021-09-01
Passing cols or rows into ctor as undefined will break terminal

### LINKED ISSUE #3329: Passing cols or rows into ctor as undefined will break terminal
This was causing https://github.com/microsoft/vscode/pull/123502 to fail CI and the reason ended up being that cols/rows set to undefined breaks. `null` works fine, as does not setting any cols/rows property.

silamon:
https://github.com/xtermjs/xterm.js/blob/554f7dab724cf63ad7b0197304eabf37992b9a68/src/common/services/BufferService.ts#L39-L40
Passing `undefined` to cols or rows would result in NaN. Passing `null` works as expected. 

Fixes #3329

### DIFF
--- a/src/common/services/OptionsService.test.ts
+++ b/src/common/services/OptionsService.test.ts
@@ -10,13 +10,23 @@ describe('OptionsService', () => {
   describe('constructor', () => {
     const originalError = console.error;
     beforeEach(() => {
-      console.error = () => {};
+      console.error = () => { };
     });
     afterEach(() => {
       console.error = originalError;
     });
+    it('uses default value if invalid constructor option values passed for cols/rows', () => {
+      const optionsService = new OptionsService({ cols: undefined, rows: undefined });
+      assert.equal(optionsService.getOption('rows'), DEFAULT_OPTIONS.rows);
+      assert.equal(optionsService.getOption('cols'), DEFAULT_OPTIONS.cols);
+    });
+    it('uses values from constructor option values if correctly passed', () => {
+      const optionsService = new OptionsService({ cols: 80, rows: 25 });
+      assert.equal(optionsService.getOption('rows'), 25);
+      assert.equal(optionsService.getOption('cols'), 80);
+    });
     it('uses default value if invalid constructor option value passed', () => {
-      assert.equal(new OptionsService({tabStopWidth: 0}).getOption('tabStopWidth'), DEFAULT_OPTIONS.tabStopWidth);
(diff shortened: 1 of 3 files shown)


src/common/services/OptionsService.test.ts
```diff
@@ -10,13 +10,23 @@ describe('OptionsService', () => {
   describe('constructor', () => {
     const originalError = console.error;
     beforeEach(() => {
-      console.error = () => {};
+      console.error = () => { };
```
meganrogge, line 13: should undo this formatting change

src/common/services/OptionsService.test.ts
```diff
@@ -10,13 +10,23 @@ describe('OptionsService', () => {
...
+      assert.equal(optionsService.getOption('cols'), DEFAULT_OPTIONS.cols);
+    });
+    it('uses values from constructor option values if correctly passed', () => {
+      const optionsService = new OptionsService({ cols: 80, rows: 25 });
+      assert.equal(optionsService.getOption('rows'), 25);
+      assert.equal(optionsService.getOption('cols'), 80);
+    });
     it('uses default value if invalid constructor option value passed', () => {
-      assert.equal(new OptionsService({tabStopWidth: 0}).getOption('tabStopWidth'), DEFAULT_OPTIONS.tabStopWidth);
+      assert.equal(new OptionsService({ tabStopWidth: 0 }).getOption('tabStopWidth'), DEFAULT_OPTIONS.tabStopWidth);
```
meganrogge, line 29:
```suggestion
      assert.equal(new OptionsService({tabStopWidth: 0}).getOption('tabStopWidth'), DEFAULT_OPTIONS.tabStopWidth);
```

src/common/services/OptionsService.ts
```diff
@@ -22,7 +22,7 @@ export const DEFAULT_OPTIONS: ITerminalOptions = Object.freeze({
   cursorStyle: 'block',
   cursorWidth: 1,
   customGlyphs: true,
-  bellSound:  DEFAULT_BELL_SOUND,
```
meganrogge, line 25: undo

src/common/services/OptionsService.ts
```diff
@@ -128,7 +128,7 @@ export class OptionsService implements IOptionsService {
         break;
       case 'cursorWidth':
         value = Math.floor(value);
-        // Fall through for bounds check
```
meganrogge, line 131: undo

meganrogge (changes requested):
looks good, thanks - a few formatting changes should be reverted

Tyriar (approved):
Thanks!

PR #3457, opened 2021-09-02
Support measuring cell size via measureText TextMetrics

Tyriar:
Not ready to look at yet


Tyriar:
Closing as there were problems here.

PR #3460, opened 2021-09-03
Adopt @playwright/test and improve integration tests

### LINKED ISSUE #3326: integration test script seriously broken
#2714 introduces several changes to the integration tests and how they get launched:
- always starts new server
- changed port to 3001

There are several issues with the new script, esp. if you want to use it locally:
- swallows errors (no feedback at all)
- server gets not always teared down
- port is blocked in next run exiting at server startup
- returns with a success returncode on next run without executing ANY test case, WTH?
- fast test case driven development is almost impossible now (server startup takes >3s)

This needs serious rework to work again, at least with proper state handling of the server process.

### LINKED ISSUE #3446: Adopt @playwright/test and use JSHandle
I'm using the new @playwright/test package in Luna Paint and it's essentially just doing what we're doing but in a much nicer way so we should adopt it. Additionally we should move away from just evaluating strings in `page.evaluate` to fetching a JSHandle to the xterm API and using that to test which would give us strong typing on the renderer side.

Tyriar:
Fixes #3446 
Fixes #3326 

Changes summary:

- Use playwright test runner instead of plain playwright with mocha
- Move from deprecated equals to strictEquals
- Use JSHandle for most things instead of raw page.evaluate(string) to strongly type renderer-side test code
  - The main thing to be careful with here is not to use async/await inside the page callback as the compiled `__awaiter` won't be on the page
- Add missing test for onBinary event
- Use `() => {}` for all tests
- Not using the string for `page.evaluate` allows us to avoid the `\\` hack
- Pretty much all write/term calls now use await and can guarantee the action is complete (eg. writes are parsed) by the time it returns in node, so we can avoid using `pollFor` for most tests
- Events are now proxied, making it more convenient to test with

Remaining:

- [ ] Migrate addon tests over
- [ ] Delete old .api tests
- [ ] Get all tests passing in CI


azure-pipelines.yml
```diff
@@ -100,10 +100,8 @@ jobs:
     displayName: 'Install Yarn'
   - script: yarn --frozen-lockfile
     displayName: 'Install dependencies and build'
-  - script: yarn test-api-chromium --headless --forbid-only
-    displayName: 'Integration tests (Chromium)'
-  - script: xvfb-run --auto-servernum -- bash -c "yarn test-api-firefox --headless --forbid-only"
-    displayName: 'Integration tests (Firefox)'
+  - script: yarn test-playwright --forbid-only
```
mxschmitt, line 103: In the config you can add: `forbidOnly: !!process.env.CI` that's how we do it, then you can drop the `--forbid-only` CLI argument everywhere.

package.json
```diff
@@ -61,7 +64,7 @@
     "mustache": "^4.2.0",
     "node-pty": "^0.10.1",
     "nyc": "^15.1.0",
-    "playwright": "^1.11.0",
+    "playwright": "^1.14.1",
```
mxschmitt, line 67: If you have `@playwright/test` installed, then you don't need additionally`playwright`. But keep in mind that you need in your CI `npx playwright install` to install the browsers.

test/playwright/TestUtils.ts
```diff
@@ -0,0 +1,346 @@
...
+export interface ITestContext {
+  page: Page;
+  termHandle: JSHandle<Terminal>;
+  proxy: TerminalProxy;
+}
+
+export async function createTestContext(browser: Browser): Promise<ITestContext> {
+  const page = await browser.newPage();
+  await page.goto('http://127.0.0.1:3000/test');
+  const proxy = new TerminalProxy(page);
```
mxschmitt, line 21: The webserver in the test-runner does automatically set the base URL. So you can change it to just `/test`.

package.json
Tyriar: npx essentially just installs it too though right? What's the benefit of using npx over a regular dev dependency?

mxschmitt: @playwright/test registers playwright as a command, I agree, kinda confusing. See here: https://unpkg.com/browse/@playwright/test@1.14.1/package.json

Pro this change:
- less dependencies, you need to update only a single one
Contra:
- you need to invoke npx playwright install manually

azure-pipelines.yml
```diff
@@ -100,10 +100,10 @@ jobs:
     displayName: 'Install Yarn'
   - script: yarn --frozen-lockfile
     displayName: 'Install dependencies and build'
-  - script: yarn test-api-chromium --headless --forbid-only
-    displayName: 'Integration tests (Chromium)'
-  - script: xvfb-run --auto-servernum -- bash -c "yarn test-api-firefox --headless --forbid-only"
-    displayName: 'Integration tests (Firefox)'
+  - script: npx playwright install
+    displayName: 'Install PlayWright browsers'
```
mxschmitt, line 104:
```suggestion
    displayName: 'Install Playwright browsers'
```

It's Playwright 🙂

Tyriar: @mxschmitt now that you're here, any idea how to solve this? Installing didn't see to fix:

```
  6) [Firefox Stable] › ../../test/playwright/CharWidth.test.ts:20:5 › CharWidth Integration Tests › getStringCellWidth › ASCII chars 

    browserType.launch: Executable doesn't exist at /home/vsts/.cache/ms-playwright/firefox-1335/firefox/firefox
    ╔═════════════════════════════════════════════════════════════════════════╗
    ║ Looks like Playwright Test or Playwright was just installed or updated. ║
    ║ Please run the following command to download new browsers:              ║
    ║                                                                         ║
    ║     npx playwright install                                              ║
    ║                                                                         ║
    ║ <3 Playwright Team                                                      ║
    ╚═════════════════════════════════════════════════════════════════════════╝
```

mxschmitt: Most likely because you have `playwright` and `@playwright/test` inside your package.json with different versions. I suggest to remove `playwright` and only have `@playwright/test` with the latest version.

(everything from `playwright` is also exported from `@playwright/test` node/package-wise)

mxschmitt: Another idea is since you are using yarn, to use `yarn run playwright install` instead.

Tyriar:
This is too stale, it's available for reference when we want to try move to @playwright/test in the future.

PR #3461, opened 2021-09-03
fix: disable pointer-events on a11y overlay

### LINKED ISSUE #1931: Copy&paste with screenReaderMode
When screenReaderMode is used it is not possible to use right-click and select copy or paste from context menu.

## Details
- Browser and browser version: Firefox 64
- OS version: Fedora 29
- xterm.js version: 3.10.1

### Steps to reproduce

1.  `var term = new Term({ screenReaderMode: true });`

2. Then right-click on the terminal and there is no option for pasting.

jonchurch:
closes #1931 

Looking for feedback here, not sure this is the best approach to fix this issue. It's something I'm willing to ship in another product, as turning on `screenReaderMode` has removed the ability to copy and paste.

From what I can tell, there shouldn't be any detrimental effect of removing pointer events in the `.xterm-accessibility` div, as the div and children aren't meant to be interacted with via pointers. 

Happy to work on this more to fit a11y needs, but I'm not aware of what problems a change like this could be creating.

### DIFF
--- a/css/xterm.css
+++ b/css/xterm.css
@@ -142,6 +142,10 @@
     cursor: crosshair;
 }
 
+.xterm .xterm-accessibility {
+    pointer-events: none;
+}
+
 .xterm .xterm-accessibility,
 .xterm .xterm-message {
     position: absolute;


Tyriar (commented):
I think this would break the screen reader's ability to read out what is being hovered?

jonchurch:
Do you have any insight into how yall are testing a11y issues here? I have been trying to test this with NVDA but am trying to learn to operate it at the same time, so my tests were inconclusive. 

I'd love to work on this issue, and want to figure out how we can both agree on an acceptance criteria 👍

PR #3462, opened 2021-09-05
fix: switch active unicode version in demo

### LINKED ISSUE #3442: Cannot switch active unicode version in demo
Right now when the unicode11 addon is activated the version remains at 6. Instead of adding UI it might be better to set `term.unicode.activeVersion = '11'` automatically

anirudh1713:
closes #3442 

Initially checks if `unicode11` is checked and sets the correct unicode `activeVersion` and does the same for change event.


Tyriar (approved):
Works great, thanks!

PR #3463, opened 2021-09-07
Fix typo in the word "Performance"

staticalliam7:


Tyriar (commented):
This wasn't a typo, I'm aware it may not be a "real word" but its use in software development is so widespread that everyone knows what it means and I happen to like it.

Issue #3138, opened 2020-10-21
Improve .npmignore of xterm-addon-search


aeschli:
You can see from
https://unpkg.com/browse/xterm-addon-search@0.7.0/

that the `xterm-addon-search` npm packages contains unnecessary resources:
- `fixtures`
- `out`
- `src` ([lib/xterm-addon-search.js.map](https://unpkg.com/browse/xterm-addon-search@0.7.0/lib/xterm-addon-search.js.map) already contains sources)

aeschli:
Same for https://unpkg.com/browse/xterm-addon-unicode11@0.2.0/ and https://unpkg.com/browse/xterm-addon-webgl@0.9.0/

jerch:
Partly also covered by the issue #3137. Guess we should clarify, what we want to end up in packages for whatever reason and exclude everything else.

kumaran-14:
I would like to take on this and related #3137
What would be the solution to this?

Tyriar:
@kumaran-14 the solution for these is to make the .npmignore for the various modules (addons+code for https://github.com/xtermjs/xterm.js/issues/3137) exclude files that we should not be shipping inside the modules. You might want to read up about how .npmignore works before jumping in - it's similar to .gitignore.

jerch:
Had another look into this - it seems our addon packages are stuck between _source distribution - nodejs target - browser prebundled_, we kinda have abit of everything but nothing to proper full extend :smile_cat: 

Not sure where to go from here, a few thoughts are:
- A source package distribution never gonna work to full extend with addon packages due their entanglement with the base repo. Furthermore full source can also be obtain from the repo itself, so `src` is not that useful.
- If we want to go with one npm package only, we prolly should move browser builds to `dist` and let `main` point to the addon entry file (addon class). This would be my preferred variant, it is just moving current content abit around.
- If we want browser builds to be distributed via npm, we should do a second package with `main` pointing to an ES6 module bundle of the addon. Those will not work in nodejs out-of-the-box, but the browser will be more than happy.

mofux:
> If we want browser builds to be distributed via npm, we should do a second package with main pointing to an ES6 module bundle of the addon. Those will not work in nodejs out-of-the-box, but the browser will be more than happy.

AFAIK we could use the`module` field in `package.json` to point to the entry file of the ES6 Module bundle. The `main` should still point to the CJS version.

jerch:
[...]

PR #3476, opened 2021-09-19
Update npmignore files

### LINKED ISSUE #3137: xterm.js npm module contains `src`
The npm node module ships with the `src` folder but also contains a source map file that includes the sources.

In order to keep the `xterm.js` npm module small and installation time minimal, I suggest to add the `src` folder to the `.npmignore` list.

The following resources can also be ignored:
`tsconfig.all.json`
`.eslintrc.json`

silamon:
This PR cleans up the npm ignore files across the repository. 
Source files have been excluded since the source mapping is available in the library folder.
Headless files have been removed from xterm since they are included in xterm-headless. 

Fixes #3137, #3138

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,10 +2,6 @@
 *
 !*/
 
-# Whitelist - entries to be included must be negated with "!"
-!*.js
-!*.json
-
 # Whitelist - css/
 !css/**/*.css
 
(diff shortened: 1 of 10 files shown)


Eugeny:
IMO sources should remain packaged, since xterm typings are artificial and purposefully limited. In particular, I'm importing types from src/ to have strong typings for xterm buffer internals

PR #3480, opened 2021-09-22
Support strikethrough in serialize addon

silamon:


Tyriar (approved):
Thanks for fixing up the other API comments and tests too 👍

PR #3491, opened 2021-10-01
Patch document role for terminal output accessibility: Move tabindex from xterm to accessibility tree root

jooyoungseo:
CC @Tyriar @meganrogge @isidorn 

This is a follow-up PR to patch [xtermjs/xterm.js#3483](https://github.com/xtermjs/xterm.js/pull/3483) to correctly address [microsoft/vscode#98918](https://github.com/microsoft/vscode/issues/98918).

I tested the recent PR (#3483) using JAWS and NVDA in the dev xterm.js with the screenreader mode, and found that it doesn't trigger document role correctly. This is because the focusable tabindex has not been placed in the same level as the document role. Moving down the tabindex 0 from xterm to xterm accessibility tree root resolves this issue (confirmed with JAWS and NVDA testing).

### DIFF
--- a/src/browser/AccessibilityManager.ts
+++ b/src/browser/AccessibilityManager.ts
@@ -53,6 +53,7 @@ export class AccessibilityManager extends Disposable {
   ) {
     super();
     this._accessibilityTreeRoot = document.createElement('div');
+    this._accessibilityTreeRoot.setAttribute('tabindex', '0');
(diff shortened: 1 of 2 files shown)


isidorn:
Thanks a lot for the PR 👏  This fix sounds good to me. Though I did not test it out.
@Tyriar and @meganrogge would be the right ones to review and merge.

PR #3494, opened 2021-10-06
sound addon

### LINKED ISSUE #3316: bell handling
(Coming from #3314.)

Consolidate bell handling as follows:
- Strip sound output handling from xterm.js.
- Propagate `onBell` as interface to hook into the bell (make sound output an outer env problem). Here integrators can do whatever they want or suits their purpose.
- Move current direct sound output to a default bell addon, that keeps working in browsers as now, but listens on `onBell`. If demanded, we can later extend the addon by other bell metaphors, like a visual bell and such.

This cannot be done until next major release version (v5), as it involves changes to the terminal options.

jerch:
WIP for a sound addon.

Currently working:
- DECPS prototype with blocking sound output (on every note currently)
- get audio permissions through user action (needed for safari)

TODO:
- move SoundService functionality to addon for `onBell` event
- make user action for audio permission configurable
- polish tone generator for DECPS
- get buffer/blocking semantics from VT520 right
- check for correct blocking timings, esp. for tiny offsets from promise code
- clarify meaning of note=0, same for defaults in all param positions

The early DECPS functionality can be tested with the example script:
```bash
$> addons/xterm-addon-sound/fixture/example.sh
```
Note: If the output stops at "C5" without any sound, click somewhere on the browser background (that click is currently hardlinked as user action to get the audio permissions).

Shall fix #3316, once we get to v5.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,6 +22,8 @@
       "addons/xterm-addon-serialize/src/tsconfig.json",
       "addons/xterm-addon-serialize/test/tsconfig.json",
       "addons/xterm-addon-serialize/benchmark/tsconfig.json",
+      "addons/xterm-addon-sound/src/tsconfig.json",
+      "addons/xterm-addon-sound/test/tsconfig.json",
       "addons/xterm-addon-unicode11/src/tsconfig.json",
       "addons/xterm-addon-unicode11/test/tsconfig.json",
       "addons/xterm-addon-web-links/src/tsconfig.json",
--- a/addons/xterm-addon-sound/LICENSE
+++ b/addons/xterm-addon-sound/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2021, The xterm.js authors (https://github.com/xtermjs/xterm.js)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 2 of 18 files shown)


j4james:
I think I found one genuine bug though. My understanding is that `printf "\e[4;32;0,~"` should produce a second of silence, but you appear to be interpreting note 0 the same as 1, i.e a C5.

jerch:
@j4james Thx for checking.

> ... my implementation seems to be interpreting the timing somewhat faster, so it ended up finishing about a second ahead of yours.

Oh I have not balanced the timings yet, there is a high chance, that the promise handling eats some microseconds here adding to a major offset in the end.

> I think I found one genuine bug though. My understanding is that printf "\e[4;32;0,~" should produce a second of silence, but you appear to be interpreting note 0 the same as 1, i.e a C5.

I wasnt sure about the meaning of zero there, so I went with "zero means one". I know thats a rather bold assumption, but not totally far fetched (as many DEC sequences have that ZDM defaulting to 1 thingy). Guess this needs to be checked against the real device as well.

Added both to the TODO/reminder list.

jerch:
@j4james Fixed the "sound of silence".

About the timings - it seems my timings have like +-10ms startup, while the runtime seems to shift by ~5ms average. I think I cannot get this more accurate for 2 reasons:
- under load the promises will schedule later, I have no way to declare RT stuff with Javascript, and promises are just a cooperative multitasking hack in JS.
- The time resolution is ~20ms on the audio context, so even in the audio contexts internal timer it will shift by 10ms average.

jerch:
Unlikely to happen, thus closing.

PR #3506, opened 2021-10-12
Report focus state on DECSET 1004

### LINKED ISSUE #2333: Report focus state on activation of DECSET 1004
See https://github.com/xtermjs/xterm.js/pull/2316#issuecomment-511900837.

Tyriar:
Fixes #2333


jerch:
@Tyriar Flakey test cases? I see this happening once in while, but only on macos and windows ingetration tests. Any clue how come, and why not on linux integration tests?

Tyriar:
One of the mouse related ones was an actual timeout and I think I split it up, maybe this is another one that needs splitting?

PR #3524, opened 2021-10-22
OSC 4/10/11/12 color handling

### LINKED ISSUE #3487: Implement OSC 10/11, prolly other OSC color commands as well
OSC 10/11 is needed for `lsix` to properly derive current SGR 39/49 colors of the terminal.

We have some prior implementation regarding the OSC color commands with OSC 4, but I think that is not feature complete compared to xterm.

### LINKED ISSUE #3549: implement OSC 104 | 110 | 111
Similar to OSC 4 | 10 | 11 these commands are meant to restore default color values.

Implementing this is a bit more involved, as we currently don't store default color values. An idea to provide default color values would be to keep anything bootstrapped from a `ColorManager.setTheme` around.

jerch:
Shall implement OSC10/11 color handling, and extend OSC 4 by the query notation.

Fixes #3487. Also closes #3549.

Beside OSC 4/10/11 there are more color commands, but I am not quite sure yet, how they are supposed to work (esp. OSC 5 with its "special colors" is a mystery to me). Not sure if the PR will deal with those as well in the end, they have no priority on my side. Most of OSC 12 -19 is def. out of scope for xterm.js.

TODO:
- [x] implement xcolor spec parser
- [x] ~support X11 color names~ (removed, to be done when #3530 got fixed)
- [x] implement OSC 10 & 11 with set & query
- [x] docs for OSC 10 & 11
- [x] unit tests
- [x] test against `lsix`
- [x] integration tests with full round trip (query & set)
- [x] extend OSC 4:
  - [x] ~apply color change immediately~ postponed: #3548
  - [x] use full xcolor spec parser
  - [x] add query functionality
  - [x] rework tests
  - [x] integration tests
- [x] goodie: implement corresponding reset sequences in 100+ (104, 110, 111)
- [x] + OSC 12 | 112


jerch:
@Tyriar Done with this PR. I postponed several aspects and created follow-up issues, see at the PR todo list at the top.

jerch:
@Tyriar Now done for real. :smile_cat:

src/browser/Color.ts
```diff
@@ -4,6 +4,9 @@
  */
 
 import { IColor } from 'browser/Types';
+import { IColorRGB } from 'common/Types';
+
+// FIXME: Move Color.ts lib to common?
```
Tyriar, line 9: Color.ts lives in browser historically as it's only required when a renderer is active, things have changed a little with this PR where InputHandler knows more about color. Sounds like a good change if `XParseColor` ends up using these utils, but if not it would be best to keep in browser as then it's not included unnecessarily in `xterm-headless`.

Tyriar (approved):

PR #3535, opened 2021-10-31
input: prevent duplicate IME input on Linux - fixes #3533

### LINKED ISSUE #3533: Support Sogou Chinese IME
Originally reported at https://github.com/microsoft/vscode/issues/135940

Characters get doubled when using Sogou.

image

Code pointer: https://github.com/xtermjs/xterm.js/blob/master/src/browser/input/CompositionHelper.ts

Eugeny:
This prevents both `input` and `keyup` triggering the same data event with Sogou IME on Linux.

The PR should be pretty safe as it only further filters the recently added `input` event handler.


Tyriar (approved):
Sorry about the delay 👍

PR #3536, opened 2021-11-01
Clear isWrapped in eraseBufferLine

### LINKED ISSUE #1882: isWrapped does not get lifted for longish bash commands

silamon:
... only if the full bufferline is being cleared. In other cases it seems logical to keep whatever isWrapped was.

Fixes #1882


Tyriar (approved):
Thanks @silamon works great, sorry about the delay

PR #3546, opened 2021-11-06
Allow setting multiple options through term.options

### LINKED ISSUE #2522: Move away from string-based options

### LINKED ISSUE #3369: windowOptions not typed in get/setOption API

silamon:
This PR builds further on #3448 and completes the second part of the linked issues. 

```
// before
term.setOption('rendererType', 'dom');
term.setOption('fontSize', 12);

// after
term.options = {
  rendererType: 'dom',
  fontSize: 12
};
```

Fixes #2522, Fixes #3369


src/browser/public/Terminal.ts
```diff
@@ -92,6 +92,9 @@ export class Terminal implements ITerminalApi {
   public get options(): ITerminalOptions {
     return this._core.options;
   }
+  public set options(options: ITerminalOptions) {
+    this._core.options = options;
+  }
```
Tyriar, line 97: Since this is just API sugar, should we iterate over the properties here and keep CoreTerminal.ts the same?

typings/xterm.d.ts
```diff
@@ -638,7 +638,7 @@ declare module 'xterm' {
     /**
      * Get the terminal options
```
Tyriar, line 639:
```suggestion
     * Gets or sets the terminal options. This supports setting multiple options.
     *
     * @example Get a single option
     * ```typescript
     * console.log(terminal.options.fontSize);
     * ```
     *
     * @example Set a single option
     * ```typescript
     * terminal.options.fontSize = 12;
     * ```
     *
     * @example Set multiple options
     * ```typescript
     * terminal.options = {
     *   fontSize: 12,
     *   fontFamily: 'Arial',
     * ;
     * ```
     */
```

Tyriar (changes requested):
Nice change 👍

Tyriar:
Also looks like this broke the tests, it would be good to add some multiple options set tests too.

Tyriar (changes requested):

src/browser/public/Terminal.test.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+import { Terminal } from 'browser/public/Terminal';
+import { assert } from 'chai';
+import { ITerminalOptions } from 'common/Types';
+
+const INIT_COLS = 80;
+const INIT_ROWS = 24;
+
+describe('Public Terminal', () => {
```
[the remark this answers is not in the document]
Tyriar: I'm working on migrating this over now

Tyriar (approved):

PR #3554, opened 2021-11-12
wide characters overflow the cache canvas on WebGL render

acid-chicken:
This change resolves microsoft/vscode#137047 in most cases.

When WebglCharAtlas draws a wide glyph (e.g. CJK chars) to their cacheCanvas, the glyph overflows the canvas if remaining of the current row width is bigger than the configured char width but less than the actual glyph width.

Base branch
--

Head branch
--

### DIFF
--- a/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
+++ b/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
@@ -14,7 +14,6 @@ import { AttributeData } from 'common/buffer/AttributeData';
 import { channels, rgba } from 'browser/Color';
 import { tryDrawCustomChar } from 'browser/renderer/CustomGlyphs';
 
-// FIXME: rendering many characters can overflow the texture rarely
 // For debugging purposes, it can be useful to set this to a really tiny value,
 // to verify that LRU eviction works.
 const TEXTURE_WIDTH = 1024;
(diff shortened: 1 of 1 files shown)


addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
```diff
@@ -14,8 +14,9 @@ import { AttributeData } from 'common/buffer/AttributeData';
 import { channels, rgba } from 'browser/Color';
 import { tryDrawCustomChar } from 'browser/renderer/CustomGlyphs';
 
-// In practice we're probably never going to exhaust a texture this large. For debugging purposes,
-// however, it can be useful to set this to a really tiny value, to verify that LRU eviction works.
+// FIXME: rendering many characters can overflow the texture rarely
```
Tyriar, line 17: Do you ever see this when not brute force testing like this? It would be better to move this to an issue rather than a comment.

Tyriar (approved):
Fix looks good 👍 just a minor comment.

addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
acid-chicken: Thank you for the review quickly.  I didn't seen ever this behavior in my usage.  Perhaps the very rarely usage like showing long CJK sentences or database entries on very large display with HiDPI causes this, however their usage may also have other performance ploblems, so it doesn't need to consider here.  I just remove this comment for now.

meganrogge:
Works well, thank you!

PR #3559, opened 2021-11-20
Resolve perf regression from OptionsService.options 

### LINKED ISSUE #3556: Resolve perf regression from OptionsService.options
https://github.com/xtermjs/xterm.js/pull/3448#issuecomment-969404249

> This PR creates a significant perf decrease in common modules, which have to ask option service for options (e.g. `Inputhandler.print` alone decreases by ~20%).
> Reason behind: https://github.com/xtermjs/xterm.js/blob/0bc02d6838a2a75048001003617bf172efcf622b/src/common/services/OptionsService.ts#L98-L107
> 
> Fix: dont use getter/setter in fast path code.

We could expose the raw object internally?

silamon:
Fixes #3556

### DIFF
--- a/src/common/services/OptionsService.test.ts
+++ b/src/common/services/OptionsService.test.ts
@@ -28,6 +28,10 @@ describe('OptionsService', () => {
     it('uses default value if invalid constructor option value passed', () => {
       assert.equal(new OptionsService({ tabStopWidth: 0 }).getOption('tabStopWidth'), DEFAULT_OPTIONS.tabStopWidth);
     });
+    it('object.keys return the correct number of options', () => {
+      const optionsService = new OptionsService({ cols: 80, rows: 25 });
+      assert.notEqual(Object.keys(optionsService.options).length, 0);
+    });
   });
   describe('setOption', () => {
     let service: OptionsService;
--- a/src/common/services/OptionsService.test.ts
+++ b/src/common/services/OptionsService.test.ts
@@ -15,11 +15,6 @@ describe('OptionsService', () => {
     afterEach(() => {
       console.error = originalError;
     });
-    it('publicOptions and options are modifying the same object', () => {
-      const optionsService = new OptionsService({ cols: undefined, rows: undefined });
-      optionsService.options.rows = 25;
-      assert.equal(optionsService.publicOptions.rows, 25);
-    });
     it('uses default value if invalid constructor option values passed for cols/rows', () => {
       const optionsService = new OptionsService({ cols: undefined, rows: undefined });
       assert.equal(optionsService.getOption('rows'), DEFAULT_OPTIONS.rows);
--- a/src/common/services/OptionsService.test.ts
+++ b/src/common/services/OptionsService.test.ts
@@ -10,10 +10,15 @@ describe('OptionsService', () => {
   describe('constructor', () => {
     const originalError = console.error;
     beforeEach(() => {
-      console.error = () => {};
+      console.error = () => { };
     });
     afterEach(() => {
       console.error = originalError;
+    });
+    it('publicOptions and options are modifying the same object', () => {
+      const optionsService = new OptionsService({ cols: undefined, rows: undefined });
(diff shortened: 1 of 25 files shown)


silamon:
I might come with a better solution, since a proxy isn't supposed to work in Internet Explorer.

src/common/services/OptionsService.ts
```diff
@@ -60,55 +60,51 @@ const FONT_WEIGHT_OPTIONS: Extract<FontWeight, string>[] = ['normal', 'bold', '1
 export class OptionsService implements IOptionsService {
   public serviceBrand: any;
 
-  private _options: ITerminalOptions;
   public options: ITerminalOptions;
```
Tyriar, line 63: I think the proxy object might be a little overkill for what we're after here, what if we had this interface:

```
interface IOptionsService {
  readonly rawOptions: Readonly<ITerminalOptions>;
}
```

This interface could be mapped to the actual underlying object and essentially be a fast readonly path for accessing options internally, since many reads is the perf concern, not writes. The implementation would look something like this:

```
export class OptionsService implements IOptionsService {
  // Note that it's not Readonly<ITerminalOptions> in the impl as the interface protects it
  readonly rawOptions: ITerminalOptions;
  // The defineProperty object (or Proxy?) that includes validation
  options: ITerminalOptions;
}
```

Lack of IE support doesn't bother me so much as we call out partial support in the readme. Someone that uses a terminal that is using IE at this point probably deserves a broken experience tbh 😉

Tyriar (changes requested):

src/common/services/OptionsService.test.ts
```diff
@@ -26,7 +26,11 @@ describe('OptionsService', () => {
       assert.equal(optionsService.getOption('cols'), 80);
     });
     it('uses default value if invalid constructor option value passed', () => {
-      assert.equal(new OptionsService({tabStopWidth: 0}).getOption('tabStopWidth'), DEFAULT_OPTIONS.tabStopWidth);
+      assert.equal(new OptionsService({ tabStopWidth: 0 }).getOption('tabStopWidth'), DEFAULT_OPTIONS.tabStopWidth);
+    });
+    it('object.keys return the correct number of options', () => {
+      const optionsService = new OptionsService({ cols: 80, rows: 25 });
+      assert.notEqual(Object.keys(optionsService.options).length, 0);
```
Tyriar, line 33: This assertion is failing

src/common/services/Services.ts
```diff
@@ -188,6 +188,7 @@ export const IOptionsService = createDecorator<IOptionsService>('OptionsService'
 export interface IOptionsService {
   serviceBrand: undefined;
 
+  readonly rawOptions: Readonly<ITerminalOptions>;
```
Tyriar, line 191:
```suggestion
  /**
   * Read only access to the raw options object, this is an internal-only fast path for accessing
   * single options without any validation as we trust TypeScript to enforce correct usage
   * internally.
   */
  readonly rawOptions: Readonly<ITerminalOptions>;
```

src/common/services/OptionsService.ts
```diff
@@ -60,55 +60,60 @@ const FONT_WEIGHT_OPTIONS: Extract<FontWeight, string>[] = ['normal', 'bold', '1
...
         } catch (e) {
           console.error(e);
         }
       }
     }
 
     // set up getters and setters for each option
-    this.options = this._setupOptions(this._options);
+    this.rawOptions = defaultOptions;
+    this.options = { ... defaultOptions };
```
Tyriar, line 85: I think this looks good if it does what it looks like, we would just now want to replace all accesses internally with `rawOptions` instead of the slower `options`. This is mainly important for "hot" code paths like within the parser/inputhandler, but we should just use it everywhere.

Tyriar (changes requested):

Tyriar (approved):
Great work @silamon 👍

PR #3560, opened 2021-11-20
Remove headless include in benchmark projects

### LINKED ISSUE #3424: Compile error: Cannot find module 'xterm-headless'
This happens after watch is run I think a couple of times:

image

Probably to do with us leaning on import `'xterm'` being the repo root but `'xterm-headless'` lives in `root/headless` and is generated.

silamon:
Fixes #3424


Tyriar (approved):
Thanks for hunting this down 👍

PR #3562, opened 2021-11-22
ES6 support

jogibear9988:
at the moment you need to manually run 
npm run package-module

I could do it for the rest of the addons (if you will merge it)

how should I include in an autmatic publish process? (so the npm package contains both ES6 and old one)


Tyriar:
Thanks for the effort, I'll mark this as reference so we have it for the future, but I'd prefer to not merge in ES6 support until I understand it better and can adopt it in my own usages, otherwise it'll probably end up breaking and go undetected.

PR #3565, opened 2021-11-24
trycatch characterJoiner handlers

### LINKED ISSUE #3547: A crashing character joiner brings down the whole terminal
In particular, `opentype.js` cannot parse the updated Monaco font file in macOS Monterey (`No valid cmap sub-tables found.`), rendering xterm unusable when the ligature addon is active.

LabhanshAgrawal:
Fixes #3547

### DIFF
--- a/src/browser/services/CharacterJoinerService.ts
+++ b/src/browser/services/CharacterJoinerService.ts
@@ -176,26 +176,25 @@ export class CharacterJoinerService implements ICharacterJoinerService {
     // At this point we already know that there is at least one joiner so
     // we can just pull its value and assign it directly rather than
     // merging it into an empty array, which incurs unnecessary writes.
-    let joinedRanges: [number, number][] = [];
+    let allJoinedRanges: [number, number][] = [];
     try {
-      joinedRanges = this._characterJoiners[0].handler(text);
+      allJoinedRanges = this._characterJoiners[0].handler(text);
     } catch (error) {
       console.error(error);
     }
     for (let i = 1; i < this._characterJoiners.length; i++) {
       // We merge any overlapping ranges across the different joiners
-      let joinerRanges: [number, number][] = [];
       try {
-        joinerRanges = this._characterJoiners[i].handler(text);
+        const joinerRanges = this._characterJoiners[i].handler(text);
+        for (let j = 0; j < joinerRanges.length; j++) {
+          CharacterJoinerService._mergeRanges(allJoinedRanges, joinerRanges[j]);
+        }
       } catch (error) {
         console.error(error);
       }
-      for (let j = 0; j < joinerRanges.length; j++) {
-        CharacterJoinerService._mergeRanges(joinedRanges, joinerRanges[j]);
-      }
     }
-    this._stringRangesToCellRanges(joinedRanges, lineData, startCol);
-    return joinedRanges;
+    this._stringRangesToCellRanges(allJoinedRanges, lineData, startCol);
+    return allJoinedRanges;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/browser/services/CharacterJoinerService.ts
```diff
@@ -176,10 +176,20 @@ export class CharacterJoinerService implements ICharacterJoinerService {
     // At this point we already know that there is at least one joiner so
     // we can just pull its value and assign it directly rather than
     // merging it into an empty array, which incurs unnecessary writes.
-    const joinedRanges: [number, number][] = this._characterJoiners[0].handler(text);
+    let joinedRanges: [number, number][] = [];
```
Tyriar, line 179: This should go back to `const` since it's assigned only once. I was confused reviewing this due to the similarly of `joinerRanges` and `joinedRanges`, how about we rename `joinedRanges` -> `allJoinedRanges` to clarify?

src/browser/services/CharacterJoinerService.ts
```diff
@@ -176,10 +176,20 @@ export class CharacterJoinerService implements ICharacterJoinerService {
...
+    try {
+      joinedRanges = this._characterJoiners[0].handler(text);
+    } catch (error) {
+      console.error(error);
+    }
     for (let i = 1; i < this._characterJoiners.length; i++) {
       // We merge any overlapping ranges across the different joiners
-      const joinerRanges = this._characterJoiners[i].handler(text);
+      let joinerRanges: [number, number][] = [];
```
Tyriar, line 187: This should go back to `const` since it's assigned only once.

src/browser/services/CharacterJoinerService.ts
```diff
@@ -176,10 +176,20 @@ export class CharacterJoinerService implements ICharacterJoinerService {
...
+      try {
+        joinerRanges = this._characterJoiners[i].handler(text);
+      } catch (error) {
+        console.error(error);
```
Tyriar, line 191: If this one fails we'll need to skip the for loop as well, let's move that into the `try` block

Tyriar (changes requested):

src/browser/services/CharacterJoinerService.ts
LabhanshAgrawal: Since we have to try catch the first handler call as well, we'll need to instantiate this with empty array first.
If we want to keep it as `const` then, should I push the results from first handler? In that case, it'd be better to just have the loop start from index 0

src/browser/services/CharacterJoinerService.ts
LabhanshAgrawal: sure

src/browser/services/CharacterJoinerService.ts
LabhanshAgrawal: sure

src/browser/services/CharacterJoinerService.ts
Tyriar: Makes sense as is 👌, we should do the rename though

Tyriar (approved):
Thanks @LabhanshAgrawal 👍

LabhanshAgrawal:
Can we get a patch release, please 🙏
4.16 is nearly 2 weeks later.

Tyriar:
@LabhanshAgrawal any reason you don't just pick up the beta? I only really have time to release during our "debt week" which is in 2 weeks

LabhanshAgrawal:
Oh, didn't think about that. Will test with that and see. Thanks.

PR #3574, opened 2021-12-09
Clear unprocessed dead key state on input or press

### LINKED ISSUE #3573: Key stroke ignored after quote on US international keyboards
VS Code issue: https://github.com/microsoft/vscode/issues/133597

Repro:

- Type a ' or " (followed by a space, to make it appear)
- Hit an arrow key, Enter, Backspace or Delete and observe that nothing happens 🐛 
- Hit an arrow key, Enter, Backspace or Delete again, this should work

Caused by: https://github.com/xtermjs/xterm.js/pull/3430

The arrow key returns here:

https://github.com/xtermjs/xterm.js/blob/d54c40a3edb97fc57ec144fc65eb5b87760eefad/src/browser/Terminal.ts#L1075

Tyriar:
Fixes #3573

@Eugeny this works for the problem in #3573, I'm not too familiar with AltGr though, could you verify that I didn't break anything you fixed in #3430?

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -1169,6 +1169,10 @@ export class Terminal extends CoreTerminal implements ITerminal {
 
     this._keyPressHandled = true;
 
+    // The key was handled so clear the dead key state, otherwise certain keystrokes like arrow
+    // keys could be ignored
+    this._unprocessedDeadKey = false;
+
     return true;
   }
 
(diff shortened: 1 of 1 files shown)


Eugeny:
Swedish layout from the original issue still works fine, I'll give it another test on a Windows machine tonight

Eugeny (approved):
French on Windows is good too.

PR #3576, opened 2021-12-15
Added support for custom regex to WebLinkProvider

### LINKED ISSUE #3540: Make Web Links Regex Configurable
Currently, the Web Links Addon has a hard-coded Regex used to match URLs, and this regex isn't always sufficient / comprehensive (related: #3539). It would be helpful to introduce some method to provide an alternative Regex, something along the lines of a `setUrlRegex(regex: RegExp)` that would de-register the existing link provider etc. and add a new one with the provided regex.

Heysunk:
Fixes #3540


Tyriar (approved):
I made a couple of small changes, but looks good 👍

PR #3596, opened 2022-01-05
Extend TLD matching to 14 chars

wesbos:
I ran into this issue with hyper wasn't highlighting my `.localhost` domains, which is a legitimate TLD and used by caddy server to allow SSL certs locally.

Then I went down the rabbit hole of TLDs and found currently the longest TLDs are 14 chars which include .BANANAREPUBLIC, .COOKINGCHANNEL, .KERRYLOGISTICS, .WEATHERCHANNEL

I'm not regex expert, but I *think* this should work? 😀


wesbos:
Sorry just found `.TRAVELERSINSURANCE` which is 18 chars :)

Bowero:
Well I am a self-declared Regex expert and this should definitely work.

Tyriar (approved):
Looks good 🙂

PR #3616, opened 2022-01-26
Implemented smooth scrolling.

PoetaKodu:
Hello everyone.

I implemented smooth scrolling. I was annoyed that every time I scroll terminal in VS Code I lose track (see #1140 and https://github.com/microsoft/vscode/issues/125950).
The option is disabled by default but can be enabled using terminal options.

Preview:
https://youtu.be/BCYdZwydmLw

### DIFF
--- a/demo/client.ts
+++ b/demo/client.ts
@@ -353,6 +353,8 @@ function initOptions(term: TerminalType): void {
       } else if (o === 'scrollSensitivity') {
         term.options.scrollSensitivity = parseFloat(input.value);
         updateTerminalSize();
+      } else if (o === 'smoothScrollingSpeed') {
+        term.options.smoothScrollingSpeed = parseFloat(input.value);
       } else if(o === 'scrollback') {
         term.options.scrollback = parseInt(input.value);
         setTimeout(() => updateTerminalSize(), 5);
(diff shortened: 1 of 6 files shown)


Tyriar (commented):
@PoetaKodu very cool 👍 will try get a review done this week/early next week. The main thing I want to check out is how the settings work, trackpad vs wheel and how close the experience is to monaco.

Eugeny:
Tested and trackpad scrolling isn't affected by the new option :+1:

Tyriar (commented):
Looking into this now, my first impressions are:

- The scroll speed is too slow, looks like monaco has a smooth scroll duration of 125ms and it feels a lot more responsive because of that
- `smoothScrollingStepInterval` seems like we don't need it and it should instead be replaced by `requestAnimationFrame`
- A duration would be preferable to `smoothScrollingSpeed`, so the pointer device tells the terminal how much to scroll and smooth scroll animates that over x milliseconds. If the animation happens based on a number of lines each frame then scrolling a large amount of lines seems like it would be a bad experience?

Tyriar (changes requested):
Yeah the above is my main feedback, it would be better to move to duration-based scrolling and we should use `requestAnimationFrame` instead of `setTimeout`/`setInterval`. Interested in your thoughts, thanks!

Tyriar:
Implemented the needed changes in https://github.com/xtermjs/xterm.js/pull/3940

Issue #3617, opened 2022-01-29
Window contents not cleared properly


LabhanshAgrawal:
## Details
- Browser and browser version: Chrome 97
- OS version: macOS 11.6.2
- xterm.js version: 4.16

### Steps to reproduce

Run `ls -G && nvim` with webgl renderer

### With WebGL
<img width="700" alt="Screenshot 2022-01-29 at 11 09 25" src="https://user-images.githubusercontent.com/16598275/151649042-4083d275-28ec-4794-ba5b-973bfbb1cdfc.png">

### With Canvas
<img width="700" alt="Screenshot 2022-01-29 at 11 09 45" src="https://user-images.githubusercontent.com/16598275/151649040-046493d7-c9af-4dec-8c3d-7c5916c8d18c.png">

## Other Examples

![CleanShot 2022-01-16 at 10 12 01](https://user-images.githubusercontent.com/16598275/151650663-4f87d3db-bcf6-49c1-bc18-907d735a363a.gif)

https://github.com/vercel/hyper/issues/6242
https://github.com/vercel/hyper/issues/6239

I'm able to consistently reproduce using the steps above for webgl renderer, the hyper issue I've linked also mentions the same for canvas but I'm not able to get a consistent repro

### Finding
54375713bc2b049648f03540e611f9c3abd58561 is the first bad commit

@Tyriar I'm not familiar enough with xterm core code to fix this, I can try if you can provide some direction as to what might be wrong from the commit I've mentioned.

Till we can get this, we can use 4.15 since this affects usability a lot. But that version suffers from #3547, I've opened #3618 to work around that in ligatures addon itself

jerch:
Seems `model.clear()` is too much here: https://github.com/xtermjs/xterm.js/blob/6c18b7f2cc91b923e765b03e572f9038e294d368/addons/xterm-addon-webgl/src/WebglRenderer.ts#L119-L134

Wild guess (because idk the cache/redraw-on-change logic behind) - the model clear in `setColors` leaves old cell drawings on screen, if it got called along with huge buffer changes, which is the case for neovim due to an altbuffer switch.

Indeed, can repro this "forgotten-onscreen-cells" issue with this line:
```bash
echo -e '\x1b[?1049h\x1b]11;rgb:00/00/ff\x07\x1b[31m\x1b[1;15HH\x1b[BE\x1b[BL\x1b[BL\x1b[BO'
```

With `this._model.clear();` commented out above the issue is gone.

[...]

PR #3618, opened 2022-01-29
trycatch font parsing in ligatures addon

LabhanshAgrawal:
Adding the fix for #3547 to ligatures addon directly so that it can be used with older xterm.js versions
Ref #3617

### DIFF
--- a/addons/xterm-addon-ligatures/src/font.ts
+++ b/addons/xterm-addon-ligatures/src/font.ts
@@ -90,9 +90,19 @@ export default async function load(fontFamily: string, cacheSize: number): Promi
     if (fonts.hasOwnProperty(family) && fonts[family].length > 0) {
       const font = fonts[family][0];
       if ('blob' in font) {
-        return loadBuffer(await (await font.blob()).arrayBuffer(), { cacheSize });
+        try {
+          return loadBuffer(await (await font.blob()).arrayBuffer(), { cacheSize });
+        } catch (err) {
+          console.error(err);
+          return undefined;
+        }
       }
-      return await loadFile(font.path, { cacheSize });
+      try {
+        return await loadFile(font.path, { cacheSize });
+      } catch (err) {
+        console.error(err);
+        return undefined;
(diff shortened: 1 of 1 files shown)


LabhanshAgrawal:
Tests failing on

https://github.com/xtermjs/xterm.js/blob/6c18b7f2cc91b923e765b03e572f9038e294d368/addons/xterm-addon-ligatures/src/index.test.ts#L142

and

https://github.com/xtermjs/xterm.js/blob/6c18b7f2cc91b923e765b03e572f9038e294d368/addons/xterm-addon-ligatures/src/index.test.ts#L182

As these expect errors being thrown which this pr suppresses

Tyriar (commented):
> to ligatures addon directly so that it can be used with older xterm.js versions

@LabhanshAgrawal isn't the right fix here to use the correct ligatures addon for the xterm.js version and update xterm.js if it's not fixed before? Throwing here and using the existing catch does seem like the more correct thing to do (instead of having a catch in xterm that would never be hit).

Tyriar:
Closing out since https://github.com/xtermjs/xterm.js/issues/3617 can now be fixed by updating to latest webgl beta

LabhanshAgrawal:
> @LabhanshAgrawal isn't the right fix here to use the correct ligatures addon for the xterm.js version and update xterm.js if it's not fixed before?

The issue was due to opentype not being able to parse some font on macOS 12, so none of the earlier ligatures addon versions would be able to work in that scenario even with previous xterm versions.

> Throwing here and using the existing catch does seem like the more correct thing to do (instead of having a catch in xterm that would never be hit).

Didn't really get what you're referring to here.

> Closing out since #3617 can now be fixed by updating to latest webgl beta

👍

Tyriar:
> Didn't really get what you're referring to here.

Would just prefer not to have try/catches everywhere when only one of them should be used in practice. We prefer to take a more evergreen approach as supporting older versions takes a bunch of effort.

PR #3619, opened 2022-01-29
HTML serialization

### LINKED ISSUE #1883: Copying HTML-formatted output
Please consider adding an API to copy terminal contents as a HTML string.

silamon:
Fixes #1883 
Makes use of the existing serialize addon and the templates


meganrogge:
Thanks for working on this.

Could you please:
- add information about how embedders are expected to hook this up to copy from selection 
- revert the formatting changes made to `client.ts`
- maybe I'm doing something wrong, but the demo doesn't seem to work for me 
  - the button says it copies, but it doesn't. 
  - when I copied the output manually and pasted it in:
      -  the vscode editor, it didn't produce the expected result or any red text
      Screen Shot 2022-02-01 at 12 13 01 PM
      - apple notes, it didn't produce the expected result - all of the text was colored red
Screen Shot 2022-02-01 at 12 12 28 PM

silamon:
Selection has been added, but unfortunately broke some existing tests. I'll fix that when I have time. :)
I've used the new asynchronous clipboard api, but a lot of browsers don't support it. I've replaced that with a deprecated alternative, which seems to be more supported anyway.

meganrogge:
from what I've read, it looks like an embedder would have to intercept the `copy` event, set the data format as `'text/html'` and provide the serialized data. is that how you are testing this @silamon ?

silamon:
When you click the 'serialize' button, it should add a html code onto your clipboard of the selected part of the terminal. Don't copy the html in the visible window, that's there for debugging or to save as html manually if copy to clipboard doesn't work. From as soon as the button is clicked, you should be able to paste the selected part in an editor that supports pasting html like word or something like that.

meganrogge:
Doesn't work for me in edge, but I see it working for some cases that I've tried in Chrome 👍🏼.

Eugeny:
Copying itself does work reliably for me in Chrome, however when the alternate screen is active, it still copies random content from the primary screen.

meganrogge:
It actually does work in edge for me as well. I was expecting the background color to match that of the terminal, but perhaps that's not necessary.

demo/client.ts
```diff
@@ -447,6 +448,23 @@ function serializeButtonHandler(): void {
   }
 }
 
+function htmlSerializeButtonHandler(): void {
+  const output = addons.serialize.instance.htmlserialize();
```
Tyriar, line 452: We'll need the new API changes reflected in https://github.com/xtermjs/xterm.js/blob/master/addons/xterm-addon-serialize/typings/xterm-addon-serialize.d.ts

demo/client.ts
```diff
@@ -447,6 +448,23 @@ function serializeButtonHandler(): void {
...
+  function listener(e) {
+    e.clipboardData.setData("text/html", output);
+    e.clipboardData.setData("text/plain", output);
+    e.preventDefault();
+  }
+  document.addEventListener("copy", listener);
+  document.execCommand("copy");
+  document.removeEventListener("copy", listener);
+  document.getElementById("htmlserialize-output-result").innerText
+            = "Copied to clipboard";
```
Tyriar, line 465:
```suggestion
  document.getElementById("htmlserialize-output-result").innerText = "Copied to clipboard";
```

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -469,3 +512,154 @@ interface ISerializeOptions {
...
+
+    if (fgChanged || bgChanged || flagsChanged) {
+      const fgHexColor = this._getHexColor(cell, true);
+      if (fgHexColor) {
+        content.push('color: ' + fgHexColor + ';');
+      }
+
+      const bgHexColor = this._getHexColor(cell, false);
+      if (bgHexColor) {
+        content.push('background-color: ' + bgHexColor + ';');
```
Tyriar, line 617: Can background color be disabled? I think in VS Code copying with syntax highlighting originally had this but it got removed as it was annoying.

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -469,3 +512,154 @@ interface ISerializeOptions {
...
+  }
+
+  private _getHexColor(cell: IBufferCell, isFg: boolean): string | undefined {
+    const color = isFg ? cell.getFgColor() : cell.getBgColor();
+    if (isFg ? cell.isFgRGB() : cell.isBgRGB()) {
+      const rgb = [
+        color >>> 0xFF0000 & 255,
+        color >>> 0xFF00 & 255,
+        color & 255
+      ];
```
Tyriar, line 593: I don't think this does what you expect, do you mean this instead?

```suggestion
      const rgb = [
        (color >> 16) & 255,
        (color >>  8) & 255,
        (color      ) & 255
      ];
```

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -469,3 +512,154 @@ interface ISerializeOptions {
...
+    targetLength = targetLength - target.length;
+    if (targetLength > padString.length) {
+      padString += padString.repeat(targetLength / padString.length);
+    }
+    return padString.slice(0, targetLength) + target;
+  }
+
+  protected _beforeSerialize(rows: number, start: number, end: number): void {
+    this._htmlContent += '<html><head><meta name=\'generator\' content=\'xtermjs\'/>'
+      + '<meta http-equiv=\'Content-Type\' content=\'text/html; charset=UTF-8\'/></head><body><!--StartFragment--><pre>';
```
Tyriar, line 563: Is the extra metadata/comments here needed? In particular: `meta name=generator`, `StartFragment`, `EndFragment`

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -469,3 +512,154 @@ interface ISerializeOptions {
...
+  ) {
+    super(buffer);
+
+    // https://github.com/xtermjs/xterm.js/issues/3601
+    this._colors = (_terminal as any)._core._colorManager.colors;
+  }
+
+  private _padStart(target: string, targetLength: number, padString: string): string {
+    targetLength = targetLength >> 0;
+    padString = String(typeof padString !== 'undefined' ? padString : ' ');
```
Tyriar, line 549: You can use `padString !== undefined` here, I stopped caring about people who want to override `undefined` years ago 😉. Also I think the `String()` is redundant and this could be simplified:

```suggestion
    padString = padString ?? ' ';
```

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -469,3 +512,154 @@ interface ISerializeOptions {
...
+  private _htmlContent = '';
+
+  private _inverseStyle = 'color: #000000; background-color: #BFBFBF;';
+  private _blinkStyle = 'text-decoration: blink;';
+  private _boldStyle = 'font-weight: bold;';
+  private _italicStyle = 'font-style: italic;';
+  private _underlineStyle = 'text-decoration: underline;';
+  private _strikethroughStyle = 'text-decoration: line-through;';
+  private _invisibleStyle = 'visibility: hidden;';
+  private _dimStyle = 'opacity: 0.5;';
```
Tyriar, line 533: If these are fixed we can put them in a `const enum`, this will make tsc inline all the values so the members won't end up existing:

```ts
const enum CssStyles {
	Inverse = 'color: #000000; background-color: #BFBFBF;',
	Blink = 'text-decoration: blink;',
	Bold = 'font-weight: bold;',
	Italic = 'font-style: italic;',
	Underline = 'text-decoration: underline;',
	Strikethrough = 'text-decoration: line-through;',
	Invisible = 'visibility: hidden;',
	Dim = 'opacity: 0.5;'
}
```

addons/xterm-addon-serialize/src/SerializeAddon.ts
```diff
@@ -460,6 +495,14 @@ export class SerializeAddon implements ITerminalAddon {
     return content;
   }
 
+  public htmlserialize(options?: Partial<IHtmlSerializeOptions>): string {
```
Tyriar, line 498: `serializeAsHtml` seems like a better more descriptive option

demo/client.ts
```diff
@@ -447,6 +448,23 @@ function serializeButtonHandler(): void {
   }
 }
 
+function htmlSerializeButtonHandler(): void {
+  const output = addons.serialize.instance.htmlserialize();
+  document.getElementById('htmlserialize-output').innerText = output;
+
+  // Deprecated, but the most supported for now.
```
Tyriar, line 455: Since we're a devtool I think it's fine to use the newer APIs like the clipboard API, writing to the clipboard API should work on all modern browser, just not IE. That will future proof this code and give a better example of how to use the new feature.

demo/client.ts
```diff
@@ -447,6 +448,23 @@ function serializeButtonHandler(): void {
...
+  function listener(e) {
+    e.clipboardData.setData("text/html", output);
+    e.clipboardData.setData("text/plain", output);
```
Tyriar, line 458: I think `text/plain` should actually be set to the original string, that means the application in question can decide which mime type it prefers.

Tyriar (changes requested):
Thanks for the PR @silamon, looks pretty close 😍

Eugeny:
BG/FG colors were correct in the HTML I got, although when I first tested it by pasting into Word, it turned out inverted - that's Word's problem though, the HTML itself is correct and opening it in Chrome gets the colors right.

meganrogge:
Yes @Eugeny I saw the same - it was inverted in apple notes, but not in chrome.

addons/xterm-addon-serialize/src/SerializeAddon.ts
silamon: StartFragment and EndFragment seems needed, meta tags have been removed.

demo/client.ts
silamon: I've left the code as it is, since it seems more supported. 
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write vs
https://developer.mozilla.org/en-US/docs/Web/API/Element/copy_event

silamon:
I can't seem to reproduce the text being pasted inverted. Do you see colored text in that paste or is it just clear text in black?

Eugeny:
@silamon it's all colored with the exception of black and white being inverted. I don't even think it's your fault - my custom HTML copy implementation suffers from the same if background and foreground are #000 and #fff respectively.

demo/client.ts
```diff
@@ -448,6 +449,21 @@ function serializeButtonHandler(): void {
...
+  document.getElementById('htmlserialize-output').innerText = output;
+
+  // Deprecated, but the most supported for now.
+  function listener(e: any) {
+    e.clipboardData.setData("text/html", output);
+    e.preventDefault();
+  }
+  document.addEventListener("copy", listener);
+  document.execCommand("copy");
+  document.removeEventListener("copy", listener);
```
Tyriar, line 463: > I've used the new asynchronous clipboard api, but a lot of browsers don't support it. I've replaced that with a deprecated alternative, which seems to be more supported anyway.

It would be better if we moved on to more modern APIs like `navigator.clipboard`, it looks like just IE doesn't support writing to the clipboard which doesn't matter https://caniuse.com/async-clipboard

demo/index.html
```diff
@@ -49,6 +49,10 @@ <h4>SerializeAddon</h4>
             <button id="serialize">Serialize the content of terminal</button>
             <label><input type="checkbox" id="write-to-terminal">Write back to terminal</label>
             <div><pre id="serialize-output"></pre></div>
+
+            <button id="htmlserialize">Serialize the content of terminal in HTML</button>
```
Tyriar, line 53: I'm seeing Edge not copy the whole buffer when no selection is made. Should this button be changed to say serialize the selection instead?

addons/xterm-addon-serialize/typings/xterm-addon-serialize.d.ts
```diff
@@ -24,14 +24,23 @@ declare module 'xterm-addon-serialize' {
...
-     * 
+     *
      * @param options Custom options to allow control over what gets serialized.
      */
     public serialize(options?: ISerializeOptions): string;
 
+    /**
+     * Serializes terminal rows into a HTML string.  The output of this function can be written
+     * to the OS clipboard. If an application supports pasting HTML, the content of the terminal
+     * is pasted with style options retained.
```
Tyriar, line 38:
```suggestion
     * Serializes terminal content as HTML, which can be written to the clipboard using the
     * `text/html` mimetype. For applications that support it, the pasted text should then retain
     * its colors/styles.
```

addons/xterm-addon-serialize/typings/xterm-addon-serialize.d.ts
```diff
@@ -56,4 +65,24 @@ declare module 'xterm-addon-serialize' {
...
     excludeAltBuffer?: boolean;
   }
+
+  export interface IHTMLSerializeOptions {
+    /**
+     * The number of rows in the scrollback buffer to serialize, starting from the bottom of the
+     * scrollback buffer. When not specified, all available rows in the scrollback buffer will be
+     * serialized.
+     */
+    scrollback: number;
```
Tyriar, line 75: This should mention `{@link IHTMLSerializeOptions.onlySelection}`

addons/xterm-addon-serialize/typings/xterm-addon-serialize.d.ts
```diff
@@ -56,4 +65,24 @@ declare module 'xterm-addon-serialize' {
...
+
+    /**
+     * Whether to only serialize the selection. If false, the whole active buffer is serialized in HTML.
+     * True by default.
+     */
+    onlySelection: boolean;
```
Tyriar, line 81: I don't think this should mention the false case as it overrides that. Also having a true by default _optional_ setting is a bit weird, let's make this false by default.

addons/xterm-addon-serialize/typings/xterm-addon-serialize.d.ts
```diff
@@ -56,4 +65,24 @@ declare module 'xterm-addon-serialize' {
...
+
+    /**
+     * Whether to include the global background of the terminal. True by default.
+     */
+    includeGlobalBackground: boolean;
```
Tyriar, line 86: Let's also change this to false by default.

Tyriar:
Looks good other than those comments 👍

demo/client.ts
silamon: I've tested with navigator.clipboard in the first commit and Firefox doesn't support it. The caniuse link does mention that as well.

addons/xterm-addon-serialize/typings/xterm-addon-serialize.d.ts
```diff
@@ -24,14 +24,23 @@ declare module 'xterm-addon-serialize' {
...
     public serialize(options?: ISerializeOptions): string;
 
+    /**
+     * Serializes terminal rows into a HTML string.  The output of this function can be written
+     * to the OS clipboard. If an application supports pasting HTML, the content of the terminal
+     * is pasted with style options retained.
+     *
+     * @param options Custom options to allow control over what gets serialized.
+     */
+    public serializeAsHTML(options?: Partial<IHTMLSerializeOptions>): string;
```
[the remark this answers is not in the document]
silamon: I'll write some unit tests as soon as I have time. :)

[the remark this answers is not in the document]
silamon: Tests have been added. Unfortunately, not as playwright tests since I couldn't get it to work at this moment.

PR #3621, opened 2022-01-30
fix text baseline for Legacy Edge

felixse:
Legacy Edge (or in my case the UWP WebView) has similar as issues as Firefox had with 'ideographic'


Tyriar (approved):
Thanks @felixse 👍

PR #3628, opened 2022-02-04
dispose of markers on buffer clear

meganrogge:
While working on #1852, I noticed that when `clear` is called, the decorations aren't disposed of because their markers aren't disposed.

### DIFF
--- a/src/common/buffer/Buffer.ts
+++ b/src/common/buffer/Buffer.ts
@@ -16,6 +16,7 @@ import { DEFAULT_CHARSET } from 'common/data/Charsets';
 import { ExtendedAttrs } from 'common/buffer/AttributeData';
 
 export const MAX_BUFFER_SIZE = 4294967295; // 2^32 - 1
+const enum BufferState { CLEARING = 'clearing' }
 
 /**
  * This class represents a terminal buffer (an internal state of the terminal), where the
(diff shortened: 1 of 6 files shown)


src/common/buffer/Buffer.ts
```diff
@@ -43,6 +44,7 @@ export class Buffer implements IBuffer {
   private _whitespaceCell: ICellData = CellData.fromCharData([0, WHITESPACE_CELL_CHAR, WHITESPACE_CELL_WIDTH, WHITESPACE_CELL_CODE]);
   private _cols: number;
   private _rows: number;
+  private _state: string | undefined;
```
Tyriar, line 47: Was this to avoid many splices and instead remove them all in bulk? Would be a little nicer as `_isClearing: boolean`

Tyriar (changes requested):
Looks like there's a test related change needed to implement an interface

src/common/buffer/Buffer.ts
meganrogge: Yes it was

Tyriar (approved):
Looks good, an API test that creates 2 markers, scrolls them into scrollback, then creates 2 more and makes sure they get disposed on clear would be great too

PR #3630, opened 2022-02-08
add Decorations

### LINKED ISSUE #1852: Allow embedders to decorate ranges in the buffer

meganrogge:
fixes #1852


meganrogge:
originally, we were thinking that we'd need `IGutterDecorationOptions` and `IBufferDecorationOptions`, but now think that both can be described using the same set of options - `IDecorationOptions`

test/api/Terminal.api.ts
```diff
@@ -731,6 +731,28 @@ describe('API Integration Tests', function(): void {
...
+      assert.notEqual(await page.evaluate(`document.querySelector('.xterm-screen .xterm-decoration')`), undefined);
+    });
+    it('should return undefined when the marker has already been disposed of', async () => {
+      await openTerminal(page);
+      await page.evaluate(`window.marker = window.term.addMarker(1)`);
+      await page.evaluate(`window.marker.dispose()`);
+      assert.equal(await page.evaluate(`window.decoration = window.term.registerDecoration({ marker: window.marker });`), undefined);
+      assert.equal(await page.evaluate(`document.querySelector('.xterm-screen .xterm-decoration')`), undefined);
+    });
+    it.skip('should throw when a negative x offset is provided', async () => {
```
meganrogge, line 748: this should be passing, but doesn't - saying the function doesn't throw. 

when I provide a -x value in the demo, it throws. maybe the error isn't propagated for some reason?

src/browser/services/DecorationService.ts
```diff
@@ -0,0 +1,134 @@
...
+class Decoration extends Disposable implements IDecoration {
+  private static _nextId = 1;
+  private _marker: IMarker;
+  private _element: HTMLElement | undefined;
+  private _id: number = Decoration._nextId++;
+  public isDisposed: boolean = false;
+
+  public get id(): number { return this._id; }
+  public get element(): HTMLElement { return this._element!; }
+  public get marker(): IMarker { return this._marker; }
```
Tyriar, line 70: Something I've started doing recently to avoid duplication is to just make properties like these public as we only expose `IDecoration` outside this file so setting them in protected already. That way we don't need to create the property and get the (small) perf hit associated with them.

So this instead and removing the prop:

```
readonly id: number = Decoration._nextId++;
```

src/browser/services/DecorationService.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (this.isDisposed) {
+      return;
+    }
+    this._screenElement.removeChild(this.element);
+    this.isDisposed = true;
+    this._marker.dispose();
+    // Emit before super.dispose such that dispose listeners get a change to react
+    this._onDispose.fire();
+    super.dispose();
+  }
```
Tyriar, line 100: I've been preferring to move the dispose handling for objects that extend `Disposable` into the ctor in the form of:

```
this.register({ dispose: () => {
  ...
}});
```

That way we won't accidentally forget to call `super.dispose()` which has happened before.

typings/xterm.d.ts
```diff
@@ -380,29 +380,90 @@ declare module 'xterm' {
...
+
+    /**
+     * The HTMLElement that gets created after the 
+     * first _onRender call, or undefined if accessed before
+     * that.
+     */
+    readonly element: HTMLElement | undefined;
+  }
+
+  export interface IDecorationOptions {
```
Tyriar, line 437: Let's add jsdoc

typings/xterm.d.ts
```diff
@@ -870,6 +931,15 @@ declare module 'xterm' {
      */
     addMarker(cursorYOffset: number): IMarker | undefined;
 
+    /**
+     * (EXPERIMENTAL) Adds a decoration to the terminal using
+     *  @param decorationOptions, which takes a marker and an optional anchor, 
+     *  width, height, and x offset from the anchor. Returns the decoration or
+     *  undefined if the alt buffer is active or the marker has already been disposed of.
+     *  @throws if the @param decorationOptions includes a negative x offset.
```
Tyriar, line 939:
```suggestion
     *  @throws when options include a negative x offset.
```

Tyriar (changes requested):

test/api/Terminal.api.ts
Tyriar: Until we move to `@playwright/test` we need to do something like this to check throws: https://github.com/Tyriar/xterm.js/blob/65182176b9de63a4d369dfc4cb33b040d41983a7/test/api/Terminal.api.ts#L36-L43

```ts
    await page.evaluate(`
      try {
        window.term.buffer;
      } catch (e) {
        window.throwMessage = e.message;
      }
    `);
    await pollFor(page, 'window.throwMessage', 'You must set the allowProposedApi option to true to use proposed API');
```

src/browser/Terminal.ts
```diff
@@ -513,6 +515,9 @@ export class Terminal extends CoreTerminal implements ITerminal {
     this.register(this._renderService.onRenderedBufferChange(e => this._onRender.fire(e)));
     this.onResize(e => this._renderService!.resize(e.cols, e.rows));
 
+    this._decorationService = this.register(this._instantiationService.createInstance(DecorationService, this.screenElement));
+    this.register(this._renderService.onRefreshRequest(() => this._decorationService?.refresh()));
```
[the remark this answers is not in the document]
meganrogge: that's outdated

css/xterm.css
```diff
@@ -173,3 +173,8 @@
 .xterm-strikethrough {
     text-decoration: line-through;
 }
+
+.xterm-screen canvas .xterm-decoration {
```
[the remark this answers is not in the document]
meganrogge: yeah that's outdated. it's now 

https://github.com/microsoft/vscode/blob/7adaf3ab0ef1d5c6a120b0b69af2a643f04a0b1c/src/vs/workbench/contrib/terminal/browser/media/xterm.css#L181

demo/client.ts
```diff
@@ -525,3 +526,10 @@ function loadTest() {
     term._core._onData.fire('\x03');
   });
 }
+
+function decoration() {
+  const marker = term.addMarker(1);
+  const decoration = term.registerDecoration({ marker });
+  decoration.element.style.backgroundColor = 'red';
+  decoration.element.style.position = 'absolute';
```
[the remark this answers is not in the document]
meganrogge: outdated

src/browser/public/Terminal.ts
```diff
@@ -171,6 +171,10 @@ export class Terminal implements ITerminalApi {
     this._verifyIntegers(cursorYOffset);
     return this._core.addMarker(cursorYOffset);
   }
+  public registerDecoration(decorationOptions: IDecorationOptions): IDecoration | undefined {
+    this._checkProposedApi();
+    return this._core.registerDecoration(decorationOptions);
```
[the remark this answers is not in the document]
meganrogge: I had the validation in Terminal.ts to begin with and thought you said to move it within `DecorationService`? Also don't we want to allow non-integers?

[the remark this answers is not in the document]
meganrogge: I think we want to allow decimal values for width and height - since you might want it to be smaller than the cell width/height. I'll do that for decorationOptions.x though

src/browser/services/DecorationService.ts
```diff
@@ -0,0 +1,134 @@
...
+    return decoration;
+  }
+
+  public refresh(): void {
+    for (const decoration of this._decorations) {
+      if ((decoration.marker.line - this._bufferService.buffers.active.ydisp) < 0 || (decoration.marker.line - this._bufferService.buffers.active.ydisp) > this._bufferService.rows) {
+        // outside of viewport
+        decoration.element.style.display = 'none';
+      } else {
+        decoration.element.style.top = `${(decoration.marker.line - this._bufferService.buffers.active.ydisp) * this._renderService.dimensions.scaledCellHeight}px`;
```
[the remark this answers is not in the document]
meganrogge: thought we wanted to avoid consts

src/browser/services/DecorationService.ts
```diff
@@ -0,0 +1,134 @@
...
+  public get onDispose(): IEvent<void> { return this._onDispose.event; }
+
+  private _onRender = new EventEmitter<HTMLElement>();
+  public get onRender(): IEvent<HTMLElement> { return this._onRender.event; }
+
+  constructor(
+    private readonly _decorationOptions: IDecorationOptions,
+    private readonly _screenElement: HTMLElement,
+    private readonly _renderService: IRenderService,
+    private readonly _bufferService: IBufferService
```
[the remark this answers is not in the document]
meganrogge: outdated

css/xterm.css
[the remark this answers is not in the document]
Tyriar: Just so they're more contained, inspecting screen is a pretty common thing we do right now when investigating renderer problems.

src/browser/services/DecorationService.ts
```diff
@@ -0,0 +1,143 @@
...
+    this._resolveDimensions();
+    this._element.style.width = `${this._decorationOptions.width}px`;
+    this._element.style.height = `${this._decorationOptions.height}px`;
+    this._element.style.top = `${(this.marker.line - this._bufferService.buffers.active.ydisp) * this._renderService.dimensions.scaledCellHeight}px`;
+    if (this._decorationOptions.x && this._decorationOptions.x < 0) {
+      throw new Error(`Decoration options x value cannot be negative, but was ${this._decorationOptions.x}.`);
+    }
+
+    if (this._decorationOptions.anchor === 'right') {
+      this._element.style.right = this._decorationOptions.x ? `${this._decorationOptions.x * this._renderService.dimensions.scaledCellWidth}px` : '';
```
Tyriar, line 123: I think this is equivalent but simpler:

```ts
this._element.style.right =`${(this._decorationOptions.x ?? 0) * this._renderService.dimensions.scaledCellWidth}px`;
```

Maybe better actually is to store the options properties including their default values during the ctor and not keep options as a property. Otherwise embedders could change decoration options after registering the decoration which could lead to weird behavior:

```ts
  constructor(
    options: IDecorationOptions,
    private readonly _screenElement: HTMLElement,
    @IBufferService private readonly _bufferService: IBufferService,
    @IRenderService private readonly _renderService: IRenderService
  ) {
    super();
    this._marker = options.marker;
    this.x = options.x ?? 0;
    ...
```

src/browser/services/DecorationService.ts
```diff
@@ -0,0 +1,143 @@
...
+        continue;
+      }
+      const line = decoration.marker.line - this._bufferService.buffers.active.ydisp;
+      if (line < 0 || line > this._bufferService.rows) {
+        // outside of viewport
+        decoration.element.style.display = 'none';
+      } else {
+        decoration.element.style.top = `${line * this._renderService.dimensions.scaledCellHeight}px`;
+        decoration.element.style.display = 'block';
+      }
```
Tyriar, line 58: Could be simpler to defer all this to the Decoration so this just ends up being:

```ts
for (const decoration of this._decorations) {
  decoration.render();
}
```

Then it's up to the decoration to lazily create the element if needed, show/hide, update position, etc.

src/browser/services/DecorationService.ts
```diff
@@ -0,0 +1,143 @@
...
+  public get onDispose(): IEvent<void> { return this._onDispose.event; }
+
+  private _onRender = new EventEmitter<HTMLElement>();
+  public get onRender(): IEvent<HTMLElement> { return this._onRender.event; }
+
+  constructor(
+    private readonly _decorationOptions: IDecorationOptions,
+    private readonly _screenElement: HTMLElement,
+    @IBufferService private readonly _bufferService: IBufferService,
+    @IRenderService private readonly _renderService: IRenderService
```
Tyriar, line 89: Actually since we want to allow registering decorations before open is called (so shell integration works on backgrounded terminals that haven't been opened yet), these may be undefined so we'll want to pass them into the `Decoration.render` call instead of keeping them as properties.

Tyriar (changes requested):

test/api/Terminal.api.ts
```diff
@@ -731,6 +731,35 @@ describe('API Integration Tests', function(): void {
...
+      try {
+        window.decoration = window.term.registerDecoration({ marker: window.marker, x: -2 });
+      } catch (e) {
+        window.throwMessage = e.message;
+      }
+    `);
+      await pollFor(page, 'window.throwMessage', 'Decoration options x value cannot be negative, but was -2.');
+      assert.equal(await page.evaluate(`document.querySelector('.xterm-screen .xterm-decoration')`), undefined);
+    });
+  });
```
meganrogge, line 761: add a test to make sure decorations can be registered before open is called

Tyriar (approved):

Issue #3634, opened 2022-02-11
Links are not opened if xterm.js is inside an iframe on safari. 


BigDwarf:
Links are not opened if xterm.js is inside an iframe on safari. 
<img width="745" alt="window doesn't open" src="https://user-images.githubusercontent.com/4678964/153626320-45b71415-9f7b-4e59-a8ae-c8585fb8f329.png">

## Details
- Safari Version 14.1.1 (16611.2.7.1.4) 
- Mac OS Big Sur, Version 1.4 (20F71): 
-   "xterm": "4.16.0",
    "xterm-addon-web-links": "0.5.0"

### Steps to reproduce

1. Install `xterm-addon-web-links `extension
2. Open `xterm.js` instance in an iframe
3. Set permissions to an iframe to allow popups and scripts: `sandbox="allow-top-navigation-by-user-activation allow-popups allow-popups-to-escape-sandbox allow-scripts"`
4. Type any website name, f.e `https://google.com`
5. Hover and click on the link above

This is caused by setting window.opener = null https://github.com/xtermjs/xterm.js/blob/1fb11707a527836083d0bdd1275075c3ccf71db7/addons/xterm-addon-web-links/src/WebLinksAddon.ts#L33 . What is the value of opening a windows and setting URL separately?  It could be smth like this (i anyway don't need a window without a URL opened).

```
function handleLink(event: MouseEvent, uri: string): void {
    try {
      window.open(uri, 'noopener');
    } catch (error){
      console.warn('Opening link blocked due to: ', error);
    }
} 

```

#### Minimalistic example to reproduce same behaviour without xterm.js: 

index.html
```
<!DOCTYPE html>
<html>
<body>
  <button onclick="newWindow = window.open(); newWindow.opener = null; newWindow.location.href='https://google.com' ">Button doesn't open a URL</button>
  <button onclick="window.open('https://google.com', 'noopener')">Button opens a URL</button>
</body>
</html>
```
iframe.html
```
<!DOCTYPE html>
<html>
<body>
<iframe
        src="index.html"
        sandbox="allow-top-navigation-by-user-activation allow-popups allow-popups-to-escape-sandbox allow-scripts">
</iframe>
</body>
</html>

```

yiliang114:
I ran into the same problem. I use Webview in VS Code extension and use xterm.js in Webview.

Manojtuguru:
[...]

PR #3635, opened 2022-02-11
Links are not opened if xterm.js is inside an iframe on safari.

BigDwarf:
https://github.com/xtermjs/xterm.js/issues/3634


addons/xterm-addon-web-links/src/WebLinksAddon.ts
```diff
@@ -27,19 +27,14 @@ const end = ')($|' + negatedPathCharacterSet + ')';
...
-    try {
-      newWindow.opener = null;
-    } catch {
-      // no-op, Electron can throw
-    }
-    newWindow.location.href = uri;
-  } else {
-    console.warn('Opening link blocked as opener could not be cleared');
+  try {
+    window.open(uri, 'noopener');
```
Tyriar, line 31: https://developer.mozilla.org/en-US/docs/Web/API/Window/open shows the 3rd arg in `window.open` is `windowFeatures `, not the second?

I'm no expert in which method is more correct, I'm guessing the `window.opener = null` approach is compatible with older versions of browsers which we don't care about anymore. Support for `rel=noopener` which I'd expect to share support here looks good https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener and no callout for 
it on window.open https://developer.mozilla.org/en-US/docs/Web/API/Window/open

Tyriar (commented):
After testing, this actually re-introduces the security issue that this change fixed in the first place so we can't merge it.

PR #3637, opened 2022-02-11
polish decorations

### LINKED ISSUE #3633: make sure to clear markers when eraseInDisplay is called
the current hack so that when `clear` is run in non-windows terminals, the markers for the decorations get disposed of and the decoration elements are removed
https://github.com/microsoft/vscode/blob/0ecbd025a8438bda49fce41afca14398461cdbb6/src/vs/workbench/contrib/terminal/browser/xterm/decorationAddon.ts#L82-L93

### LINKED ISSUE #3636: Decorations should be hidden when the alt buffer is active
Issue described here 

https://github.com/microsoft/vscode/issues/142859

meganrogge:
fixes #3633
fixes #3636

### DIFF
--- a/src/browser/services/DecorationService.ts
+++ b/src/browser/services/DecorationService.ts
@@ -88,6 +88,7 @@ export class Decoration extends Disposable implements IDecoration {
     super();
     this.x = options.x ?? 0;
     this._marker = options.marker;
+    this._marker.onDispose(() => this.dispose());
     this.anchor = options.anchor || 'left';
     this.width = options.width || 1;
     this.height = options.height || 1;
--- a/src/browser/services/DecorationService.ts
+++ b/src/browser/services/DecorationService.ts
@@ -54,16 +54,14 @@ export class DecorationService extends Disposable implements IDecorationService 
     for (const decoration of this._decorations) {
       decoration.dispose();
     }
-    if (this._container) {
-      this._screenElement?.removeChild(this._container);
+    if (this._screenElement && this._container) {
+      this._screenElement.removeChild(this._container);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


src/browser/services/DecorationService.ts
```diff
@@ -44,28 +45,26 @@ export class DecorationService extends Disposable implements IDecorationService
...
+    if (this._screenElement && this._container) {
+      this._screenElement.removeChild(this._container);
     }
   }
 }
 export class Decoration extends Disposable implements IDecoration {
-  private static _nextId = 1;
   private readonly _marker: IMarker;
   private _element: HTMLElement | undefined;
-  private readonly _id: number = Decoration._nextId++;
```
meganrogge, line 68: unused and can just use the marker's ID as I do here

https://github.com/microsoft/vscode/pull/142866/files#diff-f26cf22b79d035fac45646c72f194cc92cf9629bd5cb58e8d8a39d220bb0d5dcR119

src/common/InputHandler.ts
```diff
@@ -1292,6 +1292,7 @@ export class InputHandler extends Disposable implements IInputHandler {
           this._resetBufferLine(j);
         }
         this._dirtyRowService.markDirty(0);
+        this._bufferService.buffer.clearMarkers();
```
Tyriar, line 1295: This might be overly aggressive clearing everything when only the viewport is cleared (`J 2`), any application that echos this sequence intentionally will then clear all our markers. A safer approach would be to clear only the lines that get cleared, so all scrollback for `J 3`, only markers in the viewport for `J 2`, etc. Maybe inside `_resetBufferLine`/`_eraseInBufferLine`?

PR #3649, opened 2022-02-18
on buffer clear, update markers array once each has been disposed of

meganrogge:
found this issue while debugging https://github.com/microsoft/vscode/issues/143330

### DIFF
--- a/src/common/buffer/Buffer.ts
+++ b/src/common/buffer/Buffer.ts
@@ -591,10 +591,12 @@ export class Buffer implements IBuffer {
       for (const marker of this.markers.filter(m => m.line === y)) {
         marker.dispose();
       }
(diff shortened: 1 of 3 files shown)


src/common/buffer/Buffer.ts
```diff
@@ -591,10 +591,12 @@ export class Buffer implements IBuffer {
       for (const marker of this.markers.filter(m => m.line === y)) {
         marker.dispose();
       }
+      this.markers = this.markers.filter(m => m.line !== y);
```
Tyriar, line 594: `markers.splice` inside the for loop should be a bit better for perf

Tyriar (approved):
Makes sense 👍

PR #3650, opened 2022-02-20
Update `node-pty`'s `on` eventListener to `IPty.onData`

jsanders-capstone:
According to `node-pty`s type interface, `on("data")` is deprecated, and should be replaced with `onData`, instead.

- Additionally, to the best of my understandings, the tests are running successfully.

Reference to `node-pty.d.ts`:

```javascript
export interface IPty {
   /* ... */

    /**
     * Adds a listener to the data event, fired when data is returned from the pty.
     * @param event The name of the event.
     * @param listener The callback function.
     * @deprecated Use IPty.onData
     */
    on(event: 'data', listener: (data: string) => void): void;
}
```

I appreciate all that everyone does with the `xterm.js`-- thank you to the maintainers and contributors for actively supporting such a great project.

### DIFF
--- a/demo/server.js
+++ b/demo/server.js
@@ -114,7 +114,7 @@ function startServer() {
     }
     const send = USE_BINARY ? bufferUtf8(ws, 5) : buffer(ws, 5);
 
-    term.on('data', function(data) {
+    term.onData(function (data) {
       try {
         send(data);
       } catch (ex) {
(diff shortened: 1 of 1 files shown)


jsanders-capstone:
Please disregard the previous failures -- it seems they were cancelled due to my latest commit (I updated some spacing to reflect consistent styling).

PR #3655, opened 2022-02-22
get Decoration dispose to actually work and revert document role accessibility change

meganrogge:


src/browser/services/DecorationService.ts
```diff
@@ -122,23 +122,6 @@ export class Decoration extends Disposable implements IDecoration {
...
-        if (!this.marker.isDisposed) {
-          this.marker.dispose();
-        }
-        if (this._element && this._container.contains(this._element)) {
-          this._container.removeChild(this._element);
-        }
-        this.isDisposed = true;
-        // Emit before super.dispose such that dispose listeners get a change to react
-        this._onDispose.fire();
-        super.dispose();
```
Tyriar, line 139: I see `super.dispose()` in the registered one, that shouldn't have been there, was that the problem you're trying to solve here? That would end up calling `super.dispose()` twice.

src/browser/services/DecorationService.ts
```diff
@@ -122,23 +122,6 @@ export class Decoration extends Disposable implements IDecoration {
...
       this._element.style.left = this.x ? `${this.x * renderService.dimensions.actualCellWidth}px` : '';
     }
-    this.register({
-      dispose: () => {
-        if (this.isDisposed) {
-          return;
-        }
-        if (!this.marker.isDisposed) {
-          this.marker.dispose();
-        }
```
Tyriar, line 132: Isn't this a good thing?

meganrogge: it causes problems - I think because of

https://github.com/meganrogge/xterm.js/blob/44bc9a1f588993059819d541e2e68bac2f3e5711/src/browser/services/DecorationService.ts#L87

Tyriar (approved):

src/browser/services/DecorationService.ts
meganrogge: will remove that call. 

This register of dispose() was not working / getting called, which is why this change is necessary.

src/browser/services/DecorationService.ts
meganrogge: I would've thought so too; however, if you have an outline decoration with marker `x` and want to create a diff't decoration at that same marker, disposing of the marker prevents that from happening

PR #3661, opened 2022-02-24
set decoration line height

meganrogge:
So that it has the correct vertical alignment
Screen Shot 2022-02-24 at 12 19 58 PM


meganrogge:
I'm still testing to see what happens when the height is set to something other than 1.

PR #3666, opened 2022-02-28
v4.18.0

meganrogge:


silamon:
@meganrogge, I think you may have forgotten to release the addons along with this version.

Tyriar:
Yeah looks like we need new releases for serialize and ligatures

Issue #3526, opened 2021-10-22
Replace all getOption/setOption usage with options


Tyriar:
Follow up from https://github.com/xtermjs/xterm.js/pull/3448, we should move off the now deprecated `getOption`/`setOption`

PR #3667, opened 2022-03-01
Replace all getOption/setOption usage with options

childrentime:
#3526


typings/xterm.d.ts
```diff
@@ -1060,108 +1060,6 @@ declare module 'xterm' {
      */
     paste(data: string): void;
 
-    /**
-     * Retrieves an option's value from the terminal.
-     * @param key The option key.
-     * @deprecated Use `options` instead.
-     */
-    getOption(key: 'bellSound' | 'bellStyle' | 'cursorStyle' | 'fontFamily' | 'logLevel' | 'rendererType' | 'termName' | 'wordSeparator'): string;
```
Tyriar, line 1068: We'll have to bring this, the xterm-headless types and the public/Terminal.ts implementation back as we can only remove a deprecated feature in a major version.

Tyriar (changes requested):
Looks good, just one comment

typings/xterm.d.ts
childrentime: got it. done.

Tyriar (approved):
Thanks 👍

Issue #1456, opened 2018-05-16
Triple click should select the fully wrapped line if it spans multiple lines


Tyriar:
VS Code: https://github.com/Microsoft/vscode/issues/49935

roryokane:
Possibly related code:

https://github.com/xtermjs/xterm.js/blob/f4da8af2e948f8f7ae48163153f243152dd3efce/src/browser/services/SelectionService.ts#L552-L563

https://github.com/xtermjs/xterm.js/blob/f4da8af2e948f8f7ae48163153f243152dd3efce/src/browser/services/SelectionService.ts#L999-L1009

Some code that might be related to the similar issue https://github.com/microsoft/vscode/issues/135567:

https://github.com/xtermjs/xterm.js/blob/f4da8af2e948f8f7ae48163153f243152dd3efce/src/browser/services/SelectionService.ts#L601-L610

Tyriar:
Moved to track this edge case in https://github.com/xtermjs/xterm.js/issues/3670 instead of reopening

PR #3669, opened 2022-03-02
Fix triple select edge case

### LINKED ISSUE #3670: Triple click and drag up/down should keep the initially clicked wrapped line selected
Special case of https://github.com/xtermjs/xterm.js/issues/1456

Recording 2022-03-02 at 06 13 23

VS Code: https://github.com/microsoft/vscode/issues/135567

silamon:
Keep the length of the (wrapped) line in selectionStartLength instead of selectionEnd. Upon dragging, the selection end is modified, but the wrapped line remains selected thanks to the selectionStartLength. ( Same behavior as selecting a word. )
Tests have been adjusted to include this.

<s>Note: Dragging still select lines, instead of wrapped lines.  </s> EDIT: Fixed as well.

Fixes #3670
Related [microsoft/vscode/#135567](https://github.com/microsoft/vscode/issues/135567)

See also https://github.com/xtermjs/xterm.js/issues/1456#issuecomment-1054849470

### DIFF
--- a/src/browser/selection/SelectionModel.test.ts
+++ b/src/browser/selection/SelectionModel.test.ts
@@ -116,6 +116,12 @@ describe('SelectionModel', () => {
       model.selectionStartLength = 4;
       assert.deepEqual(model.finalSelectionEnd, [2, 3]);
     });
+    it('should return the end on a different row when start + length overflows onto a following row with selectionEnd inbetween', () => {
+      model.selectionStart = [78, 2];
+      model.selectionEnd = [79, 2];
+      model.selectionStartLength = 4;
+      assert.deepEqual(model.finalSelectionEnd, [2, 3]);
+    });
     it('should return selection end if selection end is after selection start + length', () => {
       model.selectionStart = [2, 2];
       model.selectionStartLength = 2;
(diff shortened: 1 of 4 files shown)


Tyriar (changes requested):
Looked into this and it still doesn't look quite right:
Recording 2022-03-02 at 06 11 13

Notice:

- The flash when you drag up
- The wrapped selection is only retained when you drag up
- Dragging down doesn't keep the line selected

Ideally once triple click happens that initially selected wrapped line will always be selected. We're basically after exactly what monaco does:

Recording 2022-03-02 at 06 18 46

silamon:
> Looked into this and it still doesn't look quite right: Recording 2022-03-02 at 06 11 13
> 
> Notice:
> 
> * The flash when you drag up
> * The wrapped selection is only retained when you drag up
> * Dragging down doesn't keep the line selected
> 
> Ideally once triple click happens that initially selected wrapped line will always be selected. We're basically after exactly what monaco does:
> 
> Recording 2022-03-02 at 06 18 46

That's exactly the note which was mentioned in the PR description. It's now also fixed on this branch.

Tyriar (approved):
Works great 👍

PR #3671, opened 2022-03-02
during buffer clear, don't dispose of markers on first line

meganrogge:
discovered via https://github.com/microsoft/vscode/issues/144142

### DIFF
--- a/src/common/buffer/Buffer.ts
+++ b/src/common/buffer/Buffer.ts
@@ -586,7 +586,8 @@ export class Buffer implements IBuffer {
   }
 
   /**
-   * Clears markers on line @param y
+   * Clears markers on single line.
+   * @param y The line to clear.
    */
   public clearMarkers(y: number): void {
     this._isClearing = true;
--- a/src/common/buffer/Buffer.ts
+++ b/src/common/buffer/Buffer.ts
@@ -601,8 +601,8 @@ export class Buffer implements IBuffer {
   }
 
   /**
-   * Clears markers on all lines except for
-   * those on @param excludeY
+   * Clears markers on all lines except for those on a particular line.
+   * @param excludeY The line to exclude.
    */
    public clearAllMarkers(excludeY: number): void {
     this._isClearing = true;
--- a/src/common/buffer/Buffer.ts
+++ b/src/common/buffer/Buffer.ts
@@ -591,12 +591,12 @@ export class Buffer implements IBuffer {
    */
   public clearMarkers(y: number): void {
     this._isClearing = true;
-      for (let i = 0; i < this.markers.length; i++) {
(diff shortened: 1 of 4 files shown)


src/common/buffer/Buffer.ts
```diff
@@ -595,10 +595,14 @@ export class Buffer implements IBuffer {
...
     } else {
-      for (const marker of this.markers) {
-        marker.dispose();
+      // buffer has been cleared
+      // don't dispose of markers on the current line (0)
+      for (let i = 0; i < this.markers.length; i++) {
+        if (this.markers[i].line !== 0) {
+          this.markers[i].dispose();
+          this.markers.splice(i--, 1);
+        }
```
Tyriar, line 604: The method name is a bit confusing now, I suggest we break this into two methods such as:

```ts
clearMarker(y: number): void;
clearAllMarkers(excludeY: number): void;
```

We should also add some jsdoc like most of the other methods in Buffer

Tyriar (changes requested):

src/common/buffer/Buffer.ts
```diff
@@ -585,24 +585,31 @@ export class Buffer implements IBuffer {
     return x >= this._cols ? this._cols - 1 : x < 0 ? 0 : x;
   }
 
-  public clearMarkers(y?: number): void {
+  /**
+   * Clears markers on line @param y
+   */
```
Tyriar, line 590:
```suggestion
  /**
   * Clears markers on single line.
   * @param y The line to clear.
   */
```

src/common/buffer/Buffer.ts
```diff
@@ -585,24 +585,31 @@ export class Buffer implements IBuffer {
...
-    } else {
-      // buffer has been cleared
-      // don't dispose of markers on the current line (0)
+    this._isClearing = false;
+  }
+
+  /**
+   * Clears markers on all lines except for
+   * those on @param excludeY
+   */
```
Tyriar, line 605:
```suggestion
  /**
   * Clears markers on all lines except for those on a particular line.
   * @param excludeY The line to exclude.
   */
```

src/common/buffer/Types.d.ts
```diff
@@ -45,7 +45,8 @@ export interface IBuffer {
   getNullCell(attr?: IAttributeData): ICellData;
   getWhitespaceCell(attr?: IAttributeData): ICellData;
   addMarker(y: number): IMarker;
-  clearMarkers(y?: number): void;
+  clearMarkers(y: number): void;
+  clearAllMarkers(excludeY: number): void;
```
Tyriar, line 49: I don't think this is getting called yet

Tyriar (changes requested):

PR #3673, opened 2022-03-03
disallow input during buffer clear

### LINKED ISSUE #3672: input parsed during a buffer clear is incorrectly positioned and duplicated
Reproducible in powershell:
- run some commands
- run terminal: clear (or trigger buffer clear)
- quickly type some input
- 🐛  it's not positioned correctly

reported here https://github.com/microsoft/vscode/issues/144292#event-6174822792

meganrogge:
fixes #3672

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -106,6 +106,12 @@ export class Terminal extends CoreTerminal implements ITerminal {
    */
   private _unprocessedDeadKey: boolean = false;
 
+  /**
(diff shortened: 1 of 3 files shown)


meganrogge:
ideally it would allow the input and store/queue it, but not write it until the cursor position has been updated/ all of the rows have been refreshed.

meganrogge:
can actually still repro the issue 😦

Tyriar:
See https://github.com/microsoft/vscode/issues/144292, nothing we can do about this afaik

PR #3676, opened 2022-03-09
add overview ruler 

### LINKED ISSUE #3632: Allow decorations to show up in the scroll bar
Similar to monaco:

image

meganrogge:
fixes #3632


src/browser/Terminal.ts
```diff
@@ -471,6 +472,12 @@ export class Terminal extends CoreTerminal implements ITerminal {
     this._viewportElement = document.createElement('div');
     this._viewportElement.classList.add('xterm-viewport');
     fragment.appendChild(this._viewportElement);
+
+    // TODO: make this opt in, must be done before the scroll area in order to show up
```
meganrogge, line 476: not sure how can I make this opt in given that this canvas element must be appended to the viewport before the scroll area is to work

src/browser/services/DecorationService.ts
```diff
@@ -3,72 +3,196 @@
...
+import { addDisposableDomListener } from 'browser/Lifecycle';
 import { IDecorationService, IRenderService } from 'browser/services/Services';
 import { EventEmitter, IEvent } from 'common/EventEmitter';
 import { Disposable } from 'common/Lifecycle';
+import { BufferService } from 'common/services/BufferService';
 import { IBufferService, IInstantiationService } from 'common/services/Services';
 import { IDecorationOptions, IDecoration, IMarker } from 'xterm';
 
+const enum ScrollbarConstants {
+  WIDTH = 7
```
meganrogge, line 15: should this be `actualCellWidth` instead?

meganrogge: or passed in via the `DecorationOptions`?

src/browser/Terminal.ts
meganrogge: seems like this messes with things/ causes api tests to fail

Tyriar: Before the `this._viewportScrollArea` you mean? You can use `insertBefore` instead of `appendChild`. That combined with an option like `overviewRuler? boolean` should do the trick.

src/browser/services/DecorationService.ts
Tyriar: Here's monaco's API:

https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.IModelDecorationOverviewRulerOptions.html

You can set a color, a darkColor and a position (full, left, right, center). Getting that same granularity around positioning would be good, I don't think we need a dark color though.

src/browser/services/DecorationService.ts
```diff
@@ -3,72 +3,198 @@
...
 import { EventEmitter, IEvent } from 'common/EventEmitter';
 import { Disposable } from 'common/Lifecycle';
 import { IBufferService, IInstantiationService } from 'common/services/Services';
 import { IDecorationOptions, IDecoration, IMarker } from 'xterm';
 
+const enum ScrollbarConstants {
+  WIDTH = 7
+}
+
 export class DecorationService extends Disposable implements IDecorationService {
```
Tyriar, line 17: It looks like this class does a little too much now, how about splitting it up like this?

Screen Shot 2022-03-11 at 12 06 50 pm

That way all DecorationService does is collects and manages all the decorations and fires events for components that depend on decorations.

src/browser/services/DecorationService.ts
```diff
@@ -3,72 +3,198 @@
...
-  private _screenElement: HTMLElement | undefined;
   private _renderService: IRenderService | undefined;
   private _animationFrame: number | undefined;
 
-  constructor(@IInstantiationService private readonly _instantiationService: IInstantiationService) { super(); }
+  private _screenElement: HTMLElement | undefined;
+  private _viewportElement: HTMLElement | undefined;
+  private _bufferDecorationContainer: HTMLElement | undefined;
+  private _scrollbarDecorationCanvas: CanvasRenderingContext2D | null = null;
+  private _scrollbarDecorationNode: HTMLCanvasElement | undefined;
```
Tyriar, line 26: Naming that aligns with other areas:

```ts
  private _scrollbarDecorationCtx: CanvasRenderingContext2D | null = null;
  private _scrollbarDecorationCanvas: HTMLCanvasElement | undefined;
```

src/browser/services/DecorationService.ts
```diff
@@ -3,72 +3,198 @@
...
+
+  private readonly _bufferDecorations: BufferDecoration[] = [];
+  private _scrollbarDecorations: ScrollbarDecoration[] = [];
```
Tyriar, line 29: There should only be a single list of decorations imo, it's up to the consumer of the decoration to filter out appropriately:

```ts
// Imagine after overview ruler refactor
for (const decoration of decorationService.decorations) {
  if (!decoration.scrollbarDecorationColor) {
    continue;
  }
  // do the thing
}
```

src/browser/services/DecorationService.ts
```diff
@@ -3,72 +3,198 @@
...
-    this.register(this._renderService.onRenderedBufferChange(() => this.refresh()));
-    this.register(this._renderService.onDimensionsChange(() => this.refresh(true)));
+    this._viewportElement = viewportElement;
+    this._scrollbarDecorationNode = scrollbarDecorationNode;
+
+    this.register(this._renderService.onRenderedBufferChange(() => this._refresh()));
+    this.register(this._renderService.onDimensionsChange(() => this._refresh(true)));
+    this.register(addDisposableDomListener(window, 'resize', () => this._refreshScollbarDecorations()));
+    this.register(this._bufferService.buffers.onBufferActivate(() => {
+      this._scrollbarDecorationNode!.style.display = this._bufferService.buffer === this._bufferService.buffers.alt ? 'none' : 'block';
```
Tyriar, line 43: `!` is unsafe here since `_scrollbarDecorationNode` is not read only. Another reason splitting out the service from view is good is we could initialize the overview ruler only after `open` has been called so it could be readonly.

src/browser/services/DecorationService.ts
```diff
@@ -3,72 +3,198 @@
...
+export class ScrollbarDecoration extends Disposable implements IDecoration {
+  private readonly _marker: IMarker;
+  private _element: HTMLCanvasElement | undefined;
+  private _color: string | undefined;
+
+  public isDisposed: boolean = false;
+
+  public get element(): HTMLCanvasElement { return this._element!; }
+  public get marker(): IMarker { return this._marker; }
+  public get color(): string { return this._color!; }
```
Tyriar, line 146: More `!` we want to avoid

src/browser/services/DecorationService.ts
```diff
@@ -3,72 +3,198 @@
...
+    super();
+    this._marker = options.marker;
+    this._color = options.scrollbarDecorationColor;
+    this._marker.onDispose(() => this.dispose());
+    this.render();
+  }
+  public render(): void {
+    if (!this._element) {
+      this._element = this._canvas;
     }
```
Tyriar, line 169: `_element` could be merged into `_canvas`?

src/browser/services/DecorationService.ts
```diff
@@ -3,72 +3,198 @@
...
-    if (this._screenElement && this._container && this._screenElement.contains(this._container)) {
-      this._screenElement.removeChild(this._container);
+    this._ctx.lineWidth = 1;
+    this._ctx.strokeStyle = this.color;
+    this._ctx.strokeRect(
+      0,
+      this.element.height * (this.marker.line / this._bufferService.buffers.active.lines.length),
+      this.element.width,
+      window.devicePixelRatio
```
Tyriar, line 176: This will make the line blurry when 1 < dpr < 2.

src/browser/services/DecorationService.ts
```diff
@@ -3,72 +3,198 @@
...
+  public override dispose(): void {
+    if (this._isDisposed) {
+      return;
     }
+    this._ctx.clearRect(
+      0,
+      this.element.height * (this.marker.line / this._bufferService.buffers.active.lines.length),
+      this.element.width,
+      window.devicePixelRatio
+    );
```
Tyriar, line 190: Because of the blurriness, this probably won't be sufficient. If multiple decorations are around the same zone and devicePixelRatio is a decimal, this `clearRect` could end up clearing other decorations as well. I think instead of getting the exact decimal value, you'll want to get the closest pixel on the canvas and fill that in, then maybe use `Math.round(window.devicePixelRatio)` for the line size to ensure that's a crisp pixel-perfect rectangle regardless of zoom.

Tyriar (changes requested):
One theme I've noticed that I think we should move away from is the idea of buffer decorations vs scroll bar decorations. Instead there are decorations, all decorations are rendered in the buffer and some get rendered in the scroll bar. I think the latter is how it works but most of the code is structured in the other direction.

css/xterm.css
```diff
@@ -178,3 +178,16 @@
 	z-index: 6;
 	position: absolute;
 }
+
+.xterm-decoration-scrollbar {
+    z-index: 7;
+    position: sticky;
```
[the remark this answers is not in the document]
meganrogge: as we discussed, `sticky` can work here, but going to go with absolute in order to avoid the native scrollbar overlay issue

src/browser/services/DecorationService.ts
meganrogge: thought we wanted `decoration.element` to be undefined until the first `render` happens?

src/browser/services/DecorationService.ts
meganrogge: not sure we need to support `center` as I don' think that's practical for the terminal.

src/browser/services/DecorationService.ts
meganrogge: didn't want to have to filter them out, but yeah the services refactor will prevent that

src/browser/services/DecorationService.ts
meganrogge: by the line size, do you mean `lineWidth`?

meganrogge:
To do or to defer:

- api tests

Tyriar (commented):
Sorry forgot to post this, most are probably out of date

src/browser/Decorations/OverviewRulerRenderer.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+    if (!decoration.options.overviewRulerOptions?.color) {
+      this._decorationElements.delete(decoration);
+      return;
+    }
+    this._ctx.lineWidth = 1;
+    this._ctx.strokeStyle = decoration.options.overviewRulerOptions.color;
+    const size = Math.floor(this._optionsService.options.overviewRulerWidth / 3);
+    const position = decoration.options.overviewRulerOptions.position;
+    this._ctx.strokeRect(
```
Tyriar, line 86: In monaco the height of the decoration depends on the position:

image

Looks like full has a height of 2 CSS pixels and non-full has a height of 6 CSS pixels.

The decorations are also really large when the file has little contents, I think we can avoid doing that though. This is showing a 5 line file with a single word highlighted:

image

src/browser/Terminal.ts
```diff
@@ -71,12 +74,18 @@ export class Terminal extends CoreTerminal implements ITerminal {
...
 
   private _customKeyEventHandler: CustomKeyEventHandler | undefined;
 
+  // TODO: Move into CoreTerminal.ts
+  // common services
+  private _decorationService: DecorationService;
+
   // browser services
   private _charSizeService: ICharSizeService | undefined;
   private _mouseService: IMouseService | undefined;
```
Tyriar, line 91: Actually, I changed my mind about this overnight. Since we're not supporting decorations in xterm-headless at all, this should live in the `browser/` file so code that never runs doesn't end up showing up in the headless bundle. I'm a little on the fence about whether `DecorationService` should live in common or browser, let's keep it in common for now since it won't be included in the compiled bundle unless it's referenced.

```suggestion
  // browser services
  private _decorationService: DecorationService;
  private _charSizeService: ICharSizeService | undefined;
  private _mouseService: IMouseService | undefined;
```

src/browser/Terminal.ts
```diff
@@ -576,8 +586,7 @@ export class Terminal extends CoreTerminal implements ITerminal {
     this.register(this.onScroll(() => this._mouseZoneManager!.clearAll()));
     this.linkifier.attachToDom(this.element, this._mouseZoneManager);
     this.linkifier2.attachToDom(this.screenElement, this._mouseService, this._renderService);
-
-    this.decorationService.attachToDom(this.screenElement, this._renderService, this._bufferService);
+    this._instantiationService.createInstance(BufferDecorationRenderer, this.screenElement);
```
Tyriar, line 589:
```suggestion
    this.register(this._instantiationService.createInstance(BufferDecorationRenderer, this.screenElement));
```

src/browser/tsconfig.json
```diff
@@ -18,7 +18,7 @@
   "include": [
     "./**/*",
     "../../typings/xterm.d.ts"
-  ],
+],
```
Tyriar, line 21: I might have caused this 🤷 

```suggestion
  ],
```

typings/xterm.d.ts
```diff
@@ -266,6 +266,11 @@ declare module 'xterm' {
      * All features are disabled by default for security reasons.
      */
     windowOptions?: IWindowOptions;
+
+    /**
+     * The width, in pixels, of the canvas for the overview ruler.
```
Tyriar, line 271:
```suggestion
     * The width, in pixels, of the canvas for the overview ruler. The overview
     * ruler will be hidden when not set.
```

typings/xterm.d.ts
```diff
@@ -394,7 +399,7 @@ declare module 'xterm' {
   }
 
   /**
-   * Represents a disposable with an 
+   * Represents a disposable with an
```
Tyriar, line 402:
```suggestion
   * Represents a disposable that tracks is disposed state.
```

typings/xterm.d.ts
```diff
@@ -427,16 +432,16 @@ declare module 'xterm' {
     readonly onRender: IEvent<HTMLElement>;
 
     /**
-     * The HTMLElement that gets created after the 
-     * first _onRender call, or undefined if accessed before
+     * The HTMLElement that gets created or drawn to (for scrollbar decorations)
+     * after the first _onRender call, or undefined if accessed before
```
Tyriar, line 436:
```suggestion
     * The element that the decoration is rendered to. This will be undefined
     * until it is rendered for the first time by @{link IDecoration.onRender}.
```

typings/xterm.d.ts
```diff
@@ -455,21 +460,30 @@ declare module 'xterm' {
...
+     */
     x?: number;
 
 
     /**
-     * The width of the decoration in cells, which defaults to 
-     * cell width
+     * The width of the decoration in cells, which defaults to
+     * cell width or the width in pixels, when an overlayRulerItemColor
+     * is provided.
```
Tyriar, line 470:
```suggestion
     * The width of the decoration in cells, defaults to 1.
```

typings/xterm.d.ts
```diff
@@ -455,21 +460,30 @@ declare module 'xterm' {
...
      */
     width?: number;
 
     /**
-     * The height of the decoration in cells, which defaults to 
+     * The height of the decoration in cells, which defaults to
      * cell height
```
Tyriar, line 476:
```suggestion
     * The height of the decoration in cells, defaults to 1.
```

typings/xterm.d.ts
```diff
@@ -455,21 +460,30 @@ declare module 'xterm' {
...
      */
     height?: number;
+
+    /**
+     * Renders the decoration in the scrollbar
+     * with the given @param color and optional @param position.
+     * If @param position is not set, it will span the full @param overviewRulerWidth, which
+     * must be provided via @TerminalOptions for this to work.
+     */
```
Tyriar, line 485:
```suggestion
    /**
     * When defined, renders the decoration in the overview ruler to the right
     * of the terminal. {@link ITerminalOptions.overviewRulerWidth} must be set
     * in order to see the overview ruler.
     * @param color The color of the decoration
     * @param position Tge position of the decoration.
     */
```

Tyriar (changes requested):

src/browser/Decorations/OverviewRulerRenderer.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  private _refreshStyle(decoration: IInternalDecoration): void {
+    if (!this._ctx || !this._optionsService.options.overviewRulerWidth) {
+      return;
+    }
+    if (decoration.options.anchor === 'right') {
+      this._canvas.style.right = decoration.options.x ? `${decoration.options.x * this._renderService.dimensions.actualCellWidth}px` : '';
+    } else {
+      this._canvas.style.left = decoration.options.x ? `${decoration.options.x * this._renderService.dimensions.actualCellWidth}px` : '';
+    }
```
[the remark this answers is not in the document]
meganrogge: we also want to do this the first time the decoration is rendered I think? went with `updateAnchor`

src/browser/Decorations/OverviewRulerRenderer.ts
meganrogge: it looks really weird when I do 2 and 6 pixels for the height
Screen Shot 2022-03-15 at 6 02 27 PM

demo/client.ts
```diff
@@ -546,7 +547,20 @@ function loadTest() {
 function addDecoration() {
   const marker = term.addMarker(1);
   const decoration = term.registerDecoration({ marker });
```
Tyriar, line 549: Can we add overview ruler options to this one too?

demo/client.ts
```diff
@@ -546,7 +547,20 @@ function loadTest() {
...
 }
+
+function addOverviewRuler() {
+  term.options['overviewRulerWidth'] = 15;
+  term.registerDecoration({marker: term.addMarker(1), overviewRulerOptions: { color: '#ef2929' }});
+  term.registerDecoration({marker: term.addMarker(3), overviewRulerOptions: { color: '#8ae234' }});
+  term.registerDecoration({marker: term.addMarker(5), overviewRulerOptions: { color: '#729fcf' }});
+  term.registerDecoration({marker: term.addMarker(7), overviewRulerOptions: { color: '#ef2929', position: 'left' }});
+  term.registerDecoration({marker: term.addMarker(7), overviewRulerOptions: { color: '#8ae234', position: 'center' }});
+  term.registerDecoration({marker: term.addMarker(7), overviewRulerOptions: { color: '#729fcf', position: 'right' }});
```
Tyriar, line 564: The alignment seems off with these:

image

Also blue/right is being drawn largest?

Tyriar (approved):

demo/client.ts
meganrogge: yeah thought we shouldn't add a bunch of listeners to update that based on when zoom happens etc. it is correctly aligned when zoom level is 100%
Screen Shot 2022-03-15 at 8 44 28 PM

meganrogge: spent awhile and am sort of stumped about what's going on with the ones on the same line - they're all too large, but the widths that get set and x positions are correct/ what I'd expect
Screen Shot 2022-03-15 at 6 10 25 PM

PR #3689, opened 2022-03-16
allow changing the color for overviewRuler decorations

meganrogge:


src/browser/Decorations/OverviewRulerRenderer.ts
```diff
@@ -92,7 +92,7 @@ export class OverviewRulerRenderer extends Disposable {
       return;
     }
     this._ctx.lineWidth = 1;
-    this._ctx.fillStyle = decoration.options.overviewRulerOptions.color;
+    this._ctx.fillStyle = decoration.overviewRulerDecorationColor || decoration.options.overviewRulerOptions.color;
```
Tyriar, line 95: Let's initialize `overviewRulerDecorationColor` to `options.color`  in `Decoration's` ctor, we can then avoid the `||` here and the `| undefined` in the API.

typings/xterm.d.ts
```diff
@@ -438,6 +438,13 @@ declare module 'xterm' {
      * that.
      */
     element: HTMLElement | undefined;
+
+    /**
+     * The color to be used for the overview ruler decoration.
+     * This will only take effect when @param overviewRulerOptions
+     * were provided initially.
+     */
+    overviewRulerDecorationColor: string | undefined;
```
Tyriar, line 447: Just realized we were putting an overview ruler thing into IDecoration. I think keeping it in its own object like in options would be better for consistency, like this:

```ts
overviewRuler?: { color: string }
```

Them may as well give `position` the same treatment and pull it into an interface:

```ts
readonly overviewRuler?: IDecorationOverviewRulerOptions;

...

interface IDecorationOverviewRulerOptions {
  color: string;
  position?: 'left' | 'center' | 'right';
}
```

Changing position could be useful for switching between the "cursor" overview ruler decoration into a command one as well.

Tyriar (changes requested):

typings/xterm.d.ts
meganrogge: yeah wondered about that too. will do

src/browser/Decorations/OverviewRulerRenderer.ts
meganrogge: can't avoid | undefined in the API because if `overviewRulerOptions` aren't provided, and thus `options.overviewRulerOptions.color` is undefined, that should be undefined

meganrogge: also can't avoid the || here now either

src/browser/Decorations/OverviewRulerRenderer.ts
```diff
@@ -92,7 +92,7 @@ export class OverviewRulerRenderer extends Disposable {
       return;
     }
     this._ctx.lineWidth = 1;
-    this._ctx.fillStyle = decoration.options.overviewRulerOptions.color;
+    this._ctx.fillStyle = decoration.overviewRulerOptions?.color || decoration.options.overviewRulerOptions.color;
```
meganrogge, line 95: this can't be avoided since we don't know if it's been set since the last refresh

typings/xterm.d.ts
```diff
@@ -438,8 +438,21 @@ declare module 'xterm' {
...
+     * were provided initially.
+     */
+     overviewRulerOptions?: Pick<IDecorationOverviewRulerOptions, 'color'>;
   }
 
+
+interface IDecorationOverviewRulerOptions {
+  color: string;
+  position?: 'left' | 'center' | 'right';
+}
```
Tyriar, line 454: Just indentation and jsdoc here

typings/xterm.d.ts
```diff
@@ -438,8 +438,21 @@ declare module 'xterm' {
      * that.
      */
     element: HTMLElement | undefined;
+
+    /**
+     * The options for the overview ruler that can be updated.
+     * This will only take effect when @param overviewRulerOptions
+     * were provided initially.
+     */
```
Tyriar, line 446:
```suggestion
    /**
     * The options for the overview ruler that can be updated. This will only
	 * take effect when {@link IDecorationOptions.overviewRulerOptions} were
	 * provided initially.
     */
```

PR #3693, opened 2022-03-17
clear overview ruler canvas when decoration is disposed

### LINKED ISSUE #3691: Overview ruler decorations don't disappear when the decoration is disposed of
Notice the blue outline disappears but the overview ruler doesn't get re-rendered:

Recording 2022-03-17 at 08 57 48

meganrogge:
fix #3691

### DIFF
--- a/src/browser/Decorations/OverviewRulerRenderer.ts
+++ b/src/browser/Decorations/OverviewRulerRenderer.ts
@@ -67,12 +67,7 @@ export class OverviewRulerRenderer extends Disposable {
 
   public override dispose(): void {
     for (const decoration of this._decorationElements) {
-      this._ctx?.clearRect(
-        0,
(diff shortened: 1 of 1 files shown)


src/browser/Decorations/OverviewRulerRenderer.ts
```diff
@@ -119,6 +114,15 @@ export class OverviewRulerRenderer extends Disposable {
     const element = this._decorationElements.get(decoration);
     if (!element) {
       this._decorationElements.set(decoration, this._canvas);
+      decoration.onDispose(() => {
```
Tyriar, line 117: I think `decoration.onDispose(() => this.queueRefresh())` is what we're after here, otherwise the repro would cause the overview ruler to be empty until it's next rendered.

Tyriar (changes requested):

PR #3698, opened 2022-03-18
Show all matches when searching

### LINKED ISSUE #1653: Search addon: Introduce a new selection type so that multiple ranges can be highlighted at once
This would allow showing all matches of a search in the viewport and also not hijacking the regular selection.

Plan issue: https://github.com/xtermjs/xterm.js/issues/705

meganrogge:
Fixes #1653
Part of microsoft/vscode#134790


addons/xterm-addon-search/typings/xterm-addon-search.d.ts
```diff
@@ -50,19 +50,27 @@ declare module 'xterm-addon-search' {
...
     /**
-     * Search forwards for the next result that matches the search term and
-     * options.
+     * Find all instances of the term, selecting the next one with each
+     * enter. If it doesn't exist, do nothing.
      * @param term The search term.
      * @param searchOptions The options for the search.
      */
-    public findNext(term: string, searchOptions?: ISearchOptions): boolean;
+      public find(term: string, searchOptions?: ISearchOptions): boolean;
```
Tyriar, line 58: Is this new API needed? VS Code builds on the current API with the `incremental` option to implement this behavior.

I was expecting a new `highlightAllMatches?: boolean` option or something to turn it on.

css/xterm.css
```diff
@@ -179,6 +179,10 @@
 	position: absolute;
 }
 
+.xterm-find-result-decoration {
+   opacity: 60%;
```
Tyriar, line 183: nit

```suggestion
   opacity: 0.6;
```

addons/xterm-addon-search/typings/xterm-addon-search.d.ts
Tyriar, line 58: `overviewRulerResultDecorationColor` and `overviewRulerSelectionDecorationColor` are missing from the options

demo/client.ts
```diff
@@ -200,6 +203,11 @@ function createTerminal(): void {
 
   addDomListener(paddingElement, 'change', setPadding);
 
+  addDomListener(actionElements.find, 'keyup', (e) => {
+    term.options.overviewRulerWidth = 10;
```
Tyriar, line 207: Let's leave this out and only turn the overview ruler on when the existing button is pressed.

typings/xterm.d.ts
```diff
@@ -956,7 +956,7 @@ declare module 'xterm' {
      * @param cursorYOffset The y position offset of the marker from the cursor.
      * @returns The new marker or undefined.
      */
-    registerMarker(cursorYOffset?: number): IMarker | undefined;
+    registerMarker(cursorYOffset?: number, row?: number): IMarker | undefined;
```
Tyriar, line 959: I've contemplated an API like this but it couldn't be another optional arg like this. Can create an issue about it to discuss there and in the meantime always do the math (`ybase + cursory - row`)

src/browser/Types.d.ts
```diff
@@ -61,7 +61,7 @@ export interface IPublicTerminal extends IDisposable {
   registerLinkProvider(linkProvider: ILinkProvider): IDisposable;
   registerCharacterJoiner(handler: (text: string) => [number, number][]): number;
   deregisterCharacterJoiner(joinerId: number): void;
-  addMarker(cursorYOffset: number): IMarker | undefined;
+  addMarker(cursorYOffset: number, col?: number, row?: number): IMarker | undefined;
```
Tyriar, line 64: Accident to include col? Also see other comment about this.

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -53,10 +62,83 @@ export class SearchAddon implements ITerminalAddon {
 
   public activate(terminal: Terminal): void {
     this._terminal = terminal;
+    this._terminal.onData(() => this._dataChanged = true);
```
Tyriar, line 65: Need to track the event listener disposable and dispose it when the addon is disposed.

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -3,13 +3,17 @@
...
-import { Terminal, IBufferLine, IDisposable, ITerminalAddon, ISelectionPosition } from 'xterm';
+import { Terminal, IDisposable, ITerminalAddon, ISelectionPosition, IDecoration } from 'xterm';
 
 export interface ISearchOptions {
   regex?: boolean;
   wholeWord?: boolean;
   caseSensitive?: boolean;
   incremental?: boolean;
+  startRow?: number;
+  startCol?: number;
```
Tyriar, line 14: Why are these needed?

src/browser/Terminal.ts
```diff
@@ -1012,12 +1012,14 @@ export class Terminal extends CoreTerminal implements ITerminal {
...
   }
 
-  public addMarker(cursorYOffset: number): IMarker | undefined {
+  public addMarker(cursorYOffset: number, row?: number): IMarker | undefined {
     // Disallow markers on the alt buffer
     if (this.buffer !== this.buffers.normal) {
       return;
     }
-
+    if (row) {
```
meganrogge, line 1020: currently improving the name

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -463,4 +557,29 @@ export class SearchAddon implements ITerminalAddon {
...
+
+    const findResultDecoration = terminal.registerDecoration({ marker, overviewRulerOptions: this._resultDecorations.get(marker.line) && !this._dataChanged ? undefined : { color, position: 'center' } });
+    findResultDecoration?.onRender((e) => {
+      if (!e.classList.contains('xterm-find-result-decoration') && result.term.length && e.clientWidth > 0) {
+        e.classList.add('xterm-find-result-decoration');
+        // decoration's clientWidth = actualCellWidth
+        e.style.left = `${e.clientWidth * result.col}px`;
+        e.style.width = `${e.clientWidth * result.term.length}px`;
+        e.style.backgroundColor = color;
+        e.style.color = color;
```
Tyriar: Think we only want bg color

Tyriar (changes requested):

src/browser/Terminal.ts
meganrogge: actually, I think `row` is the most appropriate here.

addons/xterm-addon-search/typings/xterm-addon-search.d.ts
meganrogge: I thought we wanted them in terminal.options, but believe you said add them to ISearchOptions instead?

addons/xterm-addon-search/src/SearchAddon.ts
meganrogge: try:
- search for something "a"
- add lines to the buffer
- press enter
- the decorations are added to the existing ones and the search only had to start at the last covered row/col

https://github.com/meganrogge/xterm.js/blob/66dcd1694c915bcf56e08b354d0037407b0e4194/addons/xterm-addon-search/src/SearchAddon.ts#L100-L105

addons/xterm-addon-search/typings/xterm-addon-search.d.ts
meganrogge: since this API won't exist, it makes sense to put them in the terminal options instead

demo/client.ts
meganrogge: That would make it hard to test from a clean state though
Screen Shot 2022-03-18 at 3 30 18 PM

addons/xterm-addon-search/typings/xterm-addon-search.d.ts
meganrogge: okay will think about that

meganrogge: working on it

addons/xterm-addon-search/src/SearchAddon.ts
Tyriar: Oh I see, I think this is done with `incremental` using the selection right now? How about tracking the active marker line and offset to know where to pick up instead?

meganrogge: that's essentially what this is doing, but yeah can set those as class variables instead of passing them in

Tyriar: Better to keep state like that out of the DOM, can it be tracked by keeping a reference to the decoration, marker or index of decoration in an array, etc.?

meganrogge: that's what this is doing, but yeah I've made that more globally stored now

```
// set start row and col to avoid redoing work
      const key = Array.from(this._searchResults.keys()).pop()?.split('-');
      if (key?.length === 2) {
        startRow = Number.parseInt(key[0]);
        startCol = Number.parseInt(key[1]) + 1;
      }
```

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -3,13 +3,19 @@
...
 
 export interface ISearchOptions {
   regex?: boolean;
   wholeWord?: boolean;
   caseSensitive?: boolean;
   incremental?: boolean;
+  decorations?: IDecorationColor;
+}
+
+interface IDecorationColor {
```
Tyriar, line 16: Let's make this `IDecorationOptions` so it's more flexible if we need to add non-color things in the future without breaking the API.

addons/xterm-addon-search/typings/xterm-addon-search.d.ts
Tyriar, line 58: We removing this one in favor of findNext/findPrevious?

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -53,9 +66,30 @@ export class SearchAddon implements ITerminalAddon {
...
   public activate(terminal: Terminal): void {
     this._terminal = terminal;
+    this._onDataDisposable = this._terminal.onData(() => {
+      this._dataChanged = true;
+      setTimeout(() => {
+        if (this._lastSearchOptions?.decorations && this._cachedSearchTerm && this._resultDecorations.size > 0 && this._lastSearchOptions) {
+          this._highlightAllMatches(this._cachedSearchTerm, this._lastSearchOptions,'previous');
+        }
+      }, 200);
+    });
```
Tyriar, line 76: This should get debounced to ensure we're not doing this every 200 ms, but only 200ms after no event

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -53,9 +66,30 @@ export class SearchAddon implements ITerminalAddon {
 
   public activate(terminal: Terminal): void {
     this._terminal = terminal;
+    this._onDataDisposable = this._terminal.onData(() => {
+      this._dataChanged = true;
+      setTimeout(() => {
+        if (this._lastSearchOptions?.decorations && this._cachedSearchTerm && this._resultDecorations.size > 0 && this._lastSearchOptions) {
+          this._highlightAllMatches(this._cachedSearchTerm, this._lastSearchOptions,'previous');
+        }
+      }, 200);
```
Tyriar, line 75: This create a new timeout on every onData event, there should only be a single timeout active at any time.

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -53,9 +66,30 @@ export class SearchAddon implements ITerminalAddon {
 
   public activate(terminal: Terminal): void {
     this._terminal = terminal;
+    this._onDataDisposable = this._terminal.onData(() => {
+      this._dataChanged = true;
+      setTimeout(() => {
+        if (this._lastSearchOptions?.decorations && this._cachedSearchTerm && this._resultDecorations.size > 0 && this._lastSearchOptions) {
+          this._highlightAllMatches(this._cachedSearchTerm, this._lastSearchOptions,'previous');
```
Tyriar, line 73:
```suggestion
          this._highlightAllMatches(this._cachedSearchTerm, this._lastSearchOptions, 'previous');
```

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -68,9 +102,60 @@ export class SearchAddon implements ITerminalAddon {
     if (!this._terminal) {
       throw new Error('Cannot use addon until it has been loaded');
     }
+    this._lastSearchOptions = searchOptions;
+    return searchOptions?.decorations ? this._highlightAllMatches(term, searchOptions, 'next') : this._findAndSelectNext(term, searchOptions);
+  }
 
+  private _highlightAllMatches(term: string, searchOptions: ISearchOptions, selectionType: 'next' | 'previous'): boolean {
```
Tyriar, line 109: I prefer const enums for strictly internal params like `selectionType`. I believe equality checks in Blink at least are O(1) as strings of the same value all share the same backing string, but I don't know if that's global for all javascript engines. For a const enum the backing values could be 0/1 which are guaranteed to be fast.

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -53,9 +66,30 @@ export class SearchAddon implements ITerminalAddon {
...
+    });
   }
 
-  public dispose(): void { }
+  public dispose(): void {
+    this.clear();
+    this._onDataDisposable?.dispose();
+  }
+
+  public clear(): void {
```
Tyriar, line 84: Does this need to be on the .d.ts? Should this be named `clearDecorations`?

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -53,9 +66,30 @@ export class SearchAddon implements ITerminalAddon {
...
+    this._selectedDecoration?.dispose();
+    this._terminal?.clearSelection();
+    this._searchResults.clear();
+    this._resultDecorations.forEach(decorations => decorations.forEach(d=> d.dispose()));
+    this._resultDecorations.clear();
```
Tyriar, line 89: This is done in 2 places, let's move it into a helper private function and also switch to using for const of loops

Tyriar (changes requested):

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -3,13 +3,19 @@
...
  */
 
-import { Terminal, IBufferLine, IDisposable, ITerminalAddon, ISelectionPosition } from 'xterm';
+import { Terminal, IDisposable, ITerminalAddon, ISelectionPosition, IDecoration } from 'xterm';
 
 export interface ISearchOptions {
   regex?: boolean;
   wholeWord?: boolean;
   caseSensitive?: boolean;
   incremental?: boolean;
```
Tyriar, line 12: Incremental isn't working correctly when highlighting all matches. Notice how the selection changes on every incremental search.

Recording 2022-03-21 at 11 40 13

Unless enter is pressed the active search term should remain the same but just expand to the right

addons/xterm-addon-search/src/SearchAddon.ts
meganrogge: thought about that, but we already have an `IDecorationOptions`... will think of a different name so it's not confusing and is more generic though

meganrogge: went with `ISearchDecorationOptions`

addons/xterm-addon-search/typings/xterm-addon-search.d.ts
meganrogge: yep good catch, unused

addons/xterm-addon-search/src/SearchAddon.ts
meganrogge: tried to do that and it wouldn't let me because it's a map

meganrogge: made a helper though

css/xterm.css
```diff
@@ -179,6 +179,10 @@
 	position: absolute;
 }
 
+.xterm-find-result-decoration {
+   opacity: 0.6;
+}
+
```
Tyriar, line 185: The addons should be self contained, we should move this and do `element.opacity = '0.6'`

Tyriar (changes requested):

css/xterm.css
Tyriar: Also should we assume the color that's used as the background includes the alpha channel instead?

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -463,4 +612,44 @@ export class SearchAddon implements ITerminalAddon {
...
+
+  /**
+   * Creates a decoration for the result and applies styles
+   * @param result the search result for which to create the decoration
+   * @param color the color to use for the decoration
+   * @returns the {@link IDecoration} or undefined if the marker has already been disposed of
+   */
+  private _createResultDecoration(result: ISearchResult, decorations: ISearchDecorationOptions): IDecoration | undefined {
+    const terminal = this._terminal!;
+    const marker = terminal.registerMarker(-terminal.buffer.active.baseY - terminal.buffer.active.cursorY + result.row);
```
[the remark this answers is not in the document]
meganrogge: hmm, but the decoration gets disposed of here anytime `highlightAllMatches` gets called

https://github.com/meganrogge/xterm.js/blob/7350d1b4a32edf4d4c0ec90cc22e8931457c5233/addons/xterm-addon-search/src/SearchAddon.ts#L93-L97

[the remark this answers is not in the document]
meganrogge: oh I see yeah the marker not the decoration, yeah that seems like a problem

css/xterm.css
Tyriar: I'll move forward and bring this up with you tomorrow.

Tyriar (approved):

PR #3702, opened 2022-03-22
refresh overviewRuler canvas dimensions on refresh of rows

meganrogge:
fixes https://github.com/microsoft/vscode/issues/145214

### DIFF
--- a/src/browser/Decorations/OverviewRulerRenderer.ts
+++ b/src/browser/Decorations/OverviewRulerRenderer.ts
@@ -38,9 +38,10 @@ export class OverviewRulerRenderer extends Disposable {
   }
   private _animationFrame: number | undefined;
 
-  private _canvasHeight: number | undefined;
-  private _canvasWidth: number | undefined;
   private _shouldUpdateDimensions: boolean | undefined = true;
+  private _shouldUpdateAnchor: boolean | undefined = true;
+
+  private _containerHeight: number | undefined;
 
   constructor(
     private readonly _viewportElement: HTMLElement,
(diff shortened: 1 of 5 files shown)


src/browser/Terminal.ts
```diff
@@ -906,6 +909,7 @@ export class Terminal extends CoreTerminal implements ITerminal {
    */
   public refresh(start: number, end: number): void {
     this._renderService?.refreshRows(start, end);
+    this._overviewRulerRenderer?.refreshCanvasDimensions();
```
Tyriar, line 912: I think this is equivalent to listening to `onRenderedBufferChange` inside `OverviewRulerRenderer`?

src/browser/Decorations/OverviewRulerRenderer.ts
```diff
@@ -129,7 +129,7 @@ export class OverviewRulerRenderer extends Disposable {
     );
   }
 
-  private _refreshCanvasDimensions(): void {
+  public refreshCanvasDimensions(): void {
```
Tyriar, line 132: This is too expensive to do on any render, unless we track the canvas width, height and window.devicePixelRatio and make sure they've changed before running the rest of the function

Tyriar (changes requested):

src/browser/Terminal.ts
meganrogge: that's what I tried first and it didn't fix the issue

Tyriar: Oh I see, it's only firing that when it actually changed, not redraws:

https://github.com/xtermjs/xterm.js/blob/fea1079b1ec91b2406ea73a3cbda0e940436bf73/src/browser/services/RenderService.ts#L123-L126

How about exposing a new event on `RenderService` so there is both `onRender: IEvent<{ start: number, end: number }>` and `onRenderedBufferChange: IEvent<{ start: number, end: number }>`?

src/browser/Decorations/OverviewRulerRenderer.ts
meganrogge: don't think we need to track window.devicePixelRatio because that's calling queueRefresh directly here https://github.com/meganrogge/xterm.js/blob/21e1c5ef584737a5a27d517fbd41a5f1f2df270a/src/browser/Decorations/OverviewRulerRenderer.ts#L71

meganrogge: we also don't need to track width because that's just a product of device pixel ratio and options.overviewRulerWidth

meganrogge: hmm 🤔  although maybe we should always check each of these before updating

Tyriar: We'd need to track everything that `refreshCanvasDimensions` and `_refreshDrawConstants` uses to derive their numbers, I think that's just canvas width/height and dpr

src/browser/Decorations/OverviewRulerRenderer.ts
```diff
@@ -165,10 +186,11 @@ export class OverviewRulerRenderer extends Disposable {
 
   private _queueRefresh(updateCanvasDimensions?: boolean, updateAnchor?: boolean): void {
     if (this._animationFrame !== undefined) {
+      this._shouldUpdateDimensions = updateCanvasDimensions || this._shouldUpdateDimensions;
```
meganrogge, line 189: prevents losing the `true` that gets passed in here.

Tyriar, line 189: This this needs to be done regardless of whether the animation frame exists yet or not:

```suggestion
  private _queueRefresh(updateCanvasDimensions?: boolean, updateAnchor?: boolean): void {
    this._shouldUpdateDimensions ||= updateCanvasDimensions;
    if (this._animationFrame !== undefined) {
```

src/browser/Decorations/OverviewRulerRenderer.ts
```diff
@@ -122,24 +131,36 @@ export class OverviewRulerRenderer extends Disposable {
...
   }
 
   private _refreshCanvasDimensions(): void {
-    this._canvas.style.width = `${this._width}px`;
-    this._canvas.style.height = `${this._screenElement.clientHeight}px`;
-    this._canvas.width = Math.round(this._width * window.devicePixelRatio);
-    this._canvas.height = Math.round(this._screenElement.clientHeight * window.devicePixelRatio);
-    this._refreshDrawConstants();
+    let updated = false;
+    if (this._canvasWidth !== this._width) {
```
Tyriar, line 143: Pretty sure we also need to track devicePixelRatio? The function could also be made much simpler by just doing everything if any one of the values changed, I don't think it's worth splitting up width and height as it adds extra complexity without much performance benefit.

```ts
if (width === _width && height === _height && dpr === _dpr) {
  return;
}
// Do everything
```

Tyriar (changes requested):

src/browser/Decorations/OverviewRulerRenderer.ts
meganrogge: this is already covered, but the code is now a little confusing. going to better organize it so it's very clear what's going on

src/browser/Decorations/OverviewRulerRenderer.ts
meganrogge: we don't need to track that because we listen to this event and update when that happens

https://github.com/meganrogge/xterm.js/blob/550b19f334f672e5367383ce0784ce4c5e4e8b15/src/browser/Decorations/OverviewRulerRenderer.ts#L106-L109

src/browser/Decorations/OverviewRulerRenderer.ts
```diff
@@ -57,22 +62,53 @@ export class OverviewRulerRenderer extends Disposable {
...
-    this.register(this._decorationService.onDecorationRegistered(() => this._queueRefresh(undefined, true)));
-    this.register(this._decorationService.onDecorationRemoved(decoration => this._removeDecoration(decoration)));
+  }
+  /**
+   * On dimension change, update canvas dimensions
+   * and then redraw
+   */
+  private _registerDimensionChangeListeners(): void {
+    // container height changed
+    this.register(this._renderService.onRender((): void => {
```
Tyriar, line 94:
```suggestion
    this.register(this._renderService.onRender(() => {
```

src/browser/Decorations/OverviewRulerRenderer.ts
```diff
@@ -131,44 +167,48 @@ export class OverviewRulerRenderer extends Disposable {
...
       this._decorationElements.set(decoration, this._canvas);
       decoration.onDispose(() => this._queueRefresh());
     }
-    this._refreshStyle(decoration, updateAnchor);
+    this._refreshStyle(decoration);
   }
 
   private _queueRefresh(updateCanvasDimensions?: boolean, updateAnchor?: boolean): void {
+    this._shouldUpdateDimensions = updateCanvasDimensions || this._shouldUpdateDimensions;
+    this._shouldUpdateAnchor = updateAnchor || this._shouldUpdateAnchor;
```
Tyriar, line 206: nit: ||= shorthand is more concise

Tyriar (approved):

PR #3704, opened 2022-03-23
add ST sequence, fix emacs

### LINKED ISSUE #3700: Emacs xterm specific initialization does not complete properly
The emacs cannot define xterm configuration of the background color.
I've debugged the issue down to the point when [emacs queries the xterm background color](https://github.com/emacs-mirror/emacs/blob/emacs-27/lisp/term/xterm.el#L737-L738). 

Unfortunately, I cannot provide more exact reasons and errors due to a lack of emacs debugging experience.

## Details
- Browser and browser version: Electron 17
- OS version: macOS 12.3 
- xterm.js version: 4.17.0 and 4.18. Possibly, 4.16.0 affected, too.
- emacs: 26 or 27. Most probably, older versions are affected, too

### Steps to reproduce

1. Run the `emacs` command
2. Observe the current window
3. `*scratch*` is opened. Expected that the `*GNU Emacs*` window is opened.

meganrogge:
fixes #3700 

https://user-images.githubusercontent.com/29464607/159740259-35468672-fdc2-4c47-8757-76c0f6428e17.mov

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -26,7 +26,7 @@ import { IRenderer } from 'browser/renderer/Types';
 import { CompositionHelper } from 'browser/input/CompositionHelper';
 import { Viewport } from 'browser/Viewport';
 import { rightClickHandler, moveTextAreaUnderMouseCursor, handlePasteEvent, copyHandler, paste } from 'browser/Clipboard';
-import { C0 } from 'common/data/EscapeSequences';
+import { C0, C1 } from 'common/data/EscapeSequences';
 import { WindowsOptionsReportType } from '../common/InputHandler';
 import { Renderer } from 'browser/renderer/Renderer';
 import { Linkifier } from 'browser/Linkifier';
@@ -224,7 +224,7 @@ export class Terminal extends CoreTerminal implements ITerminal {
             const channels = color.toColorRGB(acc === 'ansi'
               ? this._colorManager.colors.ansi[req.index]
               : this._colorManager.colors[acc]);
-            this.coreService.triggerDataEvent(`${C0.ESC}]${ident};${toRgbString(channels)}${C0.BEL}`);
+            this.coreService.triggerDataEvent(`${C0.ESC}]${ident};${toRgbString(channels)}${C1.ST}`);
             break;
           case ColorRequestType.SET:
             if (acc === 'ansi') this._colorManager.colors.ansi[req.index] = rgba.toColor(...req.color);
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -26,7 +26,7 @@ import { IRenderer } from 'browser/renderer/Types';
 import { CompositionHelper } from 'browser/input/CompositionHelper';
 import { Viewport } from 'browser/Viewport';
 import { rightClickHandler, moveTextAreaUnderMouseCursor, handlePasteEvent, copyHandler, paste } from 'browser/Clipboard';
-import { C0, C1 } from 'common/data/EscapeSequences';
+import { C0, C1_ESCAPED } from 'common/data/EscapeSequences';
 import { WindowsOptionsReportType } from '../common/InputHandler';
 import { Renderer } from 'browser/renderer/Renderer';
 import { Linkifier } from 'browser/Linkifier';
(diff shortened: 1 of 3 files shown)


src/browser/Terminal.ts
```diff
@@ -224,7 +224,7 @@ export class Terminal extends CoreTerminal implements ITerminal {
             const channels = color.toColorRGB(acc === 'ansi'
               ? this._colorManager.colors.ansi[req.index]
               : this._colorManager.colors[acc]);
-            this.coreService.triggerDataEvent(`${C0.ESC}]${ident};${toRgbString(channels)}${C0.BEL}`);
+            this.coreService.triggerDataEvent(`${C0.ESC}]${ident};${toRgbString(channels)}ESC \9c`);
```
Tyriar, line 227: Think this is right:

```suggestion
            this.coreService.triggerDataEvent(`${C0.ESC}]${ident};${toRgbString(channels)}${C1.ST}`);
```

Tyriar (changes requested):

src/browser/Terminal.ts
Tyriar: Hmm, maybe I don't know how to use ESC notation 🤔

jerch: Ah with `\` I meant the backslash ASCII char, which is at position 0x5c (or `\x5c` as hex-char). So ST would become `{C0.ESC}\x5c`, or `{C0.ESC}\\` (javascript needs double \\ to print one backslash). Maybe it is a good idea to create another convenient mapping for the escaped C1 chars in `common/data/EscapeSequences`, like this:
```typescript
const C1_escaped = {
  ST: `{C0.ESC}\\`,
  // add wotever other C1 is needed in escaped form
}
```
Then you can use it directly as `{C1_escaped.ST}` and so on.

The escaped form of C1 chars of interest can be formalized as: 
```typescript
function escape_char(c) {
  return '\x1b' + String.fromCharCode(c.charCodeAt(0) - 0x40)
}
```

Tyriar (approved):

jerch (approved):
LGTM :+1:  (tests still need to be adjusted)

test/api/InputHandler.api.ts
```diff
@@ -399,36 +399,36 @@ describe('InputHandler Integration Tests', function(): void {
       });
       it('query single color', async () => {
         await writeSync(page, '\x1b]4;0;?\x07');
-        assert.deepEqual(await page.evaluate('window._recordedData'), ['\x1b]4;0;rgb:2e2e/3434/3636\x07']);
+        assert.deepEqual(await page.evaluate('window._recordedData'), ['\x1b]4;0;rgb:2e2e/3434/3636{C0.ESC}\\']);
```
jerch, line 402: @meganrogge Those strings are no template strings, thus the {...} parts will not get eval'ed. Either make them template strings + importing `C0`, or use the hex-string notation as '\x1b\\'.

meganrogge: thanks, just realized that. What's surprising to me is that some of the tests are passing despite that 🤔

meganrogge: Oh I see that's because window.recordedData is a string

Screen Shot 2022-03-23 at 3 45 22 PM

jerch: Hmm thats weird and should not happen. Idk - maybe "printf debugging" is your friend?

jerch: Woops - isnt there a `$` sign missing in `C1_ESCAPED.ST`?

meganrogge: that was a good thought, but using the name space didn't help either 🤔 .

meganrogge: but that probably won't fix the issue that you mentioned for other encodings.

meganrogge: If using C0.ESC isn't an issue here, shouldn't this just use that same thing for the terminating escape?

https://github.com/meganrogge/xterm.js/blob/3457fb48726cea910a3c212bd4943ad360fe3f2e/src/browser/Terminal.ts#L227

jerch: > Maybe I just need to change the expected to match this? Screen Shot 2022-03-23 at 4 49 05 PM

Yeah, maybe it works with that namespace stuff instead. But the template string should contain `${C0.ESC}\\` (note the double backslashes and the $ sign, which was missing above).

meganrogge: it gets interpreted as a literal
🤷🏼‍♀️ 
Screen Shot 2022-03-23 at 5 16 17 PM
Screen Shot 2022-03-23 at 5 17 59 PM

jerch: Almost - there is one superfluous $ char :sweat_smile:

meganrogge: 😅  I took this literally

> But the template string should contain ${C0.ESC}\\

Works now 👍🏼  thanks so much for the help

PR #3712, opened 2022-03-25
PoC for protected flag

jerch:
Currently just a quickly hacked PoC for protected flag support (so far only tested manually on canvas renderer).

@stuboydl: Maybe try this one out?

### DIFF
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -99,6 +99,9 @@ export enum WindowsOptionsReportType {
   GET_CELL_SIZE_PIXELS = 1
 }
 
+// FIXME: add to terminal settings somewhere
+let is_protected = false;
+
 // create a warning log if an async handler takes longer than the limit (in ms)
 const SLOW_ASYNC_LIMIT = 5000;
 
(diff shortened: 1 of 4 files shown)


Tyriar (commented):
I know this is just a proof of concept, do you think it's worth the extra complexity to support this one? If we can support it so the perf overhead is negligible on print when no protected flags happen then I'm not opposed.

jerch:
Tbh - I dont know, but lean somewhat more towards no support. Indeed the overhead is not that much than I thought in the beginning, also the code duplication of the PoC can be shortened alot by using some handler arguments.

But I am not sure yet whether the PoC already covers all aspects of the protected flag thingy, thus should be tested first by someone with an actual use case. Without any real world test case there is a high chance we get it partially wrong, or introduce more ugly reworks down the road. Thus my hesitation.

jerch:
@Tyriar I think I will give this another go, once the other PR #4095  is done. The needed changes seem so small so I dont see a reason not to support it. Still I dont quite grasp the full domain of those protection flags, so this needs some digging through old DEC manuals.

Tyriar:
@jerch sounds good, it's just one bit and we've got extended attributes working pretty well now

jerch:
Superseeded by #4098.

PR #3713, opened 2022-03-25
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.


CommanderRoot:
The current codebase has more `slice()` usage than `substring()`. Also generally `slice()` is a bit faster than `substring()` and about the same as `substr()`. Slice is also bit shorter which means it the code is smaller.
If you still want to rather use `substring()` I can change it hower, your repo, your rules.

PR #3716, opened 2022-03-28
Add search result count to SearchAddon

### LINKED ISSUE #1659: Search addon: Return the number of results of a search
This request may impact performance, this should be evaluated.

Plan issue: #705

meganrogge:
fixes #1659

### DIFF
--- a/addons/xterm-addon-search/src/SearchAddon.ts
+++ b/addons/xterm-addon-search/src/SearchAddon.ts
@@ -280,15 +280,13 @@ export class SearchAddon implements ITerminalAddon {
       result = this._findInLine(term, searchPosition, searchOptions);
     }
 
-    if (!currentSelection || currentSelection?.startColumn !== result?.col || currentSelection?.startRow !== result?.row) {
-      if (this._searchResults) {
-        if (!this._resultIndex) {
+    if (this._searchResults) {
+      if (!this._resultIndex) {
+        this._resultIndex = 1;
+      } else {
+        this._resultIndex++;
+        if (this._resultIndex > this._searchResults.size) {
           this._resultIndex = 1;
-        } else {
-          this._resultIndex++;
-          if (this._resultIndex > this._searchResults.size) {
-            this._resultIndex = 1;
-          }
         }
       }
     }
@@ -391,15 +389,13 @@ export class SearchAddon implements ITerminalAddon {
       }
     }
 
-    if (!currentSelection || currentSelection?.startColumn !== result?.col || currentSelection?.startRow !== result?.row) {
-      if (this._searchResults) {
-        if (!this._resultIndex) {
+    if (this._searchResults) {
+      if (!this._resultIndex) {
+        this._resultIndex = this._searchResults?.size;
+      } else {
+        this._resultIndex--;
+        if (this._resultIndex === 0) {
           this._resultIndex = this._searchResults?.size;
-        } else {
-          this._resultIndex--;
-          if (this._resultIndex === 0) {
-            this._resultIndex = this._searchResults?.size;
-          }
         }
       }
     }
--- a/addons/xterm-addon-search/src/SearchAddon.ts
+++ b/addons/xterm-addon-search/src/SearchAddon.ts
@@ -131,6 +131,8 @@ export class SearchAddon implements ITerminalAddon {
   private _cursorMoveListener: IDisposable | undefined;
   private _resizeListener: IDisposable | undefined;
 
+  private _resultIndex: number | undefined;
+
   private readonly _onDidChangeResults = new EventEmitter<void>();
   public readonly onDidChangeResults = this._onDidChangeResults.event;
 
@@ -176,30 +178,26 @@ export class SearchAddon implements ITerminalAddon {
   }
 
   /**
-   * @returns the last search result count or undefined
-   * if decorations aren't enabled
-   */
-  public get resultCount(): number | undefined {
-    return this._searchResults?.size || undefined;
-  }
-
-  /**
    * Find the next instance of the term, then scroll to and select it. If it
    * doesn't exist, do nothing.
    * @param term The search term.
    * @param searchOptions Search options.
    * @return Whether a result was found.
    */
-  public findNext(term: string, searchOptions?: ISearchOptions): boolean {
+  public findNext(term: string, searchOptions?: ISearchOptions): boolean | { resultIndex: number, resultCount: number } {
     if (!this._terminal) {
       throw new Error('Cannot use addon until it has been loaded');
     }
     this._lastSearchOptions = searchOptions;
-    const findNextResult = this._findNextAndSelect(term, searchOptions);
     if (searchOptions?.decorations) {
       this._highlightAllMatches(term, searchOptions);
     }
-    return findNextResult;
+    // if (term !== this._cachedSearchTerm) {
+    //   this._resultIndex = undefined;
+    //   this.clearDecorations();
+    // }
+    const findNextResult = this._findNextAndSelect(term, searchOptions);
+    return this._resultIndex ? { resultIndex: this._resultIndex!, resultCount: this._searchResults?.size || 0 } : findNextResult;
   }
 
   private _highlightAllMatches(term: string, searchOptions: ISearchOptions): void {
@@ -298,6 +296,16 @@ export class SearchAddon implements ITerminalAddon {
       this.clearDecorations();
       return false;
     }
+    if (this._searchResults) {
+      if (!this._resultIndex) {
+        this._resultIndex = 1;
+      } else {
+        this._resultIndex++;
+        if (this._resultIndex > this._searchResults.size) {
+          this._resultIndex = 1;
+        }
+      }
+    }
 
     let startCol = 0;
     let startRow = 0;
@@ -352,7 +360,6 @@ export class SearchAddon implements ITerminalAddon {
       searchPosition.startCol = 0;
       result = this._findInLine(term, searchPosition, searchOptions);
     }
-
     // Set selection and scroll if a result was found
     return this._selectResult(result, searchOptions?.decorations);
   }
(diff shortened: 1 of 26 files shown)


meganrogge:
We can either do it this way or could return the count from findNext/findPrevious instead of a boolean

Tyriar (commented):
Do we only know the total count if decorations are on? What would this return if it's off so only a single result is selected?

meganrogge:
Yes we only know this if decorations are on. I think that's what we want because otherwise don't want to do all of this processing. When decorations are on, all of that work has already been done, so might as well leverage it.

meganrogge:
We discussed and agreed upon:
- using a property instead of a function  (✔️ )
- only enabling this when decorations are provided as an option (✔️ )
- returning from findNext/findPrevious instead of a separate API (decided against to keep things clear)

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -4,6 +4,69 @@
  */
 
 import { Terminal, IDisposable, ITerminalAddon, ISelectionPosition, IDecoration } from 'xterm';
+interface IListener<T, U = void> {
```
Tyriar, line 7: You can see the webgl addon for how to pull in parts for xterm.js core:

https://github.com/Tyriar/xterm.js/blob/eca36458b394ce980957518f23d53c10af7e13ff/addons/xterm-addon-webgl/src/tsconfig.json#L31-L38

Just need to be very careful only to include low level platform things as all code pulled in will get duplicated in the search addon's js bundle.

Tyriar (changes requested):

addons/xterm-addon-search/src/SearchAddon.ts
meganrogge: not sure if I did that right - bunch of errors

addons/xterm-addon-search/tsconfig.json
```diff
@@ -3,6 +3,6 @@
   "include": [],
   "references": [
     { "path": "./src" },
-    { "path": "./test" }
+    { "path": "./test" },
```
meganrogge, line 6:
```suggestion
    { "path": "./test" }
```

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -90,14 +96,18 @@ export class SearchAddon implements ITerminalAddon {
 
   public clearDecorations(): void {
     this._selectedDecoration?.dispose();
-    this._terminal?.clearSelection();
-    this._searchResults.clear();
+    // this._terminal?.clearSelection();
```
meganrogge, line 99:
```suggestion
```

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -164,6 +181,9 @@ export class SearchAddon implements ITerminalAddon {
     if (this._searchResults.size > 0) {
       this._cachedSearchTerm = term;
     }
+    if (dataDrivenChange && lastSearchCount !== this._searchResults.size) {
+      this._onDidChangeResults.fire();
```
meganrogge, line 185:
```suggestion
      // When data is added to the buffer that yields new search results, fire this so the match result count can be updated
      this._onDidChangeResults.fire();
```

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -210,12 +233,22 @@ export class SearchAddon implements ITerminalAddon {
...
+        }
+      }
+    }
 
     let startCol = 0;
     let startRow = 0;
     let currentSelection: ISelectionPosition | undefined;
     if (this._terminal.hasSelection()) {
-      const incremental = searchOptions ? searchOptions.incremental : false;
+      const incremental = searchOptions && !(searchOptions.decorations && this._previousWasLastCall) ? searchOptions.incremental : false;
```
meganrogge, line 251: This fixes a problem sometimes where the selection is wrong when switching between find previous/next. I''m not sure why that's required and will investigate further. 

The result index is also sometimes not quite right when switching between find next / previous

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -76,8 +83,8 @@ export class SearchAddon implements ITerminalAddon {
         window.clearTimeout(this._highlightTimeout);
       }
       this._highlightTimeout = setTimeout(() => {
-        if (this._lastSearchOptions?.decorations && this._cachedSearchTerm && this._resultDecorations.size > 0 && this._lastSearchOptions) {
-          this._highlightAllMatches(this._cachedSearchTerm, this._lastSearchOptions);
+        if (this._lastSearchOptions?.decorations && this._cachedSearchTerm && this._resultDecorations?.size && this._resultDecorations.size > 0 && this._lastSearchOptions) {
```
meganrogge, line 86: move decorations above the timeout

meganrogge: override search options incremental

{ ...this._lastSearchOptions, incremental: true }

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -90,14 +97,17 @@ export class SearchAddon implements ITerminalAddon {
 
   public clearDecorations(): void {
     this._selectedDecoration?.dispose();
-    this._terminal?.clearSelection();
-    this._searchResults.clear();
+    this._searchResults?.clear();
```
meganrogge, line 100: set searchResults to undefined and resultDecorations

meganrogge: after dipsoseDecorations, also clear resultDecorations

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -118,17 +128,29 @@ export class SearchAddon implements ITerminalAddon {
...
     this._lastSearchOptions = searchOptions;
-    const findNextResult = this._findNextAndSelect(term, searchOptions);
     if (searchOptions?.decorations) {
       this._highlightAllMatches(term, searchOptions);
     }
-    return findNextResult;
+    const next = this._findNextAndSelect(term, searchOptions);
+    if (next && this._resultIndex && this._searchResults?.size) {
+      this._onDidChangeResults.fire({ resultIndex: this._resultIndex, resultCount: this._searchResults?.size });
+    } else if (searchOptions?.decorations) {
```
meganrogge, line 137: pull if above other if

addons/xterm-addon-search/src/SearchAddon.ts
meganrogge: check if incremental is true here

addons/xterm-addon-search/src/tsconfig.json
```diff
@@ -13,10 +13,26 @@
...
     "types": [
       "../../../node_modules/@types/mocha"
-    ]
+    ],
+    "paths": {
+      "common/*": [
+        "../../../src/common/*"
+      ],
+      "browser/*": [
+        "../../../src/browser/*"
```
meganrogge, line 22: try to remove this

addons/xterm-addon-search/src/tsconfig.json
```diff
@@ -13,10 +13,26 @@
...
   "include": [
     "./**/*",
     "../../../typings/xterm.d.ts"
+  ],
+  "references": [
+    {
+      "path": "../../../src/common"
+    },
+    {
+      "path": "../../../src/browser"
```
meganrogge, line 35: try to remove browser

addons/xterm-addon-search/typings/xterm-addon-search.d.ts
```diff
@@ -95,6 +95,8 @@ declare module 'xterm-addon-search' {
      * options.
      * @param term The search term.
      * @param searchOptions The options for the search.
+     * @returns the result index and result count if decorations are provided or a boolean
```
meganrogge, line 98: fix

addons/xterm-addon-search/typings/xterm-addon-search.d.ts
```diff
@@ -103,12 +105,20 @@ declare module 'xterm-addon-search' {
...
+     * indicating if there was a match
```
meganrogge, line 109: revert

addons/xterm-addon-search/typings/xterm-addon-search.d.ts
```diff
@@ -103,12 +105,20 @@ declare module 'xterm-addon-search' {
...
     /**
      * Clears the decorations and selection
      */
     public clearDecorations(): void;
+
+    /**
+     * Fired when the search results or the selected result changes.
+     * Provides undefined when there are no results.
+     */
+    onDidChangeResults: IEvent< { resultIndex: number, resultCount: number } | undefined>;
```
meganrogge, line 122: only fires when decorations are enabled
readonly and fix space

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -68,6 +70,11 @@ export class SearchAddon implements ITerminalAddon {
   private _cursorMoveListener: IDisposable | undefined;
   private _resizeListener: IDisposable | undefined;
 
+  private _resultIndex: number | undefined;
+
+  private readonly _onDidChangeResults = new EventEmitter<{resultIndex: number, resultCount: number} | undefined>();
```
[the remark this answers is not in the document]
meganrogge: I don't think this actually needs to be registered because isn't used anywhere other than for testing / logging

addons/xterm-addon-search/src/SearchAddon.ts
meganrogge: and do same for the previous

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -68,19 +70,25 @@ export class SearchAddon implements ITerminalAddon {
...
-          this._highlightAllMatches(this._cachedSearchTerm, this._lastSearchOptions);
-        }
-      }, 200);
+      if (this._cachedSearchTerm && this._lastSearchOptions?.decorations) {
+        this._highlightTimeout = setTimeout(() => {
+          this.findPrevious(this._cachedSearchTerm!,  { ...this._lastSearchOptions, incremental: true });
+        }, 200);
+      }
     });
+    this.onDidChangeResults((e) => console.log(e));
```
meganrogge, line 91: remove

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -87,6 +87,7 @@ export class SearchAddon implements ITerminalAddon {
         }, 200);
       }
     });
+    this.onDidChangeResults((results) => console.log(results));
```
Tyriar: Missed this one

Tyriar (approved):

addons/xterm-addon-search/src/SearchAddon.ts
meganrogge: think that was outdated as I don't see it there

PR #3719, opened 2022-03-30
Expose urlRegex in the public .d.ts file.

LuckyGeck:
This makes it possible to pass a regex when working against a .d.ts file instead of directly against a WebLinksAddon.ts


Tyriar (approved):
Thanks for spotting this. I also removed duplication with the interface since I noticed they existed in 2 forms as well.

PR #3723, opened 2022-04-01
Improve performance of overview ruler by computing and rendering color zones

### LINKED ISSUE #3703: Optimize rendering of decorations in overview ruler
VS Code issue: https://github.com/microsoft/vscode/issues/145751

Could repro this slowness by filling the buffer and then searching for a single letter, this is on a 12th gen i7 (fast modern CPU):

image

A similar thing happens when pressing enter in this state since we don't do any diffing, just clear and re-render.

Ideas to explore:

- Move from rendering each decoration to calculating colored zones which merge similar rectangles into the same zone and render them
- If the buffer is full, explore shifting the texture up depending on how much scrolling occurs (might be tricky to get right)

Tyriar:
Fixes #3703

Perf of the problem case when finding in a full buffer:

image

### DIFF
--- a/demo/client.ts
+++ b/demo/client.ts
@@ -556,8 +556,8 @@ function loadTest() {
 function addDecoration() {
   term.options['overviewRulerWidth'] = 15;
   const marker = term.addMarker(1);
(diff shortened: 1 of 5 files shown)


meganrogge (approved):
Looks great 👍🏻 I can test later

PR #3728, opened 2022-04-06
Fix Caps lock bug when using some IME on macOS

### LINKED ISSUE #2457: Caps lock isn't working correcting when using Pinyin IME on MacOS
Ah I wasn't reproducing the right thing, this appears to be specific behavior of Chinese IMEs. Repro:

1. Enable PinYin IME
2. Type (Chinese auto completions appear)
3. Caps lock on
4. Type - This should be lowercase in the terminal but it's uppercase

VS Code issue: https://github.com/microsoft/vscode/issues/69367

### LINKED ISSUE #3155: Terminal always show uppercase letters when using caps lock to switch input source
## Details
- Browser and browser version: Found in VSCode 1.50.1, I think it's browser unrelated.
- OS version: 10.15.7
- xterm.js version: Found in VSCode 1.50.1, I think it's xterm.js version unrelated.

### Steps to reproduce

1. Set system input source to shuangpin, before which add shuangpin as an input source.
image
2. Lock caps lock, with caps lock unlock, you will input Chinese character, which is an expected behavior.
3. Input something in the terminal, I can only input upper case letters. But when I input something in some other application, I can input lower case letters.

serkodev:
Fixes #2457
Fixes #3155

The reason for this bug is that when Chrome uses some specific input methods and keydown, it will enter a key that is inconsistent with the actual one.

But in `onkeypress`, it can return the correct `keyCode`. Therefore, the solution to this problem is to only process characters (A-Za-z) during `onkeypress`.

In addition to the macOS built-in Pinyin input method proposed in the earlier issue, some other macOS third-party input methods such as [OpenVanilla](https://openvanilla.org/) will encounter the same problem.

---
Fixes https://github.com/microsoft/vscode/issues/109565
Fixes https://github.com/microsoft/vscode/issues/69367
Fixes https://github.com/microsoft/vscode/issues/138028


serkodev:
@Tyriar Please let me know if you have any comments on this as this bug is really annoying when using VSCode.

Tyriar (commented):
Won't this break non-US keyboard layouts that use characters beyond a-zA-Z?

serkodev:
I've tried Taiwan and Japanese layout keyboard and there is no problems.

serkodev:
@Tyriar If there are some other keyboard layouts or something else need me to help to test, just let me know. 😊
And welcome to discuss if there is other way to fix this bug.

This hug has affect me for at least 6 years when I was new to develop web using VSCode.

serkodev:
If concern about `event.key` is non-English char, I think it can also add a condition to confirm the `event.key` is `a-zA-Z` before returning.

```diff
if (event.key && !event.ctrlKey && !event.altKey && !event.metaKey && event.key.length === 1 && (
    (event.keyCode >= 65 && event.keyCode <= 90) ||
    (event.keyCode >= 97 && event.keyCode <= 122)
  )) {
    // Include only keys in A-Z/a-z.
    // HACK: Need process these key events in 'onkeypress' to fix a IME bug on macOS Chrome. fix for #2457
-   return true;
+   let c = event.key.charCodeAt(0)
+   if ((c >= 65 && c <= 90) || (c >= 97 && c <= 122)) {
+       return true;
+   }
}
```

> I did a benchmark test on using `event.key.charCodeAt(0)` and Regex `/[a-z]/i.test(event.key)`, the result is using `charCodeAt` is 91.65% faster then Regex.

If you think this solution is acceptable, I can push a commit with this.

serkodev:
@Tyriar May I know is there feed back? Thank you

src/browser/Terminal.ts
```diff
@@ -1134,6 +1134,15 @@ export class Terminal extends CoreTerminal implements ITerminal {
       return true;
     }
 
+    if (event.key && !event.ctrlKey && !event.altKey && !event.metaKey && event.key.length === 1 && (
+      (event.keyCode >= 65 && event.keyCode <= 90) ||
+      (event.keyCode >= 97 && event.keyCode <= 122)
+    )) {
+      // Include only keys in A-Z/a-z.
+      // HACK: Need process these key events in 'onkeypress' to fix a IME bug on macOS Chrome. fix for #2457
+      return true;
```
serkodev, line 1143: If concern about event.key is non-English char, I think it can also add a condition to confirm the event.key is a-zA-Z before returning.

```suggestion
      const c = event.key.charCodeAt(0)
      if ((c >= 65 && c <= 90) || (c >= 97 && c <= 122)) {
        return true;
      }
```

serkodev:
@Tyriar @meganrogge World you please take a look on this? Hope to give some advice if possible.
This bug is still bothering me a lot.

Tyriar (approved):
@serkodev thanks a bunch for looking into this and sorry about the delay. I pushed a change to move off the deprecated keyCode and it still works great. This should land in the next version of VS Code 🙂

PR #3730, opened 2022-04-08
get overview ruler to update when color is changed

### LINKED ISSUE #3729: overview ruler doesn't update when color is changed
originally reported https://github.com/microsoft/vscode/issues/145821

meganrogge:
fixes #3729


typings/xterm.d.ts
```diff
@@ -444,7 +444,7 @@ declare module 'xterm' {
      * This will only take effect when {@link IDecorationOptions.overviewRulerOptions}
      * were provided initially.
```
meganrogge, line 445: since the properties we don't want to allow updating are readonly, i think this is fine

PR #3740, opened 2022-04-19
exclude powerline char range from contrast demands

### LINKED ISSUE #3739: exclude powerline characters from contrast ratio change
originally reported https://github.com/microsoft/vscode/issues/146406

meganrogge:
fixes #3739

Based the range on this [documentation](https://apw-bash-settings.readthedocs.io/en/latest/fontpatching.html)

### DIFF
--- a/src/browser/renderer/BaseRenderLayer.ts
+++ b/src/browser/renderer/BaseRenderLayer.ts
@@ -430,7 +430,7 @@ export abstract class BaseRenderLayer implements IRenderLayer {
   private _getContrastColor(cell: CellData): IColor | undefined {
     // skip empty cells
     if (cell.content !== 4194304) {
-      if (cell.getAsCharData()[3] >= 57344 && cell.getAsCharData()[3] <=  63743) {
+      if (cell.getCode() >= 57344 && cell.getCode() <=  63743) {
         // powerline chars #3739
         return undefined;
       }
--- a/src/browser/renderer/BaseRenderLayer.ts
+++ b/src/browser/renderer/BaseRenderLayer.ts
@@ -431,7 +431,7 @@ export abstract class BaseRenderLayer implements IRenderLayer {
     // skip empty cells
     if (cell.content !== 4194304) {
       if (cell.getAsCharData()[3] >= 57344 && cell.getAsCharData()[3] <=  63743) {
-        powerline characters https://github.com/xtermjs/xterm.js/issues/3739
+        // powerline chars #3739
         return undefined;
       }
     }
--- a/src/browser/renderer/BaseRenderLayer.ts
+++ b/src/browser/renderer/BaseRenderLayer.ts
@@ -428,12 +428,9 @@ export abstract class BaseRenderLayer implements IRenderLayer {
   }
 
   private _getContrastColor(cell: CellData): IColor | undefined {
-    // skip empty cells
-    if (cell.content !== 4194304) {
-      if (cell.getCode() >= 57344 && cell.getCode() <=  63743) {
-        // powerline chars #3739
-        return undefined;
-      }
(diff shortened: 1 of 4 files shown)


src/browser/renderer/BaseRenderLayer.ts
```diff
@@ -428,6 +428,11 @@ export abstract class BaseRenderLayer implements IRenderLayer {
   }
 
   private _getContrastColor(cell: CellData): IColor | undefined {
+    if (cell.content !== 4194304) {
```
meganrogge, line 431:
```suggestion
    // skip empty cells
    if (cell.content !== 4194304) {
```

src/browser/renderer/BaseRenderLayer.ts
```diff
@@ -428,6 +428,11 @@ export abstract class BaseRenderLayer implements IRenderLayer {
...
+      if (cell.getAsCharData()[3] >= 57344 && cell.getAsCharData()[3] <=  63743) {
+        return undefined;
```
meganrogge, line 433:
```suggestion
        powerline characters https://github.com/xtermjs/xterm.js/issues/3739
        return undefined;
```

meganrogge:
still a wip - need to test more

jerch (commented):
Nonetheless a few early remarks from my side:

I'd suggest not to use `getAsCharData` for several reasons:
- outdated and should be removed (`CharData` was the cell type of the old bufferline impl, those methods only existed for interim transition, but never got cleaned up afterwards because of reasons :blush:) - ah right, too convenient in tests :smile_cat: 
- the method creates an interim array for every tested cell (creating GC pressure for no good reason)

Maybe use `cell.getCode()` directly instead?

A note on powerline glyphs - most should reside in the Unicode's  PUA (private use area), as you correctly test for. But if I remember right there are powerline-like fonts that also use other areas or even single codepoints all over the place (was it nerdfont? idk). Not sure how you want to deal with that, it is a bit like chasing a rabbit, and if you overdo it, you might end up excluding too much ...

meganrogge:
> A note on powerline glyphs - most should reside in the Unicode's PUA (private use area), as you correctly test for. But if I remember right there are powerline-like fonts that also use other areas or even single codepoints all over the place (was it nerdfont? idk). Not sure how you want to deal with that, it is a bit like chasing a rabbit, and if you overdo it, you might end up excluding too much ...

Yes good point and I considered covering additional symbols, but I think that arrows are the main problem case as they look pretty bad ATM.

Screen Shot 2022-04-19 at 1 41 30 PM

src/browser/renderer/BaseRenderLayer.ts
```diff
@@ -428,6 +428,10 @@ export abstract class BaseRenderLayer implements IRenderLayer {
   }
 
   private _getContrastColor(cell: CellData): IColor | undefined {
+    if (cell.getCode() >= 57344 && cell.getCode() <=  63743) {
```
jerch, line 431:
```suggestion
      const codepoint = cell.getCode();
      if (57344 <= codepoint && codepoint <=  63743) {
```
avoiding a second function call and maybe more readable? (I know thats a tiny micro optimization here, but it will evtl. add up for all cells, so catching the value locally might be a good idea.)

meganrogge: yes wasn't sure if the memory tradeoff would be worth it or not

jerch: Oh idk either. If the JIT kicks in here, then it is prolly a nonsense optimization (would do its own stack code reorganization thing anyway). For semi-hot code (that gets not or very late optimized) I found local vars often perform better. Under chrome it is further important to keep things strictly local, so the minor GC can wipe things fast. As soon as things leave the local scope, the GC cleanup gets much heavier and more scarce (major GC).

PR #3742, opened 2022-04-20
exclude powerline chars from min contrast for webgl and dom 

meganrogge:
related https://github.com/xtermjs/xterm.js/pull/3740/files

### DIFF
--- a/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
+++ b/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts
@@ -216,8 +216,8 @@ export class WebglCharAtlas implements IDisposable {
     }
   }
 
-  private _getForegroundCss(bg: number, bgColorMode: number, bgColor: number, fg: number, fgColorMode: number, fgColor: number, inverse: boolean, bold: boolean): string {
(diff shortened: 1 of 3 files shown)


jerch:
@meganrogge Just a minor headsup - the webgl code considers a smaller codepoint range to be powerline glyphs: https://github.com/meganrogge/xterm.js/blob/341ca24dda672475f683eaf8f0ff6318e3143513/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts#L379

Is this intended? (If not, maybe the check should be coded in one central place to avoid incongruence?)

meganrogge:
Yes I was wondering about that, but not sure. I might wait until @Tyriar is back to make further changes. Sound good?

meganrogge:
Actually, should probably just be consistent. Will do

jerch:
> Yes I was wondering about that, but not sure. I might wait until @Tyriar is back to make further changes. Sound good?

> Actually, should probably just be consistent. Will do

Well I dont mind, whatever suits your plans. My first reflex would be consistency over code repetitions, but I dont know if the webgl code has the smaller range for a good but non-obvious reason. Maybe go the consistency route in the first place, and introduce a special case in webgl code later, if @Tyriar had a specific reason for the smaller range?

PR #3743, opened 2022-04-20
align isPowerlineGlyph across renderers

meganrogge:


jerch (approved):
Yepp, coolio :+1:

This separation has another advantage - if the wider nerdfont codepoints create issues in the future, they simply can be added there a well.

jerch:
@Tyriar Oh well, then - ah already fixed. :smiley_cat:

PR #3750, opened 2022-04-27
clear `cachedSearchTerm` when decorations are cleared

### LINKED ISSUE #3749: highlight all matches triggered by onData even after clearDecorations is called
Reported here https://github.com/microsoft/vscode/issues/148295)

meganrogge:
fix #3749


addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -102,7 +102,7 @@ export class SearchAddon implements ITerminalAddon {
     this._onDataDisposable?.dispose();
   }
 
-  public clearDecorations(): void {
+  public clearDecorations(internal?: boolean): void {
```
meganrogge, line 105: should be more descriptive like `clearCachedSearchTerm`

PR #3756, opened 2022-05-02
use correct powerline glyph range

meganrogge:
follow up from #3743

### DIFF
--- a/src/browser/renderer/RendererUtils.ts
+++ b/src/browser/renderer/RendererUtils.ts
@@ -11,7 +11,5 @@ export function throwIfFalsy<T>(value: T | undefined | null): T {
 }
 
 export function isPowerlineGlyph(codepoint: number): boolean {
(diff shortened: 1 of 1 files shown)


src/browser/renderer/RendererUtils.ts
```diff
@@ -11,7 +11,5 @@ export function throwIfFalsy<T>(value: T | undefined | null): T {
 }
 
 export function isPowerlineGlyph(codepoint: number): boolean {
-  // This range was established via
-  // https://apw-bash-settings.readthedocs.io/en/latest/fontpatching.html
-  return 0xE000 <= codepoint && codepoint <=  0xF8FF;
+  return 0xE0A0 <= codepoint && codepoint <=  0xE0D6;
```
Tyriar, line 14: A comment describing the decision would be good

Tyriar (approved):

PR #3757, opened 2022-05-02
refine name for clearDecorations param

meganrogge:


addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -99,7 +99,7 @@ export class SearchAddon implements ITerminalAddon {
     this._onResizeDisposable?.dispose();
   }
 
-  public clearDecorations(internal?: boolean): void {
+  public clearDecorations(retainCachedSearchTerm?: boolean): void {
```
Tyriar, line 102: The inverse would be better imo, like `clearCachedSearchTerm`. Nicer to assume that the default of undefined acts the same as false

Tyriar (approved):

addons/xterm-addon-search/src/SearchAddon.ts
meganrogge: but then we'd have to modify the API

PR #3765, opened 2022-05-09
addon-web-links: strip trailing closing parenthesis

r10r:
This fixes link selection in markdown documents.

I've tested the fix in the demo server with the following markdown document:

```markdown
[google](http://google.com)
```
Without the fix link navigation won't work because the trailing parenthesis is included in the selection.


Tyriar (commented):
This would unfortunately break other links that validly include/end in a `)` character. It's one of the class of issues that would need more robust link detection to handle which can be achieved via the link provider API.

PR #3779, opened 2022-05-11
Layout dimension left/right after renderer dims change

### LINKED ISSUE #3773: Going from webgl renderer to dom renderer will not reposition decorations
While searching for "se", the background moves as it's my WIP for https://github.com/xtermjs/xterm.js/issues/3770 but the border remains in the old renderer's position.

Repro:

1. Open demo
2. Enable highlight all matches
3. Set DOM renderer
4. Enable webgl renderer
5. Find "se" or something else
6. Disable webgl renderer

Tyriar:
Fixes #3773


meganrogge (changes requested):
shouldn't `_dimensionsChanged` get reset to false somewhere?

Tyriar:
@meganrogge I knew there was a reason I requested a review 😉

PR #3785, opened 2022-05-12
Support setting the layer of a decoration

### LINKED ISSUE #3778: bg/fg decorations should be able to render on top of the selection
With https://github.com/xtermjs/xterm.js/pull/3775 decorations can override the buffer fg/bg, when using in the search addon with the webgl renderer this leads to weird results though since the selection is drawn on top of the decoration.

When using find:

image

If you click the terminal (removing the selection):

image

This is mainly an issue if the selection is opaque

Tyriar:
Fixes #3778

### DIFF
--- a/src/browser/renderer/BaseRenderLayer.ts
+++ b/src/browser/renderer/BaseRenderLayer.ts
@@ -443,13 +443,18 @@ export abstract class BaseRenderLayer implements IRenderLayer {
     // exist but applied after inverse
     let bgOverride: number | undefined;
     let fgOverride: number | undefined;
+    let isTop = false;
     for (const d of this._decorationService.getDecorationsAtCell(x, y)) {
+      if (d.options.layer !== 'top' && isTop) {
+        continue;
+      }
       if (d.backgroundColorRGB) {
         bgOverride = d.backgroundColorRGB.rgba;
       }
       if (d.foregroundColorRGB) {
         fgOverride = d.foregroundColorRGB.rgba;
       }
+      isTop = d.options.layer === 'top';
     }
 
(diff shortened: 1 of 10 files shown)


src/browser/renderer/BaseRenderLayer.ts
```diff
@@ -443,13 +443,18 @@ export abstract class BaseRenderLayer implements IRenderLayer {
...
+      if (d.options.layer !== 'top' && isTop) {
+        continue;
+      }
       if (d.backgroundColorRGB) {
         bgOverride = d.backgroundColorRGB.rgba;
       }
       if (d.foregroundColorRGB) {
         fgOverride = d.foregroundColorRGB.rgba;
       }
+      isTop = d.options.layer === 'top';
```
meganrogge, line 457: is `isTop` here tracking whether the prior cell had a top layer?

if so, could this be named `priorCellHadTop` or something?

also how/why does the prior cell's option impact the current one?

src/browser/renderer/TextRenderLayer.ts
```diff
@@ -179,10 +179,15 @@ export class TextRenderLayer extends BaseRenderLayer {
 
       // Get any decoration foreground/background overrides, this must be fetched before the early
       // exist but applied after inverse
+      let isTop = false;
       for (const d of this._decorationService.getDecorationsAtCell(x, this._bufferService.buffer.ydisp + y)) {
+        if (d.options.layer !== 'top' && isTop) {
+          continue;
+        }
```
meganrogge, line 186: same question as above

src/browser/renderer/BaseRenderLayer.ts
Tyriar: It's for the same cell, when `isTop` is set it starts ignoring the bottom layer decorations to get the correct order.

meganrogge (approved):

PR #3786, opened 2022-05-12
Highlight all matches in demo by default

Tyriar:


meganrogge (commented):
would be nice if the overview ruler decorations showed up immediately, but they don't because we don't create that unless it's explicitly created
https://user-images.githubusercontent.com/29464607/168161395-fd939977-b6e2-4e34-a5fb-5dcaf85c2b78.mov

meganrogge (approved):

PR #3790, opened 2022-05-12
Fix case of Decorations folder

Tyriar:


Tyriar:
Changes happened since

Tyriar:
https://github.com/xtermjs/xterm.js/pull/3791

PR #3793, opened 2022-05-12
reset search result index when buffer changes/ terminal is resized

### LINKED ISSUE #3724: Search result numbering seems illogical
It seems that the "first" search result as reported by `onDidChangeResults` is arbitrary, e.g. here the result #0 is the 3rd one from the end: 
image

meganrogge:
fix #3724

### DIFF
--- a/addons/xterm-addon-search/src/SearchAddon.ts
+++ b/addons/xterm-addon-search/src/SearchAddon.ts
@@ -79,6 +79,7 @@ export class SearchAddon implements ITerminalAddon {
     this._terminal = terminal;
     this._onDataDisposable = this._terminal.onData(() => this._updateMatches());
     this._onResizeDisposable = this._terminal.onResize(() => this._updateMatches());
+    this.onDidChangeResults((results) => console.log(results));
   }
 
   private _updateMatches(): void {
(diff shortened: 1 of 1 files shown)


addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -79,6 +79,7 @@ export class SearchAddon implements ITerminalAddon {
     this._terminal = terminal;
     this._onDataDisposable = this._terminal.onData(() => this._updateMatches());
     this._onResizeDisposable = this._terminal.onResize(() => this._updateMatches());
+    this.onDidChangeResults((results) => console.log(results));
```
meganrogge, line 82:
```suggestion
```

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -88,6 +88,7 @@ export class SearchAddon implements ITerminalAddon {
     if (this._cachedSearchTerm && this._lastSearchOptions?.decorations) {
       this._highlightTimeout = setTimeout(() => {
         this.findPrevious(this._cachedSearchTerm!,  { ...this._lastSearchOptions, incremental: true, noScroll: true });
+        this._resultIndex = this._searchResults ? this._searchResults.size -1 : -1;
```
Tyriar, line 91: So this happens after `onData` or `onResize` to re-highlight what's already searched? Couldn't the active result after that be in the middle of the buffer somewhere?

meganrogge: we always run `findPrevious` intentionally here so things don't get complicated with the index - it's always the last one in the buffer

Tyriar (approved):

PR #3796, opened 2022-05-12
share animation frame with renderService

### LINKED ISSUE #3795: decorationService and renderService should share an animation frame
originally reported in microsoft/vscode#145801

meganrogge:
fixes  #3795

### DIFF
--- a/src/browser/RenderDebouncer.ts
+++ b/src/browser/RenderDebouncer.ts
@@ -30,7 +30,9 @@ export class RenderDebouncer implements IRenderDebouncerWithCallback {
   public addRefreshCallback(callback: FrameRequestCallback): number {
     this._refreshCallbacks.push(callback);
     if (!this._animationFrame) {
-      this._animationFrame = window.requestAnimationFrame(() => this._innerRefresh());
+      // this._animationFrame = window.requestAnimationFrame(() => this._innerRefresh());
+      // throw new Error('no animation frame');
+      return 0;
     }
     return this._animationFrame;
   }
--- a/src/browser/RenderDebouncer.ts
+++ b/src/browser/RenderDebouncer.ts
@@ -52,6 +52,8 @@ export class RenderDebouncer implements IRenderDebouncerWithCallback {
   }
 
   private _innerRefresh(): void {
+    this._animationFrame = undefined;
+
     // Make sure values are set
     if (this._rowStart === undefined || this._rowEnd === undefined || this._rowCount === undefined) {
       this._runRefreshCallbacks();
@@ -65,7 +67,6 @@ export class RenderDebouncer implements IRenderDebouncerWithCallback {
     // Reset debouncer (this happens before render callback as the render could trigger it again)
     this._rowStart = undefined;
     this._rowEnd = undefined;
-    this._animationFrame = undefined;
 
     // Run render callback
     this._renderCallback(start, end);
--- a/src/browser/RenderDebouncer.ts
+++ b/src/browser/RenderDebouncer.ts
@@ -3,7 +3,7 @@
  * @license MIT
  */
 
-import { IRenderDebouncer, IRenderDebouncerWithCallback } from 'browser/Types';
+import { IRenderDebouncerWithCallback } from 'browser/Types';
 
 /**
  * Debounces calls to render terminal rows using animation frames.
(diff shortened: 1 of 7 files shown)


src/browser/decorations/BufferDecorationRenderer.ts
```diff
@@ -44,13 +43,10 @@ export class BufferDecorationRenderer extends Disposable {
...
 
   private _queueRefresh(): void {
-    if (this._animationFrame !== undefined) {
+    if (!this._renderService.getAnimationFrame()) {
       return;
     }
-    this._animationFrame = window.requestAnimationFrame(() => {
-      this.refreshDecorations();
-      this._animationFrame = undefined;
-    });
```
Tyriar, line 53: I was expecting a similar structure to `window.requestAnimationFrame` here:

```ts
if (this._animationFrame !== undefined) {
  return;
}
this._animationFrame = this._renderService.requestAnimationFrame(() => {
  this.refreshDecorations();
  this._animationFrame = undefined;
});
```

I don't think the current change will fix the problem as refreshing the buffer decoration renderer would be skipped if the renderer is doing something, otherwise it would refresh in the frame that requested it, which is bad since it's not an animation frame.

Tyriar (changes requested):

src/browser/Types.d.ts
```diff
@@ -308,4 +308,5 @@ export interface ICharacterJoiner {
 
 export interface IRenderDebouncer extends IDisposable {
   refresh(rowStart: number | undefined, rowEnd: number | undefined, rowCount: number): void;
+  requestAnimationFrame(callback: FrameRequestCallback): number | undefined;
```
Tyriar, line 311: I guess it doesn't make sense to have this on `TimeBasedDebouncer`, shall we remove it from the interface and have `RenderService` use the concrete `RenderDebouncer` instead?

src/browser/RenderDebouncer.ts
```diff
@@ -26,6 +26,15 @@ export class RenderDebouncer implements IRenderDebouncer {
     }
   }
 
+  public requestAnimationFrame(callback: FrameRequestCallback): number | undefined {
```
Tyriar, line 29: We want to keep track of an array of callbacks here and call them all at the end of `_innerRefresh`, that will guarantee the callbacks are run in the same animation frame as `_renderCallback`

Tyriar (changes requested):

src/browser/Types.d.ts
meganrogge: I'm not sure what you mean by use the concrete `RenderDebouncer`, but i've added an interface which extends IRenderDebouncer instead

Tyriar: @meganrogge the time based debouncer doesn't use an animation frame but a set timeout instead, so you should not be able to request an animation frame for it

meganrogge: yeah updated the code so I don't have that on there anymore

src/browser/RenderDebouncer.ts
```diff
@@ -26,6 +27,11 @@ export class RenderDebouncer implements IRenderDebouncer {
     }
   }
 
+  public addRefreshCallback(callback: FrameRequestCallback): number | undefined {
+    this._refreshCallbacks.push(callback);
+    return this._animationFrame;
+  }
```
Tyriar, line 33: This should return `number` instead of `number | undefined` and trigger a new animation frame if _animationFrame isn't set yet. This could happen if a decoration is added but there are no calls to trigger a refresh in the renderer, we may have other code that prevents that from happening elsewhere by decorations forcing a refresh but that could change/be refined:

```suggestion
  public addRefreshCallback(callback: FrameRequestCallback): number | undefined {
    this._refreshCallbacks.push(callback);
    if (!this._animationFrame) {
          this._animationFrame = window.requestAnimationFrame(() => this._innerRefresh());
    }
    return this._animationFrame;
  }
```

That alone won't work though since currently refreshes are only valid if they have rowStart/end/count set:

https://github.com/xtermjs/xterm.js/blob/2fa125904a5677c42d8676900dc9a3a858179fa7/src/browser/RenderDebouncer.ts#L46-L49

I'm thinking for that case we should still trigger all the callbacks and clear them out. Something like:

```ts
private _innerRefresh(): void {
  if (this._rowStart !== undefined && this._rowEnd !== undefined && this._rowCount !== undefined) {
    // refresh buffer
  }
  for (const callback of this._refreshCallbacks) {
    callback(0);
  }
  this._refreshCallbacks = [];
}
```

Tyriar (changes requested):

src/browser/RenderDebouncer.ts
meganrogge: okay yeah that makes sense.

meganrogge:
tests are passing when I run them locally.

Tyriar (approved):

PR #3800, opened 2022-05-13
Dom renderer: top decorations above selection

### LINKED ISSUE #3799: Decoration top layer backgroundColor is drawn underneath selection on DOM renderer
Oversight, should be easy to support in DOM renderer.

Tyriar:
Fixes #3799

### DIFF
--- a/src/browser/renderer/dom/DomRendererRowFactory.ts
+++ b/src/browser/renderer/dom/DomRendererRowFactory.ts
@@ -195,6 +195,11 @@ export class DomRendererRowFactory {
         isTop = d.options.layer === 'top';
       }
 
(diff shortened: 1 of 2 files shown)


src/browser/renderer/dom/DomRendererRowFactory.ts
```diff
@@ -195,6 +195,11 @@ export class DomRendererRowFactory {
         isTop = d.options.layer === 'top';
       }
 
+      // If it's a top decoration, render on above the selection
```
meganrogge, line 198:
```suggestion
      // If it's a top decoration, render above the selection
```

meganrogge (approved):

PR #3811, opened 2022-05-14
update search results `onBufferContentsChange`

### LINKED ISSUE #3794: incomplete search results
https://user-images.githubusercontent.com/29464607/168178520-66fe5943-918a-44d5-bb2d-728075aa2c68.mov

meganrogge:
fixes #3794

### DIFF
--- a/addons/xterm-addon-search/src/SearchAddon.ts
+++ b/addons/xterm-addon-search/src/SearchAddon.ts
@@ -498,7 +498,10 @@ export class SearchAddon implements ITerminalAddon {
       return this._findInLine(term, searchPosition, searchOptions);
     }
     let cache = this._linesCache?.[row];
-    if (!cache) {
+    if (!cache || cache[0] === '') {
+      // sometimes the search refresh happens before the
+      // buffer has content for a given line, so if the cached
+      // line is empty, check it again to avoid issues like #3794
       cache = this._translateBufferLineToStringWithWrap(row, true);
       if (this._linesCache) {
         this._linesCache[row] = cache;
--- a/addons/xterm-addon-search/src/SearchAddon.ts
+++ b/addons/xterm-addon-search/src/SearchAddon.ts
@@ -77,7 +77,7 @@ export class SearchAddon implements ITerminalAddon {
 
   public activate(terminal: Terminal): void {
     this._terminal = terminal;
-    this._onDataDisposable = this._terminal.onBufferContentsChange(() => this._updateMatches());
+    this._onDataDisposable = this._terminal.onWriteParsed(() => this._updateMatches());
     this._onResizeDisposable = this._terminal.onResize(() => this._updateMatches());
   }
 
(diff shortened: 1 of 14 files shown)


addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -498,7 +498,10 @@ export class SearchAddon implements ITerminalAddon {
       return this._findInLine(term, searchPosition, searchOptions);
     }
     let cache = this._linesCache?.[row];
-    if (!cache) {
+    if (!cache || cache[0] === '') {
+      // sometimes the search refresh happens before the
+      // buffer has content for a given line, so if the cached
+      // line is empty, check it again to avoid issues like #3794
```
Tyriar, line 504: This will end up causing a lot of re-checking empty rows, the root of the issue seems to be listening to the public `onData` API (ie. onKey press). I think we actually want to listen debounced to either when write data is parsed or when the buffer changes.

Tyriar (changes requested):

addons/xterm-addon-search/src/SearchAddon.ts
meganrogge: ok, will have to add a new event for either of those as currently (and to my surprise), `onBufferChange` currently fires when the buffer type changes from `active` to `alt` or vice-versa - not when the buffer changes.

listening to when write data is parsed is probably better bc I think it would be the same issue with `onBufferChange` since the buffer would change (a new empty row would be added) and that value of `""` would be stored and then never updated.

Tyriar: It's `onBufferChange` vs `onRenderedBufferChange` that might be confusing? We don't want either at the latter is only when the viewport changes but we want when anywhere in the buffer changes. Let's talk about better names and the right fix for this offline.

meganrogge: `onRenderedBufferChange` => `onRenderedViewportChange`

new one: `onBufferContentsChange`

meganrogge: fire event at end of innerWrite in writeBuffer
onWriteBufferParsedData

expose public API `onParseData` in

meganrogge: track aboslute cursor position and then always refresh the line that it searched up to and everything below

typings/xterm.d.ts
```diff
@@ -839,6 +839,12 @@ declare module 'xterm' {
      */
     onRender: IEvent<{ start: number, end: number }>;
 
+    /**
+     * Adds an event listener for when a chunk of data has been
+     * processed in the write buffer.
+     */
+    onBufferContentsChange: IEvent<void>;
```
Tyriar, line 846: Was just writing up some jsdoc to also call out the fact that this isn't entirely correct. I think the better thing here is to make the event name more correct so we don't need to call out caveats.

```suggestion
    /**
     * Adds an event listener for when data has been parsed by the terminal,
     * after {@link write} is called. This event is useful to listen for any
     * changes in the buffer.
     * 
     * This fires at most once per frame, after data parsing completes. Note
     * that this can fire when there are still writes pending if there is a lot
     * of data.
     */
    onWriteParsed: IEvent<void>;
```

Tyriar (changes requested):

Tyriar (approved):

Issue #3821, opened 2022-05-19
registerLinkProvider integration tests are quite flaky


Tyriar:
On firefox it seems to flake more.

```
API Integration Tests
       registerLinkProvider
         should fire activate events when clicking the link
```

Tyriar:
Seems to be a race condition where `Linkifier2._handleNewLink` happens after `_onClick`

Tyriar:
The root cause actually seems to be page.evaluate doesn't wait for `registerLinkProvider` to run before running page.mouse.down/up later on in the test 😕

PR #3822, opened 2022-05-19
Deflake registerLinkProvider tests

Tyriar:
The root cause of this flakiness problem was related to the click event
handler. For some reason it would not go off on some links but it would
when clicking elsewhere in the terminal. I tried a bunch of things to
keep the click handler to no avail, instead opting to move to using
mousedown and mouseup events for activating links which comes with the
nice benefit of ensuring that the link on mouseup is the same one on
mousedown.

Fixes https://github.com/xtermjs/xterm.js/issues/3821


package.json
```diff
@@ -33,8 +33,8 @@
     "test": "npm run test-unit",
     "posttest": "npm run lint",
     "test-api": "npm run test-api-chromium",
-    "test-api-chromium": "node ./bin/test_api.js --browser=chromium --timeout=20000",
-    "test-api-firefox": "node ./bin/test_api.js --browser=firefox --timeout=20000",
+    "test-api-chromium": "node ./bin/test_api.js --browser=chromium --timeout=200000",
```
meganrogge, line 36:
```suggestion
    "test-api-chromium": "node ./bin/test_api.js --browser=chromium --timeout=20000",
```

package.json
```diff
@@ -33,8 +33,8 @@
...
+    "test-api-firefox": "node ./bin/test_api.js --browser=firefox --timeout=200000",
```
meganrogge, line 37:
```suggestion
    "test-api-firefox": "node ./bin/test_api.js --browser=firefox --timeout=20000",
```

meganrogge (approved):

package.json
Tyriar: Opps 😅

PR #3824, opened 2022-05-20
Standardize css.toColor output to #rrggbb[aa]

Tyriar:
This will help reduce the chance of runtime exceptions that would
break the render loop.

### DIFF
--- a/src/common/Color.test.ts
+++ b/src/common/Color.test.ts
@@ -151,45 +151,45 @@ describe('Color', () => {
     });
   });
 
-  describe('css', () => {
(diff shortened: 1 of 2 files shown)


src/common/Color.test.ts
```diff
@@ -151,45 +151,45 @@ describe('Color', () => {
     });
   });
 
-  describe('css', () => {
+  describe.only('css', () => {
```
meganrogge, line 154:
```suggestion
  describe('css', () => {
```

meganrogge (approved):

PR #3844, opened 2022-05-31
exclude box chars from contrast ratio demands

### LINKED ISSUE #3843: Exclude box drawing characters from min contrast ratio
VS Code issue: https://github.com/microsoft/vscode/issues/150777

This framework uses box drawing characters as a scroll bar, these characters should probably be excluded from the min contrast ratio feature:

image

meganrogge:
fix #3843

### DIFF
--- a/src/browser/renderer/RendererUtils.ts
+++ b/src/browser/renderer/RendererUtils.ts
@@ -16,3 +16,11 @@ export function isPowerlineGlyph(codepoint: number): boolean {
   // ratio standards
   return 0xE0A0 <= codepoint && codepoint <=  0xE0D6;
 }
+
+function isBoxGlyph(codepoint: number): boolean {
+  return 0x2500 <= codepoint && codepoint <= 0x257F;
+}
+
+export function excludeFromContrastRatioDemands(codepoint: number): boolean {
+  return isPowerlineGlyph(codepoint) || isBoxGlyph(codepoint);
+}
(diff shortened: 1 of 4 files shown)


src/browser/renderer/RendererUtils.ts
```diff
@@ -16,3 +16,11 @@ export function isPowerlineGlyph(codepoint: number): boolean {
   // ratio standards
   return 0xE0A0 <= codepoint && codepoint <=  0xE0D6;
 }
+
+function isBoxGlyph(codepoint: number): boolean {
```
meganrogge, line 20: not sure if this is the correct range. used this as a reference

https://unicodemap.org/range/52/Box_Drawing/

meganrogge: seems I'm missing some
Screen Shot 2022-05-31 at 3 43 18 PM

meganrogge: actually maybe those are block chars?

Tyriar: Yep, also "block elements" https://en.wikipedia.org/wiki/Box-drawing_character#Block_Elements

src/browser/renderer/RendererUtils.ts
```diff
@@ -17,10 +17,10 @@ export function isPowerlineGlyph(codepoint: number): boolean {
   return 0xE0A0 <= codepoint && codepoint <=  0xE0D6;
 }
 
-function isBoxGlyph(codepoint: number): boolean {
-  return 0x2500 <= codepoint && codepoint <= 0x257F;
+function isBoxOrBlockGlyph(codepoint: number): boolean {
+  return (0x2500 <= codepoint && codepoint <= 0x257F) || (0x2580 <= codepoint && codepoint <= 0x259F);
```
Tyriar, line 21: Can simplify this::

```suggestion
  return (0x2500 <= codepoint && codepoint <= 0x259F);
```

Tyriar (approved):

Tyriar:
Ha, the tests failed because it's a box drawing character used in the webgl api tests. You can skip the test and create an issue to handle in debt week.

PR #3856, opened 2022-06-11
Custom common powerline glyph rendering

### LINKED ISSUE #3827: Powerline symbols should not use sub-pixel anti-aliasing
This happens on all renderers and makes it look suboptimal:

image

Tyriar:
Fixes #3827

Canvas renderer, no powerline symbols in font:

Screen Shot 2022-06-11 at 4 18 29 pm

Webgl renderer, no powerline symbols in font:

Screen Shot 2022-06-11 at 4 19 09 pm
Screen Shot 2022-06-11 at 4 21 47 pm

I ended up just doing the powerline symbols as it was hard to convert the font glyph's spline set to canvas instructions and the main reason we're doing this is to get the powerline to blend in with background cells correctly.

### DIFF
--- a/demo/client.ts
+++ b/demo/client.ts
@@ -585,6 +585,32 @@ function powerlineSymbolTest() {
   term.writeln(`0xA_  ${s('\ue0a0')}${s('\ue0a1')}${s('\ue0a2')}`);
   term.writeln(`0xB_  ${s('\ue0b0')}${s('\ue0b1')}${s('\ue0b2')}${s('\ue0b3')}`);
   term.writeln('');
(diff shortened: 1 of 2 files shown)


Tyriar:
Clipping is making the line look a bit off:

Screen Shot 2022-06-11 at 6 00 07 pm

Not clipping will probably make it look weird, moving it in each side by 1/2 dpr would change the angle slightly but probably look best.

meganrogge (approved):

Tyriar:
Tweaked the outline to have padding on either side:

image

PR #3865, opened 2022-06-16
Fix a typo in serialize addon test

mmis1000:
Fix a typo in serialize addon test.

The two line in this test are mean to represent exact content with line wrap different only would be detected by test utilities.

But due to a typo. The two lines end up having different visible text content.

This PR fix such issue.


Tyriar (approved):
Thanks for spotting 🙂

PR #3866, opened 2022-06-17
Fix clipping of non-powerline chars

Tyriar:
See microsoft/vscode#152478


Tyriar:
Merging without review as it's a pretty bad issue

PR #3868, opened 2022-06-20
clear all markers

meganrogge:
fixes https://github.com/microsoft/vscode/issues/151645 and reverts the changes meant to fix https://github.com/microsoft/vscode/issues/144142 without causing that issue again because the actual fix is now in https://github.com/microsoft/vscode/blob/2d4988b53b9d7d8c309ffe9f7ef457b4b8fd672e/src/vs/workbench/contrib/terminal/browser/xterm/xtermTerminal.ts#L397-L404


Tyriar (approved):
Nice find 👍

Issue #3848, opened 2022-06-02
Wheel event frequency in mouse mode too high when using a touchpad


pfitzseb:
When the vt200 mouse mode is enabled, touch scrolling emits wheel events with a very high frequency. I suspect this happens because
https://github.com/xtermjs/xterm.js/blob/a5f2ea336eff69c784414f371baacde641f03517/src/browser/Terminal.ts#L707-L709
only checks that `deltaY !== 0`, instead of summing multiple `deltaY`s until the sum is greater than some threshold.

The real-world consequence of this is that tmux mouse-scrolling (well, touchpad scrolling) is basically unusable because the sensitivity is way too high.

## Details
- Browser and browser version: Electron 17.4.1/Chromium 98/VS Code 1.67
- OS version: Manjaro
- xterm.js version: Whatever is shipped in VS Code, but also doesn't really matter

### Steps to reproduce
1. `printf "\e[?1000h"` in an xterm.js terminal and a more native terminal (I was using alacritty, the original report was about whatever terminal is the default on Macs)
2. Use a touchpad to scroll
3. Check the frequency of the emitted scroll events

Or just start `tmux` and enable mouse mode via `set -g mouse on`; you should notice a difference in scroll speed between the two terminals.

Tyriar:
👍 we'd want to do something like viewport's handling of the wheel event:

https://github.com/xtermjs/xterm.js/blob/8adf9473b76a146112c67b4e06ef746935a826db/src/browser/Viewport.ts#L32-L35

https://github.com/xtermjs/xterm.js/blob/8adf9473b76a146112c67b4e06ef746935a826db/src/browser/Viewport.ts#L224-L241

PR #3869, opened 2022-06-22
Only send wheel mouse event when scrolling a line or more

pfitzseb:
Fixes https://github.com/xtermjs/xterm.js/issues/3848.

Haven't added a test since all of the mouse event functionality seems to be untested and I'm not sure how to set up the necessary scaffolding.

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -703,10 +703,10 @@ export class Terminal extends CoreTerminal implements ITerminal {
           but = ev.button < 3 ? ev.button : CoreMouseButton.NONE;
           break;
         case 'wheel':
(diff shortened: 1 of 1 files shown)


pfitzseb:
Re-using viewport's `_wheelPartialScroll` isn't technically correct, but shouldn't have any real consequences; the same implementation is used elsewhere too:
https://github.com/xtermjs/xterm.js/blob/52044e9aadd9eeaab702b88bd6e177b4ce336e80/src/browser/Terminal.ts#L863

Tyriar (approved):
Looks good, I agree it's not ideal we're using this from Viewport but as you say it's used elsewhere already.

PR #3871, opened 2022-06-24
Add Render to list of xterm.js users

crcastle:
Hi folks- Copying this over from an erroneous [PR](https://github.com/xtermjs/xtermjs.org/pull/177) I made on the xtermjs.org website repo:

> Hello xterm.js folks- Thanks for creating and maintaining this great project. This PR adds [Render](https://render.com) to the list of xterm.js real world uses.
> 
> We have been using xterm.js for a few years to give our users a command prompt for their deployed container. We also use it to show streaming build and runtime logs for user services.
> 
> CleanShot 2022-06-24 at 10 41 48@2x


Tyriar (approved):
Thanks 🙂

PR #3876, opened 2022-06-27
Add CloudTTY to list of xterm users

yankay:
CloudTTY is a newly oss project to provide a  Friendly Kubernetes CloudShell (Web Terminal).It uses xterm.js for displaying information.(with ttyd or gotty)

Existing project already provides great feature to share terminal over the web. But Kubernetes cloud native enviroment requires a way to run web-tty via kubernetes way ( running as pod, and generated by CRD). So you can try cloudtty 🎉.

image


Tyriar (approved):

yankay:
Thank you

PR #3904, opened 2022-07-17
OSC hyperlink support

### LINKED ISSUE #1134: Support hyperlink ansi escapes

Tyriar:
Weekend project 🤓 

Fixes #1134

TODO

- [ ] Resolve TODOs
- [ ] How do wrapped lines get handled when reflow occurs?
- [ ] Share underline status between links with the same id and uri
- [ ] General clean up
- [ ] Tests
- [ ] Webgl renderer
- [ ] Hover/activate handlers in API

### DIFF
--- a/css/xterm.css
+++ b/css/xterm.css
@@ -167,6 +167,10 @@
     text-decoration: underline;
 }
 
+.xterm-link {
+    text-decoration: underline dashed;
+}
+
 .xterm-strikethrough {
     text-decoration: line-through;
 }
--- a/src/browser/Linkifier2.test.ts
+++ b/src/browser/Linkifier2.test.ts
@@ -11,7 +11,7 @@ import { ILink } from 'browser/Types';
 
 class TestLinkifier2 extends Linkifier2 {
   constructor(bufferService: IBufferService) {
-    super(bufferService);
+    super(null!, bufferService);
   }
 
   public set currentLink(link: any) {
--- a/src/browser/Linkifier2.ts
+++ b/src/browser/Linkifier2.ts
@@ -134,7 +134,6 @@ export class Linkifier2 extends Disposable implements ILinkifier2 {
     // If there is an OSC link, use that
     const oscLinks = this._oscLinkStore.getLinksByLine(position.y - 1);
     const oscLink = oscLinks.find(e => e.ranges.some(p => position.x >= p.x && position.x <= p.x + p.length));
-    console.log('oscLink hovered!', oscLink);
     if (oscLink) {
       const linkWithState = this._oscLinkToLinkWithState(oscLink, position);
       if (linkWithState) {
--- a/src/browser/Linkifier2.ts
+++ b/src/browser/Linkifier2.ts
@@ -4,13 +4,12 @@
  */
 
 import { ILinkifier2, ILinkProvider, IBufferCellPosition, ILink, ILinkifierEvent, ILinkDecorations, ILinkWithState, IBufferRange } from 'browser/Types';
-import { IDisposable } from 'common/Types';
+import { IDisposable, IOscLink } from 'common/Types';
 import { IMouseService, IRenderService } from './services/Services';
-import { IBufferService } from 'common/services/Services';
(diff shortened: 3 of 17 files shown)


jerch:
@Tyriar Why dont you use the extended attributes for marking the cells? This would automatically deal with line wrapping, reflow and such as it is an attribute directly on a cell like SGR. It also would avoid bad runtime from the event dispatching of `onPrintChar` for every single character.

For reference: #2904 contains fully working hyperlink code using extended attributes. If added into core the needed changes are very few. Only missing feature back then was the multi range support of the linkifier, which is needed for proper spec conform behavior (and which you refused to add to the new linkfier).

Tyriar:
@jerch I completely forgot about your PR here 😅

> Why dont you use the extended attributes for marking the cells? This would automatically deal with line wrapping, reflow and such as it is an attribute directly on a cell like SGR.

My thought process was trying to make it a very standalone feature as opposed to being integrated into everything, but I ended up hitting a bit of a wall how to do it properly in the webgl renderer because I didn't do the attribute approach. It turns out your achieves better in this respect since it's an actual addon.

Correct me if I'm wrong but we never ended up getting underline color/style working in the renderers, just the buffer model? Seems like we could do this in webgl by specifying the number of bits for each extended attribute and then making glyphs accessible via 3 keys (fg, bg, extended packed into a single number) here:

https://github.com/Tyriar/xterm.js/blob/a8405b86defb32743c3834a19741cbc117bbda5e/addons/xterm-addon-webgl/src/atlas/WebglCharAtlas.ts#L175-L181

> For reference: https://github.com/xtermjs/xterm.js/pull/2904 contains fully working hyperlink code using extended attributes. If added into core the needed changes are very few.

When I resume this work I think it'd probably be better to base it off yours and bring over any useful bits from this PR.

> Only missing feature back then was the multi range support of the linkifier, which is needed for proper spec conform behavior (and which you refused to add to the new linkfier).

I don't recall this discussion but I'd like to support the shared underline for all link ranges with the same id before pushing this in.

Tyriar:
> So yeah - I did not address the underline issue at all in the hyperlink draft, I just used normal underline similar to the weblink addon. With proper underline styles in all renderers a dotted/dashed default repr turning into a full underline on hover might work for a better UX.

Seems like I should tackle underline style/color rendering before continuing on this. I don't mind so much if the underlines look a little ugly in the dom renderer if that's just how the browser renders the underlines. As long as webgl looks great as it's the primary renderer

jerch:
> Seems like I should tackle underline style/color rendering before continuing on this.

While technically not really needed to get OSC 8 support, its prolly a good idea, as those underline styles/colors are pretty widespread in cmdline applications already.

Tyriar:
With https://github.com/xtermjs/xterm.js/pull/3921 we'll have full support for all underline styles/color. I tried to merge master into https://github.com/xtermjs/xterm.js/pull/2904 but it's pretty old and I struggled with that. So based off #3921 I'll bring in the best parts of both PRs, probably adopting for the addon + extended cell data approach. Also the buffer doesn't actually need to know about the link if we change the underline styles, so this could be in extended cell data but not in the new packed `ext` which is used as the glyph cache key.

jerch:
@Tyriar You already fixed the styles in the renderers - wow that was really fast. About #2904 - yeah, thats a very old old PR, might be easier to just pull some ideas from the code there, instead of messing around with the whole PR?

Tyriar:
Was a particularly productive weekend lol

Tyriar:
Closing for now, will use as a reference for the next try

PR #3907, opened 2022-07-18
Fix devcontainer docker build

silamon:
Moves the devcontainer from Debian stretch to Debian buster. 
* The node:14 docker container already moved to Debian buster some time ago, but not all required packages are available so it was failing to build the devcontainer completely. 
* Removes the puppeteer dependencies instead of finding replacements. 

Reference: https://github.com/nodejs/docker-node/commit/4f1d6a3875d1833ef57d87ed1999c41dbe6bd82c

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14
+FROM node:14-buster
 
 # Configure apt
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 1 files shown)


Tyriar (commented):
Should this move to stretch instead? That would get puppeteer working again?

silamon:
> Should this move to stretch instead? That would get puppeteer working again?
I don't see any references anymore to puppeteer, it's been replaced by playwright some years ago?
Debian stretch has been in LTS until June 2022. Therefore it was replaced in the node:14 docker image.

Tyriar (approved):
Oh right, I got mixed up

PR #3908, opened 2022-07-18
Add Cratecode to Real-world uses

uellenberg:
Hey all,

First off, I'd like to give my thanks to the team behind xterm.js (and whoever made the attach addon :laughing: ). Integrating xterm.js has been a breeze and it works perfectly.

This PR adds [Cratecode](https://cratecode.com) to the list of real-world applications. Cratecode is a service that helps people learn how to program through interactive lessons. Xterm.js's part in this is providing users with a full Linux environment. This is mostly used for running code, but users can also use it as a standard terminal, and I'm quite excited to see what else it'll be used for in the future!

image


Tyriar (approved):
Good to hear it was easy to use 🙂

PR #3912, opened 2022-07-19
Allow unsetting extending ansi colors

silamon:
Following up PR to address review comments in #3905.

### DIFF
--- a/src/browser/ColorManager.test.ts
+++ b/src/browser/ColorManager.test.ts
@@ -329,7 +329,7 @@ describe('ColorManager', () => {
       assert.equal(cm.colors.ansi[17].css, DEFAULT_ANSI_COLORS[17].css);
     });
 
-    it('should set extended ansi colors to the default after setting to something', () => {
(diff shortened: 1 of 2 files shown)


Tyriar (approved):

Tyriar:
@jerch in case I missed anything 😛

jerch (approved):
LGTM :smiley_cat:

There is just one thing idk for sure atm - when I was doing the OSC 4 implementation with hot color changes, there was a awkward color object reference thingy going on (not sufficiently re-grabbing colors from changes on individual slots). I am pretty sure I fixed it, still the issue kept me scratching my head back then.

So to be on the safe - it might be good to explicitly check, whether a `setTheme` call works as before, and also that extended colors shift properly. The old was only in canvas and webgl renderer.

Tyriar:
Tested running these on each renderer and it worked as expected:

```js
term.options.theme = { extendedAnsi: ['#ffffff']};

term.options.theme = { };
```

PR #3916, opened 2022-07-21
Fix memory leak

SvanT:
https://github.com/microsoft/vscode/issues/155232

After doing v8 heap profiling I found out that using terminal references
in the atlas cache makes it impossible for js to garbage collect the
terminals after disposal, causing both gpu and ram memory leak, therefore
switching to a simpler solution by just using an immutable id to break
the reference chain.

### DIFF
--- a/addons/xterm-addon-webgl/src/WebglRenderer.ts
+++ b/addons/xterm-addon-webgl/src/WebglRenderer.ts
@@ -6,7 +6,7 @@
 import { GlyphRenderer } from './GlyphRenderer';
 import { LinkRenderLayer } from './renderLayer/LinkRenderLayer';
 import { CursorRenderLayer } from './renderLayer/CursorRenderLayer';
-import { acquireCharAtlas } from './atlas/CharAtlasCache';
+import { acquireCharAtlas, removeTerminalFromCache } from './atlas/CharAtlasCache';
 import { WebglCharAtlas } from './atlas/WebglCharAtlas';
 import { RectangleRenderer } from './RectangleRenderer';
 import { IWebGL2RenderingContext } from './Types';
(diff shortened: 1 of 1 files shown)


Tyriar (commented):
Thanks for digging into this, do you think we could use [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef#avoid_where_possible) here instead to fix it so we could avoid adding the id prop?

SvanT:
Thanks for the fast review :)

I think that could be a good fit here perhaps (I haven't used it before but seems to tackle this kind of problems), but it seems like you have kind of IE11 support. Introducing a polyfill and [probably using ESNext](https://stackoverflow.com/a/66256279) as target for typescript could be a bit overkill for just an "identifier"?

Whatever you choose I can give it a try if you want, or feel free to refactor if you wish.

Tyriar (changes requested):
IE11's end of life was in June so I wouldn't let that stop us. I'm going to remove that note in the README now 😎

Let's try `WeakRef`, I haven't used it before so let me know if there are problems.

addons/xterm-addon-webgl/src/WebglRenderer.ts
```diff
@@ -112,6 +112,7 @@ export class WebglRenderer extends Disposable implements IRenderer {
       l.dispose();
     }
     this._canvas.parentElement?.removeChild(this._canvas);
+    removeTerminalFromCache(this._terminal);
```
Tyriar, line 115: Wonder why we didn't do that to begin with... Maybe because dispose in WebGL was only really properly done after your initial report? Thanks for looking into this! 👍

Tyriar (approved):

PR #3919, opened 2022-07-23
Dispose events

SvanT:
https://github.com/microsoft/vscode/issues/155232

The events that is setup for needs to be disposed for js to be able
to garbage collect the terminals.


SvanT:
@Tyriar this combined with https://github.com/microsoft/vscode/pull/155976 seems to finally make it so that vscode garbage collects terminals directly after they are closed :)

addons/xterm-addon-webgl/src/renderLayer/CursorRenderLayer.ts
```diff
@@ -54,6 +54,14 @@ export class CursorRenderLayer extends BaseRenderLayer {
     this.onOptionsChanged(terminal);
   }
 
+  public override dispose(): void {
+    if (this._cursorBlinkStateManager) {
+      this._cursorBlinkStateManager.dispose();
+      this._cursorBlinkStateManager = undefined;
+    }
```
Tyriar, line 61: Little more concise:

```suggestion
    this._cursorBlinkStateManager?.dispose();
    this._cursorBlinkStateManager = undefined;
```

src/browser/Linkifier.ts
```diff
@@ -53,6 +53,12 @@ export class Linkifier implements ILinkifier {
     };
   }
 
+  public dispose(): void {
+    this._onShowLinkUnderline.dispose();
+    this._onHideLinkUnderline.dispose();
+    this._onLinkTooltip.dispose();
+  }
```
Tyriar, line 60: The preferred way here is to have the class `extend Disposable` and register the emitters:

```ts
private _onShowLinkUnderline = this._register(new EventEmitter<ILinkifierEvent>());
```

src/browser/Types.d.ts
```diff
@@ -201,6 +201,7 @@ export interface ILinkifier {
   linkifyRows(start: number, end: number): void;
   registerLinkMatcher(regex: RegExp, handler: LinkMatcherHandler, options?: ILinkMatcherOptions): number;
   deregisterLinkMatcher(matcherId: number): boolean;
+  dispose(): void;
```
Tyriar, line 204: If we extend `Disposable` this interface could extend `IDisposable`

src/browser/Terminal.ts
```diff
@@ -247,8 +247,12 @@ export class Terminal extends CoreTerminal implements ITerminal {
     super.dispose();
     this._renderService?.dispose();
     this._customKeyEventHandler = undefined;
+    this._overviewRulerRenderer?.dispose();
+    this._overviewRulerRenderer = undefined;
     this.write = () => { };
     this.element?.parentNode?.removeChild(this.element);
+    this.linkifier.dispose();
+    this.linkifier2.dispose();
```
Tyriar, line 255: When extending Disposable:

```ts
    this.linkifier = this.register(this._instantiationService.createInstance(Linkifier));
    this.linkifier2 = this.register(this._instantiationService.createInstance(Linkifier2));
```

Looks like `linkifier2` is already getting registered.

src/browser/Terminal.ts
```diff
@@ -247,8 +247,12 @@ export class Terminal extends CoreTerminal implements ITerminal {
     super.dispose();
     this._renderService?.dispose();
     this._customKeyEventHandler = undefined;
+    this._overviewRulerRenderer?.dispose();
+    this._overviewRulerRenderer = undefined;
```
Tyriar, line 251: Best registered when it's created too

Tyriar (changes requested):
Great finds, just some clean up/best practices

SvanT:
@Tyriar Thanks for the feedback, I've updated the PR with the changes and I did a test that the disposal still works as expected.

Just a quick question, what is the difference of creating instances of a class through the InstantiationService instead of creating them using new Class() ?

Tyriar:
@SvanT the instantiation service is used for dependency injection, when you call createInstance it will automatically fill in the params with the `@I...Service` decoration, like this:

https://github.com/xtermjs/xterm.js/blob/d2be5f992a639884f93520dd5f1ae2e9ae8c4730/src/browser/renderer/dom/DomRendererRowFactory.ts#L36-L44

Tyriar:
I merged in the big v5 changes into master and this PR, `Linkifier` doesn't exist anymore so that's no longer needed here

Tyriar (approved):

PR #3921, opened 2022-07-23
Underline style and color rendering

### LINKED ISSUE #1145: Curly and colored underlines

Tyriar:
Fixes #1145 

This sets up the infrastructure for 3-keyed glyphs where the new key is `ext`. Currently `ext` is set to `underlineStyle` but will be moved to use the packed `ext` format, similar to `fg`/`bg`.

DOM

Screen Shot 2022-07-23 at 9 22 19 am

Canvas

Screen Shot 2022-07-23 at 9 22 41 am

Webgl

Screen Shot 2022-07-23 at 9 22 55 am


Tyriar:
Colors done too:

DOM

Screen Shot 2022-07-23 at 11 20 44 am

Canvas

Screen Shot 2022-07-23 at 11 20 25 am


Webgl

Screen Shot 2022-07-23 at 11 19 12 am

Tyriar:
No more overlapping of text/underline:
Screen Shot 2022-07-23 at 2 28 53 pm

Tyriar:
Continuous curves 😍 

Screen Shot 2022-07-23 at 2 49 04 pm

Tyriar:
dpr 1 curly rendering on Windows isn't great atm:

image
image

Tyriar:
Updated, it was bad when line width was 1:

image

image

Tyriar:
Glyph threshold clearing is causing an issue for underlines:

image

PR #3924, opened 2022-07-24
Allow markers on the alt buffer

### LINKED ISSUE #3918: Search highlights don't work in the alt buffer
run vi command
image

run cat  command
image

hightlight nor work in vi mode

## Details
- Browser and browser version:  Chrome v103.0.5060.134
- OS version: Big Sur 11.6.7
- xterm.js version:  v4.19.0
   xterm-addon-search v0.10.0-beta.3

silamon:
Fixes #3918 

An additional event to trigger on buffer change wasn't needed, since the onWriteParsed is being triggered.


Tyriar (changes requested):
One special thing that comes to mind here is for alt buffers we want to dispose all buffers as soon as the active buffer switches back to the normal buffer. I think the fix for that would be to call `clearAllMarkers()` here:

https://github.com/xtermjs/xterm.js/blob/055bcce18de91f329f61261aa1c81c1da231f7a9/src/common/buffer/BufferSet.ts#L83-L86

silamon:
> 

Without clearAllMarkers, you can see that the alt buffer decorations persist on the screen until the normal buffer is rendered. The search addon disposes the decorations/markers then on the onWriteParsed() event. Adding clearAllMarkers disposes them before clearing the all buffer. 

It might be even better to add clearAllMarkers to the clear function of the buffer (?)

Tyriar:
@silamon since markers are changing to work also in the alt buffer we'll need to cover all marker use cases, this means making them invalid when alt -> normal happens. To validate which solution is best you could also run this while the alt buffer is up:

```ts
const m = term.registerMarker();
```

Switch the the normal buffer and make sure `m` is now disposed.

src/common/buffer/BufferSet.test.ts
```diff
@@ -71,4 +71,17 @@ describe('BufferSet', () => {
       assert.equal(bufferSet.active.y, 10);
     });
   });
+
+  describe('markers', () => {
+    beforeEach(() => {
+    });
+
```
Tyriar, line 78:
```suggestion
```

Tyriar (approved):
Seems to work as expected 👍

Tyriar:
> The search-addon decorations are very hard to see with the background changes.

Do you mean the demo theme changes? The demo's options.theme isn't tied to the search theme currently and it's probably overkill for the demo. You can always switch the theme now though:

Screen Shot 2022-07-31 at 2 38 25 pm

Tyriar:
> It might be even better to add clearAllMarkers to the clear function of the buffer (?)

Adding to `clear` would call it in the normal buffer as well and I think there might be a problem with that. Just doing it for the alt buffer is the safer approach.

PR #3966, opened 2022-07-30
Fix mocha test explorer 

silamon:
Introduced by #3934. Sort of alternative solution to keep the mocha test explorer working for unit tests.
Somebody needs to check if I didn't break the api tests again. 
When mocha doesn't have a config flag passed, it takes by default the nearest file it can find.


.mocharc-default.yml
```diff
@@ -0,0 +1 @@
+# empty
```
Tyriar, line 1: Can we avoid including this file to keep the root directory a little cleaner?

Tyriar (changes requested):
Sorry about the revert, test-api still works with this though👍

Tyriar (approved):
Oh nice, super clean 🙂

silamon:
Don't merge this branch yet. The mocha test explorer didn't got all the functionality back it had with the config file.

Tyriar:
@silamon np, mark ready for review when it's ready to make sure I don't merge early

Tyriar (approved):
Still works 👍

Issue #3948, opened 2022-07-27
ITerminalOptions: Make fields required


silamon:
In #3667, all getOptions were replaced by their new equivalent: 
`this.dimensions.scaledCharLeft = Math.floor(this._terminal.getOption('letterSpacing') / 2);`
became
`this.dimensions.scaledCharLeft = Math.floor(this._terminal.options.letterSpacing! / 2);`

Mind the "!" there, which is a shortcut to let typescript know that the variable is for sure defined. In fact, that can be safely assumed since the default parameters are filled in. 
If the class ITerminalOptions is changed to have required fields which are not optional, it would be possible to have 
``this.dimensions.scaledCharLeft = Math.floor(this._terminal.options.letterSpacing / 2);``

It was kind of explored in #3448 that
``options: ITerminalOptions;``
can be changed to 
`get options(): ITerminalOptions;`
`set options(options: Partial<ITerminalOptions>): void;`
which would be an equivalent of the current code. Just ITerminalOptions is changed to have required fields. 

There are some settings though which can be "undefined": 
* fastScrollModifier: Perhaps we can introduce "none" here
* overviewRulerWidth: Use 0 to have this disabled?

This would also make the code more aligned, since the typings have ITerminalOptions with all fields optional and the internal ITerminalOptions have required fields. 

The breaking change part here lies in the fact that addons can use ITerminalOptions as type. If this is passed as a parameter, they would need to change the parameter to Partial\<ITerminalOptions\>.

Tyriar:
I like it. Also yes we should get rid of undefined in `fastScrollModifier` and `overviewRulerWidth` as I'm not even sure it's possible to unset them in the current state, I like your proposals.

Let's do this as part of v5 so we can make the (minor) breaking change

Tyriar:
Let me know if you don't have time in the next week and I can pick this up

PR #3968, opened 2022-07-30
Remove cols/rows from internal term.options

silamon:
This branch helps towards #3948. Keeping terminal.rows/terminal.cols and terminal.options.rows/terminal.options.cols in sync shouldn't be a problem then. 

Also aligns xterm with xterm-headless after #3960.

### DIFF
--- a/src/browser/Terminal.test.ts
+++ b/src/browser/Terminal.test.ts
@@ -1399,20 +1399,4 @@ describe('Terminal', () => {
       assert.deepEqual(markers.map(el => el.line), [-1, -1, 0, 1, 2]);
     });
   });
-
-  describe('options', () => {
-    beforeEach(async () => {
-      term = new TestTerminal({});
-    });
-    it('get options', () => {
-      assert.equal(term.options.cols, 80);
-      assert.equal(term.options.rows, 24);
-    });
-    it('set options', async () => {
-      term.options.cols = 40;
-      assert.equal(term.options.cols, 40);
-      term.options.rows = 20;
-      assert.equal(term.options.rows, 20);
-    });
-  });
 });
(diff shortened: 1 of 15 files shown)


src/browser/Terminal.test.ts
```diff
@@ -1399,20 +1399,4 @@ describe('Terminal', () => {
...
-      assert.equal(term.options.cols, 80);
-      assert.equal(term.options.rows, 24);
-    });
-    it('set options', async () => {
-      term.options.cols = 40;
-      assert.equal(term.options.cols, 40);
-      term.options.rows = 20;
-      assert.equal(term.options.rows, 20);
-    });
-  });
```
Tyriar, line 1417: We'll still want to keep these tests, just change to use different options

src/common/services/BufferService.ts
```diff
@@ -36,8 +36,8 @@ export class BufferService extends Disposable implements IBufferService {
     @IOptionsService private _optionsService: IOptionsService
   ) {
     super();
-    this.cols = Math.max(_optionsService.rawOptions.cols || 0, MINIMUM_COLS);
-    this.rows = Math.max(_optionsService.rawOptions.rows || 0, MINIMUM_ROWS);
+    this.cols = Math.max(80 || 0, MINIMUM_COLS);
+    this.rows = Math.max(24 || 0, MINIMUM_ROWS);
```
Tyriar, line 40: Do the initial dimensions need to be passed into the ctor now?

Tyriar (changes requested):
> Also aligns xterm with xterm-headless after https://github.com/xtermjs/xterm.js/pull/3960.

Opps 🤦

silamon:
Removing cols/rows from ITerminalOptions isn't going to work. Closing this PR and we will have to keep rows and cols in the internal ITerminalOptions.

PR #3972, opened 2022-07-30
Apply row/col changes from xterm to xterm-headless

silamon:
Pulled this out of closed #3968. 
Align xterm with xterm-headless after https://github.com/xtermjs/xterm.js/pull/3960.


Tyriar (approved):
Thanks for catching this 👍

PR #3982, opened 2022-07-31
ITerminalOptions: Make fields required

### LINKED ISSUE #3948: ITerminalOptions: Make fields required

silamon:
Fixes #3948.

When writing the issue, I was convinced that it would lead to better code. Now, after setting up all the code changes, it looks like it introduces a lot of Partials. I'm not really convinced that we should merge this as it is now.

### DIFF
--- a/addons/xterm-addon-serialize/src/SerializeAddon.test.ts
+++ b/addons/xterm-addon-serialize/src/SerializeAddon.test.ts
@@ -208,7 +208,7 @@ describe('xterm-addon-serialize', () => {
       const output = serializeAddon.serializeAsHTML({
         includeGlobalBackground: true
       });
-      assert.equal((output.match(/color: #ffffff; background-color: #000000; font-family: courier-new, courier, monospace; font-size: 15px;/g) || []).length, 1, output);
+      assert.equal((output.match(/color: #FFFFFF; background-color: #000000; font-family: courier-new, courier, monospace; font-size: 15px;/g) || []).length, 1, output);
     });
   });
 });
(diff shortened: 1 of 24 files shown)


Tyriar (changes requested):
Yeah that is a lot of changes, does switching it around and using `Required<ITerminalOptions>` instead lead to nicer code?

silamon:
> Yeah that is a lot of changes, does switching it around and using `Required<ITerminalOptions>` instead lead to nicer code?

That would be a lot less code changes, and I think the breaking change wouldn't be there too since ITerminalOptions doesn't change.

PR #3983, opened 2022-07-31
Fix demo

silamon:
WebGL throws error on initialization in some cases, breaking the demo.


Tyriar (commented):
Do you know what the error is that causes this?

silamon:
> Do you know what the error is that causes this?

Nothing fixable, it's just not supported (browser or hardware).

Tyriar (approved):

PR #3991, opened 2022-08-01
print colors

### LINKED ISSUE #3850: Add helper button to print a wide range of ansi colors
Similar to the box drawing and powerlines buttons in the test tab of the demo, a helper for printing out different color modes would be of great help. Something like this to give an idea, but that prints every color (a separate button for each color mode?)

```
echo -e '\x1b[0mDefault'
echo -e '\x1b[7mDefault (inverse)\x1b[0m'
echo -e '\x1b[33mPalette 16 (color #3, yellow)\x1b[0m'
echo -e '\x1b[38;5;50mPalette 256 (color #50)\x1b[0m'
echo -e '\x1b[38;2;255;0;0mRGB/True color (color #FF0000)\x1b[0m'
```

deepanshu44:
Implemented ansi-colors button.
Fixes #3850


deepanshu44:
@Tyriar Output looks something like below, is it what it should be? 
ansi-colors2

Tyriar (approved):
Thanks @deepanshu44, I tweaked the layout and reduce duplicate.

PR #4004, opened 2022-08-05
ITerminalOptions: Make fields required

### LINKED ISSUE #3948: ITerminalOptions: Make fields required

silamon:
Fixes #3948. Pull request #3982 but the other way around, introducing Required. 

It's still a breaking change though, thanks to the options changes in fastScrollModifier and overviewRulerWidth.

### DIFF
--- a/typings/xterm-headless.d.ts
+++ b/typings/xterm-headless.d.ts
@@ -84,7 +84,7 @@ declare module 'xterm-headless' {
     /**
      * The modifier key hold to multiply scroll speed.
      */
-    fastScrollModifier?: 'alt' | 'ctrl' | 'shift' | undefined;
+    fastScrollModifier?: 'none' | 'alt' | 'ctrl' | 'shift';
 
     /**
      * The spacing in whole pixels between characters.
@@ -538,11 +538,12 @@ declare module 'xterm-headless' {
      * ```typescript
      * console.log(terminal.options.fontSize);
      * ```
+     */
+    get options(): Required<ITerminalOptions>;
+
+    /**
+     * Gets or sets the terminal options. This supports setting multiple options.
      *
-     * @example Set a single option
-     * ```typescript
-     * terminal.options.fontSize = 12;
-     * ```
      *
      * @example Set multiple options
      * ```typescript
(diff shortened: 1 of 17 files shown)


typings/xterm-headless.d.ts
```diff
@@ -538,11 +538,12 @@ declare module 'xterm-headless' {
...
+     */
+    get options(): Required<ITerminalOptions>;
+
+    /**
+     * Gets or sets the terminal options. This supports setting multiple options.
      *
-     * @example Set a single option
-     * ```typescript
-     * terminal.options.fontSize = 12;
-     * ```
```
Tyriar, line 545: Should we move this example to `get options` now?

Tyriar (approved):
Looking good 👍

typings/xterm-headless.d.ts
silamon: Yeah, indeed.

Tyriar (approved):
👍

PR #4005, opened 2022-08-07
OSC hyperlink support

### LINKED ISSUE #1134: Support hyperlink ansi escapes

Tyriar:
Previous attempts: https://github.com/xtermjs/xterm.js/pull/2904

Fixes #1134

### DIFF
--- a/src/browser/OscLinkProvider.ts
+++ b/src/browser/OscLinkProvider.ts
@@ -1,0 +1,72 @@
+/**
+ * Copyright (c) 2022 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { ILink, ILinkProvider } from 'browser/Types';
+import { CellData } from 'common/buffer/CellData';
+import { IBufferService, IOscLinkService } from 'common/services/Services';
+
+export class OscLinkProvider implements ILinkProvider {
+  constructor(
+    @IBufferService private readonly _bufferService: IBufferService,
+    @IOscLinkService private readonly _oscLinkService: IOscLinkService
+  ) {
+  }
+
+  public provideLinks(y: number, callback: (links: ILink[] | undefined) => void): void {
+    const line = this._bufferService.buffer.lines.get(y - 1);
+    if (!line) {
+      callback(undefined);
+      return;
+    }
+
+    const result: ILink[] = [];
+    const cell = new CellData();
+    const lineLength = line.getTrimmedLength();
+    let currentLinkId = -1;
+    let currentStart = -1;
+    let finishLink = false;
+    for (let x = 0; x < lineLength; x++) {
+      if (!line.hasContent(x)) {
+        continue;
+      }
+
+      line.loadCell(x, cell);
+      if (cell.extended.urlId) {
+        if (currentStart === -1) {
+          currentStart = x;
+          currentLinkId = cell.extended.urlId;
+          continue;
+        } else {
+          finishLink = cell.extended.urlId !== currentLinkId;
+        }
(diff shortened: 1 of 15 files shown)


Tyriar:
This is basically done, I'm going to wait until next month to push it though considering the amount of changes that are already in this month.

ssbarnea:
@Tyriar I know lots of people looking forward to get hyperlink support enabled in xterm. Lost of vscode users will be thrilled to see this in sooner.

Tyriar:
@ssbarnea should land in 1.72 🙂

src/browser/OscLinkProvider.ts
```diff
@@ -0,0 +1,110 @@
...
+    const newWindow = window.open();
+    if (newWindow) {
+      try {
+        newWindow.opener = null;
+      } catch {
+        // no-op, Electron can throw
+      }
+      newWindow.location.href = uri;
+    } else {
+      console.warn('Opening link blocked as opener could not be cleared');
```
glensc, line 107: the message does not seem correct. this else code block is reached when `window.open()` returns `null`, not `newWindow.opener=null` assignment throws.

jerch:
@Tyriar Oh well abit late to the party - do you think this should get a review? Sorry the PR kinda went unnoticed to me...

Tyriar:
@jerch sure a review is always welcome! Can always react to feedback retroactively. I'm trying to integrate it with VS Code atm which is what https://github.com/xtermjs/xterm.js/pull/4087 is about

src/browser/OscLinkProvider.ts
Tyriar: This is based on:

https://github.com/Tyriar/xterm.js/blob/c5861d512c41b179fe944c1c032ad13d3a070dc0/addons/xterm-addon-web-links/src/WebLinksAddon.ts#L29-L41

I think the thing here is it can only be cleared if the object is returned. Been a while since I touched the code though.

glensc: the message there is wrong too then.

Tyriar: I disagree, this warning is about the case where the (older) browser does not return a proxy object in which case we cannot clear the opener. "Opening link blocked as window proxy object not returned" is not as helpful as this as it doesn't talk about what we need to do.

I'm not even sure what browsers this covers, probably ones we don't support. It's not worth spending time on refining the warning to be perfect because of that anyway.

glensc: In my mind "opener could not be cleared" would be for this line of code: `window.opener = null`, and the `else` part never covers that code line, i.e window.opener is never even attempted to assigned `null` when the warning is printed.

Tyriar: I see what you're saying, your way would be technically more correct but the current way is more useful to the user of the library imo.

glensc: All right, good that you understood what I had in mind!

Issue #3600, opened 2022-01-08
Erratic text output from typing into xterm console on Chrome on Android devices


Bardia95:
We are using xterm at replit to provide users with a console, and our app runs on mobile as well. A significant number of our Android users are reporting that they are having issues with typing into their console, specifically, they are seeing their characters get jumbled up, duplicate characters, and weird text.

I found this from CodeMirror which would explain it:
> Chrome Android, especially in combination with GBoard, not only doesn't reliably fire regular key events, but also often surrounds the effect of enter or backspace with a bunch of composition events that, when interrupted, cause text duplication or other kinds of corruption.

They've documented the workaround they use for their code editor [here](https://github.com/codemirror/view/blob/46c9816df1ab0987235174e6e136f7952d5ca473/src/domobserver.ts#L226), which you can use as reference for fixing this bug around the Android Chrome limitation.

We would be very appreciative if you could prioritize this fix given the significant number of people affected by it.

## Details
- Browser and browser version: Chrome (latest)
- OS version: Android (Galaxy S9)
- xterm.js version: 4.14.1

### Steps to reproduce

1. Type a number into the console.
2. See duplicated text output of the number.

Tyriar:
> We would be very appreciative if you could prioritize this fix given the significant number of people affected by it.

You're welcome to contribute improvements as this is very low on my priority list.

luxi78:
+1
This problem also bothered me and other people using android devices

mwllgr:
Same here, even with Firefox on Android. Coming from MeshCentral.

HALOCORE:
Same here, the default keyboard on Android is not usable so I am trying to combining customized virtual keyboard (such as  with this. However, virtual keyboard (such as KioskBoard) expose event handlers in JavaScript and there need to be a way to inject those key events to xterm.js. 

I think if there is an API like the SendKey mentioned here, a lot of problems can be solved externally.
#3578

HALOCORE:
Actually by adding JavaScript code that creates virtual keyboard to the webpage or put the original webpage containing xterm.js in an iframe, it is working on android chrome. No need to focus the element when typing on the virtual keyboard and the android keyboard will not pop up. Events created by the virtual keyboard are sent to `document.getElementById("terminal").querySelector(".xterm-screen").querySelector(".xterm-helpers").querySelector(".xterm-helper-textarea")`. Anyway, it would be more convenient if xterm.js has such keyboard builtin.
[...]

PR #4007, opened 2022-08-08
Handle Delete in `CompositionHelper`

ink404:
#### Browser
Tested in chrome on Android device running SDK 12.

#### Issue - relates to https://github.com/xtermjs/xterm.js/issues/3600
Deleting characters doesn't work as expected on Android since the `CompositionHelper` expects the diff to have > 1 length if there's a difference.

Steps:
- type: `1`
- tap backspace button

expected:
- character is removed

actual:
- character remains, tapping backspace once more then removes character

Example:
+ single character entry: https://user-images.githubusercontent.com/3418052/183525959-2bb604e9-76cb-4e91-89a8-da14663e7218.mp4
+ many + press & hold to delete: https://user-images.githubusercontent.com/3418052/183528031-449451f7-eb3e-4e59-9970-e52922660ea7.mp4

### DIFF
--- a/src/browser/input/CompositionHelper.ts
+++ b/src/browser/input/CompositionHelper.ts
@@ -196,6 +196,8 @@ export class CompositionHelper {
           this._coreService.triggerDataEvent(diff, true);
         } else if (newValue.length < oldValue.length) {
           this._coreService.triggerDataEvent(`${C0.DEL}`, true);
+        } else if ((newValue.length === oldValue.length) && (newValue !== oldValue)) {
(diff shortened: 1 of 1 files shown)


ink404:
@Tyriar Hi! This fixes a long standing for our use case. 

Is there any chance this can be looked at in the next week?

Tyriar (approved):
Thanks for looking into this, I didn't test on Android but it doesn't seem to regress desktop and makes sense 👍

PR #4055, opened 2022-08-23
Fix typo and remove unnecessary eslint comments

GreenMashimaro:
### Description
fix typo and remove unnecessary eslint comments

### What is the purpose of this pull request?

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/src/browser/LocalizableStrings.ts
+++ b/src/browser/LocalizableStrings.ts
@@ -3,8 +3,5 @@
  * @license MIT
  */
 
-// eslint-disable-next-line prefer-const
-export let promptLabel = 'Terminal input';
-
-// eslint-disable-next-line prefer-const
-export let tooMuchOutput = 'Too much output to announce, navigate to rows manually to read';
+export const promptLabel = 'Terminal input';
+export const tooMuchOutput = 'Too much output to announce, navigate to rows manually to read';
--- a/src/browser/LocalizableStrings.ts
+++ b/src/browser/LocalizableStrings.ts
@@ -3,10 +3,8 @@
  * @license MIT
  */
 
-/*
- * note: intentional design, it's exposed to embedders of xterm.js
- * so they can change the strings to localize xterm.js. #4055
- */
+// This file contains strings that get exported in the API so they can be localized
+
 // eslint-disable-next-line prefer-const
 export let promptLabel = 'Terminal input';
 
--- a/src/browser/LocalizableStrings.ts
+++ b/src/browser/LocalizableStrings.ts
@@ -3,5 +3,12 @@
  * @license MIT
  */
 
-export const promptLabel = 'Terminal input';
-export const tooMuchOutput = 'Too much output to announce, navigate to rows manually to read';
+/*
+ * note: intentional design, it's exposed to embedders of xterm.js
+ * so they can change the strings to localize xterm.js. #4055
+ */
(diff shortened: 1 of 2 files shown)


src/browser/LocalizableStrings.ts
```diff
@@ -3,8 +3,5 @@
  * @license MIT
  */
 
-// eslint-disable-next-line prefer-const
-export let promptLabel = 'Terminal input';
-
-// eslint-disable-next-line prefer-const
-export let tooMuchOutput = 'Too much output to announce, navigate to rows manually to read';
```
Tyriar, line 10: This actually needs to be `let`, it's exposed to embedders of xterm.js so they can change the strings to localize xterm.js. It's exposed (as non-readonly) here https://github.com/xtermjs/xterm.js/blob/c0ef2ccb5a0b2f22f28ed594f3962d75cc83da00/src/browser/public/Terminal.ts#L242-L244

src/common/buffer/BufferSet.ts
```diff
@@ -58,14 +58,14 @@ export class BufferSet extends Disposable implements IBufferSet {
...
    */
   public get active(): Buffer {
     return this._activeBuffer;
   }
 
   /**
-   * Returns the currently active Buffer of the BufferSet
+   * Returns the normal Buffer of the BufferSet
    */
   public get normal(): Buffer {
```
Tyriar, line 70: 👍 this is a problem

Tyriar (changes requested):

src/browser/LocalizableStrings.ts
GreenMashimaro: Thanks for your reply, I have a few questions.

I verified it myself within the project.

### Question1: From the compilation results, the two are the same.

original code
``` typescript
export let promptLabel = 'Terminal input';
export let tooMuchOutput = 'Too much output to announce, navigate to rows manually to read';
```
after compilation code
``` javascript
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.tooMuchOutput = exports.promptLabel = void 0;
exports.promptLabel = 'Terminal input';
exports.tooMuchOutput = 'Too much output to announce, navigate to rows manually to read';
//# sourceMappingURL=LocalizableStrings.js.map
```
------
original code
``` typescript
export const promptLabel = 'Terminal input';
export const tooMuchOutput = 'Too much output to announce, navigate to rows manually to read';
```
after compilation code
``` javascript
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.tooMuchOutput = exports.promptLabel = void 0;
exports.promptLabel = 'Terminal input';
exports.tooMuchOutput = 'Too much output to announce, navigate to rows manually to read';
//# sourceMappingURL=LocalizableStrings.js.map
```


------
### Question2: Use `let` to declare variables; the value cannot be modified in the import place.

image

Tyriar: Ah I see, still using const shows the wrong intent here. `const` means this value should not change, and while currently TypeScript outputs is as mutable, it may not in the future. So I think we should keep this as it was

GreenMashimaro: I think your idea is good, and I suggest adding a comment. I re-updated the submission, please review it.

Tyriar (approved):
Thanks!

PR #4069, opened 2022-08-24
Make terminal rendering work in popout windows

### LINKED ISSUE #3758: Terminal cannot display text when created in a react portal window.
I am working on a react project, where we want to popout a terminal into a separate window. ([Here is the basic premise of how it works](https://medium.com/hackernoon/using-a-react-16-portal-to-do-something-cool-2a2d627b0202))

In my main project, though the terminal in the new window does open and all other functions involved within the component work, for some reason the text does not display in that popout window. 

The cursor will be in correct position and spaces will be created for lines, as if the text is there. I know that the terminal is receiving data but for some reason, it appears as if the canvas part of the terminal is unable to draw the text.

I was able to get the popout terminal to display partial text once or more by chance, when i created a lot of terminals in the popout window (I am unable to provide a consistent way of doing this). Leading me to believe there could be a problem with how other parts of the xterm.js source code handles getting the document element, weather it grabs from the default main browser window, by simply calling `document` or if someway it can get the window/ document object from the container element that is passed to terminal on `terminal.open()`.

As you will see in the test repo I provided, I had to ensure I got the correct document by accessing the new window in the `<Portal/>` component in App.js. But that is my guess as to what is happening. Please let me know if you need any more information.

## Details
- Browser and browser version: FireFox 99.0.1 (64-bit)
- OS version: Windows 21H2
- xterm.js version: ^4.18.0"
xterm-bug-test

### Steps to reproduce
1. Clone the test repo i have created ([here](https://github.com/JoshuaH-sudo/window-xterm.js-test))
2. run "npm i"
3. run "npm start"
4. should then see the terminal is created on the main browser window with the text showing, then in the popout window there will be a terminal present but no text displayed.

mihaip:
Make terminal rendering work in popout windows

Add support for the parent element (what is passed to Terminal.open)
being in a different (same origin) window. Accesses to DOM APIs such
as requestAnimationFrame and devicePixelRatio thus need to be scoped
to the corrent window, instead of assuming that it's the same as the
window/global scope where the code is running.

This is done by inferring a parent window at the creation time, and
then storing it in CoreBrowserService, which is already passed to most
places that need it.

To catch future regressions, an ESLint rule that checks for global
accesses is added (it uses AST selectors via the no-restricted-syntax
rule).

This should also be applicable when the parent element is an iframe.

Fixes #3758

| Before | After |
|-|-|
| xterm-poput-before | xterm-poput-after |

### DIFF
--- a/src/browser/services/CoreBrowserService.ts
+++ b/src/browser/services/CoreBrowserService.ts
@@ -9,12 +9,17 @@ export class CoreBrowserService implements ICoreBrowserService {
   public serviceBrand: undefined;
 
   constructor(
-    private _textarea: HTMLTextAreaElement
+    private _textarea: HTMLTextAreaElement,
+    public readonly window: Window & typeof globalThis
   ) {
   }
 
+  public get dpr(): number {
+    return this.window.devicePixelRatio;
+  }
+
   public get isFocused(): boolean {
-    const docOrShadowRoot = this._textarea.getRootNode ? this._textarea.getRootNode() as Document | ShadowRoot : document;
-    return docOrShadowRoot.activeElement === this._textarea && document.hasFocus();
+    const docOrShadowRoot = this._textarea.getRootNode ? this._textarea.getRootNode() as Document | ShadowRoot : this._textarea.ownerDocument;
+    return docOrShadowRoot.activeElement === this._textarea && this._textarea.ownerDocument.hasFocus();
   }
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -147,6 +147,29 @@
         ]
       }
     ],
+    "no-restricted-syntax": [
+      "warn",
+      {
+        "selector": "CallExpression[callee.name='requestAnimationFrame']",
+        "message": "The global requestAnimationFrame() should be avoided, call it on the parent window from ICoreBrowserService."
+      },
+      {
+        "selector": "CallExpression[callee.name='cancelAnimationFrame']",
+        "message": "The global cancelAnimationFrame() should be avoided, call it on the parent window from ICoreBrowserService."
+      },
+      {
+        "selector": "CallExpression > MemberExpression[object.name='window'][property.name='requestAnimationFrame']",
+        "message": "window.requestAnimationFrame() should be avoided, call it on the parent window from ICoreBrowserService."
+      },
+      {
+        "selector": "CallExpression > MemberExpression[object.name='window'][property.name='cancelAnimationFrame']",
+        "message": "window.cancelAnimationFrame() should be avoided, call it on the parent window from ICoreBrowserService."
+      },
+      {
+        "selector": "MemberExpression[object.name='window'][property.name='devicePixelRatio']",
+        "message": "window.devicePixelRatio should be avoided, get it from ICoreBrowserService."
+      }
+    ],
     "no-trailing-spaces": "warn",
     "no-unsafe-finally": "warn",
     "no-var": "warn",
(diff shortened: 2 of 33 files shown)


mihaip:
@Tyriar was wondering if you have any thoughts on this PR. I realize it's a bit invasive, I'm happy to brainstorm other approaches. The main use case that I'm looking for is to allow an application to open terminal sessions in a child/popout window (without requiring that the popout have a full copy of the app -- it should just be a rendering surface).

Tyriar (commented):
@mihaip it is indeed invasive. It's just the xterm.js lib you need to load into the other side and then some tweaks to allow your app to communicate with it isn't it? That doesn't seem like that much overhead.

jerch:
Beside what @Tyriar already said I also what to question that idea from a security perspective. Cross window/document handling is very hard to secure properly and might not even be possible due to pending or upcoming bugs in browser engines. For something like a terminal this is really bad to rely on, as we might open attack vectors that we cannot address/mitigate at all.

Furthermore we cannot easily introduce a potential security risk in xterm.js and tell webdevs to deal with it on their own, as many webdevs simply dont care (mind you - already had some heated discussions about risky CDN usage, where ppl will tell you "I always do it that way" - very problematic mindset with a terminal component).

mihaip:
@Tyriar there is some overhead in that the application now has to have a different chunk that is loaded independently (to avoid loading the entire bundle in the popout window) and a communication protocol/mechanism must be established with it. Each application that wishes to do this will need to replicate this setup. FWIWW other commonly used libraries (including React itself) support rendering in poput windows.

@jerch Note that this PR is only intended for use in same-origin/process popout windows. This does not relax any security constraints, and should not open any attack vectors.

Tyriar (changes requested):
The only issue I have with this is how much of a pain it will be to respect this. What about instead of passing around the parent window we create a new `IWindowService` like this:

```ts
interface IWindowService {
  window: Window;
  dpr: number;
}
```

`dpr` is here for convenience as it's used a lot. If we go this route we could maybe create an eslint rule to prevent future occurrences of `window`, as I'm sure they'll slip in unless we have some lint/compile rule to prevent it.

mihaip:
The `IWindowService` approach sounds promising, will look into reworking this PR.

mihaip:
@Tyriar I've reworked this to use a service. Instead of a new `IWindowService`, I ended up adding this to `ICoreBrowserService`, since it already seemed like the right place for such global-ish information (the existing `isFocused()` method is also per-window).

I also added an ESLint check for `window.requestAnimationFrame()`, `window.devicePixelRatio` and similar access patterns.

src/browser/services/CoreBrowserService.ts
```diff
@@ -9,12 +9,17 @@ export class CoreBrowserService implements ICoreBrowserService {
...
   constructor(
-    private _textarea: HTMLTextAreaElement
+    private _textarea: HTMLTextAreaElement,
+    public readonly window: Window & typeof globalThis
   ) {
   }
 
+  public get dpr(): number {
+    return this.window.devicePixelRatio;
+  }
```
Tyriar, line 19: @jerch FYI the perf concerns you had with getters are likely not an issue with the move from ES5 to ES2015.

Source:

```ts
class A {
  get b() { return 'b'; }
}
```

ES5:

```js
var A = /** @class */ (function () {
    function A() {
    }
    Object.defineProperty(A.prototype, "b", {
        get: function () { return 'b'; },
        enumerable: false,
        configurable: true
    });
    return A;
}());
```

ES2015:

```ts
class A {
    get b() { return 'b'; }
}
```

Guessing that's what a lot of the 17% reduction in https://github.com/xtermjs/xterm.js/pull/3992 was

Tyriar (approved):
Very nice, thanks for going the extra mile with the test button and lint rules 👍

Issue #4079, opened 2022-08-28
A large amount of CPU time is spent in loadColorsForCell


Tyriar:
Probably related to bg/fg override work, it's still not that long in the grand scheme of things, but it would be good to optimize:

![image](https://user-images.githubusercontent.com/2193314/187054907-75327d01-a60c-491b-9d01-8c593edb1a04.png)

This is when scrolling a 142x51 terminal in demo

PR #4080, opened 2022-08-28
Optimize gc from decoration-related hot code paths

Tyriar:
Fixes https://github.com/xtermjs/xterm.js/issues/4079

Before:

image

After:

image

### DIFF
--- a/addons/xterm-addon-canvas/src/BaseRenderLayer.ts
+++ b/addons/xterm-addon-canvas/src/BaseRenderLayer.ts
@@ -404,12 +404,11 @@ export abstract class BaseRenderLayer implements IRenderLayer {
 
     // Don't try cache the glyph if it uses any decoration foreground/background override.
     let hasOverrides = false;
-    for (const d of this._decorationService.getDecorationsAtCell(x, y)) {
+    this._decorationService.forEachDecorationAtCell(x, y, undefined, d => {
       if (d.backgroundColorRGB || d.foregroundColorRGB) {
         hasOverrides = true;
-        break;
       }
-    }
+    });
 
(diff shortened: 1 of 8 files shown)


jerch:
@Tyriar Maybe related (and maybe already fixed?) - when doing some perf benchmarks with canvas renderer I found this to be quite expensive: https://github.com/xtermjs/xterm.js/blob/4effa381f78f4b151e38736bb577915bb18c7d60/addons/xterm-addon-canvas/src/TextRenderLayer.ts#L190

I did not look at the code in detail, so I dont know what causes the high runtime for sure. I saw that there are some iterators/generators, maybe those are already the culprit? (They were a real showstopper in the past, but I think engines got alot better in handling those in a speedy manner). Since I had no decorations in place at all I wonder if there is at least an early fast exit possible to cut off most of the code path.

Tyriar:
@jerch it should be much faster now that it's forEach instead of the old iterator. The overhead was creating 2 iterators for each cell getting updated. Lovely to use but way too slow to create in this hot code path.

I've finished all the gc-related fixes if you want to re-try any tests.

jerch:
@Tyriar Oh well, I meant the `getDecorationsAtCell` above (wrongly linked to the new impl). The forEach impl is indeed alot faster (<<5% of renderer runtime), which is somewhat sad, as the yield code reads much nicer (at least for me coming from python).

The `w` object is an interesting hack around the gc - is that still needed with the forEach way? Would have thought, that the gc can simply fast collect it here not creating any overhead (other than with generators, where the shadowed stack frames created from yield state add more overhead for the gc).

Anyway - perf issue is fixed :+1:

Tyriar:
The work object is a common trick we already used, just new bits were added since. I don't know exactly what variables are optimized away, but `for (int i...)` for example ends up causing minor GCs

PR #4095, opened 2022-09-01
DECRQM support

### LINKED ISSUE #4091: DECRQM support
DECRQM:
Reports the state of ANSI and DEC private modes back to the application (modes from SM/RM and DECSET/DECRST).

For more advanced termlibs it is crucial to be able to request feature state from the terminal. xterm.js currently lacks support for DECRQM.
This is more and more important for app side, as more features are added to the modes list. Without DECRQM applications have no means to determine at runtime, if a certain feature is supported by the underlying terminal.

jerch:
Shall fix #4091.

### DIFF
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -2270,7 +2270,7 @@ export class InputHandler extends Disposable implements IInputHandler {
    * For modes not understood xterm.js always returns `notRecognized`. In general this means,
    * that a certain operation mode is not implemented and cannot be used.
    *
-   * Modes changing the active terminal buffer (47, 1047, 1048, 1049) are not subqueried
+   * Modes changing the active terminal buffer (47, 1047, 1049) are not subqueried
    * and only report, whether the alternate buffer is set.
    *
    * Mouse encodings and mouse protocols are handled mutual exclusive,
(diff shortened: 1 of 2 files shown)


src/common/InputHandler.ts
```diff
@@ -2245,6 +2247,66 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+    const f = (m: number, v: boolean | number): void => this._coreService.triggerDataEvent(`${C0.ESC}[?${m};${2 - +v}$y`);
+    const p = params.params[0];
+
+    if (ansi) {
+      switch (p) {
+        case 2:  f(p, -1); break;
+        case 4:  f(p, this._coreService.modes.insertMode); break;
+        case 12: f(p, -2); break;
+        case 20: f(p, -2); break;
+        default: f(p, 2);
```
Tyriar, line 2273: The `2 - +v` thing is a little confusing, how about instead creating an `fb` for boolean and `fi` for ints? A comment and/or an inline `const enum` would also work to help the reader.

jerch: Guess I should not try too hard to get small webpack output :smile_cat: (Since the code is not hot later on, I went the small path...)

Anyway this is still in the making, seems we lack several symbols in SM/RM that are interchangeable from DECSET/DECRST. Not sure yet if I gonna fix them here as well, prolly should be done in a follow up PR to not clutter things here too much.

jerch:
@Tyriar I changed 2 things in DECSET and SM:
- SM 20 - newline mode: we fully support that from the internals, but it was commented out, dunno why, re-enabled it
- DECSET 12 - blinking cursor: In the past the renderer were not able to handle that, but now all do, thus I re-enabled it.

Those 2 changes should not break anything, but you might want to revisit them and if they got disabled for other reasons in the first place.

Ready for review.

src/common/InputHandler.ts
jerch: Adopted the inline enum idea and added and explicit boolean converter.

Tyriar (approved):
👍 nice

PR #4098, opened 2022-09-03
protection flag support

### LINKED ISSUE #3651: Full support for DECSED,DECSEL,DECSCA, etc for xtermjs
Using xtermjs in VSCode to an application which respects VT220 protection sequences. DECSED sequence erases the complete screen including the protected fields. This is incorrect behaviour and currently means our developers are unable to test within the IDE. 

Protected fields are normally protected from DESCSED erasure to support behaviour like clearing input data or updating areas without having to redraw the whole screen (maybe a throwback to 75 baud accousticly coupled modems?!).

Xtermjs InputHandler.ts currentlly says "Protection attributes are not supported". 

Issue #2358 was closed on the (false) premise these sequences are unused. Any chance these could be implemented?

jerch:
Fixes #3651.

TODO:
- [x] some unit tests
- [x] DECRQSS report
- [x] consolidate DECRQSS

@Tyriar Rewrote the PR, this time without additional work in `InputHandler.print`, so this should not impact normal print data flow at all. :muscle: 

Sidenote: Due to changing DECRQSS into an InputHandler function with some inline definitions I could save >3kB on the package.

Edit: Also fixed some feature docs, thus the website should be re-deployed at some point (still on v4.14).

### DIFF
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -3243,7 +3243,8 @@ export class InputHandler extends Disposable implements IInputHandler {
     if (data === '"q') return f(`P1$r${this._curAttrData.isProtected() ? 1 : 0}"q`);
     if (data === '"p') return f(`P1$r61;1"p`);
     if (data === 'r') return f(`P1$r${b.scrollTop + 1};${b.scrollBottom + 1}r`);
-    if (data === 'm') return f(`P1$r0m`);  // FIXME: report real settings instead of 0m
+    // FIXME: report real SGR settings instead of 0m
+    if (data === 'm') return f(`P1$r0m`);
     if (data === ' q') return f(`P1$r${STYLES[opts.cursorStyle] - (opts.cursorBlink ? 1 : 0)} q`);
     return f(`P0$r`);
   }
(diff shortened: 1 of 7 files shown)


src/common/InputHandler.ts
```diff
@@ -1233,9 +1140,9 @@ export class InputHandler extends Disposable implements IInputHandler {
    * The cell gets replaced with the eraseChar of the terminal and the isWrapped property is set to false.
    * @param y row index
    */
-  private _resetBufferLine(y: number): void {
+  private _resetBufferLine(y: number, protect: boolean = false): void {
     const line = this._activeBuffer.lines.get(this._activeBuffer.ybase + y)!;
-    line.fill(this._activeBuffer.getNullCell(this._eraseAttrData()));
+    line.fill(this._activeBuffer.getNullCell(this._eraseAttrData()), protect);
```
Tyriar, line 1145: I was a little confused reading the PR initially because it seemed like `protect` here was a cell's protection attribute being applied to the entire line. Would these be better as `respectProtect`? Also `@param` docs here would help

Tyriar (changes requested):
Haven't tested but looks good other than my clean up comment

src/common/InputHandler.ts
jerch: Changed.

Tyriar (approved):

Issue #4060, opened 2022-08-23
Dotted underline looks more like dashed


Tyriar:
https://github.com/microsoft/vscode/issues/158872

- open a terminal and enter `echo -e "\x1b[4:4m"  # dotted underline`

![image](https://user-images.githubusercontent.com/6461412/186132008-617d2569-8a2d-4735-848e-c3f5830bb1a4.png)

Note that it's not really dots, but lines of various length

silamon:
Notice how the dotted line is correctly drawn in the middle under a character. At the start and the end of the character it gets connected making it look more like a dashed line. Given the screenshot above, we should look into drawing a dotted line from "-" to "e" instead of drawing a dotted line for each character. 

For the canvas renderer, we should adopt the _drawForeground method a little bit more like the _drawBackground.

jerch:
This is imho an almost similar problem to the curly underline - we get stitching artefacts due to single cell rendering. If we want to keep the single cell rendering approach because it is simpler to achieve, we might have to create a perfect tiling function for different resolutions to level out the artefacts (pretty much in the sense of Escher Tiling, thus with perfect continuation). I am not sure, if thats possible at all, as the integer rounding always introduces tiny offsets.

Tyriar:
Thinking about this and the fact that I want xterm.js rendering to be as great as possible, I think we should support multiple glyphs depending on the x-axis when necessary. For the dotted line specifically, say we had a dot of size `window.devicePixelRatio` with the same sized gap in-between, so there would be at most 2 possible variants for every glyph:

```
Even:
|* * * * |

Odd:
|* * * * *|
| * * * * |
```

To store this as a unique glyph we need to make the glyph key different, this could be done by asking what underline variant it is and then stored as a number, maybe capped at 4 (2 bits) or 8 variants (3 bits)?

This would allow us make a non-repeated curvy line and not restrict us to a single cell for both the up and down parts of the wave, if that ends up looking better of course.

jerch:
@Tyriar Wouldn't this bloat the glyph cache by far? For me the underline cases are rarely occurring edge cases, which might not justify pulling them through the whole cache logic?

Wouldn't an explicit composing step for underlines remove all the cell clipping issues? Something like that:
- pre-render underline styles for a whole line (lazy, we'd have 4 lines "glyphs" for line, dotted, dashed and curly * colors requested)
- on underlined chars: grab subarea of the requested line glyph for all subsequent chars, clip into line position
[...]

PR #4100, opened 2022-09-04
Improve dotted line for canvas

silamon:
Part of #4060.

### DIFF
--- a/addons/xterm-addon-canvas/src/BaseRenderLayer.ts
+++ b/addons/xterm-addon-canvas/src/BaseRenderLayer.ts
@@ -82,11 +82,11 @@ export abstract class BaseRenderLayer implements IRenderLayer {
     }
   }
 
-  public onOptionsChanged(): void {}
-  public onBlur(): void {}
-  public onFocus(): void {}
-  public onCursorMove(): void {}
-  public onGridChanged(startRow: number, endRow: number): void {}
+  public onOptionsChanged(): void { }
+  public onBlur(): void { }
+  public onFocus(): void { }
+  public onCursorMove(): void { }
+  public onGridChanged(startRow: number, endRow: number): void { }
 
   public onSelectionChanged(start: [number, number] | undefined, end: [number, number] | undefined, columnSelectMode: boolean = false): void {
     this._selectionStart = start;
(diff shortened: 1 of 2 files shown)


addons/xterm-addon-canvas/src/TextRenderLayer.ts
```diff
@@ -230,79 +230,145 @@ export class TextRenderLayer extends BaseRenderLayer {
     ctx.restore();
   }
 
+  /**
+   * Draws the foreground for a specified range of columns. Tries to batch adjacent
+   * cells of the same color together to reduce draw calls.
+   */
   private _drawForeground(firstRow: number, lastRow: number): void {
+    const cols = this._bufferService.cols;
```
Tyriar, line 238: Are these changes all to do with reducing draw calls? I worry it's getting a bit complicated when canvas should be used as a fallback when webgl isn't available.

addons/xterm-addon-canvas/src/BaseRenderLayer.ts
```diff
@@ -232,15 +232,12 @@ export abstract class BaseRenderLayer implements IRenderLayer {
...
     const xLeft = x * this._scaledCellWidth;
     const yMid = (y + 1) * this._scaledCellHeight - window.devicePixelRatio - 1;
     this._ctx.moveTo(xLeft, yMid);
-    for (let xOffset = 0; xOffset < width; xOffset++) {
-      // const xLeft = x * this._scaledCellWidth;
-      const xRight = (x + width + xOffset) * this._scaledCellWidth;
-      this._ctx.lineTo(xRight, yMid);
-    }
+    const xRight = (x + width) * this._scaledCellWidth;
+    this._ctx.lineTo(xRight, yMid);
```
Tyriar, line 240: How does this look for wide characters? I was a little confused when working with setLineDash as it was stretching it unexpectedly when drawing across multiple cells.

silamon: before 
image

after
image

addons/xterm-addon-canvas/src/TextRenderLayer.ts
silamon: Any suggestion on how to improve this code or if it's worth trying to get this pull request done? All code changes are related to the initial issue and the idea of this pull request is to copy the idea of the drawBackground function just above it.

addons/xterm-addon-canvas/src/BaseRenderLayer.ts
the-j0k3r: 4:1m straight The lowercase g is now touching the underline, before there was a very small gap which looks better. These underline things always going to headbutt with lower case pgy and friends, not ideal when they overlap.

Tyriar (commented):
Unfortunately too many things have changed in this area so there are too many conflicts. I tried to switch the `[dpr * 2, dpr]` on master which I think is the essence of this PR, but it ended up with this problem:

image

Do we need to more manually draw the pixels for dotted to ensure it looks correct and connects properly across cells at all zoom levels?

silamon:
> Unfortunately too many things have changed in this area so there are too many conflicts. I tried to switch the `[dpr * 2, dpr]` on master which I think is the essence of this PR, but it ended up with this problem:
> 
> image
> 
> Do we need to more manually draw the pixels for dotted to ensure it looks correct and connects properly across cells at all zoom levels?

That problem was also solved in this PR by drawing one line from "a" to "h" instead of for every character.

jerch:
On a sidenote - I had a similar issue in the image addon for pixel perfect alignment. My solution there is quite simple - the addon overdraws by flooring to the left and ceiling to the right side, which avoid striping artifacts. I think a similar technique can be used here, where the on/off state  of dots only depend on total pixel progression on the viewport?

Issue #3878, opened 2022-06-29
Previous output leeches through to nvim w/ WebGL Addon enabled


AlexMordred:
Previous commands' output leeches through to nvim. The old output is only visible on the same lines you see the nvim prompt text - see screenshot. Doesn't happen with other editors, tried `vi`, `nano`, `micro`. Only happens when using the WebGL Addon. 

## Details
- Browser and browser version: Firefox 101.0.1
- OS version: Linux 5.15.49-1-MANJARO
- xterm.js version: latest

### Steps to reproduce

1. Run some commands until at least half of your screen is filled with output bottom to top
2. Run `nvim`

![image](https://user-images.githubusercontent.com/8043551/176419989-2e8699e3-507a-42ed-a27e-972ccf9dad24.png)

Tyriar:
I've spotted something similar recently as well

Tyriar:
These dots are what I've seen sometimes:

<img width="362" alt="Screen Shot 2022-07-10 at 2 48 01 pm" src="https://user-images.githubusercontent.com/2193314/178163196-6723d0c8-5b7f-4e6a-aa3e-c18e72027c9b.png">

I don't know the repro and I can't repro with nvim on macOS or Windows.

AlexMordred:
@Tyriar I don't know much about terminals, but mine is a default Manjaro Linux terminal and yours looks customized (`.bashrc`?), I guess that makes a difference too.

Tyriar:
@AlexMordred good point, I'll check without the fancy prompt

Tyriar:
We're a bit too busy to action this issue in July

Tyriar:
I don't have a Linux machine but I tested on the latest `master` of xterm.js on macOS in Firefox and Edge and seems to work fine. Some more information that could help:

- `echo $PS1` result
- Set log level to debug and provide the logs when reproducing

AlexMordred:
```
$ echo $PS1
\[\033[01;32m\][\u@\h\[\033[01;37m\] \W\[\033[01;32m\]]\$\[\033[00m\]
```

Not sure I understand the second part

Tyriar:
@AlexMordred you an turn debug logs on which get send to the devtools console by running `term.options.logLevel = 'debug'`

JasonXJ:
[...]

PR #4101, opened 2022-09-05
WebglRenderer always clear _glyphRenderer if _model is cleared

JasonXJ:
This fix the issue mentioned by OP and [this comment](https://github.com/xtermjs/xterm.js/issues/3878#issuecomment-1221782722) in #3878.

 It can happen when neovim sets some colors and then deletes some characters, and this is what happens inside the `WebglRenderer`:
* `setColors()` is called, and it clears `this._model` but not `this._glyphRenderer`.
* `_updateModel()` is called to update `this._model` to match the terminal buffer. Because `this._model` was cleared earlier, the code [here](https://github.com/xtermjs/xterm.js/blob/93bc9bf1525f663fda72c7637ae0366c2e5a9868/addons/xterm-addon-webgl/src/WebglRenderer.ts#L420) thinks that the cell is up-to-date and there is no need to update `this._glyphRenderer`.

### DIFF
--- a/addons/xterm-addon-webgl/src/GlyphRenderer.ts
+++ b/addons/xterm-addon-webgl/src/GlyphRenderer.ts
@@ -238,21 +238,24 @@ export class GlyphRenderer  extends Disposable {
     // a_cellpos only changes on resize
   }
 
-  public clear(force?: boolean): void {
+  public clear(): void {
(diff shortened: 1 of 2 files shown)


Tyriar (approved):
Looks good, thanks!

PR #4102, opened 2022-09-06
Webgl cursor renderer makes sure a char is not drawn outside of the cell

JasonXJ:
This fixes the rendering artefact reported [here](https://github.com/xtermjs/xterm.js/issues/3878#issuecomment-1221974475) in #3878.

### DIFF
--- a/addons/xterm-addon-webgl/src/renderLayer/BaseRenderLayer.ts
+++ b/addons/xterm-addon-webgl/src/renderLayer/BaseRenderLayer.ts
@@ -226,7 +226,7 @@ export abstract class BaseRenderLayer implements IRenderLayer {
   protected _fillCharTrueColor(terminal: Terminal, cell: CellData, x: number, y: number): void {
     this._ctx.font = this._getFont(terminal, false, false);
     this._ctx.textBaseline = TEXT_BASELINE;
-    this._clipRow(terminal, y);
+    this._clipCell(x, y, cell.getWidth());
     this._ctx.fillText(
       cell.getChars(),
       x * this._scaledCellWidth + this._scaledCharLeft,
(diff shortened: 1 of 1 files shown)


Tyriar (commented):
This shouldn't actually do anything as only cursor and link render layers are used in the webgl renderer. All text drawing in the webgl renderer is done in `WebglCharAtlas` and `GlyphRenderer`.

JasonXJ:
`CursorRenderLayer` draws the char with block cursor. So let's look at its `_render()` method ([code link](https://github.com/xtermjs/xterm.js/blob/93bc9bf1525f663fda72c7637ae0366c2e5a9868/addons/xterm-addon-webgl/src/renderLayer/CursorRenderLayer.ts#L122)). On line 181, we clear the cursor cell, and at line 185, we call `_renderBockCursor()`, which renders the char. The char can be extra wide, so if we don't clip it, we will leave something on the screen the next time you move the cursor left (because the `_clearCursor()` only clear the single cell).

Tyriar (approved):
Ah I see, I should have read the linked comment properly. I can't actually reproduce this when going back to the normal buffer, but I can reproduce the artifact issue when zoomed in, but I can see how it could happen. It seems to be another case of rounding problems that get covered when devicePixelRatio < 1:

https://github.com/silamon/xterm.js/blob/e8e2f16cffe89e9a08899a2a37c2a4cfa580cca9/addons/xterm-addon-webgl/src/renderLayer/CursorRenderLayer.ts#L197-L200

I guess your fix is actually better here as a black cursor with a wide character rendering an inverted foreground color would probably look better if the part beyond the cursor was the un-inverted color.

Thanks for looking into this!

PR #4109, opened 2022-09-09
Copy extended attributes to dest in reflow

### LINKED ISSUE #4108: Extended attributes can be messed up after reflow, causing stretched canvas on resize
Repro in: https://github.com/microsoft/vscode/issues/157585#issuecomment-1242356284

image

Tyriar:
Fixes #4108

This would throw before:

Recording 2022-09-09 at 12 29 36

### DIFF
--- a/src/common/buffer/BufferLine.ts
+++ b/src/common/buffer/BufferLine.ts
@@ -60,7 +60,7 @@ const w: { startIndex: number } = {
 export class BufferLine implements IBufferLine {
   protected _data: Uint32Array;
   protected _combined: {[index: number]: string} = {};
-  protected _extendedAttrs: {[index: number]: IExtendedAttrs} = {};
+  protected _extendedAttrs: {[index: number]: IExtendedAttrs | undefined} = {};
   public length: number;
 
   constructor(cols: number, fillCellData?: ICellData, public isWrapped: boolean = false) {
@@ -181,7 +181,7 @@ export class BufferLine implements IBufferLine {
       cell.combinedData = this._combined[index];
     }
     if (cell.bg & BgFlags.HAS_EXTENDED) {
-      cell.extended = this._extendedAttrs[index];
+      cell.extended = this._extendedAttrs[index]!;
     }
     return cell;
   }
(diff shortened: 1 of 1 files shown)


jerch (approved):
LGTM.

On a perf sidenote - maybe we should remove the inner for loop above by copying CONTENT, FG and BG the same way explicitly, I think this would run much faster, as loop unrolling in not yet a thing JITs. And it is basically the same with `_combined` right below. I have not tested that, so this is just a guess maybe worth trying.

Tyriar:
Do you mean instead of this:

```ts
for (let cell = 0; cell < length; cell++) {
        for (let i = 0; i < CELL_SIZE; i++) {
```

Something like this?

```ts
for (let cell = 0; cell < length * CELL_SIZE; cell++) {
```

I think reflow is relatively fast already considering the amount of work it needs to do already, it might end up bad if you have a lot of content in a bunch of terminals though. I'm also a little scared to touch reflow tbh as I always thought it was a bit fragile.

jerch:
> Something like this?
> 
> ```ts
> for (let cell = 0; cell < length * CELL_SIZE; cell++) {
> ```
> 
> I think reflow is relatively fast already considering the amount of work it needs to do already, it might end up bad if you have a lot of content in a bunch of terminals though. I'm also a little scared to touch reflow tbh as I always thought it was a bit fragile.

Ah nope, messing around with different steppinh might be abit too risky imho, I thought more like that (untested):
```ts
      for (let cell = 0; cell < length; cell++) {
        const src = (srcCol + cell) * CELL_SIZE;
        const dst = (destCol + cell) * CELL_SIZE;
        this._data[dst + Cell.CONTENT] = srcData[src + Cell.CONTENT];
        this._data[dst + Cell.FG] = srcData[src + Cell.FG];
        this._data[dst + Cell.BG] = srcData[src + Cell.BG];
        if (srcData[src + Cell.CONTENT] & Content.IS_COMBINED_MASK) {
          this._combined[destCol + cell] = src._combined[srcCol + cell];
        }
        if (srcData[src + Cell.BG] & BgFlags.HAS_EXTENDED) {
          this._extendedAttrs[destCol + cell] = src._extendedAttrs[srcCol + cell];
        }
      }
      // + same way to reversed
```

Tyriar:
@jerch when reducing the size of the buffer, I think we could actually change this:

https://github.com/xtermjs/xterm.js/blob/f9db65c52e4702cf906c14cda1c640342a7b6dca/src/common/buffer/BufferLine.ts#L335-L336

To:

```ts
this._data = this._data.subarray(0, cols * CELL_SIZE);
```

That will keep the same ArrayBuffer, but use a different view of it.

Tyriar:
https://github.com/xtermjs/xterm.js/issues/4112

PR #4115, opened 2022-09-13
some buffer handling optimizations

### LINKED ISSUE #4112: Avoid creating new ArrayBuffers when reducing the size of the buffer
See https://github.com/xtermjs/xterm.js/pull/4109#issuecomment-1243690016

I think we could actually change this:

https://github.com/xtermjs/xterm.js/blob/f9db65c52e4702cf906c14cda1c640342a7b6dca/src/common/buffer/BufferLine.ts#L335-L336

To:

```ts
this._data = this._data.subarray(0, cols * CELL_SIZE);
```

That will keep the same ArrayBuffer, but use a different view of it. We wouldn't free the memory but maybe we should only do that either when the view is < 1/2 the size of the buffer, or defer the clean up to an idle timeout? It doesn't make sense to do so many allocations if they're being thrown away immediately after several resizes.

We could do something similar here:

https://github.com/xtermjs/xterm.js/blob/f9db65c52e4702cf906c14cda1c640342a7b6dca/src/common/buffer/BufferLine.ts#L347-L348

jerch:
Should fix #4112.

TODO:
- [x] alloc/copy-free shrinking
- [x] alloc/copy-free enlarging within bytelength
- [x] use `DebouncedIdleTask` to defer array buffer resize after shrinking
- [x] ~optimize `copyCellsFrom`, check for correct _combined and _extendedAttr handling~ not critical at all
- [x] several unit tests


Tyriar:
> shim somehow

Alternatively we can just not do this optimization for xterm-headless, adding it later on if/when `requestIdleCallback` is added.

src/common/buffer/BufferLine.ts
```diff
@@ -42,6 +42,8 @@ const w: { startIndex: number } = {
   startIndex: 0
 };
 
+const EMPTY_DATA = new Uint32Array(0);
```
Tyriar, line 45: It's probably not worth having this actually, I think the min dimensions are 2x1 so this shouldn't be possible.

src/common/buffer/BufferLine.ts
```diff
@@ -339,35 +341,47 @@ export class BufferLine implements IBufferLine {
     if (cols === this.length) {
       return;
     }
+    const fourByteCells = cols * CELL_SIZE;
```
Tyriar, line 344: `fourByteCells` -> `idealBufferSize`?

src/common/buffer/BufferLine.ts
```diff
@@ -339,35 +341,47 @@ export class BufferLine implements IBufferLine {
...
     if (cols > this.length) {
-      const data = new Uint32Array(cols * CELL_SIZE);
-      if (this.length) {
-        if (cols * CELL_SIZE < this._data.length) {
-          data.set(this._data.subarray(0, cols * CELL_SIZE));
-        } else {
+      if (this._data.buffer.byteLength >= fourByteCells * 4) {
```
Tyriar, line 346: Might be wrong, but shouldn't this be:

```ts
if (this._data.buffer.byteLength >= fourByteCells) {
```

src/common/buffer/BufferLine.ts
jerch: Hmm, I think there was a reason in the first place for that `new Uint32Array(0)`, but cant remember why. (Was it early init with cols=0?)

src/common/buffer/BufferLine.ts
jerch: Ah nope, thats just an itermediate number to avoid recalc all the time. `fourByteCells` here means "number of 4-bytes".

src/common/buffer/BufferLine.ts
jerch: See above, thats the number of 4-bytes (uint32), thus needs * 4 to get bytes.

src/common/buffer/BufferLine.ts
Tyriar: I think the restriction on dimensions was added after BufferLine was created

jerch:
About lazy resizing of the underlying array buffer - see https://github.com/xtermjs/xterm.js/pull/4144

@Tyriar I currently dont like the idea to place lines * `DebouncedIdleTask` objects on the terminal buffer, also placing a cleanup task for every single line might grow really big (easily going into several MBs for scrollback >1000 just for registering the cleanup work). Instead I will try to find a way to batch the cleanup from `Buffer` or `CircularList` side (maybe 1k steps), which only needs 1 `DebouncedIdleTask` object. This might be a rare use case for a `WeakSet`, will see if that comes handy to hold refs to lines, that need a cleanup.

src/common/buffer/BufferLine.ts
jerch: Removed the EMPTY_DATA branch (it is anyway fully covered by the other branch, just abit slower).

Tyriar:
@jerch :+1:

> This might be a rare use case for a WeakSet, will see if that comes handy to hold refs to lines, that need a cleanup.

I always think something will work for `WeakSet` and it doesn't end up working out. I tried recently but the fact that the keys aren't enumerable and they cannot be numbers meant It wasn't appropriate ☹️

jerch:
@Tyriar Yeah WeakMap and WeakSet are very awkward and mostly not of much use. Will see if its useful here (have currently benchmark issues, seems master has a bad perf issue, my typical `ls -lR /usr` benchmark takes like 4 times longer since I merged master bad here, still investigating...)

src/common/buffer/Buffer.ts
```diff
@@ -243,13 +248,35 @@ export class Buffer implements IBuffer {
...
       }
     }
 
     this._cols = newCols;
     this._rows = newRows;
+
+    if (needsCleanup) {
+      this._memoryCleanupTask.set(() => this._cleanupMemory());
+    } else {
+      // FIXME: DebouncedIdleTask has no clear method?
```
jerch, line 263: @Tyriar Imho `DebouncedIdleTask` should expose a `clear` method itself?

jerch (commented):
@Tyriar Plz have a look at the way I use `DebouncedIdleTask` on `Buffer`. Resizing is quite fast that way (~120ms), furthermore it splits its memory alloc workload into smaller batches, that dont interfere too much with mainthread latency (its at ~13ms on my machine for 5k lines, can do smaller batches if needed). Also the batching should be resilient to in-between resizes altering bufferlines again before the memory got fully fixed (well, still needs some unit tests).

NB: I think there is still a catch in `DebouncedIdleTask`, as I get `Uncaught ReferenceError: performance is not defined` on nodejs. Is that unit tested to work on nodejs?

src/common/buffer/Buffer.ts
Tyriar: This isn't exactly what I had in mind when I built that util class, it was meant as a nicer way to just defer a single task instead of dealing with a queue. The current way I'd recommend to implement this is with a regular `IdleTaskQueue` and splitting the line ranges into chunks but I can definitely understand it's a little cumbersome to do it like that.

How about we make a change to all the task callbacks to return a `boolean | void`, if true it will treat that task as not finished yet and call it again? So we wouldn't increment `_i` here if true was returned:

https://github.com/xtermjs/xterm.js/blob/a351f5758a5126308b90d60b604b528462f6f051/src/common/TaskQueue.ts#L71

Then in `_cleanupMemory` you could do this to terminate it:

```ts
if (!needsCleanup) {
  return false;
}
```

Tyriar (changes requested):
Sorry about the delay, didn't realize this one was waiting on me 😅

src/common/buffer/Buffer.ts
jerch: Oh well - true using the task queue directly here is def. the better idea. Also the return value idea sounds good to me. Will see if I can get things less hacky with those changes.

jerch:
@Tyriar Saw that you use `performance.now()` in several places in `TaskQueue.ts`. I suggest to replace it with `Date.now()`, as it is typically 4-6 times faster than `performance.now()` [*]. Imho both give enough time resolution for the tracked tasks, only downside of `Date.now()` its dependence on the wall clock, thus it would shift on machine clock changes (can be mitigated by some lower bounds, like always accounting at least 1msec).

---
[*] Observed the bad runtime of `performance.now` in my fifo pipe buffer impl in [browser-fakepty](https://github.com/jerch/browser-fakepty/blob/5410f313bcf8c38f3560e2d0384e03abe4dadd2f/src/Clock.ts), where `performance.now` throttled pipe throughput to ~200 MB/s, while it happily runs at ~1GB/s with `Date.now`.

jerch:
@Tyriar Ready for review.

Note that I changed `performance.now` to `Date.now`, which also fixes `IdleTaskQueue` for nodejs.

src/common/buffer/Buffer.ts
```diff
@@ -242,13 +247,42 @@ export class Buffer implements IBuffer {
...
+
+  private _memoryCleanupQueue = new IdleTaskQueue();
+
+  private _batchedMemoryCleanup(): boolean {
+    let counted = 0;
+    for (let i = 0; i < this.lines.length; i++) {
+      counted += this.lines.get(i)!.cleanupMemory();
+      // throttle to 5k lines at once and
+      // return true to indicate, that the task is not finished yet
+      if (counted > 5000) {
```
Tyriar, line 274: 5000 seems very high? Since we're using the task queue this could go much lower and it would then be able to respect the task callback deadline better. I'd guess something more like 50 or 100 would be a better number

src/common/buffer/Buffer.ts
```diff
@@ -242,13 +247,42 @@ export class Buffer implements IBuffer {
...
+      }
+    }
+    return false;
+  }
+
+  private _forceMemoryCleanup(): void {
+    this._memoryCleanupQueue.clear();
+    for (let i = 0; i < this.lines.length; i++) {
+      this.lines.get(i)!.cleanupMemory();
+    }
```
Tyriar, line 285: Could this just be this?

```suggestion
    this._memoryCleanupQueue.flush();
```

Tyriar (changes requested):

src/common/buffer/Buffer.ts
jerch: Sure, can set that to 100.

The only issue with a rather lowish value is - since the logic re-scans already cleaned lines (needed to be resilient against multiple resizes in a row and in between buffer shifts), this somewhat adds additional lookup time on top. Did not test with 100 yet, for 1000 vs 5000 the difference the repeated scans added ~200 ms for 100k scrollback (absolute runtime was ~550ms (1000) vs. ~350ms (5000) for the overall task time summed up). This re-scan effect on runtime will multiply the lower the batch size is. Will see where it ends up with 100 :smile_cat:

src/common/buffer/Buffer.ts
jerch: Or just remove it completely (not used by any logic atm)? The idea here was to provide a sync way to force cleanup the memory without the rescan overhead of multiple `_batchedMemoryCleanup` calls.

src/common/buffer/Buffer.ts
jerch: @Tyriar Changed the batch size to 500, as it shows the best trade off for bigger scrollbacks on a single terminal. If you think thats still too high, maybe test it in vscode with multiple tabbed terminals at once? (Note, that the idle logic would still postpone the cleanup task for a terminal announcing the it late, if the event loop is already crowded.)

Tyriar: Not sure what hardware you're on but our average user's machine is pretty awful compared to mine so I always err on the side of doing what's better for lower end hardware.

> 5k: 330 ms, average invocation taking ~20 ms
> 100: 3400 ms spread over ~4s, average invocation taking ~4 ms

The main thing here is we just want to keep it within the bounds of the idle deadline, we should probably log a warning if it is exceeded to tweak things for better responsiveness. While the 5k does it much faster, it does so at the cost of a loss of framerate as each one takes 20ms (on your machine) which is sub 60 fps. My monitors support 240 fps so anything above 4ms could result in a drop in framerate, it turns out that even Chromium doesn't care at that level though and sometimes even allocates more than 16.6ms for an idle task.

The ideal case for these tasks in my opinion currently is around a maximum of 4ms, that way if the deadline we're given is small like 4ms it's happens successfully, and if it's larger like 16ms it will give 3-4 tasks in. If this is truly just clean up the actual time to complete doesn't matter so much as we just want it not to consume the extra memory in the long run.

In other words, we want to optimize for the minimal amount of time spent working beyond the idle deadline and the maximal amount of time filling the idle deadlines.

> Since the logic re-scans already cleaned lines (needed to be resilient against multiple resizes in a row and in between buffer shifts)

Can we cache where we were up to and clear the cache when another resize happens? I haven't tested this yet, but I think 100 rows is a good number, especially if we can do this caching.

jerch: > Not sure what hardware you're on but our average user's machine is pretty awful compared to mine...

Mediocre here for a desktop system - i7 haswell, was top notch years ago, but now it is more at the lower end. It still would outperform most tablets/phones, so I totally understand your concerns.

> Chromium doesn't care at that level though and sometimes even allocates more than 16.6ms for an idle task.

Yes I already spotted that while trying to get _reliable_ profiling data - it not possible, chrome sometimes just overshoots, no clue why. The problem is further complicated by the fact, that the idle detection is usually document/JS context global in browsers, thus might schedule too many idle callbacks from different page components, that again add up to frame drops - while I only tested it in the demo holding a single terminal instance. Sadly info about the idle detection done by the browser is somewhat spare, so I cannot really tell, which constellation will run into issues again.

> My monitors support 240 fps so anything above 4ms could result in a drop in framerate

Imho all main browsers would cap fps at 60 regardless of hardware support? Though I dont know that for sure or whether its still the case...

> If this is truly just clean up the actual time to complete doesn't matter so much as we just want it not to consume the extra memory in the long run.

Yes the cleanup here is just about wasting memory over a longer time. Nothing else should be affected.

> Can we cache where we were up to and clear the cache when another resize happens? I haven't tested this yet, but I think 100 rows is a good number, especially if we can do this caching.

Caching lines already handled was my first approach (thus the WeakMap notion above), but turned out to be even more expensive due to map hashing/resolving. I think a slightly more advanced logic might avoid all those re-scan refforts:
- safe last adjusted line number
- continue from last adjusted one
- when done, recheck all again to spot in between buffer shifts
- do this until exhausted

This would work pretty much the same now, but not re-scan previous lines everytime. Instead it only does a final scan, before it is really finished. With that imho lower lines handled at once should not show that exp growing anymore.

Tyriar: > Yes I already spotted that while trying to get reliable profiling data - it not possible, chrome sometimes just overshoots, no clue why. The problem is further complicated by the fact, that the idle detection is usually document/JS context global in browsers, thus might schedule too many idle callbacks from different page components, that again add up to frame drops - while I only tested it in the demo holding a single terminal instance. Sadly info about the idle detection done by the browser is somewhat spare, so I cannot really tell, which constellation will run into issues again.

I got the impression Chromium would give more time if there was less going on, prioritizing finishing tasks up over slight frame drops. Similar to `requestAnimationFrame`, I think it's best to trust the browser in this respect.

> Imho all main browsers would cap fps at 60 regardless of hardware support? Though I dont know that for sure or whether its still the case...

Electron doesn't:

image

I checked Edge and it looks like it's capped by default, I think you can unlock it via CLI args.

jerch:
@Tyriar Oh well - should we still change the rescan logic as described here?

> I think a slightly more advanced logic might avoid all those re-scan refforts:
>
>   - safe last adjusted line number
>   - continue from last adjusted one
>   - when done, recheck all again to spot in between buffer shifts
>   - do this until exhausted
>
> This would work pretty much the same as now, but not re-scan previous lines everytime. Instead it only does a final scan, before it is really finished. With that imho lower lines handled at once should not show that exp growing anymore.

Would be a more "heuristical" approach avoiding the bad exp rescan runtime for _n_ scrollback lines ...

Tyriar:
If you think it's worth it, the PR is way better than the current state already 🙂

Tyriar (approved):
Nice 👍

PR #4137, opened 2022-09-24
Optimize contexts for reading frequently

Tyriar:
This was a recommendation from Edge.
Docs here: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently


Tyriar:
First draw to char atlas before took ~50ms:

Screen Shot 2022-09-24 at 11 02 31 am

After:
Screen Shot 2022-09-24 at 11 03 33 am

PR #4144, opened 2022-09-24
Add PriorityTaskQueue

Tyriar:
This acts as a fallback when idle callbacks aren't supported, we may also want to use it for other tasks (eg. write buffer?)


jerch:
@Tyriar Can I use this for the BufferLine resizing? If so, which is the straight forward usage pattern? `DebouncedIdleTask.set`?

Tyriar:
@jerch yes you could use it for that:

```ts
private _resizeIdleTask: DebouncedIdleTask = new DebouncedIdleTask();
resize(cols, rows) {
  // Check if there is room in the buffer or we need a new one immediately
  if (cols * CELL_SIZE * 4 > this._data.buffer.byteLength) {
    // Discard any pending resizes
    this._resizeIdleTask.clear();
    // Resize immediately
    this._data = ...
  } else {
    // Swap out the current resize task
    this._resizeIdleTask.set(() => cleanUp());
  }
}
```

This clean up stuff could be encapsulated within a class for simplicity (`DynamicUint8Array`?), basically a stripped down version of my `ArrayBufferList` exploration https://github.com/Tyriar/xterm.js/tree/array_buffer_list

PR #4145, opened 2022-09-24
Optimize critical I/O path for input latency

Tyriar:
This reduced input latency from about 10ms -> 4ms on my mac (measurement based on devtools tasks).

Before server microtask:

Screen Shot 2022-09-24 at 4 23 02 pm

After server microtask:

Screen Shot 2022-09-24 at 4 22 38 pm

After server microtask and write buffer microtask:

Screen Shot 2022-09-24 at 4 26 41 pm

### DIFF
--- a/src/common/input/WriteBuffer.ts
+++ b/src/common/input/WriteBuffer.ts
@@ -105,7 +105,7 @@ export class WriteBuffer {
     // schedule chunk processing for next event loop run
     if (!this._writeBuffer.length) {
       this._bufferOffset = 0;
-      setTimeout(() => this._innerWrite());
+      queueMicrotask(() => this._innerWrite());
     }
 
     this._pendingData += data.length;
@@ -217,7 +217,7 @@ export class WriteBuffer {
         this._callbacks = this._callbacks.slice(this._bufferOffset);
         this._bufferOffset = 0;
       }
-      setTimeout(() => this._innerWrite());
+      queueMicrotask(() => this._innerWrite());
     } else {
       this._writeBuffer.length = 0;
       this._callbacks.length = 0;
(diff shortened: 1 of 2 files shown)


src/common/input/WriteBuffer.ts
```diff
@@ -105,7 +105,7 @@ export class WriteBuffer {
     // schedule chunk processing for next event loop run
     if (!this._writeBuffer.length) {
       this._bufferOffset = 0;
-      setTimeout(() => this._innerWrite());
+      queueMicrotask(() => this._innerWrite());
```
jerch, line 108: This should be safe with `queueMicrotask` - reason: this code path is only entered, if the write queue was empty before, which guarantees, that there happened a screen refresh before.

src/common/input/WriteBuffer.ts
```diff
@@ -217,7 +217,7 @@ export class WriteBuffer {
         this._callbacks = this._callbacks.slice(this._bufferOffset);
         this._bufferOffset = 0;
       }
-      setTimeout(() => this._innerWrite());
+      queueMicrotask(() => this._innerWrite());
```
jerch, line 220: Not so sure about `queueMicrotask` here - imho this will skip any screen refreshs in between, if the write queue never drops empty?

There are more edge cases linked to direct microtask usage - imho chrome and firefox prioritize them differently (dont remember the details atm).

jerch (commented):
@Tyriar I am not so sure about `queueMicrotask` usage here - see comments below.

src/common/input/WriteBuffer.ts
Tyriar: Opps. You're right, will fix

> There are more edge cases linked to direct microtask usage - imho chrome and firefox prioritize them differently (dont remember the details atm).

I didn't test in Firefox but this should only matter if queueMicrotask ends up slower than setTimeout which I think would be against the spec?

PR #4146, opened 2022-09-25
Use <details> tag for real-world uses

dstein64:
This update puts the real-world use examples in a `<details>` tag so that it must be manually expanded.

The idea is to simplify the `README.md` page, for which the example real-world uses currently takes a large portion.

Is this a change you'd be interested in merging? If so, please let me know if there are any updates I should make.

The updated version can be viewed at:
https://github.com/dstein64/xterm.js/blob/master/README.md#real-world-uses

real_world_uses

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,8 @@ Note that some APIs are marked *experimental*, these are added to enable experim
 ## Real-world uses
 Xterm.js is used in several world-class applications to provide great terminal experiences.
 
+<details><summary>Examples</summary>
+
 - [**SourceLair**](https://www.sourcelair.com/): In-browser IDE that provides its users with fully-featured Linux terminals based on xterm.js.
 - [**Microsoft Visual Studio Code**](http://code.visualstudio.com/): Modern, versatile, and powerful open source code editor that provides an integrated terminal based on xterm.js.
 - [**ttyd**](https://github.com/tsl0922/ttyd): A command-line tool for sharing terminal over the web, with fully-featured terminal emulation based on xterm.js.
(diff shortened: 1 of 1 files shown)


Tyriar (commented):
I think this would defeat a lot of the purpose of it being there as almost no one would expand and therefore we would get less PRs/engagement coming out way.

How about instead we just move it to the bottom, below releases/contributing and above license?

dstein64:
> "How about instead we just move it to the bottom, below releases/contributing and above license?"

Sure. I can make that update to this PR.

Would you be interested in either of the following as part of the update:

- Alphabetizing the list of examples
- Adding a `<details>` tag, like was originally proposed in this PR, but having it default to open, so that the list can optionally be collapsed

Tyriar:
> Alphabetizing the list of examples (with a corresponding update to the Note, to mention ordering instead of "Please add any new contributions to the end of the list only")

It was always "first come, first serve", so the earliest adopters are more prominently presented in the list. When alphabetizing, we would still want a select few at the top of the list, so the more important entries don't get buried and it gets a little complicated selecting those. Easiest it just to keep it as is

> Adding a <details> tag, like was originally proposed in this PR, but having it default to open, so that the list can optionally be collapsed

I don't think there's much point adding interactivity like this at the bottom of a readme as it would rarely get clicked.

dstein64:
> "Sure. I can make that update to this PR."

I made the update in a separate PR, moving the real-world uses section below releases/contributing and above license.
https://github.com/xtermjs/xterm.js/pull/4152

PR #4152, opened 2022-09-26
Move real-world uses section down

dstein64:
This PR moves the _Real-world uses_ section down, as a follow-up to discussion in PR #4146.


Tyriar (approved):
Thanks 🙂

PR #4159, opened 2022-09-28
Use microtask to flush the write buffer after input

### LINKED ISSUE #4158: serious perf issues on master
Master currently has serious perf issues (comparing `ls -lR /usr` on my machine, all with webgl renderer):
- master
grafik

- version 5.0
grafik

- version 4.19
grafik

All were measured after several runs to give JIT a chance to do its work. It seems the change causing this happened after 5.0.0.

Tyriar:
Fixes #4158

Previously this was 5000ms+ scripting:

Screen Shot 2022-09-28 at 9 05 58 am

After input:

Screen Shot 2022-09-28 at 9 07 45 am

### DIFF
--- a/src/common/input/WriteBuffer.ts
+++ b/src/common/input/WriteBuffer.ts
@@ -40,10 +40,15 @@ export class WriteBuffer {
   private _bufferOffset = 0;
   private _isSyncWriting = false;
   private _syncCalls = 0;
+  private _didUserInput = false;
   public get onWriteParsed(): IEvent<void> { return this._onWriteParsed.event; }
   private _onWriteParsed = new EventEmitter<void>();
 
   constructor(private _action: (data: string | Uint8Array, promiseResult?: boolean) => void | Promise<boolean>) { }
+
+  public handleUserInput(): void {
+    this._didUserInput = true;
+  }
 
   /**
    * @deprecated Unreliable, to be removed soon.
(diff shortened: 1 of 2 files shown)


src/common/input/WriteBuffer.ts
```diff
@@ -99,7 +104,15 @@ export class WriteBuffer {
     // schedule chunk processing for next event loop run
     if (!this._writeBuffer.length) {
       this._bufferOffset = 0;
-      queueMicrotask(() => this._innerWrite());
+      // If this is the first write call after the user has done some input,
+      // parse it immediately in an upcoming microtask to minimize reduce input,
+      // otherwise schedule for the next event
+      if (this._didUserInput) {
+        this._didUserInput = false;
+        queueMicrotask(() => this._innerWrite());
```
jerch, line 112: If latency is your main concern here - I think we can call `this._innerWrite()` directly here? (after pulling the last 3 lines of the method before the call)

Tyriar: Needed to duplicate some calls to make this happen, that's one of the reasons why it was microtask

jerch:
@Tyriar Yepp, the direct call works nice for me.

Also this PR restores scripting time back to old values, my devtools profiling is still screwed up:
grafik

Note this huge greyish portion and the overall runtime. Any idea what causes this? Also note that with devtools closed I see normal runtimes. My guess what be - there is another code change, that s*cks really bad on profiler performance, but gets not really esolved by the profiler (thus shown as grey). Sadly this currently makes the devtools profiler usable for me.

Tyriar:
I see that every now and then and it's not clear to me what it means. Does reloading the page or devtools help it go away? 🤔

jerch:
Oh well found the issue - `queueMicrotask` in `server.js` basically bypassed the intermediate buffering. Without the tiny buffer there, websocket create so many messages, that my browser profiler cannot keep up. Its basically a message congestion from too many tiny messages.
I think we need to keep setTimeout there, but maybe with a smaller limit? Like 2ms instead of 5?

jerch (approved):

Tyriar:
Ah ok, we should tweak that then. Shall we do a very similar change such that we fire the event following incoming data with queueMicrotask/immediate call?

jerch:
> Ah ok, we should tweak that then. Shall we do a very similar change such that we fire the event following incoming data with queueMicrotask/immediate call?

Yepp, should work there too :smiley_cat:

PR #4160, opened 2022-09-28
tweak buffering in server.js

jerch:
Part2 fix of #4158.


jerch:
@Tyriar I'd say - its back on the fast track.

grafik


What I changed:
- reduce timeout to 2ms
- introduce a 2^18 chunk limit
- userInput flag to skip buffering after receiving

You might want to check, if the input latency stays low. If not, then there is a catch hidden somewhere :sweat_smile:

Tyriar (approved):

PR #4177, opened 2022-10-06
Revert "Merge pull request #4166 from Tyriar/event_with_emitter"

Tyriar:
This reverts commit d22f7c9edd46f5ae0bed05489c915bbde4b35759, reversing changes made to 1f8e6f095266a5236899a1ebe151983ad527b9da.

When discussed with the VS Code team they pointed out this is a bit dangerous exposing the emitters publicly everywhere. Reverting as I was convinced safety beats convenience.


jerch:
@Tyriar How is that more dangerous than the current approach? I mean a serious attacker can just follow the internal symbols to get down to the emitters anyway? (Btw thats basically what I did during the listener debugging to get a hold if their root...)

Tyriar:
It's not a security issue, but a separation of concerns issue. The risk is too great that some commit would be made that creates an instance of a class class but not assign it to a variable typed with the interface (not the class) which would expose fire.

jerch:
Ah ok, yeah this newer approach mixed functionality into one thingy, that needs a certain usage pattern. I tend to treat those things as "if in doubt: explicit is better than implicit".

PR #4180, opened 2022-10-07
Fix clearTextureAtlas call as implemented on IRenderer

Tyriar:
This is the root cause of microsoft/vscode#162996


meganrogge (approved):

Tyriar:
This is a bug where `Terminal.clearTextureAtlas` wasn't doing anything for webgl, but I think the main bug still remains

PR #4181, opened 2022-10-07
Fix race condition that could when clearing atlas on 2+ terminals

Tyriar:
This is the proper fix for microsoft/vscode#162996, after a bunch of investigation we found that this is causing the problem:

VS Code clears the texture atlas on OS resume to fix an issue where the texture could be corrupted. This calls clearTextureAtlas twice which should be fine, but it's not because inside WebglRenderer.clear we're calling updateModel which is meant to only ever be called in an animation frame. This calls potentially both renderers to have stale views of the texture atlas for those rows and our perf caching of cells (the main purpose of the model) doesn't know it's stale.


meganrogge (approved):

Tyriar:
@ahmad-511 thanks for confirming!

PR #4192, opened 2022-10-09
Switch to Ubuntu 20.04

silamon:
The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002


Tyriar (approved):
👍

PR #4193, opened 2022-10-09
Fix demo again

silamon:
Same as #3983.

### DIFF
--- a/addons/xterm-addon-canvas/src/CanvasAddon.ts
+++ b/addons/xterm-addon-canvas/src/CanvasAddon.ts
@@ -25,9 +25,7 @@ export class CanvasAddon extends Disposable implements ITerminalAddon {
   public activate(terminal: Terminal): void {
     const core = (terminal as any)._core;
     if (!terminal.element) {
-      this.register(toDisposable(() => {
-        core.onWillOpen(() => this.activate(terminal));
-      }));
+      this.register(core.onWillOpen(() => this.activate(terminal)));
       return;
     }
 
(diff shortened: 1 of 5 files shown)


addons/xterm-addon-canvas/src/CanvasAddon.ts
```diff
@@ -25,7 +25,9 @@ export class CanvasAddon extends Disposable implements ITerminalAddon {
   public activate(terminal: Terminal): void {
     const core = (terminal as any)._core;
     if (!terminal.element) {
-      this.register(core.onWillOpen(() => this.activate(terminal)));
+      this.register(toDisposable(() => {
+        core.onWillOpen(() => this.activate(terminal));
+      }));
```
Tyriar, line 30: This change may not be needed? Regardless this won't work as it will only attach the `onWillOpen` event after the webgl renderer is disposed. The intent with that is to allow loading webgl/canvas without calling `Terminal.open` yet

Tyriar (changes requested):
Sorry about the breakages 😅

addons/xterm-addon-canvas/src/CanvasAddon.ts
silamon: I think I understand it. So, I've added a try-catch around terminal.open, but it fails somewhere else then. It looks like the terminal remains broken when it needs to fallback to a different renderer such as canvas or dom.

silamon: I see the dom renderer back now, but the demo fails to see that the webgl didn't load, since the error is handled silently in terminal.ts now. It should be thrown again at the end of the terminal initialization.

src/browser/Terminal.ts
```diff
@@ -515,7 +515,10 @@ export class Terminal extends CoreTerminal implements ITerminal {
     // Performance: Add viewport and helper elements from the fragment
     this.element.appendChild(fragment);
 
-    this._onWillOpen.fire(this.element);
+    try {
+      this._onWillOpen.fire(this.element);
+    }
+    catch { /* fails to load addon for some reason */ }
```
Tyriar, line 521: I like it 👍

Tyriar (approved):

PR #4194, opened 2022-10-09
Capture data between terminal creation and connected to terminal in demo

silamon:
When I opened the demo terminal, the first line was missing. The terminal is functioning whatsoever. Looks like the terminal first line is sent before the terminal frontend is connected. Thanks to `ws.send(logs[term.pid]);` that line is sent to the terminal frontend when the terminal gets connected. 

This commit introducing this: https://github.com/xtermjs/xterm.js/commit/193d305dcd37537aa47758e580059ffd220b8723
This reverts it but keeps the higher timeout.

### DIFF
--- a/demo/server.js
+++ b/demo/server.js
@@ -17,8 +17,8 @@ function startServer() {
   expressWs(app);
 
   var terminals = {},
-      unsentOutput = {},
-      temporaryDisposable = {};
+    unsentOutput = {},
+    temporaryDisposable = {};
 
   app.use('/xterm.css', express.static(__dirname + '/../css/xterm.css'));
   app.get('/logo.png', (req, res) => {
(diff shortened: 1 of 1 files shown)


demo/server.js
```diff
@@ -54,6 +55,10 @@ function startServer() {
 
     console.log('Created terminal with PID: ' + term.pid);
     terminals[term.pid] = term;
+    logs[term.pid] = '';
+    term.on('data', function(data) {
+      logs[term.pid] += data;
+    });
```
Tyriar, line 61: Can we dispose of this listener once the terminal is connected? So after right that `ws.send` call? That will give us the best of both worlds.

We could then rename logs to something more descriptive to prevent this accident in the future, like `unsentOutput`?

Tyriar (changes requested):
Thanks! I saw this but had no idea why it was happening.

demo/server.js
jerch: Oh well - yeah I removed it because it is a nuisance during perf measuring (backend constantly dies after testing a serious amount of data) - and I wasnt aware of the race condition (never happened for me). I think removing the aggregation buffer once the connection was successful would give us both advantages.

Tyriar (approved):
Thanks 😃

PR #4200, opened 2022-10-12
Optimize _clipImageData

### LINKED ISSUE #4197: _clipImageData is GC expensive
Part of https://github.com/microsoft/vscode/issues/161622

During vscode startup:

Screen Shot 2022-10-11 at 9 18 05 am

Tyriar:
This makes clip image data around twice as fast, it's a little tricky to measure it though. Changes:

- Work on Uint32Array instead of Uint8 so it's 1 assignment per pixel instead of 1 per channel.
- Perform the clipping in-place using the original image data to avoid allocation a new buffer.

Fixes #4197

### DIFF
--- a/src/browser/renderer/shared/TextureAtlas.ts
+++ b/src/browser/renderer/shared/TextureAtlas.ts
@@ -770,20 +770,24 @@ export class TextureAtlas implements ITextureAtlas {
   }
 
   private _clipImageData(imageData: ImageData, boundingBox: IBoundingBox): ImageData {
+    // Operate on pixels instead of channels to reduce the amount of work
+    const originalData = new Uint32Array(imageData.data.buffer);
+
+    // Create a new view on the same buffer for the clipped data. The clipping operation is done in
+    // place to avoid allocating another buffer
     const width = boundingBox.right - boundingBox.left + 1;
     const height = boundingBox.bottom - boundingBox.top + 1;
-    const clippedData = new Uint8ClampedArray(width * height * 4);
-    for (let y = boundingBox.top; y <= boundingBox.bottom; y++) {
-      for (let x = boundingBox.left; x <= boundingBox.right; x++) {
-        const oldOffset = y * this._tmpCanvas.width * 4 + x * 4;
-        const newOffset = (y - boundingBox.top) * width * 4 + (x - boundingBox.left) * 4;
-        clippedData[newOffset] = imageData.data[oldOffset];
-        clippedData[newOffset + 1] = imageData.data[oldOffset + 1];
-        clippedData[newOffset + 2] = imageData.data[oldOffset + 2];
-        clippedData[newOffset + 3] = imageData.data[oldOffset + 3];
+    const clippedData = new Uint32Array(imageData.data.buffer, 0, width * height);
+
+    // Perform clipping and return the result
+    let x = 0;
+    let y = 0;
+    for (y = boundingBox.top; y <= boundingBox.bottom; y++) {
+      for (x = boundingBox.left; x <= boundingBox.right; x++) {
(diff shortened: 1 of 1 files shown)


src/browser/renderer/shared/TextureAtlas.ts
```diff
@@ -770,20 +770,24 @@ export class TextureAtlas implements ITextureAtlas {
...
-    const clippedData = new Uint8ClampedArray(width * height * 4);
-    for (let y = boundingBox.top; y <= boundingBox.bottom; y++) {
-      for (let x = boundingBox.left; x <= boundingBox.right; x++) {
-        const oldOffset = y * this._tmpCanvas.width * 4 + x * 4;
-        const newOffset = (y - boundingBox.top) * width * 4 + (x - boundingBox.left) * 4;
-        clippedData[newOffset] = imageData.data[oldOffset];
-        clippedData[newOffset + 1] = imageData.data[oldOffset + 1];
-        clippedData[newOffset + 2] = imageData.data[oldOffset + 2];
-        clippedData[newOffset + 3] = imageData.data[oldOffset + 3];
+    const clippedData = new Uint32Array(imageData.data.buffer, 0, width * height);
```
jerch, line 780: Oh well - inplace overwriting is abit dangerous, if the overlapping from rewrites is not safe (guess its ok here?)

src/browser/renderer/shared/TextureAtlas.ts
```diff
@@ -770,20 +770,24 @@ export class TextureAtlas implements ITextureAtlas {
...
+    // Perform clipping and return the result
+    let x = 0;
+    let y = 0;
+    for (y = boundingBox.top; y <= boundingBox.bottom; y++) {
+      for (x = boundingBox.left; x <= boundingBox.right; x++) {
+        clippedData[(y - boundingBox.top) * width + (x - boundingBox.left)] = originalData[y * imageData.width + x];
       }
     }
-    return new ImageData(clippedData, width, height);
+    return new ImageData(new Uint8ClampedArray(clippedData.buffer, clippedData.byteOffset, clippedData.byteLength), width, height);
```
jerch, line 790: Not sure if it is worth the trouble - since you already do inplace overwriting above, the new ImageData maybe can also be avoided?

Which me to the question - is the `_clipImageData` method needed at all? As far as I can see it just transfers a rectangular subarea over. Isnt it faster to simply take the orginal ImageData and allpy the box offsets during `drawImage`? But maybe I am overlooking something...

jerch (commented):
A few more notes on code...

src/browser/renderer/shared/TextureAtlas.ts
Tyriar: It's safe as the data is read from low to high and the writing is always the same as writing (rare) or lower

src/browser/renderer/shared/TextureAtlas.ts
Tyriar: You're right! We can remove this outright and just use boundingBox in drawImage. I'm guessing the reason it's as it is currently is because I just overlooked that or wasn't thinking drawImage had a this signature source+dest signature.

Tyriar: Was just doing this, we can't actually use drawImage as it doesn't accept ImageData, looks like we can still use putImageData though.

Tyriar: 🎉 https://github.com/xtermjs/xterm.js/pull/4201/

jerch: Oh right, was mixing the methods up (I use putImageData myself, lol) :blush:

Tyriar: Surely it's faster, it's essentially doing what was being done manually before natively

PR #4208, opened 2022-10-16
Warn/throw on unexpected attach addon socket state

Tyriar:
@jerch fixing a random TODO, does this seem reasonable?

### DIFF
--- a/addons/xterm-addon-attach/src/AttachAddon.ts
+++ b/addons/xterm-addon-attach/src/AttachAddon.ts
@@ -47,16 +47,14 @@ export class AttachAddon implements ITerminalAddon {
   }
 
   private _sendData(data: string): void {
-    // TODO: do something better than just swallowing
-    // the data if the socket is not in a working condition
-    if (this._socket.readyState !== 1) {
+    if (!this._checkOpenSocket()) {
       return;
     }
     this._socket.send(data);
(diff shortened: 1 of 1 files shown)


addons/xterm-addon-attach/src/AttachAddon.ts
```diff
@@ -65,6 +63,22 @@ export class AttachAddon implements ITerminalAddon {
...
+
+  private _checkOpenSocket(): boolean {
+    switch (this._socket.readyState) {
+      case WebSocket.OPEN:
+        return true;
+      case WebSocket.CONNECTING:
+        throw new Error('Attach addon was loaded before socket was open');
+      case WebSocket.CLOSING:
+        console.warn('Attach addon socket is closing');
+        return false;
```
jerch, line 75: Why not throwing here as well? Would this create race conditions during app shutdown?

jerch (approved):
LGTM, just a minor comment on `CLOSING`.

addons/xterm-addon-attach/src/AttachAddon.ts
Tyriar: Was worried about breaking something as we dispose on `'close'`, not closing. Probably doesn't matter 🤷

PR #4216, opened 2022-10-19
'CSI 0 m' should clear the extended attribute

JasonXJ:
Before this change, `echo -e '\033[4m\033[m\033]8;;\033\hello'` will print 'hello' with a underline. This is what happened:

- `\033[4m`: set underline style
- `\033[m`: reset style. However, in [InputHandler.ts](https://github.com/xtermjs/xterm.js/blob/0b84d17b18e836a37ed544df337630709bc7aede/src/common/InputHandler.ts#L2441), we only reset the `fg` and `bg` but not the `extended` attribute.
- `\033]8;;\033\`: This trigger `_finishHyperlink()`, which resurrects the dirty extended attribute by calling `this._curAttrData.UpdateExtended()`.

This change fixes this by also resetting the `extended` attribute for 'CSI 0 m'. It also removes `_currentLinkId` from `InputHandler` and uses `_curAttrData.extended.urlId` directly so that we don't need to worry about keep them in sync any more.

### DIFF
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -2356,6 +2356,22 @@ export class InputHandler extends Disposable implements IInputHandler {
     attr.updateExtended();
   }
 
+  private _processSGR0(attr: IAttributeData): void {
+    attr.fg = DEFAULT_ATTR_DATA.fg;
+    attr.bg = DEFAULT_ATTR_DATA.bg;
+
+    if (!attr.extended.isEmpty()) {
+      // Treat extended attrs as immutable, thus always clone from old one.
+      // This is needed since the buffer only holds references to it.
+      attr.extended = attr.extended.clone();
+      attr.extended.ext = 0;
+
+      if (attr.extended.urlId) {
+        attr.bg |= BgFlags.HAS_EXTENDED;
+      }
+    }
+  }
+
   /**
    * CSI Pm m  Character Attributes (SGR).
    *
@@ -2441,9 +2457,7 @@ export class InputHandler extends Disposable implements IInputHandler {
   public charAttributes(params: IParams): boolean {
     // Optimize a single SGR0.
     if (params.length === 1 && params.params[0] === 0) {
-      this._curAttrData.fg = DEFAULT_ATTR_DATA.fg;
-      this._curAttrData.bg = DEFAULT_ATTR_DATA.bg;
-      this._curAttrData.extended = DEFAULT_ATTR_DATA.extended.clone();
+      this._processSGR0(this._curAttrData);
       return true;
     }
 
@@ -2471,9 +2485,7 @@ export class InputHandler extends Disposable implements IInputHandler {
         attr.bg |= Attributes.CM_P16 | (p - 100) | 8;
       } else if (p === 0) {
         // default
-        attr.fg = DEFAULT_ATTR_DATA.fg;
-        attr.bg = DEFAULT_ATTR_DATA.bg;
-        attr.extended = DEFAULT_ATTR_DATA.extended.clone();
+        this._processSGR0(attr);
       } else if (p === 1) {
         // bold text
         attr.fg |= FgFlags.BOLD;
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -2360,14 +2360,14 @@ export class InputHandler extends Disposable implements IInputHandler {
     attr.fg = DEFAULT_ATTR_DATA.fg;
     attr.bg = DEFAULT_ATTR_DATA.bg;
 
+    // Reset `extended` except for the `urlId`.
     if (!attr.extended.isEmpty()) {
-      // Treat extended attrs as immutable, thus always clone from old one.
-      // This is needed since the buffer only holds references to it.
-      attr.extended = attr.extended.clone();
-      attr.extended.ext = 0;
+      const urlId = attr.extended.urlId;
+      attr.extended = DEFAULT_ATTR_DATA.extended.clone();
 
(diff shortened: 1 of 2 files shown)


jerch:
Thx for spotting this, kinda overlooked the SGR 0 case when doing the extended attr patch.

I think there is still a catch here:
Url state is carried around in extendedAttribs but should not reset on SGR 0, I think we need to nullify everything in ext attrs except the url id, otherwise SGR 0 would also finish an url?

Tyriar (commented):
Nice 👏 I think this is the cause for an issue I saw after using hyperlinks where everything would get underlined

JasonXJ:
I tested a few other terminals, and they indeed didn't end the url with SGR 0 (except for hterm). I just uploaded a new patch, which will do the same.

src/common/InputHandler.ts
```diff
@@ -2353,6 +2356,22 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+      // Treat extended attrs as immutable, thus always clone from old one.
+      // This is needed since the buffer only holds references to it.
+      attr.extended = attr.extended.clone();
+      attr.extended.ext = 0;
+
+      if (attr.extended.urlId) {
+        attr.bg |= BgFlags.HAS_EXTENDED;
+      }
+    }
+  }
```
jerch, line 2373: Imho the `attr.extended.ext = 0;` is abit dangerous, as it is not clear from the data point, that it is only holding SGR relevant data (and whether 0 is the right reset action). Currently it does the right thing, but I think we should make it more explicit to avoided coding errors later on.

Now for SGR0 there are 2 important data points in extended - underlineStyle and underlineColor. Both need to be reset to default, everything else in extended should be left untouched.

Suggestion:
- clone current ext attr (as you already do)
- explicitly reset underlineStyle by setting to `UnderlineStyle.NONE`
- reset underline color: Normally resetting color mode to `CM_DEFAULT` should do, to be on the safe side, you can also clear `RGB_MASK` (thus `underlineColor &= ~(Attributes.CM_MASK | Attributes.RGB_MASK)`).
- call `attr.updateExtended();`, which should update the BG flag correctly (stays true if there is an url id, otherwise should turn false)
- done

jerch (commented):
> I tested a few other terminals, and they indeed didn't end the url with SGR 0 (except for hterm).

Specwise OSC 8 is not linked to SGR at all. It is just an implementation detail, that we store it along with attributes handled by SGR.
Our extended attrs is abit like a "Resterampe" (german word for a collection of things of low/rare interest that dont fit anywhere else). So things ending up there share the low interest feature and the need to be applied for several input chars, but not necessarily the same action semantics.

> I just uploaded a new patch, which will do the same.

Basically LGTM, but I have a few readability comments/suggestions below.

JasonXJ:
What about just cloning the default extended and set the urlId on it? This seems to address your concern, and it is simpler.

I also changed `attr.bg |= BgFlags.HAS_EXTENDED` to `attr.updateExtended()`. But IMO the former one is better since 1) it avoid the extra check on `extended` and 2) we are already directly manipulating `attr.bg` any way. Let me know if you want me to change it back, but I don't really mind either.

jerch:
> What about just cloning the default extended and set the urlId on it? This seems to address your concern, and it is simpler.

Ah nope, I think you missed my point, that SGR functionality should not know anything about other data points in extended. While your idea is simpler atm, it breaks badly in the future, when more data points got added to extended.
Imagine I add a "fancy feature xy" to the terminal, with a data point in extended. With your idea you force me to revisit SGR code, while SGR is completely independent of "fancy feature xy". This code entanglement leads directly to maintenance hell.

So this is a matter of **isolation of concerns**, which is normally enforced by OOP measures. Thats currently not the case for extended, as it creates a big perf overhead. (But I start to wonder, if we should interface extended more strictly.)

My suggestion above in code (untested):
```typescript
private _sgrReset(attr: IAttributeData): void {
  // reset FG/BG to default colors
  attr.fg = DEFAULT_ATTR_DATA.fg;
  attr.bg = DEFAULT_ATTR_DATA.bg;
  // clone extended attr to get a fresh ref
  attr.extended = attr.extended.clone();
  // reset underline style and color
  attr.extended.underlineStyle = UnderlineStyle.NONE;
  attr.extended.underlineColor &= ~(Attributes.CM_MASK | Attributes.RGB_MASK);
  // update extended flag in BG (note: true if there still was other data in extended)
  attr.updateExtended();
}
```

JasonXJ:
ok. updated.

jerch (approved):
:+1: LGTM. Thx for spotting and fixing it.

JasonXJ:
Thanks. Is there anything else I need to do here? I don't think I can merge this myself.

JasonXJ:
@Tyriar , can you merge this branch? Also, do I need to squash the commits? I am actually quite confused by how github works. I just pressed a rebase button on this page, and now it says "[JasonXJ](https://github.com/JasonXJ) [force-pushed](https://github.com/xtermjs/xterm.js/compare/538461ff9b4a173efa747e28d2bf7a83778121ee..81a3c80bc49ddce022562ff167a7ee8312a55a40) the fix-csi-0-m branch from [538461f ](https://github.com/xtermjs/xterm.js/commit/538461ff9b4a173efa747e28d2bf7a83778121ee)to [81a3c80](https://github.com/xtermjs/xterm.js/commit/81a3c80bc49ddce022562ff167a7ee8312a55a40)". Hopefully, I am not breaking anything.

Tyriar (approved):
Thanks @JasonXJ 👍

Issue #3260, opened 2021-03-22
ANSI OSC 52 support?


olegkrutov:
ANSI OSC 52 is a sequence to transport some data from remote console app to client's clipboard.  For example, tmux can use it on remote side, and xterm (terminal app), Windows terminal and a number of others -- on client side. It would be very convenient to have such a feature in xterm.js.
Here is a simple test which sends a string from remote to local clipboard (taken [here](https://github.com/realh/roxterm/issues/206#issue-433783926)):
`printf "\033]52;c;$(printf "%s" "blabla" | base64)\a"`

Tyriar:
This would need some thinking around how embedders can conditionally enable it, like how browsers ask if they can access the clipboard.

olegkrutov:
Maybe it can be done as plugin so it can be totally disabled, otherwise raising a pop-up window when such a sequence arrives first time in session?

jerch:
Just tested this with Firefox and Chrome:
```javascript
setTimeout(() => navigator.clipboard.writeText('Hello there!'), 1000);
```

which seems to work in both as long as something on the page has the focus. Not quite sure about the focus state, it also works if the browser window is in the background, but does not work, if the focus is in the console.
No clue about Safari.

What would be possible:
- make it optional via terminal settings or as addon
- use async handler for OSC 52, with this we can stop terminal progression and show a "allow clipboard access" popup, if the attempt fails in the first place due to missing focus (`clipboard.writeText` returns a promise, easy to get that up with an async handler now :smile_cat:)

Tyriar:
> which seems to work in both as long as something on the page has the focus.

I think browsers treat this differently and it may get blocked if it wasn't caused by a "user event" (ie. `KeyboardEvent`, `MouseEvent`).

> make it optional via terminal settings or as addon

The way VS Code copies depends on the environment it's in, so this might need an event. If it was an event the embedder could just not listen to it if it doesn't want to support it then.

> with this we can stop terminal progression and show an "Allow clipboard access?" popup,

We probably wouldn't want to pause the terminal, just request it and not await.

olegkrutov:
Isn't it worth to assume that since we are selecting text here in xterm session, we have a focus right now? Focus will be lost in situations like some "mirroring" of terminal output but I don't think this is 1) majority of cases and 2) we should manipulate with clipboard in these situations...

jaraco:
[...]

PR #4220, opened 2022-10-19
Add support to ANSI OSC52

aymanbagabas:
Add support to ANSI OSC52 sequence to manipulate selection and clipboard
data. The sequence specs supports multiple clipboard selections but we
only support the common ones, system and primary clipboard selections.

This adds a new event listener to the common terminal module
`onClipboard` to allow external implementations to hook into it.

The default ClipboardProvider in `xterm-addon-clipboard` uses the browser
Clipboard API to read/write from and to the clipboard. The default
`ClipboardProvider` uses the browser Clipboard API. This means it only
supports read/write to and from the system clipboard. This default
`ClipboardProvider` is limited to 1MB of data to write (zero means no limit).

TODO:
- [x] Add OSC52 handler to `/common`
- [x] Implement a clipboard addon that uses [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard)
- [x] Add `allowClipboardAccess` option. Disabled by default 
- [x] Add tests
- [ ] ~Find a workaround for the browser focus issue~

Reference: https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands
Fixes: #3260


jerch:
@aymanbagabas Thx for looking into this. A few notes from my side:

The service cannot go into `/common`, as this is also used for headless lib builds without a browser frontend. Thus the service with navigator notions should be under `/browser`, while `/common` may hold a stub, where integrators can hook in their own clipboard backend.

Furthermore clipboard access needs some thinking about the security implications. Imho it should be disallowed by default, and only granted explicitly by a user interaction (maybe also customizable, so integrators can disable the user feedback, if they have other measures in place). This is especially important for websites with xterm.js, as the browser clipboard might hold totally different content, that should not be accessible by some random terminal app.

aymanbagabas:
> @aymanbagabas Thx for looking into this. A few notes from my side:
> 
> The service cannot go into `/common`, as this is also used for headless lib builds without a browser frontend. Thus the service with navigator notions should be under `/browser`, while `/common` may hold a stub, where integrators can hook in their own clipboard backend.

Very reasonable, do you think exposing `attachClipboardService(IClipboardService)` in `/common/CoreTerminal.ts` is reasonable?

> Furthermore clipboard access needs some thinking about the security implications. Imho it should be disallowed by default (xterm even puts it under windowOps security setting), and only granted explicitly by a user interaction (maybe also customizable, so integrators can disable the user feedback, if they have other measures in place). This is especially important for websites with xterm.js, as the browser clipboard might hold totally different content, that should not be accessible by some random terminal app.
> 

I think adding an option to allow clipboard access makes sense. Perhaps add `allowClipboardAccess` as a terminal option? In `/browser` for more security, something like `confirm('Allow xterm.js to read from the clipboard?')` could be added in the service implementation when a clipboard read is performed. 

> Edit: Also the focus issue mentioned in https://github.com/xtermjs/xterm.js/issues/3260 needs some workaround for the browser part.

Will look into that once the API is in place.

ghuntley:
Howdy from https://coder.com ala https://github.com/coder. Dropping on by to say _thank-you_ for picking up this item. OSC52 support has been [requested by our community](https://github.com/coder/coder/issues/5001) and I'm opening up an open-source bounty payable when this item ships. Our way of saying thank-you. ❤️

Tyriar:
I probably won't be able to review this until early December, sorry about the delay.

src/common/InputHandler.ts
```diff
@@ -123,6 +123,7 @@ export class InputHandler extends Disposable implements IInputHandler {
   private _iconName = '';
   private _currentLinkId?: number;
   private _dirtyRowTracker: IDirtyRowTracker;
+  private _clipboardService: IClipboardService | undefined;
```
Tyriar, line 126: Instead of an optional service here it would be better to first an onRequest event

Tyriar (changes requested):

src/common/InputHandler.ts
```diff
@@ -3027,6 +3035,65 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+    const pc = args[0];
+    const pd = args[1];
+    switch (pc) {
+      case ClipboardType.CLIPBOARD:
+      case ClipboardType.PRIMARY:
+        if (pd === '?') {
+          // Reply with the current clipboard contents encoded in base64
+          return this._clipboardService.readData(pc)
+            .then(data => {
+              this._onClipboard.fire(Buffer.from(data).toString('base64'));
```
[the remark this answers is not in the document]
aymanbagabas: I'm using `Buffer` here to encode strings in base64. The browser comes with `atob()` and `btoa()`. We can achieve the same with `Buffer` here though. We can't encode strings to base64 with ArrayBuffer.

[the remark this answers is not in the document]
jerch: Why base64 in the first place? Is this needed by the clipboard interfaces?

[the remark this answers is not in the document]
aymanbagabas: Yes, base64 is needed by the OSC52 clipboard protocol.

> The second parameter, Pd, gives the selection data.  Normally
          this is a string encoded in base64 (RFC-4648).  The data
          becomes the new selection, which is then available for pasting
          by other applications.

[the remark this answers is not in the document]
jerch: Then we need another workaround for this, prolly based on `atob/btoa` in the browser and `Buffer` in nodejs. We cannot use a foreign `Buffer` impl brigding this, as it would introduce a 3rd-party dependency at runtime. 

Furthermore our underlying data is codepoints in UTF-32, which lets me think that we might need a custom base64 encoder here, means  encode to utf-8 first (is the byte level meant to be in utf-8?), then encode to base64. Such an encoder also might have to deal with special characters like NUL (from empty cells), cannot find anything regarding that in the OSC 52 docs. Does anyone know?

src/common/InputHandler.ts
```diff
@@ -3027,6 +3035,65 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+   * - `s` - select
+   * - `0-7` - cut-buffers 0-7
+   * Only the `c` selection (clipboard) is supported by xterm.js. The browser
+   * Clipboard API only supports the clipboard selection.
+   *
+   * Pd is the base64 encoded data.
+   * If Pd is `?`, the terminal returns the current clipboard contents.
+   * If Pd is neither base64 encoded nor `?`, then the clipboard is cleared.
+   */
+  public setOrReportClipboard(data: string): Promise<boolean> {
```
jerch, line 3061: I'm not sure, if we are all happy with the promise return value here - as mentioned in last paragraph here: https://github.com/xtermjs/xterm.js/issues/3260#issuecomment-804228282 (note: the promise return here basically stops in-band progression of the terminal until the promise got resolved the good or the bad way).

On overall return values:
Where possible, the return value should by a sync value of `true|false` to not slow down the data processing too much. Furthermore, since this is a default handler in `InputHandler`, it should always return a true-ish value to not into exhaustion check in the parser.

src/common/InputHandler.ts
```diff
@@ -3027,6 +3035,65 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+    return this._setOrReportClipboard(data);
+  }
+
+  private _setOrReportClipboard(data: string): Promise<boolean> {
```
jerch, line 3065: Return value should be `Promise<boolean> | boolean` (or just `boolean`, if we dont want to stop in-band processing at all).

src/common/InputHandler.ts
```diff
@@ -3027,6 +3035,65 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+    if (!this._clipboardService || !this._optionsService.options.allowClipboardAccess) {
+      return Promise.resolve(false);
```
jerch, line 3067: Return value should be flat `true`.

src/common/InputHandler.ts
```diff
@@ -3027,6 +3035,65 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+    }
+    const args = data.split(';');
+    if (args.length < 2) {
+      return Promise.resolve(false);
```
jerch, line 3071: Return value should be flat `true`.

src/common/InputHandler.ts
```diff
@@ -3027,6 +3035,65 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+        }
+        // Verify that the data is base64 encoded before writing it to the clipboard
+        // otherwise clear the clipboard
+        const buf = Buffer.from(pd, 'base64');
+        if (buf.toString('base64') === pd) {
+          return this._clipboardService.putData(pc, buf.toString()).then(() => true);
+        }
+        return this._clipboardService.clearData(pc).then(() => true);
+    }
+    return Promise.resolve(false);
```
jerch, line 3094: Return value should be flat `true`.

src/common/InputHandler.ts
```diff
@@ -3027,6 +3035,65 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+    switch (pc) {
+      case ClipboardType.CLIPBOARD:
+      case ClipboardType.PRIMARY:
+        if (pd === '?') {
+          // Reply with the current clipboard contents encoded in base64
+          return this._clipboardService.readData(pc)
+            .then(data => {
+              this._onClipboard.fire(Buffer.from(data).toString('base64'));
+              return true;
+            });
```
jerch, line 3084: If we stick with promises in OSC52 - it also needs a `.catch(e => true)` clause.

(Also below in more places.)

src/common/InputHandler.ts
```diff
@@ -3027,6 +3035,65 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+        }
+        // Verify that the data is base64 encoded before writing it to the clipboard
+        // otherwise clear the clipboard
+        const buf = Buffer.from(pd, 'base64');
+        if (buf.toString('base64') === pd) {
+          return this._clipboardService.putData(pc, buf.toString()).then(() => true);
```
jerch, line 3090: Needs a `.catch(e => true)` clause.

src/common/InputHandler.ts
```diff
@@ -3027,6 +3035,65 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+        }
+        return this._clipboardService.clearData(pc).then(() => true);
```
jerch, line 3092: needs a `.catch(e => true)` clause.

jerch (commented):
Just commenting for now, as I think that this still needs several iterations...

src/common/InputHandler.ts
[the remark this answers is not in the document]
aymanbagabas: To avoid introducing the bridging dependency, we could detect if we're running NodeJS using `typeof window === 'undefined'` and use `Buffer.from`, otherwise, use `atob/btoa`. With this approach, we still need https://github.com/xtermjs/xterm.js/pull/4220/commits/a8856804ca6b937b6c46923e89c4f82afdfdab6c to make it compatible with both NodeJS and the browser. cc/ @Tyriar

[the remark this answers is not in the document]
jerch: Well the distinction via `window` is unfortunate, as it might lead to issues in workers. How about something like that:
```js
const b64enc = (typeof btoa !== 'undefined') ? btoa : s => Buffer.from(s).toString('base64');
```
(Still would need some testing, whether it produces the same output for all allowed chars/control codes in OSC 52.)

src/common/InputHandler.ts
aymanbagabas: Reading/writing from and to the [clipboard is an asynchronous](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#instance_methods) operation, thus the promise returns here.

src/common/InputHandler.ts
aymanbagabas: Could you please expand on this?

src/common/InputHandler.ts
jerch: I think we dont need in-band processing here (means - synchronous to position in the input data), imho we can just spawn an internal event here, that gets digested by follow-up code (sync or async handled). Ofc this might lead to issues, if an app would spawn multiple OSC 52 sequences in a row (might stack up), or even expects it to be synchronous to a certain terminal buffer state. But the latter assumption should never be made on app side, as the terminal interface for back reporting is async by definition (and might not even respond to all requests).

Pros: We dont stop the parser at all (all sync return values), instead delegate processing to a later event handler invocation.
Cons: Possible stacking of OSC 52 requests, thus this needs a logic to throw away older not yet processed requests (maybe enqueued if it makes sense for the OSC52 actions).

src/common/InputHandler.ts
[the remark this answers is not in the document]
jerch: Ah sorry for the sloppy snippet - it still misses the unicode string to utf8 bytestring conversion for the `btoa` path. (would be needed for both `Base64.decode` and `Base64.encode`).

jerch:
@aymanbagabas To make that event idea more clear:

Imho all it needs in `common`, is a sync inputhandler function for OSC52 with two abstract interface "endpoints" - retrieve and place clipboard data. For maximum flexibility the endpoints can be feed from an clipboard event, to illustrate that:

```typescript
const enum ClipboardEventType {
  QUERY = 0,
  SET = 1
}
interface IClipboardEvent {
  type: ClipboardEventType;
  payload: string;
}

// in OSC52 function
// for request: emit {type: ClipboardEventType.QUERY, payload: ''}
// for set: emit {type: ClipboardEventType.SET, payload: 'unicode string data'}
```
Now all that needs to be done, is to interface that in way, so embedders can hook into it with their clipboard logic. Ideally we create a public API endpoint for this, e.g. `onClipboard((ev: IClipboardEvent) => void)`.

In a second step we can discuss, whether a default implementation should be provided from core repo  under `/browser` (I am not a big fan of this idea for security reasons), it is prolly a better idea to outsource the clipboard handler logic into an addon. This way embedders have explicit to hook their logic (hoopefully with security in mind), or explicit to load the addon, where we can place further security warnings and such.

aymanbagabas:
@jerch I've made the default implementation into a new addon `xterm-addon-clipboard`. You still need to enable the `allowClipboardAccess` option before the addon can work. Using the addon is as simple as loading it into xtermjs.

src/common/Base64.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Encode UTF-8 string to base64 encoded string.
+ * @param data The string to encode.
+ * @returns The base64 encoded string.
+ */
+export const encode = (data: string): string => {
+  try {
+    return typeof btoa !== 'undefined' ?
+      btoa(data) :
```
jerch, line 29: Both `Base64.decode` and `Base64.encode` still need the patch for correctly handling higher unicode chars, eg. in encode:
- grab unicode string
- encode it to utf-8 bytes
- convert bytes to bytestring
- apply `btoa`

src/common/Base64.ts
```diff
@@ -0,0 +1,34 @@
...
+      Buffer.from(data, 'binary').toString('base64');
```
jerch, line 30: Shouldn't `binary` be `utf-8` here? I think we lose unicode support with binary here.

src/common/InputHandler.ts
```diff
@@ -3037,6 +3042,67 @@ export class InputHandler extends Disposable implements IInputHandler {
...
+          selection: pc,
+          type: ClipboardEventType.SET
+        };
+        if (pd === '?') {
+          ev.type = ClipboardEventType.QUERY;
+        } else {
+          // Verify that the data is base64 encoded before writing it to the
+          // clipboard.
+          const str = Base64.decode(pd);
+          if (Base64.encode(str) === pd) {
```
jerch, line 3096: This decode->encode comparison will most likely create issues later on, as base64 is not stable/strict in its encoding rules, thus it might reject good cases, where the padding/whitespace settings used differ.

typings/xterm.d.ts
```diff
@@ -1736,4 +1748,31 @@ declare module 'xterm' {
...
+
+  /**
+   * Clipboard event interface. This is used to communicate with the the
+   * clipboard addon to access the system clipboard.
+   */
+  export interface IClipboardEvent {
+    type: ClipboardEventType;
+    selection: ClipboardSelectionType;
+    payload?: string;
+  }
```
jerch, line 1777: I think `const enum` is not a good idea in d.ts? Also it hides the real values, which makes it much harder to work with from JS side. Suggestion:
```typescript
  export interface IClipboardEvent {
    type: 'query' | 'set';
    selection: 'clipboard' | 'primary';
    payload?: string;
  }
```
and remove the enums.

Also what is this selection clipboard/primary thingy about? Is this important to keep working on all platforms or can we just drop the X11 only definition?

jerch (commented):
Nice, looks really good already. Have not yet tested it, so first a few code comments from my side.

aymanbagabas:
> Nice, looks really good already. Have not yet tested it, so first a few code comments from my side.

I'm thinking of moving the whole Osc logic to the addon and relying on an external package like [js-base64](https://github.com/dankogai/js-base64) for base64. This way all the security stuff is left out of the base package and implementers would have to opt-in and activate the addon for it to work.

addons/xterm-addon-clipboard/src/ClipboardAddon.ts
```diff
@@ -0,0 +1,67 @@
...
+  public writeText(data: string): Promise<void> {
+    if (navigator.clipboard) {
+      return navigator.clipboard.writeText(data);
+    }
+    return Promise.reject(new Error('Clipboard API not available'));
+  }
+
+  public readText(): Promise<string> {
+    if (navigator.clipboard) {
+      return navigator.clipboard.readText();
```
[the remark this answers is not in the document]
aymanbagabas: Chrome uses Permissions API ([`clipboard-read`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText#security) and [`clipboard-write`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText#security)) and then prompts the user to allow read access every time the clipboard is read, writing to the clipboard is permitted without permission in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/User_activation).

Firefox allows writing text to clipboard in secure contexts. Reading the clipboard requires enabling the `dom.events.asyncClipboard.readText` config. Then, every time the clipboard is read, the browser pops up a "paste" option to read from clipboard.

hh:
Is there anything I can do to help test this PR? Is there general agreement on the approach?

PerBothner:
Whether the clipboard api is available is going to depend on which browser is used and its security settings. So we will almost certainly want a way to plug in an alternative mechanism. This might mean having to run a command, such as `pbpaste` (macOS) or `wl-copy` (Wayland).

aymanbagabas:
@Tyriar @jerch I've updated the PR to use an addon for the base64 implementation and read/write logic. Now, the `InputHandler` implements a clipboard event listener that listens to OSC52 events only if the `allowClipboardAccess` option is enabled (disabled by default). Any external module can hook into the listener to receive OSC52 events.

The changes made to `src/browser` allow registering a `ClipboardProvider` that handles incoming read/write events from the clipboard event listener. It then uses the registered `ClipboardProvider` to handle the clipboard event.

Let me know if you have any feedback :)

PerBothner:
It would be helpful for the `README.md` to include skeletal code for how to call out to an external service - which may be asynchronous. For example on some platforms DomTerm will send a side-channel request to the backend server, which will execute a helper program, which will send a response using a special escape sequence.  For "paste"  the example could call a `requestClipboardData` hook (implemented by the application), and handle a `handleRequestedClipboardData` callback (called by the application, implemented by the example).

aymanbagabas:
> It would be helpful for the `README.md` to include skeletal code for how to call out to an external service - which may be asynchronous. For example on some platforms DomTerm will send a side-channel request to the backend server, which will execute a helper program, which will send a response using a special escape sequence. For "paste" the example could call a `requestClipboardData` hook (implemented by the application), and handle a `handleRequestedClipboardData` callback (called by the application, implemented by the example).

There is a `README.md` for the addon. I just updated it to include an example of using a custom clipboard provider.

PerBothner:
Thanks for the example. It's not fully clear to be what I would need to do, but I think I can figure it out, when teh time comes.

src/common/InputHandler.ts
```diff
@@ -320,6 +323,7 @@ export class InputHandler extends Disposable implements IInputHandler {
     //  50 - Set Font to Pt.
     //  51 - reserved for Emacs shell.
     //  52 - Manipulate Selection Data.
+    this._parser.registerOscHandler(52, new OscHandler(data => this.setOrReportClipboard(data)));
```
Tyriar, line 326: @jerch I was thinking this should maybe be in the `Terminal.parser` API, but it might be better to just make it entirely an addon which registers the OSC 52 handler? I think that follows our general way of core has things that almost all terminals would want, and addons (whose code can easily be lazy loaded via code splitting) include optional behavior that not all terminals may want. If we did this we could remove `allowClipboardAccess` as well as a warning could be on the clipboard readme. Thoughts?

If we did this I think the base64 dependency should just be in the readme example, not as a true dep, since vscode for example would probably share its impl.

Tyriar (changes requested):
Sorry, I neglected this PR a bit 🙁

addons/xterm-addon-clipboard/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "repository": "https://github.com/xtermjs/xterm.js",
+  "license": "MIT",
+  "keywords": [
+    "terminal",
+    "xterm",
+    "xterm.js"
+  ],
+  "scripts": {
+    "build": "../../node_modules/.bin/tsc -p .",
+    "prepackage": "npm run build",
```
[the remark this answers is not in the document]
aymanbagabas: This already matches the format of other addon packages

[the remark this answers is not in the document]
Tyriar: Oh I was looking at image which is one of the different ones 😅

typings/xterm.d.ts
```diff
@@ -1842,4 +1861,29 @@ declare module 'xterm' {
...
+    writeText(selection: ClipboardSelection, data: string): Promise<void>;
+  }
+
+  /**
+   * Clipboard selection type.
+   */
+  export const enum ClipboardSelection {
+    SYSTEM = 'c',
+    PRIMARY = 'p',
+  }
```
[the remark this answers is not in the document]
aymanbagabas: Added a little comment explaining each enum. IMO system clipboard is clearer as it refers to the _system_ clipboard which is the same on all platforms. Plus `ClipboardSelection.CLIPBOARD` stutters and feels redundant.

hh:
woohoo!

src/common/InputHandler.ts
jerch: Yes making it fully provided from addon side sounds good to me.

hh:
@Tyriar I think merge is blocked until your review is cleared?

hh:
Or maybe it's merged but the Pr is stuck open?

image

Tyriar:
Trying to update branch now

Tyriar (changes requested):
I brought this up to date, including to use the new `@xterm` scoped modules. The only thing left to do here is move the stuff out of core and into the addon, so this:

```ts
this._parser.registerOscHandler(52, new OscHandler(data => this.setOrReportClipboard(data)));
```

And would be implemented on the addon side since it's a purely optional addition that we shouldn't include in core as the currently addon API is sufficient to implement it. I like the clipboard provider abstraction which would let an embedder use their own `base64` lib and customize the clipboard calls 👍

aymanbagabas:
> I brought this up to date, including to use the new `@xterm` scoped modules. The only thing left to do here is move the stuff out of core and into the addon, so this:
> 
> ```ts
> this._parser.registerOscHandler(52, new OscHandler(data => this.setOrReportClipboard(data)));
> ```
> 
> And would be implemented on the addon side since it's a purely optional addition that we shouldn't include in core as the currently addon API is sufficient to implement it. I like the clipboard provider abstraction which would let an embedder use their own `base64` lib and customize the clipboard calls 👍

We still need to access `this._onClipboard` to send clipboard events. Can we do that using the addon API? In the case of a clipboard read, we need to access `this.coreService.triggerDataEvent` to print the clipboard content to terminal see [this](https://github.com/xtermjs/xterm.js/pull/4220/files#diff-324424747eb9d66c9516d29665ad2ceadd440fcb275ea8f1128f146deb7ec11cR1309).

jerch:
> We still need to access `this._onClipboard` to send clipboard events. Can we do that using the addon API? In the case of a clipboard read, we need to access `this.coreService.triggerDataEvent` to print the clipboard content to terminal see [this](https://github.com/xtermjs/xterm.js/pull/4220/files#diff-324424747eb9d66c9516d29665ad2ceadd440fcb275ea8f1128f146deb7ec11cR1309).

Imho the easiest solution here would be to expose `triggerDataEvent` on the external API, maybe as `xterm.input(...)`? (Technically thats the "opposite" of `term.write` and the same as writing input to the backend)

l1n:
@aymanbagabas anything that would be useful for others to help with? would love to see this get in

hh:
I'l like to ring in here and keep this PR visible. Is there mailing list or community meeting we could use to get some final reviewers? OR create a new PR with necessary changes?

jerch:
reminder: Starting from v 5.4 `this.coreService.triggerDataEvent` is exposed as `Terminal.input` on the official API.

aymanbagabas:
I've updated the branch, however, I couldn't get the demo to work :/ Am I missing something? I keep getting this error:

```
TypeError: Cannot read properties of undefined (reading 'writeText')
    at OscHandler._setOrReportClipboard [as _handler] (ClipboardAddon.ts:51:33)
    at OscHandler.end (OscParser.ts:223:18)
    at OscParser.end (OscParser.ts:159:45)
    at EscapeSequenceParser.parse (EscapeSequenceParser.ts:777:43)
    at InputHandler.parse (InputHandler.ts:486:35)
    at WriteBuffer._action (CoreTerminal.ts:146:99)
    at WriteBuffer._innerWrite (WriteBuffer.ts:164:27)
    at WriteBuffer.ts:124:29
```

CC/ @Tyriar @jerch

jerch:
@aymanbagabas Sorry just saw your last comment - does that error still happen?

aymanbagabas:
> @aymanbagabas Sorry just saw your last comment - does that error still happen?

yes

addons/addon-clipboard/src/ClipboardAddon.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export class ClipboardAddon implements ITerminalAddon {
+  private _terminal?: Terminal;
+  private _disposable?: IDisposable;
+
+  constructor(private _provider: IClipboardProvider = new ClipboardProvider()) {}
+
+  public activate(terminal: Terminal): void {
+    this._terminal = terminal;
+    this._disposable = terminal.parser.registerOscHandler(52, this._setOrReportClipboard);
```
jerch, line 18: You error happens, because you insert `this._setOrReportClipboard` here unbound, thus the access to `this._provider` will fail. (And all other `this...` attempts)

Fix: Set the method via an arrow function, that way `this` will not mutate:
```typescript
terminal.parser.registerOscHandler(52, data => this._setOrReportClipboard(data));
```

aymanbagabas, line 18:
```suggestion
    this._disposable = terminal.parser.registerOscHandler(52, data => this._setOrReportClipboard(data));
```

aymanbagabas:
This is ready for squash and merge 🙂

jerch (changes requested):
Really nice, guess it is almost there. :+1: 

Sorry for playing the nitpicker here, but I have a few restructuring ideas, esp. around the custom provider API (I really like that concept):

- To me it seems reasonable to move the base64 stuff out of the provider, this way the provider would only have to care for its main task - read or write data from clipboard system xy (e.g. move base64 handling one level up close to the OSC handler).
- Provider API should also allow to return values in sync (e.g. return type would be `string | Promise<string>` for read), a certain provider impl can then pick whatever is suitable for its clipboard backend.
- Imho there are superfluous try-catch blocks - we typically dont catch sync errors in parser handlers, just let them bubble (terminal is still partially self-healing due to its chunk/callback-based nature).
- Other than sync errors, async errors (`.catch` path on a promise) should be handled explicitly, otherwise the stack trace gets messed up and hard to debug later on.
- The OSC handler should not filter out anything yet, but instead forward `pc` to the provider - this way a provider can also implement those other buffer types mentioned in the xterm docs. The default browser provider can just skip/ignore those requests and stick to handling `'c'` as clipboard selection type.
- default provider: I'd rename it to `BrowserClipboardProvider` to make clear, that it implements what a browser offers. The base64 argument should be removed, I'd also apply the limit setting to the reading side and give it a sane default (maybe 1 MB).

aymanbagabas:
Pushed the changes

> * default provider: I'd rename it to `BrowserClipboardProvider` to make clear, that it implements what a browser offers. The base64 argument should be removed, I'd also apply the limit setting to the reading side and give it a sane default (maybe 1 MB).

I've removed the limit because the xterm.js parser already limits the size of an OSC sequence data there's no need to have one in the default browser provider, custom providers can define their implementation with a limit.

Tyriar (approved):
This is looking good to me, I was careful to review the data we send over the pty and it looks like it's safe. It also doesn't touch anything in `./src` so it's low risk to merge.

@jerch any other comments before we merge?

jerch:
@Tyriar Nope, I think the PR is now good to be merged. I dont have a good understanding of the browser security measures applied for clipboard access. I think there might be a chance, that an earlier given permission will be auto-revoked by the browser, which might cause frictions for long running apps like vim, that try to do something with OSC 52. But as I am unsure about the impact of that, I think we should not try to fix it upfront, but if and when we get followup issues.

@aymanbagabas Thx a lot for getting this done, and your patience with us :heart:

jerch (approved):

abliss:
Hi! I and many others are very excited to try this out, but it still does not seem to be part of any release. Is our only option to build our own copy of xterm.js and the clipboard-addon from master?

jerch:
@abliss You can try `npm i @xterm/addon-clipboard`. For beta versions, see the list here: https://www.npmjs.com/package/@xterm/addon-clipboard?activeTab=versions.

Addons have a lifecycle if its own to some degree, so they are not tightly coupled to xterm.js releases.

PR #4233, opened 2022-10-27
Fix resize in demo

### LINKED ISSUE #4113: revamp resize logic in demo
If the browser has a zoomlevel (prolly for any pixelDeviceRatio != 1) the element resizing in `client.ts` does not yield the same values as `FitAddon.fit` leading to additional resize adjustments done by the fit addon (see https://github.com/xtermjs/xterm.js/issues/4112#issuecomment-1245231518). This prolly also happens in the opposite direction for shrinking and a zoomlevel <100% (not tested though).

We might want to fix that, so ppl dont c&p faulty code resulting in nonsense resize events with poor performance in the end.

(@Tyriar - you might want to test if something similar happens in vscode, if the pixelDeviceRatio is not perfectly 1).

Tyriar:
A double resize could happen due to the mismatch in canvas vs device pixel widths

Fixes #4113

### DIFF
--- a/demo/client.ts
+++ b/demo/client.ts
@@ -615,10 +615,8 @@ function addDomListener(element: HTMLElement, type: string, handler: (...args: a
 }
 
 function updateTerminalSize(): void {
-  const cols = parseInt((document.getElementById(`opt-cols`) as HTMLInputElement).value, 10);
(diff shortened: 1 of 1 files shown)


Tyriar:
This didn't actually fix the problem entirely

PR #4253, opened 2022-11-03
Clear extended attribute on 0

### LINKED ISSUE #4238: Underline attribute isn't being cleared correctly in demo buttons
It should be getting cleared here:

https://github.com/xtermjs/xterm.js/blob/25551a4b98615f34428c079c5cba7ad42feeeaea/demo/client.ts#L861

image

Tyriar:
Fixes #4238

@jerch this fixes it, is this the best way to clear out the extended attribute?


jerch:
I think #4216 will fix it already?

Tyriar:
Ah thanks, I thought that discussion was in an issue but couldn't find it

PR #4261, opened 2022-11-11
fix: fix link range in the double byte character scenario

robinfai:
image


jerch:
@Tyriar Is this a regression with the new linkifier? I thought this got fixed years ago by several PRs, but could have been with the old linkifier.

Issue #4173, opened 2022-10-03
Deletion on Android IME re-propigates previous composition


ink404:
After completing a composition on Android devices deleting re-triggers a composition and propagates the input with the old composition text. (see video below).

I was able to achieve the desired behavior by clearing the `_textarea` (by calling `_textarea.value = '';`) when compositions are finalized in `CompositionHelper` here: https://github.com/xtermjs/xterm.js/blob/master/src/browser/input/CompositionHelper.ts#L171.

Unfortunately this breaks the following test cases for Korean.
```
 1) CompositionHelper
       Input
         Should insert simple characters:

      Uncaught AssertionError: expected 'ㅇㅇㅇ' to equal 'ㅇㅇ'
      + expected - actual

      -ㅇㅇㅇ
      +ㅇㅇ
      
      at Timeout._onTimeout (src/browser/input/CompositionHelper.test.ts:65:22)
      at listOnTimeout (node:internal/timers:557:17)
      at processTimers (node:internal/timers:500:7)

  2) CompositionHelper
       Input
         Should insert complex characters:

      Uncaught AssertionError: expected '앙앙앙' to equal '앙앙'
      + expected - actual

      -앙앙앙
      +앙앙
      
      at Timeout._onTimeout (src/browser/input/CompositionHelper.test.ts:101:30)
      at listOnTimeout (node:internal/timers:557:17)
      at processTimers (node:internal/timers:500:7)

  3) CompositionHelper
       Input
         Should insert complex characters that change with following character:

      Uncaught AssertionError: expected '아' to equal '아아'
      + expected - actual

      -아
      +아아
      
      at Timeout._onTimeout (src/browser/input/CompositionHelper.test.ts:133:24)
      at listOnTimeout (node:internal/timers:557:17)
      at processTimers (node:internal/timers:500:7)
```

## Details
- Browser and browser version: Chrome, 106.x
- OS version: Android 13, GBoard
- xterm.js version: 5.x

### Steps to reproduce
1. type 'a'
2. tap space (finishes composition)
3. tap delete
4. Observe there are two a's now

##### Video:

https://user-images.githubusercontent.com/3418052/193627499-1c94d946-a772-4658-9110-7df57dcf2124.mp4

ink404:
@Tyriar If you have any suggestions for what changes to make to solve this I'm happy to make a PR, also please let me know if there's any additional information I can provide that would make this easier to debug.

RahjIII:
[...]

PR #4265, opened 2022-11-15
Clear text area after finalizing composition

ink404:
## Clear text area after finalizing composition
- Related issue: https://github.com/xtermjs/xterm.js/issues/4173

Previously stale text would display from text area in the composition helper when it heard the composition character. On some IMEs like GBoard this would manifest as the text propitiating after tapping delete, on others like the Samsung keyboard it would happen from tapping space. See attached videos for examples of before and after.

This PR clears the text area by setting `_textArea.value = ''` after finalizing the composition so when the composition is next started it does not contain the stale text and display it upon starting a new composition.

Previously I thought this change broke character composition, but after looking closer at `Compostionhelper.test.ts` and some additional manual testing to ensure everything works it seems the tests were checking for the presence of the uncleared `_textArea` text.

### Before

Gboard Example:

https://user-images.githubusercontent.com/3418052/201976368-f7215af3-a215-45b9-8aef-560f36b84418.mp4

Samsung Keyboard Example:

https://user-images.githubusercontent.com/3418052/201978752-ef45f952-e17b-4a79-ac77-28b91431dbe5.mp4

### After

Samsung Keyboard (kr):

https://user-images.githubusercontent.com/3418052/201982227-849909f2-6577-48eb-bdd5-1c0cc68fd1fd.mp4

Samsung Keyboard (en):

https://user-images.githubusercontent.com/3418052/201982181-934a24d5-4d47-42c7-b11c-bccb5ac33492.mp4

### DIFF
--- a/src/browser/input/CompositionHelper.test.ts
+++ b/src/browser/input/CompositionHelper.test.ts
@@ -58,7 +58,7 @@ describe('CompositionHelper', () => {
           // Second character 'ㅇ'
           compositionHelper.compositionstart();
           compositionHelper.compositionupdate({ data: 'ㅇ' });
-          textarea.value = 'ㅇㅇ';
+          textarea.value = 'ㅇ';
           setTimeout(() => { // wait for any textarea updates
             compositionHelper.compositionend();
             setTimeout(() => { // wait for any textarea updates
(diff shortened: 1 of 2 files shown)


Tyriar (commented):
Isn't this case meant to be handled by `currentCompositionPosition.start`? I'm trying to get us to clear the textarea less and make it more complete via shell integration to improve screen reader accessibility of the prompt.

Tyriar:
Closing for now as I believe this approach has negative impacts on screen reader accessibility.

PR #4271, opened 2022-11-22
Fixes disappearance of characters in webgl renderer

jeanp413:
Fixes https://github.com/microsoft/vscode/issues/166158


jeanp413:
Because `ITextureAtlas` is shared across terminals, if a GlyphRenderer for terminal A sets `hasCanvasChanged` to `false` then other terminals B, C won't call ` _bindAtlasPageTexture` and update the texture object in its respective webgl context

addons/xterm-addon-webgl/src/WebglUtils.ts
```diff
@@ -51,3 +51,13 @@ export function expandFloat32Array(source: Float32Array, max: number): Float32Ar
   }
   return newArray;
 }
+
+export class GLTexture {
+  public texture: WebGLTexture;
+  public dirtyId: number;
```
Tyriar, line 57: I think `version` would make more sense here

Tyriar (dismissed):
Makes sense, thanks. Just one comment that would make it a lot easier to understand imo

PR #4275, opened 2022-11-24
Don't rely on the cursor textarea for a11y

### LINKED ISSUE #4269: When screenReaderMode=true, keypress triggers unwanted default browser behaviors
## Details
- Browser and browser version: Chrome 109
- OS version: ChromeOS and others
- xterm.js version: ToT

### Steps to reproduce

1. set screenReaderMode=true
2. Press keys such as ctrl+p / ctrl+f. Observe that the printing / find dialog pops up, steals the focus, and makes the user's life miserable 😢 

This is caused by the code [here](https://github.com/xtermjs/xterm.js/blob/a351f5758a5126308b90d60b604b528462f6f051/src/browser/Terminal.ts#L1063), which cancels the key event only if screenReaderMode is off. I wonder if this is to allow the screen reader to handle some key bindings. I am not sure about other OSes, but on ChromeOS, screen reader bindings (e.g. Search+up) works fine even if the key event is canceled. I think we should just always cancel the key event. If there is indeed a need to pass through the key event, the user of xterm.js can set a custom key handler with `attachCustomKeyEventHandler()`. What do you think?

JasonXJ:
... but to use the AccessibilityManager live region to announce what the user is typing. This fixes #4269.

### DIFF
--- a/src/browser/AccessibilityManager.ts
+++ b/src/browser/AccessibilityManager.ts
@@ -34,17 +34,6 @@ export class AccessibilityManager extends Disposable {
   private _topBoundaryFocusListener: (e: FocusEvent) => void;
   private _bottomBoundaryFocusListener: (e: FocusEvent) => void;
 
-  /**
-   * This queue has a character pushed to it for keys that are pressed, if the
-   * next character added to the terminal is equal to the key char then it is
-   * not announced (added to live region) because it has already been announced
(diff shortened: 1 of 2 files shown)


Tyriar:
Pretty sure we can't do this because of https://github.com/xtermjs/xterm.js/issues/4269#issuecomment-1331375325

PR #4284, opened 2022-12-02
Fix a11y issues #4269

JasonXJ:


JasonXJ:
Hi @Tyriar ,  can you take a look at this new version?

Tyriar (approved):
Thanks, seems reasonable 👍

Issue #1824, opened 2018-12-10
Make scroll-to-bottom on input behaviour configurable


Eugeny:
xterm will unconditionally scroll to bottom on user input: https://github.com/xtermjs/xterm.js/blob/master/src/Terminal.ts#L1518

It'd be handy if this could be programmatically disabled.

Tyriar:
@Eugeny are you sure that's the right fix for https://github.com/Eugeny/terminus/issues/543? This particular case in VS Code is handled by this option which let's you force normal selection (not apply tmux selection) and copy that way:

https://github.com/xtermjs/xterm.js/blob/b5f1c3384ab8af6bd770cefdaa8274ee3d42e397/typings/xterm.d.ts#L150-L157

Eugeny:
Judging by 

>But with Terminus, when you press space, it doesn't let you enter select mode. Instead, it jumps you to the bottom of the tmux window.

I'd say that's the case

Tyriar:
@Eugeny but tmux doesn't have access to the scrollback, it maintains it's own scrollback buffer which is why this is confusing me. If you were using tmux's copy mechanism, you must be at the bottom of the (xterm.js) viewport?

Tyriar:
After more discussion in https://github.com/Eugeny/terminus/issues/1578, it's not to do with tmux apparently, people just want the option to disable it so accidentally hitting a key doesn't scroll.

Proposal for new option:

```ts
scrollOnInput?: boolean; // default true
```

jerch:
@Tyriar Not sure how this is currently coupled to input - is the scrolling triggered by keyboard input or by pty output (thus `term.write` input)? Imho most other emulator couple this only to latter one like xterm with its `Scroll to Bottom on TTY output` setting.

Tyriar:
@jerch :+1: the `Terminal` options should probably override the DECSET modes though?

```
  scrollOnWrite?: boolean;    // default undefined, respect DECSET 1010
  scrollOnKeypress?: boolean; // default true, override DECSET 1011
```

jerch:
@Tyriar xterm seems to allow to change that by apps on the fly. Imho for those things the user should always be in control - I agree with you.

Tyriar:
> Another way to deal with it and have DECSET support would be a 3-state option

[...]

PR #4289, opened 2022-12-06
Add option scrollOnUserInput

JasonXJ:
Note that we also change the existing behavior a bit: CoreService.triggerDataEvent() will not scroll to the bottom unless `wasUserInput` is also true. This actually seems to be the intended behavior according to the doc just above
ICoreService.triggerDataEvent().

Also see issue #1824

### DIFF
--- a/src/common/services/Services.ts
+++ b/src/common/services/Services.ts
@@ -87,7 +87,7 @@ export interface ICoreService {
    * @param data The data that is being emitted.
    * @param wasFromUser Whether the data originated from the user (as opposed to
    * resulting from parsing incoming data). When true this will also:
-   * - Scroll to the bottom of the buffer.s
+   * - Scroll to the bottom of the buffer if option scrollOnKeypress is true.
    * - Fire the `onUserInput` event (so selection can be cleared).
    */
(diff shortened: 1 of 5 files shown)


src/common/services/Services.ts
```diff
@@ -87,7 +87,7 @@ export interface ICoreService {
    * @param data The data that is being emitted.
    * @param wasFromUser Whether the data originated from the user (as opposed to
```
Tyriar, line 88: Could you fix this name?

typings/xterm.d.ts
```diff
@@ -193,6 +193,12 @@ declare module 'xterm' {
      */
     scrollback?: number;
 
+    /**
+     * Whether to scroll to the bottom whenever a key is pressed. The default is
+     * true.
+     */
+    scrollOnKeypress?: boolean;
```
Tyriar, line 200: I think `scrollOnUserInput` would be more accurate here as it's not necessarily a keypress that could trigger this.

Tyriar (changes requested):
Looks good, just some minor comments

JasonXJ:
I think this is good to merge now?

Tyriar (approved):
Looks good, thanks!

PR #4290, opened 2022-12-06
Send corresponding escape code for alt+space and ctrl+alt+space

JasonXJ:
The previous behavior is that this combinations do nothing.


JasonXJ:
Thew new behavior matches what gnome-terminal and hterm do.

Tyriar (commented):
@jerch does this seem correct to you?

jerch (approved):
Yes looks correct to me. I cannot really test any of them, as my window manager eats those combinations. But the basic idea of the Alt+mechanics is correct: send byte value of the pressed keys with `ESC` prepended:
- "space" ---> `0x20` == `SP`
- "alt-space" --> `0x1b20` == `ESC SP`
- "ctrl-space" ---> `0x00` == `NUL`
- "alt-ctrl-space" ---> `0x1b00` == `ESC NUL`

So I'd give this a go until someone claims, thats not the right Alt expansion for that particular key combinations.

PR #4324, opened 2022-12-16
Add allowNonHttpProtocols option to ILinkHandler

Ry0taK:
This pull request adds an option called `allowNonHttpProtocols` to `ILinkHandler`.
When this option is false, all non-HTTP URLs will be ignored in OscLinkProvider.

### DIFF
--- a/src/browser/OscLinkProvider.ts
+++ b/src/browser/OscLinkProvider.ts
@@ -69,8 +69,13 @@ export class OscLinkProvider implements ILinkProvider {
 
           let ignoreLink = false;
           if (!linkHandler?.allowNonHttpProtocols) {
-            const parsed = new URL(text);
-            if (!['http:', 'https:'].includes(parsed.protocol)) {
+            try {
+              const parsed = new URL(text);
+              if (!['http:', 'https:'].includes(parsed.protocol)) {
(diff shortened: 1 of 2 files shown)


Tyriar (approved):
Thanks 🙂

Tyriar:
Marking as (soft) breaking change to remember to include a note in the release notes

jerch:
Sorry to say that - but I find this addition w'o any discussion abit weird. Why is `http(s)` treated with prefiltering here? I know that hyperlinks are mostly about HTTP stuff, but thats not necessarily true anymore for some embedder with a limited use case.

If we want prefiltering on protocols, as this PR suggests, I think we should introduce a whitelist option like `allowed_protocols = ['http', ...]`, maybe defaulting to `['http', 'https']`. This way an embedder can customize the prefilter, eg. skip 'http' if their security policy demands it. Or allow only 'mailto' for whatever reason...

Tyriar:
This was a result of an email discussion. I wanted to ensure ultimate flexibility which this allows (just say true) while at the same time being safe by default, even if the embedder ignores the doc comment on linkHandler.

Tyriar:
The main concern is links like javascript:... can run

PR #4326, opened 2022-12-16
Fix NPE in webgl renderer

### LINKED ISSUE #4314: addon-webgl exception at line.loadCell(x, cell), line undefined
JS Exception: Cannot read properties of undefined (reading 'loadCell')
at [mapped JS].ISearchResult (addons/xterm-addon-webgl/src/WebglRenderer.ts:427)
at [mapped JS].ISearchResult (addons/xterm-addon-webgl/src/WebglRenderer.ts:394)
at [mapped JS].IColorZone (browser/services/RenderService.ts:153)
at [mapped JS].IColorZone (browser/services/RenderService.ts:86)
at [mapped JS].IColorZone (browser/RenderDebouncer.ts:91)
at [mapped JS].IColorZone (browser/RenderDebouncer.ts:70)

`line.loadCell(x, cell)`

https://github.com/xtermjs/xterm.js/blob/master/addons/xterm-addon-webgl/src/WebglRenderer.ts#L372

## Details
- Browser and browser version: Chrome 107, 108
- OS version: Win/Mac/ChOs/Linux
- xterm.js version: 5.0.0

### Steps to reproduce

Unclear, likely a race condition, happens to 10% of users 1-2 times per session

Tyriar:
Theory is this can happen when a resize down happens as end will be too big

See microsoft/vscode#166878
Fixes #4314

### DIFF
--- a/addons/xterm-addon-webgl/src/WebglRenderer.ts
+++ b/addons/xterm-addon-webgl/src/WebglRenderer.ts
@@ -351,7 +351,7 @@ export class WebglRenderer extends Disposable implements IRenderer {
     let lastBg: number;
     let y: number;
     let row: number;
-    let line: IBufferLine;
+    let line: IBufferLine | undefined;
     let joinedRanges: [number, number][];
     let isJoined: boolean;
     let lastCharX: number;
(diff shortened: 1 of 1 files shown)


jerch:
@Tyriar Would it be worth to further track down the reason for the lines numbers provided being off?

I'd suspect a race condition from resizing as well, but doesnt that mean that by opposite resizing the numbers are also off? This ofc would not run into an NPE, but still apply wrong numbers for whatever reason `_updateModel` was called, prolly missing model updates (eg. from resizing - skipping updates on higher lines).

Tyriar:
I don't think the opposite is a concern as a full screen update would have been requested after the resize, I guess we could remove this workaround and then validate the end value before using it in render service or debouncer?

Tyriar:
Reopened the issue to do that, that way we would still get the exception if something else is happening

PR #4335, opened 2022-12-19
Cleanup Buffer.ts

### LINKED ISSUE #4331: Remove Buffer.iterator
@Tyriar

Since the old linkifier is gone, `Buffer.iterator` (and subsequentially `BufferStringIterator` and `Buffer.stringIndexToBufferIndex`) is not used by any code anymore. I think it can be removed. Also it is not exposed to outside, furthermore I think that this Java-ish iterator was not the best idea. (In case we want a similar feature in the future, maybe it could be achieved more idiomatic by a generator pattern.)

jerch:
Fixes #4331.


Tyriar (approved):
I like deleted code 🎉

jerch:
> I like deleted code :tada: 

`rm -rf /` is your friend :see_no_evil:
(Seriously - happened to me once in the past on a remote server connection, and I think the keyboard got bite marks from that incident. So yeah backups are a pretty decent thing, if you have one :rofl:)

PR #4342, opened 2022-12-20
Fix link event listener leak

### LINKED ISSUE #4341: Terminal can lock up when typing if the cursor is hovering a link
VS Code issue: https://github.com/microsoft/vscode/issues/169614

Repros in xterm.js:

1. Write a link
2. Hover it
3. Type slowly, it will get exponentially slower each character

Tyriar:
The bug here is that _clearCurrentLink was being called only on the line(s) that changed, but _askForLink was then being triggered regardless. This caused more onRenderedViewportChange listeners to be registered and for the thread blocking to get exponentially worse.

Fixes #4341

### DIFF
--- a/src/browser/Linkifier2.ts
+++ b/src/browser/Linkifier2.ts
@@ -309,8 +309,8 @@ export class Linkifier2 extends Disposable implements ILinkifier2 {
         }
       });
 
-      // Listen to viewport changes re-render the link under the cursor when the viewport line
-      // changes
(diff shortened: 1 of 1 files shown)


jerch:
Oh wow - kudos to your test batteries finding this one. :scream: :see_no_evil:

Tyriar:
@jerch yeah it took a while to figure this one out, also spotted a race while digging into it https://github.com/microsoft/vscode/pull/169696

Issue #2488, opened 2019-10-20
Character size computation is affected by CSS transforms


cebamps:
When a Terminal is opened in an element subject to CSS transformations such as `scale` or `rotate`, the computation of the character size is affected by that transformation, which is wrong because the rendering should be transform-agnostic.

As a result, this makes the integration of Xterm.js into reveal.js slides render incorrectly when the slides are scaled down to fit the browser viewport.

Changing the measurement method in `CharSizeService` to use `getComputedStyle` instead of `getBoundingClientRect` makes the computation independent of CSS transforms.

## Details
- Browser and browser version: Firefox 69.0 and Chromium 77.0
- OS version: Arch Linux
- xterm.js version: current master (6311076)

### Steps to reproduce

1. Add `style="transform: rotate(90deg);"` to the `<body>` tag in `demo/index.html`
2. Notice that character boxes in the demo have the wrong dimensions

### Workaround

With the `rotate(90deg)` CSS transform Xterm.js renders character cells with the width and height swapped around:
![getBoundingClientRect](https://user-images.githubusercontent.com/12797742/67163766-28a1ed00-f373-11e9-953d-f3894e5550e3.png)

The cause is in the measurement methods in `CharSizeService`: it uses `getBoundingClientRect`, which returns the size of the bounding box *after* CSS transforms.
https://github.com/xtermjs/xterm.js/blob/631107698ec1f28577101cc570540f097a83af68/src/browser/services/CharSizeService.ts#L71-L86

My understanding is that the rendering of the terminal should not be affected by said transforms. And indeed, replacing the `geometry` definition in the code above with (POC)
```ts
    const computedStyle = getComputedStyle(this._measureElement);
    const geometry = {
      height: Number(computedStyle.getPropertyValue('height').replace('px', '')),
      width: Number(computedStyle.getPropertyValue('width').replace('px', '')),
    }
```
yields a correct rendering:
![getComputedStyle](https://user-images.githubusercontent.com/12797742/67163767-28a1ed00-f373-11e9-9f16-289156334758.png)

Note that my proof-of-concept fix above may be insufficient in cases where the element or its parent are `display: none`.

Tyriar:
@cebamps the workaround seems reasonable.

> Note that my proof-of-concept fix above may be insufficient in cases where the element or its parent are display: none.

I don't think this should matter.

leomoty:
I am not really able to reproduce. The selection with cursor is broken when you rotate the terminal.

[...]

PR #4366, opened 2023-01-06
Fix for #2488. Use element offsetHeight rather than getBoundingRect

arekouzounian:
This was a simple fix for #2488, as using the offsetHeight/Width resolved the issue. 

I tried to create a quick test for the patch by having a button that would apply the transform to the body, but applying the transform after the page had loaded didn't cause any issues with character size, even without the patch applied. 

I decided just to leave out the button, because it didn't seem to help as a test.

This is my first open-source contribution ever, so any and all feedback is much appreciated. Thank you!


Tyriar (approved):
Thanks @arekouzounian! Sorry about taking so long for your first ever contributor

jerch:
Eww, looking at this PR again - plz note that `offsetWidth` & `offsetHeight` are defined as integer, where `getBoundingClientRect` gives fractions. Ofc ideally we want the cells to be spaced in integers, but I am not sure if the dimension calculations in the renderers still will yield the same numbers here.

PR #4382, opened 2023-01-23
attach and detach accessibility tree root element

meganrogge:
Explains why navigation mode was not working and gets the output reading back to the state it was in - only announcing when text is too long to read

[172007](https://github.com/microsoft/vscode/issues/172007)

### DIFF
--- a/src/browser/AccessibilityManager.ts
+++ b/src/browser/AccessibilityManager.ts
@@ -93,7 +93,7 @@ export class AccessibilityManager extends Disposable {
     this._accessiblityBuffer.ariaLabel = Strings.accessibilityBuffer;
     this._accessiblityBuffer.classList.add('xterm-accessibility-buffer');
 
-    // TODO: this is needed when content editable is false
+    // HACK: this is needed when content editable is false
     this._refreshAccessibilityBuffer();
     this._accessiblityBuffer.addEventListener('focus', () => this._refreshAccessibilityBuffer());
     this._terminal.element.insertAdjacentElement('afterbegin', this._accessiblityBuffer);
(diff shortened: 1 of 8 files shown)


src/browser/AccessibilityManager.ts
```diff
@@ -57,6 +57,7 @@ export class AccessibilityManager extends Disposable {
   ) {
     super();
     this._accessibilityTreeRoot = document.createElement('div');
+    this._terminal.element?.appendChild(this._accessibilityTreeRoot);
```
Tyriar, line 60: Can we switch this to what it was before?

https://github.com/xtermjs/xterm.js/blob/19c760a20470c26666efad770fe82496c4a9bf1e/src/browser/AccessibilityManager.ts#L87

If that position isn't good, at least to the same line?

src/browser/AccessibilityManager.ts
```diff
@@ -121,6 +122,7 @@ export class AccessibilityManager extends Disposable {
     // media query that drives the ScreenDprMonitor isn't supported
     this.register(addDisposableDomListener(window, 'resize', () => this._refreshRowsDimensions()));
     this.register(toDisposable(() => {
+      this._terminal.element?.removeChild(this._accessibilityTreeRoot);
       this._accessiblityBuffer.remove();
       this._accessibilityTreeRoot.remove();
```
Tyriar, line 127: This should be covered already by `this._accessibilityTreeRoot.remove();`

Tyriar (changes requested):

src/browser/AccessibilityManager.ts
meganrogge: good find, yes 👍🏼 looks like that happened in Dec

src/browser/AccessibilityManager.ts
```diff
@@ -121,6 +122,7 @@ export class AccessibilityManager extends Disposable {
     // media query that drives the ScreenDprMonitor isn't supported
     this.register(addDisposableDomListener(window, 'resize', () => this._refreshRowsDimensions()));
     this.register(toDisposable(() => {
+      this._terminal.element?.removeChild(this._accessibilityTreeRoot);
```
meganrogge, line 125:
```suggestion
```

src/browser/AccessibilityManager.ts
meganrogge, line 60:
```suggestion
    this._terminal.element.insertAdjacentElement('afterbegin', this._accessibilityTreeRoot);
```

Tyriar: Looks like my prototype broke it: https://github.com/xtermjs/xterm.js/commit/41dc93a008dc7bc4044020cfa483210891770b50

meganrogge:
so this was due to our accessibility buffer changes

https://github.com/xtermjs/xterm.js/commit/41dc93a008dc7bc4044020cfa483210891770b50#diff-7904a8cc1e295a78532b41608480fddf931f80c485c53b78f99ae8cd5759dedcL86

PR #4383, opened 2023-01-24
fix accessibility dom structure

meganrogge:
when this is adjacent, it messes up the accessibility buffer such that arrow keys do nothing 

https://github.com/microsoft/vscode/issues/172222

### DIFF
--- a/src/browser/AccessibilityManager.ts
+++ b/src/browser/AccessibilityManager.ts
@@ -57,7 +57,6 @@ export class AccessibilityManager extends Disposable {
   ) {
     super();
     this._accessibilityTreeRoot = document.createElement('div');
-    this._terminal.element?.appendChild(this._accessibilityTreeRoot);
     this._accessibilityTreeRoot.classList.add('xterm-accessibility');
     this._accessibilityTreeRoot.tabIndex = 0;
 
@@ -89,6 +88,7 @@ export class AccessibilityManager extends Disposable {
     if (!this._terminal.element) {
       throw new Error('Cannot enable accessibility before Terminal.open');
     }
+    this._terminal.element?.appendChild(this._accessibilityTreeRoot);
 
     this._accessiblityBuffer = document.createElement('textarea');
     this._accessiblityBuffer.ariaLabel = Strings.accessibilityBuffer;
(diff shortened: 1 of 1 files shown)


src/browser/AccessibilityManager.ts
```diff
@@ -57,7 +57,7 @@ export class AccessibilityManager extends Disposable {
   ) {
     super();
     this._accessibilityTreeRoot = document.createElement('div');
-    this._terminal.element?.insertAdjacentElement('afterbegin', this._accessibilityTreeRoot);
+    this._terminal.element?.appendChild(this._accessibilityTreeRoot);
```
Tyriar, line 60: This line was meant to go after:

```ts
    this._accessibilityTreeRoot.appendChild(this._liveRegion);
    if (!this._terminal.element) {
      throw new Error('Cannot enable accessibility before Terminal.open');
    }
```

It's best to add the element to the dom after it's all set up to avoid extra overhead that may come with changing elements when they're attached.

Tyriar (changes requested):

src/browser/AccessibilityManager.ts
```diff
@@ -89,6 +88,7 @@ export class AccessibilityManager extends Disposable {
     if (!this._terminal.element) {
       throw new Error('Cannot enable accessibility before Terminal.open');
     }
+    this._terminal.element?.appendChild(this._accessibilityTreeRoot);
```
Tyriar, line 91: It needs to be inserted as the first element via `afterbegin` in order to be before the textarea's focus, so shift+tab naturally focuses it. This also helps with the layers, it's before the renderer so it's underneath it, unless z-index is set.

Tyriar (changes requested):

src/browser/AccessibilityManager.ts
meganrogge: that breaks the accessibility buffer arrow navigation, will have to investigate more after testing

Tyriar: But that's how it was before when it worked? 😕 Moving it in front could cause other issues

meganrogge: we didn't have an accessibility buffer before

Tyriar: This is for _accessibilityTreeRoot which is the element that holds the navigation mode element too though

Tyriar (approved):

PR #4384, opened 2023-01-25
defer focus when accessibility buffer is active

meganrogge:
When `escape` or `tab` is pressed, exit the accessibility buffer. Otherwise, disallow focusing of the terminal text area 

https://github.com/microsoft/vscode/issues/172200

### DIFF
--- a/src/browser/AccessibilityManager.ts
+++ b/src/browser/AccessibilityManager.ts
@@ -93,10 +93,16 @@ export class AccessibilityManager extends Disposable {
     this._accessiblityBuffer = document.createElement('textarea');
     this._accessiblityBuffer.ariaLabel = Strings.accessibilityBuffer;
     this._accessiblityBuffer.classList.add('xterm-accessibility-buffer');
+    this.register(addDisposableDomListener(this._accessiblityBuffer!, 'keydown', (ev: KeyboardEvent) => {
+      if (ev.key === 'Escape' || ev.key === 'Tab') {
+        this._terminal?.textarea?.focus();
+      }}
+    ));
     this._accessiblityBuffer.addEventListener('focus', () => this._refreshAccessibilityBuffer());
     this._accessibilityTreeRoot.appendChild(this._accessiblityBuffer);
 
     this.register(this._renderRowsDebouncer);
+    this.register(this._terminal.onFocus(() => this._refreshAccessibilityBuffer()));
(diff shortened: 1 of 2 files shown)


src/browser/AccessibilityManager.ts
```diff
@@ -265,7 +271,7 @@ export class AccessibilityManager extends Disposable {
     // Only add the char if there is no control character.
     if (!/\p{Control}/u.test(keyChar)) {
       this._charsToConsume.push(keyChar);
-    }
+    }
```
meganrogge, line 274:
```suggestion
    }
```

src/browser/Terminal.ts
```diff
@@ -259,6 +259,9 @@ export class Terminal extends CoreTerminal implements ITerminal {
    * Focus the terminal. Delegates focus handling to the terminal's DOM element.
    */
   public focus(): void {
+    if (document.querySelector('.xterm-accessibility-buffer:focus')) {
```
meganrogge, line 262: better way? don't like that it's in this general code area

Tyriar: What is this needed for? We don't want Terminal to know about internals of AccessibilityManager

Tyriar (changes requested):

src/browser/Terminal.ts
meganrogge: needed to prevent the `textarea` from stealing the focus when the `.xterm-accessibility-buffer` is focused. what's weird is that the `z-index` should be preventing that from happening on click

meganrogge: also tried 

```
 this._accessiblityBuffer.addEventListener('click', (ev: MouseEvent) => {
      ev.stopImmediatePropagation();
    });
```

but it still goes to the `textarea`

Tyriar: There should be some way of preventing this, we definitely don't want to mention the `xterm-accessibility-buffer` class outside of AccessibilityManager.ts

Tyriar: There should be some way of preventing this, we definitely don't want to mention the `xterm-accessibility-buffer` class outside of AccessibilityManager.ts

meganrogge: yep I agree, wip

PR #4386, opened 2023-01-28
split extremely wide ligatures to fit atlas

### LINKED ISSUE #4362: Infinite loop when drawing large ligatures to char atlas
some fonts provide extremely long ligatures, e.g. Fira Code for `===================================================`
The char atlas code tries to find a row to fit it but fails and gets stuck in an infinite loop
https://github.com/xtermjs/xterm.js/blob/fe8981766769c19ab1903dffa19f1a64ad83da13/src/browser/renderer/shared/TextureAtlas.ts#L792-L799
^Here we're checking if there's room on the current row and creating a new one if not.

## Details
- Browser and browser version: chrome 108
- OS version: macOS 12
- xterm.js version: 5.1.0 (also on latest master branch)

### Steps to reproduce

1. Install Fira Code font, Run the xterm demo and enable ligatures
2. start typing `=`
3. Keep looking at the char atlas drawn below, once the glyph is almost the same width as char atlas, typing the next `=` will freeze everything.

Screenshot 2023-01-02 at 13 32 47

LabhanshAgrawal:
Fixes #4362 

Have fixed it for the canvas renderer, got confused with the webgl code, can try it again if someone can give few pointers on how to go about it.


LabhanshAgrawal:
@meganrogge @Tyriar can you please review this and let me know if you have any suggestions for webgl renderer?

LabhanshAgrawal:
got webgl to also work somehow. It's working only when there's something else at the end e.g. '=========a' but the last char disappears. Needs some more work I guess, but still not fully clear on how the webgl renderer works. @Tyriar any suggestions?

src/browser/renderer/shared/TextureAtlas.ts
```diff
@@ -678,157 +681,183 @@ export class TextureAtlas implements ITextureAtlas {
...
-    const imageData = this._tmpCtx.getImageData(
-      0, 0, this._tmpCanvas.width, this._tmpCanvas.height
-    );
 
-    // Clear out the background color and determine if the glyph is empty.
-    let isEmpty: boolean;
-    if (!this._config.allowTransparency) {
-      isEmpty = clearColor(imageData, backgroundColor, foregroundColor, enableClearThresholdCheck);
+    // split the image into multiple images to fit into the texture
+    const images: ImageData[] = [];
```
Tyriar, line 686: Returning an array of images here will complicate things a lot, can we instead force glyphs to have a maximum width of `_textureSize`? 512px is a lot even on a high device pixel ratio and seems like a reasonable place to stop

Tyriar (changes requested):

src/browser/renderer/shared/TextureAtlas.ts
LabhanshAgrawal: for most glyphs, it'd be an array with a single item, we will have splits only when the image is bigger than texture size. Limiting glyphs to texture size would work mostly, but what to do with the huge ligatures, then?

jerch: To do one step back - is the potentially endless stacking into a bigger and bigger glyph the expected behavior from ligatures, it that transporting the right sematics? How do other output systems (TEs or editors) deal with that?

If its indeed intended, then we have only one "natural" upper bound - the current width of the viewport. Not sure how much hassles it imports - could those big glyphs be rendered from a slow path (eg. skipping the texture atlas logic, instead doing on-the-fly composition)?

Tyriar: Considering how unimportant a ligature for `================================================>` seems to be, I think we should just not support that case and revert to disabling ligatures when it would exceed the width of a texture atlas page. Too much complexity would be introduced (particularly dealing with the special texture in webgl) having a special path that creates a new larger special texture just for this obscure case.

LabhanshAgrawal: Sorry, was caught up with things, I will try to update the pr this week to limit ligature size instead.
Thanks

Tyriar:
Closing as it's stale

PR #4387, opened 2023-01-29
Enable ligatures addon on electron

LabhanshAgrawal:
No ligatures were being detected on electron as it was getting skipped due to `process` being defined.


LabhanshAgrawal:
@meganrogge @Tyriar can you please review this

Tyriar (approved):

PR #4388, opened 2023-01-30
make the accessibility buffer readonly

meganrogge:
but still accept arrow keys

### DIFF
--- a/src/browser/AccessibilityManager.ts
+++ b/src/browser/AccessibilityManager.ts
@@ -94,6 +94,9 @@ export class AccessibilityManager extends Disposable {
     this._accessiblityBuffer.ariaLabel = Strings.accessibilityBuffer;
     this._accessiblityBuffer.classList.add('xterm-accessibility-buffer');
     this._accessiblityBuffer.addEventListener('focus', () => this._refreshAccessibilityBuffer());
+    this._accessiblityBuffer.addEventListener('keydown', (e) => this._handleAccessibilityBufferKey(e));
+    this._accessiblityBuffer.addEventListener('keyup', (e) => this._handleAccessibilityBufferKey(e));
+    this._accessiblityBuffer.addEventListener('keypress', (e) => this._handleAccessibilityBufferKey(e));
     this._accessibilityTreeRoot.appendChild(this._accessiblityBuffer);
 
     this.register(this._renderRowsDebouncer);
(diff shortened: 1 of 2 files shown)


src/browser/AccessibilityManager.ts
```diff
@@ -126,6 +129,14 @@ export class AccessibilityManager extends Disposable {
     }));
   }
 
+  private _handleAccessibilityBufferKey(event: KeyboardEvent): void {
+    if (['ArrowUp', 'ArrowLeft', 'ArrowRight', 'ArrowDown', 'Tab', 'Shift'].includes(event.key)) {
+      return;
+    }
+    event?.preventDefault();
+    event?.stopImmediatePropagation();
```
meganrogge, line 137:
```suggestion
    event.preventDefault();
    event.stopImmediatePropagation();
```

src/browser/AccessibilityManager.ts
```diff
@@ -126,6 +129,14 @@ export class AccessibilityManager extends Disposable {
     }));
   }
 
+  private _handleAccessibilityBufferKey(event: KeyboardEvent): void {
+    if (['ArrowUp', 'ArrowLeft', 'ArrowRight', 'ArrowDown', 'Tab', 'Shift'].includes(event.key)) {
+      return;
+    }
+    event.preventDefault();
+    event.stopImmediatePropagation();
```
Tyriar, line 137: There are other ways to navigate text (page up, home, etc.). Do we need to prevent writing like this? Restricting it in this way seems more harmful than just allowing editing it?

meganrogge: We have heard from screen reader users in the issue linked and similar that mutating the buffer is unexpected and not desirable.

meganrogge: By only allowing certain keys, we can make sure that doesn't happen while not messing up the scrolling via the `readOnly` property

Tyriar: Something I was thinking about recently was that with the move to a textarea, that makes the move to a semantic buffer not possible as you cannot have elements inside a textarea's value. So I think we should go back to a tabindex div with a contenteditable setting anyway (for non-screen readers, until we get F7 caret browsing mode)

meganrogge: when it's not content editable, screen readers cannot do anything w a div, but we can do something like this PR does to prevent unwanted edits

meganrogge: I also don't think non-screen reader users will gain anything from this unless it is content editable, so we should probably just go with that for everyone.

PR #4389, opened 2023-01-31
accessibilityBuffer: textarea -> div & prevent mouse click from exiting 

meganrogge:
fixes https://github.com/microsoft/vscode/issues/172200


src/browser/AccessibilityManager.ts
```diff
@@ -36,6 +36,8 @@ export class AccessibilityManager extends Disposable {
   private _topBoundaryFocusListener: (e: FocusEvent) => void;
   private _bottomBoundaryFocusListener: (e: FocusEvent) => void;
 
+  public accessibilityBufferActive: boolean = false;
```
Tyriar, line 39: private boolean and public getter would be best to avoid mutation from outside

Tyriar (changes requested):

PR #4400, opened 2023-02-03
pull `AccessibleBuffer` out of `AccessibilityManager`

meganrogge:
Not sure where the code that makes `Shift Tab` enter this mode is, but it doesn't work when not using screen reader mode. Perhaps that's fine though and will leave it up to the embedders

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -33,6 +33,7 @@ import * as Browser from 'common/Platform';
 import { addDisposableDomListener } from 'browser/Lifecycle';
 import * as Strings from 'browser/LocalizableStrings';
 import { AccessibilityManager } from './AccessibilityManager';
+import { AccessibleBuffer } from './AccessibleBuffer';
 import { ITheme, IMarker, IDisposable, ILinkProvider, IDecorationOptions, IDecoration } from 'xterm';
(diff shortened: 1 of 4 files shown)


src/browser/Terminal.ts
```diff
@@ -576,6 +578,8 @@ export class Terminal extends CoreTerminal implements ITerminal {
     // Listen for mouse events and translate
     // them into terminal mouse protocols.
     this.bindMouse();
+
```
meganrogge, line 581: I had to put this here so that dimensions from the render service wouldn't be 0. lmk if there's a better place

Tyriar: This is the right function, I'd put it just below the `AccessibilityManager`s init to be a little more organized. If you needed to move it here then something must not have been migrated to the new file correctly as it was working before when bundled into `AccessibilityManager`.

Unless that only worked when toggling screenReaderMode on after Terminal was `open`ed? In that case there's probably some other event we're not listening to?

meganrogge: that's where I put it originally and it didn't work, so seems like we're missing something

meganrogge: it needs to be after this to work

```
    this._charSizeService.measure();
```

PR #4402, opened 2023-02-06
add basics

meganrogge:


typings/xterm.d.ts
```diff
@@ -1108,6 +1108,13 @@ declare module 'xterm' {
      * @param addon The addon to load.
      */
     loadAddon(addon: ITerminalAddon): void;
+
+
+    /**
+     * Sets the accessibility buffer children.
```
meganrogge, line 1114:
```suggestion
     * Sets the accessibility buffer's children elements.
```

typings/xterm.d.ts
```diff
@@ -1108,6 +1108,13 @@ declare module 'xterm' {
      * @param addon The addon to load.
      */
     loadAddon(addon: ITerminalAddon): void;
+
+
+    /**
+     * Sets the accessibility buffer's child elements.
+     * @param elements The elements to add to the accessibility buffer.
+     */
+    setAccessibilityBufferElements(elements: HTMLElement[]): DocumentFragment;
```
Tyriar, line 1117: Copying my comment from https://github.com/microsoft/vscode/pull/173605

Is it possible for this API to be the other way around? Like the below such that xterm.js requests the elements only when needed?

```ts
registerBufferElementProvider(): DocumentFragment;
```

meganrogge: yeah that makes more sense 👍🏼

src/browser/AccessibleBuffer.ts
```diff
@@ -51,19 +55,43 @@ export class AccessibleBuffer extends Disposable {
...
+    return {
+      dispose: () => {
+        const providerIndex = this._providers.indexOf(bufferProvider);
+
+        if (providerIndex !== -1) {
+          this._providers.splice(providerIndex, 1);
+        }
+      }
+    };
+  }
```
Tyriar, line 69: Hard to think about what multiple of these would do, should we throw when a second is attempted to register instead of trying to support multiple?

src/browser/Terminal.ts
```diff
@@ -193,6 +193,13 @@ export class Terminal extends CoreTerminal implements ITerminal {
     }));
   }
 
+  public registerBufferElementProvider(bufferProvider: IBufferElementProvider): IDisposable {
+    if (!this._accessibleBuffer) {
+      throw new Error ('No accessible buffer');
```
Tyriar, line 198: Should this be the error we throw when the terminal isn't opened yet?

typings/xterm.d.ts
```diff
@@ -1108,6 +1108,13 @@ declare module 'xterm' {
      * @param addon The addon to load.
      */
     loadAddon(addon: ITerminalAddon): void;
+
+
+   /**
+    * Registers a buffer element provider
+    * @param bufferProvider 
+    */
+    registerBufferElementProvider(bufferProvider: IBufferElementProvider): IDisposable;
```
Tyriar, line 1117:
```suggestion
    * Registers a buffer element provider which will provide a custom element to
    * use in the terminal's accessible buffer. Using this an xterm.js embedder
    * has full control over how the buffer is represented in the DOM.
    * @param bufferElementProvider The buffer element provider.
    */
    registerBufferElementProvider(bufferElementProvider: IBufferElementProvider): IDisposable;
```

typings/xterm.d.ts
```diff
@@ -1360,6 +1367,14 @@ declare module 'xterm' {
     getNullCell(): IBufferCell;
   }
 
+  export interface IBufferElementProvider {
+    /**
+     * Provides a buffer element provider
+     * @param callback The callback to be fired when the buffer elements are requested
+     */
+    provideBufferElements(callback: (fragment: DocumentFragment) => void): void;
```
Tyriar, line 1375: Does this need to be a callback instead of just returning `DocumentFragment`?

Tyriar (changes requested):

Tyriar (approved):

PR #4403, opened 2023-02-07
ensure `line` is defined via clamp

meganrogge:
fixes https://github.com/microsoft/vscode/issues/161047


addons/xterm-addon-webgl/src/WebglRenderer.ts
```diff
@@ -369,10 +369,13 @@ export class WebglRenderer extends Disposable implements IRenderer {
...
     for (y = start; y <= end; y++) {
       row = y + terminal.buffer.ydisp;
-      line = terminal.buffer.lines.get(row)!;
+      line = terminal.buffer.lines.get(row);
       this._model.lineLengths[y] = 0;
       joinedRanges = this._characterJoinerService.getJoinedCharacters(row);
       for (x = 0; x < terminal.cols; x++) {
+        if (!line) {
+          continue;
+        }
```
Tyriar, line 378: This would be better just below where `line` gets set, but we could probably fix this in a nicer way by `clamping` both start and `end` between `0` and `rows - 1`?

This function could be copied into this file to do that: https://github.com/xtermjs/xterm.js/blob/d5b5f1a9a4b58098dc52fa6d09bfde3d92f40de0/src/browser/renderer/shared/CustomGlyphs.ts#L644-L646

Tyriar (changes requested):

Tyriar (approved):
:+1:

If we ever see this again after this is merged then something is really wrong as `lines.get()` should always return something between `ydisp` and `ydisp + rows - 1`

Issue #702, opened 2017-06-13
Add character dimensions API


Tyriar:
VS Code has it's own character measure implementation for example. @parisk thoughts?

Related: https://github.com/sourcelair/xterm.js/issues/696

Tyriar:
Closing in favor of https://github.com/xtermjs/xterm.js/issues/1128#issuecomment-394142177 which will necessitate opening up the API more.

Tyriar:
Just did a bit of research into what we actually want here, this is what I came up with.

- VS Code:
  - TerminalConfigHelper.getFont
    - Uses charMeasure.width and charMeasure.height
    - Bug: Panes don't take line height and letter spacing into account
  - TerminalInstance._evaluateColsAndRows
    - Uses getFont
    - Calculates scaledCharWidth (with letter spacing for canvas) and scaledLineHeight. These are used to set cols and rows. These are scaledCellWidth/Height
    - Could use actualCellHeight/Width
    - Dimensions is not ready when we need it in _evaluateColsAndRows. Related to https://github.com/xtermjs/xterm.js/issues/1158, vscode actually calls open before it's attached and while it's probably possible to attach earlier I don't think I could make it visible which breaks CharMeasure.measure as well.
- fit
  - Uses actualCellWidth
  - Uses actualCellHeight
- Hyper
  - Calls CharMeasure.measure manually (possibly for the same open issue as vscode)
  - Uses fit addon

```ts
interface IDimensions {
  width: number;
  height: number;
}

export class Terminal {
  dimensions: {
    cell: IDimensions;
    // Convenience: container.width = cell.width*terminal.cols
    // Generic language
    container: IDimensions;
  }
}
```

My proposal to fix the `.open` issue is to allow a helper container to be passed into the constructor as an option.

```ts
export interface ITerminalOptions {
  measureContainer: HTMLElement;
}
```

`measureContainer` could be used as a container to host (and share) the `.xterm-char-measure-element`. Allowing characters to be measured during the call to `open`. The only thing to worry about here is CSS styles that differ on `measureContainer` and `Terminal.element`. We will probably want to be very explicit in the comment for `measureContainer` as well as adding and needed styles to `.xterm-char-measure-element` in `xterm.css`.

Tyriar:
We probably actually want to expose the dimensions on the renderer and eventually phase out `CharSizeService` all together if possible.

Tyriar:
[...]

PR #4421, opened 2023-02-22
Add measureParent flag to FitAdd.

PerBothner:
The default is true, which causes the 'fit' function to measure the xterm.element.parentElement (the old behavior); if the setting is false, the xterm.element itself is measured. One or the other may be preferable, depending on how xterm is embedded.

Background: I have added an option to [DomTerm](https://domterm.org) to use xterm.js for the terminal emulator (in places of DomTerm's native terminal emulator): [see here](https://github.com/PerBothner/DomTerm/blob/master/README-xtermjs.md).  DomTerm supports tabs and tiles, using [my fork of GoldenLayout](https://github.com/PerBothner/golden-layout),  The way these tiles are managed clashes with how the `FitAddon` checks the the size of the parent of the xterm element. The changes adds an optional boolean to the `FitAddon` constructor. The default (`true`) preserves the old behavior.  Passing `false` to the constructor fixes `fit`.

It is possible a default of `false` would make more sense, but that might break some embeddings. It is possible I am missing some fundamental assumtion that the xterm element is the only child of its parent - though I am hoping we can avoid that extra level of nesting.

### DIFF
--- a/addons/xterm-addon-fit/src/FitAddon.ts
+++ b/addons/xterm-addon-fit/src/FitAddon.ts
@@ -23,8 +23,11 @@ const MINIMUM_ROWS = 1;
 
 export class FitAddon implements ITerminalAddon {
   private _terminal: Terminal | undefined;
+  private _measureParent: boolean;
 
-  constructor() {}
+  constructor(measureParent: boolean = true) {
+    this._measureParent =  measureParent;
+  }
 
   public activate(terminal: Terminal): void {
     this._terminal = terminal;
@@ -68,9 +71,12 @@ export class FitAddon implements ITerminalAddon {
     const scrollbarWidth = this._terminal.options.scrollback === 0 ?
       0 : core.viewport.scrollBarWidth;
 
-    const parentElementStyle = window.getComputedStyle(this._terminal.element.parentElement);
-    const parentElementHeight = parseInt(parentElementStyle.getPropertyValue('height'));
-    const parentElementWidth = Math.max(0, parseInt(parentElementStyle.getPropertyValue('width')));
+    const measureElement = this._measureParent
+      ? this._terminal.element.parentElement
+      : this._terminal.element;
+    const measureElementStyle = window.getComputedStyle(measureElement);
+    const measureElementHeight = parseInt(measureElementStyle.getPropertyValue('height'));
+    const measureElementWidth = Math.max(0, parseInt(measureElementStyle.getPropertyValue('width')));
     const elementStyle = window.getComputedStyle(this._terminal.element);
     const elementPadding = {
       top: parseInt(elementStyle.getPropertyValue('padding-top')),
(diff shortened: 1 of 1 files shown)


jerch:
@PerBothner I checked with all renderers - the outer element is not populated by anything else beside a single div as the main terminal element, where happens all the stuff. I have not checked event listeners and if there is any attached to the parent element, but if not - then the parent element is indeed superfluous. Idk why things are nested that way on DOM level, could be just for historical/evolutionary reasons (esp. the development of the input and scroll helpers has been bumpy and shifted alot in the earlier days).

On the other hand - is another div container around really an issue? Wouldn't the layout engine quickly fold things together, if it finds only one child?

Well it might be worth to think about removing the superfluous element, if it helps on the perf front (I really doubt that), but this also removes the additional level of encapsulation. Not sure if and how integrations make use of that.

PerBothner:
It's partly a matter of elegance - I dislike superfluous clutter in the DOM.

However, where the extra element most annoys me is when debugging: when  I use the Web Inspector that extra level of nesting requires an extra click on a unfold/unhide button and an extra level of indentation in the tree view (which is unfortunate when width is limited).

PerBothner:
> The whole point of the fit addon is to fit xterm.js within some parent element.

That is not what `fit` does, based on my reading of the code.  The fit addon calculates size in rows*columns, based on available size in pixels. It does not change the size of the xterm.element.

> It's not a good solution to check the size of term.element as that element is meant to be "owned" by xterm.js once you pass it in. ie. it's the element that is sized by xterm.js itself, the fit addon sets that element's size indirectly based on the container dimensions.

Actually, I think it is simpler and cleaner if xterm.js is responsible for the contents of the element (including padding and possible background), while the "embedding context" is responsible for the position and size of the xterm element. (Fit does not check for border sizing; perhaps it should.) No need for an extra parent element - that just adds an extra place where something can go wrong or where you might need to fiddle with css. 

> Additionally, in your approach there's a circular dependency where the fit addon reads term.elements dimensions in order to set the term.elements dimensions. When fit is called again, it would be called on the dimensions used by that it set itself, not the container which is was designed to do.

The actual fit addon measures the dimensions (in pixels) of the element (my recommendation) or its parent (old implementation), adjusts for padding and scrollbarWidth, and figures out how many rows*columns will fit, given the available dimensions. The fit addon does not set the term.element dimensions.  Maybe something else does?  It doesn't seem to be: With my change, fit seems to work just fine: I can resize the main window; adjust the slider between  the xterm.js window and other windows; and change the zoom level.  If there were such a circularity as you mention, I don't think it would work.

Tyriar:
> The fit addon calculates size in rows*columns, based on available size in pixels. It does not change the size of the xterm.element.

It's size is defined by it's descendants, anything under `term.element` is internal so with your proposal if you need the size of the terminal after calling `term.resize`, you wouldn't be able to measure element and instead have to access internal things that could change like `xterm-viewport` or `xterm-screen`.

> The fit addon does not set the term.element dimensions. Maybe something else does?

The circular logic I was calling out is:

- Call fit()
  - Fit addon queries term.element's dimensions
  - Fit addon calls term.resize
  - xterm.js sets term.element's dimensions (indirectly via its children)
- Call fit() again
  - Fit addon queries term.element's dimensions which were originally set by the fit addon via term.resize

You may be setting the width explicitly on your end before the second fit call, which would work, but that's the problem here as then multiple things are setting its dimensions.

I think this is a case where you need to just accept the extra element or roll your own resize logic like VS Code does or fork the fit addon if it bothers you that much, you will need to touch private API though until https://github.com/xtermjs/xterm.js/issues/702 is done. Changing the default is out of the question as it's a disruptive breaking change, and adding an API makes the API more complicated for very little reason.

PR #4426, opened 2023-03-01
on tab or escape, allow leaving the buffer

meganrogge:
part of the fix for
https://github.com/microsoft/vscode/issues/175764
https://github.com/microsoft/vscode/issues/175765

### DIFF
--- a/src/browser/AccessibleBuffer.ts
+++ b/src/browser/AccessibleBuffer.ts
@@ -38,6 +38,7 @@ export class AccessibleBuffer extends Disposable {
     this.register(addDisposableDomListener(this._element, 'keydown', (ev: KeyboardEvent) => {
       if (ev.key === 'Tab' || ev.key === 'Escape') {
         this._isAccessibleBufferActive = false;
+        this._element.classList.remove('active');
         if (ev.key === 'Escape') {
           this._terminal.focus();
         }
@@ -83,6 +84,7 @@ export class AccessibleBuffer extends Disposable {
     } else {
       this._element.replaceChildren(bufferElements);
     }
+    this._element.classList.add('active');
   }
 
   private _handleColorChange(colorSet: ReadonlyColorSet): void {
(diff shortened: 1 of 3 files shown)


src/browser/public/Terminal.ts
```diff
@@ -242,7 +242,7 @@ export class Terminal implements ITerminalApi {
   public static get strings(): ILocalizableStrings {
     return Strings;
   }
-
+  // to do
```
meganrogge, line 245:
```suggestion
```

src/browser/public/Terminal.ts
```diff
@@ -242,7 +242,7 @@ export class Terminal implements ITerminalApi {
   public static get strings(): ILocalizableStrings {
     return Strings;
   }
-
+
```
meganrogge, line 245:
```suggestion

```

src/browser/AccessibleBuffer.ts
```diff
@@ -36,17 +36,16 @@ export class AccessibleBuffer extends Disposable {
...
     this.register(addDisposableDomListener(this._element, 'keydown', (ev: KeyboardEvent) => {
-      if (ev.key === 'Tab') {
+      if (ev.key === 'Tab' || ev.key === 'Escape') {
         this._isAccessibleBufferActive = false;
+        this._element.classList.remove('active');
+        if (ev.key === 'Escape') {
+          this._terminal.focus();
+        }
       }
     }
```
Tyriar, line 46: Can this instead of handled by the browser naturally for tab (listen to blur/focusout to react) and vscode keybindings for escape?

Tyriar (changes requested):

src/browser/AccessibleBuffer.ts
meganrogge: Sure re escape in VS Code, though we might want that hear as well.

As for handling tab naturally, I don't think we want to do that because then when you focus out IE click in the editor, the xterm's text-area will get focused - making the accessible buffer hide. That's specifically what this fixes.

meganrogge:
we've decided that `AccessibleBuffer` is best kept out of xterm.js

PR #4427, opened 2023-03-02
remove `AccessibleBuffer`

meganrogge:
It only complicated things to have this in xterm.js. For example, see

https://github.com/xtermjs/xterm.js/pull/4426#discussion_r1123437456


src/browser/Terminal.ts
```diff
@@ -778,9 +767,7 @@ export class Terminal extends CoreTerminal implements ITerminal {
      */
     this.register(addDisposableDomListener(el, 'mousedown', (ev: MouseEvent) => {
       ev.preventDefault();
-      if (this._accessibleBuffer?.isAccessibleBufferActive) {
-        return;
-      }
+
```
meganrogge:
```suggestion
```

Tyriar (approved):

PR #4438, opened 2023-03-15
fix BCE render bug in canvas renderer

### LINKED ISSUE #4120: Clearing formatting results in different background color for different renderers
VS Code issue: https://github.com/microsoft/vscode/issues/160497

image

The rendering of the repro in the linked issue is different in webgl+canvas vs dom

jerch:
Fixes #4120.


Tyriar (approved):
Thanks for looking into it 🙂

PR #4442, opened 2023-03-22
WIP: fork attach plugin to parse messages

hkolvenbach:
Needed to fork the attach plugin to be able to parse the message before forwarding it to the xterm terminal. This allows us to send control messages (e.g. resize) over the same WebSocket channel

Work in progress, needs some cleanup


jerch:
@hkolvenbach
Just a minor headsup - the mangling through JSON + base64 will kill IO throughput.
If IO speed is important to you, you'd get better results with a binary transport protocol. Most simple one - place a leading byte as message type identifier in front of the data.

hkolvenbach:
@jerch thanks for the heads-up, this would indeed make it a lot more performant. As you can imagine, this PR was opened by accident, but thanks for the feedback :-)

Tyriar:
@hkolvenbach GH doesn't allow deleting PRs, only close

Issue #4441, opened 2023-03-21
Unfocused line/underline cursor should show the hollow rectangle


Tyriar:
Currently it's easy to make the line and underline cursor styles display the same when focused as when not focused. We should just make it consistently show the rectangle when not focused.

Tyriar:
We're going back on the decision made in https://github.com/microsoft/vscode/issues/67410 and making unfocused show the rectangle regardless of cursor style.

PR #4443, opened 2023-03-22
Uniform cursor unfocused rendering. #4441

tisilent:
#4441

### DIFF
--- a/addons/xterm-addon-canvas/src/CursorRenderLayer.ts
+++ b/addons/xterm-addon-canvas/src/CursorRenderLayer.ts
@@ -149,11 +149,7 @@ export class CursorRenderLayer extends BaseRenderLayer {
       this._ctx.save();
       this._ctx.fillStyle = this._themeService.colors.cursor.css;
       const cursorStyle = this._optionsService.rawOptions.cursorStyle;
-      if (cursorStyle && cursorStyle !== 'block') {
-        this._cursorRenderers[cursorStyle](cursorX, viewportRelativeCursorY, this._cell);
-      } else {
-        this._renderBlurCursor(cursorX, viewportRelativeCursorY, this._cell);
-      }
+      this._renderBlurCursor(cursorX, viewportRelativeCursorY, this._cell);
       this._ctx.restore();
       this._state.x = cursorX;
       this._state.y = viewportRelativeCursorY;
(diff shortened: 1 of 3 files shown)


Tyriar (commented):
I remembered when looking at this that we actually did the opposite a while ago, tracking it down  I see it was done 3.5 years ago:

- https://github.com/xtermjs/xterm.js/pull/2455
- https://github.com/xtermjs/xterm.js/issues/2356
- https://github.com/microsoft/vscode/issues/67410

This is something that the expected behavior isn't well defined. Personally I think it's better with the rectangle, regardless of cursor style so it's obvious that it's not focused. @jerch what do you think? Should we merge this and close any future complaints as by design?

tisilent:
I was just thinking about whether to comment or delete the code. :flushed:

jerch:
Hmm hard to tell, whats the right direction here. Imho there are two aspects that a hollow cursor will show - the current cursor pos + the fact that the terminal in not focused atm.

I cannot think of a good pendant for underline/bar cursors here - still showing the underline/bar loses the unfocused state information for the user. And while not showing any cursor would transport that info, it now lost the position info for the user. So imho the hollow rectangle is the best compromise, since it can show the user both aspects.

I also think that ppl would not mind much, if a terminal they are not actively interacting with would flip the cursor into a hollow block? (Ofc the cursor should not magically appear, if requested to be hidden before by a sequence.)

To sum things up - I think the hollow rectangle is the right way to go for unfocused state.

Tyriar (approved):

boettges:
> This is something that the expected behavior isn't well defined. Personally I think it's better with the rectangle, regardless of cursor style so it's obvious that it's not focused. 

@tyriar also regarding: https://github.com/microsoft/vscode/issues/178379#issuecomment-1487169731 

Since applying custom CSS styles is not possible for terminal's content, I think there should be a choice.

Instead of forcing "block" for all inactive terminal instances, I'd suggest a VS Code user setting (for example `terminal.integrated.inactiveCursorStyle`) with the following options:

- `outline` (default)
- `line` (non-blinking)
- `underline` (non-blinking)
- `none` (cursor hidden)

Personally, I've gotten used to the blinking line in regular input modes, while I use both, the block and outline cursor, exclusively in vim's normal and visual mode, respectively.

Tyriar:
@boettges sounds good, created https://github.com/xtermjs/xterm.js/issues/4566

Issue #4444, opened 2023-03-22
Improve search addon behavior when there are > 1000 results


Tyriar:
See https://github.com/microsoft/vscode/issues/178017 for context, `_searchResults` gets set to undefined when there are > 1000 results:

https://github.com/xtermjs/xterm.js/blob/d6599bc0486a5de61eeb86638cf5313f1a91baff/addons/xterm-addon-search/src/SearchAddon.ts#L168-L172

ulgyd0gg:
About the result "0 of -1"

https://github.com/xtermjs/xterm.js/blob/d6599bc0486a5de61eeb86638cf5313f1a91baff/addons/xterm-addon-search/src/SearchAddon.ts#L90-L101

`this._resultIndex` be set to `-1`, when `_searchResults`  already been set to `undefine` because of `_serachResults.size` > 1000 results.

jeanp413:
Created a PR for it https://github.com/xtermjs/xterm.js/pull/4504

PR #4452, opened 2023-03-27
Fix for #4444. remove serach result size limit.

ulgyd0gg:


ulgyd0gg:
This is the first time I tried to contribute code. I don't think I can fix this problem anytime soon. But I'll still try to fix it and trace this problem.

Tyriar:
@ulgyd0gg no problems, I'll close this off as the fix should look pretty differently but feel free to create a new one

PR #4458, opened 2023-03-28
Clear line wrapped state on explicit line feed

### LINKED ISSUE #4457: Links can be computed on the wrong line data on Windows
VS Code issue: https://github.com/microsoft/vscode/issues/178014

Repro:

Run this in demo:

```
echo "https://github.com/microsoft/vscode/issues/178014/github.com/microsoft/vscode/issues/178014
abc https://github.com/microsoft/vscode/issues/178014/github.com/microsoft/vscode/issues/178014 def
[10:45:12] https://github.com/microsoft/vscode/issues/178014/github.com/microsoft/vscode/issues/178014: alksmdlkasd"
```

Resize wider/narrower, this is at 32 cols:

image

The line data is correct here but the underline wrong:

image

Another repro I did saw the data very messed up:

image

Tyriar:
Fixes #4457
See microsoft/vscode#178014

This fixes the long standing issue of strange underlines on Windows after resizing. I believe it's safe to do this as the linefeed is explicit so will merge now. @jerch your opinions would be helpful here as we can make a new option or extend windowsMode if you don't think it's safe.

### DIFF
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -711,6 +711,13 @@ export class InputHandler extends Disposable implements IInputHandler {
       this._bufferService.scroll(this._eraseAttrData());
     } else if (this._activeBuffer.y >= this._bufferService.rows) {
       this._activeBuffer.y = this._bufferService.rows - 1;
+    } else {
+      // There was an explicit line feed (not just a carriage return), so clear the wrapped state of
+      // the line. This is particularly important on conpty/Windows where revisiting lines to
+      // reprint is common, especially on resize. Note that the windowsMode wrapped line heuristics
+      // can mess with this so windowsMode should be disabled, which is recommended on Windows build
+      // 21376 and above.
+      this._activeBuffer.lines.get(this._activeBuffer.ybase + this._activeBuffer.y)!.isWrapped = false;
     }
     // If the end of the line is hit, prevent this action from wrapping around to the next line.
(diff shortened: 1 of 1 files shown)


src/common/InputHandler.ts
```diff
@@ -711,6 +711,13 @@ export class InputHandler extends Disposable implements IInputHandler {
       this._bufferService.scroll(this._eraseAttrData());
     } else if (this._activeBuffer.y >= this._bufferService.rows) {
       this._activeBuffer.y = this._bufferService.rows - 1;
+    } else {
+      // There was an explicit line feed (not just a carriage return), so clear the wrapped state of
+      // the line. This is particularly important on conpty/Windows where revisiting lines to
+      // reprint is common, especially on resize. Note that the windowsMode wrapped line heuristics
+      // can mess with this so windowsMode should be disabled, which is recommended on Windows build
+      // 21376 and above.
+      this._activeBuffer.lines.get(this._activeBuffer.ybase + this._activeBuffer.y)!.isWrapped = false;
```
jerch, line 720: Imho it makes sense to even apply this for all NLs, so a line after an NL has always the wrapped state cleared. At least I cannot think of any exception to this? (And really wonder why we did not have it in code yet, or got test failures, haha...)

Tyriar: Cool, well this else case should cover everything as the 2 cases above are if it's the last row and then a sanity check to clamp the value. Nice to see this bug finally fixed!

PR #4472, opened 2023-04-10
Fix for #4444.

ulgyd0gg:
When the `_searchResults.size` is bigger than 1000 will skip the loop and decorations the `_searchResults` and also skip both the `_findPreviousAndSelect`  and `_findNextAndSelect` update the `_resultIndex`. So that when size bigger than limit `this._onDidChangeResults.fire(undefined)` will be call.


ulgyd0gg:
@Tyriar I see this problem had been done. plz help to close this PR.

Tyriar:
@ulgyd0gg ah sorry about not getting to this, I took a break for a few weeks and am still catching up 😅

ulgyd0gg:
@Tyriar Never mind 👻

Issue #4474, opened 2023-04-11
Font size is not updated when changed in browser DOM rendering


kelvinhammond:
The font size is not updated when changed in browser DOM rendering.

## Details
- Browser and browser version: Any
- OS version: Any
- xterm.js version: 5.1.0

### Steps to reproduce

1. Start the demo and open it in a browser
2. Disable the webgl addon
3. Change the fontSize to a large value like 23
4. Notice the fontSize didn't actually change but the spacing is incorrect
5. View the style and notice it is still the previous value and not updated.

This can be fixed by re-rendering or updating the style textContent.

PR #4475, opened 2023-04-11
Re-render browser DOM style onOptionsChanged, fixes xtermjs#4474

kelvinhammond:
Fixes https://github.com/xtermjs/xterm.js/issues/4474


Tyriar:
Thanks @kelvinhammond, looks good, I just moved the inject call into _handleOptionsChanged.

Tyriar (approved):

PR #4479, opened 2023-04-16
wasm serialize & hook interface

jerch:
Early WIP...


jerch:
@Tyriar The build passed in CI :smiley:

But the runtime penalties are quite big for mac and windows (>3 min), under linux it is only +20s :thinking:

To not penalize things from the compiler sdk, I could change gitignore to include the built wasm file, thus the CI does not have to spin up a compiler at all. Downside - the wasm file would have to reside in the repo.

jerch:
Running xterm-benchmark gives these numbers:
```bash
   Context "out-benchmark/benchmark/SerializeAddon.benchmark.js"
      Context "Terminal: sh -c "dd if=/dev/urandom count=40 bs=1k | hexdump | lolcat -f""
         Context "serialize"
            Case "#1" : 5 runs - average throughput: 5.91 MB/s
```
```bash
   Context "out-benchmark/benchmark/Serialize2Addon.benchmark.js"
      Context "Terminal: sh -c "dd if=/dev/urandom count=40 bs=1k | hexdump | lolcat -f""
         Context "serialize"
            Case "#1" : 5 runs - average throughput: 64.30 MB/s
```

jerch:
@Tyriar Found a way to avoid costly SDK pulling during normal CI runs:
- wasm target gets committed into repo by default
- with `inwasm -S` recompilation can be force-skipped (compilation would pull SDKs), instead it just bundles the repo version
- later before packaging `inwasm -f` can be used to force a recompilation before doing final tests and packaging

Time to get back to the actual serialization code :smile_cat:

Tyriar:
> later before packaging inwasm -f can be used to force a recompilation before doing final tests and packaging

@jerch committing it and forcing compile during packaging sounds like a fine compromise to me. `package` and `prepackage` is always run before releasing. You could add the compile as `postpackage` as that should always run before we release, then just try remember to always commit it during PR changes so it stays up to date.

addons/xterm-addon-serialize2/src-wasm/serialize.c
Tyriar, line 1: Can we add a "serialize2 (wasm)" button to the demo for convenient testing?

addons/xterm-addon-serialize2/src/Serialize2Addon.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    terminal = new Terminal({ cols: 10, rows: 2, allowProposedApi: true });
+    terminal.loadAddon(serializeAddon);
+
+    (terminal as any)._core._themeService = new ThemeService(new OptionsService({}));
+    (terminal as any)._core._selectionService = new TestSelectionService((terminal as any)._core._bufferService);
+  });
+
+  describe('text', () => {
+    // TODO: wirte alot more test cases here...
```
Tyriar, line 82: Is it possible to just copy over the old SerializeAddon.test.ts?

addons/xterm-addon-serialize2/src/Serialize2Addon.ts
```diff
@@ -0,0 +1,39 @@
...
+    this._terminal = terminal;
+  }
+
+  public serialize(options?: ISerializeOptions): string {
+    // TODO: Add combinedData support
+    if (!this._terminal) {
+      throw new Error('Cannot use addon until it has been loaded');
+    }
+
+    return serialize((this._terminal as any)._core);
```
Tyriar, line 34: Haven't checked what you need here, but casting it to an internal interface like this will catch typing problems at compile time:

https://github.com/xtermjs/xterm.js/blob/b7d73f63cd730bd8aeb48121ee29d9f8f1c7dc2e/addons/xterm-addon-webgl/src/WebglAddon.ts#L38

addons/xterm-addon-serialize2/src/serializer.wasm.ts
```diff
@@ -0,0 +1,214 @@
...
+  const buffer = t.buffer;
+  const len = t.buffer.lines.length;
+  let wPos = 150*65536;
+
+  let line: IBufferLineInternal;
+
+  // single call is pretty wasteful? preload combines once instead?
+  writeCombinedOverload = (dst, x) => {
+    let dp = dst >> 1;
+    const comb: string = (line as any)._combined[x] || '';
```
Tyriar, line 144: Any `as any` should use internal types where possible, if we need to expose a `combined` on the interface that's fine

addons/xterm-addon-serialize2/src/serializer.wasm.ts
```diff
@@ -0,0 +1,214 @@
...
+    LUT100[i1 * 10 + i2] = (48 + i2) << 16 | (48 + i1);
+  }
+}
+
+const mem = new WebAssembly.Memory({ initial: 3000 });
+const d16 = new Uint16Array(mem.buffer);
+const d32 = new Uint32Array(mem.buffer);
+let writeCombinedOverload: (dst: number, x: number) => number = (dst, x) => dst;
+const writeCombined: (dst: number, x: number) => number = (dst, x) => writeCombinedOverload(dst, x);
+let writeLinkOverload: (dst: number, x: number) => number = (dst, x) => dst;
```
Tyriar, line 124: Instead of setting these functions via `let`, should they just be functions that take `line` and whatever else is needed?

addons/xterm-addon-serialize2/src/serializer.wasm.ts
```diff
@@ -0,0 +1,214 @@
...
+const td = new TextDecoder('UTF-16LE');
+
+
+export function serialize(t: ITerminal): string {
+  // reset FG/BG/ext
+  inst.exports.reset(0, 0, 0, 0);
+
+  const buffer = t.buffer;
+  const len = t.buffer.lines.length;
+  let wPos = 150*65536;
```
Tyriar, line 137: Surprised this isn't a lint rule:

```suggestion
  let wPos = 150 * 65536;
```

addons/xterm-addon-serialize2/README.md
```diff
@@ -0,0 +1,42 @@
+## xterm-addon-serialize
+
+An addon for [xterm.js](https://github.com/xtermjs/xterm.js) that enables xterm.js to serialize a terminal framebuffer into string or html. This addon requires xterm.js v4+.
```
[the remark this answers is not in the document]
jerch: Sure thing. I just copy-cloned from the old addon and did not bother to fix all things yet.

addons/xterm-addon-serialize2/src-wasm/serialize.c
jerch: Yepp, already did that in my local branch (not pushed yet...)

addons/xterm-addon-serialize2/src/Serialize2Addon.test.ts
jerch: I did that, and removed all HTML test code (since the addon does not implement any HTML serialization yet). Then I realized, that there are no real test cases for the text-based serializer beside a dummy one... :scream:

addons/xterm-addon-serialize2/src/serializer.wasm.ts
```diff
@@ -0,0 +1,214 @@
...
+
+// itoa LUT
+// note: performant decimal conversion of numbers is actually a hard problem
+// we use a LUT approach with 0-99 precomputed in utf16 for better performance
+const LUT100 = new Uint32Array(100);
+for (let i1 = 0; i1 < 10; ++i1) {
+  for (let i2 = 0; i2 < 10; ++i2) {
+    LUT100[i1 * 10 + i2] = (48 + i2) << 16 | (48 + i1);
+  }
+}
```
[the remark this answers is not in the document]
jerch: Yes, thats part of the bleeding eye code - see my remarks above.

addons/xterm-addon-serialize2/src/serializer.wasm.ts
jerch: Same as above :smiley_cat:

addons/xterm-addon-serialize2/src/serializer.wasm.ts
jerch: Same as above :smiley_cat: - I already reworked that one locally, it is now a static function pointer on a context object and on top level anymore.

addons/xterm-addon-serialize2/src/serializer.wasm.ts
jerch: Hmm, interesting. Yeah I fixed all linter nagging, but this one did not show up.

The hardcoded values will be replaced by proper memory management, prolly static again to avoid wasm compilation over and over, instead calling earlier into `TextDecoder.decode`.

addons/xterm-addon-serialize2/src/Serialize2Addon.ts
Tyriar: I didn't look too hard, just skimmed it 🙂

jerch: ... did *not* read ..., tss tss tss :rofl:

jerch:
reminder - also add overline attribute support.

Tyriar:
This is a bit stale so let's close as it hasn't been touched in years. It's a cool idea but I'm not sure serialize needs the speed boost tbh. If we want to embrace wasm more we should figure out how to exactly, like choosing a language (zig is probably my preference), setting up some tooling and fixing the build time problem. One of my worries here is the maintenance burden but LLMs help bridge that gap quite a bit.

jerch:
@Tyriar This was basically blocked by the inwasm issues, which got mostly resolved now.

But it is still better to keep wasm source separated from TS umbrella in this repo, as `inwasm` would create quite some burden on the frequently touched code here, so outsourcing wasm stuff (separate subfolder or other package) seems easier to deal with.

But I wont have time to finish this rewrite anytime soon either, as there are other pressing things to get done first (qoi support for image addon, then a sequence for serializing images).

About the tooling with wasm sources - `inwasm` is still the most convenient tool I have seen for small wasm additions. The tooling around emscripten, rust/wasm-bindgen or AssemblyScript are weird for one or another reason. I dont know any explicit tooling for zig, also I don't know the language itself. For me C/C++ with clang still gives the best results.

PR #4482, opened 2023-04-19
Add smoothScroll to scrollLines

### LINKED ISSUE #4121: Smooth scrolling should apply when using Terminal.scroll* APIs
VS Code issue: https://github.com/microsoft/vscode/issues/158923

tisilent:
Fix #4121


src/browser/Terminal.ts
```diff
@@ -869,11 +876,76 @@ export class Terminal extends CoreTerminal implements ITerminal {
...
+          this._smoothScrollProgressState.origin = 0;
+          this._smoothScrollProgressState.target = disp;
+          this._smoothScrollProgressState.progress = 0;
+          this._smoothScroll(suppressScrollEvent, source);
+        } else {
+          this._clearSmoothScrollState();
+        }
+      }
+    }
+  }
```
Tyriar, line 902: This general approach seems good, but we want this logic to live in Viewport. The reasons are to keep that code in the same place, out of the main Terminal class as well as sharing the smooth scroll state object.

src/browser/Terminal.ts
```diff
@@ -869,11 +876,76 @@ export class Terminal extends CoreTerminal implements ITerminal {
     }
   }
 
-  public scrollLines(disp: number, suppressScrollEvent?: boolean, source = ScrollSource.TERMINAL): void {
+  private _scrollLines(disp: number, suppressScrollEvent?: boolean, source = ScrollSource.TERMINAL): void {
     super.scrollLines(disp, suppressScrollEvent, source);
     this.refresh(0, this.rows - 1);
   }
 
+  public scrollLines(disp: number, suppressScrollEvent?: boolean, source = ScrollSource.TERMINAL): void {
```
Tyriar, line 884: We'll want to also hook up `scrollPages`, `scrollToTop`, `scrollToBottom` and `scrollToLine` which should be fairly easy once it's pulled into `Viewport`.

Tyriar (changes requested):

src/browser/Terminal.ts
tisilent: scrollToTop  scrollToBottom , 
Using smoothScroll may cause some time waiting, I'll give it a try first.

Tyriar: It's duration based, so all scrolling no matter how large or small should take the same amount of time

tisilent: add scrollline to viewport.It should be relatively easy to use.

src/browser/Viewport.ts
```diff
@@ -175,33 +181,46 @@ export class Viewport extends Disposable implements IViewport {
...
+    this._smoothScroll(({ position }) => {
+      this._viewportElement.scrollTop = this._smoothScrollState.origin + position;
+    });
+  }
+
+  private _linesSmoothScroll(): void {
+    this._smoothScroll(({ position, lastPosition }) => {
+      this._scrollLines(position - lastPosition, false);
+    });
+  }
```
Tyriar, line 222: I was expecting these to just use the logic that was already there, so the main thing being setting the `_smoothScrollState` if needed when calling into `scrollLines`?

Tyriar (changes requested):

src/browser/Viewport.ts
tisilent: I seem to have thought of something.try try.

src/browser/Terminal.ts
tisilent: scrollToTop example:
add to Viewport
`
public scrollToTop(): void {
  this.scrollLines(-this._bufferService.buffer.ydisp);
}
`

Tyriar (approved):
Thanks @tisilent, I made some changes to fix the tests and pass the scroll to top/bottom/line/pages calls through `scrollLines`

tisilent:
Oho @Tyriar , learning to write tests, I'm thinking about some of the logic of Terminal scrollLine moving to Viewport.
Terminal -> Viewport -> BufferService

Tyriar:
Test coverage is good enough currently, there are unit tests which run in node and use a Mock Viewport and then API tests that actually use the scroll apis in a browser context

tisilent:
I’am very sorry I had a wrong to sent the message. @Tyriar
<del>learning to write tests</del>
I'm learning to write tests .
:sweat:

PR #4491, opened 2023-04-24
fix urls with screenreader

jerch:
@Tyriar This small change fixes things for the url event listeners. Idk if it will break some functionality on screenreader side, thus it furthermore needs proper testing with a screenreader app being attached.

### DIFF
--- a/css/xterm.css
+++ b/css/xterm.css
@@ -149,6 +149,7 @@
     right: 0;
     z-index: 10;
     color: transparent;
+    pointer-events: none;
 }
 
 .xterm .live-region {


meganrogge:
Before the change vs after the change 👍🏼 

Using VoiceOver with this setting, I was hoping that the link would get announced, but it doesn't. It did not before the change either.

Screenshot 2023-05-03 at 8 15 30 AM

meganrogge (approved):

jerch:
> ... I was hoping that the link would get announced, but it doesn't. ...

I think this highly depends on what underlying data structure the screenreader will trigger its link announcement. While we visually mimick a DOM link here, the underlying data is still nowhere close to a DOM link firing its hover event. But it might be possible to fake that as well in the aria container. Maybe just attach a link provider listener, that places a DOM link and triggers a hover in the aria element will already do?

Tyriar:
@meganrogge it occurs to me now we probably shouldn't have ripped out the a11y tree considering we moved the accessible buffer feature into vscode's codebase https://github.com/xtermjs/xterm.js/issues/4503

RahjIII:
Thanks for the fix.  Verified in xterm 5.2.1.

PR #4500, opened 2023-05-02
Clear whole cursor layer on Firefox

### LINKED ISSUE #4487: L-shaped artifacts when using xterm instances with Canvas or WebGL addons on Firefox
Hey there! Apologies if this has been reported already. I tried searching for similar issues and came across other graphics bugs but couldn't find one that reported this exact issue.

One graphical bug we've been noticing when using xterm has been these strange L-shaped artifacts that are occasionally persisted in the terminal and overlay the output after some time. They seem to be left over from the cursor as they match it in shape/size and appear to only render in places where the cursor was previously rendered. Note that this only happens when either the Canvas or WebGL addons are loaded. Rendering to the DOM, as xterm does by default, does not seem to cause this. We've also only been able to reproduce this on Firefox so it may be a bug in Firefox itself and how it renders graphics to a canvas.

This is a screenshot from our config which was using the Canvas addon (as well as a number of other addons):

xterm-canvas-artifacts

We were also able to reproduce this in the terminal instance on the xterm [landing page](https://xtermjs.org/):

xterm-canvas-artifacts-official-website

## Details

- Browser and browser version: Firefox 112.0.1
- OS version: MacOS Ventura 13.3.1
- xterm.js version: 5.1.0 (`xterm-addon-canvas` version 0.3.0 and `xterm-addon-webgl` version 0.14.0)

### Steps to reproduce

1. Use Firefox
2. Visit an xterm instance with either the Canvas or WebGL addons loaded
3. Enter a bunch of commands that trigger output (e.g. `ls`, `pwd`, etc)
4. After a few, you should start noticing L-shaped artifacts similar to the screenshots above, especially while the cursor moves

Tyriar:
Fixes #4487

I could not reproduce this but it should fix the problem.

### DIFF
--- a/addons/xterm-addon-canvas/src/CursorRenderLayer.ts
+++ b/addons/xterm-addon-canvas/src/CursorRenderLayer.ts
@@ -13,6 +13,7 @@ import { IEventEmitter } from 'common/EventEmitter';
 import { ICoreBrowserService, IThemeService } from 'browser/services/Services';
 import { Terminal } from 'xterm';
 import { toDisposable } from 'common/Lifecycle';
+import { isFirefox } from 'common/Platform';
 
 interface ICursorState {
   x: number;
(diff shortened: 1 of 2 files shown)


MatthiasGrandl:
I am on 5.2.1 and still encounter this issue.

image

Tyriar:
@MatthiasGrandl this is surprising to hear, that's on Firefox still?

MatthiasGrandl:
@Tyriar yeah Firefox. Both MacOS and Linux. Maybe it’s font specific? This is JetBrains Mono, I’ll try another one and report back.

Tyriar:
What's your zoom level in the browser?

MatthiasGrandl:
Default so 100%, I am on a HiDPI display with scaling though. Browser `devicePixelRatio` is 2.

It does occure much less frequently with the update to 5.2. I also reproduced it with the default font, so does not appear to be font specific.

PR #4504, opened 2023-05-07
Fixes improve search addon behavior when there are > 1000 results

jeanp413:
Fixes https://github.com/xtermjs/xterm.js/issues/4444


addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -232,12 +232,12 @@ export class SearchAddon extends Disposable implements ITerminalAddon {
...
     if (this._terminal.hasSelection()) {
-      const incremental = searchOptions ? searchOptions.incremental : false;
+      // const incremental = searchOptions ? searchOptions.incremental : false;
       // Start from the selection end if there is a selection
       // For incremental search, use existing row
       currentSelection = this._terminal.getSelectionPosition()!;
-      startRow = incremental ? currentSelection.start.y : currentSelection.end.y;
-      startCol = incremental ? currentSelection.start.x : currentSelection.end.x;
+      startRow = currentSelection.end.y;
+      startCol = currentSelection.end.x;
```
jeanp413, line 240: Not sure about the purpose of `incremental` :thinking: :
- with it `should fire with correct event values (incremental)` and `should fire with correct event values (incremental)` tests fail but due to a bug with the original code as the `resultIndex` was incorrectly increased when the selection was no changing at all
- If I remove it, all test pass successfully

meganrogge: I don't remember what that was meant to do , but your changes work very well

Tyriar:
Yes it does seem to be broken, incremental should not jump to the result at the bottom

jeanp413:
I fixed the incremental code but the flag is redundant
- it will always be incremental if there's a selection in the terminal (which I like as it behaves as the vscode editor), with this I only needed to fix 1 test
- if I make the flag required, around 6 test will fail

addons/xterm-addon-search/src/SearchAddon.ts
```diff
@@ -153,32 +151,31 @@ export class SearchAddon extends Disposable implements ITerminalAddon {
...
-    this._searchResults.forEach(result => {
+    for (const result of searchResultsWithHighlight) {
       const resultDecoration = this._createResultDecoration(result, searchOptions.decorations!);
       if (resultDecoration) {
-        const decorationsForLine = resultDecorations.get(resultDecoration.marker.line) || [];
-        decorationsForLine.push(resultDecoration);
-        resultDecorations.set(resultDecoration.marker.line, decorationsForLine);
+        this._highlightedLines.add(resultDecoration.marker.line);
+        (resultDecoration as unknown as { match: ISearchResult }).match = result;
+        this._highlightDecorations.push(resultDecoration as (IDecoration & { match: ISearchResult }));
```
Tyriar, line 176: It would be better to store these as a new object to avoid the casting/intersection type:

```ts
{ decoration: IDecoration, match: ISearchResult }
```

addons/xterm-addon-search/typings/xterm-addon-search.d.ts
```diff
@@ -79,6 +79,11 @@ declare module 'xterm-addon-search' {
    * An xterm.js addon that provides search functionality.
    */
   export class SearchAddon implements ITerminalAddon {
+    /**
+     * Max number of matches when decorations are enabled
+     */
+    public readonly MATCHES_LIMIT: number;
```
Tyriar, line 85: Let's pass this in through the ctor to let embedders define it?

Tyriar (changes requested):
The behavior seems mostly correct now, I do notice the following issues though:

- Pressing shift, ctrl or alt will do a non-incremental search
- Doing an incremental search will not apply the white activeMatchBorder defined here https://github.com/jeanp413/xterm.js/blob/a1e9105ce78154eafce3500ace119843a72c46f7/demo/client.ts#L146

jeanp413:
> Pressing shift, ctrl or alt will do a non-incremental search

[Are you referring to this](https://github.com/xtermjs/xterm.js/blob/7a7f4f0add397075e31505e3c3fc237856604cb3/demo/client.ts#L140), as I mentioned the incremental flag does not have any effect now (it can be removed in a separate PR), if there's a selection in the terminal then search will be incremental, which key you press does not have any effect

> Doing an incremental search will not apply the white activeMatchBorder defined here

It's a bug in the decorations rendering, it does not take into account if it's on the `top` or `bottom` layer so it could happen that a decoration in `top` layer renders below, I'll create another PR fixing it

Tyriar:
I mean this, I'm pressing only shift, ctrl and alt in the gif. It's moving the active match but should do nothing:


Recording 2023-05-15 at 05 08 01

jeanp413:
I thinks that's an issue with the [dom events we are using in the demo](https://github.com/xtermjs/xterm.js/blob/7a7f4f0add397075e31505e3c3fc237856604cb3/demo/client.ts#LL306C1-L311C6), it's always using `keyup` for everything, vscode search widget uses `input` to detect the actual input changed and also `onKeyDown` to handle special cases like enter, up/down arrow, tab, etc

PR #4516, opened 2023-05-12
Ensure decorations in the top layer render on top of decorations in the bottom layer

jeanp413:


Tyriar (commented):
Isn't this already handled? Is there a case we're missing?

src/browser/decorations/BufferDecorationRenderer.ts
```diff
@@ -72,6 +72,7 @@ export class BufferDecorationRenderer extends Disposable {
   private _createElement(decoration: IInternalDecoration): HTMLElement {
     const element = document.createElement('div');
     element.classList.add('xterm-decoration');
+    element.classList.toggle('top-layer', decoration?.options?.layer === 'top');
```
Tyriar, line 75: We already have `.xterm-decoration-top`:

https://github.com/xtermjs/xterm.js/blob/cc5136aeb3ee478b7220efe4ab2f371ea7466e1d/css/xterm.css#L190-L193

I can't remember why it's z-index 2, but maybe that should be 7 instead? It also seems to only be used in the dom renderer atm

Tyriar (changes requested):

PR #4521, opened 2023-05-17
Avoid triggering a reflow while rendering search decorations

jeanp413:
It's making vcode terminal super slow while scrolling

image


jeanp413:
cc @Tyriar

Tyriar (approved):

PR #4526, opened 2023-05-20
Add support for SGR 53 overline and SGR 55 not overline

### LINKED ISSUE #4499: Support SGR 53 "overline"
WT completed feature request: https://github.com/microsoft/terminal/issues/6000
Original VS Code request: https://github.com/microsoft/vscode/issues/173011

Tyriar:
Defined in https://www.ecma-international.org/publications-and-standards/standards/ecma-48/

Fixes #4499

This implementation does not apply the underline color to the overline and it does not "dodge" the text like underline does. This is simpler and it's ambiguous whether that should happen anyway.

dom

image

canvas

image

webgl

image

### DIFF
--- a/css/xterm.css
+++ b/css/xterm.css
@@ -170,6 +170,16 @@
 .xterm-underline-4 { text-decoration: dotted underline; }
 .xterm-underline-5 { text-decoration: dashed underline; }
 
+.xterm-overline {
+    text-decoration: overline;
+}
+
+.xterm-overline.xterm-underline-1 { text-decoration: overline underline; }
+.xterm-overline.xterm-underline-2 { text-decoration: overline double underline; }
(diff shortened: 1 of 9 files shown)


jerch (approved):
Yet another attribute to look for in the serializer :scream: :rofl: 

LGTM :+1: 

> This implementation does not apply the underline color to the overline and it does not "dodge" the text like underline does. This is simpler and it's ambiguous whether that should happen anyway.

About the color:
Imho using FG color is the only useful option atm, underline color is explicitly defined for underline, but we have nothing like that for overline (plz lets not open Pandora's box here and add another color setting for a rarely used attribute).

About overline cutting through glyphs:
Actually this might help to get an easier distinction from underlines above. Also the use cases in the WT thread suggest this as a visual help for things like a status line - here a "hard border/line" cutting through glyphs might help to get the wanted visual separation.

PR #4540, opened 2023-05-31
Fix overline escaping

ChaseKnowlden:
Close https://github.com/microsoft/vscode/issues/181242


Tyriar (commented):
I don't think this would fix anything? The `&nbsp` was there to ensure the overline is continuous vs a plain space in an element`<span> </span>`

Tyriar:
Turns out that issue is due to the css file not being updated

PR #4542, opened 2023-05-31
Upgrade to Node18

silamon:
Refactored the devcontainer to use Node18 instead of Node 14 (eol). Also bumping the build tools to use Node 18.

For the devcontainer:
It's now using a prebuilt node docker image as base with a feature to enable yarn.
Also adding a post create command to install the dependencies already. 
Functionality of the devcontainer remains to work after the changes (debugging on port 3000, mocha test adapter, coding in general).


Tyriar (approved):
👏

PR #4543, opened 2023-05-31
Fix overline not working on certain fonts

### LINKED ISSUE #4541: Overline in DOM renderer doesn't show depending on the font
Current `vertical-align: top`

Screenshot 2023-05-31 at 6 55 25 am

It works with `vertical-align: middle`, maybe that's the workaround? I wouldn't be surprised if it hid the underlines in this case though

Screenshot 2023-05-31 at 6 56 26 am

ChaseKnowlden:
Fix #4541

### DIFF
--- a/css/xterm.css
+++ b/css/xterm.css
@@ -175,7 +175,6 @@
 .xterm-overline {
     text-decoration: none;
     border-top: 1px solid #fff;
-    padding-top: 0.05em;
 }
 
 .xterm-overline.xterm-underline-1 { text-decoration: overline underline; }
--- a/css/xterm.css
+++ b/css/xterm.css
@@ -173,7 +173,9 @@
 .xterm-underline-5 { text-decoration: dashed underline; }
 
 .xterm-overline {
-    text-decoration: overline;
+    text-decoration: none;
+    border-top: 1px solid #fff;
+    padding-top: 0.05em;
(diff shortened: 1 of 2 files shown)


css/xterm.css
```diff
@@ -173,7 +173,9 @@
 .xterm-underline-5 { text-decoration: dashed underline; }
 
 .xterm-overline {
-    text-decoration: overline;
+    text-decoration: none;
+    border-top: 1px solid #fff;
+    padding-top: 0.05em;
```
Tyriar, line 178: We shouldn't touch padding as that would change the size of the elements/grid. It's a bit of a shame not using the real overline decoration, but perhaps we should just use border-top. If we do, we would need to make sure that `box-sizing` = `border-box` and then undo all the `text-decoration`-related styles introduced for overline (eg. `.xterm-overline.xterm-underline-1`).

Tyriar (changes requested):

css/xterm.css
ChaseKnowlden: Removed the padding.

jerch:
Imho this PR addresses the issue from the wrong side (its a visual hack). We should instead make sure, that the DOM renderer paints the lines high enough to display all decorations in a proper way.

PR #4583, opened 2023-07-05
fix mouse tracking behavior

### LINKED ISSUE #4582: harmonize mouse tracking with other TEs
Coming from #4581.

We should try to establish the same mouse tracking behavior as seen on other TEs:
- UP & DRAG should be triggered from outside of the terminal area, if DOWN originates from within the terminal
- coords should be clamped to the outermost position still within the terminal area
- UP should always be triggered within the terminal area (fixing w3m bug)

jerch:
Should fix #4582.

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -747,8 +747,10 @@ export class Terminal extends CoreTerminal implements ITerminal {
 
       if (!(events & CoreMouseEventType.UP)) {
         this._document!.removeEventListener('mouseup', requestedEvents.mouseup!);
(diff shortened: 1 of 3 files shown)


jerch:
@Tyriar This might need some tests in vscode to not break with other mouse assumptions. The crucial changes done by this PR are:
- DRAG started within the terminal widget will keep working even outside of the terminal (you can test that with tmux split panes), before it would have stopped at the terminal border
- UP following a DOWN event from within the terminal will still fire, even if the pointer is outside of the terminal (you can test it with `node bin/test_mousemodes.js` - releasing a mouse button outside of the terminal will still update the cursor pos for protocol 1000+)

Issue #4580, opened 2023-07-01
Use of box-shadow for underline cusor can have aliasing issues under certain conditions


dlech:
The use of `box-shadow` to create the underline cursor can sometimes not render correctly. Here is a screenshot.

![image](https://github.com/xtermjs/xterm.js/assets/963645/3ba0766e-bbc4-47d4-826e-094884d4801f)

## Details
- Browser and browser version:  Chromium Version 114.0.5735.106 (Official Build) snap (64-bit)
- OS version: Ubuntu 22.04
- xterm.js version: 5.2.1

### Steps to reproduce

Unfortunately, I haven't been able to create a minimal reproducible test case. This started when we updated from blueprintjs 4.x to 5.x but I haven't been able to track down exactly what changed that triggered the issue. ~~For the time being, the bug can be seen at https://beta.pybricks.com. Just click in the terminal area to focus it.~~

### Suggested solution

From what I've read about `box-shadow`, aliasing issues like this happen sometimes. It seems like we could make a more robust cursor using `border` instead of `box-shadow`. This is the workaround I have come up with for now:

```scss

.xterm-dom-renderer-owner-1 .xterm-rows .xterm-cursor.xterm-cursor-underline {
    animation: pb-terminal-cursor 1s step-end infinite !important;
    box-shadow: none !important;
    border-top: 0;
    border-left: 0;
    border-right: 0;
    border-bottom: 1px;
    border-style: solid;
    border-color: black;

    .#{bp.$ns}-dark & {
        border-color: white;
    }
}

@keyframes pb-terminal-cursor {
    50% {
        border-style: hidden;
    }
}
```

PR #4584, opened 2023-07-05
use border-bottom instead of box-shadow for underline cursor

dlech:
In rare cases, box-shadow can cause aliasing artifacts on the top and sides of the cursor box when only the bottom should be filled in to create an underline.

This replaces the box-shadow style with border-bottom which should be pixel perfect in all cases.

Fixes: https://github.com/xtermjs/xterm.js/issues/4580

### DIFF
--- a/src/browser/renderer/dom/DomRenderer.ts
+++ b/src/browser/renderer/dom/DomRenderer.ts
@@ -168,7 +168,7 @@ export class DomRenderer extends Disposable implements IRenderer {
     styles +=
       `@keyframes blink_box_shadow` + `_` + this._terminalClass + ` {` +
       ` 50% {` +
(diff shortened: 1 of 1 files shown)


Tyriar:
@jerch want to check this one out considering your work on https://github.com/xtermjs/xterm.js/pull/4605?

jerch:
> @jerch want to check this one out considering your work on #4605?

Can do that. Though I wonder, why it was done with an inset box-shadow in the first place :thinking:

Tyriar:
Can't remember... Maybe just to ensure it stays within the cell?

jerch (approved):
Tested several zoom level (dpr changes) and font sizes - renders crisper than the old approach for certain settings. Could not find any glitch either.

Issue #4560, opened 2023-06-15
"Cannot read properties of undefined (reading 'getRasterizedGlyph')" when a target element is not displayed


schoonc:
After upgrading from `4.18.0` to `5.2.1`, if the mount point is not displayed (e.g. some parent has `display: none`), this causes an error and `xterm` crashes. In `4.18.0` this problem was not present.

## Details
- Browser and browser version: `Vivaldi 6.1.3035.50`
- OS version: `Ubuntu 23.04`
- xterm.js version: `5.2.1`

### Steps to reproduce

Code:

```html
<div style="display: none" id="target"></div>
```

```js
import { Terminal } from 'xterm'
import { CanvasAddon } from 'xterm-addon-canvas'

const terminal = new Terminal()
terminal.loadAddon(new CanvasAddon())
terminal.open(document.getElementById('target'))
```

Console output:
![image](https://github.com/xtermjs/xterm.js/assets/4659633/0ce22830-66df-441d-8fe1-f3ceefe3c013)

JounQin:
Any schedule to fix this issue?

I noticed there are also several `_charAtlas?.` usage, can we just add `if (!this._charAtlas) { return }`? (Sorry I'm not familiar with the source codes)

AndreSilva1993:
Will this be released? This issue is the only thing blocking me to upgrade since it breaks our automated tests.

Tyriar:
@AndreSilva1993 you can use the latest beta build which is released on every commit, VS Code is periodically updated to the latest beta

PR #4596, opened 2023-07-14
Add getRasterizedGlyph null and undefined check

lewis-sanchez:
Ensures that the `charAtlas` object is not undefined or null before getRasterizedGlyph is invoked.

This PR fixes https://github.com/xtermjs/xterm.js/issues/4560


Tyriar (approved):

Tyriar:
Thanks!

PR #4605, opened 2023-07-21
Faster DOM renderer

### LINKED ISSUE #4554: Hyperlink dotted underline disappears after hover leave
## Details
- Browser and browser version: Firefox 113.0.2
- OS version: NixOS on https://github.com/NixOS/nixpkgs/tree/7409480d
- xterm.js version: 5.1.0

### Steps to reproduce

1. Write an OSC 8 Hyperlink into a non-WebGL terminal. Notice that there's a dotted underline underneath the link (WebGL terminals don't seem to have dotted underlines at all).
   image
2. Hover over the link.
   image
3. Move the mouse away. Notice that the dotted underline is gone.
   image

<details>
<summary>Reproducing HTML</summary>

```html
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="node_modules/xterm/css/xterm.css" />
    <script src="node_modules/xterm/lib/xterm.js"></script>
  </head>
  <body>
    <div id="terminal"></div>
    <script>
      var term = new Terminal();
      term.open(document.getElementById("terminal"));
      term.write(
        "\033]8;;http://example.com\033\\This is a link\033]8;;\033\\\n"
      );
    </script>
  </body>
</html>
```

</details>

### LINKED ISSUE #4556: Possible race in DomRenderer: Cannot read properties of undefined (reading 'replaceChildren')
1. In VS Code OSS, check out 59d8f2bd9c6
2. Run a couple of unit tests, with the [selfhost test provider](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-selfhost-test-provider) for example, and open the TestResults view
3. Click on the two top-level test result nodes to switch terminals.
4. Notice you'll eventually get an error in the console. Her
e's the debug state, it looks like there might be an off-by-one error where `y` references a row element that does not yet exist

image

### LINKED ISSUE #4601: Last char on the line not visible (wrapped) for some fonts
## Details
- Browser and browser version: FireFox 115.0.1 (snap version)
- OS version: Ubuntu 22.04.2 LTS
- xterm.js version: 5.2.1

### Steps to reproduce
1. Display 80 characters on a line
2. Try some options.fontSize value until the last character disappear (wrapped).
    Size 21 or 24 in my case (may depends on installed fonts so difficult to reproduce).

If the cursor is moved on the hidden last char, it also disappear.

### Fix
The line element inline style is in my case:
```css
   {
     width: 812px;
     height: 21.3333px;
     line-height: 21.3333px;
     overflow: hidden;
   }
```
Two ways to fix:
1. Increase the width by 1 pixel (not very nice)
2. Add  `white-space: nowrap` (it is a more resilient fix)

### LINKED ISSUE #4602: Uncaught TypeError: Cannot read properties of undefined (reading 'replaceChildren')
From VS Code's error telemetry:

```javascript
TypeError: Cannot read properties of undefined (reading 'replaceChildren')
at g.renderRows in vscode-file://vscode-app/node_modules.asar/xterm/lib/xterm.js:1:81207
at g.handleSelectionChanged in vscode-file://vscode-app/node_modules.asar/xterm/lib/xterm.js:1:79465
at u._renderRows in vscode-file://vscode-app/node_modules.asar/xterm/lib/xterm.js:1:102027
at t.RenderDebouncer._renderCallback in vscode-file://vscode-app/node_modules.asar/xterm/lib/xterm.js:1:99877
at t.RenderDebouncer._innerRefresh in vscode-file://vscode-app/node_modules.asar/xterm/lib/xterm.js:1:20236
at <anonymous> in vscode-file://vscode-app/node_modules.asar/xterm/lib/xterm.js:1:19934
```

https://github.com/microsoft/vscode/issues/188370

### LINKED ISSUE #4604: faster DOM renderer
The performance of the DOM renderer is very painful. In several local attempts I tried to isolate the reasons behind this and well, it mainly boils down to the sheer amount of [span elements x CSS rules], which gives the browser engine a hard time to update & layout things.

I was able to find a solution to reduce the "element pressure" by merging cells, that share the same fg/bg settings and where the char content gets properly layouted by the browsers font renderer. This gives a very nice perf boost (test is `ls -lR /usr`):

- master
  grafik
- WIP branch
  grafik
- compared to webgl renderer
  grafik

Observations from above:
- no more red flagged frames in devtools - most frames can be updated in reasonable time and we see almost no discarded screen refreshs anymore
- the "stattering" got much better - while the DOM renderer still saturates the CPU (the pie chart is still at ~100% usage), the rendering task is like ~7 times faster lifting some of the burden
- JS runtime got much worse, prolly due to the additional merging logic (well, its just a quickly hacked WIP atm)
- overall runtime is on par with webgl renderer with ~14s (prolly hitting an OS/PTY limit here)
- ofc webgl is still far ahead finishing the test rather quick while the CPU idles for ~40% (scales much better in bigger integrations)

So while this looks rather promising, there are ofc downsides - most prominent the fact, that a cell does not strictly refer to a single span element anymore. This currently creates glitches during selection, and most likely will break with any other 3rd party code relying on the a nice cell-span matrix (already an issue for double width chars). Furthermore the merging relies on a "close-enough" threshold approximation, thus will create tiny pixel offsets for bigger merges within a line.

jerch:
Fixes #4604.

TODO:
- [x] fix render glitches:
  - [x] selection in merged spans does not show chars
  - [x] cursor within merged span does not show char
  - [x] extended attribs dont account yet for new span creation (see colored underline in underline test)
  - [x] underlines on hovered links not working
  - [x] check ligature joins (see notes below)
- [x] make metrics calc lazy / recalc on font changes
- [x] fix bad render runtime for metrics calc
- [x] fix bad runtime of merge logic
- [x] better codepoint selection for optimization
- [x] investigate whether to ignore tiny spacings - _Solved by default spacing on all spans._
- [x] ~investigate, whether `inline-block` on spans is really needed (+20% render penalty)~ - _Postponed for own PR._
- [x] make font weights configurable
- [x] ~avoid full re-render in `handleSelectionChanged`~ - _Postponed._
- [x] test cases
- [ ] create follow-up tickets/PRs from postponed issues:
  - [ ] `inline-block` vs. `inline`
  - [ ] more advanced `handleSelectionChanged` logic
  - [ ] API to share width cache across terminal instances
  - [ ] merge measure logic in CharSizeService & WidthCache with proper event listening, also needs re-eval of dimension logic due to offsetWidth change in measuring
- [x] change measuring to offsetWidth, see #4366

Also fixes #4554.
Also fixes #4602.
Also fixes #4601.
Also fixes #4556.


willmcgugan:
Looking forward to this one! 😀

jerch:
Geez, the branch/commit history is very broken (prolly from my older attempts), needs serious rebasing...

jerch:
@willmcgugan Indeed, this promises some serious perf boost for DOM renderer. Still it creates quite some glitches, idk yet if they all can be worked around.

jerch:
@Tyriar Plz have a look at this profiler stats:

grafik

Is there a specific reason, why the selection service calls into `renderRows`, although I have nothing selected at all? Note that it also takes twice as long as from the render debouncer. This kinda  looks like a big perf smell to me, but idk the inner works of the selection service.

jerch:
@Tyriar The last commit can fix the superfluous paints from the selection service, saving ~1s runtime for my test. Would be good if you can have a look at it and check, whether thats a viable way to debounce the selection refreshs.

jerch:
Eww - the link underline logic for mouse hovers needs a full rework with the merged spans, it simply cannot be done that way anymore :cry:

jerch:
@Tyriar The following test is failing on windows only for no obvious reason: https://github.com/xtermjs/xterm.js/blob/1c98037eede9b6cb9610e20372ec43b6b19e69b7/addons/xterm-addon-search/test/SearchAddon.api.ts#L245-L261

Tested it manually under linux, where it works as expected. Since I have no working windows installation I cannot look into it. Any clue why this might be failing?

jerch:
About ligature joins (all manually tested):
- View composition works as expected.
- View on selection from the left side works as expected. From the right side the right-most chars still get blacked-out until the left-most of the ligature is selected too. Thats a quite weird behavior, not quite sure yet, whats causing this.
- Moving the cursor into a a ligature shows a similar misbehavior - on the left-most cell the whole ligature gets "cursored", for any other cell to the right the cursor vanishes.

jerch:
I'd say this implementation is ready for first alpha tests. It would be very helpful if anyone interested in testing this could run it with any sort of complicated cmdline tools, so we can spot left-over regressions.

jerch:
Not sure yet what to do about the bold, italic and bold&italic font variants. Currently it is assumed, that a proper monospace font wont have different glyph widths for those, but this gonna break really bad if only the regular variant is available and the font renderer has to use a fallback font.

The issues with also calculating these variants are:
- they are pretty rare in a terminal context, might not be worth to bother at all
- multiplies calc runtime by 4 to [regular, bold, italic, bold&italic] - currently regular takes ~500ms for codepoint 32 to 1423
- merge logic gets more cumbersome

Possible solutions:
- blacklist them from the merge dropping back to single cell spans (re-introduces ugly glyph cuts in italic)
- calc them as well, but on a smaller codepoint range (e.g. ASCII only, should still address >>90% of terminal chars)

jerch:
@Tyriar Gonna blacklist bold & italic variants for now - does not seem worth it for the runtime penalty it would introduce.

mofux:
> Currently it is assumed, that a proper monospace font wont have different glyph widths for those, but this gonna break really bad if only the regular variant is available and the font renderer has to use a fallback font

Yep, this will likely happen. When I implemented my own DOM based xterm renderer a while back, I worked around this problem the following way:
- the default character width is based on the measured with of a test character, e.g. "W"
- measure the actual width of every character that should be drawn and remember it in a map for caching (most should have the same width as your test character)
- similar to how you are splitting multiple characters based on different attributes, you now also have to split them based on their width. Whenever one character has a different width to the width of its predecessor, it needs to go into a separate span.
- and now comes the trick: if the character width is not the same as your test character width, calculate the offset, and then apply this offset to the span using the `letter-spacing` css rule. This rule allows you to add a positive or negative offset to a character in `px`. This will make sure that those non-standard width characters get the same width as your default character. And it even works if you have multiple non-standard width characters in the same span, just as long as they have the SAME non-standard width.

I hope you get the idea.

jerch:
@mofux Oh hi there! Your `letter-spacing` rings a bell - have we talked about that in the past? Gonna look into that.

> When I implemented my own DOM based xterm renderer a while back...

Now I am curious - how does your implementation compare to the current DOM renderer? Do you treat all characters that way or do you pre-limit chars to account? I found that the width measuring is very expensive, thus I limited it to ASCII + latin1 supplement, while everything else still gets its own span. This helps for >90% if terminal data, but ofc not for content outside of latin.

jerch:
The width calculation is still very expensive. Currently it uses `requestIdleCallback` to do the full roundtrip of adding measure elements + getting the bounding rects, which causes alot of forced layouting. I can only do ~30 codepoints/call this way to not introduce alot of blocking time, overall it takes ~2s for 4096 codepoints. Hmm.

There are still a few ideas to get this faster:
- Bind to `requestAnimationFrame` instead and slice work into a placement and a measuring step. I hope to get higher codepoints/call by aligning the measurement to normal render cycles.
- desperate: Check if painting to a canvas yields the same widths, if so thats prolly much faster.
- desperate & last resort: Put things into an iframe. There we can operate in much bigger batches, as blocking time is no concern. Well thats only a last resort creating a lot of fuzz like issues from security settings or how to correctly set up all needed CSS rules there.

I def. wanna try mofux's `letter-spacing` idea - it opens the field for handling a bigger codepoint range in a neat way. My idea is now to warm up with ASCII range only and to do higher BMP codepoints lazy on-the-fly.

mofux:
> Now I am curious - how does your implementation compare to the current DOM renderer? Do you measure all characters that way or do you pre-limit chars to account

I'm measuring any character as it appears for the first time. No warm-up. I'm using the [monaco API](https://github.com/microsoft/vscode/blob/96ebd364f01c62305037073a1e994110d5f378c6/src/vs/editor/browser/config/charWidthReader.ts#L140-L143) for this measurement which already utilizes a caching layer, so chances are that a lot of characters are already measured because of a previously opened editor.

I did not notice any performance issue with this. Filling a 1000 line scrollback with random chars spends about 50ms on the measure call in total (over the course of 1s) - but I might be missing something. There's a lot going on behind the scenes of the monaco editor itself that is rendering out the lines.

jerch:
@mofux Thx for the pointer to the monaco API. I am pretty sure that this isnt faster than my current approach, as the API measures chars one-by-one, which I did before as well. I already optimized that to batches of 30 chars at once, which is ~5 times faster than before.
There are still perf smells in my setup though (like complicated CSS rules), monaco's code doesn't need them and I can prolly simplifiy things as well.

> I'm measuring any character as it appears for the first time. No warm-up.

Hmm, gonna try that, maybe the penalty isnt that big. Didn't even think about this obvious way, as doing the measuring in the render call creates much more relayouting stress, thus looked like an anti-pattern to me.

src/browser/renderer/dom/DomRenderer.ts
```diff
@@ -37,7 +46,10 @@ export class DomRenderer extends Disposable implements IRenderer {
   private _rowContainer: HTMLElement;
   private _rowElements: HTMLElement[] = [];
   private _selectionContainer: HTMLElement;
-  private _cellToRowElements: Int16Array[] = [];
+  private _linkState = new Uint8Array(3);
```
Tyriar, line 49: It's not clear what this represents

src/browser/renderer/dom/DomRenderer.ts
```diff
@@ -91,6 +103,57 @@ export class DomRenderer extends Disposable implements IRenderer {
...
+      container.appendChild(el);
+    }
+    parent.appendChild(container);
+
+    const collection = container.children;
+    for (let i = 0; i < collection.length; ++i) {
+      const width = collection[i].getBoundingClientRect().width;
+      this._fontMetrics[i + this._metricsPos] = +(width < lower || width > upper);
+    }
+    container.remove();
```
Tyriar, line 142: Would it be better to split this into its own class which manages cached `container` and its children until the task is done?

src/browser/renderer/dom/DomRenderer.ts
```diff
@@ -91,6 +103,57 @@ export class DomRenderer extends Disposable implements IRenderer {
...
+      el.classList.add('xterm-char-measure-element');
+      el.textContent = String.fromCharCode(i).repeat(10);
+      container.appendChild(el);
+    }
+    parent.appendChild(container);
+
+    const collection = container.children;
+    for (let i = 0; i < collection.length; ++i) {
+      const width = collection[i].getBoundingClientRect().width;
+      this._fontMetrics[i + this._metricsPos] = +(width < lower || width > upper);
```
Tyriar, line 140: 🤨 is this setting 1 if not `lower < width < upper`? A comment or clearer code would be good as I'm just guessing 😅

src/browser/renderer/dom/DomRenderer.ts
```diff
@@ -348,6 +409,7 @@ export class DomRenderer extends Disposable implements IRenderer {
     this._updateDimensions();
     // Refresh CSS
     this._injectCss(this._themeService.colors);
+    this._cacheMetrics();
```
Tyriar, line 412: `_warmUpFontMetrics`? Does this check if they aren't measured yet when it needs to be used? A class could help this too by caching a codepoint synchronously at the time of its use.

src/browser/renderer/dom/DomRenderer.ts
```diff
@@ -393,55 +464,46 @@ export class DomRenderer extends Disposable implements IRenderer {
...
-     *
-     * For negative y-values we can simply adjust x = 0,
-     * as higher up link start means, that everything from
-     * (0,0) is a link under top-down-left-right char progression
-     *
-     * Additionally there might be a small chance of out-of-sync x|y-values
-     * from a race condition of render updates vs. link event handler execution:
-     * - (sync) resize: chances terminal buffer in sync, schedules render update async
-     * - (async) link handler race condition: new buffer metrics, but still on old render state
-     * - (async) render update: brings term metrics and render state back in sync
```
Tyriar, line 409: Does this not happen anymore?

src/browser/renderer/dom/DomRenderer.ts
```diff
@@ -393,55 +464,46 @@ export class DomRenderer extends Disposable implements IRenderer {
...
       }
-      if (++x >= cols) {
-        x = 0;
-        y++;
+      if (enabled) {
+        this._linkState[1] = i === y ? x : 0;
+        this._linkState[2] = (i === y2 ? x2 : cols) - 1;
       }
+      const row = i + this._bufferService.buffer.ydisp;
+      const lineData = this._bufferService.buffer.lines.get(row);
```
Tyriar, line 492: Declaring the variables outside the loop will save some memory getting allocated

src/browser/renderer/dom/DomRendererRowFactory.test.ts
```diff
@@ -14,7 +14,9 @@ import { MockCoreService, MockDecorationService, MockOptionsService } from 'comm
 import { css } from 'common/Color';
 import { MockCharacterJoinerService, MockCoreBrowserService, MockThemeService } from 'browser/TestUtils.test';
 
-const EMPTY_ELEM_MAPPING = new Int16Array(1000);
+const EMPTY_METRICS = new Uint8Array(1024);
```
Tyriar, line 17: I was expecting `(256)`?

Tyriar (changes requested):

src/browser/renderer/dom/DomRenderer.ts
jerch: Yeah, thats already in the making, as the letter spacing idea changes too much, and I need myself something like an API :smile:

src/browser/renderer/dom/DomRenderer.ts
jerch: Yeah the old logic just separates between 1|0 state for being at default span width or outside. With letter spacing this gonna be a float offset container, thus this hackish line prolly gets removed.

src/browser/renderer/dom/DomRenderer.ts
jerch: Class is in the making, gets too confusing already for myself.

src/browser/renderer/dom/DomRenderer.ts
jerch: This still happens, guess I forgot to copy over the comments (clamping code is still the same).

src/browser/renderer/dom/DomRendererRowFactory.test.ts
jerch: To be hidden within a class (thus not being exposed anymore here).

src/browser/renderer/dom/DomRenderer.ts
jerch: Oh and thats even the wrong typed array type :blush:. To be reworked...

jerch:
The basic rework to letter-spacing is done - up for alpha tests again.

So far the performance is really good (well as long as a monospace font is set) - it is much faster than the canvas renderer on my machine, and almost as fast as the webgl renderer (but more CPU hungry). The output and things like scrolling is really smooth now.

Still left to do:
- more unit & API tests
- ligature joiner still not working correctly
- cleanup

jerch:
some profiler graphics (tested with `ls -lR /usr` with one warm up call before):
- this PR
  grafik
- webgl
  grafik
- canvas
  grafik

The DOM renderer seems to render all frames fine, even the CPU is not at 100% anymore (upper greyish graph show bulges).
The canvas renderer is the slowest now, with several frames not finishing in time (that yellow-greenish areas in frames). It also has very high JS runtime compared to the others (prolly due accounting `drawImage` under JS).

src/browser/renderer/dom/SpacingCache.ts
```diff
@@ -0,0 +1,130 @@
...
+  FLAT_SIZE = 256,      // codepoint upper bound to handle in flat cache
+  REPEAT = 32           // char repeat for measuring
+}
+
+
+export class SpacingCache implements IDisposable {
+  // flat cache for regular
+  private _flat = new Float32Array(CacheSettings.FLAT_SIZE);
+  // holey cache for bold, italic and bold&italic for any string
+  private _holey = new Map<string, number>();
```
Tyriar, line 28: Doesn't need to be done in this PR, but sharing it between terminals similar to the texture atlas would be nice. Managing them is a matter of acquiring and releasing depending on the terminals font config.

jerch: Sure, we can also abstract it further to meet a better font config interface. My font family string + a pixel size is quite lousy still (I saw that monaco expects here more style settings, which prolly helps to avoid surprising render issues by awkward css rules on the global document.)

jerch:
The single letter based letter-spacing method has one drawback - it cannot deal with kerning of proportional fonts, thus might create tiny offsets for fonts doing this. Ofc in a terminal a non-monospace font is ugly and should not be used in the first place.

I still wonder though, if we can address this issue by a second measuring of the merged span and applying a correction value. This will penalize render speed a lot, but maybe it is possible to do this additional measuring on degenerated spans only.

jerch:
The last commit disables font kerning, with this even fonts like Arial etc. align correctly to a grid. I dont see a perf penalty from this, but I am not sure if it will have other negative side effects (if so, then we should prolly make this an explicit setting to be used with certain fonts only).

jerch:
@Tyriar I changed the cache into a width cache, this way it can be re-used for the same font settings (before with px value differences it was also tied to the unicode provider).

src/browser/renderer/dom/DomRenderer.ts
jerch: I re-added the notes about the edge cases. Though the new code needs less range checks, as the data structures that would overflow from this are not needed anymore.

Also the explicit re-rendering of hyperlink lines avoids a faulty in-between state. It still cannot handle a late render update by other means, which might steal the underline/link state right after. Imho thats a rather rare edge case and not worth to work around (would need state holding of active hyperlink in renderer).

jerch:
@Tyriar Seems the faulty test from above works again, weird.

jerch:
@Tyriar Found the reason for the selection refresh above - during lots of input with terminal scrolling a refresh gets called unconditionally here: https://github.com/xtermjs/xterm.js/blob/17b8247c3dc33c457e950da9f67ad137a74de069/src/browser/Terminal.ts#L534-L537

Since there was no additional selection check in `refresh` before, it would trigger the whole update cascade into the renderers. And the DOM renderer does a full viewport re-rendering on the selection event causing significant render stress.

I dont good other options than preventing the overhead at `refresh` itself.

jerch:
@Tyriar I think the selection handling of the DOM renderer needs some rewrite. The issue currently is that my debouncer in `SelectionService.refresh` omits any (empty) updates, but these updates are also relevant in DOM renderer to adjust previous selection elements, be it for resize, font size changes or dpr changes. In the end there is logic linked to `handleSelectionChanged` that has nothing to do with changed selections.

A possible fix would involve:
- store active selection in renderer
- create a `renderSelection` method with diff update from prev to current selection
- bind method to resize, dpr and charWidth change

jerch:
There is another issue hidden in the DOM renderer, that would fix several rendering glitches and be more performant - `inline-block` vs `inline` display for spans. We currently do `inline-block` on rows with a fixed height with `overflow: hidden`, which creates several glitches like cutting off over or underlines for certain font sizes or dpr. Also `inline-block` has a render penalty:
- current PR state with `inline-block` **~53 FPS** (750 frames in ~14s)
- spans with `inline` **~56 FPS** (780 frames in ~14s)
- compared to webgl **~57 FPS** (800 frames in ~14s)

This PR already achieves really good FPS compared to master (**~17 FPS**, 420 frame in ~25s), so we already dealt with the low hanging fruit. Imho the `inline` issue can be addressed in a later PR (also this is not a regression, it is just broken the same certain way).

jerch:
@Tyriar FYI - The last 2 commits revert the `inline` display and the selection service debouncer. This adds ~1.5s to the runtime again lowers the frame rate to ~50 FPS on my machine.

jerch:
About ligatures - with the letter-spacing merging ligatures work automatically, if a ligature font is installed. In fact the ligature addon should not be used anymore, as it makes things worse (the line on the right is a box border to see if there are offset glitches):
- w'o ligature addon the cursor can address single chars (temporarily breaking the ligature):
recorded

- with ligature addon the cursor vanishs until reaching the leading ligature char, then highlighting the whole ligature:
recorded


Idk know whats the desired bahavior here, we prolly should document the difference though.

addons/xterm-addon-search/test/SearchAddon.api.ts
```diff
@@ -242,7 +242,8 @@ describe('Search Tests', function (): void {
           { resultCount: 1000, resultIndex: 1 }
         ]);
       });
-      it('should fire when writing to terminal', async () => {
+      // FIXME: skipped due to failing on windows
+      it.skip('should fire when writing to terminal', async () => {
```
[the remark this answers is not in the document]
jerch: No clue why this happened, no clue how it got resolved, its gone. :smile_cat:

src/browser/renderer/dom/SpacingCache.ts
jerch: Listed for follow-up PR.

mofux (approved):
I gave it a good test and it works very well for me (Linux with Edge and Firefox). The performance is impressive - well done! It's really only the block characters where the WebGL renderer is still superior. There certainly are ways to also add those custom-drawn block elements to the DOM renderer, but that is a task for another day :)

jerch:
@mofux Yeah the performance is quite impressive for the fact, that it is still DOM manipulations going on. Thx again for the letter-spacing trick, it is def. the better way to deal with it. As noted above, there are ~20% more perf bonus possible, if we manage to get the spans into `inline`, thus completely rely on the layouter to do its job. And it would solve all the over-/underline issues, and make Bidi partially work too.

About the block chars - yeah kinda all font renderer libs have issues with those in monospace, as far as I am aware almost all other TEs render those themselves to overcome the font lib shortcomings. I have not thought about a workaround for the DOM renderer yet, surely we can find here something.

src/browser/services/CharSizeService.ts
```diff
@@ -7,7 +7,10 @@ import { IOptionsService } from 'common/services/Services';
 import { EventEmitter } from 'common/EventEmitter';
 import { ICharSizeService } from 'browser/services/Services';
 import { Disposable } from 'common/Lifecycle';
-import { ITerminalOptions } from 'common/Types';
+
+
+const CHAR_REPEAT = 32;
```
Tyriar, line 12: Let's make this a `const enum` so it's inlined at compile time

src/browser/renderer/dom/WidthCache.ts
```diff
@@ -0,0 +1,148 @@
...
+
+import { IDisposable } from 'common/Types';
+import { FontWeight } from 'common/services/Services';
+
+
+export const enum WidthCacheSettings {
+  FLAT_UNSET = -9999,   // sentinel for unset values in flat cache
+  FLAT_SIZE = 256,      // codepoint upper bound to handle in flat cache
+  REPEAT = 32           // char repeat for measuring
+}
```
Tyriar, line 14: Can we use `/** */` style comments here so they show up in autocomplete?

src/browser/renderer/dom/WidthCache.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  private _font = '';
+  private _fontSize = 0;
+  private _weight: FontWeight = 'normal';
+  private _weightBold: FontWeight = 'bold';
+  private _container: HTMLDivElement;
+  private _measureElements: HTMLSpanElement[] = [];
+
+  constructor(
+    private readonly _document: Document
```
Tyriar, line 38: Let's remove private readonly as it's only used in the ctor

Tyriar (changes requested):
Did a thorough review, looks close

Tyriar (changes requested):
Did a thorough review, looks close

src/browser/renderer/dom/WidthCache.ts
jerch: Sure, gonna change it.

src/browser/renderer/dom/WidthCache.ts
jerch: Woops, left-over from before.

src/browser/renderer/dom/WidthCache.ts
```diff
@@ -0,0 +1,148 @@
...
+  ) {
+    this._container = _document.createElement('div');
+    this._container.style.position = 'absolute';
+    this._container.style.top = '-50000px';
+    this._container.style.width = '50000px';
```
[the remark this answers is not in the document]
jerch: Hmm idk, I basically took that idea from monaco. Can make it smaller, maybe more within typical `fontsize * measureRepeat`.

src/browser/renderer/dom/DomRenderer.ts
```diff
@@ -123,10 +136,9 @@ export class DomRenderer extends Disposable implements IRenderer {
 
     const styles =
       `${this._terminalSelector} .${ROW_CONTAINER_CLASS} span {` +
-      ` display: inline-block;` +
+      ` display: inline-block;` +   // TODO: find workaround for inline-block (creates ~20% render penalty)
```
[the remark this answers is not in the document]
jerch: Yeah, its already listed above under the "create follow-up issues" TODO.

I already tested this locally, but it touched too many aspects (e.g. dimension calc needs rework), thus I left it out of this PR.

src/browser/renderer/dom/WidthCache.ts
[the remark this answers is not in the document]
Tyriar: If it's in monaco I'm sure it's fine

src/browser/renderer/dom/DomRenderer.ts
```diff
@@ -91,7 +94,17 @@ export class DomRenderer extends Disposable implements IRenderer {
...
+      this._widthCache.dispose();
     }));
+
+    this._widthCache = new WidthCache(document);
+    this._widthCache.setFont(
+      this._optionsService.rawOptions.fontFamily,
+      this._optionsService.rawOptions.fontSize,
+      this._optionsService.rawOptions.fontWeight,
+      this._optionsService.rawOptions.fontWeightBold
+    );
```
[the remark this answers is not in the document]
jerch: Yeah well, I already thought to move the measure logic to the `CharSizeService`, which already receives the needed events. Then all thats needed, is to clear the cache on metrics changes. Should I move it over and merge with the measure logic there?

src/browser/renderer/dom/WidthCache.ts
```diff
@@ -0,0 +1,148 @@
...
+  private _weightBold: FontWeight = 'bold';
+  private _container: HTMLDivElement;
+  private _measureElements: HTMLSpanElement[] = [];
+
+  constructor(
+    private readonly _document: Document
+  ) {
+    this._container = _document.createElement('div');
+    this._container.style.position = 'absolute';
+    this._container.style.top = '-50000px';
```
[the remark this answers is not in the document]
Tyriar: Yeah, but if monaco does this it's fine

src/browser/renderer/dom/DomRenderer.ts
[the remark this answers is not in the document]
jerch: Yeah prolly better as follow-up, it also will be affected by the yet to come shared API.

Tyriar (approved):
Looks like everything is checked off and comments resolved so let's merge and I'll pull it into VS Code so we can get some early testing. @jerch can you make the followup issues when you get a chance?

jerch:
> can you make the followup issues when you get a chance?

Yeah all done. :smiley_cat:

PR #4611, opened 2023-07-27
feature: use constructed style sheets for dom render so that it works with csp style-src 'self'

### LINKED ISSUE #4445: Latest version requires unsafe-inline due to inline styles
Content Security Policies need to be set to 'unsafe-inline' to work with xterm.js. Older versions didn't use inline styles so this wasn't an issue.

Ideally xterm should stop using inline styles or support a user-provided nonce value that can be set in the CSP. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src

## Details
- Browser and browser version: all
- OS version: all
- xterm.js version: 5.1.0

### Steps to reproduce

1. Set a content security policy like "style-src 'self';"
2. Make an xterm that has a resizable container
3. Resizing causes CSP errors in the console.
4. Resizing doesn't work properly

SimonSiefke:
Fixes #4445 

This changes the dom renderer to use constructed stylesheets instead of html style elements when available. Constructed stylesheets don't require a `style-src 'unsafe-inline'` Content Security Policy

## test the change locally

1. In demo/client.ts comment out the webgl addon so that the dom renderer is used

```js
 // if (addons.webgl.instance) {
  //   try {
  //     typedTerm.loadAddon(addons.webgl.instance);
  //     term.open(terminalContainer);
  //     setTextureAtlas(addons.webgl.instance.textureAtlas);
  //     addons.webgl.instance.onChangeTextureAtlas(e => setTextureAtlas(e));
  //     addons.webgl.instance.onAddTextureAtlasCanvas(e => appendTextureAtlas(e));
  //     addons.webgl.instance.onRemoveTextureAtlasCanvas(e => removeTextureAtlas(e));
  //   } catch (e) {
  //     console.warn('error during loading webgl addon:', e);
  //     addons.webgl.instance.dispose();
  //     addons.webgl.instance = undefined;
  //   }
  // }


```

2. In demo/server.js, add a content security policy header when serving index.html:

```js
app.get("/", (req, res) => {
    res.setHeader(
      "Content-Security-Policy",
      `default-src 'none'; style-src 'self'; img-src 'self'; script-src 'self'; connect-src 'self'`
    );
    res.sendFile(__dirname + "/index.html");
  });
```

## Demo with CSP enabled

csp-header


Tyriar:
I updated the branch

SimonSiefke:
Fixed! :)

jerch:
@SimonSiefke Can you plz also check, that `replaceSync` does not create another forced layouting step, the last thing we need in the DOM renderer is additional render engine work. So the question here is, if that sync in the name means sync to DOM render state.

Tyriar:
@jerch I get the impression `replaceSync` is doing what we used to do and `replace` would batch it for later if it isn't needed immediately like in our tests. But regardless, even slightly slower here shouldn't matter as it's only on dimensions change (not resize) and changing theme which seldom happens. I double checked the time by resizing a few times on my macbook:

Screenshot 2023-07-27 at 12 44 24 pm

SimonSiefke:
@jerch As far as I understand, there is no extra synchronization in `replaceSync`, it is just the synchronous version of `replace`

<br>

>  Can you plz also check, that replaceSync does not create another forced layouting step

<br>

As @Tyriar said, it is basically the same, both methods are dom writes, there are no extra forced layouts with this method. 

Perhaps performance might even be slightly better because there are two less dom elements now, which is also something but I doubt it would make a huge difference.

jerch:
I stumbled over one aspect in this PR when merging the DOM renderer PR with master - in `StyleSheet.ts` the global `document` is accessed. In an older PR we abstracted `document` into a terminal setting to make popouts into other documents working, as some functionality is bound to the real underlying document.

Question - should the code in `StyleSheet.ts` also refer to our document variable or does it not matter here? Note I had a similar case with cross document canvas painting, but it just works for those across documents.

Tyriar:
@jerch good find, I'll fix it up

YadhuCU:
6.0.0 still relies on inline styles. A previous fix was reverted upstream, so this is a known limitation. Is there any recommended workaround or planned fix to avoid requiring style-src 'unsafe-inline'?

PR #4612, opened 2023-07-27
update typescript version to 5.1.6

SimonSiefke:
required for CssStyleSheet typings needed in #4611


src/browser/Viewport.ts
```diff
@@ -57,7 +57,7 @@ export class Viewport extends Disposable implements IViewport {
     @ICharSizeService private readonly _charSizeService: ICharSizeService,
     @IRenderService private readonly _renderService: IRenderService,
     @ICoreBrowserService private readonly _coreBrowserService: ICoreBrowserService,
-    @IThemeService themeService: IThemeService
+    @IThemeService public themeService: IThemeService
```
Tyriar, line 60: Why were these `public`s added? They aren't needed as properties, only ctor args

src/common/InputHandler.test.ts
```diff
@@ -1950,47 +1950,47 @@ describe('InputHandler', () => {
...
+      assert.deepEqual(stack, [[{ type: ColorRequestType.REPORT, index: 0 as ColorIndex }]]);
       stack.length = 0;
       await inputHandler.parseP('\x1b]4;123;?\x07');
-      assert.deepEqual(stack, [[{ type: ColorRequestType.REPORT, index: 123 }]]);
+      assert.deepEqual(stack, [[{ type: ColorRequestType.REPORT, index: 123 as ColorIndex }]]);
       stack.length = 0;
       // multiple queries
       await inputHandler.parseP('\x1b]4;0;?;123;?\x07');
-      assert.deepEqual(stack, [[{ type: ColorRequestType.REPORT, index: 0 }, { type: ColorRequestType.REPORT, index: 123 }]]);
+      assert.deepEqual(stack, [[{ type: ColorRequestType.REPORT, index: 0 as ColorIndex }, { type: ColorRequestType.REPORT, index: 123 as ColorIndex }]]);
```
Tyriar, line 1960: @jerch this fails because the only value `index` values are 256, 257 and 258. Is this cast covering up some issue?

Tyriar (changes requested):

src/browser/Viewport.ts
SimonSiefke: For some reason ESLint was showing a warning `Missing accessibility modifier on parameter property themeService`, so I just added `public` and it kind of worked.
 
What do you think would be the best way to fix this?

missing-modified

src/common/InputHandler.test.ts
Tyriar: The range thing is kind of nifty if it works, that seems preferable to `| number` if it works. Let's do as part of https://github.com/xtermjs/xterm.js/issues/4614

src/browser/Viewport.ts
Tyriar: That's weird as we explicitly say don't consider constructors:

https://github.com/xtermjs/xterm.js/blob/b4aeb1b3e7e812a82f36c48d844d8607492646bc/.eslintrc.json#L62-L70

We might need to update eslint, @typescript-eslint and/or that rule to be compatible with the new TS version?

SimonSiefke: That worked! I updated eslint, @typescript-eslint/eslint-plugin and @typescript-eslint/parser to the latest version.

Tyriar (approved):
lgtm if CI passes, thanks

PR #4618, opened 2023-07-28
Add IntRange type for ColorIndex

### LINKED ISSUE #4614: Avoid as ColorIndex in src/common/InputHandler.test.ts
Context: https://github.com/xtermjs/xterm.js/pull/4612#discussion_r1276337382

Tyriar:
Fixes #4614

### DIFF
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -2915,7 +2915,7 @@ export class InputHandler extends Disposable implements IInputHandler {
       const spec = slots.shift() as string;
       if (/^\d+$/.exec(idx)) {
         const index = parseInt(idx);
-        if (isValidColorIndex(index) && index < 256) {
+        if (isValidColorIndex(index)) {
           if (spec === '?') {
             event.push({ type: ColorRequestType.REPORT, index });
           } else {
@@ -3073,7 +3073,7 @@ export class InputHandler extends Disposable implements IInputHandler {
     for (let i = 0; i < slots.length; ++i) {
       if (/^\d+$/.exec(slots[i])) {
         const index = parseInt(slots[i]);
-        if (isValidColorIndex(index) && index < 256) {
+        if (isValidColorIndex(index)) {
           event.push({ type: ColorRequestType.RESTORE, index });
         }
       }
@@ -3431,5 +3431,5 @@ class DirtyRowTracker implements IDirtyRowTracker {
 }
 
 function isValidColorIndex(value: number): value is ColorIndex {
-  return 0 <= value && value < 259;
+  return 0 <= value && value < 256;
 }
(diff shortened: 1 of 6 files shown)


src/common/InputHandler.ts
```diff
@@ -3429,3 +3429,7 @@ class DirtyRowTracker implements IDirtyRowTracker {
     this.markRangeDirty(0, this._bufferService.rows - 1);
   }
 }
+
+function isValidColorIndex(value: number): value is ColorIndex {
+  return 0 <= value && value < 259;
+}
```
Tyriar, line 3435: @jerch should this be < 256 instead? If so what would be a good name for this color index sub-range? `IsValidInputColorIndex` or something?

jerch: Yes, for colors the index is only 0..255, so its better to test for that range in during input. The fg/bg/cursor entries on top are an amalgamation, so the events and methods dont need re-declaration and/or branching.

src/common/Types.d.ts
```diff
@@ -413,7 +413,16 @@ export const enum ColorRequestType {
...
 }
-export const enum ColorIndex {
+
+// IntRange from https://stackoverflow.com/a/39495173
+type Enumerate<N extends number, Acc extends number[] = []> = Acc['length'] extends N
+  ? Acc[number]
+  : Enumerate<N, [...Acc, Acc['length']]>;
+type IntRange<F extends number, T extends number> = Exclude<Enumerate<T>, Enumerate<F>>;
+
+type ColorIndex = IntRange<0, 259>; // number from 0 to 258
```
jerch, line 423: Would it be clearer to limit `ColorIndex` here to 0..255, and to merge into a super type with the fg/bg/cursor values in a second step? Also the test function above can then refer to that narrower `ColorIndex` type.

src/common/InputHandler.ts
```diff
@@ -2915,7 +2915,7 @@ export class InputHandler extends Disposable implements IInputHandler {
       const spec = slots.shift() as string;
       if (/^\d+$/.exec(idx)) {
         const index = parseInt(idx);
-        if (0 <= index && index < 256) {
+        if (isValidColorIndex(index) && index < 256) {
```
jerch, line 2918: Prolly can be simplified with the recent type changes?

src/common/InputHandler.ts
```diff
@@ -3073,7 +3073,7 @@ export class InputHandler extends Disposable implements IInputHandler {
     for (let i = 0; i < slots.length; ++i) {
       if (/^\d+$/.exec(slots[i])) {
         const index = parseInt(slots[i]);
-        if (0 <= index && index < 256) {
+        if (isValidColorIndex(index) && index < 256) {
```
jerch, line 3076: Same as above.

jerch (approved):
LGTM, beside the left-over doubled change check below.

PR #4648, opened 2023-08-08
fix double underline, upwards

### LINKED ISSUE #4476: canvas renderer and curly
Not sure if this is linked to the other canvas underline issues, but the curly underline does not look very curly to me:

image

(Get the same output in Chrome && Firefox...)

tisilent:
fix #4476

### DIFF
--- a/src/browser/renderer/shared/TextureAtlas.ts
+++ b/src/browser/renderer/shared/TextureAtlas.ts
@@ -534,6 +534,7 @@ export class TextureAtlas implements ITextureAtlas {
       const yTop = Math.ceil(padding + this._config.deviceCharHeight) - yOffset;
       const yMid = padding + this._config.deviceCharHeight + lineWidth - yOffset;
       const yBot = Math.ceil(padding + this._config.deviceCharHeight + lineWidth * 2) - yOffset;
+      const ySpace = lineWidth * 2;
 
       for (let i = 0; i < chWidth; i++) {
         this._tmpCtx.save();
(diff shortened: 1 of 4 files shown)


src/browser/renderer/shared/TextureAtlas.ts
```diff
@@ -1019,13 +1020,13 @@ function clearColor(imageData: ImageData, bg: IColor, fg: IColor, enableThreshol
...
-        imageData.data[offset + 2] === b) {
+      imageData.data[offset + 1] === g &&
+      imageData.data[offset + 2] === b) {
       imageData.data[offset + 3] = 0;
     } else {
       // Check the threshold based difference
       if (enableThresholdCheck &&
-          (Math.abs(imageData.data[offset] - r) +
+        (Math.abs(imageData.data[offset] - r) +
           Math.abs(imageData.data[offset + 1] - g) +
```
Tyriar, line 1030: Can you revert these? The intent here is to double-indent wrapped lines to easily differentiate them from the lines inside the block. You might need to disable prettier if you're using that.

src/browser/renderer/shared/TextureAtlas.ts
```diff
@@ -542,10 +543,10 @@ export class TextureAtlas implements ITextureAtlas {
         const xChMid = xChLeft + this._config.deviceCellWidth / 2;
         switch (this._workAttributeData.extended.underlineStyle) {
           case UnderlineStyle.DOUBLE:
+            this._tmpCtx.moveTo(xChLeft, yTop - ySpace);
+            this._tmpCtx.lineTo(xChRight, yTop - ySpace);
             this._tmpCtx.moveTo(xChLeft, yTop);
             this._tmpCtx.lineTo(xChRight, yTop);
```
Tyriar, line 549: Since we're fixing a limitation of the canvas renderer, we still want the old ideal behavior on the webgl renderer. So this should be based on a flag passed into `_drawToCache`, something like `restrictToCellHeight: boolean`

Tyriar (changes requested):

src/browser/renderer/shared/TextureAtlas.ts
tisilent: ou.....I restore it. 😶‍🌫️

Tyriar (approved):
Works great 👍 

I created https://github.com/xtermjs/xterm.js/issues/4653 as we want a similar fix to curly too, ideally we would have a generic solution for this problem that ensures all underlines are within the cell for that.

PR #4657, opened 2023-08-10
Add cursorInactiveStyle option

### LINKED ISSUE #4566: Add cursorStyleInactive option
Follow up to https://github.com/xtermjs/xterm.js/pull/4443

Since we reverted how it used to be (which I still think is the right default), some embedders/users may want it the old way still. We can solve this for everyone by adding a `cursorInactiveStyle` as suggested by @boettges in https://github.com/xtermjs/xterm.js/pull/4443#issuecomment-1595056304. The name `cursorInactiveStyle` aligns with `selectionInactiveBackground`

To implement this you would need to:

1. Add the new option to xterm.d.ts, options service, etc.
2. Implement it for dom, canvas and webgl renderers

tisilent:
fix #4566

### DIFF
--- a/addons/xterm-addon-webgl/src/WebglRenderer.ts
+++ b/addons/xterm-addon-webgl/src/WebglRenderer.ts
@@ -461,7 +461,7 @@ export class WebglRenderer extends Disposable implements IRenderer {
               y: this._terminal.buffer.active.cursorY,
               width: cell.getWidth(),
               style: this._coreBrowserService.isFocused ?
-                (terminal.options.cursorStyle || 'block') : 'blur',
+                (terminal.options.cursorStyle || 'block') : this._getInactiveCursorStyle(terminal.options.cursorInactiveStyle),
               cursorWidth: terminal.options.cursorWidth,
               dpr: this._devicePixelRatio
             };
@@ -599,6 +599,19 @@ export class WebglRenderer extends Disposable implements IRenderer {
   private _requestRedrawCursor(): void {
     const cursorY = this._terminal.buffer.active.cursorY;
     this._onRequestRedraw.fire({ start: cursorY, end: cursorY });
+  }
+
+  private _getInactiveCursorStyle(cursorInactiveStyle: 'outline' | 'line' | 'underline' | 'none'): string {
+    if (cursorInactiveStyle === 'outline') {
+      return 'blur';
+    }
+    if (cursorInactiveStyle === 'line') {
+      return 'bar';
+    }
+    if (cursorInactiveStyle === 'underline'){
+      return 'underline';
+    }
+    return 'block';
   }
 }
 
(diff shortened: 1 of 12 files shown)


tisilent:
The identifier declared by cursorInactiveStyle is somewhat different from that of cursorStyle. The displayed effect is consistent.

demo/client.ts
```diff
@@ -419,6 +419,7 @@ function initOptions(term: TerminalType): void {
   ];
   const stringOptions = {
     cursorStyle: ['block', 'underline', 'bar'],
+    cursorInactiveStyle: ['outline', 'line', 'underline', 'none'],
```
Tyriar, line 422: Let's copy cursorStyle:

```json
cursorInactiveStyle: ['outline', 'block', 'underline', 'bar', 'none']
```

We may as well add block as it's trivial and we want to align the naming of bar and line

src/browser/renderer/dom/DomRendererRowFactory.ts
```diff
@@ -27,7 +27,11 @@ export const enum RowCss {
   CURSOR_BLINK_CLASS = 'xterm-cursor-blink',
   CURSOR_STYLE_BLOCK_CLASS = 'xterm-cursor-block',
   CURSOR_STYLE_BAR_CLASS = 'xterm-cursor-bar',
-  CURSOR_STYLE_UNDERLINE_CLASS = 'xterm-cursor-underline'
+  CURSOR_STYLE_UNDERLINE_CLASS = 'xterm-cursor-underline',
+  CURSOR_INACTIVE_STYLE_OUTLINE_CLASS = 'xterm-cursor-inactive-outline',
+  CURSOR_INACTIVE_STYLE_LINE_CLASS = 'xterm-cursor-inactive-line',
+  CURSOR_INACTIVE_STYLE_UNDERLINE_CLASS = 'xterm-cursor-inactive-underline',
+  CURSOR_INACTIVE_STYLE_NONE_CLASS = 'xterm-cursor-inactive-none'
```
Tyriar, line 34: We don't need special classes for underline or bar do we, since they should look the same as the regular styles?

Tyriar (changes requested):

addons/xterm-addon-webgl/src/WebglRenderer.ts
```diff
@@ -600,6 +603,22 @@ export class WebglRenderer extends Disposable implements IRenderer {
     const cursorY = this._terminal.buffer.active.cursorY;
     this._onRequestRedraw.fire({ start: cursorY, end: cursorY });
   }
+
+  private _getInactiveCursorStyle(cursorInactiveStyle: 'outline' | 'block' | 'bar' | 'underline' | 'none'): string {
+    if (cursorInactiveStyle === 'outline') {
+      return 'blur';
+    }
```
Tyriar, line 610: We can now get rid of this function if we standardize on 'outline'

Tyriar (approved):
Great PR 👏 I made a few small changes

PR #4659, opened 2023-08-10
Dispose blink state manager

SvanT:
@Tyriar https://github.com/microsoft/vscode/issues/187082

I've verified that terminals with blinking cursor are retained before this patch, and this patch fixes it.

### DIFF
--- a/addons/xterm-addon-webgl/src/WebglRenderer.ts
+++ b/addons/xterm-addon-webgl/src/WebglRenderer.ts
@@ -136,6 +136,7 @@ export class WebglRenderer extends Disposable implements IRenderer {
       for (const l of this._renderLayers) {
         l.dispose();
       }
+      this._cursorBlinkStateManager?.dispose();
       this._canvas.parentElement?.removeChild(this._canvas);
       removeTerminalFromCache(this._terminal);
     }));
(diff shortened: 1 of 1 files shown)


addons/xterm-addon-webgl/src/WebglRenderer.ts
```diff
@@ -136,6 +136,7 @@ export class WebglRenderer extends Disposable implements IRenderer {
       for (const l of this._renderLayers) {
         l.dispose();
       }
+      this._cursorBlinkStateManager?.dispose();
```
Tyriar, line 139: Thanks for spotting this! Instead of disposing here, could you wrap the `new CursorBlinkStateManager` call in `this.register()`? That's preferable so it's lifecycle is set up where it's created

Tyriar (changes requested):

addons/xterm-addon-webgl/src/WebglRenderer.ts
SvanT: Yeah sure, I've updated the PR and verified that this works equally well.

Tyriar (approved):
Thanks! I'll pull it into vscode some time today

Issue #4646, opened 2023-08-06
better inwasm repo integration


jerch:
Issues:
- Not sure yet when exactly this happens - sometimes after restarting the watcher on the demo project the *.wasm.js files get rebuilt by the Typescript compiler, leaving the demo project in the error state `error, must run "inwasm" first`. This can be fixed currently by manually navigating to the image addon and calling `yarn inwasm` there. Needs an automated hook in some of the startup script...
- Once we have more wasm logic in the project (see faster serialize addon), the SDK save paths need adjustments to not pull them again for every inwasm invocation. Maybe place the SDK data globally into the base repo...

Tyriar:
The error is pretty annoying 😅, I actually tried to try/catch the import but it ended up resulting in a bunch of changes as the ctor was then only available as a promise.

jerch:
Hmm I can trigger the recompilation from a branch change sometimes while `yarn watch` is still running. Seems the typescript watcher sometimes decides to refresh a bigger portion of the `*.ts` files than actually needed. I dont think we can reliable hook into that, thus a fix script entry for `*.wasm.js` is needed.

tisilent:
![aq1](https://github.com/xtermjs/xterm.js/assets/48614781/c1b4bcf2-ce55-48d2-a7a5-109fa7a4dab6)
😨

jerch:
Yeah sorry for this, the fix is a bit more involved, trying to get it fixed over the weekend.

tisilent:
Not affected.
yarn add inwasm
yarn inwasm
✌️

jerch:
@tisilent 
Ah this worked "by accident" for you - `yarn add some-package` will re-run the init scripts, which also calls `yarn postsetup`.

Also simply calling `yarn inwasm` is not enough - you must at least give it a file matching pattern. To grab all potential wasm files in addon folders this would be `yarn inwasm "addons/*/out/*.wasm.js"`. But dont do that yet - it will pull SDKs and stuff for no good reason cluttering your repo folder. I need to fix https://github.com/jerch/inwasm/issues/20 first, then I can add a more appropriate script to package.json to get `inwasm` out of the way.

Until then calling `yarn postsetup` is your best bet, as @Tyriar pointed out.

Tyriar:
I got pretty confused when trying to run `yarn test-api`, it gives this error:

```
1) WebLinksAddon
       "before all" hook for ".com":
[...]

PR #4662, opened 2023-08-11
Add functional yarn wasm

Tyriar:
This is what I tried after seeing the error so it will improve the #4646 experience until a real fix is in place

Part of #4646


jerch (approved):
:+1: Yes thats a good interim fix making things a little easier to grasp.

Issue #4408, opened 2023-02-09
Warn or support transparent foreground colors


Tyriar:
It's currently possible to set the theme's foreground to a transparent color which silently breaks rendering: https://github.com/microsoft/vscode/issues/173475

![image](https://user-images.githubusercontent.com/2193314/217834024-4c26ea02-31ce-435c-81f9-6c21e4436c5a.png)

We should either warn when a transparent foreground is used, or blend it with the (opaque) background color and use that.

benmai22:
@Tyriar I want to know if this issue is still available for me to work on it. Thanks and am looking forward to hearing from you.

NwaforAugustine321:
@Tyriar I want to know if this issue is still available for me to work on it. Thanks and am looking forward to hearing from you.

Tyriar:
@NwaforAugustine321 that would be great! Here's a code pointer for where we parse the theme:

https://github.com/xtermjs/xterm.js/blob/854e2736f66ca3e5d3ab5a7b65bf3fd6fba8b707/src/browser/services/ThemeService.ts#L125

Dennnnny:
@Tyriar hello, could I take this over? I would like to try it

Tyriar:
@Dennnnny that would be great! Here's a code pointer: https://github.com/xtermjs/xterm.js/blob/master/src/browser/services/ThemeService.ts#L123

Dennnnny:
@Tyriar thanks, 
I got some idea about this issue, should I just directly create a pr and let you check my idea ?
I think this way works but I want to know does it make sense to you?

Tyriar:
@Dennnnny yep, PRs are easy to iterate on if we need to make changes

PR #4667, opened 2023-08-12
Fix transparent foreground color

Dennnnny:
fix: #4408 

Description: 

I found this transparent foreground color will affect some text when the foreground is background
And  inside the `TextureAtlas`, there are already exist some check for background transparent
So I put another check foreground transparent 

Hope this make scenes

Any feedbacks is welcome, thanks

---

Screenshot: 

For local test, I change the xtermjsTheme foreground color with transparent
```
// demo/client.ts
const xtermjsTheme = {
  foreground: '#F8F8F8CC'
}
```
And the when no doing anything, it looks like:
截圖 2023-08-12 上午12 25 04

After change:
截圖 2023-08-12 上午12 25 59

### DIFF
--- a/src/browser/renderer/shared/TextureAtlas.ts
+++ b/src/browser/renderer/shared/TextureAtlas.ts
@@ -298,7 +298,7 @@ export class TextureAtlas implements ITextureAtlas {
       case Attributes.CM_DEFAULT:
       default:
         if (inverse) {
-          result = this._config.colors.foreground;
+          result = color.opaque(this._config.colors.foreground);
         } else {
           result = this._config.colors.background;
         }
--- a/src/browser/renderer/shared/TextureAtlas.ts
+++ b/src/browser/renderer/shared/TextureAtlas.ts
@@ -277,7 +277,7 @@ export class TextureAtlas implements ITextureAtlas {
   }
 
   private _getBackgroundColor(bgColorMode: number, bgColor: number, inverse: boolean, dim: boolean): IColor {
-    if (this._config.allowTransparency) {
+    if (this._config.allowTransparency || (inverse && !color.isOpaque(this._config.colors.foreground))) {
       // The background color might have some transparency, so we need to render it as fully
       // transparent in the atlas. Otherwise we'd end up drawing the transparent background twice
       // around the anti-aliased edges of the glyph, and it would look too dark.
--- a/src/browser/renderer/shared/TextureAtlas.ts
+++ b/src/browser/renderer/shared/TextureAtlas.ts
@@ -277,7 +277,7 @@ export class TextureAtlas implements ITextureAtlas {
   }
 
   private _getBackgroundColor(bgColorMode: number, bgColor: number, inverse: boolean, dim: boolean): IColor {
-    if (this._config.allowTransparency || (inverse && !color.isOpaque(this._config.colors.foreground))) {
+    if (this._config.allowTransparency) {
       // The background color might have some transparency, so we need to render it as fully
       // transparent in the atlas. Otherwise we'd end up drawing the transparent background twice
       // around the anti-aliased edges of the glyph, and it would look too dark.
--- a/src/browser/services/ThemeService.ts
+++ b/src/browser/services/ThemeService.ts
@@ -122,7 +122,8 @@ export class ThemeService extends Disposable implements IThemeService {
    */
   private _setTheme(theme: ITheme = {}): void {
     const colors = this._colors;
-    colors.foreground = parseColor(theme.foreground, DEFAULT_FOREGROUND);
+    colors.foreground = color.opaque(parseColor(theme.foreground, DEFAULT_FOREGROUND));
+    console.warn("xterm.js is not fully support foreground colors with transparent, so it will the foreground colors alpha channel to 255.")
     colors.background = parseColor(theme.background, DEFAULT_BACKGROUND);
     colors.cursor = parseColor(theme.cursor, DEFAULT_CURSOR);
     colors.cursorAccent = parseColor(theme.cursorAccent, DEFAULT_CURSOR_ACCENT);
--- a/src/browser/services/ThemeService.ts
+++ b/src/browser/services/ThemeService.ts
@@ -122,8 +122,7 @@ export class ThemeService extends Disposable implements IThemeService {
    */
   private _setTheme(theme: ITheme = {}): void {
     const colors = this._colors;
-    colors.foreground = color.opaque(parseColor(theme.foreground, DEFAULT_FOREGROUND));
-    console.warn("xterm.js is not fully support foreground colors with transparent, so it will the foreground colors alpha channel to 255.")
+    colors.foreground = parseColor(theme.foreground, DEFAULT_FOREGROUND);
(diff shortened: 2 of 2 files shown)


Dennnnny:
After thought... 
I think it didn't resolve problem that if people really want to have a transparent foreground
This workaround just avoid people from seeing weird effect, but it didn't really accept transparent.
So, even it fix the weird effect, but ... not sure if it fit your ideal solution 🤔

src/browser/renderer/shared/TextureAtlas.ts
```diff
@@ -277,7 +277,7 @@ export class TextureAtlas implements ITextureAtlas {
   }
 
   private _getBackgroundColor(bgColorMode: number, bgColor: number, inverse: boolean, dim: boolean): IColor {
-    if (this._config.allowTransparency) {
+    if (this._config.allowTransparency || (inverse && !color.isOpaque(this._config.colors.foreground))) {
```
Tyriar, line 280: I was thinking we should instead prevent foreground colors from every being transparent. The `allowTransparency` setting is only the time when these colors should be transparent and it's only about background:

https://github.com/xtermjs/xterm.js/blob/dc390ab03afc338aa57e9849a0c30b45c01b8228/typings/xterm.d.ts#L33-L39

So I was thinking instead, when setting the theme we would force the foreground colors alpha channel to 255 and do a `logService.warn`

Tyriar (changes requested):

src/browser/renderer/shared/TextureAtlas.ts
Dennnnny: @Tyriar thanks for reviewing this, 
I want to tell why I change here:
this issue happened when it is in inverse mode, background colors will use the foreground colors instead.
so don't you think that if we prevent the background color to be setting with transparent is a workaround ? 

if you still think we should prevent foreground colors from every being transparent, that's fine,
I just want to make sure that we are on the same page

thanks again.

Tyriar: Hmm, I guess it depends on whether transparent foreground colors work already. If you set some transparent colors in the demo theme and then run some of the test buttons does it all seem to work correctly?

Dennnnny: yah, I think the foreground colors work already.
this is a screenshot with default foreground is : #F8F8F866, 
截圖 2023-08-13 上午1 00 08

and this is screenshot with default foreground color: #F8F8F8,
截圖 2023-08-13 上午1 03 20

but yah ... I just found that some background color also change in terminal wording... 
for example: colors in colors 17-231 from 256 palette seems to be also transparent a little bit.


截圖 2023-08-13 上午1 09 49
截圖 2023-08-13 上午1 03 52

Not sure 🤔 
so ... what's your opinion about this situation ?
I can just work on your suggestion earlier: force foreground to not be transparent and add log

Tyriar: Since transparent fg works, would it work if we forced the foreground color to be opaque when used as inverse background at line 301?

Tyriar: I think we have a toOpaque or similar helper

jerch:
Have not looked at the PR code, just FYI:
We have no good idea, what transparency on FG should mean or be used for in a terminal. I tried to address this several times with other terminal devs, esp. because SGR has a dedicated transparent flag (which we dont support yet), but the response on that was more like "Idk, have not yet dealt in depth with it".

So until we resolve the issue, what transparency in FG can be used for and find a real good use case, I like @Tyriar 's idea to completely strip any transparency from FG from FG/BG colors swaps, maybe with an additional lumi correction to stay on sane ground.

src/browser/renderer/shared/TextureAtlas.ts
Dennnnny: thanks for advices, I'll try again.

Dennnnny: @Tyriar  sorry for bothering again. 
now I'm trying to use the first suggestion you provided: **force fg colors alpha channel to 255**
but I think I met some issue here with test cases.

so...
when convert fg color alpha to 255 with `color.opaque` 
it seems to transform lowercase: "00FF00" -> "00ff00"
so it fail in test

so I'm thinking ... should I just directly adjust function in `color.opaque` to make it uppercase?
and, any advice if I put console.warn directly in ThemeService ?? ( I'm not sure where should I put this log 😕 

thanks

Tyriar: Sorry for all the back and forth, but since foreground seems to work based on your testing can we instead change this line:

https://github.com/xtermjs/xterm.js/blob/70f6fa2bc91e27500900d8633d7f1279feb26e24/src/browser/renderer/shared/TextureAtlas.ts#L301

To this?

```ts
result = color.opaque(this._config.colors.foreground);
```

I'm worried about regressing when users rely on this now 😅

Dennnnny: oh, I tried that solution, it actually can fix that origin issue too.
but right now it will cause some background color become transparent too.
for example : Colors in colors 17-231 from 256 palette
like attach 
截圖 2023-08-13 上午1 09 49
we don't want to these example to be transparent, right ??

so... okay!! I'll find out how to fix this too!!
then after everything is fixed, I'll ask for re-request review

it's my bad to do all the back and forth, 
I'm afraid to mess this up, so sorry 😟 
and thanks for every suggestions.

Dennnnny: so sorry for all these back and forth, I just don't want to mess this up.
And I think this one works as well.
I'll change to this solution

thanks.

by the way, do you want me to squash commits before merge ?

src/browser/services/ThemeService.ts
```diff
@@ -122,7 +122,8 @@ export class ThemeService extends Disposable implements IThemeService {
    */
   private _setTheme(theme: ITheme = {}): void {
     const colors = this._colors;
-    colors.foreground = parseColor(theme.foreground, DEFAULT_FOREGROUND);
+    colors.foreground = color.opaque(parseColor(theme.foreground, DEFAULT_FOREGROUND));
+    console.warn("xterm.js is not fully support foreground colors with transparent, so it will the foreground colors alpha channel to 255.")
```
Tyriar, line 126: Oh I meant having this at TextureAtlas.ts by changing this line:

https://github.com/xtermjs/xterm.js/blob/39ae0f45fdee2195382e36db0f32cbeb7c747679/src/browser/renderer/shared/TextureAtlas.ts#L302

To:

```ts
result = color.opaque(this._config.colors.foreground);
```

If that works.

Tyriar (changes requested):

src/browser/services/ThemeService.ts
Dennnnny: oh, sorry for the misunderstood.
yah this one should work as well, 
it will stay the foreground color to be transparent and if is inverse mode, it will make it opaque as background color.

thanks

Tyriar (approved):
Thanks for investigating this 👍

Dennnnny:
thanks for all the help and suggestions. 
it's my honor ❤️

PR #4668, opened 2023-08-12
Restrict comments to 100 characters

Tyriar:
This is a rule I try enforce in PR but it's a bit of a hassle, this will give a lint warning when it happens resulting in:

- Yellow underline in supported editors when eslint is installed
- Compilation will succeed since it's a warning not an error
- CI will fail as it does not allow lint warnings

Special case comments are ignored which includes @vt comments, table comments and commented out code.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -159,6 +159,16 @@
     "jsdoc/check-param-names": 1,
     "jsdoc/no-multi-asterisks": 1,
     "keyword-spacing": "warn",
+    "max-len": [
+      "warn",
+      {
+        "code": 1000, // Don't enforce for code
+        "comments": 100,
+        "ignoreTrailingComments": true,
+        "ignoreUrls": true,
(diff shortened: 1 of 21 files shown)


jerch:
Hmm yeah the `@vt` docs clutter the source files to some degree, this is quite obvious when scrolling through inputhandler. Still I find the idea to have that close to source very important, otherwise docs tend to fall behind all the time (also happens to doc strings, but imho to a lesser degree than for completely separate docs).
At least for me the issue is not a biggy, as the display is big enough to still give the right amount of overview on the source, even with those nasty huge markup tables. I also dont know of any good way to shorten those.
Well, at least the `@vt` could be reshaped to store their info vertically instead of horizontally, but I am unsure, if that make anything better?

Tyriar:
I like the `@vt` comments, no problems having them and the other exceptions be longer. I just want to automate some consistency so I don't have to call it out in PRs as much

PR #4673, opened 2023-08-13
DOM: Render selection color instead of cell bg

### LINKED ISSUE #4097: DOM renderer does not show selection over regular background colors
The yellow text background should be blue here:

image

Tyriar:
Summary:

- Selection cell reuse has been added to the DOM renderer
- Selection bg override is now correctly set correctly
- Selection elements 'pretend' to be a decoration in order to render bg

Fixes #4097

image
image

### DIFF
--- a/src/browser/renderer/dom/DomRendererRowFactory.ts
+++ b/src/browser/renderer/dom/DomRendererRowFactory.ts
@@ -89,6 +89,7 @@ export class DomRendererRowFactory {
     let oldExt = 0;
     let oldLinkHover: number | boolean = false;
     let oldSpacing = 0;
+    let oldIsInSelection: boolean = false;
     let spacing = 0;
     const classes: string[] = [];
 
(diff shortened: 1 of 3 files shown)


src/browser/renderer/dom/DomRendererRowFactory.ts
```diff
@@ -154,16 +155,24 @@ export class DomRendererRowFactory {
...
-          && cell.bg === oldBg && cell.fg === oldFg && cell.extended.ext === oldExt
-          && !isInSelection
+          && (
+            (isInSelection && oldIsInSelection)
+            || (!isInSelection && cell.bg === oldBg)
+          )
+          && (
+            (isInSelection && oldIsInSelection && colors.selectionForeground)
+            || (!(isInSelection && oldIsInSelection && colors.selectionForeground) && cell.fg === oldFg)
+          )
```
jerch, line 174: Hmm this looks quite complicated - I guess you have tested it, so I assume it good as it is (though I wonder if it could be simplified). Which selection state tried you to cover here, which was not by the old conditions?

src/browser/renderer/dom/DomRendererRowFactory.ts
```diff
@@ -417,7 +431,7 @@ export class DomRendererRowFactory {
       }
 
       // exclude conditions for cell merging - never merge these
-      if (!isCursorCell && !isInSelection && !isJoined && !isDecorated) {
+      if (!isCursorCell && !isJoined && !isDecorated && isInSelection === oldIsInSelection) {
```
jerch, line 434: Not sure, if  `isInSelection === oldIsInSelection`  makes much sense here - conditions here are late evals for the first span, and whether it qualifies for merging with follow-up chars.

I lazily hacked `!isInSelection` in here, which is technically too strict. Normally selections can also merge, if the edge conditions are all met, which I was not sure about, thus the shortcut.

Now your `isInSelection === oldIsInSelection` matches the same value with itself, because of line:206?

Tyriar:
Merging so I can get some testing in VS Code, @jerch let me know if something looks off

jerch (changes requested):
Oh well, that looks complicated. Please see my comments below.

src/browser/renderer/dom/DomRendererRowFactory.ts
Tyriar: Before we would just never merge cells if it was a selection which is easy but that means select all would all be the old slow rendering.

Outside of pulling things into named variables or expanding the comment I'm not sure how to simplify this, these cases are described in the comment above:

```
- bg did not change (or both are in selection)
- fg did not change (or both are in selection and selection fg is set)
```

src/browser/renderer/dom/DomRendererRowFactory.ts
Tyriar: Oh you're right this does nothing 🙈 

We may just be able to remove that condition completely then as this seemed to work well.

Tyriar: https://github.com/xtermjs/xterm.js/pull/4676

PR #4678, opened 2023-08-15
Workaround crash that can happen in Electron/Chromium

Tyriar:
Part of microsoft/vscode#189753

cc @deepak1556

### DIFF
--- a/src/browser/renderer/dom/StyleSheet.ts
+++ b/src/browser/renderer/dom/StyleSheet.ts
@@ -3,6 +3,8 @@
  * @license MIT
  */
 
+import { isElectron } from 'common/Platform';
+
 export interface IStyleSheet {
   dispose: () => void;
   setCss: (value: string) => void;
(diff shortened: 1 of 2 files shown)


jerch:
Eww, is this style element handling browser-alpha ground? Tbh I dont like where this is going - should we revert the style stuff to a more proven code pattern?

Tyriar:
We can now actually reproduce this in Chromium, so I'm also thinking we should just revert the whole CSP change as it doesn't seem like it's ready for primetime. FYI @SimonSiefke

SimonSiefke:
Sorry for the breakage caused by my changes. I also agree that reverting the changes is the best option!

> On a sidenote - not even sure how recent engines need to be to support this - did we just raise the version requirements to newest browsers only by adding by https://github.com/xtermjs/xterm.js/pull/4611?

The change uses try/catch and falls back to `<style>` elements, It seems to me it should be backwards compatible  - apart from crashing chromium ;)

PR #4681, opened 2023-08-16
Fix bug in selection handling in dom renderer

PerBothner:
If you use the dom renderer and make a selection, the visible selection extends to the end of the line.  This fixes that.

This also fixes a logic weirdo: `A || (!A && B)` should be simplified to `A || B`.

### DIFF
--- a/src/browser/renderer/dom/DomRendererRowFactory.ts
+++ b/src/browser/renderer/dom/DomRendererRowFactory.ts
@@ -166,11 +166,11 @@ export class DomRendererRowFactory {
           cellAmount
           && (
             (isInSelection && oldIsInSelection)
-            || (!isInSelection && cell.bg === oldBg)
+            || (!isInSelection && !oldIsInSelection && cell.bg === oldBg)
           )
           && (
             (isInSelection && oldIsInSelection && colors.selectionForeground)
-            || (!(isInSelection && oldIsInSelection && colors.selectionForeground) && cell.fg === oldFg)
+            || cell.fg === oldFg
           )
           && cell.extended.ext === oldExt
           && isLinkHover === oldLinkHover


src/browser/renderer/dom/DomRendererRowFactory.ts
```diff
@@ -166,11 +166,11 @@ export class DomRendererRowFactory {
...
           && (
             (isInSelection && oldIsInSelection)
-            || (!isInSelection && cell.bg === oldBg)
+            || (!isInSelection && !oldIsInSelection && cell.bg === oldBg)
           )
           && (
             (isInSelection && oldIsInSelection && colors.selectionForeground)
-            || (!(isInSelection && oldIsInSelection && colors.selectionForeground) && cell.fg === oldFg)
+            || cell.fg === oldFg
           )
```
jerch, line 174: Suggestion - for the sake of readablility lets separate concerns here:
```typescript
  // selection state must not change
  && isInSelection === oldIsInSelection
  // outside of selections BG must match
  && (!isInSelection && cell.bg === old.Bg)
  // outside of selections FG must match
  // inside selection a defined selectionForeground will override FG
  && ((!isInSelection && cell.fg === old.Fg) || colors.selectionForeground)
```

To me thats much easier to grasp than nested conditions, hopefully I covered all cases :smile_cat:.

jerch: Hmm the FG rule still could be better:
```typescript
  && (cell.fg === old.Fg || (isInSelection && colors.selectionForeground))
```
Difference to above - also in selection the fg should match by default, `selectionForeground` is the exception.

PerBothner: That re-write seems to work fine my testing. And I'm inclined to agree it is cleaner. I was aiming for a minimal fix, but I'm happy also doing minor logic simplification.

src/browser/renderer/dom/DomRendererRowFactory.ts
```diff
@@ -164,14 +164,13 @@ export class DomRendererRowFactory {
...
-            (isInSelection && oldIsInSelection && colors.selectionForeground)
-            || (!(isInSelection && oldIsInSelection && colors.selectionForeground) && cell.fg === oldFg)
-          )
+          // selection state must not change
+          && isInSelection === oldIsInSelection
+          // outside of selections BG must match
+          && (!isInSelection && cell.bg === oldBg)
+          // default FG must match
+          // exception: in selections selectionForeground may override FG
+          && (cell.fg === oldFg || (isInSelection && colors.selectionForeground))
```
Tyriar, line 173: Tests are failing pointing out less optimal results, can we handle this case too?

```
  1) DomRendererRowFactory
       createRow
         selectionForeground
           should force whitespace cells to be rendered above the background:

      AssertionError: expected '<span style="background-color:#ff0000…' to equal '<span style="background-color:#ff0000…'
      + expected - actual

      -<span style="background-color:#ff0000;" class="xterm-decoration-top"> </span><span style="background-color:#ff0000;" class="xterm-decoration-top">a</span>
      +<span style="background-color:#ff0000;" class="xterm-decoration-top"> a</span>
      
      at Context.<anonymous> (src/browser/renderer/dom/DomRendererRowFactory.test.ts:319:16)
      at processImmediate (node:internal/timers:476:21)
```

jerch: It break out of the merge here because FG is not set on the empty cell. I think this should fix it:
```typescript
&& ((chars !== ' ' && cell.fg === oldFg) || (isInSelection && colors.selectionForeground))
```
This adds another merge chance - if there is no content is a cell, it can be merged regardless of FG.

jerch: Looking at it again, I think the simplifications are all slightly off because of the way we store SGR flags on bg and fg. In fact we'd need color inspection for the selection tests here. The difference in certain flags should still break out of the merge.

PerBothner: Maybe for now just go with my original minimal fix?

PerBothner: FWIW In my `buffer-cell-cursor` branch I added to IAttributeData new accessor functions getFg and getBg which return combined getFgColor+getFgColorMode and getBgColor+getBgColorMode as a 26-bit int. I think that is a cleaner way of dealing with colors, with a new getStyleFlags function for the SGR flags and more (up to 28 bits). Maybe we can revisit further optimizations/cleanups later.

Tyriar: The initial commit does look safer, fixes problems and passes the tests so that's cool with me.

@jerch I guess we need to consider FgFlags.INVERSE for the background check and BgFlags.DIM for the foreground check? Not sure if extended underlines are handled. We need some failing tests to fix against.

jerch: > The initial commit does look safer, fixes problems and passes the tests so that's cool with me.

Im pretty sure that INVERSE gets swallowed atm, it just that we dont have a test for atm (not tested thoughm just guessing).

Tyriar: Inverse when there's no selectionForeground won't get covered 🤯

jerch: > I guess we need to consider FgFlags.INVERSE for the background check and BgFlags.DIM for the foreground check? Not sure if extended underlines are handled. We need some failing tests to fix against.

Yeah I think we should cleanup the flag association a bit and put any font altering flag into fg.

Tyriar: https://github.com/xtermjs/xterm.js/issues/4682

jerch: Can do the nasty flag rework (that affects quite some code places), but it has to wait, as I am still on Per's PR (only halfway through so far :smiley_cat:).

PerBothner:
Not sure the best way to undo the last commit. Can I remove the last commit from the PR?  Should I use a `git revert` - which will make a needless-complicted history - or can you clean that up on merge?

jerch:
@PerBothner 
If you want a clean repo log afterwards, imho revert + a force push should do.

Tyriar:
Since we've decided what to do you can go either way. I don't normally care about many commits going in PRs as you can always revert by merge commit in the off chance we need to.

PerBothner:
@jerch "imho revert + a force push should do".

Did you mean `git reset` + `git push --force`? I did that, and it seemed to work. However, the tests reported failures, which I'm looking into.

jerch:
Well its either:
```bash
git revert <commit to revert>
git push --force
```
or (`HEAD~1` assumes to undo last commit):
```bash
git reset --soft HEAD~1
git push --force
```

PerBothner:
I think it's ready for merge now.

Tyriar (approved):

PR #4684, opened 2023-08-18
Continuous underline dotted

### LINKED ISSUE #4349: Take column into account when rendering cells for perfect patterned underline rendering
Forking off from https://github.com/xtermjs/xterm.js/issues/4060

> Thinking about this and the fact that I want xterm.js rendering to be as great as possible, I think we should support multiple glyphs depending on the x-axis when necessary. For the dotted line specifically, say we had a dot of size `window.devicePixelRatio` with the same sized gap in-between, so there would be at most 2 possible variants for every glyph:
> 
> ```
> Even:
> |* * * * |
> 
> Odd:
> |* * * * *|
> | * * * * |
> ```
> 
> To store this as a unique glyph we need to make the glyph key different, this could be done by asking what underline variant it is and then store it as a number, maybe capped at 4 (2 bits) or 8 variants (3 bits)?
> 
> This would allow us make a non-repeated curvy line and not restrict us to a single cell for both the up and down parts of the wave, if that ends up looking better of course.

Current:

image
image

Desired:

image
image

tisilent:
fix #4349 
trial version
Font size above 26 is unstable.
Only processed webgl.
@Tyriar Is this idea feasible?

### DIFF
--- a/addons/xterm-addon-webgl/src/WebglRenderer.ts
+++ b/addons/xterm-addon-webgl/src/WebglRenderer.ts
@@ -17,7 +17,7 @@ import { CellData } from 'common/buffer/CellData';
 import { Attributes, Content, ExtFlags, NULL_CELL_CHAR, NULL_CELL_CODE, UnderlineStyle } from 'common/buffer/Constants';
 import { EventEmitter, forwardEvent } from 'common/EventEmitter';
 import { Disposable, getDisposeArrayDisposable, toDisposable } from 'common/Lifecycle';
-import { ICoreService, IDecorationService, IOptionsService, IUnicodeService } from 'common/services/Services';
+import { ICoreService, IDecorationService, ILogService, IOptionsService, IUnicodeService } from 'common/services/Services';
 import { CharData, IBufferLine, ICellData } from 'common/Types';
 import { IDisposable, Terminal } from 'xterm';
 import { GlyphRenderer } from './GlyphRenderer';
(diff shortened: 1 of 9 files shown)


tisilent:
emmm, file format ...... 😭

tisilent:
企业微信截图_16923578691709
企业微信截图_16923578691709
This should be a reasonable situation.
size 21  ,   22

addons/xterm-addon-webgl/src/WebglRenderer.ts
```diff
@@ -481,13 +500,28 @@ export class WebglRenderer extends Disposable implements IRenderer {
...
-            this._model.cells[i + RENDER_MODEL_FG_OFFSET] === this._cellColorResolver.result.fg &&
-            this._model.cells[i + RENDER_MODEL_EXT_OFFSET] === this._cellColorResolver.result.ext) {
+          this._model.cells[i + RENDER_MODEL_BG_OFFSET] === this._cellColorResolver.result.bg &&
+          this._model.cells[i + RENDER_MODEL_FG_OFFSET] === this._cellColorResolver.result.fg &&
+          this._model.cells[i + RENDER_MODEL_EXT_OFFSET] === this._cellColorResolver.result.ext &&
+          this._model.cells[i + RENDER_MODEL_VARIANT_OFFSET] === variantOffset
+        ) {
+          if (cell.extended.underlineStyle === UnderlineStyle.DOTTED) {
+            const s = this._charSizeService.width;
+            variantOffset = ((deviceCellWidth! * chWidth) - variantOffset) % (lineWidth * 2);
```
Tyriar, line 519: For dotted, I would expect there only to be _up to_ `2 * dot_size` variants. So when the dot size is 1px, these would be the 2 variants for `s`:

image

Tyriar (changes requested):

addons/xterm-addon-webgl/src/WebglRenderer.ts
tisilent: Yes, I also think so. The formula for calculating offset has not been written yet.

addons/xterm-addon-webgl/src/WebglRenderer.ts
```diff
@@ -481,13 +500,28 @@ export class WebglRenderer extends Disposable implements IRenderer {
...
 
         // Nothing has changed, no updates needed
         if (this._model.cells[i] === code &&
-            this._model.cells[i + RENDER_MODEL_BG_OFFSET] === this._cellColorResolver.result.bg &&
-            this._model.cells[i + RENDER_MODEL_FG_OFFSET] === this._cellColorResolver.result.fg &&
-            this._model.cells[i + RENDER_MODEL_EXT_OFFSET] === this._cellColorResolver.result.ext) {
+          this._model.cells[i + RENDER_MODEL_BG_OFFSET] === this._cellColorResolver.result.bg &&
+          this._model.cells[i + RENDER_MODEL_FG_OFFSET] === this._cellColorResolver.result.fg &&
+          this._model.cells[i + RENDER_MODEL_EXT_OFFSET] === this._cellColorResolver.result.ext &&
+          this._model.cells[i + RENDER_MODEL_VARIANT_OFFSET] === variantOffset
```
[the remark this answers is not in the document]
tisilent: 😨   I've sorted out the code and it's a bit messy. I've already figured out how to solve 2px+.

tisilent:
??...... why.

tisilent:
close,Resubmit. :sob:.

Issue #3297, opened 2021-04-06
Migrate to GitHub Actions


Tyriar:
GH Actions is fairly mature now, it would simplify our process if we moved to that for CI/CD. The biggest challenge here is migrating the publish action which would need to be done by me, but the others could be brought over by community.

PR #4693, opened 2023-08-19
GitHub Actions

Tyriar:
Fixes https://github.com/xtermjs/xterm.js/issues/3297

---

Typical `master` build on Azure Pipelines ~15 minutes:

Screenshot 2023-08-19 at 12 57 51 pm


jerch:
@Tyriar  Wanna move away from azure CI? Tbh I found the gh-action based CI more reliable (from other projects), than azure in xtermjs. I kinda never get timeout errors on gh-actions, while azure seems to cause testing issue from high load sometimes.

Tyriar:
I do, mainly because it's a little more integrated and easier to get to the failures. It's looking quite nice so far, will see how publish goes

.github/workflows/ci.yml
```diff
@@ -0,0 +1,66 @@
...
+    - uses: actions/checkout@v3
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v3
+      with:
+        node-version: ${{ matrix.node-version }}
+        cache: 'yarn'
+    - name: Install dependencies and build
+      run: yarn --frozen-lockfile
+    - name: Integration tests (Chromium)
+      run: yarn test-api-chromium --headless --forbid-only
```
jerch, line 48: Would merging the browser tests into one container defintion help to speedup things?

I also wonder, if we should add back safari/webkit testing at some point, though I dont recall, why we removed it in the first place. We kinda have a blind there atm.

Tyriar: If agents all get provisioned then as much parallelization as possible would be the fastest. The integration tests for each browser takes several minutes.

I think for webkit there were issues when it was originally set up and never looked into after that.

PR #4700, opened 2023-08-19
Make GH actions strategy more concise

Tyriar:


Tyriar:
@jerch think I'm done with the CI changes. It's much faster now and I also reduced the number of required checks to speed up merging. If we happen to get failing tests merged in we should notice pretty quickly.

jerch:
@Tyriar Is that down from ~14m to ~8m? Wow the unit tests complete faster in gh-actions (26s) than on my machine (28s) :heart_eyes: Great work!

Tyriar:
It'll vary but the slowest job we wait for now are unit tests on Windows, so this one took about 8 minutes to finish required checks, 11 minutes to finish all checks. Then when it hits master, we release immediately [so this took 3 minutes](https://github.com/xtermjs/xterm.js/actions/runs/5913890236) to release as opposed to the ~15

Tyriar:
Oh also, we now must run `yarn setup` after cloning the repo. I couldn't figure out how to split it without removing the `prepare` script

PR #4701, opened 2023-08-20
Migrate to esbuild

Tyriar:
This is a work in progress, here are the goals of the change:

- Reduce build/watch time
- Reduce bundle size
- Simplify build by removing webpack and having dev builds run the same way
- Simplify sourcemaps
- Self host with the demo in a way that's closer to real world usage by importing the modules

Currently building everything from scratch takes less than a second(!), compared to ~16 seconds for tsc on my machine:

image

Current issues:

- [ ] There are some complications particularly around inwasm, the image addon build doesn't work yet.
- [ ] Not sure how to handle tests yet, for now we can continue to build them to out/ and out-test/ and run them the old way as we need to do a full tsc build anyway to check types.
- [ ] VS Code task order could be improved
- [ ] The demo-client build can fail when run currently as it actually depends on xterm*


Tyriar:
Looks like this sort of thing will work for tests:

```
.\node_modules\.bin\esbuild src/**/*.ts --outdir=out-esbuild --format=cjs
```

jerch:
> There are some complications particularly around inwasm, the image addon build doesn't work yet.

Does this use ESM output for addon sources? Because thats not fully supported yet with `inwasm`.

Tyriar:
It does yeah as that's something else people want (and VS Code supports now). I think we could special case the image addon in the meantime

Tyriar:
Putting this on hold as it was a weekend project that didn't make it in time 🙂

PR #4703, opened 2023-08-21
Use variants for underline dotted.

### LINKED ISSUE #4349: Take column into account when rendering cells for perfect patterned underline rendering

tisilent:
fix #4349


tisilent:
hi @Tyriar , The effect should be better this time.

src/common/MultiKeyMap.ts
```diff
@@ -40,3 +40,22 @@ export class FourKeyMap<TFirst extends string | number, TSecond extends string |
     this._data.clear();
   }
 }
+
+export class FiveKeyMap<TFirst extends string | number, TSecond extends string | number, TThird extends string | number, TFourth extends string | number, TFive extends string | number, TValue> {
```
Tyriar, line 44: Think this can be removed

Tyriar (changes requested):
Thanks! I'll try get to a proper review in a week or 2

tisilent:
I need to clean up the code.
I want to fix the dashed line as well.

tisilent:
Thanks.

addons/xterm-addon-canvas/src/TextRenderLayer.ts
```diff
@@ -72,13 +73,20 @@ export class TextRenderLayer extends BaseRenderLayer {
...
-      y: number
+      y: number,
+      variantOffset: number
     ) => void
   ): void {
+    const fontSize = this._optionsService.rawOptions.fontSize;
+    const drp = this._coreBrowserService.dpr;
+    const lineWidth = Math.max(1, Math.floor(fontSize * drp / 15));
+    const deviceCellWidth = this._charAtlas?.getDeviceCellWidth();
+    let variantOffset: number = -1;
```
Tyriar, line 84: Can all this logic live inside CellColorResolver?

addons/xterm-addon-webgl/src/WebglRenderer.ts
```diff
@@ -408,11 +410,18 @@ export class WebglRenderer extends Disposable implements IRenderer {
     this._model.cursor = undefined;
     let modelUpdated = false;
 
+    const fontSize = this._optionsService.rawOptions.fontSize;
+    const drp = this._coreBrowserService.dpr;
+    const lineWidth = Math.max(1, Math.floor(fontSize * drp / 15));
+    const deviceCellWidth = this._charAtlas?.getDeviceCellWidth();
```
Tyriar, line 416: Same; can this be shared inside CellColorResolver?

addons/xterm-addon-webgl/src/WebglRenderer.ts
```diff
@@ -408,11 +410,18 @@ export class WebglRenderer extends Disposable implements IRenderer {
...
+
```
Tyriar, line 417: You should be able to pull device cell width off `this.dimensions``

tisilent:
ok.I have discovered an issue now. I will solve it first

src/browser/renderer/shared/CellColorResolver.ts
```diff
@@ -133,5 +147,27 @@ export class CellColorResolver {
...
+        const fontSize = this._terminal.options.fontSize;
+        const drp = this._coreBrowserService.dpr;
+        const lineWidth = Math.max(1, Math.floor(fontSize! * drp / 15));
+        let chWidth: number;
+        if (typeof code === 'number') {
+          chWidth = this._unicodeService.wcwidth(code);
+        } else {
+          chWidth = this._unicodeService.getStringCellWidth(code);
+        }
+        $variantOffset = computeVarinatOffset(deviceCellWidth * chWidth, lineWidth, $variantOffset);
```
Tyriar, line 167: I was expecting the variant to be based on the left-most pixel of the glyph, so `chWidth` wouldn't be needed?

src/browser/renderer/shared/CellColorResolver.ts
```diff
@@ -133,5 +147,27 @@ export class CellColorResolver {
...
+    // Reset overrides variantOffset
+    this.result.ext &= ~ExtFlags.VARIANT_OFFSET;
+    this.result.ext |= ($variantOffset << 29) & ExtFlags.VARIANT_OFFSET;
+
+    // Compute next varinatOffset
+    if (cell.extended.underlineStyle === UnderlineStyle.DOTTED) {
+      if (code !== NULL_CELL_CODE) {
+        const fontSize = this._terminal.options.fontSize;
+        const drp = this._coreBrowserService.dpr;
+        const lineWidth = Math.max(1, Math.floor(fontSize! * drp / 15));
```
Tyriar, line 160: If you pass in `IOptionsService` you can use `rawOptions.fontSize` and avoid the unsafe `!`

Tyriar, line 160:
```suggestion
        const lineWidth = Math.max(1, Math.floor(fontSize! * this._coreBrowserService.dpr / 15));
```

src/browser/renderer/shared/CellColorResolver.ts
```diff
@@ -133,5 +147,27 @@ export class CellColorResolver {
     // Use the override if it exists
     this.result.bg = $hasBg ? $bg : this.result.bg;
     this.result.fg = $hasFg ? $fg : this.result.fg;
+
+    // Reset overrides variantOffset
+    this.result.ext &= ~ExtFlags.VARIANT_OFFSET;
+    this.result.ext |= ($variantOffset << 29) & ExtFlags.VARIANT_OFFSET;
+
+    // Compute next varinatOffset
```
Tyriar, line 155:
```suggestion
    // Compute next variantOffset
```

src/browser/renderer/shared/TextureAtlas.ts
```diff
@@ -596,9 +597,28 @@ export class TextureAtlas implements ITextureAtlas {
...
-            this._tmpCtx.setLineDash([Math.round(lineWidth), Math.round(lineWidth)]);
-            this._tmpCtx.moveTo(xChLeft, yTop);
-            this._tmpCtx.lineTo(xChRight, yTop);
+            const offsetWidth = nextOffset === 0 ? 0 :
+              (nextOffset >= lineWidth ? lineWidth * 2 - nextOffset : lineWidth - nextOffset);
+            if (offsetWidth === 0) {
+              this._tmpCtx.setLineDash([Math.round(lineWidth), Math.round(lineWidth)]);
+              this._tmpCtx.moveTo(xChLeft, yTop);
+              this._tmpCtx.lineTo(xChRight, yTop);
+            } else {
```
Tyriar, line 606: I don't think you need a special case for this as the main effect is `+ offsetWidth` = `+ 0`?

src/common/buffer/AttributeData.ts
```diff
@@ -174,6 +177,18 @@ export class ExtendedAttrs implements IExtendedAttrs {
     this._urlId = value;
   }
 
+  public get underlineVarinatOffset(): number {
+    const val = (this._ext & ExtFlags.VARIANT_OFFSET) >> 29;
+    if (val < 0) {
+      return val ^ 4294967288;
```
Tyriar, line 183: Hex would be easier to read here

Tyriar (changes requested):

src/browser/renderer/shared/CellColorResolver.ts
tisilent: I need to take a look at the process here.....

tisilent: I see. 😄

Tyriar (approved):
This works really well! 👏 

image

image

image

tisilent:
Next. Curve.

Tyriar:
@tisilent that would be great, I found it a bit tricky to get the curve to look good at various zoom levels.

PR #4718, opened 2023-08-23
Add no unused variables lint warning

Tyriar:
This shows a warning while compiling, so demo bundling will still happen but it will be rejected when a PR is made. Imports are automatically cleaned up when organizing imports in VS Code (alt+shift+o).

This also ended up removing some runtime variables that weren't used.


Tyriar:
This doesn't play nice with doc references:

image

PR #4735, opened 2023-08-27
Start @playwright/test migration

Tyriar:
This is the successor to https://github.com/xtermjs/xterm.js/pull/3460, the plan is to add proper support for these types of tests and have both types in the codebase simultaneously. Instead of trying to migrate all unit tests at once, we can do it parts at a time and any new tests can be added using the new method.

This is a WIP and will only get merged when the integration feels solid.

Part of https://github.com/xtermjs/xterm.js/issues/3446
Part of https://github.com/xtermjs/xterm.js/issues/3326


Tyriar:
Up to 50% savings from the new integration test jobs 🎉 

image

Tyriar:
Actually the gain is a little nuanced as the addon tests are only done in the old tests now.

Here's an example of without the PR:

image

Tyriar:
I didn't migrate all the tests to use the proxy as it doesn't need to happen to make the change. The addon tests also have not been brought over yet.

PR #4742, opened 2023-09-01
Allow selection in the a11y tree and sync the selection to terminal

### LINKED ISSUE #4656: Unable to select text with screen reader
## Details
- OS version:  ChromeOS (but this is probably not limit to ChromeOS)
- xterm.js version: ToT

The `.xterm-accessibility-tree` element (and everything else within `.xterm`) has the css variable `user-select` set to `none`, so ChromeVox (and probably the screenreader on other platforms) is only able to select the row element but not the text inside it. The user is also not able to copy the text because of this.

I think the solution to this is:

1. override `user-select` for `.xterm-accessibility-tree` so that the text is selectable. We will also use css to hide the browser selection highlight.
2. Monitor the `selectionchange` event for the element, and sync the selection to the terminal (e.g. via `term.select()`)

What do you think about this?

JasonXJ:
Fixes #4656


JasonXJ:
Hi @Tyriar, can you take a look at this. A few notes:

* I put a `const debug` in `AccessibilityManager.ts` so that I can easily use the mouse to interact with the tree during development. I think it is quite useful so I leave it in the code. I am also happy to remove it.
* The code in `AccessibilityManager._handleSelectionChange()` to "translate" the `Selection` object to a row-column pair is quite complicated, so I do want to write some tests there. But there isn't existing tests for the manager, and setting everything up for testing is a lot of work. So I am thinking about extracting some of the code to static methods and do unit tests on the static methods instead. What do you think about this?

JasonXJ:
Hi @Tyriar, can you take a look at this? Or can you let me know how I can help to get this merge quicker?

Tyriar (approved):
Seems to work 👍 

For tests, yes it is a little tricky to get them set up for this as it's quite entangled with the main `Terminal` object (cyclic dependency). If you want to do a follow up for tests which would be greatly appreciated, you could just turn this into an exported function and test that, instead of trying to test the method on `AccessibilityManager`.

Tyriar:
FYI @meganrogge, this should allow screen readers to select text in the underlying textarea. I haven't tested it much so we'll want to keep an eye out in Insiders.

src/common/buffer/BufferLine.ts
```diff
@@ -508,16 +508,43 @@ export class BufferLine implements IBufferLine {
...
       const cp = content & Content.CODEPOINT_MASK;
-      result += (content & Content.IS_COMBINED_MASK) ? this._combined[startCol] : (cp) ? stringFromCodePoint(cp) : WHITESPACE_CELL_CHAR;
-      startCol += (content >> Content.WIDTH_SHIFT) || 1; // always advance by 1
+      const chars = (content & Content.IS_COMBINED_MASK) ? this._combined[startCol] : (cp) ? stringFromCodePoint(cp) : WHITESPACE_CELL_CHAR;
+      result += chars;
+      if (outColumns) {
+        for (let i = 0; i < chars.length; ++i) {
+          outColumns.push(startCol);
+        }
+      }
```
jerch, line 543: Just stumbled over this code change and blamed it back to here.

This looks like a serious perf smell to me - note that this function is hot-path for handlers grabbing line contents over and over.
Ideally hot-path functions are as branching/inner-loop free as possible, and never change the inner works/touched variables to not get de-optimized from JIT. This change adds all of that.

Note that I have not perf tested it yet, so its just a "looks-alike" concern from my side atm.

JasonXJ: I played with it a bit and this did not seem to be called much during normal rendering. It is called quite a lot by the SearchAddon (when you are using search) though, but it is unclear to whether it actually affect performance.

jerch: Yeah that method is not called during normal data ingestion & rendering (basic operations). But, as you already found, it gets used by several 2nd-line operations, which do buffer scraping alot, like search, serialize, in general things operating on the buffer API. While those actions are rarely triggered itself, they most likely run over bigger bufferlines batches creating some runtime pressure for a short amount of time (potentially leading to hiccups).

Will see if I find the time to perf it - tricky part here prolly is, that the de-opt might only show up, if calls alternate between `outColumns` being set and unset. Not sure yet.

PR #4754, opened 2023-09-07
When DomRenderer focus,render the cursor line.

### LINKED ISSUE #4748: DOM renderer does not re-render cursor on focus
Repro:

1. Set DOM renderer
2. Unfocus
3. Focus, 🐛 unfocus cursor is still rendered
4. Type, it's fixed

Image

tisilent:
fix #4748.


Tyriar (approved):
Makes sense, the test buttons are a nice touch too as this was harder to do in the demo than it should have been.

PR #4762, opened 2023-09-08
Add handling of invisible

### LINKED ISSUE #4758: Invisible text is visible in the DOM renderer
Image

tisilent:
fix #4758


Tyriar (approved):
Added a regression test (in a new sub-suite) after figuring out the exact problem case

PR #4763, opened 2023-09-08
DomRenderer draw selection inactive background correctly when unfocused

### LINKED ISSUE #4750: DOM renderer: Selection background is drawn incorrectly when unfocused
Repro:

1. Enable DOM renderer
2. Select some text

Actual:

Image

Expected:

All text bg should be the inactive selection color (the grey)

tisilent:
fix #4750.


Tyriar (approved):
Fix works 👍 

I reverted the demo parts as there's too many theme keys to do that (though we may add theme customization to the options tab at some point). I did set the default value for the inactive theme key to the default theme though.

Tyriar:
Added a regression test

tisilent:
I also try the playwright test.

PR #4783, opened 2023-09-10
Fix regression in force mouse selection

### LINKED ISSUE #4781: Regression in mouse events which breaks tmux
VS Code issue: https://github.com/microsoft/vscode/issues/192518

Repro:

1. Enable mouse events in tmux `set -g mouse on` in `~/.tmux.conf`
2. Open tmux
3. Hold shift (to disable passthrough mouse events) and make a selection 🐛 it shows it being selected but as soon as mouseup happens, the selection disappears

It works if mouseup occurs outside the terminal element.

Tyriar:
Shift+left click is means to force terminal selection and not passthrough any mouse events, even if a mouse mode is enabled. The change this is reverting from #4583 caused this feature to break by always installing a global listener on Terminal.element so the SelectionService never got a chance to cancel is.

Fixes #4781

@jerch FYI, we may need to reopen https://github.com/xtermjs/xterm.js/issues/4582

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -729,10 +729,8 @@ export class Terminal extends CoreTerminal implements ITerminal {
 
       if (!(events & CoreMouseEventType.UP)) {
         this._document!.removeEventListener('mouseup', requestedEvents.mouseup!);
-        el.removeEventListener('mouseup', requestedEvents.mouseup!);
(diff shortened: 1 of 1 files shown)


jerch:
@Tyriar Thats indeed an issue, have not tested yet, but I think simply removing these handlers will break half of the mouse report apps. :thinking:
Gonna re-open the old PR...

PR #4789, opened 2023-09-11
get voiceover to announce terminal output consistently

### LINKED ISSUE #3247: screen reader mode rarely reads command output with VoiceOver
When `screenReaderMode` is enabled, the output of an entered command is rarely read back by VoiceOver (e.g. is added to the live region in the DOM.)

The length of the command's output seems to determine how likely it is that the output will actually be read, for example `pwd` output will almost never be read, `ls` output maybe one in twenty times, `ls -l` output most of the time. This seems like it might be a race condition that depends on how long it takes to render the output of a command.

For some reason, I have never had this problem with VSCode's integrated terminal, only when using xterm.js standalone.

## Details
- Browser and browser version: Chrome 89, Safari 14.0.3
- OS version: macOS 11.2
- xterm.js version: 4.10

### Steps to reproduce

1. Clone and set up xterm.js
2. In demo page, enable `screenReaderMode` and VoiceOver
3. Note that entered text is spoken, but the results of commands only rarely.

meganrogge:
fix #3247

Works great in the demo. Testing in VS Code now

The "stutter" mentioned here is very minor. https://github.com/xtermjs/xterm.js/issues/3247#issuecomment-1027405267, but looking into it. 

https://github.com/xtermjs/xterm.js/assets/29464607/020d6d77-6a3f-4a68-888e-ff241c72c241

### DIFF
--- a/src/browser/AccessibilityManager.ts
+++ b/src/browser/AccessibilityManager.ts
@@ -133,26 +133,12 @@ export class AccessibilityManager extends Disposable {
           this._liveRegion.textContent += Strings.tooMuchOutput;
         }
       }
-
-      // Only detach/attach on mac as otherwise messages can go unaccounced
-      if (isMac) {
-        if (this._liveRegion.textContent && this._liveRegion.textContent.length > 0 && !this._liveRegion.parentNode) {
-          setTimeout(() => {
-            this._accessibilityContainer.appendChild(this._liveRegion);
-          }, 0);
-        }
-      }
     }
   }
 
   private _clearLiveRegion(): void {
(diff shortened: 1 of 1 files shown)


meganrogge:
Logging shows that we are doing the right thing here, so I believe the stuttering is a VoiceOver issue

Screenshot 2023-09-11 at 12 17 43 PM

meganrogge:
works in VS Code, though sometimes reads duplicate information
https://github.com/xtermjs/xterm.js/assets/29464607/745972b6-e446-480a-bd11-544c0df13a93

Tyriar (approved):

PR #4798, opened 2023-09-13
using isCursorInitialized for domrenderer

### LINKED ISSUE #4790: Domrender not using isCursorInitialized
on vscode
step:
1. using webgl render
2. create new terminal
3. reload window
terminal do not display cursor

step:
1. using dom render
2. create new terminal
3. reload window
terminal display cursor

tisilent:
fix #4790

### DIFF
--- a/src/browser/renderer/dom/DomRendererRowFactory.ts
+++ b/src/browser/renderer/dom/DomRendererRowFactory.ts
@@ -218,7 +218,7 @@ export class DomRendererRowFactory {
         }
       }
 
-      if (!this._coreService.isCursorHidden && isCursorCell) {
+      if (!this._coreService.isCursorHidden && isCursorCell && this._coreService.isCursorInitialized) {
         classes.push(RowCss.CURSOR_CLASS);
         if (this._coreBrowserService.isFocused) {
           if (cursorBlink) {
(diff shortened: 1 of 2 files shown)


tisilent:
Testing failed in Linux... :unamused:

Tyriar:
It would be initialized because for each test `Terminal.reset()` is called:

https://github.com/xtermjs/xterm.js/blob/32c616512d428cb392badfcde7d7ee69b37bfb94/test/playwright/SharedRendererTests.ts#L18

This does not reset the cursor initialization state:

https://github.com/xtermjs/xterm.js/blob/32c616512d428cb392badfcde7d7ee69b37bfb94/src/browser/Terminal.ts#L1231-L1251

Since creating a new terminal can complicate the test side effects, we could solve this by making a new inject function:

```ts
export function injectSharedRendererTestsStandalone(ctx: ISharedRendererTestContext): void {
  ...
}
````

These can then re-create the terminal before each test which is probably just a matter of calling dispose terminal and then `openTerminal`. Similar to what happens in this test but instead in a new `describe`/`beforeEach`:

```ts
  test('render when visible after hidden', async () => {
    await openTerminal(ctx);
    await ctx.page.evaluate(`
      if ('term' in window) {
        try {
          window.term.dispose();
        } catch {}
      }
    `);
    await ctx.page.evaluate(`document.querySelector('#terminal-container').style.display='none'`);
    await ctx.page.evaluate(`window.term = new Terminal()`);
    await ctx.page.evaluate(`window.term.open(document.querySelector('#terminal-container'))`);
    await ctx.page.evaluate(`document.querySelector('#terminal-container').style.display=''`);
    await pollFor(ctx.page, `window.term._core._renderService.dimensions.css.cell.width > 0`, true);
  });
```

tisilent:
I see, adjust it.

tisilent:
I still need to modify. This way of opening the page is inefficient.

Tyriar (approved):
LGTM provided it changes after my changes to also run these in canvas/webgl 👍

PR #4801, opened 2023-09-14
using viewport relative coordinate

### LINKED ISSUE #4799: Webgl renderer: Inactive cursor can render in the wrong spot
Image

This doesn't happen with DOM or canvas, and switching to and from webgl doesn't fix it.

tisilent:
fix #4799


tisilent:
Stuck while writing regression tests.
general
`
      await ctx.value.proxy.resize(20, 10);
      await ctx.value.proxy.resize(50, 20);
      await ctx.value.proxy.scrollToBottom();
      await ctx.value.proxy.scrollLines(-2);
      await ctx.value.proxy.blur();
`
Scrollbar not displayed.

Tyriar:
@tisilent I wouldn't have thought you'd need to see the scroll bar for this? Did you write enough text to full the viewport so scrollLines(-2) does something?

PR #4807, opened 2023-09-15
Move WidthCache measure container to helper container

### LINKED ISSUE #4805: WidthCache causing horizontal overflow (scrolling)
WidthCache, which is placed on the body, messes with overflows and makes a horizontal scrollbar for `50000px`. There is no way to make it not overflow and hiding it breaks the spacing entirely.

## Details
- Browser and browser version: Firefox 117.0 and Chrome 116.0.5845.179
- OS version: Linux / Windows
- xterm.js version: 
```
"xterm": "5.3.0",
"xterm-addon-fit": "0.8.0",
"xterm-addon-search": "0.13.0",
"xterm-addon-search-bar": "0.2.0",
"xterm-addon-web-links": "0.9.0",
 ```

### Steps to reproduce

1. Add a new console in a react application
2. (Not sure if changes anything) Add these options
```
const theme: ITheme = {
    background: '#21212C',
    cursor: '#21212C',
    black: '#000000',
    red: '#E54B4B',
    green: '#9ECE58',
    yellow: '#FAED70',
    blue: '#396FE2',
    magenta: '#BB80B3',
    cyan: '#2DDAFD',
    white: '#d0d0d0',
    brightBlack: 'rgba(255, 255, 255, 0.2)',
    brightRed: '#FF5370',
    brightGreen: '#C3E88D',
    brightYellow: '#FFCB6B',
    brightBlue: '#82AAFF',
    brightMagenta: '#C792EA',
    brightCyan: '#89DDFF',
    brightWhite: '#ffffff',
    selectionBackground: '#FAF089',
};

const terminalProps: ITerminalOptions = {
    disableStdin: true,
    cursorStyle: 'underline',
    allowTransparency: true,
    fontSize: 12,
    fontFamily: 'Menlo, Monaco, Consolas, monospace',
    theme: theme,
    allowProposedApi: true,
};
```
3. Observe that you can scroll `50000px` to the right, only when console loads up. Before console loading scrolling behaves normally.

### LINKED ISSUE #4806: WidthCache is added document.body, it should be a child of .xterm-helpers ideally
This could cause issues depending on styles differ between the terminal element and `body`

https://github.com/xtermjs/xterm.js/blob/32c616512d428cb392badfcde7d7ee69b37bfb94/src/browser/renderer/dom/WidthCache.ts#L77

mofux:
Fixes #4806 
Fixes #4805 

- Moves the `WidthCache` measure container into the Terminal's helper container `.xterm-helpers` (analog to `DomMeasureStrategy` used by the `CharSizeService`)
- The spans used for the different measurements now also use the `xterm-char-measure-element` class, which makes sure that those elements are not visible (`visibility: hidden`).
- Uses the `document` reference of the terminal (which refers to the `ownerDocument` of the terminal element). This makes sure we use the correct `document` reference when creating DOM elements.

Tested against latest Chrome and Firefox on Linux

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -561,7 +561,7 @@ export class Terminal extends CoreTerminal implements ITerminal {
   }
 
   private _createRenderer(): IRenderer {
-    return this._instantiationService.createInstance(DomRenderer, this.element!, this.screenElement!, this._viewportElement!, this.linkifier2);
+    return this._instantiationService.createInstance(DomRenderer, this._document!, this.element!, this.screenElement!, this._viewportElement!, this._helperContainer!, this.linkifier2);
   }
 
   /**
--- a/src/browser/renderer/dom/DomRenderer.ts
+++ b/src/browser/renderer/dom/DomRenderer.ts
@@ -47,9 +47,11 @@ export class DomRenderer extends Disposable implements IRenderer {
   public readonly onRequestRedraw = this.register(new EventEmitter<IRequestRedrawEvent>()).event;
 
   constructor(
+    private readonly _document: Document,
     private readonly _element: HTMLElement,
     private readonly _screenElement: HTMLElement,
     private readonly _viewportElement: HTMLElement,
+    private readonly _helperContainer: HTMLElement,
     private readonly _linkifier2: ILinkifier2,
     @IInstantiationService instantiationService: IInstantiationService,
     @ICharSizeService private readonly _charSizeService: ICharSizeService,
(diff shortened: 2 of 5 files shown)


Tyriar (approved):
lgtm, @jerch could you give this a look too?

jerch (approved):
LGTM, thx @mofux.

Regarding the global `WidthCache` handling we prolly have to rework things here, but lets keep that for an actual PR dealing with this.

jayarjo:
Is this fix merged and released yet? Observing overflow problem in Chrome on Mac.

PR #4809, opened 2023-09-16
Correct graphemes addon readme and typings

silamon:
Small copy-paste fix for the typings file and readme file in the new graphemes addon.


PerBothner:
I'm OK with these changes.

One might argue that it is confusing to use singular "Grapheme" in `UnicodeGraphemeProvider` and plural "Graphemes" in `UnicodeGraphemesAddon.ts` - and I would not disagree. It would probably be better to pick one or the other.

Tyriar (approved):
LGTM

For UnicodeGraphemeProvider using singular isn't a big deal as it's an internal file, I guess making it consistently plural would be better though.

Issue #4813, opened 2023-09-19
DOM clipping some emoji characters


theflyingape:
Upfront, I had to switch back to DOM renderer after this minor dot update to 5.3.0, because it introduced a weird WebGl context error similar to what was reported in #4804 -- is WebGL going away in v6 from #4779 ?

Back to this issue, I noticed *some* emoji characters under DOM do not fill both spaces, but instead clip.  I continue to have the unicode11 add-on with `term.unicode.activeVersion = '11'` and clipping did not occur under WebGL.

![image](https://github.com/xtermjs/xterm.js/assets/19166945/3dbe8b1e-7ec7-4716-a401-b865598e64e3)
Note Class, HP, Weapon,, and Armor emoji are clipped but all of the others are not.

jerch:
Hmm yeah, thats a downside of the new char handling in the DOM renderer - it highly depends on correct wcwidth values. While the canvas and the webgl renderers have special handling for allowing chars to "overprint" to the right, thats not so easy anymore with the letter-spacing cell metrics in the DOM renderer.
There might be a workaround possible though, if we isolate much wider chars than the wcwidth value indicates, and place them with a sized span and proper overflow rules. Not sure yet if thats applicable, as it might screw up non monospace fonts.

Ideally the newer unicode addon has the right wcwidth values for those codepoints. You can also check what unicode 15 has to say about them (look for east asian width property).

mofux:
I believe that this problem has always existed with the DOM renderer, and it is not easy to fix without having regressions in performance. The character is not clipped (as it is allowed to overflow), but the cell (span) after the emoji has a background color set, which draws over the overflowing emoji, making it appear as clipped.

This does **not** happen if there is no background color set as demonstrated here:

![image](https://github.com/xtermjs/xterm.js/assets/2785983/c0e341f0-45dc-4a45-9251-506ee103abf9)

The only proper fix I can think is to draw the background and the characters in separate layers, but that would introduce more DOM elements which will decrease the rendering performance.

jerch:
@mofux Youre right, I forgot about the BG issue. I also think that separately drawing the BG is not viable for the DOM renderer.

[...]

PR #4818, opened 2023-09-20
DOM Renderer: Render background separately

mofux:
This PR attempts to resolve some rendering issues with the DOM renderer when a cell has a background color set.

It does so by creating separate spans for backgrounds. These background spans are positioned absolute and are added to the row div before any of the text spans (so the text spans will always be drawn above).

Text spans can now be merged even if the background color differs, so in many scenarios we won't even produce more spans than before.

As a little bonus, the text spans are now `display: inline` which should provide a little performance boost and might be beneficial when dealing with ligatures.

**TODO**

- [ ] Do not create background span if there is no background
- [ ] Remove selection layer and draw it as part of the background
- [ ] Make sure to handle cursor and decorations properly
- [ ] Adjust tests
- [ ] Analyze how performance differs to old DOM renderer

### DIFF
--- a/css/xterm.css
+++ b/css/xterm.css
@@ -160,6 +160,20 @@
     overflow: hidden;
 }
 
+.xterm .xterm-rows div {
+    position: relative;
+}
+
+.xterm .xterm-rows span {
+    position: relative;
+}
+
+.xterm .xterm-rows span.xterm-bg {
+    position: absolute;
+    top: 0;
+    bottom: 0;
+}
+
 .xterm-dim {
     /* Dim should not apply to background, so the opacity of the foreground color is applied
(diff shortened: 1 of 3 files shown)


jerch:
> beneficial when dealing with ligatures.

And BIDI. :smile_cat:

Tyriar:
> This PR attempts to resolve some rendering issues with the DOM renderer when a cell has a background color set.

Can you elaborate on the issues this will fix?

bmeares:
Perhaps this will fix the issues with rendering double-width chars (#4813)? Linking for reference

mofux:
I‘m closing this PR because I never got it working properly without quirks.

Issue #4800, opened 2023-09-13
Proposal: more compact and flexible BufferLine data structure


PerBothner:
See [this branch for a prototype](https://github.com/PerBothner/xterm.js/tree/buffer-cell-cursor) of this proposal.  The prototype is not usable: a lot of things work; a lot don't.

The BufferLine data structure contains a `_data` field, which is a `Uint32Array` with 3 elements per column.  This makes for fast O(1) mapping from column number to character/cell information, but it has anumber of disadvantages:

* It is not memory-effecient.

* Convoluted encoding due to squeezing attribute values into available bits in the foreground/background elements.

* Tied to a terminal model with an integral number of fixed-width cells.  Supports double-width characters and grapheme clusters (somewhat clumsily), but no variable-width fonts, or any glyphs whose width isn't an integral number of cells. Many languages don't work well with fixed-width characters. For other languages being forced to use fixed-width character is unaestheic or unfriendly. (This includes to some extent English.)  (Support for variable-width fonts is not part of the current proposal, but I do have some ideas for what can be done.)

* Limited extensibility: There aren't a lot of available bits left, and there is no room for properties that require more than a few bits. Anything that doesn't fit has to be added the the extended-attribute object, which is more expensive.

* Clumsy/expensive reflow when window width changes.

### Summary

The main fields of a `BufferLine` become:

* A `_text` string that contains all the characters in the string. This is the concatention of all the content code (for simple characters), and the `_combined` elements, and subsumes both.

* The `_data` array is a combination of cell runs (represented as lengths in the `_text` string), and bg/fg/attribute values. Each element is a 4-bit "kind" (for example `SKIP_COLUMNS` for "null" columns), and 28 bits of data (such as a color value and the length of a text run).
The `_dataLength` field contains the "current" (active) length of the `_data` array, to allow for future insertions.
The `_data`  array only needs as many elements as actually used, though pre-allocating extra space for "growth" is typically worthwhile.

In the prototype, a "run" is either one or more non-composed BMP characters of the same width; a run of "null" columns; or a single "other" glyph (a cluster or a non-BMP character).  If the number columns spanned by the `_data` elements before `_dataLength`, an implicit `SKIP_COLUMNS` represents the rest of the line.

A `CellData` contains the current index into the `_data` and `_text` arrays. Since a single `_data` element can represent a run of multiple columns (when all characters are BMP and the same width) the CellData also maintains a column-offset relative to the start of the current `_data` element.

[...]

Issue #4232, opened 2022-10-27
Defer reflow for rows above viewport to idle callbacks


Tyriar:
Resize is one of the bigger bottlenecks, especially in an embedder that has many terminal tabs or when terminal columns is high. We could use a task queue to defer any pending resize operations and either drop or flush when the row is actually needed.

jerch:
Will the deferred array buffer resizes in #4115 already do here? They give me ~4 times faster resize runtime. But there is still some work needed over there (got some errors on the `DebouncedIdleTask` object there).

Tyriar:
Delaying the resizing of the buffer would help a lot, combining that with also only resizing/reflowing until either it's needed or an idle callback would speed things up even more. I had a look at this and it's quite a bit more difficult than I initially anticipated, I think a pre-requisite to this is simplifying reflow to be able to run on a range of rows and changing that code always scares me 😅

jerch:
Hmm reflowing only the viewport (in a sense of currently visible area) when resizing occurs, and postponing everything else should remove most of main thread stalls from resizing.

The ugly part would be the state holding and dealing with different resize states on the ringbuffer. I wonder if it would be possible to keep the current logic as it is, but just shard the reflow into smaller buffer pieces backwards on the scrollbuffer - by decorating the buffer with a helper structure that maintains the reflow state? Well postponing reflow will introduce several awkward situations, where reflow has to be realized immediately:
- on viewport scroll: When the user scrolls back in scrollbuffer, content must be aligned already. So there must be a way to stop the idle callback mechs and go into a forced reflow up to the scroll position.
- on data input: The line recycling during data input may not pull shorter or bigger lines from top of scrollbuffer. Thats not a biggie at all, the line length can simply be realigned as they get recycled.
- on serialize: Any buffer area for serialization must be reflowed already. Basically the same as on viewport scroll.

As shrinking reflow always removes data from the top of the buffer, we cannot do partial reflow in the middle, but always have to progress sequentially backwards. It gets interesting when several reflows come in in short progression - I think here it would be possible to skip finishing earlier reflows, and just jump to the new sizes.

To put this all together, would something like this work? Schematically:
- onResize:
  - update `.cols/.rows` immediately (meaning now: thats the supposed buffer metrics)
  - place a reflow state object on the buffer, if none existing, e.g. `ReflowState`
  - update requested metrics on `ReflowState` saving new cols/rows
[...]

PR #4825, opened 2023-09-26
Change setCellFromCodePoint API and rename to setCellFromCodepoint

PerBothner:
Replace fg, bg, and eAttrs parameters by single IAttributeData parameter. In addition to making for a simpler/cleaner API, it also increases flexibility for future CellData or BufferLine changes.
(Rename function to reduce risk of using accidentally using old function.)

Also removed redundant eraseAttr parameter in deleteCells, insertCells, and replaceCells.

These changes will be helpful for issue #4800.


jerch (approved):
This is what you get from evolving interfaces over time, was way overspecified :smile_cat: 

Thx for cleaning it up - LGTM :+1:

Tyriar (approved):

PerBothner:
Thanks! I'm continuing to work on issue #4800 - currently focusing on line-wrapping. (I'm looking into doing lazy reflow as per issue #4232.)

PR #4843, opened 2023-10-13
release 5.3.0 with our patches

szymonkaliski:
**WIP**


src/browser/input/CompositionHelper.ts
```diff
@@ -169,6 +169,7 @@ export class CompositionHelper {
           }
           if (input.length > 0) {
             this._coreService.triggerDataEvent(input, true);
+            this._textarea.value = '';
```
Tyriar, line 172: FYI clearing this causes problems with screen readers

szymonkaliski: yes thank you, I PR-ed this to the wrong upstream!

PR #4847, opened 2023-10-17
Have addons implement their API at compile time

### LINKED ISSUE #4828: Can addons implement their own interfaces?
https://github.com/xtermjs/xterm.js/issues/4821 was caused because the compiler does not require that SearchAddon implemented the `SearchAddon` interface in `xterm-addon-search`. We do this in the core lib here:

https://github.com/xtermjs/xterm.js/blob/c08062420f6604f4c25e0e0a3b436ec4a5d49c20/src/browser/public/Terminal.ts#L16-L23

tisilent:
fix #4828 
1. Missing `clearTextureAtlas` found in canvas addon
2. xterm-addon-serialize , xterm-addon-unicode-graphemes  may  `error TS2307: Cannot find module` will happen and the reason cannot be found :dizzy:


tisilent:
ok . All addons have been added.

tisilent:
@Tyriar After you handle #4859, I will resubmit it.

Tyriar:
I'm trying to fix the conflicts now

Tyriar (approved):
Just what I was after 🤌

jerch:
@tisilent Wow - thats a really interesting idea to solve the issue. :heart: 

On a sidenote - that even solves another (non?) issue: its not possible anymore to accidentally import the addon into itself due to the path masking rule, whoever would have wanted that in the first place :smiley_cat: 

I still wonder - was there a specific reason to do it via a path rule, and not by directly importing the types from the d.ts files?

tisilent:
@jerch Without specifying path rules, TS compile may encounter problems. :sweat_smile:

tisilent:
When the '. d. ts' file uses the import. declared `module`, it cannot be found during compilation. It is uncertain if there is anything else that needs to be configured.

jerch:
Ah ok, glad you found a way around. I kinda never fully understood the `declare module` ideas/limitations - whenever I tried messing with it, it was a mess :sweat_smile:.

Issue #4850, opened 2023-10-19
node v21.0.0 breaks with xterm: Cannot read properties of undefined (reading 'includes')


davidfiala:
node v21.0.0 introduces the `navigator` object where it used to be undefined.

Which causes issues in [src/common/Platform.ts](https://github.com/xtermjs/xterm.js/blob/1c880f519969ba93888480a359e83d61c6931bdb/src/common/Platform.ts#L16-L20):

```typescript
export const isNode = (typeof navigator === 'undefined') ? true : false;
const userAgent = (isNode) ? 'node' : navigator.userAgent;
const platform = (isNode) ? 'node' : navigator.platform;

export const isFirefox = userAgent.includes('Firefox');
```

There was discussion on whether using navigator to detect node here: https://github.com/nodejs/node/pull/47769#issuecomment-1601054678 but ultimately they concluded that since Deno has introduced a navigator object, so can node.

I often see checks like this instead:

```typescript
const isNode = (typeof process !== 'undefined');
// or
const isNode = (typeof global !== 'undefined');
```

which will work for v21 and below.

PR #4851, opened 2023-10-19
proper isNode detection for node v21.0.0

davidfiala:
Fix for https://github.com/xtermjs/xterm.js/issues/4850


Tyriar (approved):
Thanks 👍

Issue #4859, opened 2023-11-01
Migrate to @xterm org on npm


Tyriar:
In order to prevent problems like https://github.com/xtermjs/xterm.js/issues/4857 in the future let's move to scoped packages on npm. I don't consider this a breaking change as it needs a manual action to update anyway.

Plan:

- [x] Start publishing beta builds to `@xterm/xterm`, `@xterm/addon-search`, etc.
- [x] Publish 5.4 to new scope
- [x] Deprecate old packages with a message redirecting users to the new package name
- [x] update all docs snippets including paths in `<script>` elements here and on xtermjs.org (added by @jerch)

jerch:
@Tyriar Wow, that already created quite a refactoring noise. Thx for getting it fixed this quickly.

Tyriar:
@jerch np, it was our tech debt week so conveniently timed 🙂 ended up being a pretty chunky VS Code PR too https://github.com/microsoft/vscode/pull/197167

octoclonius:
I tried using `@xterm/addon-attach`, `@xterm/addon-fit`, and `@xterm/addon-webgl`, and I noticed they all had the same issue.

I installed them with npm and I got their beta versions.

"@xterm/addon-attach": "^0.10.0-beta.1"
"@xterm/addon-fit": "^0.9.0-beta.1"
"@xterm/addon-webgl": "^0.17.0-beta.1"

In their respective `lib` folders (`@xterm/addon-[attach/fit/webgl]/lib/addon-[attach/fit/webgl].js`), the `.js` code has a bug. When I replace `self` with `this`, it ceases to cause me issues. I don't really know if that really fixes anything or not, but I just thought I'd share.

Tyriar:
I deprecated xterm and xterm-headless. I don't think I can give it a custom message unfortunately

PR #4860, opened 2023-11-01
Start @xterm scope with core and webgl addons

Tyriar:
Part of #4859


Tyriar:
I think everything is right, any problems should surface when I try use it in VS Code. I'll try deal with all the merge conflicts this will cause in existing PRs as I don't think GH handles edits to renamed files that well.

PR #4863, opened 2023-11-02
Handle OSC 52 for clipboard manipulation

aljazerzen:
Initial implementation of #3260 

This is a proof of concept, but I would be happy to get feedback. Mainly to these questions:

- where should this code reside in? A separate service? It needs access to the output stream of the shell process if we want to implement pasting correctly.

- what's the policy here on async functions?

- are there any utilities for dealing with arbitrary base64 encoded data?

### DIFF
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -320,6 +320,7 @@ export class InputHandler extends Disposable implements IInputHandler {
     //  50 - Set Font to Pt.
     //  51 - reserved for Emacs shell.
     //  52 - Manipulate Selection Data.
+    this._parser.registerOscHandler(52, new OscHandler(data => this.manipulateSelectionData(data)));
     // 104 ; c - Reset Color Number c.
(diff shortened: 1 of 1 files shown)


jerch:
>   where should this code reside in? A separate service? It needs access to the output stream of the shell process if we want to implement pasting correctly.

Prolly an addon would be the easiest way currently. Note that there is already PR #4220 for OSC52. Plz check that out, maybe a combined effort is better here.

>  what's the policy here on async functions?

The terminal core is completely free of async functions, mainly for speed reasons. Thats especially true for the vt parser. The parser still supports async handlers, plz see https://xtermjs.org/docs/guides/hooks/#async-actions-in-parser-hooks for details and its drawbacks. 

>  are there any utilities for dealing with arbitrary base64 encoded data?

For synchronous high performance base64 decoder/encoder maybe have a look at https://github.com/jerch/xterm-wasm-parts.

Tyriar:
Thanks for the PR! This has already had a lot of effort put into it in https://github.com/xtermjs/xterm.js/pull/4220 and is quite close to getting merged at this point.

PR #4879, opened 2023-11-11
Align character position in a11y tree with the actual rendering

JasonXJ:
This helps screen readers to draw selection outline at the correct position.


JasonXJ:
Before: 
20231110_21h43m49s_grim

after:
20231110_21h41m38s_grim

Tyriar (approved):

Issue #4934, opened 2024-01-02
web link addon doesn't parse URLs containing `%20` properly


szymonkaliski:
https://github.com/xtermjs/xterm.js/blob/master/addons/addon-web-links/src/WebLinkProvider.ts seems to break on URLs containing `%20` (and other UTF-8-encoded symbols).

The issue comes from this line:
https://github.com/xtermjs/xterm.js/blob/master/addons/addon-web-links/src/WebLinkProvider.ts#L68

`decodeURI` decodes `%20` so a URL like `http://test.com?param=a%20b` gets decoded to `http://test.com?param=a b`, which makes the `text` not equal to `urlText`.
I'd be happy to approach fixing this and sending a PR, but I'm honestly not sure what this test is protecting us from -- in which case `new URL(text)` would succeed but `text` would not be `urlText` and that wouldn't be a false positive?

## Details

(I don't think this matters for this specific bug)

- Browser and browser version: Chrome 120.0.6099.129
- OS version: macOS Sonoma 14.1
- xterm.js version: 5.3.0

### Steps to reproduce

1. `echo "http://test.com"; echo "http://test.com?param=a%20b"`
2. first link is clickable, the second is not

jerch:
Thx for finding this edge case, its indeed not covered currently. Would an additional clause w'o the `decodeURI` already fix this?

Background on the weird looking additional check - `new URL` does sometimes automagic transformations on the provided url strings, which I deemed not to be equal for security reasons (beside appending a slash, which happens pretty often). This might look paranoid, but since we dont control what `new URL` does here (up to faulty URL construction impl), this check tries to ensure basic equality. The `decodeURI` part is needed for utf-8 encoded url strings to properly match, and I kinda forgot to test url-encoded strings in the first place.

Whether that additional test is really needed or a sloppier handling would be sufficient - idk for sure. I normally try to go with security first, and since we outsource the URL construction to the browser with `new URL` we would inherit all implementation differences up to faults as well. I did not dig deeper into `new URL` and whether that can be used as an attack vector by malformed url strings and auto transformation, instead went with that equality check as a basic security measure.

szymonkaliski:
> Would an additional clause w'o the `decodeURI` already fix this?
Not sure if I understand what you're proposing :) Can you give an example?

I think what we could do is to test only the `url.origin` part, for example replace the original `if` statement with:
```js
if (!text.startsWith(url.origin.toString())) {
  continue
}
```

What do you think?

[...]

PR #4937, opened 2024-01-05
fix url parsing for urls with percentage sign in them

szymonkaliski:
This PR fixes the issue outline and discussed in here https://github.com/xtermjs/xterm.js/issues/4934

### DIFF
--- a/addons/addon-web-links/src/WebLinkProvider.ts
+++ b/addons/addon-web-links/src/WebLinkProvider.ts
@@ -41,6 +41,18 @@ export class WebLinkProvider implements ILinkProvider {
   }
 }
 
+function baseUrlString(url: URL): string {
+  if (url.password && url.username) {
+    return `${url.protocol}//${url.username}:${url.password}@${url.host}`;
+  }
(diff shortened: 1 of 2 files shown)


addons/addon-web-links/src/WebLinkProvider.ts
```diff
@@ -64,8 +76,7 @@ export class LinkComputer {
       // - append /   also match domain urls w'o any path notion
```
jerch, line 76: Can you also edit the comment to reflect the code, e.g. remove the listing and mention the explicit equality testing up to the host part.

jerch (changes requested):
LGTM, just a minor nitpick from my side below.

addons/addon-web-links/src/WebLinkProvider.ts
szymonkaliski: updated the comment, lmk what you think

jerch (approved):
Yepp LGTM, thx alot! :+1:

PR #4940, opened 2024-01-13
Update global object fix

octoclonius:
The changes are to the `webpack.config.js` files in each of the addon's folders. I set the global object to `globalThis`. I think this should fix the addons because when I import them, I get an error saying `self` is not defined.


octoclonius:
The changes are to the `webpack.config.js` to set the global object to `globalThis`. I think this should fix it.

ManojkumarV (approved):

ManojkumarV (changes requested):
>

GoodbyeNJN:
Is there any progress on this PR? I am also troubled by this problem, recompiling addon after using this patch works fine for me.

FRSgit:
@Tyriar Do you think this PR is something that could be merged in nearest future? It's blocking usage of the library with nuxt (in SSR context it tends to fail).

Tyriar (approved):

PR #4953, opened 2024-02-01
Expose API method for writing to application side (#4948)

### LINKED ISSUE #4948: expose API method for writing to application side
Coming from #4220 and this comment https://github.com/xtermjs/xterm.js/pull/4220#issuecomment-1793095087

It would be nice to have a method to write to application side from xterm.js' API. Internally xterm.js uses `this.coreService.triggerDataEvent` for this, which can easily be exposed at the API as well.

_interface suggestion_
```typescript
  export class Terminal implements IDisposable {
    ...
    /**
     * Input data to application side.
     * The data is treated the same way as typed input at the terminal (will appear in the onData event).
     */
    input(data: string): void;
    ...
}
```

Such a method would help to automate data input programmatically, e.g. from the clipboard addon w'o hooking into hidden internals.

arencoskun:
This PR adds an input method which takes data as the parameter as described in the issue. If there are any issues, you can point it out and I'll fix it.

(Sorry if there's an issue, I am a beginner.)

Fixes #4948.

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -1193,9 +1193,10 @@ export class Terminal extends CoreTerminal implements ITerminal {
   }
   
   /**
-     * Input data to application side.
-     * The data is treated the same way as typed input at the terminal (will appear in the onData event).
-     */
+    * Input data to application side.
+    * The data is treated the same way as typed input at the terminal.
+    * (will appear in the onData event).
+    */
   public input(data: string): void {
     this.coreService.triggerDataEvent(data, true);
     return this.write(data);
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -1191,15 +1191,15 @@ export class Terminal extends CoreTerminal implements ITerminal {
 
     return false;
   }
-  
+
   /**
-    * Input data to application side.
-    * The data is treated the same way as typed input at the terminal.
-    * (will appear in the onData event).
-    */
+   * Input data to application side.
+   * The data is treated the same way as typed input at the terminal.
+   * (will appear in the onData event).
+   */
   public input(data: string): void {
     this.coreService.triggerDataEvent(data, true);
-    return this.write(data);
+    this.write(data);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


jerch (changes requested):
Welcome and thank you for looking into this.

Looks really good already, only a few my remarks from my side below.

arencoskun:
Thanks for the feedback, working on it

jerch (approved):
@arencoskun With fixing the remaining linter errors, this LGTM :+1:

arencoskun:
> @arencoskun With fixing the remaining linter errors, this LGTM 👍

Oops, forgot to fix those. Will work on it in a few moments

jerch:
@arencoskun The linter is still nagging about a whitespace is in `xterm.d.ts`.

Oh and btw - the declaration also needs to be added to `xterm-headless.d.ts` (sorry forgot about that before).

arencoskun:
> @arencoskun ~The linter is still nagging about a whitespace in `xterm.d.ts`.~ (guess you were faster 😸 )
> 
> Oh and btw - the declaration also needs to be added to `xterm-headless.d.ts` (sorry forgot about that before).

No problem, also adding that

arencoskun:
Sorry, forgot to mark as draft

jerch:
Whoopsie - the headless API-bundling class in `src/headless/public/Terminal.ts` would also need the impl forwarding similar to `browser/public/Terminal.ts`.

(Why? - We basically maintain 2 versions if xterm.js Terminal class - one for browser and for headless nodejs usage. So parts of the API bundling is doubled here.)

arencoskun:
> Whoopsie - the headless API-bundling class in `src/headless/public/Terminal.ts` would also need the impl forwarding similar to `browser/public/Terminal.ts`.
> 
> 
> 
> (Why? - We basically maintain 2 versions of xterm.js Terminal class - one for browser and for headless nodejs usage. So parts of the API bundling is doubled here.)
> 
> 
> 
> 

Thanks for the info, I've been busy tonight but most probably tomorrow I'll finish it off.
(Also, seems like you can't mark PRs as drafts from the mobile app. That's weird 😀 )

jerch:
@Tyriar Could you also review this PR? Because of the API addition I dont want to merge it w'o being reviewed by you.

Tyriar (approved):
Thanks! I just tweaks the docs and removed the doc strings from the internal files to reduce duplication/bitrot

PR #4963, opened 2024-02-17
Polish convertEol api jsdoc

unional:
Fix doc typo


typings/xterm.d.ts
```diff
@@ -48,7 +48,7 @@ declare module '@xterm/xterm' {
     /**
      * When enabled the cursor will be set to the beginning of the next line
      * with every new line. This is equivalent to sending '\r\n' for each '\n'.
-     * Normally the termios settings of the underlying PTY deals with the
+     * Normally the terminal settings of the underlying PTY deals with the
```
jerch, line 51: Thats not a typo - _termios_ is part of the general terminal interface as defined by [POSIX](https://pubs.opengroup.org/onlinepubs/007904975/basedefs/termios.h.html).

unional: Um... thanks.
Based on that document, `termios` is the name of `struct termios`, where the file `termios.h` is described as `...contains the definitions used by the terminal I/O interfaces...`.

It's also used in this way in [General Terminal Interface](https://pubs.opengroup.org/onlinepubs/007904975/basedefs/xbd_chap11.html#tag_11).

IMO that is too specific as a general description in the tooltip and the more general `terminal settings` would be better. The consumer of the library (e.g. me) would not know about the name of such data structure in C/CPP.

Also the description here does not refer to any specific member in the `termios` struct (e.g. `c_oflag`). So using such term does not help the consumer to know more about this property.

Alternatively, it can be changed to a code block and like to the POSIX page:

```md
Normally the underlying PTY deals...to '\r\n' through the struct [`termios`](https://pubs.opengroup.org/onlinepubs/007904975/basedefs/termios.h.html) and this setting should not be used...
```

jerch: _termios_ is the name of the struct to configure terminal IO settings, hence the name (An older incarnation was named _termio_, whether that additional _s_ stand for settings or struct, or was introduced because of the need to separate itself from the older interface, does not matter much).

> The consumer of the library (e.g. me) would not know about the name of such data structure in C/CPP.

That OS interface is really important for interactive terminal app developers, and almost every programming language has a package or module called _termios_. Also its manpage is called _termios_. The fact that you never heard about it, should not be a reason to dilude the description or to give less information. The opposite is the case - putting the term _termios_ here helps ppl to look up the concepts, if they are in doubt.

But I dont care much about the specific wording, as long as the correct semantic is transported. Your current change to

> Normally the terminal settings of the underlying PTY deals with the

is not correct, as the PTY does not hold terminal settings, but terminal IO settings. I'd attribute the term "terminal settings" more to settings in xterm.js itself (or any other terminal emulator being used behind the PTY).

Rephrased suggestion from my side:

> Normally the IO settings of the underlying PTY (termios) deals with the

unional: updated

jerch (approved):

Tyriar (approved):

PR #4973, opened 2024-02-29
Fix spacing issue when measuring before element is attached

Tyriar:
See microsoft/vscode#204690

image


jerch:
@Tyriar  Oh wow, thats awkward - any idea why the letter measuring might return 0 sometimes?

PR #4977, opened 2024-03-02
Add @xterm scope to packages

dstein64:
This adds the `@xterm` package scope to two places in the `README`.


jerch (approved):
Thx for the cleanup work :heart:

PR #4988, opened 2024-03-07
DOM Renderer: CursorCell does not use inline backgroundColor style and fix bar cursor style not blinking

### LINKED ISSUE #4987: Block style cursor not blinking after upgrade 5.3.0 -> 5.4.0
I've updated from xterm.js 5.3.0 to 5.4.0 and noticed that the cursor is no longer blinking when it is focused and in block mode.

With these options set:
           "cursorBlink": true,
           "cursorStyle": "block",
           "cursorInactiveStyle": "outline"

The cursor goes from outline to block style when the terminal is focused, but the block cursor **does not blink** as expected.  It is supposed to blink, and **does** blink under xterm.js 5.3.0.  

With cursorStyle set to "underline" and cursorBlink to true, the cursor **does** blink as expected.  If I change cursorBlink to false, then the underline cursor **does not** blink as expected.  (showing that the cursorBlink attribute is being honored for the underline cursor)

It only appears that the cursor is fails to blink when cursorBlink is true and cursorStyle is block.

## Details
- Browser and browser version: Version 121.0.6167.139 (Official Build) (64-bit)
- OS version:  Linux
- xterm.js version: 5.4.0

### Steps to reproduce

1. Set cursorStyle to block
2. Set cursorBlink to true
3. set cursorInactiveStyle to outline
4. focus the terminal to activate the cursor
5. If cursor goes active and turns into a block, but does not blink, problem is reproduced.

tisilent:
fix #4987

### DIFF
--- a/src/browser/renderer/dom/DomRenderer.ts
+++ b/src/browser/renderer/dom/DomRenderer.ts
@@ -209,8 +209,8 @@ export class DomRenderer extends Disposable implements IRenderer {
       ` animation: blink_block` + `_` + this._terminalClass + ` 1s step-end infinite;` +
       `}` +
       `${this._terminalSelector} .${ROW_CONTAINER_CLASS} .${RowCss.CURSOR_CLASS}.${RowCss.CURSOR_STYLE_BLOCK_CLASS} {` +
-      ` background-color: ${colors.cursor.css} !important;` +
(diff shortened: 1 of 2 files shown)


tisilent:
Specificity of !important and animation.
!important > animation.

PR #4997, opened 2024-03-14
Introduce opt-in glyph scaling to achieve GB18030 compliance

### LINKED ISSUE #4969: Automatic glyph scaling based on unicode width and following character content
We have a report related to GB18030 compliance (Chinese government standard) that calls out the rendering of the `Ⅻ` char overlapping with the follow character. This is because we think it's a narrow width char but render it exactly as the font renders.

image

We could handle this case and solve the problem of overlapping character once and for all by scaling down the glyph to fit in the cell. This is what Windows Terminal's old renderer did:

image

https://github.com/microsoft/terminal/blob/release-1.19/src/renderer/dx/CustomTextLayout.cpp#L795-L836

Of course when talking about overlapping we can't ignore emoji which overlap but often contain a space after as this is what most terminals do. If we started scaling down all emoji I'm sure we would get many bug reports complaining about it. To handle this particular case generically I think we should also consider the following character, if is just whitespace then render normally without scaling,

Tyriar:
Fixes #4969

I've tried to keep the overhead for the webgl renderer as small as possible, there's just a check for an option (via optimized `rawOptions`) and fetching the width of each cell from `content`.


1atAlcone (approved):
Isnt  it exciting?

Issue #4944, opened 2024-01-22
`;` inside the URL breaks the OSC hyperlink


schaffino:
Tried this in both VSCode and Jupyterlab. It seems the OSC hyperlink integration doesn't work properly when the linked URL contains a ';' 

The following submitted to terminator terminal emulator results in a link directing me to https://www.example.com/query;arg=1;arg=2;arg=3

printf '\e]8;;https://www.example.com/query;arg=1;arg=2;arg=3\e\\This is a link\e]8;;\e\\\n'  

Running the same in jupyter or VSCode, I end up being redirected to https://www.example.com/query, with the arguments truncated

## Details
- Browser and browser version: Any
- OS version:  Any
- xterm.js version: Any supporting OSC hyperlinks

### Steps to reproduce

1.  Output a valid OSC hyperlink in tooling using xterm.js (See printf statement above)
2.  Observe the link is truncated

Tyriar:
Before fixing this we should verify that `;` is expected to be supported when not URL encoded as `%3B`

jerch:
@Tyriar https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#encodings states, that additional uri params may not contain `:` or `;` due to its ambiguity with the sequence syntax. Thus I'd say this works as intended by the spec.

Tyriar:
Cool, I suspected that might be the case. Closing as designed then

schaffino:
@Tyriar @jerch 
The parameters its referencing in the doc above is referring to the optional parameter argument of the OSC-8 hyperlink, not query parameters in the URI. It states that only id is specifically defined in the spec for this argument. My reading is that the URI can still be any string in the range  32–126. Given most other terminal emulators support this i believe this is still a bug.

https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#the-escape-sequence


It explicitly says the URI can be anything in the 32-126 when stating it cannot contain : or ;, which is a seperate argument to the URI. It states the URI can be anything in the ascii range.  Without support for ; there are plenty of valid URI's that cannot be linked

Tyriar:
@schaffino makes sense, thanks for the follow up. I read the "additional parameters" part wrong.

jerch:
Eww, seems I also misread that.

josiahhudson:
Attempted a fix here: https://github.com/xtermjs/xterm.js/pull/5003

PR #5003, opened 2024-03-15
Fix for issue #4944.

josiahhudson:
Original code performed a simple `data.split(';')` which split on every ";" in the string.  This new version treats the first ";" as the only split character, ensuring that params are properly separated from the URI.

### DIFF
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -2972,7 +2972,7 @@ export class InputHandler extends Disposable implements IInputHandler {
    * feedback. Use `OSC 8 ; ; BEL` to finish the current hyperlink.
    */
   public setHyperlink(data: string): boolean {
-    const args = data.split(';');
+    const args = data.match(/^([^;]*);(.*)$/)?.slice(1) ?? [];
     if (args.length < 2) {
       return false;
     }


src/common/InputHandler.ts
```diff
@@ -2972,7 +2972,7 @@ export class InputHandler extends Disposable implements IInputHandler {
    * feedback. Use `OSC 8 ; ; BEL` to finish the current hyperlink.
    */
   public setHyperlink(data: string): boolean {
-    const args = data.split(';');
+    const args = data.match(/^([^;]*);(.*)$/)?.slice(1) ?? [];
```
jerch, line 2975: Isn't this complicating things for no good reason? Why not simply doing:
```typescript
const idx = data.indexOf(';');
if (idx === -1) {
  // malformed sequence, just return as handled
  return true;
}
const arg0 = data.slice(0, idx);
const arg1 = data.slice(idx + 1);
...
```
Should also be much much faster than any regexp...

jerch (dismissed):
@josiahhudson Thx for looking into this. I have a few remarks:

- Plz see my comment below, imho doing it with a regexp is a bit over-engineered.
- Would also be nice to have this covered by a test case.
- I just saw, that the `setHyperlink` handler returns `false` in some cases, which is suboptimal, as the parser has an optimization for the the `true` case (since it is a default handler in xterm.js it can always return `true`). Could you also change that?

josiahhudson:
lol, feedback makes sense. Happy to make the changes. I struggled to find a good place to put a test when I was looking at this, but I'll give it another go. I may not have time to look at this again till Friday, though.

Tyriar (approved):
Thanks @josiahhudson, I tweaked it to use @jerch's suggestion and added a couple of tests

PR #5009, opened 2024-03-16
Fix dom cursor blink animation and improve readability

Tyriar:
Fixes https://github.com/xtermjs/xterm.js/issues/4987


Tyriar:
Fix wasn't this easy unfortunately 😢

vecna-sh (commented):
` `

PR #5024, opened 2024-04-08
Fix duplicate input for some IMEs

### LINKED ISSUE #5023: Duplicate input when using Chinese input methods
There was a previous report at #3191 but the linked PR does not fix this problem.

I'll begin with a description of how a certain kind of Chinese input methods work (including [the Wubi method](https://en.wikipedia.org/wiki/Wubi_method), which is very popular in China):

1. The user inputs something, and the IME updates the candidate list accordingly (`compositionstart` and `compositionupdate`)
2. Until now, nothing is committed or inserted into the text area
3. Then, the user hits another key, which bears dual functions: (1) confirms the last selection (`compositionupdate` and `compositionend`), and then (2) the key itself is considered as input for the next input session.

In the following screencast, I typed "qtqtq", and at the final "q", the problem occurs:

1. 狠狠 is inserted, which is expected
2. q is inserted, which is NOT expected

https://github.com/xtermjs/xterm.js/assets/23358293/d20db910-65ee-408a-910f-07c036a1b964

From the console log, you can see that the problem is in `_compositionPosition.end`: when I typed the final "q", the `end` position was not correctly set: it thinks the end position is 4, while the correct position should be 2. However, at this point, the composition contains the new input "q", therefore "狠狠q" is inserted.

The root cause is in that xterm.js assumed that when a new composition starts, the length of the committed input remains the same, which is not true in this case. Or it could also be a Chromium bug, when it should update composition BEFORE compositionend.

## Details
- Browser and browser version: Chrome 123.0.6312.107
- OS version: macOS 14.4.1 (23E224)
- xterm.js version: [9ba6c00](https://github.com/xtermjs/xterm.js/commit/9ba6c00a195c95fcf8292a2b9084d91450e5daae)

This bug is also reproducible in Visual Studio Code.

### Steps to reproduce

1. Install [Qingg](https://qingg.im/mac/), add "清歌五笔" from Chinese (Simplified), and select it (Qingg is a popular implementation of the Wubi method in macOS)
2. In the web demo, type "qtqtx"
3. Get "狠狠" and "x" both inserted into the terminal

ksqsf:
This is a simple fix for #5023 by removing the assumption "the end position is stable". Instead, it uses the start position of the new composition as the end position of the previous composition.

fixes #5023

### DIFF
--- a/src/browser/input/CompositionHelper.ts
+++ b/src/browser/input/CompositionHelper.ts
@@ -159,8 +159,9 @@ export class CompositionHelper {
           // otherwise input characters can be duplicated. (Issue #3191)
           currentCompositionPosition.start += this._dataAlreadySent.length;
           if (this._isComposing) {
-            // Use the end position to get the string if a new composition has started.
-            input = this._textarea.value.substring(currentCompositionPosition.start, currentCompositionPosition.end);
+            // Use the start position of the new composition to get the string
+            // if a new composition has started.
+            input = this._textarea.value.substring(currentCompositionPosition.start, this._compositionPosition.start);
           } else {
             // Don't use the end position here in order to pick up any characters after the
             // composition has finished, for example when typing a non-composition character


Tyriar (approved):
Thanks!

PR #5031, opened 2024-04-13
fix: use `this` global for `addon-fit`

### LINKED ISSUE #5030: Fit-addon is not compatible with SSR/prerendering
Here is the current content of addon-fit.js shiped with `@xterm/addon-fit@0.10.0` 

```js
!(function (e, t) {
  'object' == typeof exports && 'object' == typeof module
    ? (module.exports = t())
    : 'function' == typeof define && define.amd
    ? define([], t)
    : 'object' == typeof exports
    ? (exports.FitAddon = t())
    : (e.FitAddon = t());
})(self, () => // <=============== ISSUE HERE
  (() => {
    'use strict';
    var e = {};
    return (
      (() => {
        var t = e;
        Object.defineProperty(t, '__esModule', { value: !0 }),
          (t.FitAddon = void 0),
          (t.FitAddon = class {
            activate(e) {
              this._terminal = e;
            }
            dispose() {}
            fit() {
              const e = this.proposeDimensions();
              if (!e || !this._terminal || isNaN(e.cols) || isNaN(e.rows)) return;
              const t = this._terminal._core;
              (this._terminal.rows === e.rows && this._terminal.cols === e.cols) ||
                (t._renderService.clear(), this._terminal.resize(e.cols, e.rows));
            }
            proposeDimensions() {
              if (!this._terminal) return;
              if (!this._terminal.element || !this._terminal.element.parentElement) return;
              const e = this._terminal._core,
                t = e._renderService.dimensions;
              if (0 === t.css.cell.width || 0 === t.css.cell.height) return;
              const r = 0 === this._terminal.options.scrollback ? 0 : e.viewport.scrollBarWidth,
                i = window.getComputedStyle(this._terminal.element.parentElement),
                o = parseInt(i.getPropertyValue('height')),
                s = Math.max(0, parseInt(i.getPropertyValue('width'))),
                n = window.getComputedStyle(this._terminal.element),
                l = o - (parseInt(n.getPropertyValue('padding-top')) + parseInt(n.getPropertyValue('padding-bottom'))),
                a =
                  s -
                  (parseInt(n.getPropertyValue('padding-right')) + parseInt(n.getPropertyValue('padding-left'))) -
                  r;
              return {
                cols: Math.max(2, Math.floor(a / t.css.cell.width)),
                rows: Math.max(1, Math.floor(l / t.css.cell.height)),
              };
            }
          });
      })(),
      e
    );
  })()
);
//# sourceMappingURL=addon-fit.js.map

```

As you can see the compiled form of the addon relies on `self` which is only an alias to the global on browsers. On node environment the runtime throws `self is not defined`.

JeanMeche:
Fixes #5030

### DIFF
--- a/addons/addon-fit/webpack.config.js
+++ b/addons/addon-fit/webpack.config.js
@@ -3,29 +3,30 @@
  * @license MIT
  */
 
-const path = require('path');
+const path = require("path");
 
-const addonName = 'FitAddon';
-const mainFile = 'addon-fit.js';
+const addonName = "FitAddon";
+const mainFile = "addon-fit.js";
 
 module.exports = {
   entry: `./out/${addonName}.js`,
-  devtool: 'source-map',
+  devtool: "source-map",
   module: {
     rules: [
       {
         test: /\.js$/,
         use: ["source-map-loader"],
         enforce: "pre",
-        exclude: /node_modules/
-      }
-    ]
+        exclude: /node_modules/,
+      },
+    ],
   },
   output: {
     filename: mainFile,
-    path: path.resolve('./lib'),
+    path: path.resolve("./lib"),
     library: addonName,
-    libraryTarget: 'umd'
+    libraryTarget: "umd",
+    globalObject: "this",
   },
-  mode: 'production'
+  mode: "production",
 };


Tyriar (commented):
This one conflicts with https://github.com/xtermjs/xterm.js/pull/4940, that might fix this issue?

JeanMeche:
Yes, you're right !

Issue #4904, opened 2023-12-07
Support using option key on MacOS for Escape-Codes


Tyriar:
Request for a `macOptionIsEscape` terminal option: https://github.com/microsoft/vscode/issues/127764

PR #5033, opened 2024-04-18
Added macOptionIsEscape according to #4904

louis-baranger:
Added the option to use the `option` key on MacOS for Escape-Codes according to [https://en.wikipedia.org/wiki/C0_and_C1_control_codes](url). The user input <input> is converted to ESC+<input> as shown in the first column of the **ISO/IEC 6429 and RFC 1345 C1 control codes** table.

I might have misinterpreted what the issue meant by "[using] the option key on MacOS for Escape-Codes", so if that is the case, let me know and I will make the changes.

### DIFF
--- a/src/browser/Terminal.ts
+++ b/src/browser/Terminal.ts
@@ -1007,7 +1007,7 @@ export class Terminal extends CoreTerminal implements ITerminal {
     }
 
     // Ignore composing with Alt key on Mac when macOptionIsMeta is enabled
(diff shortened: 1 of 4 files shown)


louis-baranger:
#4904

Tyriar (commented):
Closing in favor of https://github.com/xtermjs/xterm.js/pull/5723 as it's more complete

PR #5039, opened 2024-04-21
Encode all HTML entities in serialized content

Tyriar:
Follow up from https://github.com/xtermjs/xterm.js/pull/5020 as the branch was protected


Tyriar:
Actually maybe this would make the content too large and isn't necessary

PR #5054, opened 2024-05-13
Add powerline git, LN, lock symbols

joelhockey:
Symbols taken from https://github.com/powerline/fontpatcher/blob/develop/fonts/powerline-symbols.sfd

The original symbols use height and widith of approx 1060x2048, so I modified to be roughly 1000x1000 via:

fontforge -lang=ff -c 'Open($1); SelectAll(); UnlinkReference(); Scale(100, 50); Move(0, 300); Export("svg/%n-%e.svg");' powerline-symbols.sfd

Then I edited by hand to round values to approx 2 significant digits, and then used a script to convert relative SVG commands such as 'l', 'c', 's' to use only 'M', 'L' and 'C' and used expected format with comma separators rather than spaces, and divide all values by 1000 to map into a 1x1 space.


joelhockey:
@Tyriar or @jerch or anyone, is there a better way to request a review for a pull request?

joelhockey:
@meganrogge @sbatten who have committed to CustomGlyph.ts in the past

jerch:
@joelhockey Imho @Tyriar wrote the custom glyph handling code, so he might be best to review it.

Tyriar (approved):
This is fantastic! 👏 

image

Zoomed:

image

Big letter spacing and line height:

image

joelhockey:
Thanks @Tyriar, would you also be able to review https://github.com/xtermjs/xterm.js/pull/4879 by @JasonXJ.  We both work on ChromeOS Terminal app.

PR #5066, opened 2024-06-03
Speed up clearing all markers, mass event listener dispose

Tyriar:
Part of microsoft/vscode#213174
Part of microsoft/vscode#213304

Test: 1000 scrollback, decorations stress test

Before:

Screenshot 2024-06-03 at 10 59 34 AM

After:

Screenshot 2024-06-03 at 11 42 41 AM

### DIFF
--- a/src/common/EventEmitter.ts
+++ b/src/common/EventEmitter.ts
@@ -20,23 +20,18 @@ export interface IEventEmitter<T, U = void> {
 }
 
 export class EventEmitter<T, U = void> implements IEventEmitter<T, U> {
-  private _listeners: IListener<T, U>[] = [];
+  private _listeners: Set<IListener<T, U>> = new Set();
   private _event?: IEvent<T, U>;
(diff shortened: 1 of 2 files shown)


src/common/EventEmitter.ts
```diff
@@ -20,23 +20,18 @@ export interface IEventEmitter<T, U = void> {
 }
 
 export class EventEmitter<T, U = void> implements IEventEmitter<T, U> {
-  private _listeners: IListener<T, U>[] = [];
+  private _listeners: Set<IListener<T, U>> = new Set();
```
mofux, line 23: @Tyriar Just a remark, I think this could theoretically introduce a bug: if you add the same listener function more than once. The Set will filter out duplicate references, so registering the same listener callback function twice will now only fire the listener once. I don't think this will be a problem in practice because listener callbacks are most likely passed as anonymous callback functions anyway, but yeah. Just saying :D

PR #5077, opened 2024-07-02
Include base/ platform from VS Code, adopt DomScrollableElement and introduce esbuild

Tyriar:
Succeeds https://github.com/xtermjs/xterm.js/pull/5076

Issue #2878, opened 2020-04-27
ES6 module no export Terminal


elgs:
When I follow the doc to import Terminal as an ES6 module.

The recommended way to load xterm.js is via the ES6 module syntax:
```
import { Terminal } from 'xterm';
```

I got the following error. I opened the xterm.js and found it's an minified js file and it indeed didn't export Terminal.

```
The requested module '../../../node_modules/xterm/lib/xterm.js' does not provide an export named 'Terminal'
```

I'm not using any building tools. I just used plain ES6 import syntax.

Tyriar:
I don't know the details here on what TS does but TS works for me. I guess this is requesting a native ESM target which has come up before https://github.com/xtermjs/xterm.js/issues/2486

elgs:
What I am wondering is I've seen a few other libraries claiming they are supporting ES6, but they are in the same situation like this, like this one https://github.com/highlightjs/highlight.js/issues/2473. They use Rollup to build the library. I'm not familiar with Rollup.

My understand of ES6 module is the library has some code syntax like:
```js
export something = someFunction;
```

And the client is like:
```js
import { something } from 'node_modules/someLib/someLib.js';

something();
```

I wonder if I misunderstood what ES6 is.

elgs:
This is a big deal, because with the REAL ES6 support, the client can get rid of the dependency of build tools, though the build tools are awesome, but at least it's not necessary.

Tyriar:
Yeah it would be nice to be on the latest and greatest but then you lose support for other better supported ways of importing. Eventually we'll move there but shipping a library that supports many different import types is tricky.

elgs:
I understand that. That's why if you look at the lodash package, they released a separate package called lodash-es. That is the REAL ES6 support. But unfortunately the export syntax is illegal in the nodejs like module system.

You mentioned:
> but then you lose support for other better supported ways of importing

Actually, no, you don't lose anything in browser. All modern build tools support the REAL ES6 syntax. What you lose is the support to build for nodejs user. But for packages like xterm, and highlightjs, why does it matter at all. Who use them in Nodejs.

elgs:
[...]

PR #5092, opened 2024-07-05
Add support for ESM via esbuild

Tyriar:
This PR includes the esbuild-related parts from https://github.com/xtermjs/xterm.js/pull/5077

This is roughly how the dependencies work currently:

image

The big wins here are:

- ESM support shipped in parallel via the `module` properly and `lib/xterm.mjs` (https://github.com/xtermjs/xterm.js/issues/2878), vscode does not use this yet but will soon hopefully (https://github.com/microsoft/vscode/issues/160416)
- Smaller bundle size for the new core ESM module (285KB -> 253KB)
- Much faster build times for the core lib and _unit_ tests
  - tsc still runs at the same speed to verify the types and for project references, but it no longer blocks the demo file bundling so iteration time is better.
- The demo now points at all the node modules, bringing it closer to what a user of xterm and its addons would do


Tyriar:
Latest runs integration test off esbuild output 🎉

Tyriar:
Ready to merge, this is much less risky then initially as the old tsc-based webpacked output is still shipped, just the ESM output is based on esbuild.

jerch:
@Tyriar Wow, this looks like some heavy lifting to get ESM rolling, thx for doing the nasty works! :rocket: 

This diagram looks important to me - should we link that under the wiki build page? Also to me sense for me - does that mean, that internally all packages are still handled by UMD, but the final packages for delivery are ESM now? Or do we ship now UMD and ESM in the final packages?

Tyriar:
@jerch yeah it was a bit rough, but a lot of this felt necessary in order to bring the nice scroll bars over.

> This diagram looks important to me - should we link that under the wiki build page?

Good idea, I'll clean up the diagram and stick it in contributing.md and/or the wiki.

> Also to make sense for me - does that mean, that internally all packages are still handled by UMD, but the final packages for delivery are ESM now? Or do we ship now UMD and ESM in the final packages?

We ship both which is possible by using the `main` and `module` properties in the package.json. Internally all packages use the esbuild sources and webpack is not touched at all until `yarn package` happens which is where the big perf wins come from as we don't need to wait for webpack to bundle after a change happens.

I wanted `tsc` to use `noEmit` such that we could remove `out/` and then remove `-esbuild` from those folders completely to simplify things, but in order to use project references (`"composite": true` and `"references": {...}`) we must emit the output. esbuild by itself doesn't verify types but instead just strips them out so we must run it.

PR #5103, opened 2024-07-12
remove _workCell in InputHandler.

tisilent:
Trying to add 'no-unused-private-class-members' does not work. :sleepy:


Tyriar (approved):
Thanks 👍

Tyriar:
Is that's the only change needed to add that rule we should add it as a warning (ie. everything should work but no merging without it fixed).

PR #5119, opened 2024-07-28
add class for blinking text control code (\x1b[5m) to DOM renderer

CodeGuro:
My company happened to need the blinking text escape code, so I added it to the DOM renderer.

### DIFF
--- a/src/browser/renderer/dom/DomRendererRowFactory.ts
+++ b/src/browser/renderer/dom/DomRendererRowFactory.ts
@@ -24,6 +24,7 @@ export const enum RowCss {
   UNDERLINE_CLASS = 'xterm-underline',
   OVERLINE_CLASS = 'xterm-overline',
   STRIKETHROUGH_CLASS = 'xterm-strikethrough',
+  BLINK_CLASS = 'xterm-blink',
   CURSOR_CLASS = 'xterm-cursor',
   CURSOR_BLINK_CLASS = 'xterm-cursor-blink',
   CURSOR_STYLE_BLOCK_CLASS = 'xterm-cursor-block',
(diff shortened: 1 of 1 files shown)


src/browser/renderer/dom/DomRendererRowFactory.ts
```diff
@@ -24,6 +24,7 @@ export const enum RowCss {
   UNDERLINE_CLASS = 'xterm-underline',
   OVERLINE_CLASS = 'xterm-overline',
   STRIKETHROUGH_CLASS = 'xterm-strikethrough',
+  BLINK_CLASS = 'xterm-blink',
```
Tyriar, line 27: We'd also want the associated CSS class for this in xterm.css if we went with a class

Tyriar, line 27: We should also add an implementation on the webgl side for this so we're not merging in a partially complete feature. Probably the easiest way to implement this in such a way that would share the implementation across multiple renderers is to leverage the decorations service and have an interval to set the transparency of the decoration on and off periodically:

https://github.com/xtermjs/xterm.js/blob/cb3bd463a5aabda47c135481ba4dbf7ccf8680ab/src/common/services/Services.ts#L363-L368

Tyriar, line 27: I'm not sure what the right default is for this, but regardless for accessibility reasons we will want to allow the embedder to disable blinking text. So I think we need some new option to drive this.

demo/server.js
```diff
@@ -59,13 +59,16 @@ function startServer() {
...
       name: 'xterm-256color',
       cols: cols ?? 80,
       rows: rows ?? 24,
-      cwd: process.platform === 'win32' ? undefined : env.PWD,
+      cwd: isWindows ? undefined : env.PWD,
       env,
-      encoding: USE_BINARY ? null : 'utf8'
+      encoding: USE_BINARY ? null : 'utf8',
+      useConpty: isWindows,
+      useConptyDll: isWindows,
```
Tyriar, line 71: Good change, you can make a separate PR this node-pty stuff if you want to get it merged quickly.

Tyriar (changes requested):

Tyriar:
We may add this at some point but this PR stalled. It might be something that would live behind the new `quirks` option as it's often considered bad for a11y.

PR #5136, opened 2024-08-23
Fix description for WebglAddon class in @xterm/addon-webgl

esimkowitz:
The description mentions search functionality, seems like a copy/paste issue.


Tyriar (approved):
Surprised it said search for so long 😅 Thanks for spotting

Issue #3293, opened 2021-04-05
DECSCUSR() and DECSCUSR(0) should reflect the current user setting of the cursor shape


akinomyoga:
This was first reported at https://github.com/akinomyoga/ble.sh/issues/95. In Visual Studio Code (VSCode), when a Bash plugin `ble.sh` is loaded in a Bash session, the cursor shape becomes always `Block` regardless of the user setting of the cursor shape.

<img width="50%" src="https://user-images.githubusercontent.com/8982192/113637080-b7f9a800-96ae-11eb-9f4f-47ee08a98ae7.png" />

This seems to be caused by the behavior of xterm.js for the control function <kbd>DECSCUSR()</kbd> (*empty*) and <kbd>DECSCUSR(0)</kbd> (*0*). I expect these parameters, *empty* and *0*, change the cursor shape to the default one configured by the user, but, in xterm.js, *empty* and *0* are [always treated as *1* (Blinking Block)](https://github.com/xtermjs/xterm.js/blob/233b78d258417e9d6c107eb54fc4c753e1625911/src/common/InputHandler.ts#L2666) which is just the default value for the user setting.

<img width="50%" src="https://user-images.githubusercontent.com/8982192/113639867-17f34d00-96b5-11eb-9546-7e097ff020d8.png" />

## Behavior of other terminals

- In all the other terminals that I tested (including Mintty, Alacritty, RLogin, GNOME Terminal (VTE), and Windows Terminal), when the user changed the default cursor shape from Block to another, <kbd>DECSCUSR()</kbd> and <kbd>DECSCUSR(0)</kbd> change the cursor shape to the user-configured default rather than to Block. The behavior can be checked with the following commands after the user default is changed.
  ```bash
  $ printf '\e[ q'
  $ printf '\e[0 q'
  $ printf '\e[1 q'
  ```

  I couldn't check xterm's behavior because I don't know how to change the default value for the cursor shape in xterm.

- We can also find a related discussion in [a code comment of VTE](https://github.com/GNOME/vte/blob/f728689e89ef4f6b83527c9a0d340ee52b7d0cf0/src/vteinternal.hh#L174-L185):

  > We treat 0 and 1 differently, assuming that the VT510 does so too.
  >
[...]

PR #5137, opened 2024-08-25
Working on fixing Issue #3293

Arup-Chauhan:
@Tyriar Hi, I was working on #3293 and for that I have done the following things and created this draft PR to discuss.

Here are some changes I made:

1. **Terminal.ts:**
   - I added logic to save the custom cursor style (`userCursorStyle`) when the terminal initializes. This ensures that the terminal keeps my preferred cursor style even when control sequences are sent.

2. **InputHandler.ts:**
   - I adjusted the `setCursorStyle` function to apply `userCursorStyle` when handling the reset cursor style sequence (`\x1b[0 q`). I also included some console log statements to help debug the cursor style changes.

### Possible Issue with Control Sequences:

While testing, I noticed something off with how the control sequences are being processed. Specifically, the output logs don't match the expected behavior.

**Current Log Outputs:**
```
Script is running
User Cursor Style: underline
DOM fully loaded and parsed
Sending control sequence as given 1
User Cursor Style: underline
Current Style: underline
Sending control sequence as given 2
User Cursor Style: underline
Current cursor style: underline
```

**Expected Log Outputs:**
```
Script is running
User Cursor Style: underline
DOM fully loaded and parsed
Sending control sequence as given 1
User Cursor Style: underline
Current Style: bar
Sending control sequence as given 2
User Cursor Style: underline
Current cursor style: underline
```

Am I testing / emulating the xterm terminal properly? 
The current output suggests that the control sequences may not be applied correctly, particularly when resetting the cursor style. This might need a closer look to figure out what's going on.

Help would be appreciated.

### DIFF
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -2734,6 +2734,10 @@ export class InputHandler extends Disposable implements IInputHandler {
   public setCursorStyle(params: IParams): boolean {
     const param = params.params[0] || 1;
     switch (param) {
+      case 0:
+        console.log('Case 0 triggerd');
+        this._optionsService.options.cursorStyle = this._optionsService.options.userCursorStyle || 'block';
+        break;
       case 1:
       case 2:
         this._optionsService.options.cursorStyle = 'block';
--- a/.hintrc
+++ b/.hintrc
@@ -1,0 +1,8 @@
+{
+  "extends": [
+    "development"
+  ],
+  "hints": {
+    "no-inline-styles": "off"
+  }
+}
--- a/src/browser/public/Terminal.ts
+++ b/src/browser/public/Terminal.ts
@@ -35,6 +35,11 @@ export class Terminal extends Disposable implements ITerminalApi {
     this._core = this._register(new TerminalCore(options));
     this._addonManager = this._register(new AddonManager());
 
+    // Store user's cursor style
+    this._core.options.userCursorStyle = this._core.options.cursorStyle;
+
+
+
     this._publicOptions = { ... this._core.options };
     const getter = (propName: string): any => {
       return this._core.options[propName];
--- a/test_file.html
+++ b/test_file.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>xterm.js Test</title>
+    <script src="./node_modules/@xterm/xterm/lib/xterm.js"></script>
+    <link rel="stylesheet" href="./css/xterm.css">
+    <script>
+        console.log('Script is running');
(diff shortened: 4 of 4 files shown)


src/common/InputHandler.ts
```diff
@@ -2734,6 +2734,10 @@ export class InputHandler extends Disposable implements IInputHandler {
   public setCursorStyle(params: IParams): boolean {
     const param = params.params[0] || 1;
     switch (param) {
+      case 0:
+        console.log('Case 0 triggerd');
+        this._optionsService.options.cursorStyle = this._optionsService.options.userCursorStyle || 'block';
+        break;
```
Tyriar, line 2740: I think the approach you're taking it likely to cause problems (maybe what's causing the issue you mention now). Instead of trying to replace `this._optionsService.options.cursorStyle` or add a new `this._optionsService.options.userCursorStyle` to `this._optionsService.options` which is an object created by the user and not meant to be touched by us, we would want some way to setting an override which starts out at undefined. Makes sense to put this override in [`IDecPrivateModes`](https://github.com/Tyriar/xterm.js/blob/cb3bd463a5aabda47c135481ba4dbf7ccf8680ab/src/common/Types.ts#L267-L275) as mentioned in https://github.com/xtermjs/xterm.js/issues/3293#issuecomment-2094590792

Then when we go to use the actual cursor style from the renderer, we would use `devPrivateModes.cursorStyle ?? this._optionsService.options.cursorStyle ?? 'block'`

Tyriar (changes requested):

src/common/InputHandler.ts
Arup-Chauhan: > I think the approach you're taking it likely to cause problems (maybe what's causing the issue you mention now). Instead of trying to replace `this._optionsService.options.cursorStyle` or add a new `this._optionsService.options.userCursorStyle` to `this._optionsService.options` which is an object created by the user and not meant to be touched by us, we would want some way to setting an override which starts out at undefined. Makes sense to put this override in [`IDecPrivateModes`](https://github.com/Tyriar/xterm.js/blob/cb3bd463a5aabda47c135481ba4dbf7ccf8680ab/src/common/Types.ts#L267-L275) as mentioned in [#3293 (comment)](https://github.com/xtermjs/xterm.js/issues/3293#issuecomment-2094590792)
> 
> Then when we go to use the actual cursor style from the renderer, we would use `devPrivateModes.cursorStyle ?? this._optionsService.options.cursorStyle ?? 'block'`


That makes sense, I will do this and get back to you, give me a while, thanks!

Arup-Chauhan: Hi @Tyriar, wanted to update you that I got some academic commitments mid-way so paused work on this.
I am continuing working on this, will keep you updated.

Once again, thanks for your support!

Tyriar:
Closing off as the approach was wrong and author is busy with other stuff atm.

PR #5180, opened 2024-10-04
Forward integration test exit code to running proc

### LINKED ISSUE #5179: Integration tests are failing but the action isn't reporting failure
eg. https://github.com/xtermjs/xterm.js/actions/runs/10963776473/job/30445954398

Tyriar:
Fixes #5179


jerch:
@Tyriar Was able to fix the dispose error. The other errors are all timeout errors and happen at different occasions. Locally all runs fine, idk why the integration tests are this flaky in the CI env. It is really annoying if the CI is that unreliable.

Issue #4603, opened 2023-07-21
reset() not finished on sync return


jerch:
coming from https://github.com/xtermjs/xterm.js/discussions/4319#discussioncomment-6508416 - the `Terminal.reset` call does not finish the re-initialization synchronous.

possible fixes:
Either make it a promise (rather a quickfix and with ugly API change), or revisit the code paths, that are not ready yet when returning synchronously from the method and execute them in sync.

Tyriar:
This is surprising, it was always meant to be sync so hopefully we can fix it.

Tyriar:
clear too? 😕 https://github.com/xtermjs/xterm.js/discussions/4319#discussioncomment-6510734

This isn't just a re-rendering problem is it?

jerch:
> This isn't just a re-rendering problem is it?

Haven't looked into it yet, I'd say yes - its a re-rendering issue with false scrollOffset from viewport.ts (prolly due to using an animation frame there, its just a guess though atm).

Tyriar:
Repro: run this in console `term.reset(); term.write('1\n2\n3\n\r'.repeat(20));`

PR #5182, opened 2024-10-04
fix #5181

### LINKED ISSUE #5181: race condition in terminal.dispose
image

Repro:
- open demo in browser
- call `term.dispose()`

The culprit seems to be the disposal of the webgl addon:
https://github.com/xtermjs/xterm.js/blob/9186c724c484d0a46be3e2c79d1fdb32440f4bb4/addons/addon-webgl/src/WebglAddon.ts#L90-L94

Those lines recreate & attach a DOM renderer instance, which fails, if we are in the middle of terminal disposal. Imho it can be fixed by:
```typescript
this._register(toDisposable(() => {
  if ((this._terminal as any)._core._store._isDisposed) {
    return;
  }
  const renderService: IRenderService = (this._terminal as any)._core._renderService;
  renderService.setRenderer((this._terminal as any)._core._createRenderer());
  renderService.handleResize(terminal.cols, terminal.rows);
}));
```

jerch:
See #5181.

### DIFF
--- a/addons/addon-image/test/ImageAddon.test.ts
+++ b/addons/addon-image/test/ImageAddon.test.ts
@@ -163,35 +163,28 @@ test.describe('ImageAddon', () => {
     test('testdata default (scrolling with VT240 cursor pos)', async () => {
       const dim = await getDimensions();
       await ctx.proxy.write(SIXEL_SEQ_0);
-      await timeout(50);
       deepStrictEqual(await getCursor(), [0, Math.floor(TESTDATA.height/dim.cellHeight)]);
       // moved to right by 10 cells
       await ctx.proxy.write('#'.repeat(10) + SIXEL_SEQ_0);
-      await timeout(50);
       deepStrictEqual(await getCursor(), [10, Math.floor(TESTDATA.height/dim.cellHeight) * 2]);
     });
     test('write testdata noScrolling', async () => {
       await ctx.proxy.write('\x1b[?80h' + SIXEL_SEQ_0);
-      await timeout(50);
       deepStrictEqual(await getCursor(), [0, 0]);
       // second draw does not change anything
       await ctx.proxy.write(SIXEL_SEQ_0);
-      await timeout(50);
       deepStrictEqual(await getCursor(), [0, 0]);
     });
     test('testdata cursor always at VT240 pos', async () => {
       const dim = await getDimensions();
       // offset 0
       await ctx.proxy.write(SIXEL_SEQ_0);
-      await timeout(50);
       deepStrictEqual(await getCursor(), [0, Math.floor(TESTDATA.height/dim.cellHeight)]);
       // moved to right by 10 cells
       await ctx.proxy.write('#'.repeat(10) + SIXEL_SEQ_0);
-      await timeout(50);
       deepStrictEqual(await getCursor(), [10, Math.floor(TESTDATA.height/dim.cellHeight) * 2]);
       // moved by 30 cells (+10 prev)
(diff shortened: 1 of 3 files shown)


addons/addon-image/test/ImageAddon.test.ts
```diff
@@ -163,43 +163,51 @@ test.describe('ImageAddon', () => {
     test('testdata default (scrolling with VT240 cursor pos)', async () => {
       const dim = await getDimensions();
       await ctx.proxy.write(SIXEL_SEQ_0);
+      await timeout(50);
```
Tyriar, line 166: The await on write should be fine, no? If not, is there something else we could await that isn't an arbitrary time?

jerch: Sadly nope - we expose no signal/event like thingy, that would correctly mark the end of terminal processing up to the renderer.

My other guess regaring those flaky tests is about `term.reset`, which gets called over and over in  test modules (fixed for web-links tests in the other PR #5180). I am not sure, if this call is fully synchronous now (we even had an issues regaring this in the past) - if not, then heavy load on the CI machine might cause those time offsets with failing tests. Also just a wild guess atm, but with the fix the tests passed.

Tyriar: Oh right, these need rendering to have occurred. This is internal, but you could debounce listen on `IRenderService.onRender` which is what you're actually after. Alternatively I think we have a pollFor helper where we can set a much larger timeout without impacting test duration to prevent flakiness

I'm not aware of any `reset` not being sync problems.

jerch: Yes some of them need the actual rendering result, which is a bit cumbersome to follow due to our async writing/rendering. For those tests it is clear, that a busy CI machine will take very long eventually running into a timeout. Thats also the case for weblink tests, as they eval the DOM renderer results.
I will see if I can find a more reliable / deterministic way to wait for render result with `onRender`.

The issue with `reset` was brought up in #4603, and your fix kinda addressed a bad interim state of the viewport in sync code. To me this had similarities to the infamous `nextTick` issue often seen in nodejs tasks, that need eventloop+1 invocations to fully finish.
And sometimes tests on a busy CI machine fail so randomly with a weird error that I started to wonder, if we have such an eventloop+1 issue hidden anywhere. Also I see those types of failures only in integration tests, which made we wonder if terminal life cycling could cause this. Unit tests dont show those type of failures, they kinda only fail with very high timeout on really busy machines (means we got not enough CPU in time).
Ofc integration/playwright tests are much more heavy with their browser engines in the test stack, so it is not clear, if our (test) code has a hidden issue not surfacing on idle machines, or the browser engine simply stopped working somewhere in between.

jerch:
Also gonna revert this change [5623ba6](https://github.com/xtermjs/xterm.js/pull/5180/commits/5623ba6b6368392e76bde90400cd31eae0661bfb) (made in #5180). Sure it helped yesterday, but there is really an issue with lifecycling or eventloop+1 in `term.reset`, it would just hide that.

jerch:
Eww, the errors are back in the web links tests. This really s*x, as I cannot repro locally.

jerch:
@Tyriar Since weblink tests now pass with a modified fix, lets keep them for now. Imho we should investigate that separately.

Tyriar (approved):

viko16:
This is a great bug fix. Why was it merged but a new version hasn't been released yet?

Tyriar:
@viko16 yep last stable publish was over a year ago. You can use the beta build in the meantime to get on the latest

PR #5209, opened 2024-11-04
Ensure last ligature cell is updated

### LINKED ISSUE #3288: WebGL ligatures are storing wrong color for cells, causing bad rendering sometimes when type ahead is used in VS Code
Screen Shot 2021-04-02 at 3 31 25 PM

### LINKED ISSUE #5206: Ligatures background color can flash the wrong color when moving to different lines
Image

Likely the same root cause as https://github.com/xtermjs/xterm.js/issues/5205

Tyriar:
Fixes #3288
Fixes #5206
Part of microsoft/vscode#233005

### DIFF
--- a/addons/addon-webgl/src/WebglRenderer.ts
+++ b/addons/addon-webgl/src/WebglRenderer.ts
@@ -314,14 +314,6 @@ export class WebglRenderer extends Disposable implements IRenderer {
     this._updateCursorBlink();
   }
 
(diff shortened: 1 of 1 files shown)


npezza93:
This resolves a bunch of issues i was seeing with ligatures but there still seems to be one lingering issue. (notice the extra character over the > in => that shouldnt be there and the d in ::)
image

Issue #3025, opened 2020-07-31
Terminal doesn't respond to keypresses


yvantitov:
I can't seem to get input from a `Terminal` when using xterm.js in a html page. I apologize in advance if this is a silly error on my end, but I cannot figure out what I did wrong.
## Details

- Browser and browser version: Google Chrome | 84.0.4147.89 (Official Build) (64-bit)
- OS version: Ubuntu 20.04 LTS x86_64
- xterm.js version: 4.8.1

### Steps to reproduce

1. Put this into a file:
```
<!DOCTYPE html>
<html>
  <head>
   <meta charset="UTF-8">
   <title>Broken :(</title>
   <link rel="stylesheet" href="node_modules/xterm/css/xterm.css" />
   <script src="node_modules/xterm/lib/xterm.js"></script>
  </head>

  <body>
   <div id="terminal"></div>

   <script type = "text/javascript">
      var term = new Terminal();
      term.open(document.getElementById('terminal'));
      term.writeln('Hello World!');
      term.onKey = (key, ev) => {
          console.log(`Key entered: ${key}`);
      }
   </script>
  </body>
</html>
```
2. Open it in browser (same behavior also observed when using a basic HTTP server).
3. There is a properly styled terminal but I cannot type in it. Nothing happens visually. There is also no output in the console.

Tyriar:
> term.onKey = (key, ev) => {

term.onKey is a function:

```
term.onKey((key, ev) => {

});
```

Also you should typically be using `onData` instead.

yvantitov:
Working now. Thanks!

anatoly314:
> > term.onKey = (key, ev) => {
> 
> term.onKey is a function:
> 
> ```
> term.onKey((key, ev) => {
> 
> });
> ```
> 
> Also you should typically be using `onData` instead.

So, what the difference between `onKey` and `onData`? I have a similar experience where `onKey` not fired on Android's Chrome browser (except `enter`, 'backspace`) neither `onData` till I type something like `$#@!}|` then `onData` fired will all the chars been typed till now but `onKey` still keeps silence. Doubt is it a known bug or I should report it?

Tyriar:
`onKey` fires during the keypress/keydown events when `onData` should also fire (unless `disableStdin` is on), it contains the DOM event. `onData` is the keyevent serialized such that it can be passed to a pty.

I can't repro what you're seeing. Backspace, $ and | are pressed below in the demo:

[...]

PR #5221, opened 2024-11-20
AccessibilityManager's handleKey should be triggered by onData, not onKey

ikicha:
Issue:
In Android, screen reader feature sometimes generates the voice output including previous content

Reason:
`_liveRegion.textContent` from previous command is supposed be cleaned up in key stroke event, but onKey doesn't work as intended with IME(except 'enter' in Android) as mentioned in #3025. So the voice output includes previous content.

Fix:
As @Tyriars said in #3025, use `onData` instead of `onKey`

### DIFF
--- a/src/browser/AccessibilityManager.ts
+++ b/src/browser/AccessibilityManager.ts
@@ -112,7 +112,7 @@ export class AccessibilityManager extends Disposable {
     this._register(this._terminal.onA11yChar(char => this._handleChar(char)));
     this._register(this._terminal.onLineFeed(() => this._handleChar('\n')));
     this._register(this._terminal.onA11yTab(spaceCount => this._handleTab(spaceCount)));
-    this._register(this._terminal.onKey(e => this._handleKey(e.key)));
+    this._register(this._terminal.onData(e => this._handleKey(e)));
     this._register(this._terminal.onBlur(() => this._clearLiveRegion()));
     this._register(this._renderService.onDimensionsChange(() => this._refreshRowsDimensions()));
     this._register(addDisposableListener(doc, 'selectionchange', () => this._handleSelectionChange()));


ikicha:
@Tyriar @jerch , can you review this CL?

ikicha:
@Tyriar ping..

jerch:
@ikicha Does that still work the same way? Plz note, that `onData` is special, as it outputs PTY-ready data, which is not necessarily from user input.

Sorry I cannot really review this, as I literally have no clue about accessibility's best practices.

Tyriar:
We can't use `onData` here as it is not necessarily user input would therefore make the `_charsToConsume` part fall over:

https://github.com/anouartouati/xterm.js/blob/c71ef1b91a736db5fad024600f0303809215052d/src/browser/AccessibilityManager.ts#L41-L50

What this change also does is run `_handleKey` after each and every change to the terminal buffer.

> As @Tyriars said in https://github.com/xtermjs/xterm.js/issues/3025, use onData instead of onKey

This was general advice to embedders of the lib.

---

Closing since we can't accept this and a different fix would likely look pretty different.

ikicha:
Can you suggest a possible fix for that? The problem is onKey isn't called when key input comes from soft keyboard.

jerch:
Whats the general handling of input output systems by accessibility software? With a terminal we have the following IO issue:
- INPUT: user presses a key, that genuine user input
- OUTPUT: terminal echos that user input in its output

Is it normal to be mirrored on both sides, thus twice? If yes then onKey/onData is the right track. The problem remains, that `onData` is a sythesized string suitable for PTY ingestion, which may already contain control sequences even from user input. So we already lost the key-press connection at that point and the data might be in a non-undetstandable format already. And furthermore, as @Tyriar already wrote, `onData` may contain input from other sources (like clipboard pastes or programmaically places terminal input) - also not a good idea to mirror that 1:1 to accessibility software.

Imho to fix this, we might need an API in middle specially crafted for accessibility, which can still distinguish between keypresses, clipboard pastes etc. and has not yet PTY-transformations applied on the data.

Tyriar:
Perhaps listening to `compositionupdate`? This is an event that comes from IMEs and fires when you type on PC. Similar to this:

https://github.com/xtermjs/xterm.js/blob/33a5457407949a4aab9e4662015436c427cefe5a/src/browser/CoreBrowserTerminal.ts#L381-L383

ikicha:
re comment

> Perhaps listening to `compositionupdate`? This is an event that comes from IMEs and fires when you type on PC. Similar to this:
> 
> https://github.com/xtermjs/xterm.js/blob/33a5457407949a4aab9e4662015436c427cefe5a/src/browser/CoreBrowserTerminal.ts#L381-L383

Thanks for the advice.

Did you mean by calling `_handleKey` when compositionupdate is fired as well as onKey? or maybe only compositionupdate?

Tyriar:
I _think_ you want to call `_handleKey` on `compositionend` only. That would add the characters to the `_charsToConsume` queue, and hopefully prevent them from being read by the screen reader twice which is what that mechanism is for.

Issue #1727, opened 2018-10-05
Atypical "Erase In Display" behaviour


pfitzseb:
Most terminals seem to implement ED (specifically `\x1b[2J`) by not actually deleting anything from the display buffer, but rather by scrolling. Conversely, xterm.js does delete the display buffer's contents, which is a) inconsistent and b) annoying because you lose output. I don't think this behaviour qualifies as a bug because the spec doesn't say about what should happen with the scrollback buffer, but would still be nice if this were changed.

Downstream issue in https://github.com/JunoLab/atom-julia-client/issues/521.

## Details
- Browser and browser version: Electron 2.0.8
- OS version: Manjaro 
- xterm.js version: master (but also at least 3.6 and 3.7)

### Steps to reproduce
Try 
```bash
for i in {1..20}; do
  echo "$i"
done
echo '\e[1;1H\e[2J'
```
in xterm.js and another terminal emulator, and scroll up afterwards.

![image](https://user-images.githubusercontent.com/6735977/46527796-721bb900-c892-11e8-833d-58c15ef00477.png)
Expected behaviour is what e.g. Konsole does (middle window), and not how e.g. VSCode (left) or Juno's terminal (right, with xterm.js master) handles this.

Tyriar:
Sounds good, Terminal.app and iterm2 also show this behavior.

egmontkob:
[VTE](https://bugzilla.gnome.org/show_bug.cgi?id=506438) and [Konsole](https://bugs.kde.org/show_bug.cgi?id=384218) also behave as suggested here, and have bugreports to revert to xterm's behavior of deleting those lines :D

The main problem is: if someone's intent is really to clear the entire scrollback (which the `clear` command supposedly does), it no longer works completely, leaves a screenful of content in the scrollback because `clear` emits the sequences in the wrong order – well, at least wrong order for these kinds of terminal emulators.

It's either a downside here or a downside there, you pick :)

Tyriar:
@egmontkob I thought clear typically sent a different sequence to clear the scrollback as well?

seeruk:
Really keen to see this one resolved, it's the only thing holding me back from using terminal emulators like Hyper. I use Ctrl+L a lot to get rid of noise on my screen, but often want to be able to go back and see what was output. That includes things like testing. Personally, I don't see any reason not to work like iTerm2, VTE terminals, Terminal.app, etc. - there could be a separate way to clear the whole scrollback, or just push what's on screen up.

[...]

PR #5224, opened 2024-11-21
putty-style ED2 sequence handling as terminal option

Blashaq:
This PR is related to the problem discussed in https://github.com/xtermjs/xterm.js/issues/1727.

goals: 
- Allow xterm.js based applications to change CSI 2 J sequence behaviour to non-canon, which is sometimes preferred by users. 
- Keep canon handling of CSI 2 J (clear whole screen) as default.
- Alternative handling will push erased text to the scrollback and trim empty lines. This is the way PuTTY and some other terminals do it by default.

changes made:
- New configuration option: `scrollOnDisplayErase` in `ITerminalOptions`.
- InputHandler.eraseInDisplay will behave differently if this option is set.
- Added some tests.

Fixes https://github.com/xtermjs/xterm.js/issues/1727


jerch (commented):
Thx for looking into this.

Beside the minor code suggestion below - LGTM :+1:

Blashaq:
Thanks for suggestions, this is indeed cleaner (and also works, to no suprise :octocat:  ).

jerch:
@Blashaq The CI tests dont run yet, seems `scrollOnDisplayErase` is missing in src/common/services/Services.ts in ITerminalOptions?

Blashaq:
You're right, sorry. I have accidentally skipped that file in my commit.

jerch:
There are still linter some issues.

Blashaq:
Everything should be fine now 🙏

jerch (approved):
@Blashaq LGTM :+1:, many thx for the PR.

@Tyriar There seems to be an issue with `clear` under with linux with ED3 - the new scrollbar gets not updated by ED3 yet. Maybe this not enough to announce the changes: https://github.com/xtermjs/xterm.js/blob/41e8ae395937011d6bf6c7cb618b851791aed395/src/common/InputHandler.ts#L1238

Imho its not linked to this PR, thus I create a new ticket. --> #5228

Tyriar (approved):
LGTM 👏 sorry about the delay

Issue #5220, opened 2024-11-20
Terminal eats `click` events


jtbandes:
When adding `mousedown`, `mouseup`, and `click` listeners on the terminal container element, and then clicking on the terminal, I observe that the `click` is sometimes not delivered. (On macOS with a Magic Trackpad, I noticed that the `click` event almost always works when using [tap to click](https://support.apple.com/guide/mac-help/change-trackpad-settings-mchlp1226/mac) and tapping with a single finger, but when actually _pressing_ the trackpad to click it, the `click` event is usually not delivered.)

I suspect this is a clue: right-clicking on the logged `event.target` and clicking "Reveal in Elements panel" sometimes shows "Node cannot be found in the current page". Perhaps the browser is not delivering the `click` event because the mousedown/mouseup are not on matching elements, or the clicked element is removed before the click finishes.

<img width="241" alt="image" src="https://github.com/user-attachments/assets/9a6527b2-358a-46f0-a66f-e2595b03605d">
<img width="721" alt="image" src="https://github.com/user-attachments/assets/107ec28c-da8e-49c9-8920-b2a10ec1f9c1">

https://github.com/user-attachments/assets/8b90e791-8cc7-485e-8ec2-fd472db9ae84

## Details
- Browser and browser version: Chrome 131.0.6778.70
- OS version: macOS 15.1
- xterm.js version: 5.5.0

Note: same behavior seems to happen with 5.6.0 beta as well: https://codesandbox.io/p/sandbox/xtermjs-test-forked-qfng37

### Steps to reproduce

https://codesandbox.io/p/sandbox/xtermjs-test-forked-c5s8g4
Or visit https://c5s8g4.csb.app/ and view the console

```ts
import "./styles.css";
import "@xterm/xterm/css/xterm.css";
import { Terminal } from "@xterm/xterm";
import { FitAddon } from "@xterm/addon-fit";

const container = document.getElementById("app");

const terminal = new Terminal({
  cursorStyle: "bar",
  allowProposedApi: true,
});

const fitAddon = new FitAddon();
terminal.loadAddon(fitAddon);

terminal.open(container);

for (let i = 0; i < 100; i++) {
  if (i > 0) {
    terminal.write("\r\n");
  }
  terminal.write(
    `line ${i}: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua`
  );
}

[...]

PR #5249, opened 2024-12-13
Fix click event bug caused by DomRenderer replaceChildren behavior

jtbandes:
Fixes an issue where `click` events on the terminal would not be sent when using DomRenderer. (Fixes https://github.com/xtermjs/xterm.js/issues/5220)

As described in https://github.com/xtermjs/xterm.js/issues/5220#issuecomment-2524541815, as far as I can tell, the issue happens because the clicked element (`event.target`) is replaced (via `replaceChildren`) between mousedown & mouseup. This caused the browser to never deliver a `click` event at all. Adding `pointer-events: none;` makes it so the descendants of `.xterm-rows` are not considered, and the `event.target` is always the parent `.xterm-screen` element for both mousedown & mouseup events — and the click events are delivered as expected.

### DIFF
--- a/src/browser/renderer/dom/DomRenderer.ts
+++ b/src/browser/renderer/dom/DomRenderer.ts
@@ -161,6 +161,10 @@ export class DomRenderer extends Disposable implements IRenderer {
     // Base CSS
     let styles =
       `${this._terminalSelector} .${ROW_CONTAINER_CLASS} {` +
+      // Disabling pointer events circumvents a browser behavior that prevents `click` events from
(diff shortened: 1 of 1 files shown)


jerch:
@jtbandes Wow thats a very neat trick to get it working. Thx for finding and fixing this :)

jtbandes:
@Tyriar friendly ping on this when you get a chance 🙂

Tyriar (approved):
LGTM, I thought this would work, great suggestion!

PR #5251, opened 2024-12-15
progress-addon

### LINKED ISSUE #5250: ConEmu OSC progressbar support
Windows Terminal and VTE added support for a ConEmu specific OSC sequence to display process information on the taskbar button. This led the systemd folks to add it also into systemd command output for long running tasks.

We might want to offer an interface to also handle information from these ConEmu specific sequences, so integrators can work/implement their way to present that information to users. Imho this is not worth to be added to core, but maybe as an addon (maybe even considering several of the other ConEmu subsequences).

References:
- https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC
- https://github.com/microsoft/terminal/pull/8055
- https://mastodon.social/@pid_eins/113406672373007116
- https://gitlab.gnome.org/GNOME/vte/-/issues/2845

jerch:
This addon implements ConEmu's progress sequence, for more context see #5250.

The addon tries to hide most of the sequence's shenanigans behind a convenient API:
```typescript
const progressAddon = new ProgressAddon();
terminal.loadAddon(progressAddon);
progressAddon.register((state, value) => {
  // state: 0-4 integer
  // value: 0-100 integer
  
  // do your visualisation based on state/progress here
  ...
});
```

Fixes #5250.


addons/addon-progress/typings/addon-progress.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    /** register progress handler */
+    public register(handler: ProgressHandler): IDisposable;
+
+    /** getter / setter for current progress */
+    public progress: IProgress;
+  }
+  
+  /** progress object interface */
+  export interface IProgress {
+    state: 0 | 1 | 2 | 3 | 4;
```
Tyriar, line 24: Could pull this into `type ProgressState`?

addons/addon-progress/typings/addon-progress.d.ts
```diff
@@ -0,0 +1,30 @@
...
+
+declare module '@xterm/addon-progress' {
+  /** xterm.js addon providing an interface for ConEmu's progress sequence */
+  export class ProgressAddon implements ITerminalAddon {
+    constructor();
+    public activate(terminal: Terminal): void;
+    public dispose(): void;
+
+    /** register progress handler */
+    public register(handler: ProgressHandler): IDisposable;
```
Tyriar, line 16: Should this be an event instead like this?

https://github.com/xtermjs/xterm.js/blob/41e8ae395937011d6bf6c7cb618b851791aed395/addons/addon-webgl/typings/addon-webgl.d.ts#L20-L23

addons/addon-progress/typings/addon-progress.d.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    /** getter / setter for current progress */
+    public progress: IProgress;
```
Tyriar, line 19: Should this not be readonly?

Tyriar (changes requested):
Really solid 👏 

Depending on app support, this could be really nice to integrate with the VS Code tab progress:

image

addons/addon-progress/typings/addon-progress.d.ts
jerch: Yes would be better. Is there an easy way to that in an addon w'o pulling a whole bunch of code deps into it?

addons/addon-progress/typings/addon-progress.d.ts
Tyriar: I guess set is fine as it means we can restore state if needed like when reloading a window in VS Code 👌

addons/addon-progress/typings/addon-progress.d.ts
jerch: @Tyriar Changed it to an event, which bloats the addon from ~2kB to ~50kB :scream:

I think all slightly more complex addons have the same issue, which is really nasty for web terminals. Should we expose stuff from /vs/* partially via the official API to get rid of doubled impls?

jerch: Thats what gets pulled in:
```
asset addon-progress.js 41.8 KiB [emitted] [minimized] (name: main) 1 related asset
./out/ProgressAddon.js 2.87 KiB [built] [code generated]
../../out/vs/base/common/event.js 50.6 KiB [built] [code generated]
../../out/vs/base/common/lifecycle.js 21.1 KiB [built] [code generated]
../../out/vs/base/common/errors.js 7.6 KiB [built] [code generated]
../../out/vs/base/common/functional.js 1.09 KiB [built] [code generated]
../../out/vs/base/common/arrays.js 25.6 KiB [built] [code generated]
../../out/vs/base/common/collections.js 3.16 KiB [built] [code generated]
../../out/vs/base/common/map.js 4.3 KiB [built] [code generated]
../../out/vs/base/common/iterator.js 4.46 KiB [built] [code generated]
../../out/vs/base/common/linkedList.js 3.66 KiB [built] [code generated]
../../out/vs/base/common/stopwatch.js 1.23 KiB [built] [code generated]
../../out/vs/base/common/arraysFind.js 6.72 KiB [built] [code generated]
```

On code side I only need these imports:
```typescript
import { Emitter } from 'vs/base/common/event';
import { Disposable, toDisposable } from 'vs/base/common/lifecycle';
```

addons/addon-progress/typings/addon-progress.d.ts
jerch: Indeed. While that did not came to my mind, it is a strong reason to keep this way.

jerch:
@Tyriar applied your remarks, up for review again.

jerch:
@Tyriar Found a way to borrow the emitter ctor from xterm until https://github.com/xtermjs/xterm.js/issues/5283 is resolved. With this change things are back to normal with only 1.4kB minified :muscle:  :
```
asset addon-progress.js 1.4 KiB [compared for emit] [minimized] (name: main) 1 related asset
./out/ProgressAddon.js 2.25 KiB [built] [code generated]
```

addons/addon-progress/typings/addon-progress.d.ts
Tyriar: We shouldn't be referencing src from typing files. The more isolated the better as otherwise it can cause huge problems since libraries will suddenly be looking at and maybe linting against our source code, not just the types.

addons/addon-progress/README.md
```diff
@@ -0,0 +1,80 @@
...
+The addon exposes the following API endpoints:
+- `public readonly onChange: IEvent<IProgress>` \
+  Event to register your actual progress handler, where you gonna do the progress visualisation.
+  The handler will get called upon valid progress sequences with a progress argument as `({state, value}) => {}`.
+- `public progress: IProgress`
+  A getter/setter for the current progress information. Can be used to read the last seen progress information.
+  This can also be used to clean up stuck progress indicators by setting the value back to initial, e.g.: 
+  ```typescript
+  progressAddon.progress = {state: 0, value: 0};
+  ```
```
Tyriar, line 80: Best to just defer to the .d.ts for API info, less susceptible to bitrot. Also we can add `@example` comments if needed:

https://github.com/Tyriar/xterm.js/blob/408d0995dddbe7b8e078cc68fb97d3f6166df7d0/addons/addon-webgl/README.md#L22

addons/addon-progress/src/ProgressAddon.ts
```diff
@@ -0,0 +1,101 @@
...
+          break;
+        case ProgressState.INDETERMINATE:
+          this.progress = { state: st, value: this._pr };
+          break;
+      }
+      return true;
+    });
+    // FIXME: borrow emitter ctor from xterm, to be changed once #5283 is resolved
+    this._onChange = new (terminal as any)._core._onData.constructor();
+    this.onChange = this._onChange!.event;
```
Tyriar, line 85: Don' think the `!` is needed?

Tyriar (changes requested):
Looks great, just some small polish things

addons/addon-progress/typings/addon-progress.d.ts
jerch: Hmm yeah, good point. Gonna try to find a better workaround. Sadly enums are tricky to use with d.ts files.

addons/addon-progress/README.md
jerch: Agreed. I initially added this section because of the unconventional `register` handler stuff. Since its using the event type now, which is the preferred code pattern with xterm.js anyway, things are more self-explaining. Gonna remove the section.

addons/addon-progress/src/ProgressAddon.ts
jerch: Doesnt work without. My guess - the any casting above keeps the `undefined` as part of the type, so TS doesnt see this as not undefined-guarded here.

addons/addon-progress/typings/addon-progress.d.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+  
+  /** progress object interface */
+  export interface IProgress {
+    state: ProgressState;
+    value: number;
+  }
```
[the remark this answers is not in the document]
jerch: Renamed it.

addons/addon-progress/typings/addon-progress.d.ts
jerch: See my comment below.

addons/addon-progress/typings/addon-progress.d.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Copyright (c) 2024 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { Terminal, ITerminalAddon, IDisposable } from '@xterm/xterm';
+import type { Event } from 'vs/base/common/event';
+import type { ProgressState } from '../src/ProgressAddon';
```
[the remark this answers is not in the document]
jerch: Removed.

addons/addon-progress/typings/addon-progress.d.ts
```diff
@@ -0,0 +1,54 @@
...
+     * or to restore an indicator.
+     */
+    public progress: IProgressState;
+  }
+  
+  /**
+   * Progress state tracked by the addon.
+   */
+  export interface IProgressState {
+    state: 0 | 1 | 2 | 3 | 4;
```
jerch, line 51: Tried using a const enum inside the d.ts - this worked fine for TS itself and webpack build, but not with esbuild. So I changed it back to that integer union thingy.

Tyriar: That's fine 👌

addons/addon-progress/src/ProgressAddon.ts
Tyriar: All good, this is temporary

Tyriar (approved):
👏 

Will try adopt in vscode now https://github.com/microsoft/vscode/issues/237564

Tyriar:
FYI in case you didn't see this https://github.com/kovidgoyal/kitty/issues/8011#issuecomment-2452009076

Tyriar:
I see winget uses indeterminate progress, do you know of another application that implements value/% based?

jerch:
> FYI in case you didn't see this [kovidgoyal/kitty#8011 (comment)](https://github.com/kovidgoyal/kitty/issues/8011#issuecomment-2452009076)

Yeah, thats the reason why I had to exclude handle of OSC 9 that dont continue with `;4`. There is a better more advanced sepc for notifications than the old OSC 9 stuff, imho it got moved to OSC 99 with more advanced bells&whistles.


> I see winget uses indeterminate progress, do you know of another application that implements value/% based? ([cargo coming soon](https://github.com/rust-lang/cargo/pull/14615))

Nope, I dont. Systemd prolly uses that, but no clue in windows world.

Tyriar:
winget's usage is a little unexpected since it seems to break up subtasks:

image

Not how I would have done it. I guess the progress state change should be debounced a little on my end to avoid a spinner flickering.

jerch:
@Tyriar Yeah, debouncing might be needed. The sequence isnt all that good in its spec - it is also unclear, how long to show the progress indicator or when to remove it (I'd guess that it should be removed after some time of inactivity, or after a window focus toggle etc)

Tyriar:
In my implementation I considered clearing the state after a command is run. I guess it's kind of a necessity if the proc crashes/ctrl+c doesn't handle it, etc.

jerch:
I guess it needs a timeout running down, as we never know for sure when the terminal client app crashed or gave up for whatever reason.

Tyriar:
I have it clearing when shell integration tells us any command finished: https://github.com/microsoft/vscode/pull/237573/commits/86b257ea82bca74e1fd719f952f0a9eee658fedc

Seems like a fairly safe thing to do

PR #5260, opened 2024-12-20
Blend cursor with background to support alpha in webgl

### LINKED ISSUE #5241: Theme cursor alpha is being ignored for GPU renderer
Repro in VS Code:

```js
"workbench.colorCustomizations": {
    "terminalCursor.foreground": "#ffffff25",
  }
```

GPU:

Image

DOM:

Image

Tyriar:
Fixes #5241

### DIFF
--- a/demo/client.ts
+++ b/demo/client.ts
@@ -16,7 +16,7 @@ if ('WebAssembly' in window) {
   ImageAddon = imageAddon.ImageAddon;
 }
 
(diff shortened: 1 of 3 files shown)


AnouarTouati:
I think you forgot to blend the cursor accent.

Tyriar:
Just checked and you're right that did still differ between dom and webgl 👍 making another change now

PR #5272, opened 2025-01-06
WIP: Trying to fix #5176

### LINKED ISSUE #5176: Search is too slow
Currently the search addon limits itself to 1000 highlights by default. Contrast this to monaco's 20000 results before it gives up highlighting. When we increase our limit to 20000 the search itself ends up blocking the renderer pretty significantly.

Here is a profile using my i7-12700KF 3.61 GHz:

Image

Image

Note that the search is also not incremental, so all successive searches take the same amount of time even for non-regex searches.

VS Code issue: https://github.com/microsoft/vscode/issues/230365

AnouarTouati:
This draft is WIP  attempt to fix #5176 

- Search is done in chunks , in alternating manner up/down so the region the user sees first will be done first.
- Decorators/markers 's disposal is also done in chunks in the background(a new search can start without waiting for this to finish).
- Added debounce to the search function.
- Search is now cancel-able 

This is obviously a breaking change.
I tried to go with the conservative approach to keep the code changes to a minimum. But, I had to re-write most of the code.

If this is the direction you want the code to go in. I can keep working on this and maintaining the search addon from now on.
Also ignore the commits I was experimenting with stuff. Will squash them later.

https://github.com/user-attachments/assets/487bc6ec-028c-42e8-9ccf-7169324a5da6


Tyriar:
Adding responses to https://github.com/xtermjs/xterm.js/issues/5176#issuecomment-2563395791 here.

> Why does the cache have TTL ?

This is to aid in incremental searches 

> I experimented with making the search return a small set. release the main thread to render the result and then go and get the full result. i did this with a setTimeout. It helped a lot with the UX.

Something we want to keep is the highlighting of results in the "overview ruler":

Image

Limiting these results too heavily will hurt the UX by not showing additional results. Make sure you add the overview ruler in the demo when working on this:

Image

> I noticed that "incremental" search only serves to do the following :
> Suppose you have A B D C D in the terminal
> Search for C to select it
> Then search for D and it selects the second D.
> Is this intentional ?

If I'm understanding you correctly, searching for `CD` second should only check if there is a `D` after the already existing `C`:

image

This is by design, incremental search means we can avoid doing much of the work, provided regex is not enabled which I think disables incremental completely.

> Maybe another solution is to count the matches but only apply the decorators to the visible lines and a few hidden ones on top and bottom ?

The overview ruler is again an important piece here, see this result which is after running `tree` on Windows:

image

There are 19 results, and every one of these is represented in the overview ruler, allowing the user to know immediately how far back the term(s) are.

Tyriar (changes requested):
Looks promising! 👏

addons/addon-search/src/SearchAddon.ts
```diff
@@ -40,47 +40,67 @@ export interface ISearchResult {
...
+  /**
+   * This should be high enough so not to trigger a lot of searches
+   * and subsequently a lot of canceled searches which clean up their own
+   * decorations and cause flickers
+   */
+  private _debounceTimeWindow = 300;
+  /**
+   *  Using this mainly for resizing event
+   */
+  private _longerDebounceTimeWindow = 1000;
```
Tyriar, line 95: The preference for these type of constants are to be in a const enum at the top of the file. This means they're inlined in the compiled source so less bytes, as well as being all under one named object.

AnouarTouati:
> Something we want to keep is the highlighting of results in the "overview ruler":
Limiting these results too heavily will hurt the UX by not showing additional results. Make sure you add the overview ruler in the demo when working on this

I am currently highlighting all of the matches so this will not be a problem.
In the future I will  highlight only the the matches that are in the viewport. and highlight everything on the overview ruler.

AnouarTouati:
> If I'm understanding you correctly, searching for CD second should only check if there is a D after the already existing C
This is by design, incremental search means we can avoid doing much of the work, provided regex is not enabled which I think disables incremental completely.

No, this is what the incremental flag is doing  
Screenshot from 2025-01-06 14-05-20
then replace D with B in the search bar (dont backspace. Do Ctrl + A and overwrite)

Screenshot from 2025-01-06 14-05-59

Notice that the second B is selected.
matches before D are not considered for the first select.

That's all it does. It does not help performance.

I removed this behavior.
I will implement an actual incremental search in the future.

addons/addon-search/src/SearchAddon.ts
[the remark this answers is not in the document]
AnouarTouati: Here is what I understood and correct me if I am wrong.
I will be en-queuing  call back functions (Tasks) that process X number of lines. The task queue will  run as many tasks as possible until we exceed the time limit at which point it yields, then do a new setTimeout to start processing of next batch of Tasks.

> https://github.com/Tyriar/xterm.js/blob/18c9eb196010ec082ad56adf7040ed2dca5fd391/src/common/TaskQueue.ts#L104-L109

About the race condition comment. I think it arises from line 71 `this._idleCallback = undefined;` which in my opinion should be removed and inserted above line 95 `this.start()` that way new enqueue calls wont trigger new setTimeouts.

addons/addon-search/test/SearchAddon.test.ts
```diff
@@ -42,68 +56,118 @@ test.describe('Search Tests', () => {
...
-    selectionPosition = (await ctx.proxy.getSelectionPosition())!;
-    deepStrictEqual(line.substring(selectionPosition.start.x, selectionPosition.end.x + 3), 'package.json');
-    await ctx.page.evaluate(`window.search.findNext('package.jsonc', {incremental: true})`);
-    // We have to reevaluate line because it should have switched starting rows at this point
-    selectionPosition = (await ctx.proxy.getSelectionPosition())!;
-    line = await (await ctx.proxy.buffer.active.getLine(selectionPosition.start.y))!.translateToString();
-    deepStrictEqual(line.substring(selectionPosition.start.x, selectionPosition.end.x), 'package.jsonc');
-  });
+
+  // test('Incremental Find Previous', async () => {
```
AnouarTouati, line 70: The "Incremental" functionality was removed in this PR, so I commented out the test. Waiting for your feedback on whether or not to re-implement it.

addons/addon-search/test/SearchAddon.test.ts
```diff
@@ -118,273 +182,285 @@ test.describe('Search Tests', () => {
...
+        await timeout(TIMEOUT);
+        await ctx.page.evaluate(`window.search.findNext('c', { decorations: { activeMatchColorOverviewRuler: '#ff0000' } })`);
+        await timeout(TIMEOUT);
+        await ctx.page.evaluate(`window.search.findNext('c', { decorations: { activeMatchColorOverviewRuler: '#ff0000' } })`);
+        await timeout(TIMEOUT);
+
+        deepStrictEqual(await ctx.page.evaluate('window.calls[window.calls.length - 1 ]'),
+          { resultCount: 3, resultIndex: 2, searchCompleted: true  });
       });
+    //     test('should fire with correct event values (incremental)', async () => {
```
AnouarTouati, line 220: Same here with incremental functionality.

addons/addon-search/test/SearchAddon.test.ts
```diff
@@ -118,273 +182,285 @@ test.describe('Search Tests', () => {
...
+        { resultCount: 3, resultIndex: 2, searchCompleted : true }
+      );
+      await ctx.page.evaluate(`window.search.findPrevious('c', { decorations: { activeMatchColorOverviewRuler: '#ff0000' } })`);
+      await timeout(TIMEOUT);
+      deepStrictEqual(
+        await ctx.page.evaluate('window.calls[window.calls.length-1]'),
+        { resultCount: 3, resultIndex: 1, searchCompleted : true }
+      );
+    });
+    // test('should fire with correct event values (incremental)', async () => {
```
AnouarTouati, line 377: And here.

AnouarTouati (commented):
Do I need to put "incremental" functionality back?

addons/addon-search/test/SearchAddon.test.ts
```diff
@@ -118,273 +182,285 @@ test.describe('Search Tests', () => {
...
+    //       ]);
+    //       deepStrictEqual(await ctx.page.evaluate(`window.search.findNext('abcd', { incremental: true, decorations: { activeMatchColorOverviewRuler: '#ff0000' } })`), false);
+    //       deepStrictEqual(await ctx.page.evaluate('window.calls'), [
+    //         { resultCount: 3, resultIndex: 0 },
+    //         { resultCount: 2, resultIndex: 0 },
+    //         { resultCount: 2, resultIndex: 0 },
+    //         { resultCount: 2, resultIndex: 1 },
+    //         { resultCount: 2, resultIndex: 1 },
+    //         { resultCount: 0, resultIndex: -1 }
+    //       ]);
```
AnouarTouati, line 264: Commented out "Incremental" test.

addons/addon-search/test/SearchAddon.test.ts
```diff
@@ -118,273 +182,285 @@ test.describe('Search Tests', () => {
...
+    //     { resultCount: 1000, resultIndex: -1 },
+    //     { resultCount: 1000, resultIndex: -1 }
+    //   ]);
+    //   strictEqual(await ctx.page.evaluate(`window.search.findPrevious('bc', { decorations: { activeMatchColorOverviewRuler: '#ff0000' } })`), true);
+    //   deepStrictEqual(await ctx.page.evaluate('window.calls'), [
+    //     { resultCount: 1000, resultIndex: -1 },
+    //     { resultCount: 1000, resultIndex: -1 },
+    //     { resultCount: 1000, resultIndex: -1 }
+    //   ]);
+    // });
```
AnouarTouati, line 445: Commented out "Incremental" test.

AnouarTouati (commented):
I removed the "incremental" functionality while working on the code. Do I need to put  back to pass these tests? Or have it incremental as in getting a subset of the previous search result i.e, only search inside in the previous set of matches instead of the whole buffer/terminal.

addons/addon-search/src/SearchAddon.ts
```diff
@@ -37,654 +38,726 @@ export interface ISearchAddonOptions {
...
+    this._onDidChangeResults.fire({ resultIndex:this._currentMatchIndex, resultCount: this._matches.length,searchCompleted: this._searchCompleted });
   }
+  private *_chunkSearchGenerator(term: string): Generator<{direction: string,chunkSize: number}>{
 
-  private _fireResults(searchOptions?: ISearchOptions): void {
-    if (searchOptions?.decorations) {
-      let resultIndex = -1;
-      if (this._selectedDecoration.value) {
-        const selectedMatch = this._selectedDecoration.value.match;
-        for (let i = 0; i < this._highlightDecorations.length; i++) {
```
AnouarTouati, line 342: Fire whenever we have a new chunk regardless of searchOptions.decorations is set or not.

addons/addon-search/test/SearchAddon.test.ts
```diff
@@ -468,20 +573,25 @@ test.describe('Search Tests', () => {
...
-      await ctx.proxy.write('\\x1b[CHi Hi');
-      strictEqual(await ctx.page.evaluate(`window.search.findPrevious('h', { decorations: { activeMatchColorOverviewRuler: '#ff0000' } })`), true);
-      deepStrictEqual(await ctx.page.evaluate('window.calls'), [
-        { resultCount: 2, resultIndex: 1 }
-      ]);
+      await ctx.proxy.write('\\x1b[CHi Hi Hi');
+      await ctx.page.evaluate(`window.search.findPrevious('h', { decorations: { activeMatchColorOverviewRuler: '#ff0000' } })`);
+      await timeout(TIMEOUT);
+      deepStrictEqual(await ctx.page.evaluate('window.calls[window.calls.length-1]'),
+        { resultCount: 3, resultIndex: 0, searchCompleted: true }
```
AnouarTouati, line 581: We are not firing once per search, but multiple per search (on each chunk ), so we need to get the last entry which should have `searchCompleted` `true`.

AnouarTouati (commented):
This is the API changes needed for async search.

AnouarTouati:
I fixed all the bugs I encountered, except for the flicker. 
When the search result changes e.g., from "fixes" to "fixe" like in the video below. 
The new decorations being rendered and old ones being removed might overlap, because I am not synchronizing that. 
It should only flicker once, instead it is flicking multiple times. 
**Any hints of why that is happening ?**

https://github.com/user-attachments/assets/ecfa4099-217b-4e78-8c7b-e3c900fad0e3


Also I  tested forcing it to remove all decorations before a new search starts, and finding all matches before rending it (by setting `CHUNK_SIZE` and `LINE_LIMIT` to equal highlight limit), this fixes the issue.

https://github.com/user-attachments/assets/1e6b98ab-5b7a-4bae-a9f7-7b197cc22922

Issue #5270, opened 2025-01-01
@xtermjs\addon-search regex case-sensitive should behave like VSCode


esimkowitz:
Currently, the @xtermjs/addon-search plugin applies the case-sensitive flag to the full regex, which means that if it's false, meta-characters like `\S`, which is non-whitespace chars, will be interpreted as lowercase (in this case, `\s` is whitespace chars). This can cause unintended problems. Monaco uses a different method, where only non-meta chars are interpreted as case-sensitive when the regex flag is set. While it's kind of silly to treat a regex as case-insensitive, it makes sense for xtermjs and monaco to align on their implementation.

## Details
- Browser and browser version: Electron 33.2.1
- OS version: macOS Sequoia 15.2
- xterm.js version: 5.5.0

### Steps to reproduce

1. run `echo "path wrath hath"` in the VSCode terminal
2. activate the terminal search, set case-sensitive and whole-word to false, and set regex to true
3. type `\S+th` in the search bar
4. observe that nothing is highlighted
5. Open a new tab in VSCode and type `path wrath hath` into it as its only contents
6. activate the Monaco search, set case-sensitive and whole-word to false, and set regex to true
7. type `\S+th` in the search bar
8. observe that each word is being highlighted separately

esimkowitz:
I'm willing to look into a fix for this

jerch:
> Currently, the @xtermjs/addon-search plugin applies the case-sensitive flag to the full regex, which means that if it's false, meta-characters like `\S`, which is non-whitespace chars, will be interpreted as lowercase (in this case, `\s` is whitespace chars).

Sounds like a bug to me, special chars may never flip case, as it typically flips their in/out filter meaning. Shouldn't the `i` flag alone deal with that? :thinking: 

> While it's kind of silly to treat a regex as case-insensitive, ...

Yes, I second that. Imho regexps should be treated the exact way they were given by default. If someone wants to ignore case, just append the i-flag to the regexp ctor on top.

Tyriar:
Yep looks like a bug. Relevant code is here:

https://github.com/anouartouati/xterm.js/blob/c71ef1b91a736db5fad024600f0303809215052d/addons/addon-search/src/SearchAddon.ts#L551-L552

So we just need a conditional flow there that takes `this._searchOptions.regex` into account. Could lift the handling of that from monaco too if we can find it.

esimkowitz:
i can work on this if you wanna assign it to me

PR #5279, opened 2025-01-06
Fixes: #5270 regex case-sensitive should behave like monaco

AnouarTouati:


Tyriar (approved):
Works 👍 just made some small changes to match the repo's style

Issue #4486, opened 2023-04-21
Cannot enter Chinese punctuation in chrome


zhanba:
## Details
- Browser and browser version: Version 112.0.5615.121 (Official Build) (arm64)
- OS version: 13.3.1 (22E261)
- xterm.js version: test on http://xtermjs.org or vscode 1.77.3

### Steps to reproduce

1. install any third party chinese IME，like (sougou)[https://pinyin.sogou.com/mac/] input （not the mac builtin chinese simplified）
2. enter punctuation，expect chinese punctuation bug get english punctuation

jerch:
There were issue with the sogou pinyin in the past, plz also check the issue history, if there is a workaround.

PR #5282, opened 2025-01-07
Fixed CapsLock triggering input twice in MacOS.

cweijan:
Hi @Tyriar, when I use the terminal to input Chinese on MacOS, if I press CapsLock, it will trigger English input twice, but it should only trigger once. This PR fixes this problem.

Produce:
1. Enable PinYin IME
2. Type "asd"(Chinese auto completions appear)
3. Caps lock on
4. Result:
  - Expected input: asd
  - Actual input: a s dasd
image
image

My processor is Mac **M1 Pro**, and this problem occurs in https://xtermjs.org and **VS Code terminal**, but I am surprised that no one has reported this problem. This has troubled me for a long time, I'm pretty sure this is a bug, because on the **native terminal on MacOS**, it behaves as expected. Looking forward to your reply.

### DIFF
--- a/src/browser/input/CompositionHelper.ts
+++ b/src/browser/input/CompositionHelper.ts
@@ -93,7 +93,8 @@ export class CompositionHelper {
    */
   public keydown(ev: KeyboardEvent): boolean {
     if (this._isComposing || this._isSendingComposition) {
-      if (ev.keyCode === 229) {
+      if (ev.keyCode === 20 || ev.keyCode === 229) {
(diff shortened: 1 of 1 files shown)


Tyriar (approved):
Thanks! This change looks relatively safe to merge.

It may have been reported, I've consolidated all Chinese IME problems into https://github.com/xtermjs/xterm.js/issues/4486

cweijan:
Thanks!

PR #5291, opened 2025-01-09
Make ProgressAddon.onChange non undefined

### LINKED ISSUE #5288: Progress addon onChange should not be `| undefined`
If the addon exists we should set this, otherwise it complicates setting up a listener

Image

Tyriar:
Fixes #5288


Tyriar:
FYI @jerch, your work in action 🎉 looks decent without dealing with any debouncing of events.

PR #5292, opened 2025-01-10
API for addons

### LINKED ISSUE #5283: Expose important pieces of vs folder on API
Coming from https://github.com/xtermjs/xterm.js/pull/5251#discussion_r1905849453

Currently using Event, Disposable etc. from the `vs/*` path in addons adds ~50kB on addons. We should investigate, if and how we can save some binary size on addons by exposing crucial parts on the API, so the impls dont have to be copied over for every single addon.

jerch:
Attempt to fix #5283.

@Tyriar This is my first attempt to solve #5283.

It works quite well, the addon stubs under `/shared` only add ~500 bytes on the xtermjs package, and also on addons that derive their addon class from it. Only inconvenience so far is the fact, how we synthesize the final xtermjs package, which leads to an import switch:
- addons in xtermjs repo have to use `import { ... } from 'shared/shared';`
- 3rd party addons can import the shared symbols from xterm package

Currently I have only one testbed implemented with it in `ProgressAddon.ts` using the `EmitterAddon` class. Should be the same with `DisposableAddon`, but I didnt want to change too much in code before getting some feedback from you. So plz have a look at the idea and whether it goes into the right direction. Also we kinda need to cut type ropes for the vs objects, so plz also see the added types in d.ts - we prolly should keep them as small as possible while still being useful.

And last but not least the question, whether we should put these additions under a separate name in the API (maybe `shared.*`?)

### DIFF
--- a/addons/addon-progress/src/ProgressAddon.ts
+++ b/addons/addon-progress/src/ProgressAddon.ts
@@ -3,9 +3,14 @@
  * @license MIT
  */
 
-import type { Terminal, ITerminalAddon, IDisposable } from '@xterm/xterm';
+import { Terminal, ITerminalAddon, IDisposable, EmitterCtorType, IEmitter, IEvent } from '@xterm/xterm';
 import type { ProgressAddon as IProgressApi, IProgressState } from '@xterm/addon-progress';
-import type { Emitter, Event } from 'vs/base/common/event';
+
+// to use impl parts:
+// in 3rd party addons
+//import { EmitterAddon } from '@xterm/xterm';
+// in xtermjs repo addons
+import { EmitterAddon } from 'shared/shared';
 
 
 const enum ProgressType {
@@ -33,13 +38,18 @@ function toInt(s: string): number {
 }
 
 
-export class ProgressAddon implements ITerminalAddon, IProgressApi {
+export class ProgressAddon extends EmitterAddon implements ITerminalAddon, IProgressApi {
   private _seqHandler: IDisposable | undefined;
   private _st: ProgressType = ProgressType.REMOVE;
   private _pr = 0;
-  // HACK: This uses ! to align with the API, this should be fixed when 5283 is resolved
-  private _onChange!: Emitter<IProgressState>;
-  public onChange!: Event<IProgressState>;
+  private _onChange: IEmitter<IProgressState>;
+  public onChange: IEvent<IProgressState>;
+
+  constructor(protected readonly emitterCtor: EmitterCtorType) {
+    super(emitterCtor);
+    this._onChange = new this.emitterCtor<IProgressState>();
+    this.onChange = this._onChange.event;
+  }
 
   public dispose(): void {
     this._seqHandler?.dispose();
@@ -81,9 +91,6 @@ export class ProgressAddon implements ITerminalAddon, IProgressApi {
       }
       return true;
     });
-    // FIXME: borrow emitter ctor from xterm, to be changed once #5283 is resolved
-    this._onChange = new (terminal as any)._core._onData.constructor();
-    this.onChange = this._onChange!.event;
   }
 
   public get progress(): IProgressState {
--- a/src/shared/shared.ts
+++ b/src/shared/shared.ts
@@ -1,0 +1,44 @@
+/**
+ * Copyright (c) 2024 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { IDisposable, IDisposableStore, DisposableStoreCtorType, EmitterCtorType } from '@xterm/xterm';
+
+export { DisposableStore, toDisposable } from 'vs/base/common/lifecycle';
+export { Emitter } from 'vs/base/common/event';
+
+export class DisposableAddon implements IDisposable {
+  protected readonly _store: IDisposableStore;
+
+  constructor(storeCtor: DisposableStoreCtorType) {
+    this._store = new storeCtor();
+  }
+
+  dispose(): void {
+    this._store.dispose();
+  }
+}
+
+
+export class EmitterAddon {
+  constructor(
+    protected readonly emitterCtor: EmitterCtorType
+  ) {}
+}
+
+
+export class DisposableEmitterAddon implements IDisposable {
+  protected readonly _store: IDisposableStore;
+
+  constructor(
+    readonly storeCtor: DisposableStoreCtorType,
+    protected readonly emitterCtor: EmitterCtorType
+  ) {
+    this._store = new storeCtor();
+  }
+
+  dispose(): void {
+    this._store.dispose();
+  }
+}
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -1956,4 +1956,71 @@ declare module '@xterm/xterm' {
      */
     readonly wraparoundMode: boolean;
   }
+
+
+  /**
+   * Get Emitter constructor.
+   */
+  export const emitterCtor: EmitterCtorType;
+
+  /**
+   * Get DisposableStore contructor.
+   */
+  export const disposableStoreCtor: DisposableStoreCtorType;
+
+  /**
+   * Turn a function into a Disposable.
+   */
+  export const toDisposable: (fn: () => void) => IDisposable;
+
+
+  export interface IEmitter<T> {
+    dispose(): void;
+    event: IEvent<T>;
+    fire(event: T): void;
+    hasListeners(): boolean;
+  }
+
+  interface IDisposableStore extends IDisposable {
+    /**
+     * `true` if this object has been disposed of.
+     */
+    isDisposed: boolean;
+    /**
+     * Dispose of all registered disposables but do not mark this object as disposed.
+     */
+    clear(): void;
+    /**
+     * Add a new {@link IDisposable disposable} to the collection.
+     */
+    add<T extends IDisposable>(o: T): T;
+    /**
+     * Deletes a disposable from store and disposes of it. This will not throw or warn and proceed to dispose the
+     * disposable even when the disposable is not part in the store.
+     */
+    delete<T extends IDisposable>(o: T): void;
+    /**
+     * Deletes the value from the store, but does not dispose it.
+     */
+    deleteAndLeak<T extends IDisposable>(o: T): void;
+  }
+
+  export type EmitterCtorType = { new<T>(): IEmitter<T> };
+  export type DisposableStoreCtorType = { new(): IDisposableStore; }
+
+  export class DisposableAddon implements IDisposable {
+    protected readonly _store: IDisposableStore;
+    constructor(storeCtor: DisposableStoreCtorType);
+    dispose(): void;
+  }
+  export class EmitterAddon {
+    protected readonly emitterCtor: EmitterCtorType;
+    constructor(emitterCtor: EmitterCtorType);
+  }
+  export class DisposableEmitterAddon implements IDisposable {
+    protected readonly _store: IDisposableStore;
+    protected readonly emitterCtor: EmitterCtorType;
+    constructor(storeCtor: DisposableStoreCtorType, emitterCtor: EmitterCtorType);
+    dispose(): void;
+  }
 }
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -35,6 +35,7 @@ const config = {
       common: path.resolve('./out/common'),
       browser: path.resolve('./out/browser'),
       vs: path.resolve('./out/vs'),
+      shared: path.resolve('./out/shared')
     }
   },
   output: {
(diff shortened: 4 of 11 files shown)


jerch:
@Tyriar Ah well, still not there - tried to simplify the imports/exports, which again led to pulling vs stuff into addon :blush: Need to fix that again...

jerch:
Well treeshaking is not working with webpack, a 3rd part addon would still include the whole xterm.js lib. Needs a separate shared.js output in the final xterm.js (not gonna fix that today...)

Tyriar:
> 3rd party addons can import the shared classes from xterm package
> 3rd party addon would still include the whole xterm.js lib.

I wouldn't worry about 3rd party libs using our helper. They will have access to Emitter and Event via the API so they can make the tiny wrappers on their end if they need them. Not worth it imo.

webpack.config.js
```diff
@@ -35,6 +35,7 @@ const config = {
       common: path.resolve('./out/common'),
       browser: path.resolve('./out/browser'),
       vs: path.resolve('./out/vs'),
+      shared: path.resolve('./out/shared')
```
Tyriar, line 38: By `shared/` I meant inside `common` somewhere, similar to `browser/renderer/shared/`'s usage in webgl

typings/xterm.d.ts
```diff
@@ -1956,4 +1956,73 @@ declare module '@xterm/xterm' {
...
+
+  /**
+   * Get DisposableStore contructor.
+   */
+  export const disposableStoreCtor: DisposableStoreCtorType;
+
+  /**
+   * Turn a function into a Disposable.
+   */
+  export const toDisposable: (fn: () => void) => IDisposable;
```
Tyriar, line 1974: I think you should be able to `export class` here, similar to `export class Terminal`. That way we don't eed to deal with this `CtorType` stuff

typings/xterm.d.ts
```diff
@@ -1956,4 +1956,73 @@ declare module '@xterm/xterm' {
...
+  export class EmitterAddon {
+    protected readonly _emitterCtor: EmitterCtorType;
+    constructor(_emitterCtor: EmitterCtorType);
+  }
+  export class DisposableEmitterAddon implements IDisposable {
+    protected readonly _store: IDisposableStore;
+    protected readonly _emitterCtor: EmitterCtorType;
+    constructor(_storeCtor: DisposableStoreCtorType, _emitterCtor: EmitterCtorType);
+    public dispose(): void;
+  }
```
Tyriar, line 2027: I wouldn't have thought this would be necessary? Emitters are generally several props on the object, not part of the base class.

src/shared/shared.ts
```diff
@@ -0,0 +1,42 @@
...
+  protected readonly _store: IDisposableStore;
+
+  constructor(protected readonly _storeCtor: DisposableStoreCtorType) {
+    this._store = new _storeCtor();
+  }
+
+  public dispose(): void {
+    this._store.dispose();
+  }
+}
```
Tyriar, line 19: This is all this file needs to include imo. Though I'd call it `AddonDisposable`, as in a disposable for use in addons. `DisposableAddon` signals to me that it's only intended for extending on the main addon object, not other parts of the addon.

addons/addon-progress/src/ProgressAddon.ts
```diff
@@ -33,13 +40,18 @@ function toInt(s: string): number {
...
   private _seqHandler: IDisposable | undefined;
   private _st: ProgressType = ProgressType.REMOVE;
   private _pr = 0;
-  // HACK: This uses ! to align with the API, this should be fixed when 5283 is resolved
-  private _onChange!: Emitter<IProgressState>;
-  public onChange!: Event<IProgressState>;
+  private _onChange: IEmitter<IProgressState>;
+  public onChange: IEvent<IProgressState>;
+
+  constructor(protected readonly _emitterCtor: EmitterCtorType) {
```
Tyriar, line 50: Maybe it's nicest to just pass in the whole xterm API object to some addons? I think we can do this without error:

```ts
import type * as XtermApi from '@xterm/xterm';
import type { Terminal, ITerminalAddon, IDisposable } from '@xterm/xterm';
```

That way it would be nicer from the embedder side:

```ts
new ProgressAddon(xterm)
new WebglAddon(xterm)
etc.
```

Looks better than this imo:

```ts
new ProgressAddon(AddonDisposable)
new WebglAddon(AddonDisposable)
```

Tyriar, line 50: In retrospect it's obvious this would be a breaking change, but that's fine and worth it considering the wins we get in bundle size.

Tyriar (changes requested):

webpack.config.js
jerch: Yeah, gonna move that under common, having another top level folder just complicates things unnecessarily.

typings/xterm.d.ts
jerch: I am not happy yet with the emitter base class, the emitter stuff is always a composition pattern with the ctor handed over at the ctor, so a base class is not really needed. I prolly gonna remove those again and just leave the emitter ctor on the API.
When an addon needs event stuff it can simply make the emitter ctor a mandatory argument on its own ctor and do the proper event setup there.

src/shared/shared.ts
jerch: Ah ic, yeah haven`t thought about it this way.

addons/addon-progress/src/ProgressAddon.ts
jerch: Yes, thats a good idea. The dispoable + emitter ctors, if both are needed, already make this cumbersome and looking ugly. With the whole exports as one arguments it gets much nicer and easier to comprehend on caller side.
I even wonder if we should make that the first default argument on addon ctors, this way ppl wont get it wrong on certain addons (well thats a major API shift).

jerch: About breaking change - as I wrote above, I wonder whether to make the xterm exports the first ctor argument for all addons, e.g.:
```typescript
import * as XtermApi from '@xterm/xterm';
import { AddonXY } from '@term/addon-xy';

const term = XtermApi.Terminal(...);
const addonXY = new AddonXY(XtermApi);
```

Then an addon ctor is free to use the exported ctors there or to ignore that argument:
```typescript
import type * as XtermApi from '@xterm/xterm';
class AddonXY extends ... {
  contructor(xterm: XtermApi, other_args) {
    // if addon needs an event:
    this._onWhatever = xterm.Emitter<Whatever>();
    ...
    // else: just ignore xterm argument
  }
}
```

jerch:
> I wouldn't worry about 3rd party libs using our helper. They will have access to Emitter and Event via the API so they can make the tiny wrappers on their end if they need them. Not worth it imo.

True. So lets treat the shared stuff as internal helpers only, and maybe place some recipes for event or dispoable usage in the docs for external folks. Both can be done easily with composition after we added their ctors to the API.

addons/addon-progress/src/ProgressAddon.ts
jerch: Or if thats too "globalish" looking, we could also aggregate the extra exports under a `shared` API endpoint:
```typescript
import { Terminal, shared } from '@xterm/xterm';
import { AddonXY } from '@term/addon-xy';

const term = new Terminal(...);
const addonXY = new AddonXY(shared);
```

jerch: The latter has a few advantages, like keeping stuff under the Terminal umbrella and automatically gaining access to those symbols even on a terminal instance.

webpack.config.js
jerch: Moved it to `common/shared`.

src/shared/shared.ts
jerch: Changed to `AddonDisposable` under `common/shared/AddonDisposable.ts`.

addons/addon-progress/src/ProgressAddon.ts
jerch: Well the full module export type interface stubbing felt kinda wrong. My next approach looks like that:

on addon impl side:
```typescript
import { ..., ISharedExports } from '@xterm/xterm';

export class AddonXY implements ... {

  constructor(sharedExports: ISharedExports) {
    // do something with things exposed under sharedExports
  }
```

on embedding side:
```typescript
import type { Terminal, sharedExports } from '@xterm/xterm';
import { AddonXY } from '@term/addon-xy';

const term = new Terminal(...);
const addonXY = new AddonXY(sharedExports);
```

jerch:
@Tyriar Results so far:
- image addon: from ~80 kB down to ~54 kB
- progress addon: from ~45 kB down to ~2 kB
- search addon: from ~50kB down to ~11 kB
- webgl addon: from ~240 kB down to ~125 kB

addons/addon-progress/src/ProgressAddon.ts
Tyriar: > why not to load addons this way in the first place with a terminal instance as first argument

`Terminal` has always been the only thing exposed on the API, but it makes a lot of sense to pass in the full API when we start adding new things. This sort of breaking change is more impactful though as we can't just expect all addon ctors to have it as the first arg.

> Do you remember, why we have the loadAddon mechanics on the terminal the way it is implemented currently?

Was a long time ago, but one of the big things we get is `loadAddon` lets the Terminal take ownership of it. So disposing of a terminal means the addon will be destroyed. That's pretty much all `AddonManager` does.

jerch: Yes it def. smells like quite the big API change would be needed, so idk exactly how to proceed. Maybe we should go back to conceptual structuring before inventing a square wheel here? Gonna try to do a write up of what we have currently vs. what do be done about it in #5283.

typings/xterm.d.ts
Tyriar: > Isn't that pulling types from internal sources into the public API?

Everything in the API needs to be standalone, so we'd duplicate it there.

Also, we depend the other way for Terminal here to ensure our implementation matches the API:

https://github.com/xtermjs/xterm.js/blob/d81b25c3bf6c0e3352c175705f460aa39df6e374/src/browser/public/Terminal.ts#L25

So we could do the same for `DisposableStore`/`Emitter`/etc. by depending on `xterm.d.ts` from `public/Terminal.ts` again.

addons/addon-progress/src/ProgressAddon.ts
Tyriar: Posted long replies in https://github.com/xtermjs/xterm.js/issues/5283#issuecomment-2587425995

src/common/shared/AddonDisposable.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright (c) 2024 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { IDisposable, IDisposableStore, ISharedExports } from '@xterm/xterm';
+
+
+export class AddonDisposable implements IDisposable {
```
Tyriar, line 9: Should we also export `Disposable` on `ISharedExports` instead of this?

addons/addon-webgl/src/TextureAtlas.ts
```diff
@@ -80,16 +80,22 @@ export class TextureAtlas implements ITextureAtlas {
...
-  public readonly onAddTextureAtlasCanvas = this._onAddTextureAtlasCanvas.event;
-  private readonly _onRemoveTextureAtlasCanvas = new Emitter<HTMLCanvasElement>();
-  public readonly onRemoveTextureAtlasCanvas = this._onRemoveTextureAtlasCanvas.event;
+  private readonly _onAddTextureAtlasCanvas: IEmitter<HTMLCanvasElement>;
+  public readonly onAddTextureAtlasCanvas: IEvent<HTMLCanvasElement>;
+  private readonly _onRemoveTextureAtlasCanvas: IEmitter<HTMLCanvasElement>;
+  public readonly onRemoveTextureAtlasCanvas: IEvent<HTMLCanvasElement>;
 
   constructor(
+    _sharedExports: ISharedExports,
```
Tyriar, line 89: No `_` since it's not a member

typings/xterm-headless.d.ts
```diff
@@ -1358,4 +1358,31 @@ declare module '@xterm/headless' {
...
+  export interface ISharedExports {
+    readonly DisposableStore: new() => IDisposableStore;
+    readonly Emitter: new<T>() => IEmitter<T>;
+    readonly toDisposable: (fn: () => void) => IDisposable;
+  }
+
+  export const sharedExports: ISharedExports;
+
+  // FIXME: @Tyriar - plz have a look at the following interfaces and
+  // to what degree those should be exposed or get stripped down
```
Tyriar, line 1375: These look good as a starting set, can always expand it later

Tyriar:
See https://github.com/xtermjs/xterm.js/issues/5283, we can now just depend on the lightweight versions in common/

jerch:
@Tyriar Yes, this seems the better way forward. Plus once we got ESM sorted out, sub exports are easily possible.

PR #5303, opened 2025-01-28
Update CoreBrowserService.ts

frost-cy:
Move the initializer into the constructor for instance members that reference identifiers declared in the constructor.

When TypeScript outputs modern language features, the below case throws an TS error.

```
class C {
  a = this.x; // TS2729: Property 'x' is used before its initialization. 

  constructor(public x:number){}
}
```
This error is fixed by moving the initializer of such class members into the constructor.

This is a no-op change

### DIFF
--- a/src/browser/services/CoreBrowserService.ts
+++ b/src/browser/services/CoreBrowserService.ts
@@ -13,12 +13,12 @@ export class CoreBrowserService extends Disposable implements ICoreBrowserServic
 
   private _isFocused = false;
   private _cachedIsFocused: boolean | undefined = undefined;
-  private _screenDprMonitor = this._register(new ScreenDprMonitor(this._window));
+  private _screenDprMonitor;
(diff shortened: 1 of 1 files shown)


Tyriar (commented):
It would make things a lot more noisy to do this for events and emitters, and I think they don't meet the condition you're talking about as they're not referencing anything in the constructor?

I think only `_screenDprMonitor` should move? Regardless, it would be best to tackle this holistically when we start to use such features instead of just a single case in one file.

jerch:
I second @Tyriar 's remark - this should not be addressed individually, but uniformly across the codebase.

Btw this a good example, where scoping/visibility rules get hard to explain in JS/TS. Mix in a `static` at some place and even many long-term programmers will get it wrong. My take on this - JS is not very beginner friendly due to its prototype-based OOP.

PR #5308, opened 2025-02-12
Update AccessibilityManager.ts to use textContent rather than innerText

alexemorris:
The innerText property requires layout and therefore is not very efficient. For more information, see http://www.kellegous.com/j/2013/02/27/innertext-vs-textcontent/.


Tyriar (approved):
TIL 👍

PR #5316, opened 2025-03-01
feat(addon-fit): fit to specific dimensions

DSpeichert:
Allows more control by checking proposed dimensions with additional logic.

### DIFF
--- a/addons/addon-fit/src/FitAddon.ts
+++ b/addons/addon-fit/src/FitAddon.ts
@@ -32,8 +32,8 @@ export class FitAddon implements ITerminalAddon , IFitApi {
 
   public dispose(): void {}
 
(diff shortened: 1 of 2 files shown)


Tyriar:
The whole point of the fit addon is to measure the container. You should just not use it and call `Terminal.resize` with those dimensions?

DSpeichert:
My workflow was to use this addon's `proposeDimensions()` first and validate it according to some external criteria.
If it passed, I'd call `fit()` with these specific dimensions. Since `fit()` uses internal/private APIs, I preferred to avoid copying that code to force redraw and rely on the addon, in case it changes in future. You are right that I could call `Terminal.resize` directly though.

Issue #5213, opened 2024-11-15
Add setting to enable reflow for the cursor line


jtbandes:
I'm trying to use xterm.js to make a log viewer (read-only, no keyboard input from the user).

I've used `@xterm/addon-fit` to handle resizing, and automatic line wrapping seems to work nicely. However, the very last line, i.e. the "current" line, is only wrapped based on the original width when the line is written, and doesn't re-wrap when the terminal is resized. When the terminal is sized down, it seems to delete some characters, and when the terminal is sized back up, those characters are permanently gone and it leaves a blank space on the right.

Note that I haven't included a `\r\n` on the last line, because I don't want a blank line to be displayed at the bottom of the terminal.

How can I make this last line automatically wrap when the terminal is resized?

https://github.com/user-attachments/assets/d324d25b-232f-4b5a-bcbc-565374a7b3a0

## Details
- Browser and browser version: Chrome 130.0.6723.117
- OS version: macOS 15.1
- xterm.js version: 5.5.0

### Steps to reproduce

https://codesandbox.io/p/sandbox/xtermjs-test-7xvd9v

```ts
import "./styles.css";
import "@xterm/xterm/css/xterm.css";
import { Terminal } from "@xterm/xterm";
import { FitAddon } from "@xterm/addon-fit";

const container = document.getElementById("app");

const terminal = new Terminal({
  cursorStyle: "bar",
  allowProposedApi: true,
});

const fitAddon = new FitAddon();
terminal.loadAddon(fitAddon);

terminal.open(container);

for (let i = 0; i < 100; i++) {
  if (i > 0) {
    terminal.write("\r\n");
  }
  terminal.write(
    `line ${i}: Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua`
  );
}

const resizeObserver = new ResizeObserver((_entries) => {
  fitAddon.fit();
});
resizeObserver.observe(container);
```

Tyriar:
This is by design because of the customized for shells point from @jerch, we "reflow" all lines except the cursor line and expect the shell running in the terminal to do it for the line with the cursor:

https://github.com/xtermjs/xterm.js/blob/41e8ae395937011d6bf6c7cb618b851791aed395/src/common/buffer/BufferReflow.ts#L44-L49

@jerch do you think there should be a setting to just always reflow?

jerch:
[...]

PR #5317, opened 2025-03-02
docs: added usage tip for registerMarker()

tabarra:
Simple PR to add a usage tip for `term.registerMarker()`.  
Apologies if this sounds obvious, but definitely took me half an hour to figure out how to best deal with this issue.  

> [!IMPORTANT]
> The issue with using the batched line index id only happens when there are lines reflowing above the marker position, and I don't know if the issue would be better solving by using the option from issue #5213.
> Please disregard this PR if there are better options that I'm unaware.


Tyriar (approved):
Thanks for the PR. `registerMarker` isn't used too often with write calls so it clutters the docs having it there imo. I instead made the note about the callback clearer in the write* docs.

Issue #5319, opened 2025-03-14
Resize logic does not work well with conpty v1.22+


Tyriar:
Repro is to run `ls` then resize narrow and wide again. Conpty expects new lines to be added to the bottom and lines to get pushed into the scrollback:

![Image](https://github.com/user-attachments/assets/3ded63cf-6ddb-4852-a9c9-4a27d47206d9)

*The top window shows conpty's buffer state*

This was always a problem in that is deleted content from the buffer as the whole screen was reprinted. With passthrough, this problem is even work as the cursor is put in a random spot after pwsh calls the `GetConsoleCursorInfo` API that is handled by conpty, not xterm.js

VS Code issue: https://github.com/microsoft/vscode/issues/241978

Tyriar:
Reverted that change in https://github.com/xtermjs/xterm.js/pull/5358

Tyriar:
Let's track this in https://github.com/microsoft/vscode/issues/224488 for now.

Issue #4231, opened 2022-10-27
Reflow scrollback even when isWindows is enabled


Tyriar:
Reflow is currently disabled when isWindows is enabled, there's no reason we can't still reflow scrollbakc, just not in the viewport.

https://github.com/xtermjs/xterm.js/blob/a4f8b07a7fa06506b3094eb459bde1f02c085270/src/common/buffer/Buffer.ts#L254-L256

Tyriar:
This is done now when conpty is used.

Issue #3513, opened 2021-10-19
Support alternate handling of scrollback in resize for conpty


Tyriar:
See https://github.com/microsoft/vscode/issues/134448#issuecomment-947053801, conpty doesn't push scrollback down but pushed the viewport up when the rows in the terminal are increased.

Tyriar:
Reverted that change in https://github.com/xtermjs/xterm.js/pull/5358

Tyriar:
Out of scope for now

PR #5358, opened 2025-06-18
Revert conpty-specific reflow handling

### LINKED ISSUE #5357: Terminal can go blank in webgl extension
VS Code issue: https://github.com/microsoft/vscode/issues/251800

Tyriar:
This reverts commit a260f7d2889142d6566a66cb9856a07050dea611, reversing changes made to 2042bb85023714e55c0c2e986b5000e33b17c414 (https://github.com/xtermjs/xterm.js/pull/5321).

Fixes #5357
Reopens #5319, #3513, #4231

This caused some pretty major issues where the buffer could essentially get bricked (https://github.com/microsoft/vscode/issues/251800#issuecomment-2984141671). The situation was much better before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "jsdom": "^18.0.1",
     "mocha": "^10.1.0",
     "mustache": "^4.2.0",
-    "node-pty": "^1.1.0-beta31",
+    "node-pty": "1.1.0-beta19",
     "nyc": "^15.1.0",
     "source-map-loader": "^3.0.0",
     "source-map-support": "^0.5.20",
--- a/src/common/buffer/Buffer.ts
+++ b/src/common/buffer/Buffer.ts
@@ -320,37 +320,7 @@ export class Buffer implements IBuffer {
     if (toRemove.length > 0) {
       const newLayoutResult = reflowLargerCreateNewLayout(this.lines, toRemove);
       reflowLargerApplyNewLayout(this.lines, newLayoutResult.layout);
-
-      // For conpty, it has its own copy of the buffer _without scrollback_ internally. Its behavior
-      // when reflowing larger is to insert empty lines at the bottom of the buffer as when lines
-      // unwrap conpty's view cannot pull scrollback down, so it adds empty lines at the end.
-      let removedInViewport = 0;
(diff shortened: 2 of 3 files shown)


ulrrmehdi:
Is this what you suggested to revert?  I had this box checked in, as suggested by  Megan.  Also to remind the problem solution: I am NOT working in VS Code on MS Windows.

ulrrmehdi:
Screenshot 2025-07-03 at 8 17 48 AM

ulrrmehdi:
Resizing Terminal window does not do anything to me after the Terminal is cleared after the 15 min idle time. The content  in the Terminal does not come back :(. This problem is VERY annoying. First of all, I have no idea how that happened in the first place, after NOT having it before.  It is weird.

ulrrmehdi:
The only way to bring the content back is to open a new Terminal, type my work related settings, like navigating to the right workspace,  setting aliases, all my normal work flow ... until this nasty thing  happens if I do not do anything in this terminal for 15 min. Also, why 15 min, what this magic number is? Have you thought about it? Perhaps, I can set this time out be an hour, for instance, or anything much longer than 15 min?

Tyriar:
@ulrrmehdi I don't know what you're talking about, is there some context in an issue you created somewhere I'm missing? The code this changes only affects you if you're on Windows, regardless of whether that setting is checked.

ulrrmehdi:
The issue I am talking about, and it is not the first time I address this issue via my tickets, is a blanking out of the integrated terminal  in my instance of VS Code after 15 min of it's idle time. Is it not possible to check my previous tickets and see what is there? The bottom line, the suggested resizing of the terminal does not fix anything - the content of the terminal does not come back, remains blank, until I press Output tab in the Terminal, or I open up a new Terminal,  and my DEFAULT prompt reappears again, NOT the content of the Terminal before it's content had disappear without a trace!   I cannot possibly type or do something in the Terminal  all the time to make sure that after 15 min of not touching the Terminal screen, it does not disappear!  Please find out the reason of this odd effect and suggest a working solution. Like I mentioned earlier, this effect appeared somewhat a ~month ago, and never before several  years(!).

Tyriar:
@ulrrmehdi sorry if you were directed to this PR, but this has nothing to do with that. You should create or comment on a related issue in the microsoft/vscode repo.

PR #5385, opened 2025-08-20
Fix terminal find when wrapped

anthonykim1:
Resolves: https://github.com/microsoft/vscode/issues/253544

### DIFF
--- a/addons/addon-search/src/SearchAddon.ts
+++ b/addons/addon-search/src/SearchAddon.ts
@@ -351,7 +351,6 @@ export class SearchAddon extends Disposable implements ITerminalAddon , ISearchA
       let resultIndex = -1;
       if (this._selectedDecoration.value) {
         const selectedMatch = this._selectedDecoration.value.match;
-        // Problem: this.highlightDecorations.length was showing all lines of highlighted.
         for (let i = 0; i < this._searchResultsWithHighlight.length; i++) {
           const match = this._searchResultsWithHighlight[i];
           if (match.row === selectedMatch.row && match.col === selectedMatch.col && match.size === selectedMatch.size) {
--- a/addons/addon-search/src/SearchAddon.ts
+++ b/addons/addon-search/src/SearchAddon.ts
@@ -72,6 +72,7 @@ export class SearchAddon extends Disposable implements ITerminalAddon , ISearchA
   private _cachedSearchTerm: string | undefined;
   private _highlightedLines: Set<number> = new Set();
   private _highlightDecorations: IHighlight[] = [];
+  private _searchResultsWithHighlight: ISearchResult[] = [];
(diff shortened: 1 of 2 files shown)


addons/addon-search/src/SearchAddon.ts
```diff
@@ -118,6 +119,7 @@ export class SearchAddon extends Disposable implements ITerminalAddon , ISearchA
     this._selectedDecoration.clear();
     dispose(this._highlightDecorations);
     this._highlightDecorations = [];
+    this._searchResultsWithHighlight = [];
```
Tyriar, line 122: @anthonykim1 it's not clear what the issue was exactly and how this fixes it, could you explain?

anthonykim1: Yeah @Tyriar ,the issue was we were using `_highlightDecorations `  to show `x of y` where  `_highlightDecorations  = y ` . The problem is that we were using the `_highlightDecorations `, which increases in number if string overflows into the next line, whereas `_searchResultsWithHighlight ` does not. 

The fix was in _fireResults:
```
 private _fireResults(searchOptions?: ISearchOptions): void {
    if (searchOptions?.decorations) {
      let resultIndex = -1;
      if (this._selectedDecoration.value) {
        const selectedMatch = this._selectedDecoration.value.match;
        for (let i = 0; i < this._searchResultsWithHighlight.length; i++) {
          const match = this._searchResultsWithHighlight[i];
          if (match.row === selectedMatch.row && match.col === selectedMatch.col && match.size === selectedMatch.size) {
            resultIndex = i;
            break;
          }
        }
      }
      this._onDidChangeResults.fire({ resultIndex, resultCount: this._searchResultsWithHighlight.length });
    }
  }
```
Switched to use` this._searchResultsWithHighlight` instead of `this._highlightDecorations`


In cases like: Screenshot 2025-08-20 at 6 14 42 PM

The value of `this._highlightDecorations ` was four, while  value of `this._searchResultsWithHighlight` was two. 

I think the highlightDecoration counts more as text gets extended to next line, but searchResultsWithHighlight counts those as single component. 

It seems like we've already had logic to calculate `searchResultsWithHighlight`, but I just made this class variable so I can access it in other function in the file.

Tyriar: Could you add a test that covers this case? I think the search addon tests were in a decent state so agent mode will probably write it in one go.

anthonykim1: yeap, how about [95a9531](https://github.com/xtermjs/xterm.js/pull/5385/commits/95a9531ec17aa430652281e643cf603b6455c5b4)

Tyriar (approved):

PR #5386, opened 2025-08-20
fix: ensure that currentRow can not go out of the range

alexkozy:
With the code today, it is possible to get into a situation where the currentRow goes out of the range and will continue going in one direction until an out-of-memory crash.

There is no easy repro - the only one that works for me is to open an xterm inside Cursor, open the nano editor in a terminal, enter 10 one-character lines (like a, b, c, d, ...), and then make the option + click as quickly as possible.

image
image

### DIFF
--- a/src/browser/input/MoveToCell.ts
+++ b/src/browser/input/MoveToCell.ts
@@ -199,7 +199,9 @@ function bufferLine(
   let currentRow = startRow;
   let bufferStr = '';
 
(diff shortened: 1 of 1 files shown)


tmm1:
This manifests as an OOM in our usage of xterm.js

image

cc @Tyriar @meganrogge

Tyriar (approved):
This looks good, thanks for the investigation!

PR #5390, opened 2025-08-28
Fix scrollbar teleport after exiting alt buffer 

anthonykim1:
This resolves the first issue (teleport issue) from: https://github.com/microsoft/vscode/issues/224750#issuecomment-3221966394


src/common/buffer/BufferSet.ts
```diff
@@ -91,6 +91,8 @@ export class BufferSet extends Disposable implements IBufferSet {
       activeBuffer: this._normal,
       inactiveBuffer: this._alt
     });
+    // Prevent scrollbar "teleport" to top of the terminal, from previous alt buffer.
+    this._bufferService.syncScrollPosition();
```
anthonykim1: Gist: Seems like we had the correct ydisp the whole time, but just needed to update/refresh after coming back from alt buffer. 

Had to add public method `syncScrollPosition()` and call it from here since `_onScroll` was private to bufferService.

Tyriar: Instead of firing a new event, `BufferService` could listen on `onBufferActivate` and fire `onScroll`?

src/common/buffer/BufferSet.test.ts
```diff
@@ -81,4 +82,69 @@ describe('BufferSet', () => {
       assert.equal(bufferSet.alt.markers.length, 0);
     });
   });
+
+  describe('scroll position synchronization', () => {
```
Tyriar: A test here is probably overkill

src/common/buffer/BufferSet.ts
anthonykim1: Yeap, totally. I'm liking the event listener approach better, seems more clean as we don't have to explicitly call buffer service in middle of `activateNormalBuffer` 

Changes: [753ea81](https://github.com/xtermjs/xterm.js/pull/5390/commits/753ea81fc08661b4c661338461c81b3a06178d8c)

PR #5391, opened 2025-08-28
Bring back partial wheel tracking

Tyriar:
Part of microsoft/vscode#224750

### DIFF
--- a/src/common/services/CoreMouseService.ts
+++ b/src/common/services/CoreMouseService.ts
@@ -177,7 +177,7 @@ export class CoreMouseService extends Disposable implements ICoreMouseService {
   private _wheelPartialScroll: number = 0;
 
   private readonly _onProtocolChange = this._register(new Emitter<CoreMouseEventType>());
-  public readonly onProtocolChange =  this._onProtocolChange.event;
+  public readonly onProtocolChange = this._onProtocolChange.event;
 
   constructor(
     @IBufferService private readonly _bufferService: IBufferService,
@@ -270,12 +270,16 @@ export class CoreMouseService extends Disposable implements ICoreMouseService {
 
   private _applyScrollModifier(amount: number, ev: WheelEvent): number {
     const modifier = this._optionsService.rawOptions.fastScrollModifier; // QUESTION: This is always alt. Seems deprecated via: https://github.com/xtermjs/xterm.js/blob/5.5.0/src/browser/Viewport.ts
-    console.log('what is this modifier: ', modifier);
+    // Problem is that modifier always seem 'alt' even when ev.ctrlKey is true, for example.
+    // Maybe this doesnt matter because all of alt, ctrl, shift does same operation?
 
     // Multiply the scroll speed when the modifier key is pressed
     if ((modifier === 'alt' && ev.altKey) ||
-        (modifier === 'ctrl' && ev.ctrlKey) ||
-        (modifier === 'shift' && ev.shiftKey)) {
+      (modifier === 'ctrl' && ev.ctrlKey) ||
+      (modifier === 'shift' && ev.shiftKey)) {
+
+      console.log('fast scroll sensitivity: ', this._optionsService.rawOptions.fastScrollSensitivity);
+      console.log('regular scroll sensitivity: ', this._optionsService.rawOptions.scrollSensitivity);
       return amount * this._optionsService.rawOptions.fastScrollSensitivity * this._optionsService.rawOptions.scrollSensitivity;
     }
 
--- a/src/common/services/CoreMouseService.ts
+++ b/src/common/services/CoreMouseService.ts
@@ -269,18 +269,10 @@ export class CoreMouseService extends Disposable implements ICoreMouseService {
   }
 
   private _applyScrollModifier(amount: number, ev: WheelEvent): number {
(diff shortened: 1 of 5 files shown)


Tyriar:
@anthonykim1 can you pull the partial wheel tracking into CoreMouseService?

anthonykim1:
I just realized the scrolling experience differs alot when using trackpad vs. mouse when scrolling.
My scrolling speed for both mouse and trackpad is set to the median/middle in Mac system device setting.

I'm aggressively scrolling with mouse in the two videos:

https://github.com/user-attachments/assets/94c34893-c030-4804-a600-da9f7a9de2fd


https://github.com/user-attachments/assets/efe9ca30-0a29-4a32-99b6-d0b795ff3333



Seems like better experience with mac trackpad, but bit too slow imo with mouse scrolling.
```
    // Fallback to WheelEvent.DOM_DELTA_LINE
    const targetWheelEventPixels = this._renderService.dimensions.device.cell.height / this._coreBrowserService.dpr;
    let amount = 1;
    if (ev.deltaMode === WheelEvent.DOM_DELTA_PIXEL) {
      amount /= targetWheelEventPixels + 0.0; // Prevent integer division
      this._wheelPartialScroll += amount;
      amount = Math.floor(Math.abs(this._wheelPartialScroll)) * (this._wheelPartialScroll > 0 ? 1 : -1);
      this._wheelPartialScroll %= 1;
    } else if (ev.deltaMode === WheelEvent.DOM_DELTA_PAGE) {
      amount *= this._bufferService.rows;
    }
    return amount;
``` 
Maybe the amount should be somewhere around 2 instead of 1? or wondering how should we handle mouse scroll vs. trackpad scroll.

Tyriar:
@anthonykim1 I think you have vim mouse events off which is why it's scrolling more slowly? As discussed "1 line" scrolled ends up translating into 4 lines when vim mouse support is on, which was 4 times faster than how it should be scrolling.

The first video doesn't look great, could you look into what deltaY, deltaMode and targetWheelEventPixels are in your case? Different mouses/OSs behave differently.

anthonykim1:
@Tyriar I'm not sure, I just added `set mouse=a` to my `.vimrc`, but  it still feels bit sluggish.

Two constant thing I found is that `targetWhellEventPixel` seems to always be `14`  and `deltaMode` always seem `0` 

Here are more findings with mouse scroll: 

* It also seems like `deltaY` should at least be absolute value of 103 in order to scroll to next element. 
* If its a small scroll (`deltaY` abs(13)), then it seems to take MANY attempts to scroll to next).

Video for reference (only used mouse scroll):
https://github.com/user-attachments/assets/16fb95ee-999c-411e-acff-0251d8fdcbd8

I noticed that on trackpad where I have good experience, the `deltaY`  is contained around MAX absolute value of 30, otherwise btw absolute value of 10.

anthonykim1:
Update: I think I got it.
Now mouse scrolling is very similar experience to trackpad

https://github.com/user-attachments/assets/4dfc7850-9bdf-414e-8cf0-e8e3e597d32c

src/common/services/CoreMouseService.ts
```diff
@@ -229,6 +230,35 @@ export class CoreMouseService extends Disposable implements ICoreMouseService {
...
+
+    if (cellHeight === undefined || dpr === undefined) {
+      return 0;
+    }
+
+    // Fallback to WheelEvent.DOM_DELTA_LINE
+    const targetWheelEventPixels = cellHeight / dpr;
+    let amount = 1;
+    if (ev.deltaMode === WheelEvent.DOM_DELTA_PIXEL) {
+      amount = ev.deltaY / (targetWheelEventPixels + 0.0); // Prevent integer division
```
anthonykim1: This was the key 
```
amount = ev.deltaY / (targetWheelEventPixels + 0.0); // Prevent integer division
```

src/common/services/CoreMouseService.ts
```diff
@@ -229,6 +231,59 @@ export class CoreMouseService extends Disposable implements ICoreMouseService {
...
+      amount = Math.floor(Math.abs(this._wheelPartialScroll)) * (this._wheelPartialScroll > 0 ? 1 : -1);
+      this._wheelPartialScroll %= 1;
+    } else if (ev.deltaMode === WheelEvent.DOM_DELTA_PAGE) {
+      amount *= this._bufferService.rows;
+    }
+    return amount;
+  }
+
+  private _applyScrollModifier(amount: number, ev: WheelEvent): number {
+    const modifier = this._optionsService.rawOptions.fastScrollModifier; // QUESTION: This is always alt. Seems deprecated via: https://github.com/xtermjs/xterm.js/blob/5.5.0/src/browser/Viewport.ts
```
anthonykim1: The value of `modifier` is always `alt`. Maybe this is because its been deprecated https://github.com/xtermjs/xterm.js/blob/5.5.0/src/browser/Viewport.ts 

But this will prevent from going into if conditional since modifier and `ev.ctrlKey`, for example never matches.
For example, modifier would always be `alt` when `ev.ctrlKey` would be true.

anthonykim1: @Tyriar  Should I just change the conditional so that we can apply `return amount * this._optionsService.rawOptions.fastScrollSensitivity * this._optionsService.rawOptions.scrollSensitivity;` when any of `ev.altKey || ev.ctrlKey || ev.shiftKey` are true?

Tyriar: Yes since modifier is deprecated just ignore that

PR #5392, opened 2025-08-29
headless onRowChange event

jim-lake:
Add a onRowChange event so that watchers can do efficient inspection of buffer changes.

### DIFF
--- a/src/headless/Terminal.ts
+++ b/src/headless/Terminal.ts
@@ -38,6 +38,8 @@ export class Terminal extends CoreTerminal {
   public readonly onA11yChar = this._onA11yCharEmitter.event;
   private readonly _onA11yTabEmitter = this._register(new Emitter<number>());
   public readonly onA11yTab = this._onA11yTabEmitter.event;
+  private readonly _onRowChange = this._register(new Emitter<{ start: number, end: number }>());
+  public readonly onRowChange = this._onRowChange.event;
 
   constructor(
     options: ITerminalOptions = {}
@@ -53,6 +55,11 @@ export class Terminal extends CoreTerminal {
     this._register(Event.forward(this._inputHandler.onTitleChange, this._onTitleChange));
     this._register(Event.forward(this._inputHandler.onA11yChar, this._onA11yCharEmitter));
     this._register(Event.forward(this._inputHandler.onA11yTab, this._onA11yTabEmitter));
+    this._register(this._inputHandler.onRequestRefreshRows(e => {
+      if (e) {
+        this._onRowChange.fire({ start: e.start, end: e.end });
+      }
+    }));
   }
 
   /**
--- a/src/headless/public/Terminal.ts
+++ b/src/headless/public/Terminal.ts
@@ -11,7 +11,7 @@ import { Terminal as TerminalCore } from 'headless/Terminal';
 import { AddonManager } from 'common/public/AddonManager';
 import { ITerminalOptions } from 'common/Types';
 import { Disposable } from 'vs/base/common/lifecycle';
-import type { Event } from 'vs/base/common/event';
+import { Event } from 'vs/base/common/event';
 /**
  * The set of options that only have an effect when set in the Terminal constructor.
  */
@@ -81,6 +81,10 @@ export class Terminal extends Disposable implements ITerminalApi {
   public get onScroll(): Event<number> { return this._core.onScroll; }
   public get onTitleChange(): Event<string> { return this._core.onTitleChange; }
   public get onWriteParsed(): Event<void> { return this._core.onWriteParsed; }
+  public get onRowChange(): Event<{ start: number, end: number }> {
+    this._checkProposedApi();
+    return Event.map(this._core.onRowChange, e => ({ start: e.start, end: e.end }));
+  }
 
   public get parser(): IParser {
     this._checkProposedApi();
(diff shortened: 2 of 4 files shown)


jim-lake:
The goal is to expose the _onRequestRefreshRows, so you can write a renderer on top of headless, like a real renderer would in regular xterm.js.  The specific fire that is very useful in rendering is actually for a scroll event:

this._onRequestRefreshRows.fire({
        start: Math.min(viewportStart, this._bufferService.rows - 1),
        end: Math.min(viewportEnd, this._bufferService.rows - 1)
      });

This event provides just what you need to for render on headless.  onLineFeed and onCursorMove didn't definitely give me this fire every time the buffer scrolls.

src/headless/public/Terminal.ts
```diff
@@ -81,6 +81,10 @@ export class Terminal extends Disposable implements ITerminalApi {
   public get onScroll(): Event<number> { return this._core.onScroll; }
   public get onTitleChange(): Event<string> { return this._core.onTitleChange; }
   public get onWriteParsed(): Event<void> { return this._core.onWriteParsed; }
+  public get onRowChange(): Event<{ start: number, end: number }> {
+    this._checkProposedApi();
+    return Event.map(this._core.onRowChange, e => ({ start: e.start, end: e.end }));
```
Tyriar, line 86: We want to avoid `Event.map` here if possible, now sure why this would be needed?

src/headless/public/Terminal.ts
```diff
@@ -81,6 +81,10 @@ export class Terminal extends Disposable implements ITerminalApi {
   public get onScroll(): Event<number> { return this._core.onScroll; }
   public get onTitleChange(): Event<string> { return this._core.onTitleChange; }
   public get onWriteParsed(): Event<void> { return this._core.onWriteParsed; }
+  public get onRowChange(): Event<{ start: number, end: number }> {
+    this._checkProposedApi();
```
Tyriar, line 85: Can skip making it proposed since it's so simple

src/headless/public/Terminal.ts
```diff
@@ -186,7 +190,7 @@ export class Terminal extends Disposable implements ITerminalApi {
   }
   public loadAddon(addon: ITerminalAddon): void {
     // TODO: This could cause issues if the addon calls renderer apis
-    this._addonManager.loadAddon(this as any, addon);
+    this._addonManager.loadAddon(this as any, addon as any);
```
Tyriar, line 193: Why was this needed?

src/headless/Terminal.ts
```diff
@@ -53,6 +55,11 @@ export class Terminal extends CoreTerminal {
     this._register(Event.forward(this._inputHandler.onTitleChange, this._onTitleChange));
     this._register(Event.forward(this._inputHandler.onA11yChar, this._onA11yCharEmitter));
     this._register(Event.forward(this._inputHandler.onA11yTab, this._onA11yTabEmitter));
+    this._register(this._inputHandler.onRequestRefreshRows(e => {
+      if (e) {
+        this._onRowChange.fire({ start: e.start, end: e.end });
+      }
+    }));
```
Tyriar, line 62: Shouldn't the event also fire undefined? I think that means every row needs to be updated?

typings/xterm-headless.d.ts
```diff
@@ -748,6 +748,13 @@ declare module '@xterm/headless' {
      */
     onResize: IEvent<{ cols: number, rows: number }>;
 
+    /**
+     * Adds an event listener for when buffer rows change during parsing. The event
+     * value contains the range of rows that changed.
+     * @returns an `IDisposable` to stop listening.
+     */
+    onRowChange: IEvent<{ start: number, end: number }>;
```
Tyriar, line 756: Let's also add to xterm.d.ts as xterm-headless is a subset of that

typings/xterm-headless.d.ts
```diff
@@ -748,6 +748,13 @@ declare module '@xterm/headless' {
      */
     onResize: IEvent<{ cols: number, rows: number }>;
 
+    /**
+     * Adds an event listener for when buffer rows change during parsing. The event
+     * value contains the range of rows that changed.
```
Tyriar, line 753:
```suggestion
     * Adds an event listener for when buffer rows change during parsing. The
     * event value contains the range of rows that changed. This is particularly
     * useful when implementing a custom renderer.
```

typings/xterm-headless.d.ts
Tyriar, line 756: I wonder if instead xterm-headless should just forward `onRequestRefreshRows` to the existing `onRender`? Would that accomplish what you're after?

src/headless/Terminal.ts
jim-lake: I didn't love the might be undefined or might be an object as the event spec, and I never needed the undefined case so I thought it made the API cleaner.  If I remember right another event gets fired on full repaint so its redundant.  I'm indifferent to changing the interface to { start, end}|undefined.

src/headless/public/Terminal.ts
jim-lake: The typescript compiler was complaining because the interfaces on headless and regular are different.  I'm not sure why it ever worked without an as cast.

Tyriar:
Thanks for the PR, I ended up making the change to onRender so you should be able to listen to that to accomplish this: https://github.com/xtermjs/xterm.js/pull/5504

jim-lake:
Perfect, thanks.

PR #5408, opened 2025-09-23
Fix firefox selection when in shadow dom

CGNonofr:
In firefox, if `getComputedStyle` is called on the wrong window, it only return an empty style

The consequence for the user is the inability to select anything in the terminal

This PR make it so the right window object is used, both in `Mouse.ts` (the actual bug I want to fix) and in the fit addon that seems to have the same issue (although I don't know how to test it exactly)

### DIFF
--- a/addons/addon-fit/src/FitAddon.ts
+++ b/addons/addon-fit/src/FitAddon.ts
@@ -23,6 +23,17 @@ interface ITerminalDimensions {
 const MINIMUM_COLS = 2;
 const MINIMUM_ROWS = 1;
 
+function getWindow(e: Node): Window {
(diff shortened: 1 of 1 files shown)


CGNonofr:
@Tyriar https://github.com/xtermjs/xterm.js/pull/5334 wasn't enough for firefox

CGNonofr:
rebased on master, and tests should pass now

Tyriar (commented):
This makes sense, I recreated the PR in https://github.com/xtermjs/xterm.js/pull/5535 so I could fix a stack overflow that was happening in the demo.

Issue #5373, opened 2025-06-30
scrollbar bug with alternate buffer


jerch:
Repro:
- run demo
- open vim, .e.g. call `vim package.json`
- scroll with the mouse several times up and down

--> At some point a scrollbar appears and messes further scrolling with the mouse totally up.

Expected:
On alternate buffer applications a scrollbar with a scrollbuffer handling should never appear.

Observations:
Seem I can only trigger the faulty behavior once. After exit to the shell and re-running vim I cannot provoke it again.

@Tyriar My first guess was, that something with the dimensions is off in the demo. But I can also repro it in vscode, so it actually might be related to the new scrollbar impl and/or related to the fake-scrolling with arrow keys (edited).

https://github.com/user-attachments/assets/6b08b627-e7b7-44a4-8e6e-801e612b157c

Tyriar:
@jerch can't reproduce this on master or https://github.com/xtermjs/xterm.js/pull/5411's branch. Do you have anything interesting set in .vimrc? Something I do observe is when vim mouse events are off then the mouse event's default is incorrectly not prevented so it also scrolls the page.

jerch:
@Tyriar Yes basically that - mouse events are off by default (I rarely use those with vim). What bugs me - why would the scrollbar appear at all for an alternate buffer app? Normally the alternate buffer should not have a scrollbuffer at all, so something is weird here.

Tyriar:
I mean I didn't see what you saw; no scroll bar at all but scrolling would scroll the web page.

jerch:
> I mean I didn't see what you saw; no scroll bar at all but scrolling would scroll the web page.

Hmm ok, thats weird. I gonna check again on master and test your PR, maybe it fixes things.

Tyriar:
Couldn't mac or Linux, both checking vim mouse events on or off via `set mouse=a`. Closing as this isn't actionable atm.

Issue #5339, opened 2025-05-15
Buffer position becomes desynced from the scroll bar


cs278:
https://github.com/microsoft/vscode/issues/224750

This issue has been annoying me for months, I've tried to find out what the status of getting it fixed is but there doesn't seem to be an issue here despite the VSCode issue being labelled as such. So perhaps it's slipped through the cracks and if I create this, it will get fixed? 🤞 

## Details
- Browser and browser version: 
- OS version: Ubuntu 22.04
- xterm.js version: No idea - presumably 5.5.0.

rob-myers:
> When scrolling through full-screen CLI app (git log or git diff), sometimes the terminal will not let me scroll up. If I try to scroll up, I am "teleported" back to previous position. As soon as I scroll down, I can't scroll back up to the previous text.

I've seen a similar issue: (a) while Terminal is hidden write many lines, (b) show Terminal and scroll up, (c) cannot scroll down to bottom.

I found a hacky fix:

```js
// on resize observer width > 0
xterm.resize(xterm.cols + 1, xterm.rows);
xterm.resize(xterm.cols - 1, xterm.rows);
```

Tyriar:
This should be fixed now, let us know if not.

PR #5411, opened 2025-10-01
Fix teleport when exiting out of alt buffer. v2

anthonykim1:
Had people try out: https://github.com/microsoft/vscode/issues/224750#issuecomment-3324948592

Apparently https://github.com/xtermjs/xterm.js/pull/5390 wasn't enough. 
If you really interact scroll around alt buffer, after maxing out in normal buffer, you can sometimes repro this if you get lucky. 

@Tyriar  The teleport issue came back, this time it was more difficult to repro, but I was able to repro.

Steps to repro for me was:
1.  fill in bunch of things in the normal buffer with `echo -l {1..10000}` 
2. enter alt buffer,
3. scroll very aggressively in alt buffer with+without content
4. Exit vim via `:q!` 
5. notice scroll is at the top again. You have to get pretty lucky. I noticed trying to max out my normal buffer, and also filling in content inside alt buffer & scrolling aggressively there, back <-> forth between helped increase the probability of repro. Sometimes had to launch vim multiple times and exit out re-doing the steps. 

-----------------------------------------------------
I spent some time and it seems like at certain point ` undefined  !== 1000` causes true for 
```
  if (ydisp !== this._latestYDisp) {
      this._scrollableElement.setScrollPosition({
        scrollTop: ydisp * this._renderService.dimensions.css.cell.height
      });
```
 when  ydisp is undefined after exiting out of alt buffer. Hence scroll will be brought to very top. 

* undefined above is likely coming from https://github.com/xtermjs/xterm.js/blob/a1d8e96fdd5ed93329b5af09106f5a3c6cd5740b/src/browser/Viewport.ts#L97

EDIT: ydisp cannot be undefined in sync due to https://github.com/xtermjs/xterm.js/pull/5411#discussion_r2395187542 
-----------------------------------------------------

Adding `this._latestYDisp = ydisp`in `_sync`  similar to `queueSync` seems to solve the problem.

Here are some logs PRE-PR:

(Entering vim)
enterAlt

(Exit out of vim) 
backtoNorm

(start scrolling) 
startScroll

### DIFF
--- a/src/browser/Viewport.ts
+++ b/src/browser/Viewport.ts
@@ -94,7 +94,12 @@ export class Viewport extends Disposable {
     }));
 
     this._register(this._bufferService.onResize(() => this.queueSync()));
-    this._register(this._bufferService.buffers.onBufferActivate(() => this.queueSync()));
+    this._register(this._bufferService.buffers.onBufferActivate(() => {
+      // Reset _latestYDisp when switching buffers to prevent stale scroll position
+      // from alt buffer contaminating normal buffer scroll position
+      this._latestYDisp = undefined;
+      this.queueSync();
+    }));
     this._register(this._bufferService.onScroll(() => this._sync()));
 
     this._register(this._scrollableElement.onScroll(e => this._handleScroll(e)));
@@ -160,7 +165,6 @@ export class Viewport extends Disposable {
     // If ydisp has been changed by some other component (input/buffer), then stop animating smooth
     // scroll and scroll there immediately.
     if (ydisp !== this._latestYDisp) {
-      this._latestYDisp = ydisp;
       this._scrollableElement.setScrollPosition({
         scrollTop: ydisp * this._renderService.dimensions.css.cell.height
       });


src/browser/Viewport.ts
```diff
@@ -157,9 +157,10 @@ export class Viewport extends Disposable {
     });
     this._suppressOnScrollHandler = false;
 
-    // If ydisp has been changed by some other copmonent (input/buffer), then stop animating smooth
+    // If ydisp has been changed by some other component (input/buffer), then stop animating smooth
     // scroll and scroll there immediately.
     if (ydisp !== this._latestYDisp) {
+      this._latestYDisp = ydisp;
```
Tyriar: Does it work if we set this only when `this._latestYDisp` is undefined? Normally `_latestYDisp` is set when handling scroll events, not syncing.

anthonykim1: I just tested this with conditional `this._latestYDisp is undefined` , and it's not working..

 `this._latestYDisp` switches from 0 to 1000  in `queueSync`, and this `this._latestYDisp` passes down and becomes  `ydisp` in `sync` where

```
 if (ydisp !== this._latestYDisp) {    // 1. this would be 0 !== 1000
      this._scrollableElement.setScrollPosition({
        scrollTop: ydisp * this._renderService.dimensions.css.cell.height   // 2. ydisp is 0.
      });
    }
```

Afterwards, when latestYDisp is 1000, and ydisp is also 1000, we don't get to call 
``` 
  scrollTop: ydisp * this._renderService.dimensions.css.cell.height  
``` 
because ` if (ydisp !== this._latestYDisp)` is false, and this would leave the scrollTop to remain 0, instead updating to correct value (end of buffer)
---> Hence you get stuck in 0 scrollTop position.

Also, _sync 's ydisp cannot actually be undefined since it has default value, from:
```
 private _sync(ydisp: number = this._bufferService.buffer.ydisp): void {
``` 

Long story short, 
In the "teleport" scenario, I think the the problem appears because we don't get to update scrollTop to the correct place when we go from having 0 to 1000 `ydisp` when calling `_sync` 



Maybe this is also some race condition, as I can only repro sometime

Tyriar: Does `this._bufferService.buffers.onBufferActivate` need to be handled specially? It seems like it's related to buffer switching during a smooth scroll happening. You could try increasing smooth scroll duration significantly (10000ms?) to see if it's easier to repro?

anthonykim1: Yeah, maybe more proper/intuitive fix is that we set latestYDisp to undefined onBufferActivate.
This way we prevent from any contamination when switching between buffers.

Setting the smooth scroll to 50000 and also aggressively scrolling and letting the mouse scroll run right before exiting out via `q!` seemed to help me repro more consistently.

jerch:
The new vs scroller integration still has some issues get updated correctly, maybe it is related to this here? Plz see #5373
 and #5339.

Issue #5417, opened 2025-10-17
Only log character codes for verbose logging


Tyriar:
These logs make it difficult to parse logs of output:

https://github.com/xtermjs/xterm.js/blob/e969832946a97791efbc3ad520e827b65729c0ba/src/common/services/CoreService.ts#L77-L88

We should just show the string if debug log level, and show this in verbose. We could also log the character codes in a separate log calls so they can more easily be filtered out.

PR #5420, opened 2025-10-17
Only log character codes for trace log

anthonykim1:
Resolves: https://github.com/xtermjs/xterm.js/issues/5417

### DIFF
--- a/src/common/services/CoreService.ts
+++ b/src/common/services/CoreService.ts
@@ -76,7 +76,7 @@ export class CoreService extends Disposable implements ICoreService {
 
     // Fire onData API
     this._logService.debug(`sending data "${data}"`);
-    this._logService.trace(`sending data "${data}", and their character codes:`, () => data.split('').map(e => e.charCodeAt(0)));
+    this._logService.trace(`sending data (codes)`, () => data.split('').map(e => e.charCodeAt(0)));
     this._onData.fire(data);
   }
 


src/common/services/CoreService.ts
```diff
@@ -75,15 +75,17 @@ export class CoreService extends Disposable implements ICoreService {
     }
 
     // Fire onData API
-    this._logService.debug(`sending data "${data}"`, () => data.split('').map(e => e.charCodeAt(0)));
+    this._logService.debug(`sending data "${data}"`);
+    this._logService.trace(`sending data "${data}", and their character codes:`, () => data.split('').map(e => e.charCodeAt(0)));
```
Tyriar:
```suggestion
    this._logService.debug(`sending data "${data}"`);
    this._logService.trace(`sending data (codes)"`, () => data.split('').map(e => e.charCodeAt(0)));
```

src/common/services/CoreService.ts
```diff
@@ -75,15 +75,17 @@ export class CoreService extends Disposable implements ICoreService {
...
     this._onData.fire(data);
   }
 
   public triggerBinaryEvent(data: string): void {
     if (this._optionsService.rawOptions.disableStdin) {
       return;
     }
-    this._logService.debug(`sending binary "${data}"`, () => data.split('').map(e => e.charCodeAt(0)));
+    this._logService.debug(`sending binary "${data}"`);
+    this._logService.trace(`sending binary "${data}", and their character codes:`, () => data.split('').map(e => e.charCodeAt(0)));
```
Tyriar:
```suggestion
    this._logService.trace(`sending binary (codes)`, () => data.split('').map(e => e.charCodeAt(0)));
```

src/common/services/CoreService.ts
anthonykim1: I think this one has extra " after `)` Will push a commit without the `"`
You didnt want the data along with the character codes, right?
Just the codes

PR #5425, opened 2025-10-22
Only show char codes for parsing data on trace

anthonykim1:
We previously did: https://github.com/xtermjs/xterm.js/pull/5420 to log character codes for `sending data` only on trace.
Should we do the same for `parsing data`? 
Otherwise, the debug level for terminal output still is filled with bunch of character codes.

### DIFF
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -444,7 +444,7 @@ export class InputHandler extends Disposable implements IInputHandler {
     }
 
     // Log debug data, the log level gate is to prevent extra work in this hot path
-    if (this._logService.logLevel === LogLevelEnum.DEBUG) {
+    if (this._logService.logLevel <= LogLevelEnum.DEBUG) {
       this._logService.debug(`parsing data ${typeof data === 'string' ? ` "${data}"` : ` "${Array.prototype.map.call(data, e => String.fromCharCode(e)).join('')}"`}`);
     }
     if (this._logService.logLevel === LogLevelEnum.TRACE) {


src/common/InputHandler.ts
```diff
@@ -444,12 +444,11 @@ export class InputHandler extends Disposable implements IInputHandler {
     }
 
     // Log debug data, the log level gate is to prevent extra work in this hot path
-    if (this._logService.logLevel <= LogLevelEnum.DEBUG) {
```
Tyriar, line 447: We should keep the ifs for both. We want to avoid this work when > debg/trace

src/common/InputHandler.ts
```diff
@@ -444,12 +444,11 @@ export class InputHandler extends Disposable implements IInputHandler {
...
-      this._logService.debug(`parsing data${typeof data === 'string' ? ` "${data}"` : ` "${Array.prototype.map.call(data, e => String.fromCharCode(e)).join('')}"`}`, typeof data === 'string'
-        ? data.split('').map(e => e.charCodeAt(0))
-        : data
-      );
-    }
+    this._logService.debug(`parsing data ${typeof data === 'string' ? ` "${data}"` : ` "${Array.prototype.map.call(data, e => String.fromCharCode(e)).join('')}"`}`);
+    this._logService.trace(`parsing data (codes)`, () => typeof data === 'string'
+      ? data.split('').map(e => e.charCodeAt(0))
+      : Array.from(data)
```
Tyriar: Didn't need Array.from before?

anthonykim1: Facts!!  You're right.. data is Uint8Array, doing Array.from(data) feels like complete waste. 
When we use logService to print, it would look identical, right? 
I dont think I ever saw `Uint8Array(5)` that prepends these characters in the log

src/common/InputHandler.ts
```diff
@@ -444,8 +444,11 @@ export class InputHandler extends Disposable implements IInputHandler {
     }
 
     // Log debug data, the log level gate is to prevent extra work in this hot path
-    if (this._logService.logLevel <= LogLevelEnum.DEBUG) {
-      this._logService.debug(`parsing data${typeof data === 'string' ? ` "${data}"` : ` "${Array.prototype.map.call(data, e => String.fromCharCode(e)).join('')}"`}`, typeof data === 'string'
+    if (this._logService.logLevel === LogLevelEnum.DEBUG) {
```
Tyriar:
```suggestion
    if (this._logService.logLevel <= LogLevelEnum.DEBUG) {
```

Issue #5438, opened 2025-11-07
Keyboard input duplicated on Android with physical keyboards


nya3jp:
## Details

- Browser and browser version: Chrome 143.0.7499.15
- OS version: Android 16
- xterm.js version: One that is bundled in the latest VSCode

### Steps to reproduce

1. Prepare an Android device with a physical keyboard
2. Open GitHub Codespace for an arbitrary project
3. Open the integrated terminal
4. Press two keys "J" and "K" simultaneously

**Expected result**: `jk` or `kj` appears on the terminal.

**Actual behavior**: `kjj` appears on the terminal.

PR #5439, opened 2025-11-07
Fix key duplication on Android with physical keyboards

nya3jp:
This is a fix for #5438.

When typing rapidly on Android with physical keyboards, characters often duplicate on the terminal. This is because physical keyboard events always go through IME on Android, even for plain English characters, and rapid typing causes a race condition in the composition helper.

This patch fixes the issue by making sure textarea changes are handled correctly.


Tyriar (approved):
Thanks, makes sense 👍

Issue #5509, opened 2025-12-27
The demo server should be written in TS


Tyriar:
Currently there are a bunch of ts-check errors as it's not checked in CI. It would make it easier to work with if it was an actual TS file.

PR #5512, opened 2025-12-27
Move server.js to TypeScript

Tyriar:
Fixes https://github.com/xtermjs/xterm.js/issues/5509


Tyriar:
Works from clean install on both Windows/macOS using:

```
npm ci
npm run setup
npm run dev
```

PR #5561, opened 2025-12-31
Fix allowTransparency in lib and demo

### LINKED ISSUE #5558: Issue: Transparent background not working in xterm.js 6.0.0
**Problem**: `theme.background: “transparent"` with `allowTransparency: true` renders as black after upgrading from `xterm.js` 5.5.0 to 6.0.0, setting a transparent background no longer works. The terminal renders with a black background instead.

**Expected behavior**: Terminal background should be transparent, allowing the parent container's background to show through.

**Actual behavior**: Terminal background is black (`rgb(0, 0, 0)`). Inspecting the DOM shows an inline style `background-color: rgb(0, 0, 0)` on the `.xterm-scrollable-element` `div`.

**Workaround**: Use an explicit hex color matching the container background instead of "transparent".

## Details
- OS version: macOS 26.2
- xterm.js version: 6.0.0

### Steps to reproduce

```ts
const terminal = new Terminal({
  allowTransparency: true,
  theme: {
    background: "transparent",
  },
});
```

### LINKED ISSUE #5560: Toggling allowTransparency in the demo will hide all text until the termianl is focused
Image

Image

Tyriar:
Fixes #5558
Fixes #5560

New checkered background when this option is set:

image

### DIFF
--- a/css/xterm.css
+++ b/css/xterm.css
@@ -91,8 +91,6 @@
 }
 
 .xterm .xterm-viewport {
-    /* On OS X this is required in order for the scroll bar to appear fully opaque */
-    background-color: #000;
     overflow-y: scroll;
     cursor: default;
     position: absolute;
@@ -100,6 +98,11 @@
     left: 0;
     top: 0;
     bottom: 0;
+}
+
+.xterm:not(.allow-transparency) .xterm-viewport {
+    /* On OS X this is required in order for the scroll bar to appear fully opaque */
+    background-color: #000;
 }
 
 .xterm .xterm-screen {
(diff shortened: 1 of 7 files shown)


css/xterm.css
```diff
@@ -91,8 +91,6 @@
 }
 
 .xterm .xterm-viewport {
-    /* On OS X this is required in order for the scroll bar to appear fully opaque */
-    background-color: #000;
```
amanape, line 95: Unchanged for 9 years until now. This was older than my professional career 😢

Issue #5570, opened 2026-01-03
OOM crash when enabling ligatures when using CommitMono


Tyriar:
VS Code: https://github.com/microsoft/vscode/issues/239541

Repro:

1. Open 127.0.0.1:3000
2. Options tab: Change fontFamily to CommitMono
3. Addons tab: Check ligatures, 🐛 frozen -> OOM

<img width="713" height="458" alt="Image" src="https://github.com/user-attachments/assets/9bc37b93-6e00-42c9-94e0-0a215784ea71" />

PR #5573, opened 2026-01-03
Fix infinite recursion OOM crash in addon-ligatures 

Tyriar:
1. processLookahead/BacktrackPosition created separate entry objects
   for each glyph at each position (O(N^M) with N glyphs, M positions).
   Now all glyphs at a position share one entry object.
2. cloneEntry, cloneTree, flattenEntry, and mergeTreeEntry could
   infinitely recurse or duplicate work on shared/cyclic references.
   Added visited/merged tracking maps to cache results.

Note that parsing CommitMono locks up the renderer for some time still
so there is still a performance issue here, this fixes the OOM though.
This was mostly defensive generated code since the codebase is
unfamiliar to me.

Fixes https://github.com/xtermjs/xterm.js/issues/5570

Timeline of the perf problem:

image

### DIFF
--- a/addons/addon-ligatures/src/fontLigatures/flatten.ts
+++ b/addons/addon-ligatures/src/fontLigatures/flatten.ts
@@ -1,13 +1,13 @@
 import { ILookupTree, IFlattenedLookupTree, ILookupTreeEntry, IFlattenedLookupTreeEntry } from './types';
 
-export default function flatten(tree: ILookupTree): IFlattenedLookupTree {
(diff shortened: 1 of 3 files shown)


ScreamZ:
ETA on VSCODE ?

Tyriar:
@ScreamZ should be fixed in insiders, coming to stable in a week or so.

ScreamZ:
Awesome ! Just formated my computer to try to fix this bug when i got pinged for this issue ! Thanks for fixing

PR #5616, opened 2026-01-19
Fix IME composition overflow at right edge

### LINKED ISSUE #5454: Chinese IME shows input far from cursor when using AI CLIs
VS Code issue: https://github.com/microsoft/vscode/issues/282621

Image

notxx:
When the cursor is at the right-most column and IME composition text is long, the composition-view element can extend beyond the terminal boundary, causing the container to stretch horizontally.

This fix adds two layers of protection:

1. CSS: Add overflow: hidden to .xterm and .xterm .xterm-screen to prevent container stretching by internal elements.

2. Core logic: Enhance CompositionHelper.updateCompositionElements() to detect when the composition view would overflow the terminal width. When overflow is detected, the composition view is moved left and down one line to keep it within bounds, while maintaining IME candidate window alignment.

The fix includes unit tests to verify the overflow detection and position correction behavior.

Fixes #5454
Related to microsoft/vscode#282621


Tyriar:
Should be fixed by https://github.com/xtermjs/xterm.js/pull/5759 and https://github.com/xtermjs/xterm.js/pull/5747

Issue #5592, opened 2026-01-06
Implement Kitty graphics protocol support


Tyriar:
https://sw.kovidgoyal.net/kitty/graphics-protocol/

This should probably live in an addon as it's an extension. We currently have an image addon which includes the sixel support.

PR #5617, opened 2026-01-21
Add register apc handler 

anthonykim1:
Part of: https://github.com/xtermjs/xterm.js/issues/5592 

Add registerAPCHandler to parser and expose in API, add test.
This is the first part toward kitty graphics protocol.

### DIFF
--- a/src/common/parser/EscapeSequenceParser.ts
+++ b/src/common/parser/EscapeSequenceParser.ts
@@ -76,7 +76,7 @@ const NON_ASCII_PRINTABLE = 0xA0;
  * VT500 compatible transition table.
  * Taken from https://vt100.net/emu/dec_ansi_parser.
  */
-export const VT500_TRANSITION_TABLE = (function(): TransitionTable {
+export const VT500_TRANSITION_TABLE = (function (): TransitionTable {
   const table: TransitionTable = new TransitionTable(4095);
 
   // range macro for byte


src/common/parser/Constants.ts
```diff
@@ -41,7 +42,10 @@ export const enum ParserAction {
   CLEAR = 11,
   DCS_HOOK = 12,
   DCS_PUT = 13,
-  DCS_UNHOOK = 14
+  DCS_UNHOOK = 14,
+  APC_START = 15,
+  APC_PUT = 16,
+  APC_END = 17
```
jerch, line 48: For values greater bitlength:4 the transition table has to be changed. It currently works on uint8 storing 2 values in a single slot as |4|4| - quick fix would be to use uint16 instead and change the offsets here accordingly: https://github.com/xtermjs/xterm.js/blob/03b015d5d3298251c57578290cf1e0f0d54f69ff/src/common/parser/EscapeSequenceParser.ts#L20-L24

With uint16 you can store |8|8|, which should be sufficient for any upcoming state|action additions in the future.

anthonykim1: Thank you! Does [1c73006](https://github.com/xtermjs/xterm.js/pull/5617/commits/1c73006e39ee70332f8c48fa521bddf5a33df64d) seem good?

src/common/parser/EscapeSequenceParser.test.ts
```diff
@@ -1207,7 +1221,7 @@ describe('EscapeSequenceParser', () => {
       clearAccu();
     });
     it('print handler', () => {
-      parser2.setPrintHandler(function (data: Uint32Array, start: number, end: number): void {
+      parser2.setPrintHandler(function(data: Uint32Array, start: number, end: number): void {
```
Tyriar: You have some formatter on?

Tyriar (approved):
All seems pretty reasonable. Make sure you check the benchmark before and after just to make sure this didn't break perf somehow.

src/common/parser/Constants.ts
jerch: Yupp looks good to me :+1: 

I think you should add ParserState.APC_STRING in this list to get properly tested too: https://github.com/xtermjs/xterm.js/blob/b01ba7db28ce9fd10d103d2bbcf6f9fa3db5d85c/src/common/parser/EscapeSequenceParser.test.ts#L151-L161
(This unit test tests all table transitions, so any new state has to be added there to truly cover all states.)

src/common/parser/EscapeSequenceParser.test.ts
anthonykim1: I thought I just had typescript+js one, but maybe prettier is kicking in..? Let me revert some of these weird formatting ones

anthonykim1:
Performance metrics going from master (uint8) -> this current PR branch (uint16):
Overall total: 4060.18 MB/s    vs. 4090.38 MB/s      
----> +0.74%
[benchmark_results.txt](https://github.com/user-attachments/files/24782581/benchmark_results.txt)

AI generated Python script to calculate display results in terminal (You probably need to adjust the path if you want to test it on your local machine): 
[gen_results.py](https://github.com/user-attachments/files/24782591/gen_results.py)

src/common/parser/Constants.ts
anthonykim1: Thank you for this🙏
I just realized I need `APC_STRING + 1` instead of `DCS_PASSTHROUGH + 1` on states since APC_STRING is now the last element in ParserState.  Handled: [b286f21](https://github.com/xtermjs/xterm.js/pull/5617/commits/b286f21f649d432d8942751d242f9125556c3de6)

jerch: Oh wow, had overlooked this, too. Maybe create a symbol ParserState.STATE_LENGTH for this as last entry with a comment, that states have to be continuous and that entry denotes LAST_STATE+1? (This stuff is very C-ish, sorry for that :smile_cat: )

anthonykim1: Thanks! This repo is fun, I love learning here. Just created https://github.com/xtermjs/xterm.js/pull/5618

PR #5619, opened 2026-01-22
Support Kitty graphics protocol mvp

anthonykim1:
Part of https://github.com/xtermjs/xterm.js/issues/5592 

WIP 

See discussion for components in spec, their progress/discussion: https://github.com/xtermjs/xterm.js/discussions/5683 

Plan:

1. (Done with https://github.com/xtermjs/xterm.js/pull/5617) First PR: Add registerAPCHandler to parser and expose in API, add tests 
2. Scaffold addon:
    * Create addon-kitty-graphics, copy structure from other addons
    * Hook up in demo client.ts
        * Search addon-progress and ProgressAddon for how to do this
    * Get it working and activating
        * (Checked, remember to remove comments) Add a console.log to the addon's activate function
3. Come up with a simple test command to verify in kitty or ghostty (see send-png)
4. Set up APC handler, get it to trigger with the test
5. Create a playwright test (test folder) in the addon that automates it for a simple image
    * Use single 1x1 black pixel png image, write it to the terminal in your test, verify the image we see is black
    * Use 3x1 png image, 255 red first, 255 green second, 255 blue third
6. Understand what APC sequences are being sent by the program
7. Implement the handlers  
8. Draw to a canvas layer
9. Make sure we get the playwright tests to pass.

(Jan 22, 4:00 PM PST) There's a problem currently where kitty kitten +icat wont return image but using send-image file will. 

Copilot's research on why normal kitty kitten +icat doesnt work on xterm demo rn: 
```
kitty icat uses TIOCGWINSZ ioctl to get pixel dimensions from the PTY before sending any graphics. node-pty doesn't set pixel values in that ioctl, so it returns 0. kitty icat sees 0 pixels and errors out before even trying the graphics protocol.
```

(Jan 22, 6:30 PM PST) 
Things like: `kitty +kitten icat --use-window-size=80,24,1280,768 --transfer-mode=stream ~/Desktop/xterm.js/demo/logo.png` work atm

~TODO since image goes on top of text rn 🥶: 
```
Calculate how many rows/cols the image spans
Move the cursor accordingly (or reserve blank space)
```
--> Handled by image storage and c=1, c=0 logic. See discussion page. 

TODOs
- Share decoder instance: https://github.com/xtermjs/xterm.js/pull/5619#discussion_r2768463065 
- Get the MVP out on feb!!

### DIFF
--- a/demo/client/tsconfig.json
+++ b/demo/client/tsconfig.json
@@ -19,12 +19,13 @@
       "@xterm/addon-web-links": ["../../addons/addon-web-links"],
       "@xterm/addon-webgl": ["../../addons/addon-webgl"],
       "@xterm/addon-unicode11": ["../../addons/addon-unicode11"],
-      "@xterm/addon-unicode-graphemes": [
-        "../../addons/addon-unicode-graphemes"
-      ],
+      "@xterm/addon-unicode-graphemes": ["../../addons/addon-unicode-graphemes"],
       "@xterm/addon-ligatures": ["../../addons/addon-ligatures"],
       "*": ["./*"]
     }
   },
-  "include": ["./**/*", "../../typings/xterm.d.ts"]
+  "include": [
+    "./**/*",
+    "../../typings/xterm.d.ts"
+  ]
 }


jerch:
> ...node-pty doesn't set pixel values in that ioctl...

Imho this should be fixed in node-pty to populate the pixel dimensions where possible (I think there was an issue with conpty not providing it, but for other platforms it can be populated)

Another workaround for console apps is to use WinOps sequences like CSI 14t (see here for an example https://github.com/jerch/sixel-puzzle/blob/56129538af70fa4ec9441d1d3553398dc8e66f1f/termlib.py#L95) But of couse this is beyond xterm.js and has to be implemented by the console app itself.

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+  private async _decodeImage(image: IKittyImage): Promise<ImageBitmap> {
+    const format = image.format;
+    const base64Data = image.data as string;
+
+    // Decode base64 to binary
+    const binaryString = atob(base64Data);
+    const bytes = new Uint8Array(binaryString.length);
+    for (let i = 0; i < binaryString.length; i++) {
+      bytes[i] = binaryString.charCodeAt(i);
+    }
```
jerch: This has very bad runtime and is a nightmare creating memory pressure with huge images.

You might want to check the wasm based base64 decoder I used for the IIP protocol. It works on chunks, thus should be much friendlier to chunked data ingestion.

For basic usage see https://github.com/jerch/xterm-wasm-parts, for chunked usage addons/addon-image/src/IIPHandler.ts holds the needed bits.

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+
+    if (format === KittyFormat.PNG) {
+      // PNG: create blob and decode
+      const blob = new Blob([bytes], { type: 'image/png' });
+      return createImageBitmap(blob);
```
jerch: I dont know if this is still an issue - older Safari versions could not do this and needed a workaround with `new Image()` instead. (also see IIP handler)

jerch:
@anthonykim1 I start to wonder, if the kitty graphics handler should be integrated into the image addon. I don't know the requirements of this protocol for the render layering, but maybe things could be reused and unified? It would make the lifecycling easier and would not introduce another image layer.

But as I said - i dont know the kitty protocol requirements, so fusing it might be a futile attempt. Thoughts?

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+
+    if (this._debug) {
+      console.log('[KittyGraphicsAddon] Registering APC handler for G (0x47)');
+    }
+
+    // Register APC handler for 'G' (0x47) - Kitty graphics protocol
+    // APC sequence format: ESC _ G <data> ESC \
+    this._apcHandler = terminal.parser.registerApcHandler(0x47, (data: string) => {
+      return this._handleKittyGraphics(data);
+    });
```
jerch: Since image data tend to be much bigger than 10 MB, the string based handler interface is not a good choice here (the string handler interface is limited to 10 MB of string data). Also the string forth and back conversion and aggregation exhibits really bad runtime here. Better - implement a full chunk based handler class and register an instance of it directly on the parser. This is much more efficient, as the utf32 codepoints emitted by the parser can be handled borrowed without copying them. (See the IIP and Sixel handlers for how to create a proper chunk handler class.)

addons/addon-kitty-graphics/src/Types.ts
```diff
@@ -0,0 +1,29 @@
...
+  coreService: ICoreService;
+}
+
+/**
+ * Extended terminal interface that exposes the internal _core property.
+ * This is needed to send responses back to the client.
+ */
+export interface ITerminalExt extends Terminal {
+  _core: ICoreTerminal;
+}
```
Tyriar: You can use `Terminal.input(data, false)` now for this

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+      case 'f': cmd.format = parseInt(value); break;
+      case 'i': cmd.id = parseInt(value); break;
+      case 's': cmd.width = parseInt(value); break;
+      case 'v': cmd.height = parseInt(value); break;
+      case 'x': cmd.x = parseInt(value); break;
+      case 'y': cmd.y = parseInt(value); break;
+      case 'c': cmd.columns = parseInt(value); break;
+      case 'r': cmd.rows = parseInt(value); break;
+      case 'm': cmd.more = parseInt(value); break;
+      case 'q': cmd.quiet = parseInt(value); break;
```
Tyriar: Const enums for these keys, then we can describe them in the symbol name and elaborate in jsdocs

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+  for (const part of parts) {
+    const eqIdx = part.indexOf('=');
+    if (eqIdx === -1) continue;
+
+    const key = part.substring(0, eqIdx);
+    const value = part.substring(eqIdx + 1);
+
+    switch (key) {
+      // These are key(s) from: https://sw.kovidgoyal.net/kitty/graphics-protocol/#control-data-reference
+      // Question: How do we know which radix to use?
```
Tyriar: Base 10 I think? It'll blow up if there are any a-f

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+      // TODO: Support more keys like type of compression, animations, etc
+      case 'a': cmd.action = value; break;
+      case 'f': cmd.format = parseInt(value); break;
+      case 'i': cmd.id = parseInt(value); break;
```
Tyriar: Handle `'a'` in an if first, return early, and then `parseInt` before the switch?

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
Tyriar: This could live in a KittyApcHandler.ts

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+        return this._handleQuery(cmd);
+      case KittyAction.DELETE:
+        return this._handleDelete(cmd);
+      default:
+        return true;
+    }
+  }
+
+  private _handleTransmit(cmd: IKittyCommand): boolean {
+    const id = cmd.id || this._nextImageId++;
```
Tyriar: Prefer `??` in general as it prevents problems like if `id` was 0

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+
+    const bytesPerPixel = format === KittyFormat.RGBA ? 4 : 3;
+    const expectedBytes = width * height * bytesPerPixel;
+
+    if (bytes.length < expectedBytes) {
+      throw new Error(`Insufficient pixel data: got ${bytes.length}, expected ${expectedBytes}`);
+    }
+
+    // Convert to RGBA ImageData
+    const imageData = new ImageData(width, height);
```
Tyriar: Prefer `new ImageData(dataArray, width, height)`, so assemble the array and pass it in instead

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+    const data = imageData.data;
+
+    for (let i = 0; i < width * height; i++) {
+      const srcOffset = i * bytesPerPixel;
+      const dstOffset = i * 4;
```
Tyriar: Pull `let srcOffset = 0` and `let dstOffset = 0` just incase JIT doesn't optimize

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+
+      data[dstOffset] = bytes[srcOffset];         // R
+      data[dstOffset + 1] = bytes[srcOffset + 1]; // G
+      data[dstOffset + 2] = bytes[srcOffset + 2]; // B
+      data[dstOffset + 3] = format === KittyFormat.RGBA
```
Tyriar: Avoid `format === KittyFormat.RGBA` inside the hot loop

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+    const imageData = new ImageData(width, height);
+    const data = imageData.data;
+
+    for (let i = 0; i < width * height; i++) {
+      const srcOffset = i * bytesPerPixel;
+      const dstOffset = i * 4;
+
+      data[dstOffset] = bytes[srcOffset];         // R
+      data[dstOffset + 1] = bytes[srcOffset + 1]; // G
+      data[dstOffset + 2] = bytes[srcOffset + 2]; // B
```
Tyriar:
```suggestion
      data[dstOffset    ] = bytes[srcOffset    ]; // R
      data[dstOffset + 1] = bytes[srcOffset + 1]; // G
      data[dstOffset + 2] = bytes[srcOffset + 2]; // B
```

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,462 @@
...
+    // Validate the image data can be decoded
+    try {
+      // Decode base64 to verify it's valid
+      const binaryString = atob(payload);
+      const bytes = new Uint8Array(binaryString.length);
+      for (let i = 0; i < binaryString.length; i++) {
+        bytes[i] = binaryString.charCodeAt(i);
+      }
+
+      const format = cmd.format || 32;
```
Tyriar: Replace magic numbers with `KittyFormat.blah`

addons/addon-kitty-graphics/src/KittyImageRenderer.ts
```diff
@@ -0,0 +1,237 @@
...
+   */
+  public ensureCanvasLayer(): void {
+    if (this._canvas) {
+      return;
+    }
+
+    // Access internal screenElement
+    const core = (this._terminal as any)._core;
+    const screenElement = core?.screenElement;
+    if (!screenElement) {
```
Tyriar: We should make a new API that exposes this, we use it in VS Code too.

```ts
readonly screenElement: HTMLElement | undefined;
```

You can do this in a PR before this is merged, then update to use it in VS Code where we touch screenElement.

addons/addon-kitty-graphics/src/KittyImageRenderer.ts
```diff
@@ -0,0 +1,237 @@
...
+      console.warn('[KittyGraphicsAddon] Cannot create canvas: no screenElement');
+      return;
+    }
+
+    // Get dimensions from terminal
+    const dimensions = (this._terminal as any).dimensions;
+    const width = dimensions?.css?.canvas?.width || 800;
+    const height = dimensions?.css?.canvas?.height || 600;
```
Tyriar: Use `Terminal.dimensions` instead

addons/addon-kitty-graphics/src/KittyImageRenderer.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    // Handle resize
+    this._resizeDisposable = this._terminal.onResize(() => this._resizeCanvas());
+  }
+
+  /**
+   * Get cell dimensions from terminal.
+   */
+  public getCellSize(): { width: number, height: number } {
+    const dimensions = (this._terminal as any).dimensions;
```
Tyriar: `Terminal.dimensions`

addons/addon-kitty-graphics/src/KittyImageRenderer.ts
```diff
@@ -0,0 +1,237 @@
...
+    const height = dimensions?.css?.canvas?.height || 600;
+
+    if (this._canvas.width !== width || this._canvas.height !== height) {
+      this._canvas.width = width;
+      this._canvas.height = height;
+      this._draw();
+    }
+  }
+
+  public dispose(): void {
```
Tyriar: I always put dispose just below ctor

addons/addon-kitty-graphics/test/KittyGraphicsAddon.test.ts
```diff
@@ -0,0 +1,269 @@
...
+ */
+
+import test from '@playwright/test';
+import { readFileSync } from 'fs';
+import { deepStrictEqual, strictEqual } from 'assert';
+import { ITestContext, createTestContext, openTerminal, timeout } from '../../../test/playwright/TestUtils';
+
+// Load test images as base64
+const BLACK_1X1_BASE64 = readFileSync('./addons/addon-kitty-graphics/fixture/black-1x1.png').toString('base64');
+const RGB_3X1_BASE64 = readFileSync('./addons/addon-kitty-graphics/fixture/rgb-3x1.png').toString('base64');
```
Tyriar: You also want tests for f=24 and 32

send-png
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+# TODO: Consider removing when Kitty Graphics Addon is shipped.
+"""
+Minimal Kitty Graphics Protocol image sender.
+Based on: https://sw.kovidgoyal.net/kitty/graphics-protocol/#a-minimal-example
```
Tyriar: Should remove this file

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,465 @@
...
+      return true;
+    }
+
+    // Validate the image data can be decoded
+    try {
+      // Decode base64 to verify it's valid
+      const binaryString = atob(payload);
+      const bytes = new Uint8Array(binaryString.length);
+      for (let i = 0; i < binaryString.length; i++) {
+        bytes[i] = binaryString.charCodeAt(i);
```
Tyriar: Replace magic numbers with `KittyFormat.blah`

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
```diff
@@ -0,0 +1,465 @@
...
+    if (!width || !height) {
+      throw new Error('Width and height required for raw pixel data');
+    }
+
+    const bytesPerPixel = format === KittyFormat.RGBA ? 4 : 3;
+    const expectedBytes = width * height * bytesPerPixel;
+
+    if (bytes.length < expectedBytes) {
+      throw new Error(`Insufficient pixel data: got ${bytes.length}, expected ${expectedBytes}`);
+    }
```
Tyriar: Prefer `new ImageData(dataArray, width, height)`, so assemble the array and pass it in instead

addons/addon-kitty-graphics/src/KittyApcHandler.ts
```diff
@@ -0,0 +1,614 @@
...
+   * Decode base64 image data into an ImageBitmap.
+   */
+  private async _decodeImage(image: IKittyImage): Promise<ImageBitmap> {
+    const format = image.format;
+    const base64Data = image.data as string;
+
+    // TODO: This atob + charCodeAt loop has bad runtime and creates memory pressure with large
+    // images. Consider using the wasm-based base64 decoder from xterm-wasm-parts which also
+    // supports chunked data ingestion. See addon-image/src/IIPHandler.ts for chunked usage.
+    const binaryString = atob(base64Data);
```
jerch: Note on this - the wasm based base64 decoder works best, if the decoded payload size is known (it does a sanity check with this). Unfortunately the kitty protocol does not announce the expected byte size, which is bad as the payload can be any size. In JS we have only 2GB at hand, therefore I strongly suggest to cap the payload at a certain size (the image addon does this at 20MB for IIP data payload and 16M pixels by default, both is adjustable)

For the decoder it means, that you prolly should do the base64 decoding in 1 MB shards (1048576 --> 786432) and add those shards manually to the full byte data.

addons/addon-kitty-graphics/src/KittyApcHandler.ts
```diff
@@ -0,0 +1,614 @@
...
+    let srcOffset = 0;
+    let dstOffset = 0;
+    for (let i = 0; i < pixelCount; i++) {
+      data[dstOffset    ] = bytes[srcOffset    ]; // R
+      data[dstOffset + 1] = bytes[srcOffset + 1]; // G
+      data[dstOffset + 2] = bytes[srcOffset + 2]; // B
+      data[dstOffset + 3] = isRgba ? bytes[srcOffset + 3] : ALPHA_OPAQUE;
+      srcOffset += bytesPerPixel;
+      dstOffset += BYTES_PER_PIXEL_RGBA;
+    }
```
jerch: This is still subpar - for RGBA data you can simply use the bytes directly, for RGB interleaving with the 4th byte in uint32 blocks might be faster, schematically:
```typescript
if (format === KittyFormat.RGBA) {
  imgData = new ImageData(bytes, width, height);
} else {
  const bytes32 = new Uint32Array(BYTES.buffer);
  const data32 = new Uint32Array(DATA.buffer);
  let dstOffset = 0;
  let srcOffset = 0;
  // assuming little endian, PIXELS must be 4-aligned
  for (let i = 0; i < PIXELS; i += 4) {
    const bloc1 = bytes32[srcOffset++];
    const bloc2 = bytes32[srcOffset++];
    const bloc3 = bytes32[srcOffset++];
    data32[dstOffset++] = bloc1 | 0xFF000000;
    data32[dstOffset++] = (bloc1 >> 24) | (bloc2 << 8) | 0xFF000000;
    data32[dstOffset++] = (bloc2 >> 16) | (bloc3 << 16) | 0xFF000000;
    data32[dstOffset++] = (bloc3 >> 8) | 0xFF000000;
  }
  // handle leftover here...
}
```
This is 28 vs. 7 reads/writes per 4 pixels (memory access is a lot heavier than the bit manipulations).

On a sidenote - this byte interleaving is a typical task where wasm-simd would outperform by far (can do it with 3 instructions - load, swizzle, store).

---

Edit: Fixing the bit shifts. Timing is 60 ms vs. 10 ms for 4M pixels (speedup 5-6x, or 0.2 GB/s vs. 1.3 GB/s, wasm-simd would reach here >4GB/s).

addons/addon-kitty-graphics/src/KittyApcHandler.ts
Tyriar: @anthonykim1 let's move it into addon-image for this, just make sure all the kitty stuff is clearly separated in a separate folder or something

addons/addon-kitty-graphics/src/KittyApcHandler.ts
jerch: Here is a gist of different versions: https://gist.github.com/jerch/e2f7695b887228d9e703c53af23c0737

Results:
```bash
$> node lib/test.wasm.js
original 50.4 ms, 250 MB/s
slightly optimized 23.9 ms, 526 MB/s
4 byte blocks 8.1 ms, 1553 MB/s
wasm simd 2.65 ms, 4748 MB/s
```

anthonykim1: This is crazy.. 50 -> 2.6 🤯🤩🔥

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
anthonykim1: It seems to work on my work mac, safari: `Version 26.2 (21623.1.14.11.9)` . Will check again with my personal mac at home that should be more outdated!

anthonykim1:
> I start to wonder, if the kitty graphics handler should be integrated into the image addon. I don't know the requirements of this protocol for the render layering, but maybe things could be reused and unified? It would make the lifecycling easier and would not introduce another image layer.
But as I said - i dont know the kitty protocol requirements, so fusing it might be a futile attempt. Thoughts?

@jerch Thank you for the feedback :) 
Will give this a shot! I started moving things around, but not done yet 🥸

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
anthonykim1: Thanks for this. Am I going in right direction with: [bae314f](https://github.com/xtermjs/xterm.js/pull/5619/commits/bae314f67c7d0ba4d031f22d4a4ecabd49e2ffef)

jerch: Following this link https://sw.kovidgoyal.net/kitty/graphics-protocol/#the-graphics-escape-code the base form of all kitty sequences is

```text
<ESC>_G<control data>;<payload><ESC>\
```

so the delimiter between control data and payload is a semicolon. Next you should clarify, whether the semicolon is an allowed char in control data itself. If not, then the separation of control data vs. payload is unique and you can just scan for the first semicolon in the byte stream. If a semicolon is allowed in control data, then the separation gets more ugly - prolly the `=` can be used as sentinel, it will not occur in base64 beside at the very end for padding.

About the `put` method - it gets called for every single chunk of sequence payload (here `<control data>;<payload>`). It is important to know, that the data can be split arbitrarily (worst case - one byte per `put` invocation) - so you have to collect the data.

Now to solve the separation of control data vs. base64 payload - I would assign a static Uint32Array to copy chunks over until the semicolon was found, schematically:

```typescript
private _inControlCode = true; // must be reset in start()
private _controlData = new Uint32Array(4096);  // or another size that exceeds expectations
private _controlLength = 0; // must be reset in start()

public put(data: Uint32Array, start: number, end: number): void {
  if (this._aborted) return;

  // perf note: in C/C++ the following then-else branches should be flipped
  // as the payload branch is more likely (no clue if it makes a difference in JS)
  if (this._inControlCode) {
    let controlEnd = end;
    // scan for payload delimiter
    for (let i = start, i < end; ++i) {
      if (data[i] === 0x3b) {
        // ; was found, so here we have all control code data
        this._inControlCode = false;
        controlEnd = i;
        break;
      }
    }
    // copy control data over
    this._controlData.set(data.subarray(start, controlEnd), this._controlLength);
    this._controlLength += controlEnd - start;
    if (!this._inControlCode) {
      // this._controlData now contains all control data, as we finished it above
      this._parseControlData();
      // init the base64 decoder
      this._b64Decoder.init(786432);
      // we still may have base64 payload in the chunk
      // +1 is needed to skip the ; itself
      if (controlEnd+1 < end) {
        if (this._b64Decoder.put(data, controlEnd+1, end)) {
          this._aborted = true;
        }
      }
    }
  } else {
    // base64 payload
    if (this._b64Decoder.put(data, start, end)) {
      this._aborted = true;
    }
  }
}
```

Now this 1MB sharding makes the usage of the base64 decoder a bit more cumbersome. I suggest to wrap the `b64Decoder.put` calls to do the tracking transparently, schematically:

```typescript
private _shardSize = 0;
private _b64put(data: Uint32Array, start: number, end: number): boolean {
  // is there still room in the base64 shard?
  const free = BASE64_SHARD_SIZE - this._shardSize;
  if (free >= end - start) {
    this._shardSize += end - start;
    return this._b64Decoder.put(data, start, end);
  }
  // fill the shard
  if (this._b64Decoder.put(data, start, start + free)) {
    return true;
  }
  // grab the decoded data
  this._b64Decoder.end();
  const chunk = this._b64Decoder.data8;  // borrowed!
  // --> store the chunk somehow...

  // handle leftover data:
  // - restart decoder for next shard decoding
  // - feed data to decoder
  this._b64Decoder.init(786432);
  this._shardSize = 0;
  return this._b64Decoder.put(data, start + free, end));
}
```

This last snippet also shows what to do in the `end` method:
```typescript
public end(success: boolean): boolean | Promise<boolean> {
  if (this._aborted) return true;

  // finalize last base64 shard
  this._b64Decoder.end();
  const chunk = this._b64Decoder.data8;  // borrowed!
  // --> store the chunk somehow...

  if (success) {
    // process image data chunks further here...
  }
  // cleanup chunks here...
  return true;
```

jerch: Just saw that the IIP handler is still on an old version of the base64 decoder. It refers to the very first commit in this repo [9a0692b](https://github.com/jerch/xterm-wasm-parts/tree/9a0692b3d0e4962c847bb94d88d8980b23ee1d8b). Please dont change this yet, as there were changes in API. Seem I need to cleanup this...

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
jerch: There is a high chance that this is finally fixed. Last I checked was ~4 years ago and Safari already had it implemented by that time, but deactivated behind a dev toggle.

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
jerch: Just saw, that there are also sequences, where the payload is omitted or not base64 encoded. For not base64 payload you basically would need to eval the control data before spinning up the base64 stuff.

For sequences without any payload the semicolon as delimiter is also absent. To properly grab those sequences you can do it in the `end` method, schematically:

```typescript
public end(success: boolean): boolean | Promise<boolean> {
  if (this._aborted) return true;

  // sequence without any payload
  if (this._inControlCode) {
    if (!success) return true;
    this._parseControlData();
    // exec the control commands if legit
    ...
    return true;
  }

  // payload handling goes here
  ...
}
```

jerch:
Sidenote on base64 decoding - ECMA2026 brings `Uint8Array.frombase64` & `Uint8Array.setfromBase64` (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64)

Need to re-eval things for the base64 decoder + encoder, on a first glance they are reasonably fast enough (tested with node 25):
```text
   Context "lib/base64/base64.benchmark.js"
      Context "Base64 - decode"
         Context "Node - Buffer"
            Case "256" : 100 runs - average throughput: 56.09 MB/s
            Case "4096" : 100 runs - average throughput: 465.91 MB/s
            Case "65536" : 100 runs - average throughput: 1215.26 MB/s
            Case "1048576" : 100 runs - average throughput: 1663.89 MB/s
         Context "Node - Uint8Array.fromBase64"
            Case "256" : 100 runs - average throughput: 47.96 MB/s
            Case "4096" : 100 runs - average throughput: 613.69 MB/s
            Case "65536" : 100 runs - average throughput: 1398.99 MB/s
            Case "1048576" : 100 runs - average throughput: 1373.99 MB/s
         Context "Base64Decoder"
            Case "256" : 100 runs - average throughput: 72.56 MB/s
            Case "4096" : 100 runs - average throughput: 586.40 MB/s
            Case "65536" : 100 runs - average throughput: 1477.00 MB/s
            Case "1048576" : 100 runs - average throughput: 1515.69 MB/s
      Context "Base64 - encode"
         Context "Node - Buffer"
            Case "256" : 100 runs - average throughput: 30.34 MB/s
            Case "4096" : 100 runs - average throughput: 357.81 MB/s
            Case "65536" : 100 runs - average throughput: 1451.27 MB/s
            Case "1048576" : 100 runs - average throughput: 1992.45 MB/s
         Context "Node - Uint8Array.toBase64"
            Case "256" : 100 runs - average throughput: 66.50 MB/s
            Case "4096" : 100 runs - average throughput: 865.88 MB/s
            Case "65536" : 100 runs - average throughput: 2996.19 MB/s  <-- WTH?
            Case "1048576" : 100 runs - average throughput: 1086.55 MB/s
         Context "Base64Encoder"
            Case "256" : 100 runs - average throughput: 60.01 MB/s
            Case "4096" : 100 runs - average throughput: 521.15 MB/s
            Case "65536" : 100 runs - average throughput: 1415.68 MB/s
            Case "1048576" : 100 runs - average throughput: 1564.53 MB/s
```

There is an issue with the wasm impls in node 25, node 20 is much faster:

```text
   Context "lib/base64/base64.benchmark.js"
      Context "Base64 - decode"
         Context "Node - Buffer"
            Case "256" : 100 runs - average throughput: 39.19 MB/s
            Case "4096" : 100 runs - average throughput: 313.93 MB/s
            Case "65536" : 100 runs - average throughput: 504.64 MB/s
            Case "1048576" : 100 runs - average throughput: 570.46 MB/s
         Context "Base64Decoder"
            Case "256" : 100 runs - average throughput: 46.53 MB/s
            Case "4096" : 100 runs - average throughput: 654.45 MB/s
            Case "65536" : 100 runs - average throughput: 2100.95 MB/s
            Case "1048576" : 100 runs - average throughput: 2213.01 MB/s
      Context "Base64 - encode"
         Context "Node - Buffer"
            Case "256" : 100 runs - average throughput: 24.87 MB/s
            Case "4096" : 100 runs - average throughput: 331.30 MB/s
            Case "65536" : 100 runs - average throughput: 1211.42 MB/s
            Case "1048576" : 100 runs - average throughput: 1684.94 MB/s
         Context "Base64Encoder"
            Case "256" : 100 runs - average throughput: 69.87 MB/s
            Case "4096" : 100 runs - average throughput: 587.26 MB/s
            Case "65536" : 100 runs - average throughput: 2078.57 MB/s
            Case "1048576" : 100 runs - average throughput: 2259.20 MB/s
```
---

Edit: There is indeed a serious performance issue in node 24. All wasm execution suffers from this, but with SIMD it gets really bad:
```text
jerch@linpad:~/xterm-wasm-parts$ nvm use 24
Now using node v24.13.0 (npm v11.6.2)
jerch@linpad:~/xterm-wasm-parts$ node lib/base64/Base64Decoder.wasm.js
duration: 508 ms, rate: 2064 MB/s
duration: 493 ms, rate: 2126 MB/s
duration: 522 ms, rate: 2008 MB/s
jerch@linpad:~/xterm-wasm-parts$ node lib/base64/Base64Decoder.wasm.js
duration: 517 ms, rate: 2027 MB/s
duration: 482 ms, rate: 2177 MB/s
duration: 665 ms, rate: 1577 MB/s
jerch@linpad:~/xterm-wasm-parts$ nvm use 22
Now using node v22.22.0 (npm v10.9.4)
jerch@linpad:~/xterm-wasm-parts$ node lib/base64/Base64Decoder.wasm.js
duration: 254 ms, rate: 4124 MB/s
duration: 213 ms, rate: 4928 MB/s
duration: 223 ms, rate: 4700 MB/s
jerch@linpad:~/xterm-wasm-parts$ node lib/base64/Base64Decoder.wasm.js
duration: 220 ms, rate: 4767 MB/s
duration: 199 ms, rate: 5265 MB/s
duration: 234 ms, rate: 4472 MB/s
```

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
anthonykim1: We now have kittyHandler! chunk where we have similar start() put() and end() as iip and sixel.
Working directly with uint32array instead of string conversion..
We also stream in 1mb shards :) Feels lot faster after this. [e4710d3](https://github.com/xtermjs/xterm.js/pull/5619/commits/e4710d3286cf2e80e6c78a92a27571cd41a6820f)

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
anthonykim1: Thank you! Got lots of inspiration from this for [e4710d3](https://github.com/xtermjs/xterm.js/pull/5619/commits/e4710d3286cf2e80e6c78a92a27571cd41a6820f)  and [a7c4a16](https://github.com/xtermjs/xterm.js/pull/5619/commits/a7c4a160a859136944f4e17ebf545bcf6b23a761) 

Im kind of hesitant to get rid of the full shared -> decode now, since tracking the bytes separately seem intimidating tho 🥲

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -0,0 +1,731 @@
...
+const DECODER_KEEP_DATA = 4194304;
+
+// Base64 shard size (~1MB, must be divisible by 4)
+const BASE64_SHARD_SIZE = 1048576;
+
+// Decoded shard size: 1MB base64 → 768KB decoded
+const DECODED_SHARD_SIZE = 786432;
+
+// Maximum control data size
+const MAX_CONTROL_DATA_SIZE = 4096;
```
jerch: If the control data only consists of smallish key=value combinations, then this number might be too big. Does the kitty protocol state what to do on invalid keys or values? Will they just be ignored or cancel a sequence?

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -0,0 +1,731 @@
...
+      const copyLength = controlEnd - start;
+      if (this._controlLength + copyLength > MAX_CONTROL_DATA_SIZE) {
+        this._aborted = true;
+        return;
+      }
+      this._controlData.set(data.subarray(start, controlEnd), this._controlLength);
+      this._controlLength += copyLength;
+
+      if (!this._inControlData) {
+        // Found semicolon - stream remaining as payload
```
jerch: I suggest to parse the control data here already, this way you can cancel invalid sequences early, e.g. a sequence that does not expect payload can be finished/aborted here already (depends on what the spec says about malformed sequences).

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -0,0 +1,731 @@
...
+   */
+  private _streamPayload(data: Uint32Array, start: number, end: number): void {
+    if (this._aborted) return;
+
+    // Check size limit
+    const estimatedTotal = this._totalDecodedSize + Math.ceil(this._shardBufferPos * 3 / 4) + Math.ceil((end - start) * 3 / 4);
+    if (estimatedTotal > this._opts.kittySizeLimit) {
+      this._aborted = true;
+      return;
+    }
```
jerch: This does the maths for every single chunk. If you turn it around (calc it once on `this._opts.kittySizeLimit`) then you can simply compare sums.

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -0,0 +1,731 @@
...
+
+    const inputLength = end - start;
+    const freeSpace = BASE64_SHARD_SIZE - this._shardBufferPos;
+
+    if (inputLength < freeSpace) {
+      // Fits in current shard
+      this._shardBuffer.set(data.subarray(start, end), this._shardBufferPos);
```
jerch: `this._shardBuffer` introduces another copy: data from parser --> shard buffer --> base64 decoder. The decoder can handle that directly instead by feeding data up to BASE64_SHARD_SIZE to it.

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -0,0 +1,731 @@
...
+    const remaining = start + freeSpace;
+    if (remaining < end) {
+      this._streamPayload(data, remaining, end);
+    }
+  }
+
+  /**
+   * Decode a full 1MB shard (768KB output).
+   */
+  private _decodeFullShard(): boolean {
```
jerch: Decoding the full 1 MB at once is not a good idea, as it introduces blocking milliseconds. Instead feed the data chunk to the decoder with `put` up to BASE64_SHARD_SIZE, on BASE64_SHARD_SIZE call `end`. After than you can grab the data with the `data8` property.

The base64 decoder is optimized for xterm.js' parser's standard chunk size (2^17 codepoints) and would do the base64 decoding in tandem to the parser's work to not aggravate too many blocking milliseconds.

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -0,0 +1,731 @@
...
+    }
+
+    if (this._decoder.end()) {
+      this._decoder.release();
+      this._decodeError = true;
+      return false;
+    }
+
+    const chunk = new Uint8Array(this._decoder.data8);
+    this._decodedChunks.push(chunk);
```
jerch: Hmm, maybe use the blob interface here? That would save precious memory in the JS context. But has the downside of being async on retrieval (thus works only, when the final payload is only used in `end` with a promise as return value).

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -0,0 +1,731 @@
...
+      }
+      pending.leftoverLength = 0;
+    }
+
+    // Decode with 4-byte alignment, preserving leftover if m=1
+    if (this._shardBufferPos > 0) {
+      if (isMoreComing) {
+        this._decodePartialShardWithLeftover(pendingKey, cmd);
+      } else {
+        this._decodePartialShard();
```
jerch: I don't quite understand this leftover & pending handling - is this some specialty of the kitty protocol? Why prepending?

`_decodePartialShardWithLeftover` and `_decodePartialShard` look overly complicated, like you are doing partial base64 decoding yourself. Can you explain what the methods are meant to do? Is that to call `_decoder.init` with the correct size?

addons/addon-kitty-graphics/src/KittyGraphicsAddon.ts
jerch: Sorry overlooked this comment yesterday - well the full shard decoding introduces another copy, which will introduce lag spikes, when the decoder gets the shard for decoding.

Would it help to introduce a property on the decoder, that gives you the amount of loaded base64 bytes? Something like this:
```typescript
interface Decoder {
  ...
  // max allowed base64 bytes
  maxBytes: number;
  // loaded base64 bytes
  loadedBytes: number;
  // free base64 bytes (maxBytes - loadedBytes)
  freeBytes: number;
}
```

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
anthonykim1: You're right, with around 25 keys, I think we should be good with 512.

For invalid key, I think kitty only explicitly state for `Specifying both i and I keys in any command is an error. The terminal must reply with an EINVAL error message, unless silenced.` Otherwise we just ignore them

Will handle this as well!
Edit: [5f724c5](https://github.com/xtermjs/xterm.js/pull/5619/commits/5f724c59247b723c9880ea4a6025ca3c6a517f32)

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
anthonykim1: Good point! Added d44900f spec only seem to call out i and I explicitly so currently all other invalid value would be ignored, or fallback to base case atm.

Maybe we could be more opinionated + explicit and handle them explicitly like i and I?

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
anthonykim1: Tried the pre calculating and also found a bug where we were resetting per chunk, right here: [1029e19](https://github.com/xtermjs/xterm.js/pull/5619/commits/1029e19b57c9cb9bc0daab4f1f703c808a476e8c)

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
anthonykim1: Not kitty‑specific, the leftover/pending was just to preserve base64 4‑char alignment across m=1.

> Is that to call _decoder.init with the correct size?

Yeah.. I originally used 1MB shards because `decoder.init()` needs a decoded size and kitty doesn’t provide size info. But I think I can just can compute decoded size per parser chunk after alignment + padding, so we can probably drop the 1MB shard buffer and decode directly per parser chunk (keeping only 0–3 leftover chars) ???

I should try to tackle decodePartialShard into a simple “leftover merge + per‑chunk decode” path.

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
jerch: Hmm I hate sloppy specs :imp:

I basically had the same issue with IIP, in an older version the byteSize was mandatory and excess data would invalidate the sequence (this was a good and concise spec), but got watered to optional in a newer version of iTerm2. Currently IIP is still on that older stricter spec, but we might need to change that too, as people use the newer sequences more often now.

If you ask me - I am more on the strict side of things regarding this. I would abort on any malformed sequence (e.g. invalid keys, payload in a sequence that doesn't expect payload) as it greatly lowers the chance, that malicious data enters the terminal. Ofc additional limitations to the base spec need to be clearly stated in the docs.

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
jerch: As I noted in https://github.com/xtermjs/xterm.js/pull/5619#discussion_r2753964238, gonna introduce properties on the decoder for the byte loading, then you don't have to deal with this yourself. This is a base64 issue, and the decoder should deal transparently with it and not leak base64 details to the caller side. I will also remove the final size check from the decoder, so if you have no expected size given, it will just decode everything it got by `put` when calling `end`.

With these changes the last shard is easier to handle:
```typescript
// in end()
...
// error now denotes only base64 decoding errors,
// not size mismatches anymore
const error = decoder.end();
if (error) {
  // cleanup & abort
}

// grab the decoded data of the last shard
const data = decoder.data8;

// if you have a size expectation
// e.g. full shard must be at 768kB, IIP still has optional
if (data.length !== expectedSize) {
  // cleanup & abort
}
...
```

anthonykim1: Thank you! https://github.com/xtermjs/xterm.js/pull/5619#discussion_r2753964238 sounds awesome :)) I think this will simplify things a ton for me 🤩

jerch: PR for it https://github.com/jerch/xterm-wasm-parts/pull/6

Please check if the changes will work for you.

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
anthonykim1: [e95ec8d](https://github.com/xtermjs/xterm.js/pull/5619/commits/e95ec8d4bc0375882d21d29ad9cac13a67682ee7) but very likely will revisit and simplify more once I get upgraded decoder

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
anthonykim1: Got rid of _decodeFullShard and now decode per parser chunk here: [e95ec8d](https://github.com/xtermjs/xterm.js/pull/5619/commits/e95ec8d4bc0375882d21d29ad9cac13a67682ee7) but very likely will revisit and simplify more once I get upgraded decoder

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
anthonykim1: Thank you sm , will install locally and give this a try.

RUnning:
``` 
cd /tmp && git clone --branch next_version https://github.com/jerch/xterm-wasm-parts.git
cd xterm-wasm-parts && npm install && npm run build
```
then 
```
cp -r /tmp/xterm-wasm-parts/lib /Users/anthonykim/Desktop/xterm.js/node_modules/xterm-wasm-parts/
cp -r /tmp/xterm-wasm-parts/lib /Users/anthonykim/Desktop/xterm.js/addons/addon-image/node_modules/xterm-wasm-parts/
```

anthonykim1: Played around after checking it out locally, and got the image to render with the new `decoder.put(data.subarray(start, end)) ` and been trying to use the return code as defined const. like `const DECODER_SUCCESS=0`!
 I think we'll be able to trim down significally once the size check from init() goes away 🥹

Screenshot 2026-02-03 at 8 24 15 PM

jerch: > Played around after checking it out locally, and got the image to render with the new `decoder.put(data.subarray(start, end)) ` and been trying to use the return code as defined const. like `const DECODER_SUCCESS=0`! I think we'll be able to trim down significally once the size check from init() goes away 🥹

Yes, the `decoder.put` changed interface in an earlier update to the simpler `put(data: Uint8Array | Uint16Array | Uint32Array): number` style.

What do you mean by removing the size check from `init()`? You can simply always feed it `init(786432)`, schematically:

```typescript
// return means to break & abort handling

const decodedShards = [];
decoder.init(786432);

// Handler.put (for loop simulating the invocations)
for (const chunk of payloadChunks) {
  const free = decoder.freeBytes;
  if (free >= chunk.length) {
    if (decoder.put(chunk)) return;
  } else {
    if (decoder.put(chunk.subarray(0, free))) return;
    if (decoder.end()) return;
    decodedShards.push(new Uint8Array(decoder.data8));
    decoder.init(786432);
    if (decoder.put(chunk.subarray(free))) return;
  }
}

// Handler.end
if (decoder.end()) return;
decodedShards.push(decoder.data8);  // w'o Uint8Array ctor here
// now decodedShards contains all the decoded byte data
// as 786432ish shards + the last smaller one
```

I need the size in `init(size)` to prealloc memory, otherwise we'd have to make it free growing, but then it cannot be reused anymore (likely to grow big) and we enter toxic realloc & wasm respawning cycles.

And doing the last shard borrowed has the advantage, that for small images fitting within nothing gets copied (in RGBA uncompressed it would be up to 512x512 pixels, of course we could increase that limit).

jerch: @anthonykim1 Hmm I just thought about this - prolly making it free growing is indeed the better solution. The sharding basically doubles the copy needs and we just gain the guarantee that we dont have to throw away the wasm instance.

Proposal:
- free growing up to 24MB (decoded) or 32 MB (encoded)
- start with a rather big initial memory of 4MB, then only  3 reallocs might occur by doubling (-->8, -->16, -->32)
- keepSize to 4 MB

jerch: @anthonykim1 Realloc is now implemented in the PR. Usage:

```typescript

// instance with keepSize: 4 MB, maxBytes: 32MB, initialBytes: 1MB
// important: values now denote encoded bytes!
const dec = new Base64Decoder(4194304, 33554432, 1048576);
dec.init(); // allows to overwrite maxBytes & initialBytes

// now simply feed until done or abort on error
if (dec.put(chunk1)) abort();
if (dec.put(chunk2)) abort();
...
// when done with payload
if (dec.end()) abort();
const fullData = dec.data8;

// when done with data processing
dec.release(); // only release on >4MB
```

jerch: @anthonykim1 Released v0.3.0 with the new decoder. Makes it easier for you to use from the ready-to-go package. (Note: If you upgrade the package in the xterm.js project, then the IIP handler also needs changes.)

anthonykim1: YAYYY!! FREE GROWING is awesome!!! thank you 🤩 🚀 @jerch 
Will try this and start trimming a bunch :)) 
Edit --> 🏃[08f85cf](https://github.com/xtermjs/xterm.js/pull/5619/commits/08f85cf826778f9ff13f1e07cfa2776b00a6782f) 

Would it be a good idea for me to update the package first + IIP handler in different PR before this PR or just include them in this PR as well

jerch: Nice, you already patched the IIP handler. LGTM. :+1:

To keep memory profile low, it might be a good idea to share later on the decoder instance across the protocol handlers, as it is technically impossible to have intersecting sequences (here guaranteed by the escape sequence parser). This would avoid holding more and more RAM from more handlers (there is yet another one coming for image serializing also using base64 encoding). This is also the reason, why `init()` can overwrite `maxBytes` while `keepSize` is a ctor-only argument.

About `DECODER_SUCCESS` - the package also exports a `DecodeStatus` enum, might be better to use this than hardcoding the error values. You can import it like this:
```typescript
import Base64Decoder, { DecodeStatus } from 'xterm-wasm-parts/lib/base64/Base64Decoder.wasm';
```

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -0,0 +1,590 @@
...
+  private _handleCommandWithBytesAndCmd(cmd: IKittyCommand, bytes: Uint8Array, decodeError: boolean): boolean | Promise<boolean> {
+    const action = cmd.action ?? 't';
+
+    switch (action) {
+      case KittyAction.TRANSMIT:
+        return this._handleTransmit(cmd, bytes, decodeError);
+      case KittyAction.TRANSMIT_DISPLAY:
+        return this._handleTransmitDisplay(cmd, bytes, decodeError);
+      case KittyAction.QUERY:
+        return this._handleQuery(cmd, bytes, decodeError);
```
jerch: If any of these follow-up actions store the bytes for later usage, then you must make a copy of them. Currently the image data gets passed on borrowed from  the base64 decoder in `end`, which is optimal for actions that finalize the image processing before `end` returns (things like `new ImageData` or `createImageBitmap` or `new Blob` already copy the data itself).

Specifically `_handleTransmit` looks fishy to me, as it stores the data in your `this._images` map. Easy fix - just replace it with:
```typescript
...
return this._handleTransmit(cmd, new Uint8Array(bytes), decodeError);
...
```

Haven't checked the other 2 code paths `_handleTransmitDisplay` and `_handleQuery`.

jerch:
@anthonykim1 Question about your image map: https://github.com/xtermjs/xterm.js/blob/08f85cf826778f9ff13f1e07cfa2776b00a6782f/addons/addon-image/src/kitty/KittyGraphicsHandler.ts#L65

This seems to store kitty image data outside of the `ImageStorage`. We should avoid double storing data. Furthermore we have to abide to a very strict memory handling here, which `ImageStorage` tries to do. (half of the code is about proper image eviction...)

Suggestion - check if we can reuse & extend `ImageStorage` for what is missing for the kitty protocol. If that is futile and you truly need a separate data storage, then we need to offload the data somehow (e.g. by using blobs). Note, that we have only 2GB in Javascript, that will be exhausted in a few seconds, if someone tries to play a movie via the kitty protocol. :cinema:

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
anthonykim1: Good point about the sharing decoder! 

Im just realizing esbuild wont let me inline const enum values from external packages. So at runtime DecodeStatus is undefined, and DecodeStatus.OK throws.. 

I'm thinking about mirroring from type right now, but maybe in the future, we can just have `const enum -> enum` ?  [a59d527](https://github.com/xtermjs/xterm.js/pull/5619/commits/a59d527d1a31117f7e9bd157db6f8cf83b9b5135)

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
anthonykim1: Wow.. Didnt even realize how that underlying memory could be reused later by decoder for different image, thank you this is very helpful!

Done here -->  [596ab84](https://github.com/xtermjs/xterm.js/pull/5619/commits/596ab845d0c741bea4294ae792e2578f118b708b)

anthonykim1:
> @anthonykim1 Question about your image map:
> ... 
> This seems to store kitty image data outside of the ImageStorage. We should avoid double storing data. 
> Furthermore we have to abide to a very strict memory handling here, which ImageStorage tries to do. (half of the  
> code is about proper image eviction...)

> Suggestion - check if we can reuse & extend ImageStorage for what is missing for the kitty protocol. If that is futile 
> and you truly need a separate data storage, then we need to offload the data somehow (e.g. by using blobs). Note, > that we have only 2GB in Javascript, that will be exhausted in a few seconds, if someone tries to play a movie via 
> the kitty protocol. 🎦

@jerch Yeah, this is a challenging one. I don't think we can discard raw data all the time since we won't be able to re-display at different size, a=p (display previously transmitted), etc.

I'll play around with the blob offloading, 

Flow would be like:
storing raw image data as blob and then on display,  decode from blob -> imageBitmap -> pass to Imagestorage?
This way ImageStorage would handle eviction for rendered bitmaps, while blob would persist for re-rendering at different size, etc.

And then the delete command `a=d` would clean up blob. I should also clear on clear screen, transmitting with same id, etc

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
jerch: Eww ok, yeah I remember we had issues with const enums before even with the TS compiler itself, but that got fixed on typescript side way back. Seems esbuild still misses pulling values from d.ts files (const enums are only defined there).

Enums would do as a workaround, but they expose a far worse runtime, as they need "object creation" --> "object name lookup" --> "object property lookup" to retrieve the value on runtime, while a const enum is just initial static value parsing.

So perfwise your local define is the better idea, but then it would be better to copy the const enum over to again benefit from the value being inlined.

I tested the const enum pattern from foreign modules with the typescript compiler and it works there just fine. Imho this should be fixed in esbuild, not by resorting to a weaker pattern.

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
jerch: Yeah that is the idea of `keepSize` - to reuse the same memory later on to lower GC pressure. But with this borrow pattern things need to be tracked carefully. It works well, if the final action copies or transforms the data anyway in sync (e.g. `new ImageData(bytes)`).

I see that you also did it for `_handleTransmitDisplay` - is that really needed? Does the kitty protocol demand you to hold the original bytes?

jerch:
> @jerch Yeah, this is a challenging one. I don't think we can discard raw data all the time since we won't be able to re-display at different size, a=p (display previously transmitted), etc.

This is already done by the `ImageStorage` - it stores the original asset along with a maybe resized one: https://github.com/xtermjs/xterm.js/blob/b287397c4f05fcf2109e3e9446d713b4ea6d6c86/addons/addon-image/src/Types.ts#L100-L108

The resized one (`actual...` entries) kicks in, when you do a font resize on the terminal, e.g. the cell size changed. Resizing always happens from the orig one to not add up artifacts from previous resizes.

> Flow would be like:
storing raw image data as blob and then on display,  decode from blob -> imageBitmap -> pass to Imagestorage?
This way ImageStorage would handle eviction for rendered bitmaps, while blob would persist for re-rendering at different size, etc.
>
> And then the delete command `a=d` would clean up blob. I should also clear on clear screen, transmitting with same id, etc

You are reinventing what is already there. `ImageStorage` also allows to delete an image by an id. If the protocol insists of providing ids from outside, then all you need is an outsideId --> storageId map and delegate the delete to `ImageStorage`. Explicit deletion is currently not exposed by `ImageStorage` as the other protocols have no means to to do that explicitly.

Explicit later resizing by the protocol introduces several issues, esp. around cell alignment and how to evict the previous tiles rendered. This prolly needs a clever separate method on `ImageStorage` similar to `addImage`.

Edit:
I just looked up the kitty protocol - hmm, those placements from a single source need some clarification. They introduce an indirection between loading image data and actual displaying. The image storage cannot do that at the moment. Also its eviction rules are solely based on the fact, whether image tiles are actually visible (invisible deleted first, then in ascending id progression deleting older ones next). So we'd have to change that, hmm.

> Finally, you can specify the image z-index, i.e. the vertical stacking order. Images placed in the same location with different z-index values will be blended if they are semi-transparent. You can specify z-index values using the z key. Negative z-index values mean that the images will be drawn under the text.

Eww, that is not possible at all currently, as the `ImageRenderer` renders always on top. The negative z-index would prolly be possible with the webgl renderer, not sure if we ever get this working with the DOM renderer. The additional layering on top needs render composition, which is a TODO for the `ImageRenderer`.

So yeah after you sorted out the storage needs, the render side of things will get even more funny due to technical limitations.

jerch:
@anthonykim1 I think we should create a discussion about the kitty protocol listing all its features and discuss first, what can be implemented and how, and what not. It is so loaded with features, that working into a certain direction might negate other aspects, if not considered from the beginning.

anthonykim1:
@jerch Good idea! --> https://github.com/xtermjs/xterm.js/discussions/5683 

I'll try to list out the components as sub-items/lists to the discussion, their status, etc tomorrow(edit: added a bunch of things in discussion chart) . I also think its easier for us to talk about if I can number the chart/list and reference them individually (This PR is getting big and there's alot to KGP! )

jerch:
@anthonykim1 Wow this list is a lot to digest. :sweat_smile: Is the list feature complete for what the kitty protocol defines?

anthonykim1:
@jerch Yeap, it's pretty big 🥲 I've been trying to make have dedicated space at the end to track feedbacks from comments, more urgent ones - these would be located under the MVP/Feb section. 

I might still be missing some entries like in the `Control Data keys`.. there are just so many keys 😵‍💫

bin/test_integration.js
```diff
@@ -25,6 +25,7 @@ const addons = [
   'clipboard',
   'fit',
   'image',
+  'kitty-graphics',
```
Tyriar: remove

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -504,12 +504,18 @@ export class KittyGraphicsHandler implements IApcHandler, IResetHandler {
     const savedY = buffer.y;
     const savedYbase = buffer.ybase;
 
+    // Determine layer based on z-index: negative = behind text, 0+ = on top.
+    // Bottom layer only works when allowTransparency is enabled (otherwise text
+    // canvas background is opaque and hides the bottom canvas). Fall back to top.
+    const wantsBottom = cmd.zIndex !== undefined && cmd.zIndex < 0;
+    const layer: ImageLayer = (wantsBottom && this._coreTerminal.options.allowTransparency) ? 'bottom' : 'top';
```
Tyriar: You'll want to store the actual number as well, then draw the images in that order in order to properly respect `z`.

addons/addon-image/src/ImageRenderer.ts
```diff
@@ -305,34 +324,60 @@ export class ImageRenderer extends Disposable implements IDisposable {
...
+        this._canvasTop.classList.add('xterm-image-layer-top');
+        this._terminal._core.screenElement.appendChild(this._canvasTop);
+        this._ctxTop = this._canvasTop.getContext('2d', { alpha: true, desynchronized: true });
+        this.clearAll('top');
+      }
+      if (layer === 'bottom' && !this._canvasBottom) {
+        this._canvasBottom = ImageRenderer.createCanvas(
+          this.document, this.dimensions?.css.canvas.width || 0,
+          this.dimensions?.css.canvas.height || 0
+        );
```
Tyriar: Try reduce duplication from this commit

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
anthonykim1: Can I sort them and keep them like this: [677e527](https://github.com/xtermjs/xterm.js/pull/5619/commits/677e5275f9fcecc39df90f0e19cbe0836efefb76)

addons/addon-image/src/ImageRenderer.ts
anthonykim1: Trimmed some if statement, added a map that contains two entry (one for top, bottom layer) here to make things cleaner --> [4de03aa](https://github.com/xtermjs/xterm.js/pull/5619/commits/4de03aaa383bc7e5cb4b52dc937c5e7aa3ab52c3)

anthonykim1:
In the efforts to not falsely inform/respond back to clients:

<details>
<summary><strong>1) Query action (<code>a=q</code>)</strong></summary>

| Scenario | Example control data | Current behavior | Spec quote |
|---|---|---|---|
| Capability probe (no payload) | `a=q,i=1,t=d` | `OK` | "To check if a terminal emulator supports the graphics protocol the best way is to send the above query action …" |
| PNG query, valid base64 payload | `a=q,i=1,f=100,t=d;<base64>` | `OK` | "The terminal emulator must understand pixel data in three formats, 24-bit RGB, 32-bit RGBA and PNG." |
| PNG query, invalid base64 | `a=q,i=1,f=100,t=d;!!!` | `EINVAL:invalid base64 data` | "The payload is arbitrary binary data, base64 encoded …" |
| PNG query, valid base64 but malformed PNG bytes | `a=q,i=1,f=100,t=d;<malformed png bytes>` | `OK` at query-time; malformed data is rejected during render (`a=T`) | "set `a=q`. Then the terminal emulator will try to load the image and respond with either `OK` or an error …" + PNG support requirement above |
| Raw RGB/RGBA with valid dimensions and enough bytes | `a=q,i=1,f=24,s=1,v=1,t=d;AAAA` | `OK` | "the width and height are specified using the `s` and `v` keys respectively … pixel data must be … bytes." |
| Raw RGB/RGBA missing `s`/`v` | `a=q,i=1,f=24,t=d;AAAA` | `EINVAL:width and height required for raw pixel data` | Same `s`/`v` requirement quote above |
| Raw RGB/RGBA insufficient bytes | `a=q,i=1,f=24,s=10,v=10,t=d;AAAA` | `EINVAL:insufficient pixel data` | Same byte-count requirement quote above |
| Unsupported medium `t=f` | `a=q,i=1,t=f;<base64 path>` | `EINVAL:unsupported transmission medium` | "The `t` key defaults to `d` and can take the values: `d`, `f`, `t`, `s`." |
| Unsupported medium `t=t` | `a=q,i=1,t=t;<base64 path>` | `EINVAL:unsupported transmission medium` | Same quote above |
| Unsupported medium `t=s` | `a=q,i=1,t=s;<base64 shm>` | `EINVAL:unsupported transmission medium` | Same quote above |

</details>

<details>
<summary><strong>2) Non-query actions (with <code>i</code> present)</strong></summary>

| Action/Scenario | Example | Current behavior | Spec quote |
|---|---|---|---|
| Transmit (`a=t`) success | `a=t,i=1,f=100;<png base64>` | `OK` | "terminal emulator will reply after trying to load the image … The string will be `OK` if reading the pixel data succeeded or an error message." |
| Transmit (`a=t`) decode error | `a=t,i=1;!!!` | `EINVAL:invalid base64 data` | Same response quote above |
| Transmit (`a=t`) unsupported medium with `i` | `a=t,i=1,t=f;<base64 path>` | `EINVAL:unsupported transmission medium` | "The `t` key defaults to `d` and can take the values: `d`, `f`, `t`, `s`." |
| Transmit (`a=t`) unsupported medium without `i` | `a=t,t=f;<base64 path>` | No response (silent rejection) | Response contract is tied to sending `i` |
| Transmit+display (`a=T`) success | `a=T,i=1,f=100;<png base64>` | `OK` | "simultaneously transmit and display an image using the action `a=T`" |
| Transmit+display (`a=T`) decode error | `a=T,i=1;!!!` | `EINVAL:invalid base64 data` | Error response on failed load/processing |
| Transmit+display (`a=T`) render failure | `a=T,i=1,f=100;<corrupt image>` | `EINVAL:image rendering failed` | Error response on failed load/processing |
| Transmit+display (`a=T`) unsupported medium with `i` | `a=T,i=1,t=f;<base64 path>` | `EINVAL:unsupported transmission medium` | "The `t` key defaults to `d` and can take the values: `d`, `f`, `t`, `s`." |
| Transmit+display (`a=T`) unsupported medium without `i` | `a=T,t=f;<base64 path>` | No response (silent rejection) | Response contract is tied to sending `i` |
| Delete (`a=d`) | `a=d,d=a` | No response | "Images can be deleted by using the delete action `a=d`." |
| Placement (`a=p`) not implemented | `a=p,i=1` | `EINVAL:unsupported action` | Control reference includes `a=p` (put/display previous transmitted image) |
| Frame upload (`a=f`) not implemented | `a=f,i=1` | `EINVAL:unsupported action` | Control reference includes `a=f` (transmit animation frame data) |
| Animation control (`a=a`) not implemented | `a=a,i=1` | `EINVAL:unsupported action` | Control reference includes `a=a` (control animation) |
| Frame compose (`a=c`) not implemented | `a=c,i=1` | `EINVAL:unsupported action` | Control reference includes `a=c` (compose animation frames) |

</details>

<details>
<summary><strong>3) Validation & key rules</strong></summary>

| Rule | Example | Current behavior | Spec quote |
|---|---|---|---|
| Both `i` and `I` specified | `a=t,i=1,I=2` | `EINVAL:cannot specify both i and I keys` | "Specifying both `i` and `I` keys in any command is an error. The terminal must reply with an EINVAL error message, unless silenced." |
| Query with no explicit `t` | `a=q,i=1` | Treated as `t=d` | "The `t` key defaults to `d` …" |
| Query with no explicit `f` | `a=q,i=1,s=1,v=1;...` | Treated as `f=32` (RGBA) | Control data reference: `f` default is `32` |

</details>

<details>
<summary><strong>4) Quiet mode behavior (<code>q</code>)</strong></summary>

| `q` value | OK responses | Error responses | Spec quote |
|---|---|---|---|
| `q=0` (default) | Sent | Sent | Control data reference default behavior |
| `q=1` | Suppressed | Sent | "Set it to `1` to suppress `OK` responses …" |
| `q=2` | Sent | Suppressed | "… and to `2` to suppress failure responses." |

</details>

<details>
<summary><strong>5) Response-gating behavior by id</strong></summary>

| Scenario | Current behavior | Spec quote |
|---|---|---|
| `a=q` with `i` present | Sends `i=<id>;OK` or `i=<id>;EINVAL:...` (subject to `q`) | Query section + response format examples |
| `a=q` with no `i` | Sends with `i=0` | Query behavior is still expected to reply; control data ref gives `i` default `0` |
| `a=t` / `a=T` with `i` present | Sends `OK`/error response (subject to `q`) | "If it does so [uses `i`], the terminal emulator will reply … `OK` or an error message." |
| `a=t` / `a=T` with no `i` | No response on failure/success | Spec response contract is tied to sending `i` |

</details>

---

**Notes:**
- Current implementation rejects `t=f/t/s` across all actions (`a=q`, `a=t`, `a=T`) with `EINVAL:unsupported transmission medium` in browser context, with TODOs for future filesystem/shared-memory support.
- `a=p/a=f/a=a/a=c` are explicitly rejected with `EINVAL:unsupported action` (when `i` is present), so unsupported features are not silently accepted.
- Query for PNG confirms support capability; malformed PNG payload detection is enforced on actual decode/render paths (`a=T`).

addons/addon-image/src/ImageAddon.ts
```diff
@@ -148,6 +152,18 @@ export class ImageAddon implements ITerminalAddon, IImageApi {
...
+
+    // Kitty graphics handler
+    if (this._opts.kittySupport) {
+      const kittyStorage = new KittyImageStorage(this._storage!);
+      const kittyHandler = new KittyGraphicsHandler(this._opts, this._renderer!, kittyStorage, terminal);
+      this._handlers.set('kitty', kittyHandler);
+      this._disposeLater(
+        kittyStorage,
+        kittyHandler,
+        terminal._core._inputHandler._parser.registerApcHandler(0x47, kittyHandler)
```
Tyriar, line 164: This should use the public API, not `_core`

Tyriar: Oh I see the others use `_core` as well, I guess we can keep this as is for now.

Tyriar:
Let's merge and fix any other issues in separate smaller prs

PR #5645, opened 2026-01-31
Optimize Emitter.fire for 0 and 1 listeners

Tyriar:
Before:
```
      Context "Emitter.fire()"
         Context "0 listeners"
            Case "#1" : 1 runs - average throughput: 97.77 MB/s
         Context "1 listener"
            Case "#1" : 1 runs - average throughput: 49.30 MB/s
         Context "2 listeners"
            Case "#1" : 1 runs - average throughput: 35.37 MB/s
         Context "5 listeners"
            Case "#1" : 1 runs - average throughput: 15.32 MB/s
```

After:
```
      Context "Emitter.fire()"
         Context "0 listeners"
            Case "#1" : 1 runs - average throughput: 376.94 MB/s
         Context "1 listener"
            Case "#1" : 1 runs - average throughput: 84.29 MB/s
         Context "2 listeners"
            Case "#1" : 1 runs - average throughput: 32.36 MB/s
         Context "5 listeners"
            Case "#1" : 1 runs - average throughput: 15.93 MB/s
```

I think MB/s ~= 1M ops/s. Most events have 1 listener so this is a big improvement:

Screenshot 2026-01-31 at 2 44 10 pm

### DIFF
--- a/src/common/Event.test.ts
+++ b/src/common/Event.test.ts
@@ -1,0 +1,72 @@
+/**
+ * Copyright (c) 2024-2026 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import { assert } from 'chai';
+import { Emitter } from 'common/Event';
+
+describe('Emitter', () => {
+  it('should fire with 0 listeners without error', () => {
+    const emitter = new Emitter<number>();
+    emitter.fire(42);
+  });
+
+  it('should fire with 1 listener', () => {
+    const emitter = new Emitter<number>();
+    let received: number | undefined;
+    emitter.event(e => { received = e; });
+    emitter.fire(42);
+    assert.strictEqual(received, 42);
+  });
+
+  it('should fire with 1 listener using thisArgs', () => {
+    const emitter = new Emitter<number>();
+    const obj = { value: 0, handler(e: number) { this.value = e; } };
+    emitter.event(obj.handler, obj);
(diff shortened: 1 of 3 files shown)


jerch:
@Tyriar Eww - the difference is massive. I knew a loop setup is quite expensive - but that much? Btw - the same issue applies to parser handlers as well.

Very creative way of using `ThroughputRuntimeCase` :smile: (not sure, if it accounts MB or MiB).

Tyriar:
@jerch haha, I'm poking various models to come up with perf improvements, that's the benchmark it came up with. I think I'm going to be doing this type of thing more so is there a better way to do the benchmarking using our current setup? Basically just need it to run many iterations to show a number getting bigger or smaller to run before and after

jerch:
`xterm-benchmark` currently has 2 default perf case impls:
- `RuntimeCase` with `showAverageRuntime()` as summary (averaged total runtime)
- `ThroughputRuntimeCase` with `showAverageThroughput()` as summary (averaged MiB/s with 1024 * 1024)

For a MOps/s the throughput case is close, if the payload denotes the op interations, but it is in MiOps due to the 1024 factors. (see https://github.com/xtermjs/xterm-benchmark/blob/8ff2fae0b90427bea67d78d448b6bb63048bba38/src/mixins.ts#L75)

Issue #4552, opened 2023-06-09
WebGPU Renderer


dynamite-bud:
Currently, Xterm.js has:
- Canvas renderer
- WebGL renderer

Now, WebGPU is supported in Chrome v113. We can have an implementation of the WebGPU based renderer in Xterm.js

Tyriar:
I have a work in progress over at https://github.com/Tyriar/xterm.js/tree/webgpu

Tyriar:
This might happen eventually, sounds like the kind of thing I might eventually do on weekends for fun. But in the meantime let's close this off as out of scope since its very much a nice to have, we'll will likely be the more reliable one for some time still anyway.

PR #5666, opened 2026-02-02
Webgpu renderer

Tyriar:
This is a work in progress, I've still barely looked at the code but it appears to work and also be faster than webgl. This was only possible because of how good our shared renderer tests are.

Part of #4552


Tyriar:
Even better models now, if we did this it would make sense to redo from scratch

Issue #5629, opened 2026-01-28
Add API to expose screenElement


anthonykim1:
https://github.com/xtermjs/xterm.js/pull/5619#discussion_r2722602918

PR #5674, opened 2026-02-03
Add api to expose screenElement

anthonykim1:
Resolves: https://github.com/xtermjs/xterm.js/issues/5629 

Be able to avoid things like: `(this._terminal as any)._core.screenElement` + vscode things like 
```
// eslint-disable-next-line no-restricted-syntax
			const screenElement = xterm.raw.element!.querySelector('.xterm-screen')!;
```

### DIFF
--- a/typings/xterm.d.ts
+++ b/typings/xterm.d.ts
@@ -912,9 +912,9 @@ declare module '@xterm/xterm' {
     readonly element: HTMLElement | undefined;
 
     /**
-    * The screen element containing the terminal's canvas rendering layers
-    * and decorations, excluding the viewport and the scrollbar.
-    */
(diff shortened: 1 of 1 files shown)


typings/xterm.d.ts
```diff
@@ -911,6 +911,12 @@ declare module '@xterm/xterm' {
      */
     readonly element: HTMLElement | undefined;
 
+    /**
+    * The screen element containing the terminal's canvas rendering layers
+    * and decorations, excluding the viewport and the scrollbar.
+    */
```
```suggestion
    /**
     * The screen element containing the terminal's canvas rendering layers and
     * decorations, excluding the viewport and the scrollbar.
     */
```

Issue #5689, opened 2026-02-10
Migrate to tsgo


Tyriar:
`tsgo` is the new official go-based typescript compiler currently at https://www.npmjs.com/package/@typescript/native-preview. Pretty soon that will be the default, but there are some missing features so there's a bit of migration work. This should speed up builds significantly.

PR #5692, opened 2026-02-10
Update wasm decoder to 0.3.0 changes in IIPHandler.ts

anthonykim1:
Part of breaking https://github.com/xtermjs/xterm.js/pull/5619 into smaller PRs

### DIFF
--- a/addons/addon-image/src/IIPHandler.ts
+++ b/addons/addon-image/src/IIPHandler.ts
@@ -5,16 +5,19 @@
 import { IImageAddonOptions, IOscHandler, IResetHandler, ITerminalExt } from './Types';
 import { ImageRenderer } from './ImageRenderer';
 import { ImageStorage, CELL_SIZE_DEFAULT } from './ImageStorage';
-import Base64Decoder, { type DecodeStatus } from 'xterm-wasm-parts/lib/base64/Base64Decoder.wasm';
+import Base64Decoder from 'xterm-wasm-parts/lib/base64/Base64Decoder.wasm';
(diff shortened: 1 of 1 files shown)


addons/addon-image/src/IIPHandler.ts
```diff
@@ -5,12 +5,16 @@
...
 import { HeaderParser, IHeaderFields, HeaderState } from './IIPHeaderParser';
 import { imageType, UNSUPPORTED_TYPE } from './IIPMetrics';
 
-// limit hold memory in base64 decoder
+// limit hold memory in base64 decoder (encoded bytes)
 const KEEP_DATA = 4194304;
+const INITIAL_DATA = 1048576;
+
+// Local mirror of const enum (esbuild can't inline const enums from external packages)
+const DECODER_OK: DecodeStatus.OK = 0;
```
jerch: I suggest to put those const integer values into a local const enum. Would save a tiny amount of indirection at runtime.

anthonykim1: Thanks! Would it be like: [d9b5d15](https://github.com/xtermjs/xterm.js/pull/5692/commits/d9b5d150aead1fc6dcbcc76e981c5d9be7d2d6b9) ?

jerch: Yupp, looks good. The hardcoded decoder return value is annoying, but maybe it gets better with #5689. Until then - gonna put a note on the wasm decoder, that changing the return value needs adjustments on our caller sides, too.

PR #5695, opened 2026-02-11
Separate image storage per mechanism (sixel, iip)

anthonykim1:
Coming from: https://github.com/microsoft/vscode/issues/293305 
Per-protocol image storage wrapper for existing sixel , iip protocols. Prep for kitty. 

Right now `ImageStorage` seem to handle alot more than storage, including tiling , eviction, rendering, protocol-specific cursor behavior. Sixel scrolling option (DECSET 80) read in addImage() , meaning every protocol is forced through sixel cursor logic. I think this is fine for IIP, but maybe not so much for kitty.

Make room for:
 - Transmit/display separation. Later kitty image storage can have addsource (upload only, no tiling), addPlacement (tile into buffer)
 - Multiple placement per image. Later kitty imageStorage can track same source different placements.
 - explicit deletion. deleteByQuery from kitty
 - Per command cursor control. We now have addImage(img, scrolling), where we take scrolling as parameter, so kitty can pass true/false instead of relying on global option - always true.

### DIFF
--- a/addons/addon-image/src/IIPHandler.ts
+++ b/addons/addon-image/src/IIPHandler.ts
@@ -4,7 +4,8 @@
  */
 import { IImageAddonOptions, IOscHandler, IResetHandler, ITerminalExt } from './Types';
 import { ImageRenderer } from './ImageRenderer';
-import { ImageStorage, CELL_SIZE_DEFAULT } from './ImageStorage';
+import { IIPImageStorage } from './IIPImageStorage';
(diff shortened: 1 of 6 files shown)


jerch:
@anthonykim1 @Tyriar 

Yes looks good. The old storage grew a bit over time and has been loaded with functionality not fitting elsewhere. As long as all refer to the same storage instance underneath for byte quota it is fine this way.

PR #5698, opened 2026-02-12
Fix IME input when composing in the middle of the textarea

### LINKED ISSUE #5456: Japanese IME issues
VS Code report: https://github.com/microsoft/vscode/issues/255725

Quoting:

---

Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.102.0
- OS Version: macOS 15.3.1 

Steps to Reproduce:

1. Open a new terminal
2. Enter Japanse.

**Terminal**

https://github.com/user-attachments/assets/53b7c988-1610-42ae-ac4e-71924c21a435

**Claude Code**

https://github.com/user-attachments/assets/5494cbb7-4e13-4072-91bd-6b9528e21419

It works on iTerm2, so it would be a bug on IDE's terminal. I also paste a move about it.

https://github.com/user-attachments/assets/84855ddc-06f0-449a-afa7-78c5da72b1e6

Thanks for your help 🙏 

## References

- #187228

ntomoya:
Fixes #5456

When `screenReaderMode` is enabled, moving the caret inside existing text and then committing IME input can cause stale trailing text from the textarea's previous state to be re-sent.

This PR fixes composition range tracking in two places:

- At `compositionstart`, it uses `textarea.selectionStart/selectionEnd` (the actual caret/selection) instead of `textarea.value.length`.
- At finalize time, it keeps support for characters typed immediately after `compositionend` while avoiding re-sending trailing text that already existed before composition started.

This keeps composition diffing aligned with the real insertion point and prevents stale suffix insertion.

## Repro

1. Enable `screenReaderMode`.
2. Type `一二` via IME.
3. Move the caret between `一` and `二`.
4. Commit `一` via IME.

- Before: text could become `一二二` (the previous trailing `二`
is re-sent).
- After: text becomes `一一二`.

## Validation

- Reproduced with `screenReaderMode: true` before the fix; no repro after the fix.
- Confirmed no behavior change in normal mode (`screenReaderMode: false`).
- Manually verified caret movement + IME insertion in the VS Code integrated terminal scenario.

---

Related VS Code issues:

- https://github.com/microsoft/vscode/issues/255725
- https://github.com/microsoft/vscode/issues/274992

### DIFF
--- a/src/browser/input/CompositionHelper.ts
+++ b/src/browser/input/CompositionHelper.ts
@@ -72,6 +72,8 @@ export class CompositionHelper {
    */
   public compositionstart(): void {
     this._isComposing = true;
+    // It's important to use the selection here instead of textarea length to avoid conflicts with
+    // screen reader mode
     const start = this._textarea.selectionStart ?? this._textarea.value.length;
     const end = this._textarea.selectionEnd ?? start;
     this._compositionPosition.start = Math.min(start, end);


Tyriar:
Repro before:

Recording 2026-03-11 at 13 05 22

After:

Recording 2026-03-11 at 13 08 16

src/browser/input/CompositionHelper.ts
```diff
@@ -65,7 +72,11 @@ export class CompositionHelper {
    */
   public compositionstart(): void {
     this._isComposing = true;
-    this._compositionPosition.start = this._textarea.value.length;
+    const start = this._textarea.selectionStart ?? this._textarea.value.length;
+    const end = this._textarea.selectionEnd ?? start;
```
Tyriar, line 78:
```suggestion
    // It's important to use the selection here instead of textarea length to avoid conflicts with
    // screen reader mode
    const start = this._textarea.selectionStart ?? this._textarea.value.length;
    const end = this._textarea.selectionEnd ?? start;
```

Tyriar (approved):

PR #5720, opened 2026-02-19
work around Safari sending keycode 13 (enter) when typing ctrl-c

### LINKED ISSUE #5721: Safari on iPad/iPhone/AppleVisionPro with a hardware keyboard: ctrl-c sends keyCode 13 = <Enter>
Using a hardware keyboard with an iPad, iPhone or an Apple Vision Pro, hitting control-c in Safari results in a keypress event with keyCode 13, which is <Enter>.
This is problematic in general with Unix CLI, and with Emacs.

## Details
Safari on iOS, iPadOS, VisionOS. All versions.
- OS version: 
All versions
- xterm.js version: 
All versions up to 6.0.0

### Steps to reproduce

1. Use an iPhone or iPad or AppleVisionPro, with a paired bluetooth keyboard
2. startup the demo app or, for example, [ttyd](https://github.com/tsl0922/ttyd)
3. type a long running command, e.g `sleep 60`
4. try to interrupt it with control-c
5. the cursor goes to the next line, process is not interrupted.

sergezloto:
Safari has this bug on iOS, iPadOS, AppleVision Pro (but not MacOS) when using a hardware keyboard:
It sends event with `keyCode` 13 when control-c is pressed, which is obviously problematic.
However, the `key` property is correct, and can be used to work around this problem.
Fixes #5721

### DIFF
--- a/addons/addon-image/src/kitty/KittyGraphicsHandler.ts
+++ b/addons/addon-image/src/kitty/KittyGraphicsHandler.ts
@@ -323,9 +323,10 @@ export class KittyGraphicsHandler implements IApcHandler, IResetHandler, IDispos
       case KittyAction.QUERY:
         this._sendResponse(cmd.id ?? 0, 'OK', cmd.quiet ?? 0);
         return true;
+      case KittyAction.PLACEMENT:
+        return this._handlePlacement(cmd);
       default:
         // TODO: Implement remaining actions when needed:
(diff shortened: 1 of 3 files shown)


Tyriar:
Thanks, cherry picked and made some changes in https://github.com/xtermjs/xterm.js/pull/5742

PR #5724, opened 2026-02-20
Add fontAliasing option for webkit font smoothing

### LINKED ISSUE #2464: Support configuring anti-aliasing on renderer
VS Code has a mac-specific feature that allows configuring of anti-aliasing on rendered text, this is because macOS by default disabled anti-aliasing so when you plug a macbook into a regular dpi monitor is looks all choppy and awful. Additionally sub-pixel anti-aliasing can look bad on retina displays.

VS Code issue: https://github.com/microsoft/vscode/issues/36249

murataslan1:
Summary:
- add a new fontAliasing terminal option with values: default, none, antialiased, auto
- apply option to terminal root via -webkit-font-smoothing style updates
- validate option values in options service and expose in public typings
- add terminal and options service coverage for the new option

Validation:
- npm run build
- npm run esbuild
- npm run test-unit -- out-esbuild/browser/Terminal.test.js out-esbuild/common/services/OptionsService.test.js
- npm run lint-changes

Closes #2464

### DIFF
--- a/src/browser/CoreBrowserTerminal.ts
+++ b/src/browser/CoreBrowserTerminal.ts
@@ -1199,6 +1199,12 @@ export class CoreBrowserTerminal extends CoreTerminal implements ITerminal {
     return thirdLevelKey && (!ev.keyCode || ev.keyCode > 47);
   }
 
+  /**
+   * Updates the `-webkit-font-smoothing` CSS property on the terminal element.
+   * Note: This only affects the DOM renderer. Canvas and WebGL renderers
+   * handle font rendering through their own pipelines and are not affected
+   * by CSS font smoothing properties.
+   */
   private _updateFontAliasing(fontAliasing: FontAliasing): void {
     if (!this.element) {
       return;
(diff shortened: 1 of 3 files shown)


src/browser/CoreBrowserTerminal.ts
```diff
@@ -1197,6 +1199,14 @@ export class CoreBrowserTerminal extends CoreTerminal implements ITerminal {
     return thirdLevelKey && (!ev.keyCode || ev.keyCode > 47);
   }
 
+  private _updateFontAliasing(fontAliasing: FontAliasing): void {
+    if (!this.element) {
+      return;
+    }
+
+    this.element.style.setProperty('-webkit-font-smoothing', fontAliasing === 'default' ? '' : fontAliasing);
```
krassowski, line 1213: I think this would not have any effect on `canvas`?

murataslan1:
@krassowski Correct — `-webkit-font-smoothing` is a CSS property so it only affects the DOM renderer. For the canvas and WebGL renderers, font rendering is handled by the canvas 2D context / WebGL shaders respectively, and CSS font smoothing properties have no effect there.

I will update the PR description and add a note in the code/docs clarifying that `fontAliasing` only applies to the DOM renderer.

PR #5728, opened 2026-02-28
Fix #5194: Implement alternateScroll option and fix multi-line alt bu…

### LINKED ISSUE #5194: Scroll in "less" only jumps a single line
## Details
- VScode version: 1.93.1
- OS version: RHEL 9

### Steps to reproduce

1. `less file.txt`
2. Scroll mouse one click

### Expected outcome

File scrolls the same 3 lines you get in the scrollback buffer, or as commonly seen in almost all other terminals.

### Actual outcome

File scrolls a single line.

### Additional details

If I do an strace of `less`, I can see the other terminals send three `\33OB` on one mouse scroll click, whilst xterm.js only sends the one. Which fully explains things.

I couldn't find any setting in vscode that affected this. Workarounds are welcome.

agentic-sanyam:
…ffer scroll


agentic-sanyam:
Hi, the workflows are awaiting approval. Could someone please approve and trigger them?

src/browser/CoreBrowserTerminal.ts
```diff
@@ -906,9 +906,15 @@ export class CoreBrowserTerminal extends CoreTerminal implements ITerminal {
           return false;
         }
 
-        // Construct and send sequences
-        const sequence = C0.ESC + (this.coreService.decPrivateModes.applicationCursorKeys ? 'O' : '[') + (ev.deltaY < 0 ? 'A' : 'B');
-        this.coreService.triggerDataEvent(sequence, true);
+        if (this.optionsService.rawOptions.alternateScroll) {
```
Tyriar, line 909: This should actually live primarily in `IDecPrivateModes` options:

https://github.com/SergioChan/xterm.js/blob/5302e62c90ab7f0ae21518bbd5b36e6617390f11/src/common/Types.ts#L267-L280

I guess we should also have one in xterm.js' options to override it. @jerch is this what you were thinking? This is similar to `cursorStyle` where generally the terminal wants to override it.

compressed-build.zip
Tyriar, line 1: 👀 

Please re-create the PR without this zip. Also 2 separate smaller PRs would be easier to handle as mentioned above.

src/browser/CoreBrowserTerminal.ts
```diff
@@ -906,9 +906,15 @@ export class CoreBrowserTerminal extends CoreTerminal implements ITerminal {
...
+          // Construct and send sequences
+          const sequence = C0.ESC + (this.coreService.decPrivateModes.applicationCursorKeys ? 'O' : '[') + (ev.deltaY < 0 ? 'A' : 'B');
+          let data = '';
+          for (let i = 0; i < Math.abs(lines); i++) {
+            data += sequence;
+          }
+          this.coreService.triggerDataEvent(data, true);
```
Tyriar, line 916: This issue actually tracks 2 distinct changes:

- Multiply the sequence by the number of lines
- Implement `alternateScroll`

I suggest splitting them into separate PRs. as the multiply sequence change is an easy merge.

src/browser/CoreBrowserTerminal.ts
jerch: Yes this should go to `IDecPrivateModes` as there is also a corresponding DECSET mode for it - 1007. Thus needs proper impl in DECSET and DECRQM, too.

src/browser/CoreBrowserTerminal.ts
jerch: Agreed, mixing those two issues in one PR makes things hard to comprehend.

PR #5733, opened 2026-03-03
add onImageAdded to api

anthonykim1:
Part of: https://github.com/microsoft/vscode/issues/299058 

Expose an `onImageAdded` event on `ImageAddon` that fires whenever a new image is stored, regardless of protocol (Sixel, iTerm IIP, Kitty).

Wanting to keep track of usage in VS Code.

### DIFF
--- a/addons/addon-image/src/ImageAddon.ts
+++ b/addons/addon-image/src/ImageAddon.ts
@@ -5,6 +5,7 @@
 
 import type { ITerminalAddon, IDisposable } from '@xterm/xterm';
 import type { ImageAddon as IImageApi } from '@xterm/addon-image';
+import { Emitter, type IEvent } from 'common/Event';
 import { IIPHandler } from './IIPHandler';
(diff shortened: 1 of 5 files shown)


jerch:
@anthonykim1 Would it make sense to also pass the ID with it? This way an integrator can grab a newly added image directly for other purposes, e.g. render it outside as it comes in.

anthonykim1:
@jerch Great question. I thought about this... but was little hesitant since the IDs in the image storage felt internal. 

Also we have a map of kitty id -> internal storage id. I think that may make things little instable/confusing to folks? 

I def think it will be useful, but if we do this, I think we should probably also have like an api to also tell if image (given id) has been evicted, or other state changes related to image (identified with the internal ID) 

What do you think?

jerch:
Yes you are right, it introduces more questions at the current state without a clean use case and proper access pattern.

It is planned to make images still residing in scrollbuffer serializable. For that we might need more fine-grained access to storage internals anyway.

Issue #5732, opened 2026-03-03
perf optimization


jerch:
@anthonykim1 

https://github.com/xtermjs/xterm.js/blob/dedf3b15956b5185eae84d49f3fbd7593a2fc0fc/addons/addon-image/src/kitty/KittyGraphicsHandler.ts#L755-L757

In lines above imho the `& 0x00FFFFFF` parts are not needed, as the final `| 0xFF000000` would set all bits on the highest byte anyway.

PR #5739, opened 2026-03-08
Remove redundant bitmask in RGB->RGBA conversion

anthonykim1:
Resolves: https://github.com/xtermjs/xterm.js/issues/5732 
Thanks!


jerch (approved):
👍 LGTM.

Just a minor beautify nag - I would also remove the outer parentheses in the overlapping equations, as bitwise OR is commutative. You could also pull the 0x00FFFFFF in front, maybe makes the 4th byte interleaving more obvious. Both is coding taste, so just leave as it is if that is more readable for you.

anthonykim1:
always down for cleaner code! Thanks :))

Tyriar (approved):

PR #5743, opened 2026-03-09
fix(input): anchor IME composition to textarea cursor

### LINKED ISSUE #5734: IME composition window positioned incorrectly when terminal displays placeholder text
## Description

  When using Chinese IME (Input Method Editor) in xterm.js, the composition candidate window appears
  at an incorrect position when the terminal application displays placeholder/hint text (e.g., Claude
  Code CLI's "Try 'fix lint errors'" prompt). The IME window appears at the end of the placeholder
  text instead of at the actual cursor position (beginning of the line).

Image

  This issue does not occur in native terminal emulators like Windows Terminal with the same shell and
   application, suggesting it's specific to xterm.js's IME handling.

  ## Details
  - Browser and browser version: Electron 33.2.1 (Chromium 130.0.6723.120)
  - OS version: Windows 11 Pro 10.0.22621
  - xterm.js version: 6.0.0

  ## Steps to reproduce

  1. Create an Electron app with xterm.js 6.0.0
  2. Connect xterm.js to a PTY running PowerShell 7
  3. Launch Claude Code CLI (or any CLI app that displays placeholder/hint text using ANSI escape
  sequences)
  4. When the placeholder text appears (e.g., "Try 'fix lint errors'"), activate Chinese IME
  5. Start typing with IME (e.g., type "nihao" to get Chinese characters)

  ## Expected behavior

  The IME composition candidate window should appear at the actual cursor position (beginning of the
  line where input will be inserted), just like it does in Windows Terminal.

  ## Actual behavior

  The IME composition candidate window appears at the end of the placeholder text, not at the actual
  cursor position. After typing a space (which causes the placeholder to disappear), subsequent IME
  input shows the candidate window at the correct position.

  ## Additional context

  - The issue only affects the **first** IME composition when placeholder text is visible
  - After the placeholder disappears (e.g., after typing space), IME positioning works correctly
  - The same setup works correctly in Windows Terminal, indicating this is xterm.js-specific
  - The placeholder text is likely rendered using ANSI escape sequences by the terminal application
  - The `.xterm-helper-textarea` element seems to have incorrect cursor position information when
  placeholder text is present

## Workarounds attempted

  - Setting `selectionStart` and `selectionEnd` to 0 in compositionstart event → breaks input
  functionality
  - Clearing textarea value → causes input loss
  - Sending dummy characters to hide placeholder → causes input duplication

  None of these workarounds resolved the issue without breaking normal input functionality.

  ## Related issues

  This may be related to:
  - #3533 (Sogou Chinese IME support)
  - #3679 (CJK IME input broken on Windows in Electron 13)

  However, those issues describe different symptoms (input not working at all), while this issue is
  specifically about IME candidate window positioning.

SergioChan:
## Summary
- use the textarea caret position (`selectionStart`) to anchor composition start/end instead of always using `value.length`
- preserve compatibility with non-composition keystrokes while stripping pre-existing trailing helper text when forwarding composed input
- add a regression unit test covering ghost/placeholder text with the cursor at column 0

## Testing
- npm run esbuild
- npm run test-unit -- --grep="actual cursor position"

## Related
Fixes #5734

### DIFF
--- a/src/browser/input/CompositionHelper.test.ts
+++ b/src/browser/input/CompositionHelper.test.ts
@@ -31,6 +31,8 @@ describe('CompositionHelper', () => {
     } as any;
     textarea = {
       value: '',
+      selectionStart: 0,
+      selectionEnd: 0,
       style: {
         left: 0,
         top: 0
@@ -229,6 +231,26 @@ describe('CompositionHelper', () => {
         textarea.value = 'ㅇ1';
         setTimeout(() => { // wait for any textarea updates
           assert.equal(handledText, 'ㅇ1');
+          done();
+        }, 0);
+      }, 0);
+    });
+
+    it('Should use textarea selection start to anchor composition at the actual cursor position', (done) => {
+      textarea.value = "Try 'fix lint errors'";
+      textarea.selectionStart = 0;
+      textarea.selectionEnd = 0;
+
+      compositionHelper.compositionstart();
+      compositionHelper.compositionupdate({ data: '你' });
+      textarea.value = "你Try 'fix lint errors'";
+      textarea.selectionStart = 1;
+      textarea.selectionEnd = 1;
+
+      setTimeout(() => {
+        compositionHelper.compositionend();
+        setTimeout(() => {
+          assert.equal(handledText, '你');
           done();
         }, 0);
       }, 0);
--- a/src/browser/input/CompositionHelper.ts
+++ b/src/browser/input/CompositionHelper.ts
@@ -42,6 +42,11 @@ export class CompositionHelper {
   private _dataAlreadySent: string;
 
   /**
+   * Trailing textarea content that existed before the current composition started.
+   */
+  private _initialCompositionSuffix: string;
+
+  /**
    * The pending textarea change timer, if any.
    */
   private _textareaChangeTimer?: number;
@@ -58,6 +63,11 @@ export class CompositionHelper {
     this._isSendingComposition = false;
     this._compositionPosition = { start: 0, end: 0 };
     this._dataAlreadySent = '';
+    this._initialCompositionSuffix = '';
+  }
+
+  private _getTextareaCursorIndex(): number {
+    return this._textarea.selectionStart ?? this._textarea.value.length;
   }
 
   /**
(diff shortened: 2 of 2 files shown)


anthonykim1:
@SergioChan Thanks for the PR. Were you able to repro the issue on your setup? 

@Tyriar This one is tricky. I did notice that there are like two "ghost text"/suggestions sometimes show in the prompt of claude othertimes below.. https://github.com/xtermjs/xterm.js/issues/5734#issuecomment-4032923013 

I thought these ghost text were from cli or tui (ink in claude code).
Maybe Ink is not telling us the most accurate cursor position?

SergioChan:
Thanks for checking this.

I wasn't able to fully reproduce the exact Windows + Claude/Ink environment locally in this round, so this patch is based on the failure mode described in #5734 plus a focused regression test that models the reported condition:

- pre-existing trailing helper/ghost text in the textarea value
- actual cursor at column 0 (`selectionStart = 0`)
- IME composition input arriving while that mismatch exists

The key change is to anchor composition boundaries to the caret position rather than `value.length`, so we don't append composed text after unrelated trailing content.

If it helps, I can add another test variant that more closely mirrors the exact sequence from the issue comments.

Tyriar:
> Thanks for checking this.
> 
> I wasn't able to fully reproduce the exact Windows + Claude/Ink environment locally in this round, so this patch is based on the failure mode described in #5734 plus a focused regression test that models the reported condition:
> 
> * pre-existing trailing helper/ghost text in the textarea value
> * actual cursor at column 0 (`selectionStart = 0`)
> * IME composition input arriving while that mismatch exists
> 
> The key change is to anchor composition boundaries to the caret position rather than `value.length`, so we don't append composed text after unrelated trailing content.
> 
> If it helps, I can add another test variant that more closely mirrors the exact sequence from the issue comments.

I'd rather not do a speculative fix without verifying the actual problem, otherwise these fixes that we can't confirm will stack up and greatly complicate the codebase. @anthonykim1's video in https://github.com/xtermjs/xterm.js/issues/5734#issuecomment-4032923013 seems to indicate it's a more general problem of textarea position rather than textarea.value being the problem.

Let's close this and make sure we can verify before trying to fix. I'm pretty sure some of the other ime issues are tracking bad positioning of the textarea https://github.com/xtermjs/xterm.js/issues?q=state%3Aopen%20label%3A%22area%2Fime%22

PR #5746, opened 2026-03-09
  fix: prevent crash when dimensions getter is called before renderer is initialized

milandebuck:
## Problem
  The `dimensions` getter in `RenderService` crashes when `_renderer.value` is undefined:

  public get dimensions(): IRenderDimensions { return this._renderer.value!.dimensions; }

  ### Impact on VSCode
  This bug breaks the VSCode integrated terminal on macOS 26 (Tahoe). When users run interactive CLI tools that
  use raw mode (Inquirer.js prompts, Claude Code, `npx create-next-app`, etc.), the terminal becomes
  unresponsive:

  - **Enter key stops working**
  - **Backspace key stops working**
  - **Control keys stop working**
  The only workaround is to close and reopen the terminal.

  **VSCode issue:** microsoft/vscode#300192
  ### Root Cause
  A race condition occurs where `dimensions` is accessed before the renderer is initialized or after it's
  disposed. The non-null assertion (`!`) causes an uncaught exception that breaks terminal input handling.

  ## Solution
  Added a static fallback object and replaced the non-null assertion with optional chaining:

  private static readonly _defaultDimensions: IRenderDimensions = {
    css: { canvas: { width: 0, height: 0 }, cell: { width: 0, height: 0 } },
    device: { canvas: { width: 0, height: 0 }, cell: { width: 0, height: 0 }, char: { width: 0, height: 0, top: 0,
   left: 0 } }
  };
  public get dimensions(): IRenderDimensions { return this._renderer.value?.dimensions ??
  RenderService._defaultDimensions; }

  ## Testing
  - [x] Added unit tests for the `dimensions` getter (3 test cases)
  - [x] All existing tests pass (2236 tests)
  - [x] Lint passes
  - [x] Manually verified fix in VSCode (Code-OSS) integrated terminal

### DIFF
--- a/src/browser/services/RenderService.test.ts
+++ b/src/browser/services/RenderService.test.ts
@@ -1,0 +1,143 @@
+/**
+ * Copyright (c) 2019 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import jsdom = require('jsdom');
+import { assert } from 'chai';
+import { RenderService } from 'browser/services/RenderService';
+import { IRenderDimensions, IRenderer, IRequestRedrawEvent } from 'browser/renderer/shared/Types';
+import { MockBufferService, MockOptionsService, MockCoreService, MockLogService, MockDecorationService } from 'common/TestUtils.test';
+import { MockCharSizeService, MockThemeService } from 'browser/TestUtils.test';
+import { ICoreBrowserService } from 'browser/services/Services';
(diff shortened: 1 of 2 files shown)


Tyriar:
This isn't the right fix for that exception, it intentionally throws there as it should not be possible. Commented on https://github.com/microsoft/vscode/issues/300192, we need to root cause it.

Issue #5756, opened 2026-03-11
Clarify mouse services


Tyriar:
- `CoreMouseService` should handle just the mouse protocol state, dealing with events should move out of it as it's never used in headless
- `CoreMouseService` should be named `MouseProtocolService`

Related: https://github.com/xtermjs/xterm.js/issues/5754

PR #5758, opened 2026-03-11
CoreMouseService -> MouseStateService, make mouse services more focused

Tyriar:
- Renames CoreMouseService -> MouseStateService
- Moves custom mouse event handler into MouseStateService, removing the need to keep it around in CoreBrowserService
- Moves mouse event handling to MouseService (browser/) since it's only needed in xterm, not headless 

Fixes https://github.com/xtermjs/xterm.js/issues/5756

### DIFF
--- a/src/browser/services/MouseService.test.ts
+++ b/src/browser/services/MouseService.test.ts
@@ -1,0 +1,213 @@
+/**
+ * Copyright (c) 2019 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+import { assert } from 'chai';
+import { MouseService } from 'browser/services/MouseService';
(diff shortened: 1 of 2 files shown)


jerch:
@Tyriar Is there a specific reason for moving `triggerMouseEvent` out of common? Imho being able to a send mouse reports is a core feature, e.g. if some integrator wants to build a terminal multiplexer with headless, he'd certainly need that functionality. This was the reason to create own core mouse events in the first place.

The `consumeWheelEvent` was indeed misplaced there.

Tyriar:
@jerch just reorganizing since it's not exposed currently on the headless API, if we wanted to expose it there no problem moving it back.

PR #5759, opened 2026-03-11
fix(ime): resync textarea position when composition starts

### LINKED ISSUE #5734: IME composition window positioned incorrectly when terminal displays placeholder text

SergioChan:
## Summary
- sync the helper textarea position right before `compositionstart` is processed
- trigger an immediate composition element update after composition starts so IME anchoring uses the latest cursor position
- this avoids stale IME anchor coordinates when prompt UIs briefly draw helper text and restore the cursor before the first composition render tick

## Testing
- `npm run lint -- src/browser/CoreBrowserTerminal.ts` *(fails in this environment because project dev dependencies are not installed yet, so `eslint` is unavailable)*

## Related
- Fixes #5734

### DIFF
--- a/src/browser/CoreBrowserTerminal.ts
+++ b/src/browser/CoreBrowserTerminal.ts
@@ -415,9 +415,11 @@ export class CoreBrowserTerminal extends CoreTerminal implements ITerminal {
     this._register(addDisposableListener(this.textarea!, 'keydown', (ev: KeyboardEvent) => this._keyDown(ev), true));
     this._register(addDisposableListener(this.textarea!, 'keypress', (ev: KeyboardEvent) => this._keyPress(ev), true));
     this._register(addDisposableListener(this.textarea!, 'compositionstart', () => {
-      // Ensure the textarea is synced to the latest cursor location before composition begins.
-      // Some prompt UIs briefly draw helper text and then restore the cursor without triggering a
-      // new key event, which can leave IME anchoring stale until the first render tick.
+      // Ensure the textarea is synced to the latest cursor location before composition begins. This
+      // is to workaround a problem where highly dynamic TUIs like agentic CLIs reprint agressively
+      // would cause the IME to appear in the wrong position. The theory is that when the IME is
+      // triggered during a partial render the textarea position becomes locked and will not move
+      // until it is hidden and a custom move occurs.
       this._syncTextArea();
       this._compositionHelper!.compositionstart();
       this._compositionHelper!.updateCompositionElements();


src/browser/CoreBrowserTerminal.ts
```diff
@@ -414,7 +414,14 @@ export class CoreBrowserTerminal extends CoreTerminal implements ITerminal {
     this._register(addDisposableListener(this.textarea!, 'keyup', (ev: KeyboardEvent) => this._keyUp(ev), true));
     this._register(addDisposableListener(this.textarea!, 'keydown', (ev: KeyboardEvent) => this._keyDown(ev), true));
     this._register(addDisposableListener(this.textarea!, 'keypress', (ev: KeyboardEvent) => this._keyPress(ev), true));
-    this._register(addDisposableListener(this.textarea!, 'compositionstart', () => this._compositionHelper!.compositionstart()));
+    this._register(addDisposableListener(this.textarea!, 'compositionstart', () => {
+      // Ensure the textarea is synced to the latest cursor location before composition begins.
+      // Some prompt UIs briefly draw helper text and then restore the cursor without triggering a
+      // new key event, which can leave IME anchoring stale until the first render tick.
+      this._syncTextArea();
+      this._compositionHelper!.compositionstart();
```
Tyriar, line 424: Fix seems reasonable. I reproduced a problem similar to the video once but could not seem to do it again. The sync is meant to happen immediately after cursor move:

https://github.com/xtermjs/xterm.js/blob/89980fc565c178d13588e135f8eb975f44a47712/src/browser/CoreBrowserTerminal.ts#L337-L340

This happens after a full parse has occurred (not for every write):

https://github.com/xtermjs/xterm.js/blob/89980fc565c178d13588e135f8eb975f44a47712/src/common/InputHandler.ts#L519-L521

So the only thing I can think of that could cause it is a race condition where the data events are coming in separate chunks such that the cursor position is not in the final position (likely still hidden as the rendering is still occurring), triggering composition during one of these partial re-renders would cause `isComposing` to be true and the position of the element to become locked until composition has ended and the cursor moves again.

src/browser/CoreBrowserTerminal.ts
```diff
@@ -414,7 +414,14 @@ export class CoreBrowserTerminal extends CoreTerminal implements ITerminal {
     this._register(addDisposableListener(this.textarea!, 'keyup', (ev: KeyboardEvent) => this._keyUp(ev), true));
     this._register(addDisposableListener(this.textarea!, 'keydown', (ev: KeyboardEvent) => this._keyDown(ev), true));
     this._register(addDisposableListener(this.textarea!, 'keypress', (ev: KeyboardEvent) => this._keyPress(ev), true));
-    this._register(addDisposableListener(this.textarea!, 'compositionstart', () => this._compositionHelper!.compositionstart()));
+    this._register(addDisposableListener(this.textarea!, 'compositionstart', () => {
+      // Ensure the textarea is synced to the latest cursor location before composition begins.
+      // Some prompt UIs briefly draw helper text and then restore the cursor without triggering a
+      // new key event, which can leave IME anchoring stale until the first render tick.
```
Tyriar:
```suggestion
      // Ensure the textarea is synced to the latest cursor location before composition begins. This
      // is to workaround a problem where highly dynamic TUIs like agentic CLIs reprint agressively
      // would cause the IME to appear in the wrong position. The theory is that when the IME is
      // triggered during a partial render the textarea position becomes locked and will not move
      // until it is hidden and a custom move occurs.
```

src/browser/CoreBrowserTerminal.ts
Tyriar: Note that I was testing on mac, reprinting probably happens in more chunks on Windows so this theory makes sense in that respect.

PR #5762, opened 2026-03-11
Fix RTL breaking IME composition rendering

### LINKED ISSUE #5760: Compositions are backwards due to direction: rtl
Pinyin "nihao.":

Image

Caused by https://github.com/xtermjs/xterm.js/pull/5747

Tyriar:
Fixes #5760

cc @anthonykim1


anthonykim1:
Nice catch, thank you! Weird how the dots go in the left but the characters on the right.. lol
Maybe this is because punctuations are supposed to neutral and takes the direction of container

Tyriar:
`direction: rtl` is for right-to-left languages which can seem very unintuitive to English speakers but that's how they work (shuffling punctuation/words in opposite direction).

PR #5766, opened 2026-03-11
fix(addon-image): disable desynchronized resize canvas

### LINKED ISSUE #5691: Images jitter when resizing the demo window
Image

SergioChan:
## Summary
- Image addon currently creates layer canvases with { desynchronized: true }.
- During resize this can show blank-frame jitter on some platforms because repaint may miss the resize frame.
- Switch layer canvas creation to { desynchronized: false } so resize and repaint stay synchronized.

Fixes #5691.

## Validation
- npm run lint -- addons/addon-image/src/ImageRenderer.ts

### DIFF
--- a/addons/addon-image/src/ImageRenderer.ts
+++ b/addons/addon-image/src/ImageRenderer.ts
@@ -358,8 +358,7 @@ export class ImageRenderer extends Disposable implements IDisposable {
       canvas.style.zIndex = '0';
       screenElement.appendChild(canvas);
     }
-    // Keep rendering synchronized with resize to avoid blank-frame jitter on some platforms.
-    const ctx = canvas.getContext('2d', { alpha: true, desynchronized: false });
+    const ctx = canvas.getContext('2d', { alpha: true });
     if (!ctx) {
       canvas.remove();
       return;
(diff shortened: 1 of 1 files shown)


Tyriar:
Need to test this on Windows before and after to verify it's fixed.

SergioChan:
Thanks — I ran a Windows before/after check as requested.

Environment:
- Windows 10 (22H2)
- Chrome (stable)
- `xterm.js` demo page (`image` addon, `snake` sample)

What I tested:
1. **Before**: local baseline with `desynchronized: true` in `ImageRenderer.ts`
2. **After**: this PR version with `desynchronized: false`
3. Repeated rapid resize loops between narrow/wide viewport sizes + manual drag-resizing.

Result:
- I **could not reproduce visible jitter/blank-frame behavior** in my Windows environment in either baseline or patched variant.
- So from this machine I can’t provide a “before jitters / after fixed” confirmation yet.

If you have a specific repro setup (GPU/browser flags/display scale/steps) where you can reproduce reliably, I can rerun exactly that path and report back with targeted results.

Tyriar:
Let's put this on hold until I can test on the machine I know I saw it on then.

jerch:
@Tyriar Checked this on my old laptop, where it is reproducible under linux. The PR fixes it for me.

PR #5767, opened 2026-03-11
fix(addon-web-links): recover wrapped URL detection when reflow wrap flags are stale

### LINKED ISSUE #5412: xterm WebLinkAddon does not detect links with new line (broken from xterm fit)
I am trying to open `codex` from openai and authenticate through the link, but for some reason the link is not being detected to be opened:

```
https://auth.openai.com/oauth/authorize?response_type=code&client
_id=aljsdhfjkahsdkjfhakjsdjkfa&redirect_uri=http%3A%2F%2Flocalh
ost%3A1455%2Fauth%2Fcallback&scope=openid%20profile%20email%20off
line_access&code_challenge=y9F1nT-S38-R5Xa4yu0oXR3kxIq8yd42-bf8P-
Ten78&code_challenge_method=S256&id_token_add_organizations=trues
codex_cli_simplified_flow=true&state=xfkjsadjkghsjkafhgksjhdfgjk
faskjdfhajshdfkjhaskd&originator=codex_cli_rs
```

Depending on the terminal width it only detects the first line; 

## Details
- Browser and browser version: Chrome latest
- OS version: MacOS Sonoma
- xterm.js version: latest

### Steps to reproduce

1. Open OpenAI -> Codex
2. Login with Account + Click the link

SergioChan:
## Summary
- add a conservative URL-continuation fallback in `LinkComputer._getWindowedLineStrings` so web links can still be reconstructed when reflow/resize leaves stale `isWrapped` markers
- keep the existing `isWrapped` path unchanged and only join adjacent lines when both sides look URL-safe and there is HTTP(S) protocol context
- add a regression test that simulates lost wrap markers and verifies hovering a continuation row still resolves the full URL

## Testing
- `npm run lint -- addons/addon-web-links/src/WebLinkProvider.ts addons/addon-web-links/test/WebLinksAddon.test.ts`

## Related
- Fixes #5412

### DIFF
--- a/addons/addon-web-links/test/WebLinksAddon.test.ts
+++ b/addons/addon-web-links/test/WebLinksAddon.test.ts
@@ -124,9 +124,10 @@ test.describe('WebLinksAddon', () => {
       await ctx.proxy.write(uri);
 
       const wrappedRows: number[] = await ctx.page.evaluate(() => {
+        const term = (window as any).term;
         const rows: number[] = [];
-        for (let y = 0; y < window.term.rows; y++) {
-          const line = window.term.buffer.active.getLine(y);
+        for (let y = 0; y < term.rows; y++) {
+          const line = term.buffer.active.getLine(y);
           if (line?.isWrapped) {
             rows.push(y);
           }
@@ -136,8 +137,9 @@ test.describe('WebLinksAddon', () => {
       strictEqual(wrappedRows.length > 0, true);
 
       await ctx.page.evaluate(rows => {
+        const term = (window as any).term;
         for (const row of rows) {
-          const line = window.term.buffer.active.getLine(row) as any;
+          const line = term.buffer.active.getLine(row) as any;
           if (line) {
             line.isWrapped = false;
           }


SergioChan:
Pushed a follow-up commit to address the CI TypeScript errors in the new regression test.

### What changed
- In the test-only `page.evaluate` callbacks, switched direct `window.term` access to a local `const term = (window as any).term;` reference.
- This keeps runtime behavior identical but avoids TS2339 (`Property 'term' does not exist on type 'Window & typeof globalThis'`) during the build step.

### Validation
- `npm run lint -- addons/addon-web-links/src/WebLinkProvider.ts addons/addon-web-links/test/WebLinksAddon.test.ts`

I was not able to run `npm run tsc` locally in this environment due to missing `@typescript/native-preview` platform package, so I’ll rely on CI for the full matrix.

jerch:
@SergioChan This is the wrong end to fix issues around text reflow and stale `isWrapped` markers. ~Also please open an issue first to discuss things, so we can pin down the issue to the real route cause and fix it there.~ Thanks.

@Tyriar Do we have issues with line reflowing and `isWrapped` state updates from resizing?

Edit: Right, it is already stated in #5412.

PR #5770, opened 2026-03-12
fix: defer viewport DOM sync during synchronized output (DEC 2026)

mgabor3141:
## Summary

Fix viewport scroll flicker during synchronized output mode (DEC 2026).

When synchronized output is active (`CSI ?2026h`..`CSI ?2026l`), `RenderService` defers canvas paints, but `Viewport` was still applying DOM scroll updates (`scrollTop`/`scrollHeight`) immediately. This could cause visible jump/flicker when apps clear and redraw scrollback inside a synchronized output block.

This change defers viewport DOM sync while synchronized output is enabled and flushes it on the next render (after ESU), so canvas and DOM scroll state update atomically.

Also updates the demo client to avoid reading the removed internal `viewport.scrollBarWidth` property.

## Changes

- `src/browser/Viewport.ts`
  - Inject `ICoreService`
  - Skip `_sync()` while `decPrivateModes.synchronizedOutput` is true
  - Track a pending viewport sync and flush it from `renderService.onRender`
- `demo/client/client.ts`
  - Replace `_core.viewport.scrollBarWidth` usage with options-based scrollbar width calculation

## Validation

- `npm run esbuild`
- `npm run test-unit` (2242 passing)
- `npm run lint`

+2 new tests added

## Repro / verification notes

The terminal in these tests is running [pi](https://pi.dev) which notably does not use the alternate screen, but instead uses the synchronized output to delete and re-print changed lines.

In the tests I have some scrollback, and change the prompt line to go from being two lines to being one line. This causes pi to rerender lines, which used to cause flicker and now doesn't.

### xterm.js `5.5.0`

Flickers consistently, occasionally even loses the scrolled to bottom state as seen towards the end of the video.

https://github.com/user-attachments/assets/5eed1f18-198d-44c2-967a-6b122be868b2

### current `master`

Still flickers, but it is less noticeable somehow. You can see the scrollbar that there is still dom updates happening. Can't reproduce the above issue where it loses the scrolled-to-bottom state.

https://github.com/user-attachments/assets/bba43cb0-dd34-4290-965c-13df35b9f1ec

### this patch

Flicker gone completely, scrollbar not moving.

https://github.com/user-attachments/assets/e7f41fac-37e0-49dd-b867-f362e8b97299


mgabor3141:
Rewrote history to be TDD: first commit is just adding the core service to the viewport, second commit adds failing tests, third one fixes them

Tyriar:
@mgabor3141 was there an issue tracking this?

mgabor3141:
@Tyriar No, I wanted to investigate more before filing an issue, but during investigating I found the fix too.

Tyriar (approved):
Thanks, looks good! I deleted the test as I want us to eventually remove jsdom if possible and the test itself was a little too fragile.

PR #5775, opened 2026-03-14
fix(addon-image): render inline images at device resolution on HiDPI

### LINKED ISSUE #5035: Terminal Image display has low resolution
From https://github.com/microsoft/vscode/issues/210672

Does this issue occur when all extensions are disabled?: Yes/No

- VS Code Version: 1.88.1
- OS Version: Mac: Universal Intel silicon 

Steps to Reproduce:

1. imgcat "anyimage.png"

# Original image:

download

# In VScode terminal

CleanShot 2024-04-18 at 12 08 01@2x

# In Iterm2 terminal

CleanShot 2024-04-18 at 12 09 17@2x

mgabor3141:
On HiDPI displays (`devicePixelRatio > 1`) the image layer is sized in CSS pixels, so the browser upscales it — every inline image looks blurry.

Render the image layer at device resolution instead: the backing store is sized in device pixels (CSS size via `style`), tiles are placed in device cells, and the IIP/sixel/kitty decoders rasterize their sources at device resolution. At `dpr === 1` nothing is resampled, so pixel-exact images are untouched.

Adds `ImageRendering.test.ts` — a protocol × sizing × `devicePixelRatio` (1, 1.5, 2) matrix across all browser projects. It asserts the layer backing store is at device resolution and the on-screen footprint is `devicePixelRatio`-invariant (catching "half size on HiDPI"); both fail at dpr > 1 without the fix.

**Before / after at `devicePixelRatio` 2:**

Resolution chart (multi-frequency checkerboard):

proof_chart

Photo (`imgcat photo.jpg`):

proof_photo

Fixes #5035


jerch:
@mgabor3141 I looked at your PR - You sure this works correctly and prints the image at the correct size? You basically just increased the canvas internal size, so with a DPR of 2 wouldn't images get rendered half as big?

Don't we need a transform or scale adjustment on the context like this:
```typescript
ctx.setTransform(dpr, 0, 0, dpr, 0, 0);
// or
ctx.scale(dpr, dpr);
```

Note that we have to stay in the DPR 1 assumption for image pixels, as image sequences currently have no way to tell us that a different DPR was targeted. Thus an image pixel must be treated as one CSS pixel.

Edit: Another way would be to do the CSS --> device pixels resizing on the image prehand, this way we would get less rounding artifacts (given that DPR is usually >1).

mgabor3141:
@jerch that's correct. I've looked into this some more and updated the PR with a better approach. It now renders the image layer at device resolution instead of CSS resolution.

[AGENT] A couple of notes on scope and the discussion in #5861:

**Detail recovery is source-limited.** This only recovers real detail when the app sends more pixels than the on-screen footprint (i.e. a downscale) — that's the case for **IIP**, where the original encoded image is decoded straight to the device-pixel target. For **sixel** the app rasterizes to a fixed pixel grid with no way to request a larger source, so those pixels *are* the image and we can only interpolate them up to device resolution — we avoid the browser's second upscale, but there's no extra detail to draw. **kitty** has the same limitation whenever the app transmits display-sized pixels.

**`px` = device pixels.** Following @gnachman's clarification that iTerm2 treats `Npx` as device pixels, I fixed the `px` handling: `px` params are now passed through unscaled while cell/auto/% sizing scales to device. Otherwise `px`-sized images double-scaled on HiDPI. There's a regression test for it.

**Relation to #5861.** This PR only does the *device-resolution rendering* half — it still resizes in the handler rather than keeping a pristine `orig` and computing `active` late. Happy to grow it toward the full store-original design, or land this as the foundation and do that as a follow-up, whichever you prefer.

I've also added a CI-runnable Playwright suite that covers this: IIP/kitty/sixel × sizing mode × devicePixelRatio (1, 1.5, 2), asserting the layer backing store is at device resolution and the on-screen footprint stays devicePixelRatio-invariant. Before/after screenshots are in the PR description above.

Ancientkingg:
I independently reproduced #5035 with iTerm2 inline images in a macOS Tauri/WKWebView terminal. Using device-pixel backing dimensions while keeping the CSS dimensions unchanged fixed the blur and matched the same image rendered through Kitty in Ghostty.

jerch:
@mgabor3141 & @Ancientkingg - could you please test if #6145 solves your DPR blurring issues?

PR #5785, opened 2026-03-27
issues with clearing setWrapped when erasing

PerBothner:
If you clear to the end of a line (using `eraseInLine` with param 0 or 2) it seems to me you want the `isWrapped` flag of the _following_ line to be cleared, but that doesn't seem to be happening (based on the tests added to `InputHandler.test.ts`). There is a special case in `eraseInDisplay` to un-wrap the following line in the `param==1` case  - but that code is wrong, because it doesn't add `ybase` to `j`. Also, it seems highly inconsistent.

I propose for simplicity and consistency we should just clear `isWrapped` for the current line if `start==0` and clear `isWrapped` for the following line if `end` is the end of the line. I haven't verified if xterm or other terminals do it the same way, but this rule seems simple and logical. As a bonus, we can remove the `clearWrap` parameter of `_eraseInBufferLine`.


PerBothner:
This has been incorporated into PR #5797. The testsuite changes in `InputHandler.test.ts` are the same in both PRs, but the impementation in `InputHandler.ts` is slightky different

Tyriar:
Thoughts @jerch?

PerBothner:
Reminder - this is a small change that would be nice to get resolved.

PR #5791, opened 2026-04-02
Fix kitty keyboard "Report event types" behavior not matching spec

### LINKED ISSUE #38651: Shift modifier on input 0-9 blocked in VSCode terminal when upgrading from Neovim 0.11.7 to 0.12.0
### Problem

When upgrading from Neovim 0.11.7 to 0.12.0, Shift-modified number key input (0–9) stops being correctly recognized in the VSCode integrated terminal. 

Characters affected: `!@#$%^&*()`
Other shift modifiers works fine (e.g. `~_+{}`)

The issue is reproducible across multiple systems and persists even when using nightly builds (v0.13.0-dev-29+g1bcf2d7f90).

Reverting to 0.11.7 resolves the issue immediately.

This appears specific to VSCode terminal + Neovim 0.12+ input handling.

### Steps to reproduce
inside VSCode integrated terminal
```bash
nvim --clean
:edit foo
```

Test Shift-modified number input (Shift+0..Shift+9) 
Try typing any of: `!@#$%^&*()`
or try quitting using `:q!`

Upgrade from:
  Neovim 0.11.7 → 0.12.0 (or newer)

Observe that Shift-modified numeric input is no longer handled correctly.

Test again with:
  Neovim 0.11.7 → works as expected

### Expected behavior

Shift-modified number key input should be consistently recognized across versions:

- 0.11.7: works correctly
- 0.12.0: should behave identically
- nightly (v0.13.0-dev-29+g1bcf2d7f90): should behave identically

VSCode integrated terminal should not lose or block Shift-modified numeric input after upgrade.

### Nvim version (nvim -v)

0.12.0

### Vim (not Nvim) behaves the same?

No, 8.0.1763

### Operating system/version

Rocky 8.10 (WSL) and Ubuntu 24.04.2 LTS(WSL)

### Terminal name/version

VSCode integrated terminal (code --version 1.111.0)

### $TERM environment variable

xterm-256color

### Installation

build from repo

zeertzjq:
According to spec:
- The Enter, Tab and Backspace keys will not have release events unless
  "Report all keys as escape codes" is also set.
- Key events that result in text are reported as plain UTF-8 text, so
  events are not supported for them, unless the application requests
  "key report" mode.

I checked kitty's behavior. While it does report release events for keys
that result in text as CSI sequences, press and repeat events for these
are reported as plain UTF-8 text. This commit changes Xterm.js to match
this behavior.

Fix neovim/neovim#38651

### DIFF
--- a/src/common/input/KittyKeyboard.ts
+++ b/src/common/input/KittyKeyboard.ts
@@ -477,7 +477,12 @@ export class KittyKeyboard {
 
     if (flags & KittyKeyboardFlags.REPORT_ALL_KEYS_AS_ESCAPE_CODES) {
       useCsiU = true;
-    } else if (reportEventTypes) {
+    } else if (reportEventTypes && eventType === KittyKeyboardEventType.RELEASE) {
+      // Per spec, Enter/Tab/Backspace will not have release events unless "Report all keys as
+      // escape codes" (which is handled by the branch above) is also set.
+      if (keyCode === 13 || keyCode === 9 || keyCode === 127) {
+        return result;
+      }
       useCsiU = true;
     } else if (flags & KittyKeyboardFlags.DISAMBIGUATE_ESCAPE_CODES) {
       // Per spec, Enter/Tab/Backspace "still generate the same bytes as in legacy
--- a/src/common/input/KittyKeyboard.ts
+++ b/src/common/input/KittyKeyboard.ts
@@ -471,34 +471,33 @@ export class KittyKeyboard {
       return result;
     }
 
+    // Special handling for Enter/Tab/Backspace.
+    const specialKey = keyCode === 13 || keyCode === 9 || keyCode === 127;
+
+    // Per spec, Enter/Tab/Backspace will not have release events unless "Report all keys as escape
+    // codes" is also set.
+    if (specialKey && eventType === KittyKeyboardEventType.RELEASE && !(flags & KittyKeyboardFlags.REPORT_ALL_KEYS_AS_ESCAPE_CODES)) {
+      return result;
+    }
+
     const isFunc = this._functionalKeyCodes[ev.key] !== undefined || this._getNumpadKeyCode(ev) !== undefined;
 
-    let useCsiU = false;
-
-    if (flags & KittyKeyboardFlags.REPORT_ALL_KEYS_AS_ESCAPE_CODES) {
-      useCsiU = true;
-    } else if (reportEventTypes && eventType === KittyKeyboardEventType.RELEASE) {
-      // Per spec, Enter/Tab/Backspace will not have release events unless "Report all keys as
-      // escape codes" (which is handled by the branch above) is also set.
-      if (keyCode === 13 || keyCode === 9 || keyCode === 127) {
-        return result;
-      }
-      useCsiU = true;
-    } else if (flags & KittyKeyboardFlags.DISAMBIGUATE_ESCAPE_CODES) {
-      // Per spec, Enter/Tab/Backspace "still generate the same bytes as in legacy
-      // mode" and consider space to be a text-generating key, so these skip the isFunc fast-path
-      // and only get CSI u when modifiers are present (handled below).
-      const isDisambiguateLegacy = keyCode === 13 || keyCode === 9 || keyCode === 127;
-      if (isFunc && !isDisambiguateLegacy) {
-        useCsiU = true;
-      } else if (modifiers > 0) {
-        if (ev.shiftKey && !ev.ctrlKey && !ev.altKey && !ev.metaKey && ev.key.length === 1) {
-          useCsiU = false;
-        } else {
-          useCsiU = true;
-        }
-      }
-    }
+    const useCsiU = !!(
+      flags & KittyKeyboardFlags.REPORT_ALL_KEYS_AS_ESCAPE_CODES ||
+      (reportEventTypes && eventType === KittyKeyboardEventType.RELEASE) ||
+      (flags & KittyKeyboardFlags.DISAMBIGUATE_ESCAPE_CODES &&
+        (
+          // Per spec, Enter/Tab/Backspace "still generate the same bytes as in legacy mode" and
+          // consider space to be a text-generating key, so these skip the isFunc fast-path and only
+          // get CSI u when modifiers are present (handled below).
+          (isFunc && !specialKey) ||
+          (
+            (modifiers > 0 && ev.key.length !== 1) ||
+            modifiers - 1 > KittyKeyboardModifiers.SHIFT
+          )
+        )
+      )
+    );
 
     if (useCsiU) {
       result.key = this._buildCsiUSequence(ev, keyCode, modifiers, eventType, flags, isFunc, isMod);
--- a/src/common/input/KittyKeyboard.test.ts
+++ b/src/common/input/KittyKeyboard.test.ts
@@ -453,9 +453,29 @@ describe('KittyKeyboard', () => {
     describe('event types (press/repeat/release)', () => {
       const flags = KittyKeyboardFlags.DISAMBIGUATE_ESCAPE_CODES | KittyKeyboardFlags.REPORT_EVENT_TYPES;
 
-      it('press event (default, no suffix)', () => {
+      it('UTF-8 text press event', () => {
         const result = kitty.evaluate(createEvent({ key: 'a' }), flags, KittyKeyboardEventType.PRESS);
-        assert.strictEqual(result.key, '\x1b[97u');
+        assert.strictEqual(result.key, 'a');
+      });
+
+      it('Escape key press event (default, no suffix)', () => {
+        const result = kitty.evaluate(createEvent({ key: 'Escape' }), flags, KittyKeyboardEventType.PRESS);
+        assert.strictEqual(result.key, '\x1b[27u');
+      });
+
+      it('Enter key press event → legacy \\r', () => {
+        const result = kitty.evaluate(createEvent({ key: 'Enter' }), flags, KittyKeyboardEventType.PRESS);
+        assert.strictEqual(result.key, '\r');
+      });
+
+      it('Tab key press event → legacy \\t', () => {
+        const result = kitty.evaluate(createEvent({ key: 'Tab' }), flags, KittyKeyboardEventType.PRESS);
+        assert.strictEqual(result.key, '\t');
+      });
+
+      it('Backspace key press event → legacy \\x7f', () => {
+        const result = kitty.evaluate(createEvent({ key: 'Backspace' }), flags, KittyKeyboardEventType.PRESS);
+        assert.strictEqual(result.key, '\x7f');
       });
 
       it('press event explicit :1 when modifiers present', () => {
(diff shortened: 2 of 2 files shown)


scutuatua-crypto (approved):

justinmk:
can we get a review on this @Tyriar @jerch

src/common/input/KittyKeyboard.ts
jerch: I suggest to pull the change out of the big if-else cascade, similar to this:
```suggestion
    const keyCode = this._getKeyCode(ev, macOptionAsAlt);
    if (keyCode === undefined) {
      return result;
    }

    // special handing for Enter/Tab/Backspace
    const specialKey = keyCode === 13 || keyCode === 9 || keyCode === 127;

    if (reportEventTypes && eventType === KittyKeyboardEventType.RELEASE && specialKey) {
      // Enter/Tab/Backspace do not report RELEASE in REPORT_EVENT_TYPES
      return result;
    }

    const isFunc = this._functionalKeyCodes[ev.key] !== undefined || this._getNumpadKeyCode(ev) !== undefined;

    let useCsiU = false;

    if (flags & KittyKeyboardFlags.REPORT_ALL_KEYS_AS_ESCAPE_CODES ||
      (reportEventTypes && eventType === KittyKeyboardEventType.RELEASE)
    ) {
      useCsiU = true;
    } else if (flags & KittyKeyboardFlags.DISAMBIGUATE_ESCAPE_CODES) {
      // Per spec, Enter/Tab/Backspace "still generate the same bytes as in legacy
      // mode" and consider space to be a text-generating key, so these skip the isFunc fast-path
      // and only get CSI u when modifiers are present (handled below).
      if ((isFunc && !specialKey) ||
        ((modifiers > 0) &&
          (!ev.shiftKey || ev.ctrlKey || ev.altKey || ev.metaKey || ev.key.length !== 1)
        )
      ) {
        useCsiU = true;
      }
    }
```

Edit: Or even more aggressive - fold `useCsiU` into a single assignment:
```typescript
    const useCsiU = !!(
      flags & KittyKeyboardFlags.REPORT_ALL_KEYS_AS_ESCAPE_CODES ||
      (reportEventTypes && eventType === KittyKeyboardEventType.RELEASE) ||
      (flags & KittyKeyboardFlags.DISAMBIGUATE_ESCAPE_CODES &&
        (
          (isFunc && !specialKey) ||
          (
            (modifiers > 0 && ev.key.length !== 1) ||
            modifiers - 1 > KittyKeyboardModifiers.SHIFT
          )
        )
      )
    );
```

jerch (dismissed):
The control flow is a bit hard to follow in the existing code. 😱 

One code reformatting suggestion below.

src/common/input/KittyKeyboard.ts
zeertzjq: I don't think the first suggested changes are correct. It doesn't report Enter/Tab/Backspace release events even if `REPORT_ALL_KEYS_AS_ESCAPE_CODES` is set.

jerch: Why would that be? 

```typescript
    if (flags & KittyKeyboardFlags.REPORT_ALL_KEYS_AS_ESCAPE_CODES ||
      (reportEventTypes && eventType === KittyKeyboardEventType.RELEASE)
    ) {
      useCsiU = true;
    } ...
```
The condition for `REPORT_ALL_KEYS_AS_ESCAPE_CODES` did not change.

zeertzjq: No, but you added a `return` statement above it.

zeertzjq: I added a refactoring commit. I also added a few more tests, but somehow I cannot run the tests anymore:
```
> @xterm/xterm@6.0.0 test-unit
> node ./bin/test_unit.js out-esbuild/common/input/KittyKeyboard.test.js

file:///home/zeertzjq/dev/xterm.js/node_modules/yargs/yargs:3
const {applyExtends, cjsPlatformShim, Parser, Yargs, processArgv} = require('./build/index.cjs')
                                                                    ^

ReferenceError: require is not defined in ES module scope, you can use import instead
    at file:///home/zeertzjq/dev/xterm.js/node_modules/yargs/yargs:3:69
    at ModuleJobSync.runSync (node:internal/modules/esm/module_job:543:37)
    at ModuleLoader.importSyncForRequire (node:internal/modules/esm/loader:353:47)
    at loadESMFromCJS (node:internal/modules/cjs/loader:1675:24)
    at Module._compile (node:internal/modules/cjs/loader:1838:5)
    at Object..js (node:internal/modules/cjs/loader:1996:10)
    at Module.load (node:internal/modules/cjs/loader:1579:32)
    at Module._load (node:internal/modules/cjs/loader:1381:12)
    at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
    at Module.require (node:internal/modules/cjs/loader:1602:12)

Node.js v26.1.0
```

jerch: > No, but you added a `return` statement above it.

You are right, I overlooked the _unless [Report all keys as escape codes](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#report-all-keys) is **also set**_ in the specs.

About the test run issue:
Please use node 22 oder 24. Newer version are not tested yet and prolly need package upgrades.

zeertzjq: Ah yes. The tests work if I downgrade to Node 25.

Tyriar (approved):
👍

jerch:
@Tyriar I am not sure yet about the state of this fix or the kitty keboard protocol in general. Did you a quick comparison with kitty 0.46.2 - the behavior still deviate for certain cases.

Also the spec docs seems slightly off to the actual behavior, example: under https://sw.kovidgoyal.net/kitty/keyboard-protocol/#report-event-types the spec states, that 

> Report event types ... causes the terminal to report key **repeat** and key **release** events. ... Except Enter, Tab and Backspace keys will not have **release** events unless [Report all keys as escape codes](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#report-all-keys) is also set, ...

But in kitty not only release events are swallowed for Enter/Tab/Backspace, but also repeat events.

Do we have a comprehensive test suite for what is expected to be reported for the different modes and flags?

zeertzjq:
> But in kitty not only release events are swallowed for Enter/Tab/Backspace, but also repeat events.

What do you mean by "swallowed"? I tested kitty 0.46.2, and repeating Enter/Tab/Backspace still produces a plain char each time.

jerch:
@zeertzjq Try Strg-a in Report event types, we deviate there.

Edit: I think this clause is the culprit (2nd condition):
https://github.com/xtermjs/xterm.js/blob/7059a07c0920e8592e4ca1965e72e7ad84a02f10/src/common/input/KittyKeyboard.ts#L487

zeertzjq:
Ah yes, I didn't test `REPORT_EVENT_TYPES` without `DISAMBIGUATE_ESCAPE_CODES` much.

jerch:
The repeats should come in all browsers as I just verified with https://w3c.github.io/uievents/tools/key-event-viewer.html.

Also I get different codes in xterm.js and kitty, so there is more off here. That's why the question for a comprehensive test suite. This whole keyboard stuff in browsers is an annoyingly complicated issue haha.

zeertzjq:
This seems to match kitty's behavior:
```diff
diff --git a/src/common/input/KittyKeyboard.ts b/src/common/input/KittyKeyboard.ts
index f35cfebd..5b65dd39 100644
--- a/src/common/input/KittyKeyboard.ts
+++ b/src/common/input/KittyKeyboard.ts
@@ -481,20 +481,17 @@ export class KittyKeyboard {
     }
 
     const isFunc = this._functionalKeyCodes[ev.key] !== undefined || this._getNumpadKeyCode(ev) !== undefined;
+    const reportModifiers = (modifiers > 0 && ev.key.length !== 1) || modifiers - 1 > KittyKeyboardModifiers.SHIFT;
 
     const useCsiU = !!(
       flags & KittyKeyboardFlags.REPORT_ALL_KEYS_AS_ESCAPE_CODES ||
-      (reportEventTypes && eventType === KittyKeyboardEventType.RELEASE) ||
+      (reportEventTypes && (eventType === KittyKeyboardEventType.RELEASE || reportModifiers)) ||
       (flags & KittyKeyboardFlags.DISAMBIGUATE_ESCAPE_CODES &&
         (
           // Per spec, Enter/Tab/Backspace "still generate the same bytes as in legacy mode" and
           // consider space to be a text-generating key, so these skip the isFunc fast-path and only
           // get CSI u when modifiers are present (handled below).
-          (isFunc && !specialKey) ||
-          (
-            (modifiers > 0 && ev.key.length !== 1) ||
-            modifiers - 1 > KittyKeyboardModifiers.SHIFT
-          )
+          (isFunc && !specialKey) || reportModifiers
         )
       )
     );
```

zeertzjq:
> This seems to match kitty's behavior:

The modifier behavior seems correct, but the functional key behavior with only `REPORT_EVENT_TYPES` seems still not fully right. I'll take another look tomorrow...

zeertzjq:
I did a bit more testing and I think kitty's behavior is a bit strange. I've created <https://github.com/kovidgoyal/kitty/issues/9999> to check if this behavior is really intentional.

**Update:** Kovid thinks that enabling "Report event types" without "Disambiguate escape codes" doesn't make sense, which I agree.

jerch:
> **Update:** Kovid thinks that enabling "Report event types" without "Disambiguate escape codes" doesn't make sense, which I agree.

That is a doc/spec issue then. It should clearly be noted, which bit flag combinations are meant to be useful and which are left as undefined/not useful. Bit flags suggest they can be combined freely, so people will do that.

Tyriar:
@jerch I kind of expect a long tail of little issues like this as I built it relatively hastily and was heavily experimenting with LLMs just in this component (and the off-by-default win32 input mode which should arguably be ripped out as it's kind of broken and no time to fix). Shipped super fast, but it immediately broke some important behavior, I think it was ctrl+c 😅. If existing tests are passing and there are new ones I haven't been reviewing the code here super closely.

It would definitely be lovely if an official comprehensive test suite existed. I spent a bunch of time manually diffing `kitten` with what xterm.js was outputting while building this.

jerch:
> It would definitely be lovely if an official comprehensive test suite existed. I spent a bunch of time manually diffing `kitten` with what xterm.js was outputting while building this.

I imagine the hassles as I had to do exactly that yesterday, too. Just to find the Ctrl-a issue after playing around.

@zeertzjq Do you happen to a have key recorder script for different keyboard layouts in neovim? If not, I could write such a thing for my keyboard (PC style) and someone would have to provide one for Mac keyboards. Only annoying issue for me is - I have literally no clue which bit combinations are useful, so the test base explodes in a combinatorial way :scream:

zeertzjq:
> Do you happen to a have key recorder script for different keyboard layouts in neovim?

I'm not entirely sure what you are referring to. On Linux `--log-in` flag of the `script` command from util-linux logs input to a terminal program. I'm not sure about other platforms.

Although I usually just use stuff like `printf '\x1b[>3u'; cat; printf '\x1b[<u'` in the shell to test kitty keyboard protocol.

jerch:
> Although I usually just use stuff like printf '\x1b[>3u'; cat; printf '\x1b[<u' in the shell to test kitty keyboard protocol.

Yeah that is what I did yesterday for  different modes. I think about doing this in a more structured way with a script capturing the reported values under different modes as a test base to test against.

PR #5798, opened 2026-04-10
addon-search:  Add "Find Nth" functionality

boborrob:
### Summary: 
Add "Find Nth" functionality to the terminal search feature in order to access  matches arbitrarily.

### Motivation:
For large sets of search results, cycling through matches sequentially can be cumbersome and time-consuming, especially when the location of the desired match is strongly suspected or known with precision. An input allows the user to jump arbitrarily to the desired match or to get closer to it much faster. 

### Changes:
- Add an `n` field to  `searchOptions` for locating matches arbitrarily.

- Add a `findNth(...)`  method analogous to  `findNext(...)` and `findPrev(...)`.

- Add baseline validations for `n` (!isNaN, isTruthy, etc.). Rigorous validation is delegated to the client where the bounds are better understood pre-invocation.

- Update demo app with 'Find Nth' search controls.

### Example use cases:

VS Code `terminalFindWidget`

Screenshot_20260409_225319-2

### Demo App Updates

Screenshot_20260409_224725

### DIFF
--- a/addons/addon-search/src/SearchAddon.ts
+++ b/addons/addon-search/src/SearchAddon.ts
@@ -120,6 +120,41 @@ export class SearchAddon extends Disposable implements ITerminalAddon, ISearchAp
     return found;
   }
 
+  /**
+   * Find the Nth instance of the term, then scroll to and select it. If it
+   * doesn't exist, do nothing.
+   * @param term The search term.
+   * @param searchOptions Search options.
+   * @returns Whether a result was found.
+   */
+  public findNth(term: string, searchOptions?: ISearchOptions, internalSearchOptions?: IInternalSearchOptions): boolean {
+    if (!this._terminal || !this._engine) {
+      throw new Error('Cannot use addon until it has been loaded');
+    }
+
+    // If nthMatchPosition is invalid for any reason, pass along -1 as a substitute.
+    // That way, `clearDecoration()` logic is guaranteed to run and reflect match failure.
+    if (!searchOptions || isNaN(parseInt(`${searchOptions.nthMatchPosition}`, 10)) || !searchOptions.nthMatchPosition) {
+      searchOptions = { nthMatchPosition: -1, ...(searchOptions ?? {}) };
+    }
+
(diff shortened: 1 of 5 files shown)


Tyriar (commented):
Thanks for the PR, but we don't want this. If a terminal wanted to implement this, they can achieve it by calling findNext multiple times. I don't think it should or would be added to vscode though.

boborrob:
Understandable and thanks for the consideration.  

TL;DR 
Friendly rebuttal: Why loop with O(n) complexity over a contiguous, indexable array when an O(1) lookup is possible?

--

A looping approach was considered early on, but since the underlying data structure is array-like and indexable, that all seemed unnecessary. A loop would make more sense if the structure were a linked list where only 1 node at a time could be accessed in either direction. 

Some more considerations: Calling either find method N times could be costly if the total number of matches is large enough, especially if the current match is far from the target match. Then there's the question of which direction to search in.  Since the target match could be  behind or ahead of the current one, both the distance and the direction would need to be determined to minimize travel. 

Seems like more overhead to me given that the structure already lends itself really well to constant-time lookups. 

Anyway, it'd be great if this were added to VS Code. The match counts are nice to look at, but I'm not sure how  valuable the information is if it isn't actionable.

Tyriar:
@boborrob it's not O(n) after https://github.com/xtermjs/xterm.js/pull/5902, should be pretty fast now.

boborrob:
I see. I suspect that [`resultTracker.searchResults`](https://github.com/boborrob/xterm.js/blob/addon_search-arbitrary_access_to_matches/addons/addon-search/src/SearchAddon.ts#L219) is a higher-level, array-like abstraction of #5902's cache. Correct me if I'm wrong. 

In O(1) and from any start location, the desired match can be accessed directly from that structure via an  index (N - 1) . The work in #5902 builds and abstracts the cache so well that retrieving matches from it becomes trivial. One call, as opposed to many, can leverage all that work. 

It's a compliment, really.

PR #5813, opened 2026-04-13
Use element window instead of global object

CGNonofr:
If xterm is running in another frame than its dom, `window.getComputedStyle` returns an empty object in Firefox, which completely breaks the user selections


CGNonofr:
Anything I can do to help merge this very simple fix? it's really a pain to patch it after every update, especially because the released version is minified for some reasons :thinking:

Tyriar (approved):

CGNonofr:
Thanks :+1: :100:

PR #5821, opened 2026-04-17
Hold this via WeakRef in IntersectionObserver callback

### LINKED ISSUE #5820: IntersectionObserver callback retains RenderService → BufferLines after dispose
## Problem

`RenderService._registerIntersectionObserver` creates an `IntersectionObserver` whose callback closes over `this`:

https://github.com/xtermjs/xterm.js/blob/master/src/browser/services/RenderService.ts#L129

```ts
const observer = new w.IntersectionObserver(
  e => this._handleIntersectionChange(e[e.length - 1]),
  { threshold: 0 }
);
```

On dispose, `_observerDisposable` calls `observer.disconnect()`. In theory that releases the callback. In practice — verified against a production Chrome tab — the callback's closure lives on, keeping `this` (RenderService) reachable, and with it the whole service graph:

```
RenderService._coreService
  → CoreService._bufferService
  → BufferService.buffers._normal
  → Buffer.lines (CircularList)
  → BufferLine[]
  → Uint32Array (cell data, ~1.3 KB each + native ArrayBuffer backing)
```

## Evidence

Heap-snapshot diff of a multi-tile terminal app ([kolu](https://github.com/juspay/kolu)) across 30 mount/unmount cycles of 7 `Terminal` instances:

| Class | Δ count | Δ bytes |
|---|---|---|
| `native:system / JSArrayBufferData` | +175,594 | **+220 MB** |
| `object:Uint32Array` | +175,594 | +10 MB |
| `object:ArrayBuffer` | +175,594 | +9 MB |
| `object:BufferLine` (minified `Z1`) | +175,594 | +5 MB |

All 175,594 retained `Uint32Array` instances traced through the same retainer signature:

```
Window.IntersectionObserver (registry)
  → callback closure
  → Context
  → RenderService (minified ep)
  → _coreService (sp) → _bufferService (ap)
  → buffers._normal (CS) → .lines (SS)
  → ._array → BufferLine (Z1)
  → ._data → Uint32Array
```

175,594 retained BufferLines = 30 toggles × 7 terminals × ~830 lines per terminal — basically the full scrollback buffer of every disposed `Terminal` was being held past `terminal.dispose()`.

## Proposal

Hold `this` in the observer callback via `WeakRef` instead of a direct closure capture. Then even if the callback is retained somewhere past `observer.disconnect()`, the RenderService and its service graph (including `Uint32Array` cell buffers) become GC-eligible on dispose.

```diff
   private _registerIntersectionObserver(w: Window & typeof globalThis, screenElement: HTMLElement): void {
     if ('IntersectionObserver' in w) {
-      const observer = new w.IntersectionObserver(
-        e => this._handleIntersectionChange(e[e.length - 1]),
-        { threshold: 0 }
-      );
+      const weakSelf = new WeakRef(this);
+      const observer = new w.IntersectionObserver(
+        e => weakSelf.deref()?._handleIntersectionChange(e[e.length - 1]),
+        { threshold: 0 }
+      );
       observer.observe(screenElement);
       this._observerDisposable.value = toDisposable(() => observer.disconnect());
     }
   }
```

Functional behavior is preserved: while RenderService is alive, `weakSelf.deref()` returns it. Once no strong refs remain, the intersection callback becomes a no-op and the buffer graph is free to GC.

Happy to send a PR. I have not chased down *why* `disconnect()` isn't fully releasing the callback in practice (devtools extension? browser native registry? different for Chrome vs Firefox?) — but the defensive WeakRef wrap breaks the retention regardless of the underlying cause.

## Related

- #5817 (juspay/xterm.js `fix/dispose-leaks`) addresses two adjacent leaks (`CursorBlinkStateManager` + `_pausedResizeTask`) — independent from this one.
- Verified locally against the juspay fork on top of #5817.

🤖 Diagnosis assisted by [Claude Code](https://claude.com/claude-code)

srid:
## Summary

Fix for #5820. `RenderService._registerIntersectionObserver` creates an `IntersectionObserver` whose callback closes over `this` directly. Although `_observerDisposable` calls `observer.disconnect()` on dispose, the retained callback chain keeps `this` (RenderService) alive in practice — along with `_coreService → _bufferService → buffers → BufferLine → Uint32Array` cell data.

Wrap `this` in a `WeakRef`. While RenderService is alive, `deref()` returns it and the handler runs unchanged. Once no strong refs remain, the callback becomes a no-op and the BufferService graph can GC.

## Evidence (from #5820)

Heap-snapshot diff across 30 mount/unmount cycles of 7 `Terminal` instances in a multi-tile app:

| Class | Δ count | Δ bytes |
|---|---|---|
| `native:system / JSArrayBufferData` | +175,594 | **+220 MB** |
| `object:Uint32Array` | +175,594 | +10 MB |
| `object:ArrayBuffer` | +175,594 | +9 MB |
| `object:BufferLine` | +175,594 | +5 MB |

Every retained `Uint32Array` traced to the same retainer chain — global `IntersectionObserver` registry → callback closure → `RenderService` → service graph → BufferLines. After the WeakRef fix, production-equivalent local testing shows Memory Footprint flat across 30 mode-toggles (was +367 MB/30-toggles before).

## Why not just fix `disconnect()`?

Unclear why `observer.disconnect()` isn't releasing the callback in the observed environment — could be DevTools instrumentation, a Chrome extension patching `window.IntersectionObserver`, a native registry quirk, or something else. The WeakRef wrap is a defensive fix that breaks the retention chain regardless of the underlying cause and preserves functional semantics exactly.

## Test

No behavioral change to validate beyond existing renderer-pause-on-visibility tests. A regression would manifest as the pause handler not firing when visibility changes; while RenderService is strongly rooted (which it is for any live terminal), `deref()` always returns the instance.

Closes #5820

### DIFF
--- a/src/browser/services/RenderService.ts
+++ b/src/browser/services/RenderService.ts
@@ -124,9 +124,25 @@ export class RenderService extends Disposable implements IRenderService {
 
   private _registerIntersectionObserver(w: Window & typeof globalThis, screenElement: HTMLElement): void {
     // Detect whether IntersectionObserver is detected and enable renderer pause
-    // and resume based on terminal visibility if so
+    // and resume based on terminal visibility if so.
+    //
+    // The callback closes over a WeakRef to `this` rather than a strong
+    // reference. Even though `_observerDisposable` calls `observer.disconnect()`
+    // on disposal, in practice some browser-side (or devtools/extension-side)
+    // registries retain the callback past disconnect — the retained closure
+    // then keeps `this` (RenderService) alive, which transitively keeps
+    // `_coreService → _bufferService → buffers → BufferLines → Uint32Array`
+    // alive. Each leaked terminal was pinning ~1.3 KB × scrollback-lines of
+    // cell data, producing hundreds of MB of retained native buffer bytes
+    // across mode-toggle churn. WeakRef lets the RenderService (and hence
+    // its service graph and BufferLines) GC even if the callback itself
+    // is retained.
     if ('IntersectionObserver' in w) {
-      const observer = new w.IntersectionObserver(e => this._handleIntersectionChange(e[e.length - 1]), { threshold: 0 });
+      const weakSelf = new WeakRef(this);
+      const observer = new w.IntersectionObserver(
+        e => weakSelf.deref()?._handleIntersectionChange(e[e.length - 1]),
+        { threshold: 0 }
+      );
       observer.observe(screenElement);
       this._observerDisposable.value = toDisposable(() => observer.disconnect());
     }


src/browser/services/RenderService.ts
```diff
@@ -124,9 +124,25 @@ export class RenderService extends Disposable implements IRenderService {
...
+    // is retained.
     if ('IntersectionObserver' in w) {
-      const observer = new w.IntersectionObserver(e => this._handleIntersectionChange(e[e.length - 1]), { threshold: 0 });
+      const weakSelf = new WeakRef(this);
+      const observer = new w.IntersectionObserver(
+        e => weakSelf.deref()?._handleIntersectionChange(e[e.length - 1]),
+        { threshold: 0 }
+      );
       observer.observe(screenElement);
       this._observerDisposable.value = toDisposable(() => observer.disconnect());
```
Tyriar, line 147: I'm surprised this isn't handled by `observer.disconnect()` dropping the refs to the callback. Is there a way to solve this without the introduction of `WeakRef`?

Tyriar (changes requested):

src/browser/services/RenderService.ts
jerch: To me this smells like a browser bug and might be better addressed in chrome directly. Should also be tested, whether it is reproducible on other browser engines.

jerch: Checked the w3c spec - `disconnect` does indeed not nullify the callback, in fact the observer object holds onto the callback to be reusable with new `.observe(some_element)` calls. So this is indeed specced like this.

Would it help to just throw away the observer object itself for the GC to clean the callback up?

Tyriar: As in `observer = undefined`?

jerch: Yepp (and to make sure, the ref gets not bound elsewhere).

jerch:
@srid In https://github.com/juspay/kolu/pull/652 you write, that the callback retention is not reproducible locally with chrome-devtools MCP (shows always benign behavior), but with normal chrome + devtools it is and that only the weakref counters it effectively.

Can you please repeat the tests with devtools not opened, e.g. by adding the async snippet on toplevel and let it trigger the toggles by a timeout from normal load? then inspect the memory footprint again. Just want to make sure, that this is not a devtools instrumentation bug (devtools exhibits several of those, esp. during perf testing with screwed up timer events and overly big websocket load).

src/browser/services/RenderService.ts
Tyriar: That would be preferable to a `WeakRef`, every time I've thought I needed it there was something simpler like this.

srid:
> In [juspay/kolu#652](https://github.com/juspay/kolu/pull/652) you write, that the callback retention is not reproducible locally with chrome-devtools MCP (shows always benign behavior)

@jerch This was only because Kolu had already moved to an unified infinite canvas model (no more two-mode switching). And I had to go back in history to old model, and reproduce it there. Sharing results shortly, in the thread above.

src/browser/services/RenderService.ts
srid: Confirmed this to help too. 

So, we can close this PR and choose https://github.com/xtermjs/xterm.js/pull/5831 instead.

srid: Hi folks, is https://github.com/xtermjs/xterm.js/pull/5831 ready for merge, or would you like me to follow-up on anything? This is the only patch that keeps me from using upstream xterm.js, and it would be good to get it into next xterm release, so https://kolu.dev/ use the stable version.

Tyriar:
Merged https://github.com/xtermjs/xterm.js/pull/5831, thanks!

PR #5824, opened 2026-04-19
Fix text-producing keys in Kitty keyboard REPORT_EVENT_TYPES mode

### LINKED ISSUE #5823: In Neovim Shift+number keys produce digits instead of symbols with Kitty keyboard protocol
### Problem

In VS Code 1.116  integrated terminal, `Shift+number` keys (e.g., `Shift+1` for `!`) produce digits instead of symbols when using Neovim. This only affects VS Code terminal; Ghostty, and other terminals work correctly.

### Environment

- **Neovim**: nightly (post-2026-04-19)
- **VS Code**: 1.116 (released 2026-04)
- **OS**: macOS (Windows/Linux unconfirmed)
- **Terminal**: VS Code integrated terminal (xterm.js)

### Steps to reproduce

1. Open VS Code integrated terminal
2. Launch `nvim`
3. Enter `:`
4. Press `Shift+1` expecting `!`
5. Observe `1` appears instead

### Root cause

VS Code added Kitty keyboard protocol support via xterm.js. Neovim detects this and sends `CSI > 3 u` to enable:
- bit0: Disambiguate escape codes
- bit1: Report event types

wngtk:
Attempt to fix #5823 

In REPORT_EVENT_TYPES mode (CSI > 3 u), text-producing keys like Shift+1 should send the actual character ('!') for PRESS events, not CSI u escape sequences. Only REPEAT and RELEASE events should use CSI u for text-producing keys, while function keys use CSI u for all event types.

Fixes issue where Shift+1 was not outputting '!' character.

### DIFF
--- a/src/common/input/KittyKeyboard.test.ts
+++ b/src/common/input/KittyKeyboard.test.ts
@@ -453,9 +453,21 @@ describe('KittyKeyboard', () => {
     describe('event types (press/repeat/release)', () => {
       const flags = KittyKeyboardFlags.DISAMBIGUATE_ESCAPE_CODES | KittyKeyboardFlags.REPORT_EVENT_TYPES;
 
-      it('press event (default, no suffix)', () => {
+      it('text-producing key uses legacy encoding in REPORT_EVENT_TYPES mode', () => {
+        // Per spec: "In report event types mode, function keys are encoded as CSI u...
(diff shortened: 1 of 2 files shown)


niclimcy:
What is blocking this PR? Would be nice to see this fixed

jerch:
This is actually fixed in master.

PR #5825, opened 2026-04-21
perf: CSI fast-path (+42% throughput)

homanp:
## Summary

+28% parser throughput by fast-pathing CSI sequences instead of running them through the per-byte state machine. All 2261 tests pass, no behavior changes.

## Benchmarks

5 workloads, median of 3 fresh node processes (Node 23, macOS):

| workload | before | after | change |
|---|---|---|---|
| sgr_heavy | 1,189 MB/s | 1,952 MB/s | +64% |
| cursor_move | 1,522 MB/s | 2,240 MB/s | +47% |
| params_heavy | 1,050 MB/s | 1,681 MB/s | +60% |
| mixed_realistic | 1,710 MB/s | 2,445 MB/s | +43% |
| print_ascii | 3,832 MB/s | 3,873 MB/s | ~0% |
| **weighted** | **1,949 MB/s** | **2,487 MB/s** | **+28%** |

print_ascii doesn't move because the existing PRINT unrolling already handles that well. all the gains are on escape-sequence-heavy output, which is most real terminal content (colors, cursor moves, truecolor).

## What changed

When the parser is in GROUND and sees `ESC [`, the new code skips the transition table entirely and consumes the CSI sequence in one tight loop — check for private-mode char, accumulate params, dispatch on the final byte. after dispatch it scans ahead for a print run so we don't bounce back to the main loop just to handle the text that almost always follows a color change.

The state machine is still there for everything else. the fast-path only fires for CSI in GROUND, which is the overwhelmingly common case. async handlers still work (calls `_preserveStack` if a handler returns a Promise).

## Why it helps

The per-byte table lookup is correct and handles every edge case, but for a typical `ESC [ 3 8 ; 5 ; 1 9 6 m` that's 12 table lookups + 12 switch cases for what's a completely predictable state sequence (GROUND → CSI_ENTRY → CSI_PARAM → ... → GROUND). the fast-path replaces those with direct comparisons against known byte ranges.

## Tests

All 2261 unit tests pass. the fast-path is conservative — it only activates from GROUND state, so mid-stream sequences (during DCS/OSC/etc) still go through the full state machine. if anything looks unexpected in the fast-path it falls through to the normal path.

happy to split into two commits (CSI fast-path + print-scan-ahead) if that's easier to review.


jerch:
@homanp Nice :+1: 

Can you show us the perf difference from this call (this tests the parser only skipping the more expensive bookkeeping of the buffer):
```
npm run benchmark out-test/benchmark/EscapeSequenceParser.benchmark.js
```

homanp:
> @homanp Nice 👍
> 
> Can you show us the perf difference from this call:
> 
> ```
> npm run benchmark out-test/benchmark/EscapeSequenceParser.benchmark.js
> ```

Thanks @jerch! Here are the numbers from the project's own benchmark:

| benchmark | master | this PR | change |
|---|---|---|---|
| PRINT - a | 731.5 | 686.7 | -6% |
| EXECUTE - \n | 186.2 | 180.8 | -3% |
| ESCAPE - ESC E | 88.2 | 76.8 | -13% |
| ESCAPE w/ collect | 111.4 | 103.7 | -7% |
| **CSI - CSI A** | **178.4** | **297.8** | **+67%** |
| **CSI w/ collect** | **176.3** | **285.0** | **+62%** |
| **CSI params (short)** | **190.1** | **334.6** | **+76%** |
| **CSI params (long)** | **230.2** | **364.5** | **+58%** |
| OSC string (short) | 115.9 | 111.0 | -4% |
| OSC string (long) | 125.8 | 108.0 | -14% |
| OSC class (short) | 147.9 | 145.1 | -2% |
| OSC class (long) | 501.0 | 485.7 | -3% |
| DCS string (short) | 85.8 | 80.6 | -6% |
| DCS string (long) | 120.4 | 108.7 | -10% |
| DCS class (short) | 120.6 | 100.0 | -17% |
| DCS class (long) | 415.3 | 399.8 | -4% |

CSI cases see +58-76% improvement, which is where the fast-path operates. non-CSI cases show regressions of 3-17%, the fast-path check adds a branch at the top of the main loop that penalizes sequences that don't benefit from it.

For real-world terminal output (mostly SGR/CSI), the net effect is positive. but the non-CSI regressions aren't great, especially on DCS/OSC.

One idea to reduce the overhead: only enter the fast-path check when `code === 0x1b`, rather than checking on every byte. That way the branch predictor handles the common (non-ESC) case for free. Happy to iterate on this if you think the approach is worth pursuing.

homanp:
Working on a fix for this

homanp:
@jerch moved the fast-path from the top of the main loop into the CLEAR action so it only fires on ESC bytes, not every byte. regressions are gone:

| benchmark | master | this PR | change |
|---|---|---|---|
| PRINT - a | 731.5 | 812.0 | +11% |
| EXECUTE - \n | 186.2 | 197.5 | +6% |
| ESCAPE - ESC E | 88.2 | 87.6 | -1% |
| ESCAPE w/ collect | 111.4 | 112.4 | +1% |
| **CSI - CSI A** | **178.4** | **278.3** | **+56%** |
| **CSI w/ collect** | **176.3** | **325.0** | **+84%** |
| **CSI params (short)** | **190.1** | **322.9** | **+70%** |
| **CSI params (long)** | **230.2** | **370.5** | **+61%** |
| OSC string (short) | 115.9 | 117.9 | +2% |
| OSC string (long) | 125.8 | 131.6 | +5% |
| OSC class (short) | 147.9 | 148.1 | +0% |
| OSC class (long) | 501.0 | 526.6 | +5% |
| DCS string (short) | 85.8 | 89.2 | +4% |
| DCS string (long) | 120.4 | 124.9 | +4% |
| DCS class (short) | 120.6 | 125.0 | +4% |
| DCS class (long) | 415.3 | 426.2 | +3% |

no regressions on any case. CSI throughput +56-84%. averaged across 2 runs of `npm run benchmark`.

jerch:
Yes this is worth pursuing. And since CSI sequences are the second most data bytes after pure PRINT data it would nicely speed up things.

On a technical side - I wonder if we could make it similar to the OSC and DCS paths - like swallowing CSI allowed bytes in a tight loop only from within CSI states and bail out on any other. Would avoid the outer check completely.

Edit: Oh ok, you already fixed it :smile_cat:

homanp:
> Yes this is worth pursuing. And since CSI sequences are the second most data bytes after pure PRINT data it would nicely speed up things.
> 
> On a technical side - I wonder if we could make it similar to the OSC and DCS paths - like swallowing CSI allowed bytes in a tight loop only from within CSI states and bail out on any other. Would avoid the outer check completely.
> 
> Edit: Oh ok, you already fixed it 😸

If you have any other feedback, happy to take a look.

jerch:
> If you have any other feedback, happy to take a look.

Will have a closer look at it in the evening. Thanks for bringing this up.

homanp:
> > If you have any other feedback, happy to take a look.
> 
> Will have a closer look at it in the evening. Thanks for bringing this up.

Thanks for an awesome lib and work put into this!

jerch:
@homanp Had a closer look at the changes. Here are my observations:

- The main speedup results from collapsing:
  - `table lookup:ESC --> ParserAction.CLEAR`
  - `table lookup:0x5b --> ParserAction.CLEAR`
  into one `table lookup:ESC --> ParserAction.CLEAR --> 0x5b`. So this spares a table lookup and a CLEAR call.
- params parsing gets not faster much (already optimized in master), but spares the initial table lookup.
- Final byte matching spares one table lookup.
- Intermediates are not handled (bails out to normal processing with a tiny penalty).
- `_csiFast` does handle ParserAction.PRINT at the end, which I don't understand why. Looks misplaced to me, like overoptimizing a certain pattern.

**Net effect:** -3 table lookups, one CLEAR call spared

So your code optimizes the 7-bit CSI format without intermediates. It does not work with 8-bit CSI or with intermediates, which ok, as those are very rare so even the penalty is no biggie.

_Considerations for further action:_
- PRINT data is typically the overwhelming data type, often in longer batches
- CSI (esp. SGR) is the second most data, 99% with 7-bit `ESC ]` opener and no intermediates

Taking both facts into account I suggest another way for fast-pathing CSI - it actually should be a condition at the top of the codepoint loop and doing CSI in a hot inner loop there. It would penalize the first byte of PRINT data slightly, but make not a big difference, as PRINT is optimized internally already. Same for payload intensive OSC and DCS data.
Additionally EXE bytes smaller than 0x18 should get a similar optimization before the CSI fast path condition to not suffer from it too much.

homanp:
> @homanp Had a closer look at the changes. Here are my observations:
> 
> 
> 
> - The main speedup results from collapsing:
> 
>   - `table lookup:ESC --> ParserAction.CLEAR`
> 
>   - `table lookup:0x5b --> ParserAction.CLEAR`
> 
>   into one `table lookup:ESC --> ParserAction.CLEAR --> 0x5b`. So this spares a table lookup and a CLEAR call.
> 
> - params parsing gets not faster much (already optimized in master), but spares the initial table lookup.
> 
> - Final byte matching spares one table lookup.
> 
> - Intermediates are not handled (bails out to normal processing with a tiny penalty).
> 
> - `_csiFast` does handle ParserAction.PRINT at the end, which I don't understand why. Looks misplaced to me, like overoptimizing a certain pattern.
> 
> 
> 
> **Net effect:** -3 table lookups, one CLEAR call spared
> 
> 
> 
> So your code optimizes the 7-bit CSI format without intermediates. It does not work with 8-bit CSI or with intermediates, which ok, as those are very rare so even the penalty is no biggie.
> 
> 
> 
> _Considerations for further action:_
> 
> - PRINT data is typically the overwhelming data type, often in longer batches
> 
> - CSI (esp. SGR) is the second most data, 99% with 7-bit `ESC ]` opener and no intermediates
> 
> 
> 
> Taking both facts into account I suggest another way for fast-pathing CSI - it actually should be a condition at the top of the codepoint loop and doing CSI in a hot inner loop there. It would penalize the first byte of PRINT data slightly, but make not a big difference, as PRINT is optimized internally already. Same for payload intensive OSC and DCS data.
> 
> Additionally EXE bytes smaller than 0x18 should get a similar optimization before the CSI fast path condition to not suffer from it too much.

Hmm, that's interesting. Will do it that way and get back to you

jerch:
@homanp For typical terminal data perf testing you can use this call:
```bash
npm run benchmark out-test/benchmark/Terminal.benchmark.js
```
It benchmarks the `ls -lR` output against the full terminal buffer chain (thus gives a better picture of the raw throughput xterm.js can swallow and contains CSI as well).

jerch:
Not CSI relevant:

Since you are perf testing the parser anyway, could you also test if the following replacement shows a benefit for PRINT data:
```typescript
        case ParserAction.PRINT:
          // read ahead with loop unrolling
          // Note: 0x20 (SP) is included, 0x7F (DEL) is excluded
          let c = i;
          const l4 = length - 4;
          while (c < l4
            && data[++c] >= 0x20 && (data[c] <= 0x7e || data[c] >= NON_ASCII_PRINTABLE)
            && data[++c] >= 0x20 && (data[c] <= 0x7e || data[c] >= NON_ASCII_PRINTABLE)
            && data[++c] >= 0x20 && (data[c] <= 0x7e || data[c] >= NON_ASCII_PRINTABLE)
            && data[++c] >= 0x20 && (data[c] <= 0x7e || data[c] >= NON_ASCII_PRINTABLE)
          ) {}
          if (c >= l4)
            while (c < length && data[c] >= 0x20 && (data[c] <= 0x7e || data[c] >= NON_ASCII_PRINTABLE)) c++;
          this._printHandler(data, i, c);
          i = c - 1;
          break;
```

Imho this should perform better than the current loop unrolling, as it is optimized to trueish eval and spares the individual length checks in the unrolled case.

homanp:
@jerch reworked the approach based on your feedback:

- moved CSI fast-path to top of the codepoint loop as a condition on `code === 0x1b`
- added EXE fast-path for bytes < 0x18 before the CSI check so control bytes don't suffer
- replaced PRINT loop unrolling with your suggested while-based version (trueish eval, spares individual length checks)
- removed the misplaced PRINT scan-ahead from CSI dispatch

parser benchmark:

| benchmark | master | this PR | change |
|---|---|---|---|
| PRINT - a | 824.7 | 843.9 | +2% |
| EXECUTE - \n | 196.1 | 230.2 | +17% |
| ESCAPE - ESC E | 88.1 | 84.6 | -4% |
| ESCAPE w/ collect | 113.2 | 103.3 | -9% |
| **CSI - CSI A** | **189.3** | **296.4** | **+57%** |
| **CSI w/ collect** | **190.4** | **354.3** | **+86%** |
| **CSI params (short)** | **204.2** | **332.9** | **+63%** |
| **CSI params (long)** | **256.0** | **399.4** | **+56%** |
| OSC string (short) | 123.8 | 116.0 | -6% |
| OSC string (long) | 134.0 | 129.5 | -3% |
| OSC class (short) | 157.3 | 142.4 | -9% |
| OSC class (long) | 536.0 | 471.1 | -12% |
| DCS string (short) | 91.9 | 89.2 | -3% |
| DCS string (long) | 125.0 | 123.3 | -1% |
| DCS class (short) | 116.9 | 118.4 | +1% |
| DCS class (long) | 431.5 | 414.4 | -4% |

terminal benchmark:

| benchmark | master | this PR | change |
|---|---|---|---|
| write/string/async | 19.20 | 19.19 | ~0% |
| write/Utf8/async | 20.42 | 21.47 | +5% |

OSC/DCS see small regressions (3-12%) from the two extra branch checks at the top of the loop, as you predicted. real-world throughput is flat-to-+5%. happy to iterate further if needed.

homanp:
Later on if I find the time I would like to profile the Terminal and see if there's anything I can contribute there.

src/common/parser/EscapeSequenceParser.ts
```diff
@@ -702,34 +649,96 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
...
+        else this._executeHandlerFb(code);
+        this.precedingJoinState = 0;
+        continue;
+      }
+
+      // CSI fast-path: collapse ESC [ into a single entry, parse params+final in a tight loop
+      if (code === 0x1b
+        && this.currentState !== ParserState.OSC_STRING
+        && this.currentState !== ParserState.DCS_PASSTHROUGH
+        && this.currentState !== ParserState.APC_STRING
```
jerch: Hmm these extra conditions are nasty, but they are needed to propage the abort properly to the subparsers. Can we shortcut them to one clever range eval? Like smaller than the first of those?

jerch (changes requested):
One obvious perf smell :smile_cat:

src/common/parser/EscapeSequenceParser.ts
homanp: yes, def.

homanp: Done

src/common/parser/EscapeSequenceParser.ts
```diff
@@ -649,34 +649,94 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
     for (let i = start; i < length; ++i) {
       code = data[i];
 
+      // EXE fast-path: common control bytes (0x00-0x17) in non-payload states
+      if (code < 0x18 && this.currentState <= ParserState.CSI_IGNORE) {
+        if (this._executeHandlers[code]) this._executeHandlers[code]();
+        else this._executeHandlerFb(code);
```
jerch, line 659: Does changing this to:
```typescript
(this._executeHandlers[code] ?? this._executeHandlerFb)(code);
```
help? If not, then we prolly should introduce an array lookup for those lower EXE handlers. Currently it is a property lookup on an object. Array index is much faster.

homanp: The ?? trick doesn't work here — ExecuteHandlerType is () => boolean while ExecuteFallbackHandlerType is (ident: number) => void. TS errors with "Expected 0 arguments, but got 1". Let me see if I can find a better way.

homanp: Also adding the array index would be larger change surface, maybe better to add it in a follow-up PR? Wdyt?

src/common/parser/EscapeSequenceParser.ts
```diff
@@ -649,34 +649,94 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
...
+        continue;
+      }
+
+      // CSI fast-path: collapse ESC [ into a single entry, parse params+final in a tight loop
+      if (code === 0x1b
+        && this.currentState < ParserState.OSC_STRING
+        && i + 1 < length && data[i + 1] === 0x5b
+      ) {
+        this._params.reset();
+        this._params.addParam(0); // ZDM
```
jerch, line 670: This still bugs me, two separate calls to a foreign object. Maybe we should replace them with a single method call on `Params.`, e.g.:

```typescript
  /**
   * Resets and adds 0 as first param (ZDM).
   */
  public resetZdm(): void {
    this.length = 1;
    this._subParamsLength = 0;
    this._rejectDigits = false;
    this._rejectSubDigits = false;
    this._digitIsSub = false;
    this._subParamsIdx[0] = 0;
    this.params[0] = 0;
  }
```

src/common/parser/EscapeSequenceParser.ts
jerch: Yes, just leave it as it is. Same for the param remark below.

homanp: Ok, leaving it for a follow-up pr.

src/common/parser/EscapeSequenceParser.ts
homanp: Leaving this for a follow up pr as mentioned above.

jerch:
@homanp Please pull, added brackets to silence the linter.

So how do the final numbers look now? Have we introduced perf regressions somewhere? Is this still on par with your fastest solution from above? (it actually should be a bit faster for CSI, as it spares another table lookup.)

Edit: And I have to check for correctness - not sure yet, if the CSI_IGNORE state is still properly respected.

homanp:
@jerch pulled your linter fix and ran the benchmarks. numbers look solid, CSI is still on par or better than before:

parser benchmark:

| benchmark | master | this PR | change |
|---|---|---|---|
| PRINT - a | 824.7 | 857.3 | +4% |
| EXECUTE - \n | 196.1 | 243.1 | +24% |
| ESCAPE - ESC E | 88.1 | 87.1 | -1% |
| ESCAPE w/ collect | 113.2 | 107.8 | -5% |
| **CSI - CSI A** | **189.3** | **315.0** | **+66%** |
| **CSI w/ collect** | **190.4** | **368.8** | **+94%** |
| **CSI params (short)** | **204.2** | **349.8** | **+71%** |
| **CSI params (long)** | **256.0** | **372.1** | **+45%** |
| OSC string (short) | 123.8 | 105.5 | -15% |
| OSC string (long) | 134.0 | 130.3 | -3% |
| OSC class (short) | 157.3 | 144.4 | -8% |
| OSC class (long) | 536.0 | 518.7 | -3% |
| DCS string (short) | 91.9 | 88.1 | -4% |
| DCS string (long) | 125.0 | 124.9 | ~0% |
| DCS class (short) | 116.9 | 118.3 | +1% |
| DCS class (long) | 431.5 | 411.6 | -5% |

terminal benchmark:

| benchmark | master | this PR | change |
|---|---|---|---|
| write/string/async | 19.20 | 20.03 | +4% |
| write/Utf8/async | 20.42 | 22.58 | +11% |

full-stack throughput is actually up now too, +4% and +11%. no regressions on the real-world path.

jerch:
Thanks for the numbers. I actually see a much bigger jump for PRINT here (+ 15-20%). For real world data I see a similar jump (~8%).

Will check, if the test cases cover the CSI_IGNORE states correctly, otherwise this looks good to go.

homanp:
> Thanks for the numbers. I actually see a much bigger jump for PRINT here (+ 15-20%). For real world data I see a similar jump (~8%).
> 
> Will check, if the test cases cover the CSI_IGNORE states correctly, otherwise this looks good to go.

Awesome, appreciate the help getting this through.

src/common/parser/EscapeSequenceParser.ts
```diff
@@ -649,34 +649,97 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
...
+            this.currentState = ParserState.GROUND;
+            csiDone = true;
+            break;
+          } else if (ch >= 0x20 && ch <= 0x2f) {
+            this._collect <<= 8;
+            this._collect |= ch;
+            i = k;
+            this.currentState = ParserState.CSI_INTERMEDIATE;
+            csiDone = true;
+            break;
```
jerch: I think we should not try to handle intermediates at all here. Once an intermediate appeared, digits are not allowed anymore. The loop doesn't cover this right now (and the test cases don't test for that faulty state atm).

Solution: Remove intermediate handling completely, let it bail out to table lookup handling. This would indeed penalize sequences with intermediates, but they are very rare again.

Example that should not work, but would currently be parsed successfully (untested):
```text
\x1b[?1;2;3!5z  # the '5' is misplaced, sequence must abort
```

src/common/parser/EscapeSequenceParser.ts
```diff
@@ -649,34 +649,97 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
...
+        this._params.reset();
+        this._params.addParam(0); // ZDM
+        this._collect = 0;
+        let k = i + 2;
+        if (k >= length) { i = k - 1; this.currentState = ParserState.CSI_ENTRY; continue; }
+        let ch = data[k];
+        if (ch >= 0x3c && ch <= 0x3f) {
+          this._collect = ch;
+          if (++k >= length) { i = k - 1; this.currentState = ParserState.CSI_PARAM; continue; }
+        }
```
jerch: Can we get this slicker without the additional length checks by checking length once deep enough and bail out early?

Reasoning: A sequence broken over multiple chunks is very rare, most apps take care to create them in one data chunk. Getting rid of the conditions should speed up things while not showing negative impact.

jerch (changes requested):
Have a few more comments, please see below.

src/common/parser/EscapeSequenceParser.ts
homanp: Removed

src/common/parser/EscapeSequenceParser.ts
homanp: Tried cleaning this up.

src/common/parser/EscapeSequenceParser.ts
jerch: I misread earlier, it was formally correct due to the final break. But I think we should keep the fast path slick and tight, so this looks better.

src/common/parser/EscapeSequenceParser.ts
jerch: Yes better, and should give another tiny boost on the fast path.

homanp: Yep

jerch (approved):
Nice, looks clean and concise now. 👍 

Please add a remark at the method docs, that EXE < 0x18 and CSI 7-bit w'o intermediates have a fast path.

homanp:
> Nice, looks clean and concise now. 👍
> 
> Please add a remark at the method docs, that EXE < 0x18 and CSI 7-bit w'o intermediates have a fast path.

Added

jerch:
How do the final numbers look now? :smiley:

homanp:
This is what I get.

parser benchmark:

| benchmark | master | this PR | change |
|---|---|---|---|
| PRINT - a | 824.7 | 848.5 | +3% |
| EXECUTE - \n | 196.1 | 241.4 | +23% |
| ESCAPE - ESC E | 88.1 | 87.0 | -1% |
| ESCAPE w/ collect | 113.2 | 100.4 | -11% |
| **CSI - CSI A** | **189.3** | **328.4** | **+73%** |
| **CSI w/ collect** | **190.4** | **363.7** | **+91%** |
| **CSI params (short)** | **204.2** | **357.2** | **+75%** |
| **CSI params (long)** | **256.0** | **398.8** | **+56%** |
| OSC string (short) | 123.8 | 114.8 | -7% |
| OSC string (long) | 134.0 | 131.7 | -2% |
| OSC class (short) | 157.3 | 145.3 | -8% |
| OSC class (long) | 536.0 | 516.6 | -4% |
| DCS string (short) | 91.9 | 88.1 | -4% |
| DCS string (long) | 125.0 | 119.0 | -5% |
| DCS class (short) | 116.9 | 117.8 | +1% |
| DCS class (long) | 431.5 | 418.5 | -3% |

terminal benchmark:

| benchmark | master | this PR | change |
|---|---|---|---|
| write/string/async | 19.20 | 27.54 | +43% |
| write/Utf8/async | 20.42 | 28.90 | +42% |

jerch:
@homanp Ok this is incredible - thanks a lot for looking into this. :heart: 

Wanna look into the two other possible perfs smells as well?

homanp:
This is some 🔥 perf work.

homanp:
> @homanp Ok this is incredible - thanks a lot for looking into this. ❤️
> 
> Wanna look into the two other possible perfs smells as well?

yes, I will do follow up PRs for those 2. just to recap: 

- Array-indexed execute handlers
- `resetZdm()` on Params

jerch:
> yes, I will do follow up PRs for those 2. just to recap:
> 
>     * Array-indexed execute handlers
>     * `resetZdm()` on Params

Yes, exactly. The method for params should be straight forward, the array-based execute handlers could just be a second container to be filled from the setters leaving the existing container untouched.

homanp:
> > yes, I will do follow up PRs for those 2. just to recap:
> > ```
> > * Array-indexed execute handlers
> > * `resetZdm()` on Params
> > ```
> 
> Yes, exactly. The method for params should be straight forward, the array-based execute handlers could just be a second container to be filled from the setters leaving the exiting container untouched.

will do 2 separate PRs.

jerch:
@homanp Wanna also check, if loop unrolling helps in ParserAction.OSC_PUT, ParserAction.DCS_PUT and ParserAction.APC_PUT?

homanp:
> The method for params should be straight forward, the array-based execute handlers could just be a second container to be filled from the setters leaving the existing container untouched.

sure, will follow up.

PR #5829, opened 2026-04-23
refactor: add Params.resetZdm() to avoid double call on foreign object

homanp:
Follow-up from #5825, collapses the `reset()` + `addParam(0)` pattern into a single `resetZdm()` method on `Params`.

The parser calls this pair 10 times across different code paths. Having it as one method avoids two separate calls to a foreign object per sequence reset.

No perf change expected (confirmed with benchmarks), just a cleanup.


homanp:
@jerch as discussed

jerch (approved):
LGTM 👍 , thanks for the cleanup work.

PR #5830, opened 2026-04-23
perf: array-indexed execute handler lookup in hot path

homanp:
Follow-up from #5825 
The EXE fast-path uses `_executeHandlers[code]` which is a property lookup on an object. since the codes are small integers (0x00-0x1f), an array index is much faster.

Adds a parallel array (`_executeHandlersFast`) alongside the existing object container. The setters populate both, the hot path reads from the array. existing API is unchanged.

No measurable regression on any benchmark path.

### DIFF
--- a/src/common/parser/EscapeSequenceParser.ts
+++ b/src/common/parser/EscapeSequenceParser.ts
@@ -247,6 +247,7 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
   // handler lookup containers
   protected _printHandler: PrintHandlerType;
   protected _executeHandlers: { [flag: number]: ExecuteHandlerType };
+  protected _executeHandlersArr: (ExecuteHandlerType | undefined)[];
   protected _csiHandlers: IHandlerCollection<CsiHandlerType>;
   protected _escHandlers: IHandlerCollection<EscHandlerType>;
   protected readonly _oscParser: IOscParser;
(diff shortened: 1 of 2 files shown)


src/common/parser/EscapeSequenceParser.ts
```diff
@@ -651,7 +662,8 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
 
       // EXE fast-path: common control bytes (0x00-0x17) in non-payload states
       if (code < 0x18 && this.currentState <= ParserState.CSI_IGNORE) {
-        if (this._executeHandlers[code]) this._executeHandlers[code]();
+        const h = this._executeHandlersArr[code];
+        if (h) h();
         else this._executeHandlerFb(code);
```
jerch: I think the pattern
```typescript
(this._executeHandlersArr[code] ?? this._executeHandlerFb)(code);
```
would be faster here. Could you test that? You might have to change the interface for that like:
```typescript
export type ExecuteHandlerType = (indent?: number) => boolean;
export type ExecuteFallbackHandlerType = (ident: number) => void;
```
(This should be safe, as we currently don't expose EXE handlers on the API.)

homanp: agree

homanp: Done, EXECUTE throughput went from 244 → 346 MB/s with the ?? pattern

src/common/parser/EscapeSequenceParser.ts
```diff
@@ -290,11 +291,13 @@ export class EscapeSequenceParser extends Disposable implements IEscapeSequenceP
     this._errorHandlerFb = (state: IParsingState): IParsingState => state;
     this._printHandler = this._printHandlerFb;
     this._executeHandlers = Object.create(null);
+    this._executeHandlersArr = new Array<ExecuteHandlerType | undefined>(0x20).fill(undefined);
```
jerch: Only up to 0x17 (inclusive) seems to be safe. Please adjust the 0x20 here and follow-up to 0x18 (exclusive), so no misinterpretations result from the slightly off length here.

Edit:
I think the additional typing hint is not needed, for better readability imho this should be sufficient here:
```typescript
this._executeHandlersArr = new Array(0x18).fill(undefined);
```

jerch (changes requested):
The chosen length of 0x20 seems off, see below.

src/common/parser/EscapeSequenceParser.ts
homanp: Right, updated

jerch (approved):
Thanks 👍 Final speedup compared to old handling (before fast path) is ~1.8x for EXE bytes. It won't change much in real-world data, as EXE bytes are not that often and the real terminal handler for NL is actually pretty expensive.

homanp:
> Thanks 👍 Final speedup compared to old handling (before fast path) is ~1.8x.

Brr 🚀

Issue #617, opened 2017-03-23
How to get input commands


smalltimoo:
## Details
- Browser and browser version: 
- OS version: 
- xterm.js version: 

### Steps to reproduce

1. How to get input commands
For example: I enter a command in xterm.js, how do I get this command
2.

akalipetis:
Hello @smalltimoo - could you please explain a bit more what do you mean? How do you enter an input command?

If, for example, you run a command in your terminal that requires user input, this will just work, there's no action you need to take.

Tyriar:
Unfortunately this is not as easy as it would seem, as the input is passed to the shell and there's no way to differentiate whether it's a command or input. There are some techniques that can be used such as wrapping the shell process or implementing something like iTerm2's shell integration https://github.com/sourcelair/xterm.js/issues/576. I'm closing this as a duplicate of that since that seems like the best way forward at the moment.

smalltimoo:
I entered a command on terminal, and I want to get this command when I press the Enter key

Tyriar:
@smalltimoo you can try listening to the key events yourself, otherwise https://github.com/sourcelair/xterm.js/issues/617#issuecomment-288849502 explains the current situation.

smalltimoo:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

PR #5831, opened 2026-04-23
Clear IntersectionObserver reference on dispose

### LINKED ISSUE #5820: IntersectionObserver callback retains RenderService → BufferLines after dispose

srid:
## Summary

Fixes the IntersectionObserver retention path from #5820 without using `WeakRef`.

`RenderService._registerIntersectionObserver` currently creates an observer whose callback closes over `this`. The disposable disconnects the observer, but the disposable closure also keeps a JS reference to the observer. Since the observer owns its callback, the disposed `RenderService` can remain reachable through:

`RenderService -> MutableDisposable -> disposable closure -> IntersectionObserver -> callback -> RenderService`

Store the observer on `RenderService` instead, and have the disposable disconnect and null out that field. The disposable no longer captures the observer object, so after dispose there is no JS-land reference from the disposed service back to the observer/callback pair.

This is an alternative to #5821's `WeakRef` version.

## Evidence

Kolu reproduced the original leak on the pre-#617 UI and compared three arms in Chrome DevTools MCP. Each arm used a fresh `just dev`, 7 terminals created/restored through the UI, and 30 Focus<->Canvas round trips via the visible header toggle.

| Arm | xterm variant | xterm buffer retention delta / 30 toggles |
|---|---|---:|
| A | unpatched baseline, pre-#5821 | `Uint32Array +339`, `JSArrayBufferData +339` |
| B | #5821 WeakRef fix | `Uint32Array +0`, `JSArrayBufferData +0` |
| C | this observer-nullify variant | `Uint32Array +0`, `JSArrayBufferData +0` |

The baseline reproduced the retained xterm-buffer signature. Both WeakRef and this observer-nullify variant retained zero additional xterm buffer arrays.

Caveat: this MCP run did not reproduce the original +367 MB magnitude because the clean UI-created/restored terminals had only ~118 scrollback rows instead of the large restored buffers from the production repro. The retention signature still distinguishes the baseline from both fixes.

## Test

- `npm run lint-changes`

Closes #5820

### DIFF
--- a/src/browser/services/RenderService.ts
+++ b/src/browser/services/RenderService.ts
@@ -29,6 +29,7 @@ export class RenderService extends Disposable implements IRenderService {
   private _renderDebouncer: IRenderDebouncerWithCallback;
   private _pausedResizeTask: DebouncedIdleTask;
   private _observerDisposable = this._register(new MutableDisposable());
+  private _intersectionObserver: IntersectionObserver | undefined;
 
   private _isPaused: boolean = false;
   private _needsFullRefresh: boolean = false;
@@ -127,8 +128,12 @@ export class RenderService extends Disposable implements IRenderService {
     // and resume based on terminal visibility if so
     if ('IntersectionObserver' in w) {
       const observer = new w.IntersectionObserver(e => this._handleIntersectionChange(e[e.length - 1]), { threshold: 0 });
+      this._observerDisposable.value = toDisposable(() => {
+        this._intersectionObserver?.disconnect();
+        this._intersectionObserver = undefined;
+      });
+      this._intersectionObserver = observer;
       observer.observe(screenElement);
-      this._observerDisposable.value = toDisposable(() => observer.disconnect());
     }
   }
 


Tyriar (approved):
Thanks!

Issue #5845, opened 2026-04-26
image addon serialization


jerch:
Currently images printed to the terminal cannot be serialized. Previous tests regarding image serialization were done in https://github.com/jerch/xterm-addon-image/pull/48. They revealed the following cornerstones:

- add QOI image format for performance reasons
- implement an image sequence format suitable for serialization
  - separate data load from placements
  - possible to shard data parts
-  extend serialize addon with at least row-level events

**Steps to be taken**
As a possible cause of actions I suggest the following steps:
- [ ] add QOI decoder to IIP (maybe kitty too?) --> #5846
- [ ] implement serialization image sequence
  - [ ] add QOI encoder
  - [ ] learn from kitty protocol about sharding and placements
  - [ ] spec internal sequence with:
    - [ ] data load
    - [ ] placements
- [ ] extend serialization
  - [ ] add events to serialize addon
  - [ ] expose events on serialize addon API
- [ ] register image serialization to serialize events

The reason for using QOI and not one of the standard formats is mostly performance. QOI yields comparable file sizes to PNG, but is in previous tests ~4 times faster to process.

For the custom sequence we can take bits from the kitty protocol, or extend it for encoding purposes with QOI. The tricky part will be to harmonize it with SIXEL and IIP states though. Currently I don't aim to write a full serialization for all protocol types, as this seems overbloated. Instead we should extract the minimal needed bits to restore images states correctly with one sequence and image format only.

Up for discussion first, as it will introduce quite some changes across the codebase.
FYI: @Tyriar, @anthonykim1, @PerBothner

Tyriar:
> expose events on serialize addon API

Interested in what you had in mind here with the events, have a small code snippet to communicate that?

jerch:
In https://github.com/jerch/xterm-addon-image/pull/48 I created a stub serialize (text-only for demo purpose, see ImageAddon.ts:L348-L361):

```typescript
  public serialize(start: number, end: number): string[] {
    const st = Date.now();
    const lines: string[] = [];
    const buffer = this._terminal!._core.buffer;
    for (let row = start; row < end; ++row) {
      const line = buffer.lines.get(row);
      if (!line) break;
      // FIXME: hook into serialize addon instead of translateToString
      // lines.push(line.translateToString(true) + this.serializeLinePng(row));
      lines.push(line.translateToString(true) + this.serializeLineQoi(row));
    }
    console.log('duration', Date.now()-st);
    return lines;
  }
```

[...]

PR #5846, opened 2026-04-26
QOI support for IIP

jerch:
Adds QOI support to the image addon.

Part of #5845.

Some benchmarks with https://github.com/jerch/sixel-bench (all my machine™ with Chromium):

| Format | Protocol | TP in MB/s | FPS  | VSCode MB/s | VSCode FPS | VSCode Penalty |
|---|---|---|---|---|---|---|
| Sixel | Sixel |39.46 | 227.92  | 5.8 | 33.48 | 6.8x |
| PNG  | IIP  | 21.77  | 104.48  | 4.7 | 22.58 | 4.6x |
| JPEG  | IIP  | 4.71  | 200.04  | 3.0 | 127.44 | 1.6x |
| QOI  | IIP  | **43.78**  | **229.31**  | - | - | - |
| PNG  | Kitty  | 17.93  | 86.09  | 4.3 | 20.66 | 4.2x |

Observation: QOI leads in throughput and FPS. Sixel is close behind but unusable for quick encoding and due its lower quality. The kitty implementation with PNG is worse than for IIP, prolly not as heavily optimized yet.

On a sidenote: Webkit/Safari is actually faster than Chromium yielding 2.82s (46.45 MB/s, 243.31 fps) for QOI. For Kitty it performs much worse with ~12 MB/s.

_Edit: Added vscode numbers for comparison to xterm.js demo._

### DIFF
--- a/addons/addon-image/package.json
+++ b/addons/addon-image/package.json
@@ -25,6 +25,6 @@
   },
   "devDependencies": {
     "sixel": "^0.16.0",
-    "xterm-wasm-parts": "^0.3.0"
+    "xterm-wasm-parts": "^0.4.1"
   }
 }
--- a/addons/addon-image/src/IIPHandler.ts
+++ b/addons/addon-image/src/IIPHandler.ts
@@ -7,6 +7,7 @@ import { ImageRenderer } from './ImageRenderer';
 import { IIPImageStorage } from './IIPImageStorage';
 import { CELL_SIZE_DEFAULT } from './ImageStorage';
 import Base64Decoder from 'xterm-wasm-parts/lib/base64/Base64Decoder.wasm';
+import QoiDecoder from 'xterm-wasm-parts/lib/qoi/QoiDecoder.wasm';
 import { HeaderParser, IHeaderFields, HeaderState } from './IIPHeaderParser';
 import { imageType, UNSUPPORTED_TYPE } from './IIPMetrics';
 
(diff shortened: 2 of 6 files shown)


jerch:
@Tyriar, @anthonykim1 This also removes the Image-load shim around missing `createImageBitmap` in older Safari (<16.4, fixed in Safari since March 27, 2023).

jerch:
@Tyriar, @anthonykim1 Added numbers for vscode above in the table. Something in vscode terminal is sucking hard on the effective throughput, especially when the input is big (thus JPEG is less affected).

To repro yourself in vscode terminal:
```bash
git clone git@github.com:jerch/sixel-bench.git
cd sixel-bench
cat parts_* > blobs.tar.gz
tar -zxf blobs.tar.gz
reset
python run.py data/data.sixel
reset
python run.py data/data.iip.png
```
The `reset` is needed for the speed result numbers to show up. I don't know why it doesn't work without, seem the terminal tries to be smart about the prompt line?

Tyriar (approved):

Tyriar:
> Something in vscode terminal is sucking hard on the effective throughput

Quite a lot happens in VS Code on data events, the pty setup is much more involved than the demo is probably also a contributor to this. Btw I left Microsoft recently if you didn't see, still working on xterm.js (and vscode) as needed, time's a bit tight as I ramp up though 🙂

Issue #5849, opened 2026-04-28
RG colorbits not cleared in AttributeData


jerch:
In the inputhandler colorbits get only cleared for the range to be overwritten:

https://github.com/xtermjs/xterm.js/blob/3a84bcb7ae5a3471a1810db621010a46dba2085f/src/common/InputHandler.ts#L2629-L2632

Here the `PCOLOR_MASK` only clears the lower 8 bits, as only those will be significant for `CM_P16`.

**Issue:**
The char atlas caches values from the whole fg & bg int32 values:
https://github.com/xtermjs/xterm.js/blob/3a84bcb7ae5a3471a1810db621010a46dba2085f/addons/addon-webgl/src/TextureAtlas.ts#L278-L282

thus will create doubled cache entries after RGB colors, if RG bits got set by the RGB colors.

**Fix:**
Instead of `PCOLOR_MASK` we should use `RGB_MASK` in inputhandler (and other places, where char attribs are set).

PR #5850, opened 2026-04-28
Fix RG colorbits in AttributeData

jerch:
See #5849.


jerch:
@Tyriar Could it be, that the CI does not run the addon tests anymore? For the individual addons the CI seems to repeat the whole standard playwright battery (is that really needed?), but not tests defined under the addons itself.

PR #5859, opened 2026-04-29
IIP enhancements

### LINKED ISSUE #5858: image addon: make IIP more streamline
With the recently introduced FilePart in IIP our implementation works only with the legacy switch on `imgcat`:

```bash
imgcat -l addons/addon-image/fixture/palette.png
```

The chunking in several sub sequences was introduced to overcome a limitation in terminal multiplexer like `tmux`. We should extend the IIP protocol to also support the sub sequence chunking.

Edit:
There are various changes needed for IIP on top of that, see https://github.com/jerch/xterm-addon-image/issues/57 and https://github.com/jerch/xterm-addon-image/issues/58.

Related:
- #4981
- https://github.com/jerch/xterm-addon-image/issues

jerch:
Fixes #5858.

TODO:
- [x] IIPHeaderParser:
  - [x] MultipartFile
  - [x] FilePart
  - [x] ReportCellSize
  - [x] parser tests
- [x] IIPHandler:
  - [x] multipart handling
  - [x] remove strict size param handling
  - [x] ReportCellSize
- [x] cursor placement
- [x] ~keep original, move resize to storage/renderer~ postponed: #5861
- [x] integration tests
- [x] FilePart test in demo
- [x] change docs


Tyriar (approved):

addons/addon-image/src/kitty/KittyImageStorage.ts
```diff
@@ -42,6 +42,7 @@ export class KittyImageStorage implements IDisposable {
       this._images.delete(kittyId);
     }
   };
+  private _addImageOpts: IAddImageOpts = { scrolling: true, layer: 'top', zIndex: 0, cursorPos: 'iip' };
```
anthonykim1, line 45: Thank sm for the PR🙏

I believe for Kitty we override cursor position: https://github.com/xtermjs/xterm.js/blob/43e8365f0e20402ce9744e3770a97dd7a78ef877/addons/addon-image/src/kitty/KittyGraphicsHandler.ts#L666-L681 

@jerch Should we have `CursorPosition` optional?
https://github.com/xtermjs/xterm.js/blob/43e8365f0e20402ce9744e3770a97dd7a78ef877/addons/addon-image/src/Types.ts#L105-L110

intention would be to let protocol handle its own cursor logic

jerch: @anthonykim1 Sorry missed your comment.

I would not make it optional on the interface to avoid object shape shifting (will often lead to bail out of the code optimizer). Can you explain at which level you need custom cursor handling - on handler side or in the kitty storage impl?
I see two possible ways here - either give kitty its own `CursorPos` type or we remove cursor positioning completely from `addImage`, thus do it in every impl overload after calling `addImage`.

PR #5868, opened 2026-05-03
fix(addon-clipboard): honor all OSC 52 selection types in BrowserClipboardProvider

aaradhyaberisfdc:
## Problem

`BrowserClipboardProvider` silently drops OSC 52 reads/writes for any selection other than `c` (system clipboard). tmux sends selection `p` (primary) by default with `set-clipboard on`, so clipboard sync never reaches the browser - and the same goes for the empty-selection form `\e]52;;…\a`.

Relevant code before this change: [ClipboardAddon.ts#L72-L78](https://github.com/xtermjs/xterm.js/blob/master/addons/addon-clipboard/src/ClipboardAddon.ts#L72-L78).

## Fix

Browsers expose a single clipboard with no primary/secondary distinction, so the default provider should route every selection type to `navigator.clipboard`. Removed the `selection !== 'c'` guards in `readText` / `writeText`. Consumers that need per-selection behavior can still supply a custom `IClipboardProvider`.

Also added a clarifying JSDoc note on `ClipboardSelectionType`. The enum itself is unchanged to keep this PR focused.

## Tests

Added playwright coverage for previously-silent cases:
- write `\e]52;p;…\a` (primary)
- write `\e]52;;…\a` (empty / default)
- read `\e]52;p;?\a`

## Context

Surfaced downstream in ttyd (tsl0922/ttyd#1518), where tmux's `set-clipboard on` didn't reach the browser. Fixing it upstream means ttyd and similar consumers don't need to ship a custom `BrowserClipboardProvider` shim.

Verified manually with tmux `set -g set-clipboard on` + `set -g mouse on`: mouse-drag in copy-mode and vi `y` now copy to the system clipboard.


jerch:
@aaradhyaberisfdc This needs to be tested against xterm and the OSC 52 spec first. We cannot simply change sequence related things in a handwaving manner without checking against the spec.

aaradhyaberisfdc:
Thanks @jerch - fair callout. Grounding this in the spec:

Per [xterm ctlseqs](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html) `Ps=52`:
```
Ps = 5 2  ⇒  Manipulate Selection Data.  These controls may
          be disabled using the [allowWindowOps](https://invisible-island.net/xterm/manpage/xterm.html#VT100-Widget-Resources:allowWindowOps) resource.  The parameter
          Pt is parsed as
               Pc ; Pd

          The first, Pc, may contain zero or more characters from the
          set c , p , q , s , 0 , 1 , 2 , 3 , 4 , 5 , 6 , and 7 .  It is
          used to construct a list of selection parameters for
          clipboard, primary, secondary, select, or cut-buffers 0
          through 7 respectively, in the order given.  If the parameter
          is empty, xterm uses s 0 , to specify the configurable
          primary/clipboard selection and cut-buffer 0.

          The second parameter, Pd, gives the selection data.  Normally
          this is a string encoded in base64 (RFC-4648).  The data
          becomes the new selection, which is then available for pasting
          by other applications.

          If the second parameter is a ? , xterm replies to the host
          with the selection data encoded using the same protocol.  It
          uses the first selection found by asking successively for each
          item from the list of selection parameters.

          If the second parameter is neither a base64 string nor ? ,
          then the selection is cleared.
 ```

### master vs. this PR

| `Pc` | master | this PR | spec |
|---|---|---|---|
| `c` | clipboard | clipboard | clipboard |
| `p`, `q`, `s`, `0`–`7` | **dropped** | clipboard\* | primary / secondary / select / cut-buffers |
| *(empty)* | **dropped** | clipboard\* | `s0` |
| `cp`, `ps0`, … | **dropped** | clipboard\* | walk list, first hit |

\* A browser exposes one clipboard (`navigator.clipboard`) - no primary/secondary/cut-buffer surface. For the *default browser provider*, every spec-valid `Pc` collapses to that one backend. Custom `IClipboardProvider` impls can still route per-selection.

This PR only fixes `BrowserClipboardProvider`'s drop-on-non-`c` behaviour.

jerch:
@aaradhyaberisfdc Thx for the spec. I gonna integrate the PR into #5874.

PR #5874, opened 2026-05-10
clipboard addon: various fixes

jerch:
This is mostly a security update to remove an outer dependency of the clipboard addon. Also incorporates #5868.

- [x] replace base64 actions with atob/atob (also in NodeJS since v16)
- [x] remove outer dependency js-base64
- [x] adjust tests (allow Firefox in CI)
- [x] always refer to navigator.clipboard in default provider
- [x] remove `ClipboardSelectionType` as it is not sufficient for xterm's selection param (**API change**)


Tyriar (approved):
No deps 🎉

PR #5875, opened 2026-05-12
Fix keys getting swallowed when enabling only "Report event types"

zeertzjq:
Per kovidgoyal/kitty#9999, enabling "Report event types" without
"Disambiguate escape codes" doesn't really make sense, and terminals may
choose whatever behavior they think makes sense. However, keys shouldn't
get swallowed (which, after #5791, happens when the !useCsiU branch
cannot handle the event), so instead make "Report event types"
imply "Disambiguate escape codes".

### DIFF
--- a/src/common/input/KittyKeyboard.test.ts
+++ b/src/common/input/KittyKeyboard.test.ts
@@ -478,7 +478,7 @@ describe('KittyKeyboard', () => {
         assert.strictEqual(result.key, '\x7f');
       });
 
-      it('press event explicit :1 when modifiers present', () => {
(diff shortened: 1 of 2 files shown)


Tyriar (approved):

jerch:
@zeertzjq So "Report event types" is not 2, but treated as 3 now (as if the 1 bit was set)?

zeertzjq:
Yes. The behavior after this PR is still an improvement over the behavior before #5791, where enabling "Report event types" behaved more (but not completely) like 11 (as if both the 1 bit and the 8 bit are set).

Issue #5847, opened 2026-04-27
[webgl] Partial row ghosting with transparent theme background on WKWebView / Tauri (stable macOS)


longshuaige:
## Summary

On a Tauri (WKWebView) app using `@xterm/addon-webgl` with `allowTransparency: true` and a semi-transparent `theme.background`, lines with **ANSI background colors** (diff rows, cursor row, selection) progressively develop **partial ghosting / overlap** during streaming output such as Claude Code / git diff / tree. A window resize temporarily clears it.

This appears **related to but distinct from #5816** — that one is the fully broken rendering on macOS 26.5 beta Safari. Mine reproduces on **stable macOS** inside **Tauri (WKWebView)** only, without needing any beta OS, and the visual symptom is partial rather than complete corruption. Filing separately per the pattern suggested in discussion.

## Environment

- `@xterm/xterm` `6.0` (latest)
- `@xterm/addon-webgl` `^0.19.0` (latest)
- Tauri (WKWebView) on macOS stable (not 26.5 beta)
- MacBook Pro Retina display, `devicePixelRatio = 2`
- `allowTransparency: true`, `theme.background: "rgba(20, 20, 25, 0.25)"` (or similar alpha < 1)
- Custom wallpaper visible behind the terminal

## Visual symptom

[screenshot: renderErrIng1.png]

- Upper half of the screen (diff lines with red/green ANSI backgrounds): severe character overlap / ghosting
- Lower half (plain text lines): perfectly clean
- The ghosting is **localized to lines that have a non-null cell background color** — which is the key asymmetry that rules out "every frame is being overlaid globally"

## Steps to reproduce

1. Inside a Tauri WKWebView window on stable macOS, instantiate an xterm Terminal with:
   ```ts
   new Terminal({
     allowTransparency: true,
     theme: { background: 'rgba(20,20,25,0.25)', foreground: '#fff' },
   });
   term.loadAddon(new WebglAddon());


<img width="1080" height="794" alt="Image" src="https://github.com/user-attachments/assets/bf38ff6a-0647-458e-aad4-de3b35d0ed34" />

<img width="1009" height="445" alt="Image" src="https://github.com/user-attachments/assets/11d26758-237e-40b3-89b2-548be55b1f33" />

jerch:
@longshuaige Does this happen right from the start (like running ls -lR /usr) or after a while of pumping a lot of content to the screen?

longshuaige:
[...]

PR #5883, opened 2026-05-17
Fix webgl rendering corruption from atlas page merges

vsych:
Two bugs in the webgl renderer cause the corruption from #5847.

**1. Texture binding stale after a same-index page swap.** `GlyphRenderer.render` only rebinds a texture unit on `version` mismatch, but after `_createNewPage` merges 4 pages into 1, the page object at the merged index swaps to a fresh `mergedPage`. With per-page counters the new page's `version` often coincided with the version we last bound at that index (the freshly-pushed mergedPage always landed on `1`). The rebind was skipped and the texture kept the old page's canvas, so glyphs sampling that unit got garbage.

Make `AtlasPage.version` monotonic across all pages and the existing comparison detects a same-index swap, no extra bookkeeping required.

**2. Stale `texturePage` in the vertex buffer after a mid-update merge.** When a merge fires inside `_updateModel`, cached glyphs get a new `texturePage`. Cells already written in this pass (or carried over via the model-unchanged early-exit) still point at the pre-merge index. `_requestClearModel` was set on merge but `beginFrame` never reset it, so it stayed true forever; and `renderRows` only checked it before `_updateModel`, so a mid-pass merge went unhandled.

- reset `_requestClearModel` when `beginFrame` reads it
- after `_updateModel`, re-run a full update if a merge fired during the pass (capped at `MERGE_RETRY_LIMIT`)

Before:
issue
After:
image

Stress repro with the script from #5847 (full-screen rewrite, many colors, frequent merges):

Before this PR:

repro-master-no-fix

After this PR:

repro-with-fix

### DIFF
--- a/addons/addon-webgl/src/GlyphRenderer.ts
+++ b/addons/addon-webgl/src/GlyphRenderer.ts
@@ -373,6 +373,10 @@ export class GlyphRenderer extends Disposable {
 
   public setAtlas(atlas: ITextureAtlas): void {
     this._atlas = atlas;
+    this.invalidateAtlasTextures();
+  }
+
+  public invalidateAtlasTextures(): void {
     for (const glTexture of this._atlasTextures) {
       glTexture.version = -1;
     }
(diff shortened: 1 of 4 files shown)


Tyriar (approved):
Legend 👏 I can repro the problem on master and not on your branch.

This has been an issue for a long time, I think it's popped up more recently due to agentic CLIs using a much wider range of character styles.

Kitenite:
This is awesome!

PR #5885, opened 2026-05-20
fix: remove aria-hidden from DOM renderer row container

### LINKED ISSUE #5881: aria-hidden on row container makes DOM renderer output inaccessible to screen readers
## Summary

The DOM renderer sets `aria-hidden="true"` on the row container, making all terminal output invisible to screen readers.

## Location

`browser/renderer/dom/DomRenderer.ts`, constructor:

```typescript
this._rowContainer.setAttribute('aria-hidden', 'true');
```

## Reproduction

Reproduced in a minimal xterm.js app with no dependencies beyond the library itself. The DOM is well-structured — output is rendered as readable text inside `div` and `span` elements — but the `aria-hidden` attribute on `.xterm-rows` makes it completely invisible to screen readers.

Steps:
1. Create a minimal xterm.js page (DOM renderer is used by default)
2. Enable a screen reader (tested with NVDA on Windows)
3. Type any command and press Enter
4. Terminal output is not announced and cannot be read

## Impact

Screen reader users have no access to terminal output. The `xterm-helper-textarea` handles input accessibly, but there is no equivalent for output, leaving the implementation half-accessible.

This affects any application embedding xterm.js with the DOM renderer, including OpenShift Web Terminal.

## Workaround

Removing the attribute makes the content readable with screen readers without any visible side effects:

```javascript
document.querySelector('.xterm-rows').removeAttribute('aria-hidden');
```

## Proposed fix

Remove the `aria-hidden` attribute from `_rowContainer` in `DomRenderer.ts`.

jpavonabian:
## Summary

- Removes `aria-hidden="true"` from `_rowContainer` (`.xterm-rows`) in `DomRenderer.ts`
- The DOM renderer renders actual text in the DOM as divs/spans — screen readers can read it directly once the attribute is removed
- `_selectionContainer` keeps its `aria-hidden="true"` (correct — it is a visual-only overlay)
- The `screenReaderMode` option and `AccessibilityManager` are unchanged and continue to provide an enhanced experience (live region, structured accessible tree) when explicitly enabled

## Motivation

`aria-hidden="true"` was unconditionally set on the row container, making all terminal output completely invisible to screen readers when `screenReaderMode` is `false` (the default). Since `AccessibilityManager` only activates when `screenReaderMode: true`, the default configuration left screen reader users with zero access to terminal output.

Accessibility should not depend on the embedding developer knowing to opt in via `screenReaderMode`.

Closes #5881

### DIFF
--- a/src/browser/renderer/dom/DomRenderer.ts
+++ b/src/browser/renderer/dom/DomRenderer.ts
@@ -80,7 +80,6 @@ export class DomRenderer extends Disposable implements IRenderer {
     this._rowContainer = this._document.createElement('div');
     this._rowContainer.classList.add(Constants.ROW_CONTAINER_CLASS);
     this._rowContainer.style.lineHeight = 'normal';
-    this._rowContainer.setAttribute('aria-hidden', 'true');
     this._refreshRowElements(this._bufferService.cols, this._bufferService.rows);
     this._selectionContainer = this._document.createElement('div');
     this._selectionContainer.classList.add(Constants.SELECTION_CLASS);


Tyriar:
It's there for a reason. The list is virtualized and as a result it's simply not accessible without enabling `screenReaderMode`

jpavonabian:
> It's there for a reason. The list is virtualized and as a result it's simply not accessible without enabling `screenReaderMode`
But that is a mistake. The accessibility of the component depends on a developer setting a variable. Most applications have the component set to false, which is the default behavior. Removing that aria-hidden at least allows the output text from the terminal to be read.

Tyriar:
Agree to disagree. Ideally it would be the case, but a big goal of the project was high performance. If it's important to you, you should set that option. If possible on your platform you could detect a screen reader and set it intelligently.

Issue #5443, opened 2025-11-17
Search is slow when there are many logs  --Help


bo-zn:
**Expected Behavior:**  
Make search faster

**Actual Behavior:**  
Non-linear performance degradation:  
- 30k lines → 3s delay  
- 50k lines → 8s delay  
- 100k lines → 30s delay  

**Environment:**  
- @xterm : 5.3.0 
- @xterm/addon-search : 0.15.0

Code Example: 
```  js  
import { Terminal } from 'xterm'
import { SearchAddon } from '@xterm/addon-search'

// Component-level variables
let terminal: Terminal | null = null;
let searchAddon: SearchAddon | null = null;

onMounted(() => {
  terminal = new Terminal({ /* options */ ,scrollback: 30000,});
  searchAddon = new SearchAddon();
  terminal.loadAddon(searchAddon);

  // Generate test data (30k lines in this example)
  for (let i = 0; i < 30000; i++) {
    terminal.write(`Log line ${i}\n`);
  }
});

// Search function
function handleSearch() {
  if (!searchAddon) {
    console.error("SearchAddon not initialized!");
    return;
  }
  searchAddon.findNext(20000);
}  
```

jerch:
Plz see #5176 for further details on the slow search.

Issue #5177, opened 2024-10-03
Search is not incremental


Tyriar:
See https://github.com/xtermjs/xterm.js/issues/5176 for context

All searches take the same amount of time when the latter ones should take much less time due to being incremental:

Test file:

[test.txt](https://github.com/user-attachments/files/17243265/test.txt)

![Image](https://github.com/user-attachments/assets/34f484ce-b9e3-4645-98bc-42588238fb1b)

Poonam1216:
Hi! I'd love to work on this issue. I have experience with TypeScript and have worked on performance-sensitive real-time systems. Could you point me to where the search state is currently managed in the codebase? Happy to dig in.

Issue #4911, opened 2023-12-11
Mass decoration registration and disposable is slow


Tyriar:
VS Code issue: https://github.com/microsoft/vscode/issues/199593

Lots of GC happening when registering and disposing many decorations. Here's an example showing and hiding a command guide in VS Code which creates and then disposed ~16k decorations:

![image](https://github.com/xtermjs/xterm.js/assets/2193314/62943f68-4785-426d-b407-8362732fa5db)
![image](https://github.com/xtermjs/xterm.js/assets/2193314/a51182cd-08c0-4907-aae1-8927780f20dd)

https://github.com/Tyriar/xterm.js/blob/1943636f023b0496c2b95aeb31f2d08d7c0e4dce/src/common/services/DecorationService.ts#L41-L60

Some ideas to speed up:

- Don't use `splice`(!) in `SortedList`
  - Doing something more clever here would probably vastly improve registration
- Use an object pool
- Defer clean up to idle callback

jerch:
My guess also goes to the splice needs here, `SortedList` basically implements the insert sort idea, which is nice for insertion/removal at the end with _O(1)_, but soon saturates to _O(n²)_ for average (I guess this happens alot here, when lines drop off the scroll buffer).

Tyriar:
Yes RB or AVL is probably the fix. I've done both before but it looks like I never bothered porting the RB tree from js to ts yet. https://github.com/gwtw/ts-avl-tree, https://github.com/gwtw/js-data-structures/blob/master/lib/red-black-tree.js

> A balanced tree based impl will perform alot worse for typical insertion/removal at the end. Maybe partitioning of the list (like a skiplist) will already do the job here?

It's still just log n though, the main thing that matters is that it scales well imo

tisilent:
I see if there is anything I can do on the demo to facilitate subsequent verification. 😀

jerch:
@Tyriar How would an easy repro look like? Is it enough to spam tons of decorations all over the lines, or do they need to follow a certain pattern?

Which action triggered the toxic removals? Was that just from scrolling off at the top? Because if thats the case, we could prolly lift some burden here by batching the removals with interim "dead keys" on the sorted list and doing a final cleanup copy. Difference would be _O(n log n)_ vs. currently _O(n*m log n)_ for _m_ elements to be removed from _n_ elements (the _m_ multiplier comes from the splice copy overhead for every single removal in our current bisearch&splice approach).

Tyriar:
[...]

Issue #4902, opened 2023-12-07
Search is very slow when there is a lot of content and many wrapped lines


Tyriar:
Repro:

Run the following script in powershell and try search:

```
[...]

PR #5902, opened 2026-05-26
Optimize forEachDecorationAtCell by maintaining cache against buffer line index

### LINKED ISSUE #5176: Search is too slow

Tyriar:
## Summary

Speeds up `DecorationService.forEachDecorationAtCell` and `getDecorationsAtCell` by indexing decorations per logical buffer line (via marker coordinates), so render-time lookup scales with decorations on the queried line instead of the total decoration count. Addresses renderer blocking when search (and similar features) register very large numbers of highlights.

Resolves #5176

Refs #3809, #5443, [microsoft/vscode#230365](https://github.com/microsoft/vscode/issues/230365)

## Problem

`forEachDecorationAtCell` previously iterated every registered decoration on each call. The DOM renderer invokes it per cell during row refresh, so cost is roughly:

`O(viewport cells × total decorations)`

With thousands of search highlights (and higher highlight limits in clients like VS Code), a single viewport refresh can scan the full decoration set tens of thousands of times and noticeably block the main thread. This is the bottleneck called out in [#5176](https://github.com/xtermjs/xterm.js/issues/5176).

Prior approaches (e.g. [#5886](https://github.com/xtermjs/xterm.js/pull/5886)) used a fast path only when no multi-line decorations exist; a single multi-line decoration forced a full scan again.

Related issues this does **not** fully resolve:

- [#5177](https://github.com/xtermjs/xterm.js/issues/5177) — incremental search
- [#4902](https://github.com/xtermjs/xterm.js/issues/4902) — slow search with wrapped lines (scan cost)
- [#4911](https://github.com/xtermjs/xterm.js/issues/4911) — mass decoration registration/disposal (`SortedList` / GC)

### Per-line index (`DecorationLineCache`)

- Exported `DecorationLineCache` in `src/common/services/DecorationService.ts` maintains `Map<logicalLine, decorations[]>`.
- Keys use **`marker.line`** (stable logical buffer coordinates), not `CircularList` ring slots.
- Multi-line decorations (`height > 1`) are stored in every line bucket they span, so mixed single-line + multi-line workloads stay on the fast path.
- `DecorationService` delegates add/remove/lookup to the cache; the existing `SortedList` remains for sorted iteration elsewhere.

### Buffer line mutations

The cache listens to active buffer `trim` / `insert` / `delete` events (via `IBufferService`) and keeps indexes aligned with marker movement:

| Event | Strategy |
|--------|----------|
| **Trim** | Shift all map keys down synchronously; update `_indexedStartLine` |
| **Insert** | After markers update (microtask): shift keys at/above insert index; sync start lines; full re-index only for rare multi-line decorations that span the insert point |
| **Delete** | After markers update (microtask): drop deleted line keys, shift keys below; full re-index only for rare multi-line decorations that span the deleted range but survive |

Insert/delete maintenance is **O(unique indexed lines)** for key shifts plus a light pass over decorations, not a full remove/re-add per decoration.

### `MicrotaskTimer`

- New `MicrotaskTimer` in `src/common/Async.ts` mirrors `TimeoutTimer` ergonomics (`set`, `cancel`, `dispose`) but schedules via `queueMicrotask`.
- `DecorationLineCache` uses it to batch multiple insert/delete sync callbacks into one microtask after marker line updates.

### Internal API

- `IInternalDecoration._indexedStartLine` supports bucket removal when `marker.line` is cleared on dispose.

## Benchmarks

New `test/benchmark/DecorationService.benchmark.ts`:

- **Single-line dense / sparse hit** — 20k decorations, repeated lookup on line 0
- **Mixed single-line + multi-line** — 19,999 single-line + one 2-line decoration
- **Viewport grid scan** — 30×80 grid × 20 iterations with 20k decorations

Run benchmarks:

    npm run benchmark -- -s "out-test/benchmark/DecorationService.benchmark.js|DecorationService.forEachDecorationAtCell" out-test/benchmark/DecorationService.benchmark.js

Results:

| Benchmark case | Before (ms) | After (ms) | Delta (ms) | Delta (%) |
|----------------|-------------|------------|------------|-----------|
| Single-line dense / sparse line hit | 7.37 | 0.02 | −7.35 | −99.7% |
| Mixed single-line + multi-line / sparse line hit | 3.49 | 0.02 | −3.47 | −99.4% |
| Viewport grid scan (30×80 × 20 iter, 20k decorations) | 6385.83 | 1.80 | −6384.03 | −100.0% |

## Manual perf profile

Test case:

1. Set scrollback to 100000
2. Enable overview ruler (probably doesn't make a difference)
3. Run `ls -lR .` 4 times to fill scrollback
4. Find next "d" (daniel is username so there are many results)

Before:

Screenshot 2026-05-26 at 10 47 53 AM

After:

Screenshot 2026-05-26 at 10 45 34 AM

## Tests

Extended `src/common/services/DecorationService.test.ts`:

- Existing single/multi-line and x-range behavior
- Dense single-line set with a multi-line decoration on the same line
- Line index maintenance: buffer trim, trim-dispose, line insert (`splice`)
- Direct `DecorationLineCache` smoke test

## Commits on this branch (vs `master`)

1. **Speed up decoration cell lookup with a per-line index** — core index, `IBufferService` wiring, tests, benchmark
6. **Refactor decoration line index into DecorationLineCache** — extract cache class; optimized insert/delete key shifting
7. **Add MicrotaskTimer and use it for decoration line index sync** — shared async helper; cache uses it for deferred sync

## Test plan

- [x] `npm run build && npm run esbuild`
- [x] `npm run test-unit -- out-esbuild/common/services/DecorationService.test.js`
- [x] `npm run benchmark -- -s "out-test/benchmark/DecorationService.benchmark.js|DecorationService.forEachDecorationAtCell" out-test/benchmark/DecorationService.benchmark.js`
- [x] `npm run lint-changes`
- [x] Manual: search addon with decorations enabled on a large buffer; verify highlights render correctly while scrolling and after buffer reflow
- [x] Manual: decoration with `height > 1` still highlights all spanned rows

## Notes

- Complements broader marker/buffer work in [#5853](https://github.com/xtermjs/xterm.js/pull/5853); scoped to the current `forEachDecorationAtCell` API used by DOM/WebGL render paths today.
- Makes higher highlight limits (see [#3809](https://github.com/xtermjs/xterm.js/issues/3809)) practical for the renderer; registration/disposal performance ([#4911](https://github.com/xtermjs/xterm.js/issues/4911)) is a separate concern.

### DIFF
--- a/src/common/Async.ts
+++ b/src/common/Async.ts
@@ -73,6 +73,42 @@ export class TimeoutTimer implements IDisposable {
   }
 }
 
+/**
+ * Schedules a single runner on the microtask queue. Unlike {@link TimeoutTimer}, a scheduled
+ * microtask cannot be unqueued; {@link cancel} prevents the runner from executing if it has not
+ * run yet.
+ */
+export class MicrotaskTimer implements IDisposable {
+  private _isScheduled = false;
+  private _isDisposed = false;
+
+  public dispose(): void {
+    this.cancel();
+    this._isDisposed = true;
+  }
+
+  public cancel(): void {
+    this._isScheduled = false;
+  }
+
+  public set(runner: () => void): void {
+    if (this._isDisposed) {
+      throw new Error('Calling set on a disposed MicrotaskTimer');
+    }
+    if (this._isScheduled) {
+      return;
+    }
+    this._isScheduled = true;
+    queueMicrotask(() => {
+      if (!this._isScheduled) {
+        return;
+      }
+      this._isScheduled = false;
+      runner();
+    });
+  }
+}
+
 export class IntervalTimer implements IDisposable {
   private _disposable: IDisposable | undefined;
   private _isDisposed = false;
--- a/src/common/services/DecorationService.test.ts
+++ b/src/common/services/DecorationService.test.ts
@@ -4,7 +4,7 @@
  */
 
 import { assert } from 'chai';
-import { DecorationService } from './DecorationService';
+import { DecorationLineCache, DecorationService } from './DecorationService';
 import { IMarker } from 'common/Types';
 import { Disposable } from 'common/Lifecycle';
 import { Emitter } from 'common/Event';
@@ -146,6 +146,13 @@ describe('DecorationService', () => {
       assert.strictEqual([...service.getDecorationsAtCell(0, 6)].length, 1);
       assert.strictEqual([...service.getDecorationsAtCell(0, 7)].length, 1);
       assert.strictEqual([...service.getDecorationsAtCell(0, 8)].length, 0);
+    });
+  });
+
+  describe('DecorationLineCache', () => {
+    it('should return undefined for lines with no indexed decorations', () => {
+      const cache = new DecorationLineCache();
+      assert.isUndefined(cache.getDecorationsOnLine(0));
     });
   });
 
--- a/src/common/services/DecorationService.test.ts
+++ b/src/common/services/DecorationService.test.ts
@@ -8,7 +8,9 @@ import { DecorationService } from './DecorationService';
 import { IMarker } from 'common/Types';
 import { Disposable } from 'common/Lifecycle';
 import { Emitter } from 'common/Event';
-import { MockLogService } from 'common/TestUtils.test';
+import { MockLogService, MockBufferService, MockOptionsService } from 'common/TestUtils.test';
+import { Buffer } from 'common/buffer/Buffer';
+import { DEFAULT_ATTR_DATA } from 'common/buffer/BufferLine';
 
 function createFakeMarker(line: number): IMarker {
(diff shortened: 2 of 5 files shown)


Tyriar:
@anthonykim1 big improvement to a long standing issue btw

PR #5961, opened 2026-05-29
Enable native touch copy and paste on iOS

### LINKED ISSUE #3727: copy and paste do not work on ipad
## Details
- Browser and browser version: Safari on an iPad
- OS version: all versions of iPadOS (and iPhoneOS too)
- xterm.js version: all versions

### Steps to reproduce

1. Go to the demo at https://xtermjs.org
2. The terminal works. However, it is not possible to select text at all by touching the screen, so copy (and paste) are impossible.
3. Using a keyboard and trackpad on the iPad, you can select text, but when you copy with (cmd+c), it does NOT copy, leaving the clipboard unchanged.  Paste does work with a keyboard.
4. I also tried both the canvas and dom, and the webgl addon.  I couldn't select text with any of those.

I use xtermjs, in https://cocalc.com for desktop, but for mobile I've been using an ancient version of term.js... almost entirely because copy paste actually always fully worked with term.js on ipad.

williamstein:
## Summary

Fixes #3727 by letting iOS use native DOM selection and the native paste menu for terminals rendered with DOM rows.

The change is intentionally narrow:

- Detect Apple touch devices with a coarse primary pointer and add `xterm-native-touch-selection` to the terminal element.
- In that mode, make `.xterm-rows` selectable and avoid xterm's internal gesture/selection mousedown handling when mouse reporting is inactive.
- Keep mouse reporting behavior intact for applications that enable mouse events.
- Move the helper textarea to the cursor with a usable touch target so long-press paste works natively.
- Prevent default textarea insertion when the paste event already provides clipboard data, avoiding duplicate pasted text.
- Make the demo prefer the DOM renderer on coarse-pointer devices, since native iOS selection cannot select text rendered only to a WebGL canvas.

## Verification

- `npm run build`
- `npm run lint-changes`
- `npm run test-unit`
- `npm run esbuild`
- Chromium integration suites for core and all addons, run against system Chromium on this host

I also manually verified on iPhone Safari that terminal output can be selected/copied with the native iOS UI and that long-pressing the prompt cursor offers Paste and inserts text once.

Note: the local host is Ubuntu 26.04, and `npx playwright install` currently refuses to install Playwright browser binaries for this OS. Because of that I could run Chromium integration coverage via system Chromium, but not local Firefox/WebKit integration coverage.

### DIFF
--- a/demo/client/client.ts
+++ b/demo/client/client.ts
@@ -118,6 +118,11 @@ const xtermjsTheme = {
   white: '#F8F8F8',
   brightWhite: '#FFFFFF'
 } satisfies ITheme;
+
+function hasCoarsePrimaryPointer(): boolean {
+  return window.matchMedia?.('(hover: none) and (pointer: coarse)').matches ?? false;
+}
+
 function setPadding(): void {
   term!.element!.style.padding = parseInt(paddingElement.value, 10).toString() + 'px';
   addons.fit.instance!.fit();
@@ -242,11 +247,13 @@ if (document.location.pathname === '/test') {
   controlBar.setTabVisible('addon-serialize', true);
   controlBar.setTabVisible('addon-web-fonts', true);
   controlBar.setTabVisible('addon-web-links', !!addons.webLinks.instance);
-  controlBar.setTabVisible('addon-webgl', true);
-  addonWebglWindow.setTextureAtlas(addons.webgl.instance!.textureAtlas!);
-  addons.webgl.instance!.onChangeTextureAtlas(e => addonWebglWindow.setTextureAtlas(e));
-  addons.webgl.instance!.onAddTextureAtlasCanvas(e => addonWebglWindow.appendTextureAtlas(e));
-  addons.webgl.instance!.onRemoveTextureAtlasCanvas(e => addonWebglWindow.removeTextureAtlas(e));
+  controlBar.setTabVisible('addon-webgl', !!addons.webgl.instance);
+  if (addons.webgl.instance) {
+    addonWebglWindow.setTextureAtlas(addons.webgl.instance.textureAtlas!);
+    addons.webgl.instance.onChangeTextureAtlas(e => addonWebglWindow.setTextureAtlas(e));
+    addons.webgl.instance.onAddTextureAtlasCanvas(e => addonWebglWindow.appendTextureAtlas(e));
+    addons.webgl.instance.onRemoveTextureAtlasCanvas(e => addonWebglWindow.removeTextureAtlas(e));
+  }
 
   paddingElement.value = '0';
   addDomListener(paddingElement, 'change', setPadding);
@@ -304,10 +311,12 @@ function createTerminal(): Terminal {
   addons.progress.instance = new ProgressAddon();
   addons.unicodeGraphemes.instance = new UnicodeGraphemesAddon();
   addons.clipboard.instance = new ClipboardAddon();
-  try {  // try to start with webgl renderer (might throw on older safari/webkit)
-    addons.webgl.instance = new WebglAddon();
-  } catch (e) {
-    console.warn(e);
+  if (!hasCoarsePrimaryPointer()) {
+    try {  // try to start with webgl renderer (might throw on older safari/webkit)
+      addons.webgl.instance = new WebglAddon();
+    } catch (e) {
+      console.warn(e);
+    }
   }
   addons.webLinks.instance = new WebLinksAddon();
   addons.webFonts.instance = new WebFontsAddon();
--- a/src/browser/CoreBrowserTerminal.ts
+++ b/src/browser/CoreBrowserTerminal.ts
@@ -60,6 +60,12 @@ import { Emitter, EventUtils, type IEvent } from 'common/Event';
 import { addDisposableListener } from 'browser/Dom';
 import { MutableDisposable, toDisposable } from 'common/Lifecycle';
 
+function shouldUseNativeTouchSelection(targetWindow: Window): boolean {
+  const navigator = targetWindow.navigator as Navigator & { maxTouchPoints?: number };
+  return (/iPad|iPhone|iPod/.test(navigator.userAgent) || (navigator.platform === 'MacIntel' && (navigator.maxTouchPoints ?? 0) > 1))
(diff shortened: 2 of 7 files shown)


williamstein:
Heh @Tyriar (Daniel) this is William (lead dev of cocalc, sagemath guy, we had lunch at Microsoft once), and I was considering a WASM competitors to xtermjs because it supported iphone copy/paste, which is something that didn't get supported when you built xterm.js out of termjs.  https://github.com/xtermjs/xterm.js/issues/3727 is about this and has been open since I created it in 2022.    I would much rather switch with xterm.js, so I wondered if codex could figure this copy/paste problem out, and after a bunch of false starts and some interactive debugging, it did I think properly solve it.  This PR isn't a workaround like https://github.com/xtermjs/xterm.js/issues/3727#issuecomment-3857343838 but an actual fix.

I don't use android and have no idea how this is relevant to that. 

 -- william

williamstein:
Note that the PR includes a few related fixes in this commit: iOS native selection/paste, hidden helper caret, and hidden measurement DOM.  I noticed these when fully testing the integration with cocalc. Basically there was an issue where the cursor was confusing and this resolved it.

src/browser/CoreBrowserTerminal.ts
```diff
@@ -60,6 +60,12 @@ import { Emitter, EventUtils, type IEvent } from 'common/Event';
 import { addDisposableListener } from 'browser/Dom';
 import { MutableDisposable, toDisposable } from 'common/Lifecycle';
 
+function shouldUseNativeTouchSelection(targetWindow: Window): boolean {
+  const navigator = targetWindow.navigator as Navigator & { maxTouchPoints?: number };
+  return (/iPad|iPhone|iPod/.test(navigator.userAgent) || (navigator.platform === 'MacIntel' && (navigator.maxTouchPoints ?? 0) > 1))
+    && (targetWindow.matchMedia?.('(hover: none) and (pointer: coarse)').matches ?? true);
+}
```
Tyriar, line 67: Could we have a new option `nativeSelection: 'auto' | boolean` for this? It could explain that it only applies to DOM renderer and can only select within viewport by design. In the option's validation logic it could also resolve to true/false such that the webgl addon/demo can check it for the resolved value?

src/browser/CoreBrowserTerminal.ts
```diff
@@ -60,6 +60,12 @@ import { Emitter, EventUtils, type IEvent } from 'common/Event';
 import { addDisposableListener } from 'browser/Dom';
 import { MutableDisposable, toDisposable } from 'common/Lifecycle';
 
+function shouldUseNativeTouchSelection(targetWindow: Window): boolean {
+  const navigator = targetWindow.navigator as Navigator & { maxTouchPoints?: number };
+  return (/iPad|iPhone|iPod/.test(navigator.userAgent) || (navigator.platform === 'MacIntel' && (navigator.maxTouchPoints ?? 0) > 1))
```
Tyriar, line 65: We should support Android too, if you get AI to do this I can try test it. Is just checking course pointer/no hover the approach for this instead of getting into the UA?

src/browser/CoreBrowserTerminal.ts
```diff
@@ -60,6 +60,12 @@ import { Emitter, EventUtils, type IEvent } from 'common/Event';
...
+    && (targetWindow.matchMedia?.('(hover: none) and (pointer: coarse)').matches ?? true);
```
Tyriar, line 66: Comment about the `hover: none` part making this stricter would be good

demo/client/client.ts
```diff
@@ -304,10 +311,12 @@ function createTerminal(): Terminal {
   addons.progress.instance = new ProgressAddon();
   addons.unicodeGraphemes.instance = new UnicodeGraphemesAddon();
   addons.clipboard.instance = new ClipboardAddon();
-  try {  // try to start with webgl renderer (might throw on older safari/webkit)
-    addons.webgl.instance = new WebglAddon();
-  } catch (e) {
-    console.warn(e);
+  if (!hasCoarsePrimaryPointer()) {
```
Tyriar, line 314: Instead of demo customization, could we check the option inside the webgl addon and throw when touch/native selection is enabled?

src/browser/CoreBrowserTerminal.ts
```diff
@@ -60,6 +60,12 @@ import { Emitter, EventUtils, type IEvent } from 'common/Event';
 import { addDisposableListener } from 'browser/Dom';
 import { MutableDisposable, toDisposable } from 'common/Lifecycle';
 
+function shouldUseNativeTouchSelection(targetWindow: Window): boolean {
```
Tyriar, line 63: I think `MouseService` should own the detection logic and set the class, if other classes need it, a method should be exposed on `MouseService`

Tyriar (changes requested):
Yep I remember out lunch 😃

williamstein:
thanks @Tyriar -- I'm extremely busy right now and also don't have access to android directly.  I think the best thing is to let this sit for a bit until I've launched it in production and got some feed from cocalc users (which will have android).    I'll follow up later.

ibetitsmike:
@williamstein hey, have you had a chance to look at this change again?

CaneTLOTW:
A data point from a downstream integration: we used the same native DOM-selection direction from this PR to solve iOS copy/paste in the Home Assistant Companion WebView around ttyd/xterm.

In our case normal mobile mode still needs touch/swipe terminal behavior, so we expose an explicit Sel toggle: entering it switches from the configured renderer to DOM rows, enables native WebKit selection/callouts, suppresses competing terminal touch handling, and leaves the helper textarea available for native Paste; leaving Sel restores the configured renderer and normal terminal gestures. We also route insertFromPaste through xterm's normal paste() path to avoid duplicate insertion.

This has now been tested successfully on iPhone in Home Assistant Companion: native long-press selection/Copy and native Paste both work. We kept the adaptation at ttyd's integration layer rather than carrying a separate xterm fork.

Downstream implementation notes: https://github.com/CaneTLOTW/ha-codex/blob/main/codex/ttyd-mobile-keys/README.md

Thanks — this PR was very useful in validating the approach.

PR #5964, opened 2026-05-30
Avoid "fake markers" in DecorationService.test.ts.

PerBothner:
It is better to test the actual Marker implementation in the context of the actual Buffer implementation, instead of constructing "fake Markers". For one, it works with the PR #5853 re-implenation of markers.


src/common/services/DecorationService.test.ts
```diff
@@ -113,7 +99,7 @@ describe('DecorationService', () => {
       const buffer = bufferService.buffer;
       (buffer as Buffer).fillViewportRows();
 
-      for (let i = 0; i < 100; i++) {
+      for (let i = 0; i < buffer.lines.length; i++) {
```
Tyriar: The buffer being used here is wrong, should be good if the services aren't reconstructed inside the tests?

Tyriar (changes requested):

src/common/services/DecorationService.test.ts
PerBothner: Not sure I understand what you're asking. As I see it, creating a marker on a non-existing line is not meaningful - it is undefined behavior. A test that expects a specific result from undefined behavior does not seem very useful. Specifically, note that PR #5853 attaches Markers to lines; only "disposed" Marker can exist not attached to a marker. I guess if the requested line number is out of range we can have `addMarker` create a disposed marker (which has line number `-1`).

PerBothner: Or do you mean this test should be using the `bufferService` created in the `beforeEach`, like the other tests? Haven't tried that, bt it makes sense - I think I overlooked changing that test to use `bufferService`.

PerBothner: I updated 4 more tests to use the shared `bufferService` variable.

Tyriar (approved):
Yep that's what I meant 👍

PR #5972, opened 2026-06-01
fix(buffer): clear stale attrs and combined data in BufferLine

Tyriar:
## Summary

When a buffer cell is overwritten with a plain character or copied from another cell, leftover extended-attribute and combining-character state could remain in `_combined` / `_extendedAttrs` even after `_data` no longer indicates combined or extended content. That desync can affect rendering, width calculation, string translation, and `loadCell` when reusing a `CellData` work object.

This PR fixes that by making **`_data` flags the source of truth** and treating the sparse maps as caches:

- **`IS_COMBINED_MASK` / `HAS_EXTENDED`** decide whether map entries are read.
- **`setCellFromCodepoint`** (input hot path) only updates `_data`, plus `_extendedAttrs` when the new attrs carry extended data. It does not touch `_combined` or call `delete` per codepoint.
- **`loadCell`** clears `CellData.combinedData` when not combined; when extended flags are off it assigns `DEFAULT_ATTR_DATA.extended.clone()` so a reused `CellData` cannot mutate a line map entry (fixes addon-serialize underline `4m` vs `24m` failures).
- **`setCell`** writes map entries only when the new cell needs them.
- **`copyCellsFrom`** copies map data per cell from src `_data` flags (bounded range, no `Object.keys` scan, no dest `delete`).
- **`clone` / `copyFrom`** rebuild maps from flags; **`fill` / `resize`** still reset or trim maps on structural changes.

## Why this approach

The first fix attempt cleared stale entries with **`delete` on every overwrite**. That is correct but expensive: V8 penalizes repeated `delete` on dictionary-mode objects. Microbenchmarks showed ~8× regression on `setCellFromCodepoint` and ~3.5× on a 1 MiB headless `writeSync`.

**Guarded `delete`** (`index in this._combined`) reduced that cost but still added branches and measurable overhead on the hot path (~20–30% on isolated benchmarks, ~6–8% on `writeSync`).

**Flag-gated caches** avoid map maintenance on the common case (plain ASCII). Stale map keys may linger but are **inert** until `_data` says otherwise; readers gate on flags. Structural operations (`fill`, `resize`) still reclaim map memory.

## Performance

Re-measured on **HEAD** (`ad1e02808`, after `loadCell` clone fix) vs `master` and earlier iterations. Median of 9 runs per case; harness: `bin/bufferline_perf_compare.mjs` / `bin/run_bufferline_perf_matrix.sh`.

| Benchmark | master | Original (`delete` every write) | Guarded `delete` | Flag-gated (this PR) |
|---|--:|--:|--:|--:|
| `setCellFromCodepoint` (2M plain) | 7.84 ms | 64.05 ms (+717%) | 9.59 ms (+22%) | 7.86 ms (+0.2%) |
| `setCellFromCodepoint` (extended) | 12.54 ms | 41.25 ms (+229%) | 13.74 ms (+10%) | 12.57 ms (+0.3%) |
| `setCell` (plain overwrite) | 6.79 ms | 64.88 ms (+856%) | 8.55 ms (+26%) | 6.83 ms (+0.6%) |
| `setCell` (clear combined+extended) | 7.55 ms | 123.21 ms (+1532%) | 59.08 ms (+682%) | 7.60 ms (+0.7%) |
| `copyCellsFrom` (50k × 40 cells) | 10.43 ms | 28.63 ms (+174%) | 11.82 ms (+13%) | 9.68 ms (−7%) |
| Headless `writeSync` (1 MiB ASCII) | 238.14 ms | 841.08 ms (+253%) | 251.22 ms (+5.5%) | 261.46 ms (+9.8%) |

**Headless `writeSync` repeatability** (3 extra runs, same harness):

| Variant | Run 1 | Run 2 | Run 3 |
|---|--:|--:|--:|
| master | 247.9 ms | 238.1 ms | 237.8 ms |
| flag-gated (HEAD) | 264.0 ms | 260.7 ms | 270.6 ms |

**Interpretation**

- vs **original delete-based fix**: large win on every benchmark; not viable on the input path.
- vs **`master`**: hot-path microbenchmarks match `master` within noise; `copyCellsFrom` is slightly faster.
- **End-to-end** `writeSync` is ~6–10% above `master` (matrix +3-run median ~265 ms vs ~241 ms). The `loadCell` clone change does not materially shift microbench numbers; remaining gap is likely parser/scroll paths outside per-codepoint `setCellFromCodepoint`.

## Testing

- `npm run test-unit -- **/out-esbuild/**/BufferLine*.test.js` (51 passing)
- `npm run test-integration-chromium -- --suite=addon-serialize` (40 passing)

<div><a href="https://cursor.com/agents/bc-b07b05e2-53ac-4b21-b53e-64e93cc46c2c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b07b05e2-53ac-4b21-b53e-64e93cc46c2c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


jerch:
@Tyriar Was there a real issue behind this PR? Because the _combined and _extendedAttrs entries were always meant to be eval'ed only when the corresponding flag is set. I skipped the more defensive deletion because of the performance penalty. Same for a `workCell` - its interfaces must always test the flags, so a stale entry does not hurt.

I am asking because in data heavy input I see a rather big performance penalty of roughly 20% from the changes to `copyFrom`. That method does the line recycling during ringbuffer reusage, which is guaranteed to always have empty _combined and _extendedAttrs entries from the `newLine` blueprint. That is a rather big dent in performance.

Suggestion:
- We should check if we can get rid of the defensive deletion/copy pattern again by making sure, that the interfaces always eval the flags first.
- Maybe create a `copyFromBlank`, where we can skip the `_copyCellMapsFrom` call in total by the guarantee, that a blank line should never contain _combined and _extendedAttrs entries.

Also created #6101 to track the `copyFrom` issue.

PR #5987, opened 2026-06-02
fix(webgl): avoid glyph atlas mipmaps

### LINKED ISSUE #5986: WebGL glyph atlas corruption on Linux/Wayland when generating mipmaps
## Details
- Browser and browser version: Chromium 148.0.7778.96 (Arch Linux) for the Hermes downstream observation; Electron/Chromium in Orca for the separate downstream observation
- OS version: Linux + Wayland. Observed downstream on Arch Linux rolling / kernel 7.0.10 and separately on Ubuntu 26.04 with Intel Arc graphics
- xterm.js version: downstream packages using `@xterm/addon-webgl@0.19.0` and `@xterm/addon-webgl@0.20.0-beta.215`; current `master` still has the same `gl.generateMipmap(gl.TEXTURE_2D)` call in `addons/addon-webgl/src/GlyphRenderer.ts`

### What went wrong

Downstream applications using the WebGL addon can render dense terminal/TUI output as black rectangular cell blocks, diagonal smears, or corrupted glyphs on Linux + Wayland GPU stacks.

The common path appears to be the glyph atlas upload in `GlyphRenderer`, which currently uploads the canvas atlas and then calls:

```ts
gl.generateMipmap(gl.TEXTURE_2D);
```

In one downstream app, Electron/ANGLE logged `GL_INVALID_OPERATION` while allocating mipmaps for generation. Removing the mipmap generation and setting explicit non-mipmapped filters fixed the corruption while keeping WebGL enabled.

### Steps to reproduce

I have not reduced this to a standalone xterm.js demo repro. The failure appears GPU/driver/backend-specific, so I want to be explicit about that rather than overstate it.

Downstream reproductions:

1. Open a WebGL-backed xterm terminal on Linux + Wayland.
2. Run dense/scrolling TUI output such as `htop`, `nvim`, or an agent TUI.
3. Scroll or let output update heavily.
4. Observe black cell blocks / corrupted glyphs over terminal text.

Evidence from downstreams:

- Orca issue: https://github.com/stablyai/orca/issues/1579
- Orca downstream fix: https://github.com/stablyai/orca/pull/1743
- Hermes Agent issue: https://github.com/NousResearch/hermes-agent/issues/37203
- Hermes downstream workaround PR: https://github.com/NousResearch/hermes-agent/pull/37205

### Expected behavior

The WebGL renderer should keep glyph text visually stable during dense output and scrolling.

### Actual behavior

On affected Linux + Wayland GPU stacks, the glyph atlas can corrupt and render terminal text as black boxes / smears.

### Proposed fix

Avoid generating mipmaps for the terminal glyph atlas and explicitly use non-mipmapped linear filters:

```ts
gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
```

The glyph atlas is sampled at terminal glyph scale; mipmaps do not appear necessary for this path. I will open a small PR with only that change so maintainers can evaluate it directly.

Whamp:
## Summary

Fixes #5986.

This removes mipmap generation from the WebGL glyph atlas upload path and explicitly sets non-mipmapped linear filters for atlas pages:

```ts
gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
```

The goal is to avoid glyph atlas corruption seen in downstream Linux + Wayland WebGL terminals while preserving the WebGL renderer.

## Context

Two downstream applications have hit the same failure class: dense terminal text can turn into black cell blocks, smears, or corrupted glyphs on Linux + Wayland GPU stacks when the WebGL addon is active.

- xterm.js issue: https://github.com/xtermjs/xterm.js/issues/5986
- Orca downstream issue: https://github.com/stablyai/orca/issues/1579
- Orca downstream fix: https://github.com/stablyai/orca/pull/1743
- Hermes Agent downstream issue: https://github.com/NousResearch/hermes-agent/issues/37203
- Hermes Agent downstream workaround PR: https://github.com/NousResearch/hermes-agent/pull/37205

In the Orca case, Electron/ANGLE logged `GL_INVALID_OPERATION` while allocating mipmaps for generation. Removing `gl.generateMipmap(gl.TEXTURE_2D)` and using explicit non-mipmapped filters resolved the corruption there. The Hermes downstream workaround applies the same change to its installed `@xterm/addon-webgl` package.

I have not reduced this to a standalone xterm.js demo repro; it appears GPU/driver/backend-specific. This PR is intentionally narrow so the exact WebGL state change can be reviewed directly.

## Testing

Ran in a clean `/tmp/xterm.js-pr` checkout to avoid parent-directory `node_modules` pollution:

- `npm ci`
- `npm run build`
- `npm run lint-changes`
- `npm run setup`
- `npm run test-unit` — 2367 passing
- Started the demo with `PORT=3310 npm start`
- Built the demo client with `npm run esbuild-demo-client`
- Loaded `http://127.0.0.1:3310/` in Chromium and opened the `webgl` panel; confirmed the page loaded and a WebGL2 canvas was present
- Verified the built demo bundle contains `TEXTURE_MIN_FILTER` / `TEXTURE_MAG_FILTER` and no `generateMipmap`

## Notes

The repository contribution guide asks for tests when changes are easy to test or likely to regress. I did not add a new automated test because this failure is GPU/driver-specific and the existing unit test surface does not directly exercise browser WebGL texture state. The focused build, lint, unit test suite, and demo smoke test are included above.

### DIFF
--- a/addons/addon-webgl/src/GlyphRenderer.ts
+++ b/addons/addon-webgl/src/GlyphRenderer.ts
@@ -389,8 +389,9 @@ export class GlyphRenderer extends Disposable {
     gl.bindTexture(gl.TEXTURE_2D, this._atlasTextures[i].texture);
     gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
     gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
     gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, atlas.pages[i].canvas);
-    gl.generateMipmap(gl.TEXTURE_2D);
     this._atlasTextures[i].version = atlas.pages[i].version;
   }
 


jerch (approved):
Can repro the issue with this stress test:

```javascript
const TEXT = '!"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~';

async function stress() {
  for (let fr = 0; fr < 256; fr += 2) {
    for (let fg = 0; fg < 256; fg += 2) {
      for (let fb = 0; fb < 256; fb += 2) {
        console.log('\x1b[H\x1b[2J\x1b[mThis should be readable with white FG color.');
        console.log(`r: ${fr} g:${fg} b:${fb}`);
        console.log(`\x1b[38;2;${fr};${fg};${fb}m${TEXT}`);
        await new Promise(r => setTimeout(r, 20));
      }
    }
  }
}

stress();
```

This PR solves it on my old laptop. 👍

mgabor3141:
Confirming this on another GPU/driver — **AMD Radeon 680M (radeonsi via ANGLE)** on Linux:

```
renderer: ANGLE (AMD, AMD Radeon 680M (radeonsi rembrandt ACO), OpenGL ES 3.2)
vendor:   Google Inc. (AMD)
maxTextureSize: 16384
GL_INVALID_OPERATION: 0x00000502 ... allocateMipmapLevelsForGeneration ... Unexpected driver error.  (repeating)
```

Two details that may help confirm the diagnosis:
- The atlas was only **512×512** (one page), nowhere near the 16384 max — so it's not a texture-size limit, the `generateMipmap` call itself is what the driver rejects.
- The context was **never lost** (no `webglcontextlost`) — it's a pure live-context atlas-upload failure, so `onContextLoss`-based recovery doesn't help here; only dropping the mipmap path fixes it.

We shipped the **identical change** (`TEXTURE_MIN_FILTER=LINEAR` + remove `generateMipmap`) as an interim `pnpm patch` on `@xterm/addon-webgl@0.20.0-beta.194`, and the affected user confirms the black rectangles are gone with no visual regression. 👍 for this PR — we'll drop our patch once it lands in a published release.

jerch:
> * so it's not a texture-size limit, the `generateMipmap` call itself is what the driver rejects.

I think this is a recent bug in ANGLE itself, because it works on the same machine with webkit and firefox, but not with chromium.

Whamp:
Thanks, both. The ANGLE root cause makes sense to me especially with Firefox/WebKit not hitting it on the same machine. Glad the smaller fix works for the stress case too.

jerch:
@Tyriar Since you wrote the webgl parts and introduced the mipmap usage for a reason - could you check if the linear approach does not degrade output too much? (I see no difference here, but my old laptop is DPR 1 with 1:1 pixel alignment) If the quality degrades e.g. on mac, then we might have to do an OS/browser switch here.

@Whamp 
Would you be able to do a minimal webgl repro w'o xterm.js? With that and proper plattform/driver conditions we should issue it to the chromium folks.

Tyriar (approved):
Very cool 👏 not sure why mipmaps were being generated at all as I know what they are. The goal is that the glyph will never need to be sampled so things are pixel perfect.

Whamp:
Put together a standalone repro harness here:

https://gist.github.com/Whamp/875d0f2184a5a4b43e017f439cf6cd00

It does not import xterm.js. It just creates a 512x512 canvas-backed RGBA texture, uploads it with `texImage2D`, calls `generateMipmap`, checks `gl.getError()`, then compares against the fixed path with `TEXTURE_MIN_FILTER=LINEAR` / `TEXTURE_MAG_FILTER=LINEAR` and no mipmap generation.

I tested it locally on my desktop and it does not reproduce there:

```text
ANGLE (AMD, AMD Radeon 780M Graphics (radeonsi phoenix ACO), OpenGL ES 3.2)
after generateMipmap: NO_ERROR
stress result: NO_ERROR after 2048 iterations
after fixed upload: NO_ERROR
RESULT: not reproduced on this browser/driver
```

So the useful next step is probably to run that same page on the affected 680M/radeonsi/ANGLE setup and attach the output to a Chromium issue if it shows the `INVALID_OPERATION` path.

jerch:
@Whamp Thank you for looking into a shorter version. It indeed does not crash for me either. I wonder if the 512 --> 1024 texture size transition in xterm.js might trigger it, while the not yet working repro stays on 512.

PR #6001, opened 2026-06-04
fix atlas callstack exception

jerch:
Eureka - my infamous atlas stress test finally runs through without error, generating ~197M glyphs.

Well after this patch here :smiley_cat: 

Fixes: The pattern `this._glyphs.push(...p.glyphs)` runs into callstack exception for very late page merges (`p.glyphs` spreading into too many arguments).

@Tyriar Added the stress test under `bin/test_atlas_stress.js`. Not sure if we should add it to CI, it takes quite long to run (can be made a lot faster w'o the promise await, but I don't know how to grab the output glitches it produced before, so maybe it has to stay manual testing).

Manually tested on Ubuntu 24 with:
- Chromium 148
- Firefox 151
- WebKit 605.1.15 (epiphany)


bin/test_atlas_stress.js
```diff
@@ -0,0 +1,17 @@
...
+        console.log('\x1b[H\x1b[2J\x1b[mThis should be readable with white FG color.');
+        console.log(`r: ${r} g:${g} b:${b}`);
+        console.log(`\x1b[38;2;${r};${g};${b}m${TEXT}`);
+        await new Promise(r => setTimeout(r, 0));
+      }
+    }
+  }
+}
+
+stress();
```
Tyriar: Think this could easily be turned into a test button on the demo? More accessible there if so and co-located with similar things.

addons/addon-webgl/src/TextureAtlas.ts
```diff
@@ -1080,17 +1080,15 @@ class AtlasPage {
...
-        this._usedPixels += p._usedPixels;
-      }
-    }
     this.canvas = createCanvas(document, size, size);
     // The canvas needs alpha because we use clearColor to convert the background color to alpha.
     // It might also contain some characters with transparent backgrounds if allowTransparency is
     // set.
     this.ctx = throwIfFalsy(this.canvas.getContext('2d', { alpha: true }));
+    if (sourcePages) {
+      this._glyphs = this._glyphs.concat(...sourcePages.map(p => p.glyphs));
```
Tyriar: I've learned about spread causing call stack issues since writing this.

Can we avoid the spread here too? Also map/reduce causes extra GC churn, safest as a for loop since `p.glyphs` could be large?

Tyriar (changes requested):
Let's keep to manual testing, this sort of thing is tough to reliably test in CI and don't want to push the time out further.

addons/addon-webgl/src/TextureAtlas.ts
jerch: Well the concat at once over all `p.glyphs` arrays is the fastest without causing render frame stall warnings anymore. An interim loop version did the concat one-by-one which still showed the stall warnings.

Will see what I can do about it.

jerch: @Tyriar Separated the default case (quadmerge) from others. This way the spread operator and map/reduce can be omitted. The fallback does concat-loop joining, which shows worse runtime (again GC overhead).

bin/test_atlas_stress.js
jerch: The stress test is now under webgl.

jerch:
@Tyriar The stress test reveals one more thing - the cached data gets rather big after a while. I wonder if we could find a less steep growing data structure. Maybe that is worth an investigation, as it hogs quite some memory in JS. But not in this PR.

Tyriar:
Well it's shared between all terminals in the same window and this is meant to be an edge case. I'm sure agentic CLIs have increased the chance of texture atlas merging but I'm still not sure this stress test is realistic and worth optimizing for in practice. This cache is in the end easy to throw away also, this does happen I think when there's no more room for a certain glyph. If you have ideas to reduce memory usage without major changes though.

Tyriar (approved):

PR #6002, opened 2026-06-04
fix non-ascii clipboard handling

### LINKED ISSUE #6000: addon-clipboard: Base64.encodeText/decodeText use raw btoa/atob, corrupting non-ASCII (UTF-8) clipboard over OSC 52
`@xterm/addon-clipboard`'s default `Base64` provider uses raw `btoa`/`atob`, so any non-ASCII (UTF-8) clipboard text is corrupted over OSC 52: on **decode** each UTF-8 byte becomes a separate Latin-1
  code unit; on **encode** `btoa` throws for code units > 255.

  ## Details
  - Browser and browser version: Edge 148.0.3967.96 (also reproduces in VS Code 1.123 / Electron)
  - OS version: macOS Tahoe 26.5.1
  - xterm.js version: `@xterm/addon-clipboard` — broken since #5874 (commit `c8dddddb`, 2026-05-10) through current `0.3.0-beta.276` (also reproduces in the `master` demo). Last good: `0.3.0-beta.213`
  (VS Code 1.122.x); first observed broken: `0.3.0-beta.220` (VS Code 1.123).

  ### Steps to reproduce

  1. Minimal (no terminal). `decodeText` is literally `return atob(data)`, so in any JS console:

     ```js
     atob('6L2u6K+i')   // 6L2u6K+i = base64 of "轮询" (UTF-8 bytes E8 BD AE E8 AF A2)
     // => "è½®è¯¢"   (expected "轮询")
     ```

  2. End-to-end (OSC 52). With the clipboard addon loaded in the demo, write a set-clipboard sequence to the terminal:

     ```bash
     printf '\x1b]52;c;6L2u6K+i\x07'
     ```

     The system clipboard receives `è½®è¯¢` instead of `轮询`.

  **Expected:** clipboard contains `轮询`.
  **Actual:** clipboard contains `è½®è¯¢` (each UTF-8 byte decoded as Latin-1).
  (Conversely, copying a non-ASCII selection makes `encodeText` → `btoa` throw `InvalidCharacterError`.)

  ## Root cause

  `addons/addon-clipboard/src/ClipboardAddon.ts`, class `Base64`:

  ```ts
  public encodeText(data: string): string { return btoa(data); }
  public decodeText(data: string): string { try { return atob(data); } catch { return ''; } }
  ```

  `atob`/`btoa` operate on Latin-1 "binary strings" (one code unit per byte), with no UTF-8 handling. This regressed in #5874 (commit `c8dddddb`, "remove js-base64 dependency", 2026-05-10): the removed
  `js-base64` did UTF-8 (de/en)coding by default, so dropping it for raw `atob`/`btoa` silently changed the behavior for any non-ASCII text.

  ## Suggested fix (keeps native btoa/atob; no new dependency)

  Wrap with the built-in `TextEncoder`/`TextDecoder`:

  ```ts
  public encodeText(data: string): string {
    const bytes = new TextEncoder().encode(data);
    let bin = '';
    for (let i = 0; i < bytes.length; i++) bin += String.fromCharCode(bytes[i]);
    return btoa(bin);
  }
  public decodeText(data: string): string {
    let bin: string;
    try { bin = atob(data); } catch { return ''; }
    return new TextDecoder().decode(Uint8Array.from(bin, c => c.charCodeAt(0)));
  }
  ```

  I verified this locally against the built addon: encode→decode round-trips correctly for Chinese / Japanese / Cyrillic / emoji, and `decodeText('6L2u6K+i')` returns `"轮询"`.

  Downstream reports: microsoft/vscode#319853, microsoft/vscode#319821.

jerch:
Fixes #6000.

### DIFF
--- a/addons/addon-clipboard/src/ClipboardAddon.ts
+++ b/addons/addon-clipboard/src/ClipboardAddon.ts
@@ -82,16 +82,16 @@ export class BrowserClipboardProvider implements IClipboardProvider {
  * TODO: Once the base64 handling on Uint8Arrays is more common,
  * remove the btoa/atob fallbacks below.
  */
-interface Uint8ArrayB64 extends Uint8Array {
+interface IUint8ArrayB64 extends Uint8Array {
   toBase64(): string;
 }
-interface Uint8ArrayB64Ctor extends Uint8ArrayConstructor {
-  fromBase64(s: string): Uint8ArrayB64;
+interface IUint8ArrayB64Ctor extends Uint8ArrayConstructor {
+  fromBase64(s: string): IUint8ArrayB64;
 }
 
 export class Base64 implements IBase64 {
   public encodeText(data: string): string {
-    const bytes = new TextEncoder().encode(data) as Uint8ArrayB64;
+    const bytes = new TextEncoder().encode(data) as IUint8ArrayB64;
     if (bytes.toBase64 !== undefined) {
       return bytes.toBase64();
     }
@@ -102,9 +102,9 @@ export class Base64 implements IBase64 {
     return btoa(bin);
   }
   public decodeText(data: string): string {
-    if ((Uint8Array as Uint8ArrayB64Ctor).fromBase64 !== undefined) {
+    if ((Uint8Array as IUint8ArrayB64Ctor).fromBase64 !== undefined) {
       try {
-        return new TextDecoder().decode((Uint8Array as Uint8ArrayB64Ctor).fromBase64(data));
+        return new TextDecoder().decode((Uint8Array as IUint8ArrayB64Ctor).fromBase64(data));
       } catch {}
       return '';
     }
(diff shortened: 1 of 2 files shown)


jerch:
@anthonykim1 This might be a showstopper in vscode and prolly needs to be added rather quick.

addons/addon-clipboard/src/ClipboardAddon.ts
```diff
@@ -78,13 +78,43 @@ export class BrowserClipboardProvider implements IClipboardProvider {
   }
 }
 
+/**
+ * TODO: Once the base64 handling on Uint8Arrays is more common,
+ * remove the btoa/atob fallbacks below.
```
Tyriar, line 83: Just node min version we need to wait for? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toBase64#browser_compatibility

Tyriar (approved):

addons/addon-clipboard/src/ClipboardAddon.ts
jerch: Yes and a higher setting in TS target. But there is no need to rush things, the fallback can stay and we re-eval in ... years haha.

PR #6011, opened 2026-06-09
fix: scroll lines into scrollback on SU (CSI S) when scroll region top is at row 1

### LINKED ISSUE #6010: SU (CSI Ps S) deletes lines instead of moving them to scrollback, inconsistent with LF scrolling (scroll region top = 1)
### Description

When a scroll region is set whose top is row 1 (e.g. `CSI 1;20r` — typical for TUIs that protect a status/input panel at the bottom), xterm.js handles the two scroll opcodes inconsistently:

- **LF at the region bottom** → the line scrolled out of the top **is moved to scrollback**. (`BufferService.scroll` handles `scrollTop === 0` by splicing a blank line in below `scrollBottom` and incrementing `ybase`, so the top line is preserved.)
- **`CSI Ps S` (SU / Scroll Up)** → the line scrolled out of the top **is deleted**. (`InputHandler.scrollUp` does `lines.splice(ybase + scrollTop, 1)` unconditionally — the line never reaches scrollback.)

Same buffer state, same scroll region, same visual result — but whether the departing line survives in scrollback depends only on which opcode the application/multiplexer happened to emit.

### Real-world impact

tmux coalesces multi-line pane scrolls into `CSI n S` when redrawing attached clients. Modern AI CLI TUIs (OpenAI Codex CLI with `--no-alt-screen`, Claude Code, etc.) drive exactly this pattern: partial scroll region + heavy redraw. The result is that xterm.js-based terminals (browser terminals, VS Code integrated terminal) intermittently lose a few lines from scrollback/copy while the same byte stream replayed into tmux keeps everything. See e.g. openai/codex#14277 for user-visible fallout of this class.

We hit this in production, captured the failing byte stream, and bisected it: every lost line in our capture is accounted for by a `CSI 2 S` event; every LF-scrolled line survived. Reproduced identically on `@xterm/headless` 5.5.0, 6.0.0 and 6.1.0-beta.285.

### Minimal reproduction (node, @xterm/headless)

```js
const { Terminal } = require('@xterm/headless');

function run(useSU) {
  return new Promise(resolve => {
    const term = new Terminal({ cols: 80, rows: 24, scrollback: 1000 });
    let data = '\x1b[1;20r\x1b[20;1H';                     // region rows 1-20, cursor at bottom
    for (let i = 1; i <= 30; i++) data += `LINE_${String(i).padStart(3,'0')}\r\n`;
    data += useSU ? '\x1b[2S'                              // scroll 2 lines via SU
                  : '\x1b[20;1H\r\n\r\n';                  // scroll 2 lines via LF (control)
    data += '\x1b[r';
    term.write(data, () => {
      const buf = term.buffer.active, seen = new Set();
      for (let i = 0; i < buf.length; i++) {
        const m = buf.getLine(i).translateToString(true).match(/LINE_(\d{3})/);
        if (m) seen.add(Number(m[1]));
      }
      const missing = [];
      for (let i = 1; i <= 30; i++) if (!seen.has(i)) missing.push(i);
      resolve(missing);
    });
  });
}

(async () => {
  console.log('LF scroll, missing:', await run(false));   // -> []
  console.log('SU scroll, missing:', await run(true));    // -> [12, 13]
})();
```

Output (6.0.0):

```
LF scroll, missing: []
SU scroll, missing: [ 12, 13 ]
```

### Expected

`CSI Ps S` with `scrollTop === 0` should preserve scrolled-out lines into scrollback the same way LF scrolling does (i.e. delegate to the same path `BufferService.scroll` uses), so retention does not depend on which opcode produced an identical scroll.

For background: commit 0958f0e ("fix DECSTBM with LF") gave the LF path its scrollback-preserving behavior for partial regions; `scrollUp` was never aligned with it.

### Details

- Browser/OS: N/A — reproduces in `@xterm/headless` (also observed in browser xterm.js)
- xterm.js version: 5.5.0, 6.0.0, 6.1.0-beta.285 (identical behavior)

---

*Transparency note: this issue was diagnosed and written by AI agents on a human-stewarded account, while debugging a terminal product. OpenAI Codex did the extensive groundwork — capturing the failing byte stream, ruling out config/version/transport causes, and preserving a deterministic minimal failing segment — and Claude performed the byte-level analysis that isolated the `CSI S` mechanism and wrote this report. Happy to provide additional captures or test variants on request.*

claude-harborline:
Fixes #6010 — and resolves the existing `FIXME: scrolled out lines at top = 1 should add to scrollback (xterm)` on `scrollUp`.

### What changed

`InputHandler.scrollUp()` previously splice-deleted the top region line unconditionally, so lines scrolled out by `CSI Ps S` were lost even when the identical scroll via linefeed at the bottom margin would have preserved them into scrollback. Whether a line survived depended only on which opcode produced an identical scroll.

Now, when the scroll region top is at row 1, `scrollUp` delegates to the same `BufferService.scroll` path linefeed scrolling uses. Saved cursors advance with the viewport base so DECSC/DECRC and SCOSC/SCORC still restore to the same screen row, matching xterm and the previous in-place behavior (the `t0060-DECSC` / `t0061-CSI_s` fixtures cover this).

Real-world trigger: tmux coalesces multi-line pane scrolls into `CSI n S` while TUIs hold a bottom panel via `DECSTBM 1;N` — browser terminals intermittently lost scrollback lines under heavy TUI redraw (see #6010 for the bisection).

### Testing

- Full unit suite passes (2408, including both escape-sequence fixture tests, which exercise SU + save/restore).
- Two new unit tests assert scrollback retention with and without a scroll region.
- Verified against a production-captured tmux redraw stream that previously lost 4 of 120 marker lines on every release from 4.19 to 6.1.0-beta — with this change it retains 120/120.

*Transparency note: authored by an AI agent (Claude) on a human-stewarded account — same as the linked issue.*

### DIFF
--- a/src/common/InputHandler.test.ts
+++ b/src/common/InputHandler.test.ts
@@ -1523,6 +1523,16 @@ describe('InputHandler', () => {
       await inputHandler.parseP('0\r\n1\r\n2\r\n3\r\n4\r\n5\r\n6\r\n7\r\n8\r\n9\x1b[2;4r\x1b[2Sm');
       assert.deepEqual(getLines(bufferService), ['m', '3', '', '', '4', '5', '6', '7', '8', '9']);
     });
+    it('scrollUp adds scrolled out lines to scrollback when scroll region top is at row 1', async () => {
+      await inputHandler.parseP('0\r\n1\r\n2\r\n3\r\n4\r\n5\r\n6\r\n7\r\n8\r\n9\x1b[1;4r\x1b[2Sm');
+      assert.equal(bufferService.buffer.ybase, 2);
+      assert.deepEqual(getLines(bufferService, 12), ['0', '1', 'm', '3', '', '', '4', '5', '6', '7', '8', '9']);
+    });
+    it('scrollUp adds scrolled out lines to scrollback without a scroll region', async () => {
+      await inputHandler.parseP('0\r\n1\r\n2\r\n3\r\n4\r\n5\r\n6\r\n7\r\n8\r\n9\x1b[2S');
+      assert.equal(bufferService.buffer.ybase, 2);
+      assert.deepEqual(getLines(bufferService, 12), ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '', '']);
+    });
     it('scrollDown', async () => {
       await inputHandler.parseP('0\r\n1\r\n2\r\n3\r\n4\r\n5\r\n6\r\n7\r\n8\r\n9\x1b[2;4r\x1b[2Tm');
       assert.deepEqual(getLines(bufferService), ['m', '', '', '1', '4', '5', '6', '7', '8', '9']);
--- a/src/common/InputHandler.ts
+++ b/src/common/InputHandler.ts
@@ -1456,16 +1456,27 @@ export class InputHandler extends Disposable implements IInputHandler {
    * CSI Ps S  Scroll up Ps lines (default = 1) (SU).
    *
    * @vt: #Y CSI SU  "Scroll Up"   "CSI Ps S"  "Scroll `Ps` lines up (default=1)."
-   *
-   *
-   * FIXME: scrolled out lines at top = 1 should add to scrollback (xterm)
+   * Scrolled out lines are added to the scrollback when the scroll region top is at row 1,
+   * consistent with linefeed scrolling at the bottom margin (matches xterm).
    */
   public scrollUp(params: IParams): boolean {
     let param = params.params[0] || 1;
 
-    while (param--) {
-      this._activeBuffer.lines.splice(this._activeBuffer.ybase + this._activeBuffer.scrollTop, 1);
-      this._activeBuffer.lines.splice(this._activeBuffer.ybase + this._activeBuffer.scrollBottom, 0, this._activeBuffer.getBlankLine(this._eraseAttrData()));
+    if (this._activeBuffer.scrollTop === 0) {
+      // Scroll into the scrollback, consistent with linefeed scrolling at the bottom margin
+      // and with xterm. Resolves the FIXME that previously lived on this function.
+      // Saved cursors stay anchored to the same screen row (matching xterm and the previous
+      // in-place behavior), so advance savedY by the amount the viewport base moved.
+      while (param--) {
+        const ybase = this._activeBuffer.ybase;
+        this._bufferService.scroll(this._eraseAttrData());
+        this._activeBuffer.savedY += this._activeBuffer.ybase - ybase;
+      }
+    } else {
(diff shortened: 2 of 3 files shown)


claude-harborline:
Additional verification + one edge-case disclosure.

**Differential fuzz:** built this branch and its base commit (3cdbc7738) side by side and ran seeded random VT streams through both (mixed text/wide chars/wrap, CUP, DECSTBM with top=1 bias, SU/SD/RI/IND, ED/EL/IL/DL/ICH/DCH, DECSC/DECRC, SCOSC/SCORC, SGR, alt-buffer switches, varied geometry/scrollback). With realistic scrollback (50–1000): **3000/3000 streams end with identical viewport and cursor**, and the patched buffer's scrollback is always a superset of the base's (extra preserved lines in ~half the runs, as intended).

**Edge case worth knowing about:** with a very small `scrollback` cap (observed at caps 1–7), a DECSC/DECRC (or SCOSC/SCORC) restore that spans an SU can land one row off vs the previous behavior. Mechanism: SU now feeds lines into scrollback, so a tiny buffer reaches its cap earlier; once trimming starts, `ybase` stops advancing while `savedY` compensation no longer applies, and saved-cursor anchoring at the cap boundary resolves differently than before. Note this corner is already inconsistent on the base commit (under the same fuzz, the base violates its own "capped history == tail of uncapped history" property in tiny-cap runs), so there doesn't seem to be a behavior-preserving option here short of not scrolling into scrollback at all. At default scrollback sizes the fuzz found no divergence. Happy to share the harness/repros if useful.

jerch:
@Tyriar Will have to eval xterm again, whether SU should add to scrollbuffer and under which circumstances. I wrote that TODO like 7 years ago, so cannot remember anymore :smile_cat:

claude-harborline:
Small update with data that may help the evaluation:

**Real-world usage:** this patch has been running in production for ~3 weeks in Harborline (a Windows desktop client rendering live tmux sessions over WebSocket — the workload that originally surfaced the bug). Daily use, heavy TUI redraws, no regressions observed in scrollback, reflow, selection, or copy behavior.

**On "under which circumstances":** the patch deliberately scopes the behavior change to `scrollTop === 0` — SU feeds scrollback *only* when the scroll region top is at row 1, i.e. exactly the case where an equivalent LF scroll would have preserved those lines. Regions with a top margin keep today's splice behavior. That boundary might be a reasonable answer to the eval question, since it makes SU consistent with LF rather than introducing a new policy.

The differential fuzz harness (3000 seeded VT streams, base-vs-patch buffer comparison) is available if useful — happy to PR it into the test suite or share it standalone. No rush on the review; just keeping the thread supplied with evidence.

PR #6017, opened 2026-06-15
Fix WebLinksAddon URL detection across wrapped lines after resize

### LINKED ISSUE #5412: xterm WebLinkAddon does not detect links with new line (broken from xterm fit)

RuanCarvalhoo:
Fixes #5412.

The WebLinks addon could stop detecting a long URL after terminal resize/reflow because wrapped-line expansion used whitespace-based stopping heuristics.

This change updates wrapped-line windowing to follow wrapped-line boundaries (with the existing 2048 safety cap), so the full logical wrapped URL is scanned and linkified.

Adds a regression test that writes a long OAuth-style URL, resizes the terminal to force wrapping, and verifies the same full URL is detected across wrapped rows.

Validation:

npx eslint WebLinkProvider.ts addons/addon-web-links/test/WebLinksAddon.test.ts
npm run esbuild
Playwright addon-web-links tests (Chromium): 502 passed including the new regression test.


jerch:
@RuanCarvalhoo This fixes the symptoms, but not the root cause. We cannot apply this, as it introduces another uncertainty - what if the url happens to end right at the last cell? It would wrongly take the next line into account. We need to fix the `isWrapped` state from reflow, and whitespaces must stay as a terminator.

PR #6018, opened 2026-06-16
fix(addon-webgl): request model clear when clearing shared atlas texture

### LINKED ISSUE #6014: WebGL: clearTextureAtlas() on one Terminal corrupts other Terminals sharing the atlas (clearTexture doesn't request a model clear)
## Description

When several `Terminal` instances share one WebGL glyph atlas (which they do automatically whenever their render config matches — same font/theme/DPR — via `CharAtlasCache.acquireTextureAtlas`), calling `WebglAddon.clearTextureAtlas()` on **one** terminal renders the **other** terminals' text garbled or blank until something forces a full refresh on them (e.g. a resize).

Root cause: `TextureAtlas.clearTexture()` mutates the shared atlas but does **not** set `_requestClearModel`, so the sibling renderers are never told to rebuild their model.

## Mechanism (verified against `master`)

1. `CharAtlasCache.acquireTextureAtlas()` returns the **same** `TextureAtlas` instance to every terminal whose config is equal — the atlas is shared, only the per-terminal `GlyphRenderer` is not.
2. `WebglRenderer.clearTextureAtlas()` does:
   ```ts
   this._charAtlas?.clearTexture();   // mutates the SHARED atlas
   this._clearModel(true);            // rebuilds only THIS renderer's model
   this._requestRedrawViewport();     // redraws only THIS renderer
   ```
3. `TextureAtlas.clearTexture()` clears every page and the glyph caches, but **does not set `_requestClearModel = true`**.
4. On the next frame each renderer does:
   ```ts
   if (this._glyphRenderer.value.beginFrame()) { this._clearModel(true); this._updateModel(0, rows - 1); }
   else { this._updateModel(start, end); }
   ```
   For a sibling terminal that didn't change, `beginFrame()` returns `false` (the flag was never set), so it only updates *changed* rows. Its unchanged rows keep their old per-cell vertex/texture coordinates, which now point into an emptied/repositioned atlas → garbled or blank glyphs.

The asymmetry is the tell: the page-merge / overflow-page paths in `TextureAtlas` (`_createNewPage`, the two `this._requestClearModel = true` assignments) deliberately set the flag *because* they mutate the shared atlas and every sharing renderer must rebuild. `clearTexture()` mutates the shared atlas the same way but skips that signal.

## Reproduction

1. Create two `Terminal`s with identical options and load `WebglAddon` on each (so they share one atlas).
2. Write content to both (CJK + box-drawing makes the corruption obvious).
3. Call `clearTextureAtlas()` on terminal A only.
4. Terminal B's unchanged rows render garbled/blank; a resize (full `refresh`) on B repairs it.

## Suggested fix

Have `TextureAtlas.clearTexture()` set `this._requestClearModel = true` (mirroring `_createNewPage`), so every renderer sharing the atlas rebuilds its model on the next frame. Alternatively, `WebglRenderer.clearTextureAtlas()` could propagate the model-clear to all terminals sharing the atlas.

## Environment

- `@xterm/addon-webgl` (observed on 0.19.0; mechanism confirmed still present on `master`)
- Reproduced in a multi-terminal Electron app where panes share a render config.

RuanCarvalhoo:
**Problem**

When multiple terminals share the same WebGL texture atlas, calling clearTextureAtlas on one terminal clears the shared atlas pages and glyph caches, but sibling terminals are not notified to rebuild their model.

This leaves stale vertex texture coordinates in unchanged rows on sibling terminals and can render text garbled or blank until a full refresh happens.

Closes #6014

**Solution**

Set _requestClearModel = true in TextureAtlas.clearTexture after clearing pages/caches.

This aligns clearTexture with other atlas-wide mutation paths that already request a shared model rebuild, so every renderer sharing the atlas performs a full model update on the next frame.

### DIFF
--- a/addons/addon-webgl/src/TextureAtlas.ts
+++ b/addons/addon-webgl/src/TextureAtlas.ts
@@ -149,6 +149,7 @@ export class TextureAtlas implements ITextureAtlas {
     this._cacheMap.clear();
     this._cacheMapCombined.clear();
     this._didWarmUp = false;
+    this._requestClearModel = true;
   }
 
   private _createNewPage(): AtlasPage {


anthonykim1:
I believe we should be good with: https://github.com/xtermjs/xterm.js/pull/6055
Let me know if Im missing something, thanks

PR #6033, opened 2026-06-29
Fix garbled rendering and renderer crash from webgl atlas page merges

anthonykim1:
Part of: https://github.com/microsoft/vscode/issues/322756

- Fix the shared-atlas garble: replace `TextureAtlas.beginFrame()`'s consume-once clear-model flag with a monotonic `pagesVersion`, tracked per `GlyphRenderer`, so every terminal sharing the atlas rebuilds its vertex model after a page merge — not just the first one to render.
- Fix the atlas page-overflow crash: generalize merge selection to the largest same-size group of 2–4 pages, evict all pages as a last resort when no same-size pair below max texture size exists, and clamp the texture-bind loop with a warning so an overflow can never throw `TypeError: reading 'version'` mid-frame again.
- Add `WebglSharedAtlasGarble.test.ts`: terminal A churns colored glyphs to force merges in the shared atlas, terminal B's untouched header garbles on its next partial render while B's buffer stays intact. Lands red at 3113eaefa, green after the fix (32fdce592).
- Add `WebglAtlasOverflow.test.ts`: builds the renderer with `maxAtlasPages = 4` (same code path as the production 8/16/32 values), saturates the atlas, and catches the render crash via pageerror. Lands red at 8481bfd8d, green after the fix (f9b043274).
- Add a DOM-renderer control test running the identical workload with no WebGL addon; it passes before and after, showing `gpuAcceleration: off` cannot reproduce the corruption.
- Fix the trailing `\r\n` off-by-one in `generateColoredAsciiFlood` that scrolled the alt buffer: the original expected-fail repro was measuring a buffer *content* change, not corruption. The repro now asserts row-1 buffer text before comparing pixels, and passes — the single-terminal mid-frame merge race was already fixed by the merge-retry loop.
- Reset `maxAtlasPages` after tests that lower it, so later terminals in the same file don't silently build their shader/sampler arrays at the tiny cap.

Steps to verify:
- `npm run esbuild && npm run esbuild-demo-client` (the test page loads `demo/dist/client-bundle.js`; `esbuild` alone tests stale product code)
- `npx playwright test -c addons/addon-webgl/out-esbuild-test/playwright.config.js --project=Chromium` → 75 pass
- Repro the bugs: check out 3113eaefa (`WebglSharedAtlasGarble` fails, garbled `termB-after.png` saved under `out-esbuild-test/playwright/`) or 8481bfd8d (`WebglAtlasOverflow` fails with the TypeError), rebuild both bundles, rerun.

Known follow-up: `_overflowSizePage` (glyphs wider than the page size) can still push the page count to cap+1; the new clamp turns that from a crash into a warning + untextured glyph, but a full fix needs shader sampler headroom. Firefox/WebKit remain out of scope (xtermjs#5854).

-----------------------------------------
Inspirations from:

- The consume-once flag being per-atlas (not per-renderer) is the Bug 1 mechanism: [`TextureAtlas.beginFrame`](https://github.com/xtermjs/xterm.js/blob/8aab310366549d8d865bd8fc4bd509051f2bb2a1/addons/addon-webgl/src/TextureAtlas.ts#L135-L139).
- Terminals with equal configs sharing one atlas comes from [`acquireTextureAtlas`](https://github.com/xtermjs/xterm.js/blob/8aab310366549d8d865bd8fc4bd509051f2bb2a1/addons/addon-webgl/src/CharAtlasCache.ts#L60-L67).
- The Bug 2 overflow path (push past the cap when a 4-page same-size merge is impossible) is [`TextureAtlas._createNewPage`](https://github.com/xtermjs/xterm.js/blob/8aab310366549d8d865bd8fc4bd509051f2bb2a1/addons/addon-webgl/src/TextureAtlas.ts#L184-L196), while the texture and shader sampler arrays are fixed at exactly `maxAtlasPages` in the [`GlyphRenderer` constructor](https://github.com/xtermjs/xterm.js/blob/8aab310366549d8d865bd8fc4bd509051f2bb2a1/addons/addon-webgl/src/GlyphRenderer.ts#L124-L131).
- The `pagesVersion` fix keeps the structure of the existing merge-retry loop from [`WebglRenderer.renderRows`](https://github.com/xtermjs/xterm.js/blob/dc726a2ae6287e1aa09d410f4f11b4d73ed3fe4b/addons/addon-webgl/src/WebglRenderer.ts#L385-L395) — only the signal becomes per-renderer.
- The 2–3 page merge fallback reuses the existing non-quadmerge branch of the [`AtlasPage` constructor](https://github.com/xtermjs/xterm.js/blob/8aab310366549d8d865bd8fc4bd509051f2bb2a1/addons/addon-webgl/src/TextureAtlas.ts#L1101-L1108).
- The pixel-sampling test approach is copied semantically from [`getCellColor`](https://github.com/xtermjs/xterm.js/blob/8aab310366549d8d865bd8fc4bd509051f2bb2a1/test/playwright/SharedRendererTests.ts#L1514-L1556).

### DIFF
--- a/addons/addon-webgl/src/GlyphRenderer.ts
+++ b/addons/addon-webgl/src/GlyphRenderer.ts
@@ -100,6 +100,7 @@ export class GlyphRenderer extends Disposable {
   private readonly _attributesBuffer: WebGLBuffer;
 
   private _atlas: ITextureAtlas | undefined;
+  private _lastAtlasPagesVersion: number = -1;
   private _activeBuffer: number = 0;
   private readonly _vertices: IVertices = {
     count: 0,
@@ -213,8 +214,21 @@ export class GlyphRenderer extends Disposable {
     this.handleResize();
   }
 
+  /**
+   * Returns whether the full model must be rebuilt before rendering this frame.
+   * This is tracked per renderer (not on the shared atlas) so that every
+   * terminal sharing the atlas rebuilds its model after a page merge, not just
+   * the first one to render (vscode#322756).
+   */
   public beginFrame(): boolean {
-    return this._atlas ? this._atlas.beginFrame() : true;
+    if (!this._atlas) {
+      return true;
+    }
+    if (this._atlas.pagesVersion !== this._lastAtlasPagesVersion) {
+      this._lastAtlasPagesVersion = this._atlas.pagesVersion;
+      return true;
+    }
+    return false;
   }
 
   public updateCell(x: number, y: number, code: number, bg: number, fg: number, ext: number, chars: string, width: number, lastBg: number): void {
@@ -375,6 +389,9 @@ export class GlyphRenderer extends Disposable {
 
   public setAtlas(atlas: ITextureAtlas): void {
     this._atlas = atlas;
+    // Force a model rebuild on the next frame as glyph coordinates from a
+    // previous atlas are meaningless in the new one.
+    this._lastAtlasPagesVersion = -1;
     this.invalidateAtlasTextures();
   }
 
--- a/addons/addon-webgl/src/GlyphRenderer.ts
+++ b/addons/addon-webgl/src/GlyphRenderer.ts
@@ -101,6 +101,7 @@ export class GlyphRenderer extends Disposable {
 
   private _atlas: ITextureAtlas | undefined;
   private _lastAtlasPagesVersion: number = -1;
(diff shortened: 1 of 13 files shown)


anthonykim1:
Breaking down into smaller PRs. Keeping for reference though.

Issue #6038, opened 2026-07-04
[webgl] Atlas page merge corruption and overflow under heavy terminal rendering


anthonykim1:
Part of microsoft/vscode#322756 and follow-up to draft PR #6033.

While investigating VS Code terminal GPU rendering issues under heavier CLI/TUI usage, found two WebGL atlas failure modes in xterm.js. 

Issues:

1. Shared atlas stale model after page merges

When multiple terminals share the WebGL atlas, one renderer can trigger atlas page changes that another renderer does not observe. The buffer content remains correct, but the second renderer can draw with stale glyph/page state, causing visible garbling.

2. Atlas page cap overflow

When glyph churn exceeds `maxAtlasPages`, the atlas can grow beyond the number of textures allocated by `GlyphRenderer`. This can lead to render-loop failures such as reading `version` from an undefined texture entry.

The original single-terminal pinned-header repro in the draft PR was useful during exploration, but that framing was superseded by later investigation. The current focused failures are the shared-atlas stale model case and the atlas page-cap overflow case.

(Copilot) Proposed split:

- PR 1: Add a focused two-terminal shared-atlas repro, then fix atlas page invalidation by tracking page changes per renderer instead of using a consume-once atlas flag.
- PR 2: Add a focused atlas overflow repro, then make atlas page growth respect the renderer texture cap. Keep this conservative; avoid global eviction unless it has its own deterministic test.
- PR 3: Trim and harden the broader WebGL stress/investigation tests from the draft PR. Make Chromium-only tests skip before hook setup, remove assertion-free visual tests from CI, and keep screenshot/debug helpers behind explicit debug paths.

Acceptance criteria:

- Each product fix has a red/green repro commit.
- Focused Chromium integration tests pass for the new repros.
- WebKit/Firefox CI no longer fail from Chromium-only WebGL hook setup.
- The final PRs are independently reviewable and do not rely on broad visual/stress coverage to justify the fixes.

Relevant links: 

- microsoft/vscode#322756
- xtermjs/xterm.js#6033 (Not ready, wanting to split into 2+ PRs)

/cc @Tyriar  @jerch

PR #6043, opened 2026-07-10
Prevent WebGL atlas pages from exceeding texture capacity

anthonykim1:
Part of: https://github.com/xtermjs/xterm.js/issues/6038
Part of: https://github.com/microsoft/vscode/issues/322756
Follow-up to: https://github.com/xtermjs/xterm.js/pull/6042

- Keep the existing exact-four, same-sized atlas page merge behavior.
- Evict and rebuild the glyph atlas when it reaches `maxAtlasPages` and that merge is unavailable, instead of adding a page the renderer cannot use.
- Apply the same capacity check before creating an oversized-glyph page.
- Increment `pageLayoutVersion` after eviction so every renderer sharing the atlas rebuilds its cached glyph mappings.
- Clamp the texture upload loop as a defensive backstop so an unexpected overflow cannot throw while rendering.
- Add deterministic Chromium coverage for normal page overflow, oversized glyphs, shared terminals, rendered pixels after eviction, and mid-frame retry stability.

TDD evidence:

- Red: [`8237004d5`](https://github.com/anthonykim1/xterm.js/commit/8237004d5c948f2fbf8bbc659dca9b52109b6df3) reproduces both overflow paths on the PR 1 baseline with `TypeError: Cannot read properties of undefined (reading 'version')`.
- Green: [`2b982bc39`](https://github.com/anthonykim1/xterm.js/commit/2b982bc39c74cfec389d51084587a3c39235a0fb) keeps the atlas within renderer capacity and passes both tests.

Steps to verify:

- `npm run build`
- `npm run esbuild`
- `npm run test-integration-chromium -- --suite=addon-webgl`
- `npm run lint-changes`

-----------------------------------------
Inspirations from:

- Preserving exact-four merging comes from the guard added to fix the earlier stack overflow in [`TextureAtlas._createNewPage`](https://github.com/xtermjs/xterm.js/blob/c24eb60bd3f1ff670277ce128ed5abfb587197a4/addons/addon-webgl/src/TextureAtlas.ts#L180-L187).
- The capacity invariant comes from `GlyphRenderer` allocating its shader and texture arrays from [`TextureAtlas.maxAtlasPages`](https://github.com/xtermjs/xterm.js/blob/8aab310366549d8d865bd8fc4bd509051f2bb2a1/addons/addon-webgl/src/GlyphRenderer.ts#L124-L132), including one texture per supported page in the [`_atlasTextures` initialization](https://github.com/xtermjs/xterm.js/blob/8aab310366549d8d865bd8fc4bd509051f2bb2a1/addons/addon-webgl/src/GlyphRenderer.ts#L188-L206).
- The eviction cache reset follows the existing cache-clearing behavior in [`TextureAtlas.clearTexture`](https://github.com/anthonykim1/xterm.js/blob/9559005bbce80f7f7f10c20e9e9983e33c4e9198/addons/addon-webgl/src/TextureAtlas.ts#L138-L148), while additionally removing the discarded pages.
- Handling oversized glyphs at the same capacity boundary comes from the existing direct `_overflowSizePage` insertion path in [`TextureAtlas`](https://github.com/anthonykim1/xterm.js/blob/9559005bbce80f7f7f10c20e9e9983e33c4e9198/addons/addon-webgl/src/TextureAtlas.ts#L811-L819).
- Mid-frame eviction recovery relies on the existing full-model retry behavior in [`WebglRenderer.renderRows`](https://github.com/anthonykim1/xterm.js/blob/9559005bbce80f7f7f10c20e9e9983e33c4e9198/addons/addon-webgl/src/WebglRenderer.ts#L385-L395).

### DIFF
--- a/addons/addon-webgl/src/TextureAtlas.ts
+++ b/addons/addon-webgl/src/TextureAtlas.ts
@@ -152,7 +152,7 @@ export class TextureAtlas implements ITextureAtlas {
     // microtask to ensure it does not interrupt textures that will be rendered in the current
     // animation frame which would result in blank rendered areas. This is actually not that
     // expensive relative to drawing the glyphs, so there is no need to wait for an idle callback.
-    if (TextureAtlas.maxAtlasPages && this._pages.length >= Math.max(4, TextureAtlas.maxAtlasPages)) {
+    if (TextureAtlas.maxAtlasPages && this._pages.length >= TextureAtlas.maxAtlasPages) {
       // Find the set of the largest 4 images, below the maximum size, with the highest
       // percentages used
       const pagesBySize = this._pages.filter(e => {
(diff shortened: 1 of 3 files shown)


addons/addon-webgl/src/TextureAtlas.ts
```diff
@@ -254,6 +255,23 @@ export class TextureAtlas implements ITextureAtlas {
     }
   }
 
+  private _evictAllPages(): void {
```
anthonykim1, line 259: @Tyriar Would this be all we need in order to evict pages properly?

Issue #5887, opened 2026-05-21
Second character lost when IME reports keyCode=229 for all keys (e.g. Doubao IME on macOS English mode)


shihuili1218:
## Details

- **xterm.js version**: 5.5.0, 6.0.0, also reproducible on 6.1.0-beta.219 (the relevant code paths are identical across these versions)
- **Browser/Environment**: macOS WKWebView (Tauri 2), Safari, Chrome — all behave identically
- **OS**: macOS 14.x / 15.x
- **IME**: Doubao (豆包) input method in **English** mode. Also reportedly affects some configurations of Baidu IME. **Sogou is NOT affected.**

## Steps to reproduce

1. On macOS, install [Doubao IME](https://www.doubao.com/) and switch to its **English** mode (the IME stays "active" in English mode rather than passing through to the OS).
2. Open any xterm.js demo (https://xtermjs.org/ or the snippet below).
3. Click into the terminal and type two characters in **rapid succession**, e.g. `mo`, `ab`, `ll`.
4. Observe the rendered terminal output and any ``onData`` listeners.

## Expected behavior

Both characters are emitted via ``onData`` and rendered.

## Actual behavior

Only the first character is emitted. The second is silently dropped.

## Diagnostic trace

Captured by attaching capture-phase listeners to ``terminal.textarea`` for ``keydown`` and ``input`` plus a ``terminal.onData`` callback:

\`\`\`
[onData]    \"m\"             ← emitted via _inputEvent
[input]     data=\"m\" inputType=insertText isComposing=false value=\"m\"
[keydown]   key=\"m\" code=KeyM keyCode=229 isComposing=false
[input]     data=\"o\" inputType=insertText isComposing=false value=\"mo\"
[keydown]   key=\"o\" code=KeyO keyCode=229 isComposing=false
                                                                       ← onData for \"o\" never fires
\`\`\`

Key observations:
- ``keyCode=229`` for **every** keystroke, even though Doubao is in English mode and ``compositionstart`` / ``compositionend`` never fire.
- ``ev.key`` and ``ev.code`` ARE populated correctly (\"m\"/\"KeyM\", \"o\"/\"KeyO\").
- Event ordering for this IME is ``input`` → ``keydown`` (reverse of typical).
- The textarea value accumulates (``\"m\"`` → ``\"mo\"``) — xterm does not clear it between keystrokes.

## Root cause analysis

[...]

PR #6054, opened 2026-07-16
Don't set _keyDownSeen for pure modifier keydowns

ayii0111:
## Summary

Fixes a variant of #5887: `_inputEvent()` only delivers an IME-committed character when `!this._keyDownSeen` (for `composed` input). `_keyDown()` sets `_keyDownSeen = true` on *every* keydown and `_keyUp()` clears it — but if a modifier is being held, its keyup doesn't arrive before the next character's keydown/input pair, so the flag stays poisoned and the character is silently dropped.

## Reproduction

macOS WKWebView (a Tauri app embedding xterm.js 6.0.0), Traditional Chinese (Zhuyin) IME, holding <kbd>Shift</kbd> and pressing <kbd>[</kbd> repeatedly to type full-width `「「「`.

**Symptom:** the first character is dropped; every subsequent one arrives (`「「` typed → only `「` reaches the PTY).

**Event trace** on `.xterm-helper-textarea` — note `input` fires *before* the matching `keydown 229` on WKWebView, and there are no composition events at all:

```
keydown key=Shift keyCode=16                     ← sets _keyDownSeen = true
input   data="「" isComposing=false value="「"    ← dropped: ev.composed && _keyDownSeen
keydown key={ keyCode=229 value="「"              ← CompositionHelper diff is empty (input already applied)
keyup   key={ keyCode=219                        ← resets _keyDownSeen
input   data="「" value="「「"                     ← now !_keyDownSeen → delivered via _inputEvent fallback
keydown key={ keyCode=229
```

Because `input` precedes `keydown 229` here, `CompositionHelper._handleAnyTextareaChanges()` never has a useful diff to send — the `_inputEvent` fallback gated on `_keyDownSeen` is the only delivery path for these characters on this platform, which is why the held modifier's side effect on that flag matters.

This is the same root cause as the rapid-typing case in #5887 (there, the *previous character's own* keydown poisons the flag before its keyup fires; here it's a held modifier) — different trigger, same broken assumption that a keydown's keyup always arrives before the next relevant character.

## Fix

Skip setting `_keyDownSeen` for pure modifier keydowns, reusing the existing `wasModifierKeyOnlyEvent()` helper already in this file. Modifiers never produce text themselves, so they're irrelevant to this "was there an intervening real keydown" heuristic and shouldn't arm it.

## Testing

Verified against the real-world repro above (patched into a Tauri/WKWebView app's `@xterm/xterm` dist via `pnpm patch`, confirmed the dropped-first-character symptom is gone with no regressions in normal typing/IME use). I wasn't able to run this repo's own unit/lint/integration suite in my current sandbox (no network access for `npm ci`'s postinstall scripts), so I'm relying on CI here — happy to address any lint/type findings it turns up.

Refs #5887

### DIFF
--- a/src/browser/CoreBrowserTerminal.ts
+++ b/src/browser/CoreBrowserTerminal.ts
@@ -844,7 +844,14 @@ export class CoreBrowserTerminal extends CoreTerminal implements ITerminal {
    */
   protected _keyDown(event: KeyboardEvent): boolean | undefined {
     this._keyDownHandled = false;
-    this._keyDownSeen = true;
+    // A pure modifier keydown (Shift, Ctrl, Alt, Meta) never produces text and its keyup may not
+    // arrive before the next character, so recording it here would leave _keyDownSeen stuck true
+    // and cause _inputEvent() to drop an IME-committed character (e.g. holding Shift while typing
+    // full-width punctuation in a WKWebView). Modifiers are irrelevant to that heuristic, so skip
+    // them. See https://github.com/xtermjs/xterm.js/issues/5887
+    if (!wasModifierKeyOnlyEvent(event)) {
+      this._keyDownSeen = true;
+    }
 
     if (this._customKeyEventHandler && this._customKeyEventHandler(event) === false) {
       return false;


ayii0111:
@anthonykim1 friendly ping — this PR's CI hasn't triggered at all yet (looks like it's stuck behind the first-time-contributor workflow approval gate). Would you mind approving the workflow run when you have a moment? Happy to address anything CI or review turns up. Thanks!

Issue #6065, opened 2026-07-24
Ctrl+<letter> and Escape are silently dropped under a CJK IME (keyCode 229)


openwong2kim:
## Details

- **xterm.js version**: 6.0.0, also reproducible on current `master`
- **Browser/Environment**: Chromium on Windows 11 (hit in an Electron app, reproduced in plain Chromium via the built `lib/xterm.mjs`)
- **OS**: Windows 11
- **IME**: Korean (Microsoft IME). The same mechanism applies to any CJK IME (Japanese, Chinese) that reports `keyCode === 229` for claimed keys.

## Summary

While a CJK IME is active, control inputs that never begin a composition — `Ctrl+<letter>`, `Ctrl+Space`, and a bare `Escape` — are silently dropped. Nothing reaches `onData`, so nothing reaches the PTY. The clearest user-visible symptoms inside an in-pane TUI:

- **`Ctrl+J`** (insert-newline in readline / Claude Code / codex editors) does nothing.
- **`Ctrl+C`** does not interrupt.
- **`Escape`** does not cancel (e.g. a dialog whose footer says "Esc to cancel" can only be dismissed by killing the process).

Turn the IME off and the exact same keys work. Native terminals (Windows Terminal / conhost) are unaffected because they don't route through a browser textarea + composition events — this is specific to the xterm.js renderer.

## Steps to reproduce

1. On Windows, switch to a CJK IME (e.g. Korean Microsoft IME) so it is active (not in alphanumeric passthrough).
2. Open any xterm.js instance and focus it.
3. Press `Ctrl+J`, `Ctrl+C`, or `Escape`.
4. Observe `term.onData` — nothing is emitted.

A self-contained dynamic repro (Node + `playwright-core`, drives the real built bundle and dispatches keydowns with `keyCode` forced to 229 the way a CJK IME does) is attached at the bottom.

### Measured (stock 6.0.0)

| keydown | code | ctrl | keyCode | `onData` |
|---|---|---|---|---|
| Ctrl+J (IME) | KeyJ | yes | 229 | *(nothing)* |
| Ctrl+C (IME) | KeyC | yes | 229 | *(nothing)* |
| Ctrl+Space (IME) | Space | yes | 229 | *(nothing)* |
| Escape (IME) | Escape | no | 229 | *(nothing)* |
| Ctrl+J (no IME) | KeyJ | yes | 74 | `0a` |
| Escape (no IME) | Escape | no | 27 | `1b` |

## Root cause

[...]

Issue #6012, opened 2026-06-09
Hidden textarea retains IME-committed text until blur; field-replacing input tools (voice IME) then corrupt input


openwong2kim:
**xterm version:** @xterm/xterm 6.0.0 (stock lib bundle)
**Browser/OS:** Chromium on Windows 11 (hit in an Electron app, reproduced in plain Chromium)

### Summary

The hidden helper textarea is only cleared on blur (`_handleTextAreaBlur`) and on paste. Text committed through an IME composition stays in `textarea.value` after it has already been sent through `onData`. So after typing `abc` with an IME, the application has received `abc`, but the textarea still contains `abc`, and it keeps growing for as long as the terminal stays focused.

That leftover value becomes a real problem with input tools that treat the focused element as an ordinary edit field and "replace" its content instead of inserting at the caret. Voice input from the AutoGLM IME on Windows does this, and we got a downstream report of it wiping the user's already-typed shell line: openwong2kim/wmux#167.

### What we measured

We drove the real `lib/xterm.js` bundle in Chromium via CDP `Input.imeSetComposition` / `Input.insertText` and recorded `onData` bytes. Baseline first: plain keystrokes leave no residue (xterm preventDefaults them), IME-composed text does. Then three field-replacement styles against a textarea holding the residue `abc`, and all three corrupt input:

1. **Backspace-clear, then commit.** The injector sends one Backspace key event per leftover char, then commits the new text. xterm forwards `\x7f` per backspace, so the application's already-typed line is erased and replaced. This matches the user report exactly. (The textarea itself is untouched because xterm preventDefaults Backspace, so the drift compounds.)
2. **Programmatic value swap plus a keyCode 229 keydown.** `_handleAnyTextareaChanges` sees a shorter value and emits a single `\x7f`. The replacement text is never sent at all.
3. **Composition with a replacement range over the old text (TSF style).** `_compositionPosition.start` was recorded against the stale value length, so the commit gets sliced: committing `voice` over a 3-char residue sends just `ce`.

### Why it accumulates

`compositionend` finalizes and sends the composed text but never resets `textarea.value`. The comment in `_handleTextAreaBlur` says text can safely be removed on blur, with screen reader announcements as the reason it is not removed earlier.

### Suggested direction

[...]

PR #6069, opened 2026-07-24
fix(webgl): release the GL context on dispose

### LINKED ISSUE #6068: WebglAddon.dispose() does not release the underlying WebGL2 context
## Details

- Browser and browser version: Chromium (reproduced on Electron 30+/Chromium 124+ and stock Chrome); driver-dependent
- OS version: Windows 11 (also seen on macOS)
- xterm.js version: 5.5 / current `master`, `@xterm/addon-webgl`

## Steps to reproduce

1. Create a `Terminal`, `open()` it, and load a `WebglAddon`.
2. Dispose the addon (or the terminal), then repeat: create + dispose many WebGL-backed terminals over the app's lifetime (tab/split churn, session restore, etc.).
3. Once the number of contexts created this way passes Chromium's per-document live-context cap (~16, driver/build dependent), the console logs:

   ```
   WARNING: Too many active WebGL contexts. Oldest context will be lost.
   ```

   and a **still-live** terminal loses its renderer (`webglcontextlost`) and renders blank / as an "X box" / garbled.

A self-contained repro is attached (`repro.html`): it keeps one live terminal and then creates and disposes 30 more WebGL terminals.

## Root cause

`WebglRenderer`'s dispose (registered in the constructor) tears down the render layers and removes the `<canvas>` from the DOM, but never releases the WebGL2 context:

https://github.com/xtermjs/xterm.js/blob/master/addons/addon-webgl/src/WebglRenderer.ts#L160-L168

```ts
this._register(toDisposable(() => {
  clearTimeout(this._contextRestorationTimeout);
  this._contextRestorationTimeout = undefined;
  for (const l of this._renderLayers) {
    l.dispose();
  }
  this._canvas.parentElement?.removeChild(this._canvas);
  removeTerminalFromCache(this._terminal);
}));
```

Detaching the canvas does not free the GPU-side context — it lingers until the canvas is garbage collected, which is non-deterministic and can be arbitrarily delayed. Every disposed addon leaves a "zombie" context that still counts toward Chromium's cap. When the cap is exceeded the browser force-evicts the *oldest* context, which may belong to a terminal that is still on screen, so a live terminal goes blank rather than the disposed one being cleaned up.

## Impact

Any embedder that creates and disposes many terminals in one document hits this — VS Code-style multi-terminal UIs, terminal multiplexers, IDEs with per-tab terminals. The failure is confusing because the terminal that breaks is not the one that was disposed, and it only appears after enough churn to cross the (driver-dependent) cap.

We currently work around it downstream by reaching into private addon internals (`addon._renderer._gl`) after `dispose()` and calling `WEBGL_lose_context.loseContext()` ourselves — which is exactly the kind of thing that should live in the addon.

## Suggested fix

Call `this._gl.getExtension('WEBGL_lose_context')?.loseContext()` as part of the renderer's dispose so the context is released immediately and the live count stays bounded. PR to follow.

openwong2kim:
Fixes #6068

## What

Explicitly lose the WebGL2 context when the WebGL renderer is disposed, by
calling `WEBGL_lose_context.loseContext()` in the renderer's teardown:

```ts
this._canvas.parentElement?.removeChild(this._canvas);
removeTerminalFromCache(this._terminal);
// Detaching the canvas does not free the underlying WebGL2 context; it
// lingers until it is garbage collected. Browsers cap the number of live
// WebGL contexts per document (Chromium ~16), so in apps that create and
// dispose many terminals these orphaned contexts accumulate and the
// browser force-evicts the oldest LIVE context, blanking an active
// terminal. Explicitly lose the context so it is released immediately.
this._gl.getExtension('WEBGL_lose_context')?.loseContext();
```

## Why

`WebglAddon.dispose()` detached the canvas but never released the underlying
WebGL2 context. It lingered until garbage collection, so every disposed addon
left a zombie context behind. Browsers cap the number of live WebGL contexts
per document (Chromium ~16, driver/build dependent); once an app creates and
disposes enough terminals to cross the cap, the browser force-evicts the
*oldest* context — which can belong to a terminal that is still on screen — and
that live terminal goes blank. See #6068 for a repro and full
analysis.

`loseContext()` is the standard, documented way to release a context
deterministically. `getExtension('WEBGL_lose_context')` is optional-chained so
a browser that does not expose the extension simply degrades to the previous
(leak-on-GC) behavior instead of throwing.

## Tests

Added `addons/addon-webgl/test/WebglContextRelease.test.ts`, an integration test
that spies on `WEBGL_lose_context.loseContext` and asserts `dispose()` invokes
it exactly once. The cap-eviction itself is GPU/driver dependent and does not
reproduce under headless SwiftShader, so the test verifies the mechanism
directly rather than trying to force an eviction. Verified it fails without this
change (loseContext called 0 times) and passes with it.

- `npm run tsc` — clean
- `eslint` on the changed files — clean
- New test passes on Chromium; the pre-existing shared-renderer color-comparison
  failures under headless SwiftShader are unaffected by this change (they fail
  identically on `master`).

### DIFF
--- a/addons/addon-webgl/src/WebglRenderer.ts
+++ b/addons/addon-webgl/src/WebglRenderer.ts
@@ -166,6 +166,13 @@ export class WebglRenderer extends Disposable implements IRenderer {
       }
       this._canvas.parentElement?.removeChild(this._canvas);
       removeTerminalFromCache(this._terminal);
+      // Detaching the canvas does not free the underlying WebGL2 context; it
+      // lingers until it is garbage collected. Browsers cap the number of live
+      // WebGL contexts per document (Chromium ~16), so in apps that create and
+      // dispose many terminals these orphaned contexts accumulate and the
+      // browser force-evicts the oldest LIVE context, blanking an active
+      // terminal. Explicitly lose the context so it is released immediately.
+      this._gl.getExtension('WEBGL_lose_context')?.loseContext();
     }));
   }
 
--- a/addons/addon-webgl/test/WebglContextRelease.test.ts
+++ b/addons/addon-webgl/test/WebglContextRelease.test.ts
@@ -1,0 +1,80 @@
+/**
+ * Copyright (c) 2025 The xterm.js authors. All rights reserved.
+ * @license MIT
+ */
+
+import test, { expect } from '@playwright/test';
+import { createTestContext, openTerminal } from '../../../test/playwright/TestUtils';
+import { platform } from 'os';
+
+test.describe('WebGL context release on dispose', () => {
+  // HACK: webgl2 is often not supported in headless firefox on Linux
+  // https://github.com/microsoft/playwright/issues/11566, also disable safari (#5852)
+  if (platform() === 'linux') {
+    test.skip(({ browserName }) => browserName === 'firefox' || browserName === 'webkit');
+  }
+
+  // Detaching the canvas on dispose does not free the underlying WebGL2
+  // context; it lingers until GC. Browsers cap the number of live contexts per
+  // document (Chromium ~16) and force-evict the oldest LIVE context when the cap
+  // is exceeded, blanking an active terminal. dispose() must explicitly release
+  // the context via WEBGL_lose_context.loseContext().
+  //
+  // The cap-eviction itself is GPU/driver dependent and does not reproduce under
+  // headless SwiftShader, so instead of asserting on eviction this verifies the
+  // mechanism directly: that dispose invokes loseContext() on the context.
+  test('dispose calls WEBGL_lose_context.loseContext() on the renderer context', async ({ browser }) => {
+    const ctx = await createTestContext(browser);
+    try {
+      await openTerminal(ctx);
+
+      const result = await ctx.page.evaluate(() => {
+        const w = window as any;
+
+        // Spy on loseContext for any WEBGL_lose_context extension handed out
+        // while this addon is alive.
+        const proto = w.WebGL2RenderingContext.prototype;
+        const originalGetExtension = proto.getExtension;
+        let loseContextCalls = 0;
+        proto.getExtension = function (name: string): any {
+          const ext = originalGetExtension.call(this, name);
+          if (name === 'WEBGL_lose_context' && ext && !ext.__spied) {
+            const originalLoseContext = ext.loseContext.bind(ext);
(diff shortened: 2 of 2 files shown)


savetheclocktower:
Yeah, this is a necessary fix. I work on [Pulsar's terminal package](https://github.com/pulsar-edit/terminal) and was able to demonstrate exactly what #6068 documented: the oldest contexts are targeted first, even if they're still being used.

Thanks for the contribution; I'll be employing my own version of it until this lands.

openwong2kim:
Glad someone else landed on this — that makes two independent repros; the other is an Obsidian terminal plugin, over in #6068.

Same origin story here: I hit it building an Electron terminal multiplexer, and #6068/#6069 came out of one long debugging run against 6.0.0. Several of the others from that run are probably in Pulsar's path too, so I'll link them rather than let you rediscover them the hard way.

**Teardown — same family as this PR**

- #6070 — document-level `mouseup`/`mousemove` listeners outlive `Terminal.dispose()`. Drag out of a terminal, dispose it mid-drag, release the button → `TypeError: Cannot read properties of undefined (reading 'dimensions')`. Very reachable in a tabbed terminal.
- #6064 (PR) — the shape underneath it: `RenderService.dimensions` throws after dispose and there are 20+ unguarded call sites. #5591/#5827 patched two of them per-callsite; this returns the last known dimensions instead, so a straggling rAF/timeout callback degrades to a stale number rather than a crash.

**CJK IME**

- #6065 / #6066 (PR) — with a Korean/Japanese/Chinese IME active, `Ctrl+<letter>` and Escape are silently dropped (everything arrives as keyCode 229). Ctrl+C not reaching the shell is hard to miss once you know to look.
- #6012 — the hidden textarea holds IME-committed text until blur; field-replacing input tools (voice IME, some mobile keyboards) then re-send it and corrupt the input.

**Unicode widths**

- #6079 / #6080 (PR) — `addon-unicode-graphemes` reads the trie header via `new DataView(data.buffer)`, ignoring `byteOffset`. Decode the payload into a pooled buffer and you get silently wrong widths instead of an error.

One triage note while you're in here: #6014 presents *identically* to #6068 — right cell geometry, wrong glyph contents, on a terminal nobody touched — but it was a shared-atlas bug, fixed by #6042/#6055. I burned real time conflating the two.

PR #6083, opened 2026-07-30
Fix Option+punctuation sending composed codepoints in the Kitty keyboard protocol

### LINKED ISSUE #5819: Kitty Keyboard Protocol: Option+<punctuation> doesn't work
- OS: macOS Tahoe
- Browser: macOS WKWebView
- Version: `@xterm/xterm` 6.1.0-beta.197

## Initial Observation

A keybinding such as `Option+]` did not appear to work as expected.

## Analysis

This is what I think is happening, but I'm not very experienced with xterm.js's internals.

`KittyKeyboard._getKeyCode` sends composed Unicode codepoints for Option+punctuation on macOS

When `macOptionIsMeta` is `true` and the Kitty keyboard protocol is active, pressing Option+punctuation keys (e.g. `]`, `[`, `;`, `/`, etc.) sends the macOS-composed Unicode codepoint instead of the base key with an Alt modifier.

For example, Option+`]` on a US layout sends `CSI 8217;3u` (codepoint of `'`, U+2019) instead of the correct `CSI 93;3u` (codepoint of `]`).

## Root cause

`_getKeyCode` in `KittyKeyboard.ts` uses `ev.code` to recover the base key when `macOptionAsAlt && ev.altKey`, but only handles two prefixes:

```typescript
if ((ev.shiftKey || (macOptionAsAlt && ev.altKey)) && ev.code) {
  if (ev.code.startsWith('Digit') && ev.code.length === 6) {
    // ...handles Digit0-Digit9
  }
  if (ev.code.startsWith('Key') && ev.code.length === 4) {
    // ...handles KeyA-KeyZ
  }
}
// Falls through to ev.key — the composed character
if (ev.key.length === 1) {
  return ev.key.codePointAt(0);  // returns e.g. 8217 instead of 93
}
```

Browser `code` values for punctuation keys (`BracketLeft`, `BracketRight`, `Semicolon`, `Quote`, `Backquote`, `Minus`, `Equal`, `Backslash`, `Comma`, `Period`, `Slash`, `Space`) are not mapped, so the method falls through and returns the codepoint from `ev.key` -- the macOS-composed character.

## Affected keys

| Key | `ev.code` | `ev.key` (composed) | Sent codepoint | Expected codepoint |
|-----|-----------|---------------------|----------------|-------------------|
| `]` | `BracketRight` | `'` (U+2019) | 8217 | 93 |
| `[` | `BracketLeft` | `\u201c` (U+201C) | 8220 | 91 |
| `;` | `Semicolon` | `\u2026` (U+2026) | 8230 | 59 |
| `/` | `Slash` | `\u00f7` (U+00F7) | 247 | 47 |
| etc. | | | | |

Option+letter keys (e.g. Option+A) work correctly because `KeyA` matches the `"Key"` prefix check.

## Suggested fix

This is the local patch I'm doing to make it work. Again, I'm not familiar with xterm.js's internals enough to know if this is the *correct* patch.

I suggest adding a code-to-codepoint map for the remaining standard keys, checked inside the existing `(ev.shiftKey || (macOptionAsAlt && ev.altKey)) && ev.code` block:

```typescript
const punctuationCodes: Record<string, number> = {
  'Minus': 45, 'Equal': 61, 'BracketLeft': 91, 'BracketRight': 93,
  'Backslash': 92, 'Semicolon': 59, 'Quote': 39, 'Backquote': 96,
  'Comma': 44, 'Period': 46, 'Slash': 47, 'Space': 32,
};
const punctCode = punctuationCodes[ev.code];
if (punctCode !== undefined) {
  return punctCode;
}
```

## Steps to reproduce

1. macOS, any terminal using xterm.js with `macOptionIsMeta: true` and Kitty keyboard protocol enabled
2. Press Option+`]` (or any punctuation key)
3. Observe: CSI u sequence contains the composed Unicode codepoint, not the base key

tarann26:
Fixes #5819.

On macOS with Option-as-Alt, `KittyKeyboard._getKeyCode` fell through to the Option-composed `ev.key` for punctuation keys, so Option+punctuation emitted the composed Unicode codepoint (e.g. Option+] → CSI 8217;3u for `'`) instead of the base key with the Alt modifier (CSI 93;3u for `]`).

The mac Option (and Shift) unwind gate already recovered the base codepoint from `ev.code` for `Digit*` and `Key*` codes but not for punctuation. This adds the remaining standard punctuation codes (Backquote, Minus, Equal, Bracket[Left|Right], Backslash, Semicolon, Quote, Comma, Period, Slash, Space) to that same gate so the base key is recovered.

Recovery is based on the US-QWERTY layout, consistent with the existing letter/digit handling.

### DIFF
--- a/src/common/input/KittyKeyboard.test.ts
+++ b/src/common/input/KittyKeyboard.test.ts
@@ -929,9 +929,65 @@ describe('KittyKeyboard', () => {
         assert.strictEqual(result.key, 'ƒ');
       });
 
-      it('falls through when ev.code is not Key*/Digit* (Opt+;)', () => {
+      it('Opt+; (key=…) → CSI 59;3 u', () => {
         const result = kitty.evaluate(createEvent({ key: '…', code: 'Semicolon', altKey: true }), flags, press, true);
-        assert.strictEqual(result.key, '\x1b[8230;3u');
+        assert.strictEqual(result.key, '\x1b[59;3u');
+      });
+
+      it('Opt+] → CSI 93;3 u', () => {
+        const result = kitty.evaluate(createEvent({ key: '’', code: 'BracketRight', altKey: true }), flags, press, true);
+        assert.strictEqual(result.key, '\x1b[93;3u');
+      });
+
+      it('Opt+[ → CSI 91;3 u', () => {
+        const result = kitty.evaluate(createEvent({ key: '“', code: 'BracketLeft', altKey: true }), flags, press, true);
+        assert.strictEqual(result.key, '\x1b[91;3u');
+      });
+
+      it('Opt+/ → CSI 47;3 u', () => {
+        const result = kitty.evaluate(createEvent({ key: '÷', code: 'Slash', altKey: true }), flags, press, true);
+        assert.strictEqual(result.key, '\x1b[47;3u');
+      });
+
+      it('Opt+- → CSI 45;3 u', () => {
+        const result = kitty.evaluate(createEvent({ key: '–', code: 'Minus', altKey: true }), flags, press, true);
+        assert.strictEqual(result.key, '\x1b[45;3u');
+      });
+
+      it('Opt+= → CSI 61;3 u', () => {
+        const result = kitty.evaluate(createEvent({ key: '≠', code: 'Equal', altKey: true }), flags, press, true);
+        assert.strictEqual(result.key, '\x1b[61;3u');
+      });
+
+      it('Opt+, → CSI 44;3 u', () => {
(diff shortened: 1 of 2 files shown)


jerch:
This cannot be merged for a simple reason: US-QWERTY is wrong for all other keyboard layouts. The kitty keyboard protocol is layout dependent (i wish it wasn't). I have started to map layouts here https://github.com/jerch/xkb-layouts and a proper fix needs to use a keyboard layout provider either from system (for electron) or driven by this layout lib (browser).

Issue #5713, opened 2026-02-16
Kitty graphics: Implement animation support (a=f, a=a, a=c)


anthonykim1:
## Context

The Kitty graphics protocol supports animated images through three actions: `a=f` (frame operations), `a=a` (animation control), and `a=c` (frame composition). None of these are implemented — they all return `EINVAL:unsupported action`. The related animation keys (frame number, gap, blend mode, etc.) are also not parsed.

## What needs to happen

1. **Parse animation keys** in `KittyGraphicsTypes.ts`:
   - Add to `KittyKey` enum and `IKittyCommand`:
     - Frame-related: frame number, source frame for copy, gap (ms between frames), blend mode, composition mode, background frame
   - Reference: https://sw.kovidgoyal.net/kitty/graphics-protocol/#control-data-reference

2. **Implement `a=f` (frame operations)**:
   - Load a new frame into an existing image
   - Edit an existing frame's data
   - Delete specific frames
   - Each frame has its own pixel data, gap timing, blend/composition mode

3. **Implement `a=a` (animation control)**:
   - Start/stop animation playback
   - Set number of loops
   - Jump to a specific frame
   - Set default gap timing

4. **Implement `a=c` (frame composition)**:
   - Compose/blend one frame onto another
   - Support composition modes: overwrite, alpha blend, etc.

5. **Animation rendering loop**:
   - Timer-based frame advancement using per-frame gap values
   - Re-render the image placement each time the active frame changes
   - Respect loop count

6. **Wire up delete selector `f/F`** (delete animation frames) — currently listed as unsupported in `_handleDelete()`, depends on this work

## Spec reference

https://sw.kovidgoyal.net/kitty/graphics-protocol/#animation

## Key files

- `addons/addon-image/src/kitty/KittyGraphicsTypes.ts` — add animation key parsing
- `addons/addon-image/src/kitty/KittyGraphicsHandler.ts` — TODOs at lines 329 and 363 for `a=f`, `a=a`, `a=c`
- `addons/addon-image/src/kitty/KittyImageStorage.ts` — needs per-image frame storage
- `addons/addon-image/src/ImageRenderer.ts` — needs timer-driven re-rendering for animation
- `addons/addon-image/test/KittyGraphics.test.ts` — TODO at line 108 for animation frame tests

## Acceptance criteria

- `a=f,i=3` loads a new animation frame into image 3
- `a=a,i=3,s=1` starts animation playback for image 3
[...]

Issue #5712, opened 2026-02-16
Kitty graphics: Implement relative placements (P/Q/H/V keys)


anthonykim1:
## Context

The Kitty graphics protocol supports "relative placements" — positioning an image relative to another already-placed image rather than at the current cursor position. This uses four keys: `P` (parent image ID), `Q` (parent placement ID), `H` (horizontal offset from parent in pixels), and `V` (vertical offset from parent in pixels).

Currently none of these keys are parsed or handled anywhere in the codebase.

## What needs to happen

1. **Parse relative placement keys** in `KittyGraphicsTypes.ts`:
   - `P` — parent image ID (the image to position relative to)
   - `Q` — parent placement ID (which placement of the parent image)
   - `H` — horizontal offset in pixels from parent's top-left corner
   - `V` — vertical offset in pixels from parent's top-left corner
   - Add these to the `KittyKey` enum and `IKittyCommand` interface

2. **Implement relative positioning** in `KittyGraphicsHandler._decodeAndDisplay()`:
   - When `P` is specified, look up the parent image's placement position in the buffer
   - Calculate the target position as parent position + (H, V) pixel offsets
   - Convert pixel offsets to cell coordinates for placement
   - If the parent placement is not found, respond with an error

3. **Track placement positions** in `KittyImageStorage`:
   - Need to store the screen position of each placement so child images can reference it
   - This likely requires a placement registry mapping (imageId, placementId) → (row, col, pixelOffsetX, pixelOffsetY)

## Spec reference

https://sw.kovidgoyal.net/kitty/graphics-protocol/#relative-placement

## Key files

- `addons/addon-image/src/kitty/KittyGraphicsTypes.ts` — add `P`, `Q`, `H`, `V` key parsing
- `addons/addon-image/src/kitty/KittyGraphicsHandler.ts` — `_decodeAndDisplay()` needs relative positioning logic
- `addons/addon-image/src/kitty/KittyImageStorage.ts` — needs placement position tracking
- `addons/addon-image/test/KittyGraphics.test.ts`

## Acceptance criteria

- `a=T,i=2,P=1,H=50,V=30` places image 2 at 50px right and 30px down from image 1's top-left
- `a=T,i=2,P=1,Q=3,H=0,V=0` places image 2 at the same position as placement 3 of image 1
- If parent image/placement not found, respond with appropriate error
- Negative `H`/`V` offsets work correctly (place to the left/above parent)
- Integration tests verify relative positioning accuracy

[...]

Issue #5711, opened 2026-02-16
Kitty graphics: Implement Unicode placeholder-based image display (U+10EEEE)


anthonykim1:
## Context

The Kitty graphics protocol supports "Unicode placeholders" — a mechanism where instead of the terminal deciding cell layout, the client writes U+10EEEE characters into terminal cells with specific combining diacritics to mark which cells belong to which image placement. This enables precise control over image-in-text layouts and is activated with `U=1` in the transmission/display command.

Currently there is zero implementation: the `U` key is not parsed in `KittyGraphicsTypes.ts`, U+10EEEE characters are not recognized anywhere, and there is only a TODO comment in the test file.

## What needs to happen

1. **Parse the `U` key** in `KittyGraphicsTypes.ts`:
   - Add `UNICODE_PLACEMENT = 'U'` to the `KittyKey` enum
   - Add `unicodePlacement?: number` to `IKittyCommand`
   - Wire it into `parseKittyCommand()`

2. **Handle `U=1` during transmit/display**:
   - When `U=1`, the terminal should NOT place the image at the cursor automatically
   - Instead, store the image and wait for U+10EEEE characters to appear in the cell buffer

3. **Recognize U+10EEEE in the terminal cell buffer**:
   - When a cell contains U+10EEEE (codepoint 0x10EEEE), treat it as an image placeholder
   - The combining diacritics following U+10EEEE encode: image ID (row diacritics), placement ID, row index, and column index
   - The renderer must decode these diacritics and render the corresponding image tile

4. **Diacritic encoding scheme** (per spec):
   - First diacritic: row of the image (0x0305 + row)
   - Second diacritic: column of the image (0x0305 + col)
   - Third+ diacritics: encode image ID and placement ID

## Spec reference

https://sw.kovidgoyal.net/kitty/graphics-protocol/#unicode-placeholders

## Key files

- `addons/addon-image/src/kitty/KittyGraphicsTypes.ts` — add `U` key parsing
- `addons/addon-image/src/kitty/KittyGraphicsHandler.ts` — handle `U=1` mode
- `addons/addon-image/src/ImageRenderer.ts` — render U+10EEEE placeholder cells
- `addons/addon-image/test/KittyGraphics.test.ts` — existing TODO at line 106

## Acceptance criteria

- `a=T,U=1,i=3,f=100;...` transmits and stores image but does NOT auto-place it
- Writing U+10EEEE with appropriate diacritics into cells renders the corresponding image tiles
[...]

Issue #5710, opened 2026-02-16
Kitty graphics: Implement remaining delete selectors and case-sensitive layer distinction


anthonykim1:
## Context

The Kitty graphics protocol `a=d` (delete) action supports many selectors for targeting specific images. Currently only `a/A` (delete all) and `i/I` (delete by image ID) are implemented. All other selectors are silently ignored (see `_handleDelete()` default branch).

Additionally, the spec distinguishes lowercase selectors (delete placements only, keep stored data) from uppercase (delete placements AND free stored image data). The current implementation treats both the same — it always frees everything. There is a TODO at line 277 of `KittyGraphicsHandler.ts` noting this.

**Note:** The current `d=a` implementation also deletes ALL images including offscreen stored data. Per spec, `d=a` should only delete "all placements visible on screen" — offscreen/stored-but-not-placed images should be preserved. This needs to be fixed as part of this issue.

## What needs to happen

1. **Fix `d=a` scope**: `d=a` should only delete placements visible on screen, not all stored image data. Offscreen images (stored via `a=t` but never placed, or scrolled out of view) should be preserved. `d=A` should delete visible placements AND free their stored image data.

2. **Implement missing delete selectors** in `KittyGraphicsHandler._handleDelete()`:
   - `c/C` — delete image at cursor position
   - `n/N` — delete images with ID &gt;= the specified ID
   - `p/P` — delete image by placement ID (requires `p` key + optionally `i` key)
   - `q/Q` — delete images whose placement intersects the cursor cell
   - `r/R` — delete images intersecting a row range (uses `x`/`y` keys for row bounds)
   - `x/X` — delete images intersecting a column range
   - `y/Y` — delete images at a specific cell position
   - `z/Z` — delete images at a specific z-index
   - `f/F` — delete animation frames (depends on animation support)

3. **Implement case distinction**:
   - Lowercase selector → delete the on-screen placement only, keep the stored image data in `_images` map so `a=p` can re-place it later
   - Uppercase selector → delete placement AND free stored image data from `_images`

4. **Add positional query support to `KittyImageStorage`**:
   - Currently only has `deleteById()` and `deleteAll()` — no way to query "which images intersect row 5?" or "which images are at cursor position?"
   - Needs methods like `deleteAtCursorPosition()`, `deleteByRowRange()`, `deleteByColumnRange()`, `deleteByZIndex()`, etc.

## Spec reference

[...]

PR #6098, opened 2026-08-11
addon-image: Fix Kitty placement lifecycle

mitchdenny:
Follow-up to #5707 and #5722. Partially addresses the `a/A` live-screen scope and
lower/uppercase lifetime requirements in #5710; the remaining delete selectors
from that issue are out of scope.

## Why

Kitty image data and placements have different lifetimes. One transmitted image
can have multiple displayed placements, and a non-zero placement ID identifies
one placement by `(imageId, placementId)`.

The current addon maps a Kitty image ID to one `ImageStorage` entry. Multiple
placements therefore overwrite shared bookkeeping, same-ID replacement can
leave old cells visible, and targeted deletion can remove shared image data.
It also decodes the transmitted source again for each placement.

## Behavior

| Command | Result |
|---|---|
| Repeated `a=p,i=N,p=P` | Replaces only named placement `(N,P)` |
| `a=p,i=N` or `p=0` | Adds an independent anonymous placement |
| `a=T` without `i=`, or with `i=0` | Ignores `p=`, creates an anonymous ID-zero placement, and releases its payload with its final placement |
| ID-less `a=t` | Discards the unaddressable payload after transmission |
| Failed `a=T` | Emits one error and does not reuse an older same-ID image or move the cursor |
| `d=i,i=N,p=P` | Deletes only `(N,P)` and preserves image data |
| `d=I,i=N,p=P` | Deletes only `(N,P)` and frees data only after a matching last placement |
| `d=i/I,i=N` | Deletes all placements for image N; uppercase frees unreferenced data |
| `d=a/A` | Deletes only active-buffer placements on the live application screen; uppercase frees only data left unreferenced |
| Retransmit `i=N` | Removes all old placements and replaces the transmitted data |

Explicit replacement and deletion clear their cell metadata so stale
placeholders cannot remain.

## Implementation

- Separates image-scoped payload/decoded-source state from named and anonymous
  placement records.
- Reuses one lazily decoded source bitmap for placement crop/scale operations.
- Indexes placement cells for targeted cleanup.
- Rebuilds cell indexes after resize/reflow and handles shifted cells.
- Rebuilds only candidate image-cell indexes from authoritative `HAS_EXTENDED`
  attributes on attached normal and alternate buffer lines before explicit
  deletion or named replacement. This drops detached line identities, discovers
  replacement live lines, repairs tile counts, and prevents stale tombstones.
- Reconciles Kitty placement indexes before `d=a/A` and deletes cell-less
  placements through the normal lifecycle path.
- Preserves placement tombstones and their render layer when raster quota
  eviction needs placeholder and later deletion semantics.
- Accounts decoded source pixels in storage limits. The hard cap remains 256
  transmitted images, with placements evicted together with capped payloads.
- Invalidates async placement work on reset, retransmit, delete, and dispose;
  stale bitmaps are closed before they can reach storage.
- Keeps ID-zero images under internal negative keys, validates client IDs as
  unsigned 32-bit values, and never exposes internal keys as reusable client IDs.
- Makes transmit-and-display transactional so transmission failure cannot fall
  through to an existing same-ID payload.
- Treats `HAS_EXTENDED` as authoritative in visibility/index scans, matching
  #6131. True Kitty placements that survive text overwrite require the separate
  placement store tracked in #6132.
- Extends the image demo with a single 10-second placement-rain smoke demo. It
  transmits one transparent teardrop sprite, animates a dense randomized field
  of named placements at independent pixel velocities using sub-cell offsets,
  targeted-deletes and reuses them, adds anonymous `p=0` placements, clears
  live-screen placements with `d=a`, proves payload reuse once more, and ends
  with uppercase cleanup.
- Adds a 15-second bouncing wireframe-cube demo that pre-renders and transmits 60
  transparent 200x200 PNG frames once, then moves/replaces the active placement
  and switches frames through lowercase hide/show operations before freeing all
  stored frame data.

## Performance

- Image bytes are transmitted once and source pixels are decoded once while the
  decoded source remains cached.
- Named replacement and targeted deletion rebuild only the requested storage
  IDs in a single pass over attached normal and alternate buffer lines. Index
  maps and column sets are allocated only for matching live cells.
- `d=a/A` scans the active buffer's live application screen for targets, then
  uses the same candidate-scoped reconciliation.
- Decoded source pixels are quota-accounted and evicted before visible placement
  rasters.

## Scope and compatibility

- No public API or typings change.
- Lowercase deletion intentionally changes to preserve transmitted image data,
  as required by the Kitty protocol.
- This does not implement the remaining `c/C`, `n/N`, `p/P`, `q/Q`, `r/R`,
  `x/X`, `y/Y`, `z/Z`, or `f/F` selectors tracked by #5710.
- The existing one-image-tile-per-cell model remains. Fully overlapping
  placements cannot yet stack and reveal an older image when a newer one is
  removed; image-on-image composition requires a broader renderer/buffer design.
- Relative placements (#5712), Unicode placeholders (#5711), and animation
  (#5713) remain out of scope.

## Validation

- `npm run build`
- `npm run esbuild`
- `npm run lint-changes`
- `npm run test-unit -- addons/addon-image/out-esbuild/kitty/KittyGraphicsTypes.test.js`
  - 23 passing
- `npm run test-integration -- --suite=addon-image`
  - 743 passing, 4 skipped across Chromium, Firefox, and WebKit
- `git diff --check`

The demo client/server bundles were built, and the Playwright suite exercised
the addon through its demo server. The placement-rain button was also exercised
headlessly through its full animation with no page errors. The wireframe-cube
button completed its preload, 15-second animation, and cleanup without page
errors or viewport scrolling.

Regression coverage includes named siblings, same-placement replacement,
independently cloned/replaced live buffer lines, bidirectional move sequences,
anonymous and ID-zero placements, targeted and live-screen deletion, scrolled-up
and alternate-buffer behavior, shared decode lifetime, crop fragments,
retransmission, quota pressure, negative-z placeholders, text overwrite,
detached lines, ICH/DCH shifts, resize/reflow, alternate-only index rebuilds,
acknowledgements, failed and empty transmit-and-display commands, invalid client
IDs, pending-upload deletion, and deterministic async
reset/retransmit/delete/dispose races.

Downstream validation in Hex1b covered the shared-payload placement lifecycle
and exposed the detached-line index invariant during repeated named-placement
movement. The regression and fix in `825f5605` are project-level xterm.js buffer
and storage correctness changes; a matched immutable browser build has been
prepared for downstream revalidation.

### DIFF
--- a/addons/addon-image/src/ImageStorage.ts
+++ b/addons/addon-image/src/ImageStorage.ts
@@ -273,6 +273,7 @@ export class ImageStorage implements IDisposable {
 
   public deleteImages(ids: Iterable<number>): number {
     const uniqueIds = new Set(ids);
+    this._rebuildImageCellIndex(uniqueIds);
     const clearedCells = this._clearImageCells(uniqueIds);
     for (const id of uniqueIds) {
       const spec = this._images.get(id);
@@ -313,88 +314,7 @@ export class ImageStorage implements IDisposable {
   }
 
   public reconcileImageCellIndexes(imageIds: Iterable<number>): void {
-    const uniqueIds = new Set(imageIds);
-    const lineImageIds = new Map<IBufferLineExt, Set<number>>();
-    const liveTileCounts = new Map<number, number>();
-    for (const imageId of uniqueIds) {
-      const lines = this._imageCells.get(imageId);
-      if (!lines) {
-        continue;
-      }
-      liveTileCounts.set(imageId, 0);
-      for (const line of lines.keys()) {
-        let ids = lineImageIds.get(line);
-        if (!ids) {
-          ids = new Set();
-          lineImageIds.set(line, ids);
-        }
-        ids.add(imageId);
-      }
-    }
-
-    const attachedLines = new Set<IBufferLineExt>();
-    const buffers = [this._terminal._core.buffers.normal, this._terminal._core.buffers.alt];
-    for (const buffer of buffers) {
-      for (let row = 0; row < buffer.lines.length; row++) {
-        const line = buffer.lines.get(row) as IBufferLineExt | undefined;
-        if (line && lineImageIds.has(line)) {
-          attachedLines.add(line);
-        }
-      }
-    }
-
-    for (const [line, imageIdsOnLine] of lineImageIds) {
-      if (!attachedLines.has(line)) {
-        for (const imageId of imageIdsOnLine) {
-          this._imageCells.get(imageId)?.delete(line);
-        }
-        continue;
-      }
-      const liveColumns = new Map<number, Set<number>>();
-      for (const key of Object.keys(line._extendedAttrs)) {
-        const column = Number(key);
-        if (
-          column >= line.length ||
-          !(line._data[column * Cell.SIZE + Cell.BG] & BgFlags.HAS_EXTENDED)
-        ) {
-          continue;
-        }
-        const imageId = line._extendedAttrs[column]?.imageId;
-        if (imageId === undefined || !imageIdsOnLine.has(imageId)) {
-          continue;
-        }
-        let columns = liveColumns.get(imageId);
-        if (!columns) {
-          columns = new Set();
-          liveColumns.set(imageId, columns);
-        }
-        columns.add(column);
-      }
-      for (const imageId of imageIdsOnLine) {
-        const lines = this._imageCells.get(imageId);
-        const columns = liveColumns.get(imageId);
-        if (columns?.size) {
-          lines?.set(line, columns);
-          liveTileCounts.set(imageId, (liveTileCounts.get(imageId) ?? 0) + columns.size);
-        } else {
-          lines?.delete(line);
-        }
-      }
-    }
-
-    for (const [imageId, tileCount] of liveTileCounts) {
-      const spec = this._images.get(imageId);
-      if (spec) {
-        spec.tileCount = tileCount;
-      }
-    }
-    const emptyImageIds = [...uniqueIds].filter(imageId => {
-      const lines = this._imageCells.get(imageId);
-      return !!lines && !lines.size;
-    });
-    if (emptyImageIds.length) {
-      this.deleteImages(emptyImageIds);
-    }
+    this._rebuildImageCellIndex(new Set(imageIds));
   }
 
   /**
@@ -915,14 +835,24 @@ export class ImageStorage implements IDisposable {
     return cleared;
   }
 
-  private _rebuildImageCellIndex(): void {
-    this._imageCells.clear();
-    for (const [id, spec] of this._images) {
-      this._imageCells.set(id, new Map());
-      spec.tileCount = 0;
+  private _rebuildImageCellIndex(imageIds?: ReadonlySet<number>): void {
+    const targetIds = imageIds ?? new Set([
+      ...this._images.keys(),
+      ...this._evictedImages.keys(),
+      ...this._imageCells.keys()
+    ]);
+    if (!imageIds) {
+      this._imageCells.clear();
     }
-    for (const id of this._evictedImages.keys()) {
-      this._imageCells.set(id, new Map());
+    for (const imageId of targetIds) {
+      const spec = this._images.get(imageId);
+      if (!spec && !this._evictedImages.has(imageId) && !this._imageCells.has(imageId)) {
+        continue;
+      }
+      this._imageCells.set(imageId, new Map());
+      if (spec) {
+        spec.tileCount = 0;
+      }
     }
     const buffers = [this._terminal._core.buffers.normal, this._terminal._core.buffers.alt];
     for (const buffer of buffers) {
@@ -941,7 +871,7 @@ export class ImageStorage implements IDisposable {
             continue;
           }
           const imageId = line._extendedAttrs[x]?.imageId;
-          if (imageId === undefined || !this._imageCells.has(imageId)) {
+          if (imageId === undefined || !targetIds.has(imageId) || !this._imageCells.has(imageId)) {
             continue;
           }
           this._trackCell(imageId, line, x);
@@ -952,9 +882,10 @@ export class ImageStorage implements IDisposable {
         }
       }
     }
-    const emptyImageIds = [...this._imageCells]
-      .filter(([, lines]) => !lines.size)
-      .map(([imageId]) => imageId);
+    const emptyImageIds = [...targetIds].filter(imageId => {
+      const lines = this._imageCells.get(imageId);
+      return !!lines && !lines.size;
+    });
     for (const imageId of emptyImageIds) {
       const marker = this._images.get(imageId)?.marker ?? this._evictedImages.get(imageId)?.marker;
       marker?.dispose();
--- a/addons/addon-image/src/ImageStorage.ts
+++ b/addons/addon-image/src/ImageStorage.ts
@@ -312,6 +312,91 @@ export class ImageStorage implements IDisposable {
     return result;
   }
 
+  public reconcileImageCellIndexes(imageIds: Iterable<number>): void {
+    const uniqueIds = new Set(imageIds);
+    const lineImageIds = new Map<IBufferLineExt, Set<number>>();
+    const liveTileCounts = new Map<number, number>();
+    for (const imageId of uniqueIds) {
+      const lines = this._imageCells.get(imageId);
+      if (!lines) {
+        continue;
+      }
+      liveTileCounts.set(imageId, 0);
+      for (const line of lines.keys()) {
+        let ids = lineImageIds.get(line);
+        if (!ids) {
+          ids = new Set();
+          lineImageIds.set(line, ids);
+        }
+        ids.add(imageId);
+      }
+    }
+
+    const attachedLines = new Set<IBufferLineExt>();
+    const buffers = [this._terminal._core.buffers.normal, this._terminal._core.buffers.alt];
+    for (const buffer of buffers) {
+      for (let row = 0; row < buffer.lines.length; row++) {
+        const line = buffer.lines.get(row) as IBufferLineExt | undefined;
+        if (line && lineImageIds.has(line)) {
+          attachedLines.add(line);
+        }
+      }
+    }
+
+    for (const [line, imageIdsOnLine] of lineImageIds) {
+      if (!attachedLines.has(line)) {
+        for (const imageId of imageIdsOnLine) {
+          this._imageCells.get(imageId)?.delete(line);
+        }
+        continue;
+      }
+      const liveColumns = new Map<number, Set<number>>();
+      for (const key of Object.keys(line._extendedAttrs)) {
+        const column = Number(key);
+        if (
+          column >= line.length ||
+          !(line._data[column * Cell.SIZE + Cell.BG] & BgFlags.HAS_EXTENDED)
+        ) {
+          continue;
+        }
+        const imageId = line._extendedAttrs[column]?.imageId;
+        if (imageId === undefined || !imageIdsOnLine.has(imageId)) {
+          continue;
+        }
+        let columns = liveColumns.get(imageId);
+        if (!columns) {
+          columns = new Set();
+          liveColumns.set(imageId, columns);
+        }
+        columns.add(column);
+      }
+      for (const imageId of imageIdsOnLine) {
+        const lines = this._imageCells.get(imageId);
+        const columns = liveColumns.get(imageId);
+        if (columns?.size) {
+          lines?.set(line, columns);
+          liveTileCounts.set(imageId, (liveTileCounts.get(imageId) ?? 0) + columns.size);
+        } else {
+          lines?.delete(line);
+        }
+      }
+    }
+
+    for (const [imageId, tileCount] of liveTileCounts) {
+      const spec = this._images.get(imageId);
+      if (spec) {
+        spec.tileCount = tileCount;
+      }
+    }
+    const emptyImageIds = [...uniqueIds].filter(imageId => {
+      const lines = this._imageCells.get(imageId);
+      return !!lines && !lines.size;
+    });
+    if (emptyImageIds.length) {
+      this.deleteImages(emptyImageIds);
+    }
+  }
+
   /**
    * Method to add an image to the storage.
    * @param img - The image to add (canvas or bitmap).
--- a/addons/addon-image/src/ImageStorage.ts
+++ b/addons/addon-image/src/ImageStorage.ts
@@ -18,6 +18,12 @@ export const CELL_SIZE_DEFAULT: ICellSize = {
   height: 14
 };
 
+export interface IImageStoragePixelCache {
+  readonly pixelCount: number;
+  evict(pixels: number): number;
+  clear(): void;
+}
+
 /**
  * Extend extended attribute to also hold image tile information.
  *
@@ -116,6 +122,9 @@ const EMPTY_ATTRS = new ExtendedAttrsImage();
 export class ImageStorage implements IDisposable {
   // storage
   private _images: Map<number, IImageSpec> = new Map();
+  private _evictedImages: Map<number, Pick<IImageSpec, 'marker' | 'bufferType'>> = new Map();
+  private _imageCells: Map<number, Map<IBufferLineExt, Set<number>>> = new Map();
+  private _pixelCaches: Set<IImageStoragePixelCache> = new Set();
   // last used id
   private _lastId = 0;
   // last evicted id
@@ -155,13 +164,28 @@ export class ImageStorage implements IDisposable {
   }
 
   public reset(): void {
+    for (const cache of this._pixelCaches) {
+      cache.clear();
+    }
     for (const spec of this._images.values()) {
+      spec.marker?.dispose();
+    }
+    for (const spec of this._evictedImages.values()) {
       spec.marker?.dispose();
     }
     // NOTE: marker.dispose above already calls ImageBitmap.close
     // therefore we can just wipe the map here
     this._images.clear();
+    this._evictedImages.clear();
+    this._imageCells.clear();
     this._renderer.clearAll();
+  }
+
+  public registerPixelCache(cache: IImageStoragePixelCache): IDisposable {
+    this._pixelCaches.add(cache);
+    return {
+      dispose: () => this._pixelCaches.delete(cache)
+    };
   }
 
   public getLimit(): number {
@@ -173,6 +197,10 @@ export class ImageStorage implements IDisposable {
       throw RangeError('invalid storageLimit, should be at least 0.5 MB and not exceed 1G');
     }
     this._pixelLimit = (value / 4 * 1000000) >>> 0;
+    this._evictOldest(0);
+  }
+
+  public enforceLimit(): void {
     this._evictOldest(0);
   }
 
@@ -190,15 +218,20 @@ export class ImageStorage implements IDisposable {
         }
       }
     }
+    for (const cache of this._pixelCaches) {
+      storedPixels += cache.pixelCount;
+    }
     return storedPixels;
   }
 
   private _delImg(id: number): void {
     const spec = this._images.get(id);
-    if (!spec) return;
+    if (!spec && !this._evictedImages.has(id) && !this._imageCells.has(id)) return;
     this._images.delete(id);
+    this._evictedImages.delete(id);
+    this._imageCells.delete(id);
     // FIXME: really ugly workaround to get bitmaps deallocated :(
-    if (window.ImageBitmap && spec.orig instanceof ImageBitmap) {
+    if (window.ImageBitmap && spec?.orig instanceof ImageBitmap) {
       spec.orig.close();
     }
     this.onImageDeleted?.(id);
@@ -216,6 +249,12 @@ export class ImageStorage implements IDisposable {
         zero.push(id);
       }
     }
+    for (const [id, spec] of this._evictedImages.entries()) {
+      if (spec.bufferType === 'alternate') {
+        spec.marker?.dispose();
+        zero.push(id);
+      }
+    }
     for (const id of zero) {
       this._delImg(id);
     }
@@ -228,12 +267,49 @@ export class ImageStorage implements IDisposable {
    * Delete an image by its internal storage ID.
    * Used by protocols that support explicit deletion (e.g. Kitty a=d).
    */
-  public deleteImage(id: number): void {
-    const spec = this._images.get(id);
-    if (spec) {
-      spec.marker?.dispose();
-      this._delImg(id);
+  public deleteImage(id: number): number {
+    return this.deleteImages([id]);
+  }
+
+  public deleteImages(ids: Iterable<number>): number {
+    const uniqueIds = new Set(ids);
+    const clearedCells = this._clearImageCells(uniqueIds);
+    for (const id of uniqueIds) {
+      const spec = this._images.get(id);
+      const evicted = this._evictedImages.get(id);
+      if (spec || evicted || this._imageCells.has(id)) {
+        (spec?.marker ?? evicted?.marker)?.dispose();
+        this._delImg(id);
+      }
     }
+    if (uniqueIds.size) {
+      this._needsFullClear = true;
+      this._fullyCleared = false;
+      this._terminal._core._inputHandler._dirtyRowTracker.markAllDirty();
+    }
+    return clearedCells;
+  }
+
+  public getVisibleImageStorageIds(): Set<number> {
+    const result = new Set<number>();
+    const buffer = this._terminal._core.buffer;
+    const end = Math.min(buffer.ydisp + this._terminal.rows, buffer.lines.length);
+    for (let row = buffer.ydisp; row < end; row++) {
+      const line = buffer.lines.get(row) as IBufferLineExt | undefined;
+      if (!line) {
+        continue;
+      }
+      for (let col = 0; col < this._terminal.cols; col++) {
+        if (!(line._data[col * Cell.SIZE + Cell.BG] & BgFlags.HAS_EXTENDED)) {
+          continue;
+        }
+        const imageId = line._extendedAttrs[col]?.imageId;
+        if (imageId !== undefined && imageId !== -1 && this._imageCells.has(imageId)) {
+          result.add(imageId);
+        }
+      }
+    }
+    return result;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


mitchdenny:
I spiced up the demo app to show how useful the ability retain and re-place images can be:

https://github.com/user-attachments/assets/2f58256f-cc75-43a2-bf8a-3eeb86f38824

The reason I started working on this is I was trying to hook up my .NET terminal emulator/UI stack project to xterm.js' KGP support and found a few glitches in my rendering which turned out to be incomplete implementation on the xterm.js side. Here is an example of xterm.js in action in a demo web-UX that hosts my TUI framework.

https://github.com/user-attachments/assets/fa16195d-02bc-4d87-a1d7-0ee09cb6d246

addons/addon-image/src/ImageStorage.ts
```diff
@@ -595,45 +671,194 @@ export class ImageStorage implements IDisposable {
...
+          const x = Number(key);
+          if (x >= line.length) {
+            delete line._extendedAttrs[x];
+            continue;
+          }
+          const imageId = line._extendedAttrs[x]?.imageId;
+          if (imageId === undefined || !this._imageCells.has(imageId)) {
+            continue;
+          }
+          this._trackCell(imageId, line, x);
```
anthonykim1: These new paths treat stale `_extendedAttrs` entries as authoritative placement data, which now conflicts with #6131.

- Both this loop and `getVisibleImageStorageIds` read `line._extendedAttrs[col]` without checking `HAS_EXTENDED`. Per `BufferLine.ts`, that map is a sparse cache that is only valid while the flag is set. Entries left after text cleared the flag are stale cache, and here they drive deletion scope, `tileCount`, and image disposal.
- #6131 (just merged) removed the flagless render fallback, so master now disagrees. A rebase won't surface it: `ImageStorage.ts` auto-merges with no conflict markers, and the only signal will be [KittyGraphics.test.ts:1517-1550](https://github.com/xtermjs/xterm.js/blob/d153c858b7f6bd1f1555f4c471fa2d3444dfa8ab/addons/addon-image/test/KittyGraphics.test.ts#L1517-L1550) failing (it asserts the old overwrite-survival behavior and needs rewriting anyway).
- If we want placements that truly survive text (what kitty actually does), that's the separate placement store tracked in #6132, not stale cell metadata.

addons/addon-image/src/ImageStorage.ts
```diff
@@ -228,12 +267,47 @@ export class ImageStorage implements IDisposable {
...
+      this._terminal._core._inputHandler._dirtyRowTracker.markAllDirty();
+    }
+    return clearedCells;
+  }
+
+  public getVisibleImageStorageIds(): Set<number> {
+    const result = new Set<number>();
+    const buffer = this._terminal._core.buffer;
+    const end = Math.min(buffer.ydisp + this._terminal.rows, buffer.lines.length);
+    for (let row = buffer.ydisp; row < end; row++) {
```
anthonykim1: `d=a` should target the live application screen (`ybase`), not the user's scroll viewport (`ydisp`).

- kitty filters `d=a` purely on live-screen coordinates ([`clear_filter_func_noncell`, graphics.c L2354-L2358](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/kitty/graphics.c#L2354-L2358)); the user's scrollback view never reaches delete handling.
- With the user scrolled up, this scan deletes the viewed history and misses the actual screen.
- Potential fix: anchor at `ybase` (this function's only caller is the kitty delete path, so no collateral) and add a `ydisp < ybase` test.

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -661,7 +671,10 @@ export class KittyGraphicsHandler implements IApcHandler, IResetHandler, IDispos
       }
 
       const zIndex = cmd.zIndex ?? 0;
-      this._kittyStorage.addImage(image.id, bitmap, true, layer, zIndex);
+      if (!this._kittyStorage.isPlacementOperationCurrent(operationGeneration, image)) {
+        throw new Error('image placement was invalidated');
+      }
+      this._kittyStorage.addImage(image.id, cmd.placementId ?? 0, bitmap, true, layer, zIndex);
```
anthonykim1, line 696: `a=T,i=31,p=42` is acked as `Gi=31;OK`, but the spec wants `Gi=31,p=42;OK`.

- The placement is stored under `cmd.placementId` here, but the `a=T` ack in `_handleTransmitDisplay` ([line 439](https://github.com/xtermjs/xterm.js/blob/d153c858b7f6bd1f1555f4c471fa2d3444dfa8ab/addons/addon-image/src/kitty/KittyGraphicsHandler.ts#L439)) never passes it to `_sendResponse`. Spec: the placement id "will be added to the acknowledgement code" ([rst L487-L488](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L487-L488)).
- Potential fix: pass `cmd.placementId` through to `_sendResponse`; it already takes the param and the `a=p` path passes it. Worth an `a=T,...,p=42` response assertion too; the only `p`-echo test today is the standalone `a=p` one.
- To be fair, this bug is pre-existing (the `a=T` ack path predates this PR); flagging it since it bites the placement lifecycle this PR builds.

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -500,28 +500,33 @@ export class KittyGraphicsHandler implements IApcHandler, IResetHandler, IDispos
...
       case 'i':
+        if (cmd.id !== undefined) {
+          const pending = this._pendingTransmissions.get(cmd.id);
+          if (pending) {
+            pending.decoder.release();
+          }
+          this._removePendingEntry(cmd.id);
+          this._kittyStorage.deletePlacements(cmd.id, cmd.placementId);
+        }
+        break;
```
anthonykim1, line 545: Any delete command must abort the pending upload, not just `a/A` (all) and `i/I` (matching id).

- Spec: "when **any** delete command is received, the partial upload must be aborted" ([rst L789-L790](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L789-L790)); kitty does it as the first statement of `handle_delete_command`, before looking at the selector.
- Potential fix: move the cleanup to the top of `_handleDelete` and drop the per-selector logic. (Remaining delete selectors are tracked in #5710.)
- The two-concurrent-uploads test at [KittyGraphics.test.ts:385-398](https://github.com/xtermjs/xterm.js/blob/d153c858b7f6bd1f1555f4c471fa2d3444dfa8ab/addons/addon-image/test/KittyGraphics.test.ts#L385-L398) codifies a state a conforming client can't produce, since chunked uploads can't interleave ([rst L405-L410](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L405-L410)); consider replacing it with "any `a=d` clears the pending upload".

addons/addon-image/src/ImageStorage.ts
```diff
@@ -595,45 +671,194 @@ export class ImageStorage implements IDisposable {
...
         if (spec.actual && spec.orig !== spec.actual) {
           current -= spec.actual.width * spec.actual.height;
         }
-        spec.marker?.dispose();
-        this._delImg(this._lowestId);
+        this._images.delete(this._lowestId);
+        this._evictedImages.set(this._lowestId, {
+          marker: spec.marker,
+          bufferType: spec.bufferType
+        });
```
anthonykim1, line 621: Evicting a `z<0` placement with `showPlaceholder: true` puts its checkerboard *above* the text.

- The tombstone keeps only `marker`/`bufferType`, dropping the placement's `layer`/`zIndex`, and placeholders always draw on the top canvas, inverting the spec's "negative z-index … drawn under the text" ([rst L543-L544](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L543-L544)).
- Potential fix: carry enough layer info in the tombstone to keep the placeholder underneath, or skip placeholders for bottom-layer evictions; a negative-z eviction test would cover it.
- Background on this placeholder pattern: #5731.

addons/addon-image/src/kitty/KittyImageStorage.ts
```diff
@@ -53,102 +75,301 @@ export class KittyImageStorage implements IDisposable {
...
+    }
+  }
+
+  public deleteImage(kittyId: number, placementId?: number): void {
+    this._invalidatePlacementOperations();
+    this._deletePlacements(kittyId, placementId);
+    if (!this._hasPlacements(kittyId)) {
+      this._deleteImageData(kittyId);
+    }
+  }
```
anthonykim1, line 267: Non-blocking: `d=I,i=N,p=<unmatched>` frees image data kitty would keep.

- An unmatched `P` is a no-op in `_deletePlacements`, then `_hasPlacements` is false, so the payload is freed. kitty frees an uppercase-targeted image only when a ref actually matched, and its no-placements special case requires `p` to be absent ([graphics.c L2208](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/kitty/graphics.c#L2208), [L2424-L2426](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/kitty/graphics.c#L2424-L2426)).
- Spec prose doesn't cover this case, so not a hill to die on, but matching kitty avoids a surprise `ENOENT` on the next `a=p,i=N`.

addons/addon-image/src/kitty/KittyImageStorage.ts
```diff
@@ -53,102 +75,301 @@ export class KittyImageStorage implements IDisposable {
...
-    if (!this._images.has(imageId) && this._images.size >= KittyImageStorage._maxStoredImages) {
+    if (this._images.size >= Constants.MAX_STORED_IMAGES) {
       this._evictUndisplayedImages();
+      while (this._images.size >= Constants.MAX_STORED_IMAGES) {
+        const oldestImageId = this._images.keys().next().value;
+        if (oldestImageId === undefined) {
+          break;
+        }
+        this._deleteImageData(oldestImageId);
+      }
```
anthonykim1, line 126: Non-blocking: hard-cap eviction keeps the pixels on screen but kills the id.

- The 256-image cap frees the payload while the placement stays visible, so a later `a=p,i=<id>` gets `ENOENT`. kitty evicts image + refs together ([`apply_storage_quota`](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/kitty/graphics.c#L308-L325)), though the spec only prescribes a byte quota ([rst L1038-L1042](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L1038-L1042)), so a count cap has no exact analog.
- If keeping pixels alive is intentional (the [test at 1717-1737](https://github.com/xtermjs/xterm.js/blob/d153c858b7f6bd1f1555f4c471fa2d3444dfa8ab/addons/addon-image/test/KittyGraphics.test.ts#L1717-L1737) suggests so), a short comment documenting the tradeoff would do; otherwise consider evicting the placements too.

anthonykim1:
Thanks for the PR! @mitchdenny 
Also very cool demo https://github.com/xtermjs/xterm.js/pull/6098#issuecomment-5253535207 🔥!

 It would be really cool for us to eventually support animation from kitty as well :) 

Placement feels challenging atm because kitty shares the cell-backed storage with sixel/IIP (where #6131 made text overwrite the rule, see #5860) and truly independent placements need a separate store (#6132). We can consider rebase, keep the lifecycle work, and leave text-survival to #6132 ?

anthonykim1:
@jerch  Could you take a look when you get a chance as well 🙏 especially changes to `ImageStorage.ts`

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
mitchdenny: Moved `_cleanupAllPending()` to the start of `_handleDelete`, before selector handling, and removed the per-selector cleanup. Replaced the interleaved-upload expectation with a regression proving even an unsupported delete selector aborts the active partial upload.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
mitchdenny: Passed `cmd.placementId` through every `a=T` acknowledgement path, including successful display, decode failure, unsupported transmission medium, and rendering failure. Added success and error response assertions for `p=`.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

addons/addon-image/src/ImageStorage.ts
mitchdenny: Changed the delete-all discovery range to `ybase..ybase+rows`. The regression explicitly scrolls the viewport into history (`ydisp < ybase`) and verifies `d=a/A` delete the live-screen placement while retaining the historical placement.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

addons/addon-image/src/ImageStorage.ts
mitchdenny: Rebased onto master with #6131 and removed the flagless-cell behavior. `HAS_EXTENDED` is authoritative again in visibility discovery, cell replacement/untracking, deletion cleanup, and index rebuilding, and the stale text-survival regression is gone. True text-surviving Kitty placements remain scoped to #6132.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

addons/addon-image/src/kitty/KittyImageStorage.ts
mitchdenny: Changed the hard-cap behavior rather than documenting the inconsistency: eviction now removes all placements for the oldest image before deleting its payload. The regression verifies both the placement map and buffer cell are cleared with the image ID.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

addons/addon-image/src/ImageStorage.ts
mitchdenny: Eviction tombstones now retain their image layer, layer discovery includes tombstones, and placeholders draw on the retained layer. Added a quota-eviction regression verifying a negative-z placeholder is opaque on the bottom canvas and absent from the top canvas.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

addons/addon-image/src/kitty/KittyImageStorage.ts
mitchdenny: Fixed this by making targeted placement deletion report whether it matched. `d=I,...,p=` now returns without freeing image data when the placement is absent, with a regression that reuses the retained payload afterward.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -403,7 +403,7 @@ export class KittyGraphicsHandler implements IApcHandler, IResetHandler, IDispos
     const transmission = cmd.transmission ?? 'd';
     if (transmission !== 'd') {
       if (cmd.id !== undefined) {
-        this._sendResponse(cmd.id, 'EINVAL:unsupported transmission medium', cmd.quiet ?? 0);
+        this._sendResponse(cmd.id, 'EINVAL:unsupported transmission medium', cmd.quiet ?? 0, cmd.placementId);
```
anthonykim1, line 425: `a=T` can still send this `EINVAL`, then display an older image with the same id and send `OK`.

- The Kitty spec defines `a=T` as one transmit-and-display operation ([rst L465-L478](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L465-L478)) and says terminals must not display anything until a chunked sequence has been received and validated ([rst L405-L412](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L405-L412)).
- `_handleTransmitDisplay` ignores `_handleTransmit`'s result and continues into [`getImage(id)`](https://github.com/xtermjs/xterm.js/blob/5b65c03690770673f407931c767f72ec908dce2c/addons/addon-image/src/kitty/KittyGraphicsHandler.ts#L431-L439). If `i=N` already exists, `a=T,t=f,i=N,p=P;<path>` sends `EINVAL`, places the old image, then sends `OK`.
- The new test uses a fresh id (`310`) and only checks `startsWith`, so it cannot catch the stale placement or second response.
- Potential fix: make `_handleTransmit` report success/failure and stop here on failure. A same-id regression should assert exactly one `EINVAL`, no new placement, and no cursor movement.
- This flow is pre-existing; flagging it here because this commit changes the `a=T` response paths and adds coverage for this error case.

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
```diff
@@ -661,7 +661,10 @@ export class KittyGraphicsHandler implements IApcHandler, IResetHandler, IDispos
       }
 
       const zIndex = cmd.zIndex ?? 0;
-      this._kittyStorage.addImage(image.id, bitmap, true, layer, zIndex);
+      if (!this._kittyStorage.isPlacementOperationCurrent(operationGeneration, image)) {
+        throw new Error('image placement was invalidated');
+      }
+      this._kittyStorage.addImage(image.id, cmd.placementId ?? 0, bitmap, true, layer, zIndex);
```
anthonykim1, line 696: **(new from this PR, commit [`d6f791a`](https://github.com/xtermjs/xterm.js/commit/d6f791a143cc5962a8bc0c52591232a34ac66e8b))**

Images transmitted without `i=` lose their ID-0 semantics here.

- The Kitty spec says `p=` must be ignored when the image has no id; the resulting placement remains `(image id=0, placement id=0)` ([rst L485-L494](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L485-L494)).
- `storeImage(undefined)` assigns a synthetic internal id, then this passes `cmd.placementId`, so `a=T,p=7` becomes a named placement under that synthetic id.
- The documented reuse path is `a=p,i=N`, using an id supplied during transmission ([rst L465-L470](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L465-L470)). After the final ID-0 placement disappears, this client has no id with which to reuse the retained payload. Kitty removes such unreferenced ID-0 images ([graphics.c L2197-L2209](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/kitty/graphics.c#L2197-L2209)).
- Base deleted the payload with the shared-storage entry.
- Potential fix: preserve whether the client supplied `i=`, force `p=0` when it did not, and free an ID-0 payload when its last placement disappears. A regression with id-less `a=T,p=7` followed by lowercase deletion would cover both cases.

addons/addon-image/src/kitty/KittyImageStorage.ts
```diff
@@ -53,102 +75,307 @@ export class KittyImageStorage implements IDisposable {
...
+  public deletePlacements(kittyId: number, placementId?: number): void {
+    this._invalidatePlacementOperations();
+    this._deletePlacements(kittyId, placementId);
+  }
+
+  public deleteVisiblePlacements(freeData: boolean): void {
+    this._invalidatePlacementOperations();
+    const storageIds = new Set<number>();
+    const kittyIds = new Set<number>();
+    for (const storageId of this._storage.getVisibleImageStorageIds()) {
```
anthonykim1, line 239: **(new from this PR, commit [`d6f791a`](https://github.com/xtermjs/xterm.js/commit/d6f791a143cc5962a8bc0c52591232a34ac66e8b))**

Text overwrite can leave a placement and its payload in the maps with no live cells through which `d=A` can discover them.

- Kitty models each display as a placement ([rst L465-L470](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L465-L470)) with its own z-order; negative-z placements are drawn under text rather than deleted by it ([rst L540-L548](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L540-L548)). Full independent placement rendering is already deferred to #6132.
- In the current cell-backed fallback, #6131 makes text remove the live tiles. But ordinary text writes do not call `_untrackCell` or `_handleStorageImageDeleted`, so `_storageIdToPlacement` still contains the placement.
- `d=A` is defined to delete visible placements and free their image data once no placements remain ([rst L756-L770](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L756-L770), [L784-L787](https://github.com/kovidgoyal/kitty/blob/7f71461a9de44f71f631db29d50bcb3e866988b8/docs/graphics-protocol.rst?plain=1#L784-L787)). Once every cell has been overwritten, this scan returns nothing, so neither the stale placement mapping nor the payload is cleaned up.
- Base used `deleteAll()` here, so this retained state comes from the new visible-placement deletion path.
- This does not require restoring flagless rendering or implementing #6132. Could we reconcile placements with no remaining live `HAS_EXTENDED` cells and add a `place -> overwrite all cells -> d=A` regression?

mitchdenny: Fixed in c34b04b5. Before `d=a/A` selection, `ImageStorage` now reconciles only the indexed placement lines against attached normal/alternate buffer lines and authoritative `HAS_EXTENDED` cells, repairing shifted columns/tile counts and deleting cell-less storage entries through the normal deletion callback. `d=A` snapshots the associated Kitty IDs first so payloads are freed once reconciliation removes the last placement. Regressions cover full text overwrite and a detached line created by `CSI M`; both leave no placement, storage entry, or payload after `d=A`.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
mitchdenny: Fixed in c34b04b5. Omitted `i=` and `i=0` now use anonymous ID-zero semantics: their internal storage keys live in a negative namespace, `p=` is forced to zero, and unreferenced payloads are released after placement deletion, failed display, or transmit-only use. Client IDs are validated as unsigned 32-bit values so they cannot collide with internal keys. The regression covers both omitted `i=` and explicit `i=0`, verifies anonymous placement tracking, then confirms lowercase deletion removes the placement and payload.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

addons/addon-image/src/kitty/KittyGraphicsHandler.ts
mitchdenny: Fixed in c34b04b5. `_handleTransmit` now returns the stored image ID or `undefined`, and `a=T` stops immediately when transmission fails instead of looking up an older same-ID image. The regression seeds `i=310`, sends a failing `a=T,t=f,i=310,p=43`, and asserts one exact `EINVAL`, no placement, retained source data, and unchanged cursor position. Empty `a=T` payloads now also receive `EINVAL` rather than falling through silently.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

mitchdenny:
Follow-up fix pushed in `825f5605`: before clearing an internal image storage ID, `ImageStorage` now rebuilds that candidate index from authoritative attached normal and alternate buffer cells. This handles legitimate buffer-line replacement/reflow where `_imageCells` still points at detached line objects, preventing named `(imageId, placementId)` replacement from leaving stale storage IDs/tombstone placeholders.

The regression independently clones live line attributes, reuses the same placement without retransmission across repeated bidirectional moves, and checks after every move that live/stored/indexed/placement ID sets are identical singletons, live cell count equals `tileCount`, every indexed line is attached, and historical IDs do not accumulate. The rebuild is candidate-scoped and a single pass across attached buffers; decoded image data remains shared/reused.

Validation: `npm run build`, `npm run lint-changes`, parser unit tests (23 passing), and addon-image Playwright across Chromium/Firefox/WebKit (743 passing, 4 skipped).

mitchdenny:
@jerch any chance you could take a look? Looking forward to getting KGP support on the path to full compliance - I've got a vendored version of this PR working in the Aspire repo and it's working pretty well. Once/if this goes in then we can look at KGP animation support.

jerch:
@mitchdenny Tbh I am not very happy how the kitty graphics protocol does not integrate well with the rest of the image addon currently. It duplicates tons of work and does things in the handler that should be storage or renderer responsibility. Furthermore it misuses the cell-based placements currently done in `ImageStorage.addImage`, which is clearly not how it is meant to be done with kitty graphics (I think this also applies to IIP, have not tested yet).

I already started #6145 to solve the css vs. device pixel issue and doing a basic reshape of internal storage to renderer abstraction. The next step regarding KGP would be to consolidate things, so the handler does only sequence parsing and data preparing, while the storage holds all bits for ID based image storing (I also need that for the planned serialization) and dedicated placement rules. Things like rescaling, cropping and even layered composing output should be the renderer's responsibility in a form of "painting operations" always drawing from a single source of truth - the original image data only.

I don't want to derail your efforts here but I currently shy away from adding another +700 LOC in the wrong places. We should get the ground work done first.

mitchdenny:
Thanks for the candid architectural feedback. That direction makes sense, and we don't want to keep adding substantial placement logic in layers that are expected to change. We'll close this in favor of the image-addon groundwork in #6145 and the follow-up storage/placement redesign, keep an eye on that sequence of work, and look for opportunities to help once the intended extension points are clearer.

PR #6137, opened 2026-08-27
Add an experimental WebGPU renderer addon

wangdiandao:
## Summary

This PR adds `@xterm/addon-webgpu`, an experimental WebGPU renderer addon for xterm.js.

The goal of this iteration is functional completeness against the current renderer behavior and shared renderer test coverage. It is **not** presented as a replacement for `@xterm/addon-webgl`, and it does **not** claim a clear performance advantage at this stage.

Part of #4552.

## Design

The addon follows the existing renderer-addon model while accounting for WebGPU's asynchronous initialization and device lifecycle:

- The existing renderer remains active until adapter/device acquisition and renderer initialization complete successfully.
- `ready`, `onRendererError`, and `onDeviceLoss` expose initialization and runtime state to embedders.
- Initialization failure leaves the current renderer untouched.
- Unexpected device loss restores xterm.js's default renderer. Retrying requires loading a new addon instance.
- WebGL and WebGPU are mutually exclusive in the demo.

The renderer is split into focused paths:

- `RectangleRenderer` handles cell backgrounds and rectangle-based terminal decorations.
- `GlyphRenderer` uses packed per-cell GPU records, WGSL shaders, render bundles, and a multi-page glyph atlas.
- Dirty-cell tracking uploads only planned cell-buffer ranges instead of rewriting the full viewport on every frame.
- Atlas changes retain bounded dirty-rectangle history so independent renderers sharing an atlas can upload changes safely.
- Atlas uploads choose between partial rectangles, a bounding rectangle, and a full upload based on estimated copy cost.
- Atlas page limits, merging, overflow glyphs, and eviction are covered explicitly.
- Custom box-drawing, block, Braille, Powerline, progress, and legacy-computing glyphs are supported.
- Link rendering remains a DOM overlay, consistent with the existing renderer architecture.

The addon is wired into the demo, build, packaging, publishing, artifact upload, and Chromium integration-test paths.

## Experimental status and boundaries

`@xterm/addon-webgpu` should be treated as experimental:

- The current target is functional completeness, not proven performance leadership over WebGL.
- WebGPU availability depends on a secure context, browser support, GPU adapter availability, drivers, and platform policy.
- CI invokes the addon suite only in the Chromium matrix using Playwright's bundled Chromium. Broad browser/OS/GPU-driver coverage has not been established yet.
- The implementation currently integrates through xterm.js renderer internals, so internal renderer changes may require corresponding addon updates.
- Device loss falls back safely, but the addon does not recreate a device automatically.
- This PR does not claim lower power consumption, lower memory usage, or better performance across machines.
- WebGL remains the mature accelerated renderer and the appropriate default while this addon is evaluated.

## Performance observations

I ran a deterministic local comparison using identical workloads and viewport settings for DOM, WebGL, and WebGPU. Each renderer was warmed up first; execution order was reversed on alternating iterations. The table below reports the median of 5 runs per renderer for 3,000 operations with batches of 50.

Environment: Windows 10, headless stable Chrome 151, NVIDIA WebGPU adapter (`blackwell` architecture as reported by WebGPU), 32 logical CPUs, 32 GiB device memory, DPR 1, 1280×720 browser viewport, approximately 55×22 terminal cells.

Throughput is processed input bytes divided by terminal write/parse time. Renderer CPU is cumulative renderer callback CPU time for one scenario run; it is not GPU execution time.

| Scenario | WebGL MB/s | WebGPU MB/s | WebGPU throughput delta | WebGL renderer CPU | WebGPU renderer CPU |
| --- | ---: | ---: | ---: | ---: | ---: |
| Interactive shell | 1.449 | 1.455 | +0.4% | 8.1 ms | 10.5 ms |
| Build output | 0.959 | 0.946 | -1.4% | 7.4 ms | 9.3 ms |
| Streaming service logs | 1.169 | 1.157 | -1.0% | 6.0 ms | 7.2 ms |
| Unicode / emoji | 0.987 | 0.970 | -1.7% | 10.1 ms | 11.9 ms |
| Live dashboard updates | 1.295 | 1.286 | -0.7% | 6.8 ms | 8.5 ms |
| Full-screen editor updates | 0.529 | 0.525 | -0.8% | 6.7 ms | 9.0 ms |
| SSH reconnect/history burst | 44.474 | 45.138 | +1.5% | 0.3 ms | 0.4 ms |
| Multi-job progress updates | 0.582 | 0.574 | -1.4% | 3.5 ms | 5.2 ms |

In this run, WebGPU throughput ranged from 1.7% below to 1.5% above WebGL, while cumulative renderer CPU time was higher for WebGPU in all eight scenarios. WebGPU's final GPU-idle wait was generally lower (roughly 27%–42%), but that did not translate into a consistent end-to-end throughput advantage. Both accelerated renderers had similar maximum sampled frame intervals (about 6.1–6.2 ms) and no observed long tasks.

These results should be read as a snapshot from one machine, not a general benchmark claim. They support the current positioning: the addon is functionally useful and provides a WebGPU implementation for further evaluation, but performance has not separated clearly from WebGL yet.

## Validation

- TypeScript build and esbuild bundles pass.
- Full lint and changed-file lint pass.
- WebGPU unit tests: 40 passing.
- Local stable-Chrome WebGPU integration run: 67 passing, 9 skipped by shared renderer coverage.
- CI is configured to invoke the same addon suite in its bundled-Chromium job; the PR check remains the source of truth for that runner environment.
- Integration coverage includes initialization failure, dispose races, device loss, renderer behavior, custom glyph adoption, dirty uploads, cursor uploads, shared-atlas invalidation, atlas overflow, and atlas stress.
- The demo can switch between WebGL and WebGPU without leaving both renderers active.

## Follow-up work

- Expand OS, browser, GPU-vendor, and driver coverage.
- Profile renderer CPU overhead and reduce WebGPU-side bookkeeping.
- Add reproducible performance tracking in CI where stable GPU-backed runners are available.
- Investigate memory, power, and multi-terminal behavior independently from throughput.
- Revisit API stability and fallback/recovery policy based on maintainer feedback.


daonb:
Thanks for your work. My terminal regularly runs out of canvases so my two cents are to fold it in.

